PR #62, opened 2022-01-12
proof of work based on bitcoin

b-yap:
~~This will depend on https://github.com/mintlayer/mintlayer-core/pull/70, so this will not actually compile just yet.~~
proof of work based on bitcoin core.

This is not for merging to master branch, but for discussion purposes.


block_production/src/lib.rs
```diff
@@ -19,25 +20,30 @@ pub trait Chain {
...
+
 impl From<BlockCreationError> for BlockProductionError {
     fn from(e: BlockCreationError) -> Self {
         BlockProductionError::BlockCreationError(e)
     }
 }
 
-#[derive(PartialEq, Eq, Clone, Debug)]
+#[derive(PartialEq, Eq, Debug)]
 pub enum ConsensusParams {
```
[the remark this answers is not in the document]
b-yap: ^ forget my previous comment. This is already addressed by the netupgrades: https://github.com/mintlayer/mintlayer-core/pull/70

common/src/primitives/id.rs
```diff
@@ -29,6 +30,21 @@ impl From<GenericArray<u8, typenum::U32>> for H256 {
...
+    fn reverse_inner(&self) -> [u8; 32] {
+        let mut h256_inner = self.0;
+        h256_inner.reverse();
+        h256_inner
+    }
+}
+
+impl From<H256> for Uint256 {
+    fn from(hash: H256) -> Self {
+        let x = hash.reverse_inner();
```
TheQuantumPhysicist, line 43: This isn't an operation I'm particularly happy about and I think we should discuss the reason why this is necessary. Since we're starting from scratch, let's try to make both H256 and Uint256 use the same order. No need to reverse things every time. This can get very confusing and error-prone very quickly.

TheQuantumPhysicist: I can see that H256 uses big-endian order... we may have to fix this...

consensus/src/pow/config.rs
```diff
@@ -0,0 +1,146 @@
...
+        ]),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use common::chain::{NetUpgradeError, UpgradeVersion};
+
+    impl TryFrom<(UpgradeVersion, ChainType)> for Config {
```
TheQuantumPhysicist, line 66: I don't think the UpgradeVersion is necessary here.

consensus/src/pow/constants.rs
```diff
@@ -0,0 +1,14 @@
+use common::Uint256;
+
+pub const TARGET_TIMESPAN_SECS: u32 = 14 * 24 * 60 * 60; // 2 weeks
+pub const TARGET_TIMESPAN_UINT256: Uint256 =
+    Uint256([0x0000000000127500, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000]);
+
+pub const TARGET_SPACING: u32 = 10 * 60;
+pub const DIFFICULTY_ADJUSTMENT_INTERVAL: u32 = TARGET_TIMESPAN_SECS / TARGET_SPACING;
+
+/// taken from: https://github.com/bitcoin/bitcoin/blob/master/src/pow.cpp#L56
```
TheQuantumPhysicist, line 10: Whenever you take a link from github, make sure you use a permalink, which contains the commit number. You can do this by clicking on the `...` on the left of the line number. Why? Because the master branch can always change.

altonen:
I haven't looked at the code yet but I hope we aren't merging this as-is but that the history is squashed before merge. I think this is way too many commits. I can take a look if some of the duplicate commits can be squashed so we wouldn't lose all of this work but I think 383 commits is too much. Thoughts?

consensus/src/pow/constants.rs
```diff
@@ -0,0 +1,14 @@
+use common::Uint256;
+
+pub const TARGET_TIMESPAN_SECS: u32 = 14 * 24 * 60 * 60; // 2 weeks
```
TheQuantumPhysicist, line 3: This whole file should probably go into ChainConfig.

- TARGET_TIMESPAN_SECS should be in ChainConfig, and should be returned as rust Duration.
- Uint256 version should be derived from that, and not hardcoded.
- TARGET_SPACING should certainly be in ChainConfig.
- DIFFICULTY_ADJUSTMENT_INTERVAL should be a function.
- TIMESPAN_ADJUSTMENT_FACTOR: This should be called MAX_DIFFICULTY_ADJUSTMENT_PER_INTERVAL, which is basically the max increase/decrease in work that can happen. This protects from having huge swings in difficulty. Also should go into ChainConfig
- Finally, the upper and lower things also should be functions.

Preferably, find a way to group them all under PoW in ChainConfig instead of having them scattered. Maybe just prefix? Maybe substruct? I don't know. Please propose something.

TheQuantumPhysicist:
> I haven't looked at the code yet but I hope we aren't merging this as-is but that the history is squashed before merge. I think this is way too many commits. I can take a look if some of the duplicate commits can be squashed so we wouldn't lose all of this work but I think 383 commits is too much. Thoughts?

Taking micro-adjustments to another level, huh? :-D

Yeah, I agree. Carla, please squash before merging. You can do that locally too. Please make sure to backup your branch first.

b-yap:
This is not for merging to master, at all. I have added that in the description.
But yes, on the next PRs I will squash them before merging.

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+        false
+    }
+}
+
+impl Config {
+    pub fn check_for_work_required(
+        &self,
+        prev_block_index: &BlockIndex,
+        _height: BlockHeight,
```
TheQuantumPhysicist, line 27: The block index contains the height

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,111 @@
...
+    ) -> Result<Compact, Error> {
+        let pow_limit = self.limit;
+        let prev_block_bits = prev_block_index.data.bits();
+
+        if self.no_retargeting {
+            return Ok(prev_block_bits);
+        }
+
+        // limit adjustment step
+        let actual_timespan_of_last_2016_blocks =
```
TheQuantumPhysicist, line 56: call it "interval" instead of 2016 blocks

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,111 @@
...
+            actual_timespan_of_last_2016_blocks,
+            prev_block_bits,
+            pow_limit,
+        )
+    }
+
+    fn next_work_required_for_testnet(&self, time: u32, prev_block_index: &BlockIndex) -> Compact {
+        let pow_limit = Compact::from(self.limit);
+
+        if self.allow_min_difficulty_blocks {
```
TheQuantumPhysicist, line 70: So why do we have a separate function for testnet if we do this check which is true anyway for testnet? Please combine both functions into one function and make them depend on ChainConfig.

common/src/primitives/id.rs
muursh: As we briefly discussed earlier I think forking fixed_hash is the best way to approach this, my 30s peak makes it look like we should be dandy dependency wise so it's just a case of fixing the macros and tests and we should be good to go.

block_production/src/pow/network.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    pub fn allow_min_difficulty_blocks(&self) -> bool {
+        match self {
+            Network::MainNet | Network::SigNet => false,
+            Network::TestNet | Network::RegTest => true,
+        }
+    }
+
+    pub fn limit(&self) -> Uint256 {
```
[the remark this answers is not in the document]
b-yap: already moved: https://github.com/mintlayer/mintlayer-core/pull/95

b-yap:
continue to PR https://github.com/mintlayer/mintlayer-core/pull/99

PR #91, opened 2022-02-01
net/libp2p: Implement mDNS peer discovery

altonen:
This PR implements mDNS peer discovery and essentially allows the Mintlayer node to find other nodes in the local network. It adds `mdns` field to the `NetworkBehaviour` struct, waits for incoming mDNS events from the network and sends them to the P2P object for further processing. This PR does not implement any peer management so currently the P2P object receives the mDNS events and discards the records.

An unfortunate thing about `libp2p-mdns` is that there is no way during runtime to enable/disable it so it's either always on or never on and that decision is made when the code is compiled and there is no way to shut down the mDNS object or tell it to no longer query the network. It's not an attack vector or performance concern but "disabling" mDNS is not quite as elegant as one could hope: we basically just ignore incoming events and that's it. If `Libp2pStrategy::MulticastDns` is provided when the network service provider is being created, the P2P object will receive mDNS events and if the mDNS strategy is not specified, those events are still received by the libp2p backend but they are not relayed to the P2P object for further processing.

I don't know if there ever a use-case where you'd need to disable mDNS though. You'd have to nodes running locally at the same time and you wouldn't want them to connect to each other?

### DIFF
--- a/p2p/src/net/libp2p/backend.rs
+++ b/p2p/src/net/libp2p/backend.rs
@@ -21,7 +21,8 @@ use crate::{
 };
 use futures::StreamExt;
 use libp2p::{
-    core::connection::ConnectedPoint,
+    core::{connection::ConnectedPoint, either::EitherError},
+    mdns::MdnsEvent,
     streaming::{OutboundStreamId, StreamHandle, StreamingEvent},
     swarm::{NegotiatedSubstream, ProtocolsHandlerUpgrErr, Swarm, SwarmEvent},
     Multiaddr, PeerId,
@@ -53,6 +54,9 @@ pub struct Backend {
         OutboundStreamId,
         oneshot::Sender<error::Result<StreamHandle<NegotiatedSubstream>>>,
     >,
+
+    /// Whether mDNS peer events should be relayed to P2P manager
+    relay_mdns: bool,
 }
 
 impl Backend {
@@ -60,6 +64,7 @@ impl Backend {
         swarm: Swarm<common::ComposedBehaviour>,
         cmd_rx: Receiver<common::Command>,
         event_tx: Sender<common::Event>,
+        relay_mdns: bool,
     ) -> Self {
         Self {
             swarm,
@@ -68,6 +73,7 @@ impl Backend {
             dials: HashMap::new(),
             conns: HashMap::new(),
             streams: HashMap::new(),
+            relay_mdns,
         }
     }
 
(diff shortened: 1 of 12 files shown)


p2p/src/net/libp2p/mod.rs
```diff
@@ -43,8 +45,10 @@ mod common;
 // Maximum message size of 10 MB
 const MESSAGE_MAX_SIZE: u32 = 10 * 1024 * 1024;
 
-#[derive(Debug)]
-pub enum LibP2pStrategy {}
+#[derive(Debug, PartialEq, Eq)]
+pub enum Libp2pStrategy {
+    MulticastDns,
+}
```
iljakuklic, line 51: I'd like to see a brief doc comment explaining what this enum represents.

p2p/src/net/libp2p/mod.rs
```diff
@@ -447,4 +507,63 @@ mod tests {
             Err(P2pError::DecodeFailure("Message is too big".to_string()))
         );
     }
+
+    #[test]
+    fn test_parse_discovered_addr() {
+        let peer_id: PeerId =
+            "12D3KooWE3kBRAnn6jxZMdK1JMWx1iHtR1NKzXSRv5HLTmfD9u9c".parse().unwrap();
```
iljakuklic, line 514: This is probably ingrained into libp2p too much and almost certainly not worth the effort but I'd find bech32 peer IDs much nicer.

iljakuklic (approved):
I'm fine with this but preferably if someone with more networking knowledge could have a look as well.

p2p/src/net/libp2p/backend.rs
```diff
@@ -160,6 +169,40 @@ impl Backend {
...
+                let peers: Vec<(PeerId, Multiaddr)> = expired.collect::<_>();
+                if peers.is_empty() {
+                    return Ok(());
+                }
+
+                self.event_tx
+                    .send(common::Event::PeerExpired { peers })
+                    .await
+                    .map_err(|_| P2pError::ChannelClosed)
+            }
```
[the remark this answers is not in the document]
altonen: I'll try refactoring it into a helper function

p2p/src/net/libp2p/mod.rs
altonen: bech32 would certainly be nicer but as you said, this is a libp2p thing and I think we're stuck with these.

p2p/src/net/libp2p/backend.rs
```diff
@@ -86,10 +92,33 @@ impl Backend {
         }
     }
 
+    /// Collect peers into a vector and send appropriate event to P2P
+    async fn send_discovery_event(
+        &mut self,
+        peers: Vec<(PeerId, Multiaddr)>,
+        discovered: bool,
```
iljakuklic, line 99: Not a massive fan of bool flags, as it's not very clear at the call site what is the meaning. Not sure what would be better short of a custom enum, which seems like an overkill. Here's one possible trick using a function instead of bool:
```rust
    event_fn: impl FnOnce(Vec<(PeerId, Multiaddr)>) -> common::Event,
(... blah blah ...)
    self.event_tx.send(event_fn(peers)).(...)
```
And then call it like so
```rust
    self.send_discovery_event(peers.collect(), common::Event::PeerDiscovered)
```

iljakuklic (approved):

muursh (approved):

p2p/src/net/libp2p/backend.rs
altonen: Very nice, thanks. Fixed

PR #92, opened 2022-02-02
storage: Separate out mutable and immutable operations

iljakuklic:
* Method signature specifies mutability in a more precise way
* Database transactions distinguish between read-only transactions and read-write transactions
* Introduced traits abstracting over low-level storage operations

TODO:

- [x] In-memory storage could now use `RwLock` rather than `Mutex`
- [ ] Blockchain storage still uses one transaction type, not taking
       advantage of the separation between ro and rw transactions.
- [x] Update docs

### DIFF
--- a/blockchain_storage/src/mock.rs
+++ b/blockchain_storage/src/mock.rs
@@ -9,12 +9,12 @@ mockall::mock! {
     pub Store {}
 
     impl crate::BlockchainStorage for Store {
-        fn get_storage_version(&mut self) -> crate::Result<u32>;
+        fn get_storage_version(&self) -> crate::Result<u32>;
         fn set_storage_version(&mut self, version: u32) -> crate::Result<()>;
-        fn get_best_block_id(&mut self) -> crate::Result<Option<Id<Block>>>;
+        fn get_best_block_id(&self) -> crate::Result<Option<Id<Block>>>;
         fn set_best_block_id(&mut self, id: &Id<Block>) -> crate::Result<()>;
         fn add_block(&mut self, block: &Block) -> crate::Result<()>;
-        fn get_block(&mut self, id: Id<Block>) -> crate::Result<Option<Block>>;
+        fn get_block(&self, id: Id<Block>) -> crate::Result<Option<Block>>;
         fn del_block(&mut self, id: Id<Block>) -> crate::Result<()>;
         fn set_mainchain_tx_index(
             &mut self,
@@ -23,24 +23,21 @@ mockall::mock! {
         ) -> crate::Result<()>;
 
         fn get_mainchain_tx_index(
-            &mut self,
+            &self,
             tx_id: &Id<Transaction>,
         ) -> crate::Result<Option<TxMainChainIndex>>;
 
         fn del_mainchain_tx_index(&mut self, tx_id: &Id<Transaction>) -> crate::Result<()>;
 
         fn get_mainchain_tx_by_position(
-            &mut self,
+            &self,
             tx_index: &TxMainChainPosition,
         ) -> crate::Result<Option<Transaction>>;
 
-        fn get_mainchain_tx(
-            &mut self,
-            txid: &Id<Transaction>,
-        ) -> crate::Result<Option<Transaction>>;
+        fn get_mainchain_tx(&self, txid: &Id<Transaction>) -> crate::Result<Option<Transaction>>;
 
         fn get_block_id_by_height(
-            &mut self,
+            &self,
             height: &BlockHeight,
         ) -> crate::Result<Option<Id<Block>>>;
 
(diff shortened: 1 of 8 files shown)


blockchain_storage/src/mock.rs
```diff
@@ -53,9 +50,11 @@ mockall::mock! {
...
     }
 
-    impl storage::Transactional<'_> for Store {
-        type Transaction = MockStoreTx;
-        fn start_transaction(&mut self) -> MockStoreTx;
+    impl storage::traits::Transactional<'_> for Store {
+        type TransactionRo = MockStoreTx;
+        type TransactionRw = MockStoreTx;
+        fn start_transaction_ro(&self) -> MockStoreTx;
+        fn start_transaction_rw(&self) -> MockStoreTx;
```
TheQuantumPhysicist, line 57: Are you sure this is correct, taking a `&self` even though it's r/w?

iljakuklic: This follows the [interior mutablility pattern](https://doc.rust-lang.org/book/ch15-05-interior-mutability.html), same as `Mutex`or `LockRw`, or `RefCell`. The basic storage implementation uses a `RwLock` under the hood in this iteration. If you look for example at [RefCell doc](https://doc.rust-lang.org/std/cell/struct.RefCell.html), notice that both `borrow` and `borrow_mut` take `&self` wihtout mut.

There are basically three similar methods, taking `RefCell` as an example again:

1. `borrow(&self) -> Ref<'_, T>`
2. `borrow_mut(&self) -> RefMut<'_, T>`
3. `get_mut(&mut self) -> &mut T`

The `get_mut` method guarantees (due its type signature) that the RefCell is the only reference to the data. It is only usable upon initialization, once you start sharing the cell, you have to use `borrow` and `borrow_mut` which provide a runtime mechanism to enforce the aliasing rule. With storage, it is similar situation. The aliasing rules are enforced by a mutex internally so the storage can be shared by multiple subsystems.

In the real implementation (non-mock) the mutable and immutable case are differentiated by the type of transaction that is returned, i.e. `start_transaction_ro` returns a `Transactional::TransactionRo`, whereas `start_transaction_rw` returns a `Transactional::TransactionRw`, These two are analogous to (1) and (2). With mock I took the liberty to take a shortcut and simplify things a bit by having just one transaction type representing both mutable and immutable transactions.

TheQuantumPhysicist: Well, nevertheless, I'm wondering whether it nominally makes more sense to make it `&mut self` just for the meaningfulness of it. I mean, we're using the rust mutability properties to map them to database operation mutability properties, and we don't need to do that anyway, but it's for our own pleasure and safety in our code. So, it's going through my head whether we should just make it `&mut self` because we're writing, to distinguish from reading. Just like we make `get` a `&self` while a `set` is `&mut self`.

blockchain_storage/src/store.rs
```diff
@@ -436,26 +459,24 @@ mod test {
...
                     assert!(tx.set_storage_version(v + 3).is_ok());
                     assert!(tx.commit().is_ok());
                 })
             };
             let thr1 = {
-                let mut store = store.clone();
+                let store = Store::clone(&store);
                 common::thread::spawn(move || {
-                    let mut tx = store.start_transaction();
+                    let mut tx = store.start_transaction_rw();
```
TheQuantumPhysicist, line 479: No tests for read-only?

blockchain_storage/src/store.rs
```diff
@@ -416,9 +439,9 @@ mod test {
                 })
             };
             let thr1 = {
-                let mut store = store.clone();
+                let store = Store::clone(&store);
                 common::thread::spawn(move || {
-                    let tx_result = store.transaction(|tx| {
+                    let tx_result = store.write(|tx| {
```
TheQuantumPhysicist, line 444: I'm not a big fan of using `write` as a member function name for this... it's a little confusing.

iljakuklic: I am open to suggestions. I was following the naming scheme from `std::sync::RwLock` but `transaction_{ro,rw}` sounds sensible to me as well

blockchain_storage/src/store.rs
iljakuklic: Will revisit this once again and potentially add more tests or make some read-only if possible.

blockchain_storage/src/store.rs
TheQuantumPhysicist: `transaction_{ro,rw}` sounds good to me. And to be honest, I'd prefer calling this "inline transactions", but maybe `inline_transaction_{ro,rw}` is an overkill?

TheQuantumPhysicist (approved):

blockchain_storage/src/store.rs
iljakuklic: To be honest, I don't like the `inline_transaction_x` suggestion after thinking about that for a bit. Currently my preferred variant is `start_transaction_x` and `transaction_x`. I find that fairly descriptive: `start_transaction_x` starts a transaction, `transaction_x` defines the whole transaction. That is merely a different interface to the same thing, not a different "kind" of transaction, as the `inline_` prefix seems to suggest.

TheQuantumPhysicist: That's OK. Use what you suggested.

blockchain_storage/src/mock.rs
iljakuklic: Turns out having `transaction_rw` take `&mut self` is not nicely doable until blockchain storage also uses two separate types for read-only vs. read-write transactions. I.e. it makes it difficult to have one transaction type that implements both `TransactionRw` functionalisy and `TransactionRo` functionality.

PR #95, opened 2022-02-08
Proof of Work Config from ChainConfig

b-yap:
Based on the comment: https://github.com/mintlayer/mintlayer-core/pull/62/files#r801270588
Removed all `const`.
All PoW related config can be derived from `ChainConfig`, using the `chain_type` field. 

Computed values like the `UPPER_TARGET_TIMESPAN_SECS` and `LOWER_TARGET_TIMESPAN_SECS` and `DIFFICULTY_ADJUSTMENT_INTERVAL` will be functions in the PoW package of consensus module.

### DIFF
--- a/common/src/chain/config.rs
+++ b/common/src/chain/config.rs
@@ -40,6 +40,10 @@ pub struct ChainConfig {
 }
 
 impl ChainConfig {
+    pub fn chain_type(&self) -> ChainType {
+        self.chain_type
+    }
+
     pub fn address_prefix(&self) -> &str {
         &self.address_prefix
     }
@@ -102,6 +106,7 @@ pub fn create_mainnet() -> ChainConfig {
     }
 }
 
+#[cfg(test)]
 mod tests {
 
     #[test]
--- a/common/src/chain/config.rs
+++ b/common/src/chain/config.rs
@@ -2,7 +2,7 @@ use crate::address::Address;
 use crate::chain::block::Block;
 use crate::chain::transaction::Transaction;
 use crate::chain::upgrades::NetUpgrades;
-use crate::chain::UpgradeVersion;
+use crate::chain::{PoWChainConfig, UpgradeVersion};
 use crate::primitives::consensus_data::ConsensusData;
 use crate::primitives::id::{Id, H256};
 use crate::primitives::{version::SemVer, BlockHeight};
@@ -40,10 +40,6 @@ pub struct ChainConfig {
 }
 
 impl ChainConfig {
-    pub fn chain_type(&self) -> ChainType {
-        self.chain_type
-    }
-
     pub fn address_prefix(&self) -> &str {
         &self.address_prefix
     }
@@ -62,6 +58,10 @@ impl ChainConfig {
 
     pub fn net_upgrade(&self) -> &NetUpgrades<UpgradeVersion> {
         &self.net_upgrades
+    }
+
+    pub const fn get_proof_of_work_config(&self) -> PoWChainConfig {
+        PoWChainConfig::new(self.chain_type)
     }
 }
 
--- a/common/src/chain/pow.rs
+++ b/common/src/chain/pow.rs
@@ -1,68 +1,73 @@
 use crate::chain::config::ChainType;
-use crate::chain::ChainConfig;
 use crate::Uint256;
 use std::time::Duration;
 
-/// Chain Parameters for Proof of Work, as found in
-/// https://github.com/bitcoin/bitcoin/blob/eca694a4e78d54ce4e29b388b3e81b06e55c2293/src/chainparams.cpp
+/// Chain Parameters for Proof of Work.
+///
+/// See in Bitcoin's [chainparams.cpp](https://github.com/bitcoin/bitcoin/blob/eca694a4e78d54ce4e29b388b3e81b06e55c2293/src/chainparams.cpp)
+#[derive(Debug)]
 pub struct PoWChainConfig {
     no_retargeting: bool,
+    /// Checks whether minimum difficulty can be used for the block
     allow_min_difficulty_blocks: bool,
+    /// The lowest possible difficulty
     limit: Uint256,
 }
 
-impl ChainConfig {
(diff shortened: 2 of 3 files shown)


common/src/chain/config.rs
```diff
@@ -40,6 +40,10 @@ pub struct ChainConfig {
 }
 
 impl ChainConfig {
+    pub fn chain_type(&self) -> ChainType {
```
altonen, line 43: should this return a reference? All other getters of `ChainConfig` seem to return a reference

TheQuantumPhysicist: Yes and no. Generally, yes, it should be. But since this is just an empty enum with no contents other than the numeric arms, it supports the copy trait, and is just an integer, so it doesn't really matter.

common/src/chain/pow.rs
```diff
@@ -0,0 +1,112 @@
...
+
+/// Chain Parameters for Proof of Work, as found in
+/// https://github.com/bitcoin/bitcoin/blob/eca694a4e78d54ce4e29b388b3e81b06e55c2293/src/chainparams.cpp
+pub struct PoWChainConfig {
+    no_retargeting: bool,
+    allow_min_difficulty_blocks: bool,
+    limit: Uint256,
+}
+
+impl ChainConfig {
```
TheQuantumPhysicist, line 14: I'm not a big fan of scattering this around. Please put this in config.rs, then you're free to keep the implementation here through a proper link with a global function such as pow::xyz(&chain_config).

common/src/chain/pow.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn get_pow_config(&self) -> PoWChainConfig {
```
TheQuantumPhysicist, line 15: So every time we call this function, we'll instantiate a new one with all the relevant values? Kind of suboptimal. Can you think of a better way to do this?

common/src/chain/pow.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    pub fn target_timespan(&self) -> Duration {
+        Duration::new(14 * 24 * 60 * 60, 0)
+    }
+
+    pub fn target_spacing(&self) -> Duration {
+        Duration::new(10 * 60, 0)
+    }
+
+    /// https://github.com/bitcoin/bitcoin/blob/eca694a4e78d54ce4e29b388b3e81b06e55c2293/src/pow.cpp#L56
```
iljakuklic, line 45: I'm going to be a nitpicker again. This is a doc comment and as such will show up in documentation when you generate it using `cargo doc`. The first line is the main brief description of what the function does and appears in lists of functions and such. It is quite ugly to have just a URL there.

common/src/chain/pow.rs
```diff
@@ -0,0 +1,112 @@
+use crate::chain::config::ChainType;
+use crate::chain::ChainConfig;
+use crate::Uint256;
+use std::time::Duration;
+
+/// Chain Parameters for Proof of Work, as found in
+/// https://github.com/bitcoin/bitcoin/blob/eca694a4e78d54ce4e29b388b3e81b06e55c2293/src/chainparams.cpp
```
iljakuklic, line 7: Similar comment like below. To make it render nicely in the docs, maybe have a description and the link in the detailed section separately, roughly like that:
```
/// Chain Parameters for Proof of Work.
///
/// As found in Bitcoin [chainparams.cpp](https://github.com/bitcoin/bitcoin/blob/eca694a4e78d54ce4e29b388b3e81b06e55c2293/src/chainparams.cpp)
```

iljakuklic (approved):
While I and others have left comments, I don't see a showstopper

common/src/chain/pow.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn allow_min_difficulty_blocks(&self) -> bool {
+        self.allow_min_difficulty_blocks
+    }
+
+    pub fn limit(&self) -> Uint256 {
+        self.limit
+    }
+
+    pub fn target_timespan(&self) -> Duration {
+        Duration::new(14 * 24 * 60 * 60, 0)
```
muursh, line 38: While it may be obvious now I feel a short comment describing these numbers might prove helpful, to someone, in the future. Same for `target_spacing` below

common/src/chain/pow.rs
b-yap: would you prefer converting the whole `ChainConfig` into `PoWChainConfig` ?

TheQuantumPhysicist: The problem here is the constant reservation of memory on every call to `get_pow_config`. I don't have a specific solution in mind. These values should be reserved once, only when the ChainConfig created. Then, they should be requested on demand.

iljakuklic: I am not convinced optimising this is worth the effort. If this is done only once per block, it's rather trivial operation compared to the rest of the processing of the block. I'd just leave as it is and only revisit if it actually shows up during profiling (which I quite doubt it will). Also compilers tend to be really good at optimizing this stuff, all the temporaries will likely be gone when you compile this.

TheQuantumPhysicist: Well, it's just bad practice. Sure, the compiler may be nice enough to fix this, but why play that game if a better design can come with this? For proof of stake, this is going to be much bigger and more complex. It's a win if Carla can come up with a better way to do this for all consensus algorithms. Besides, this is included in the first line of defense where we verify the work of a block (assuming we're proof of work chain, but again, we're seeking things to be done nominally right), hence, any spam attempt will face this first. Sure, we shouldn't spend a week planning this. But since we're in the design stage, let's aim for this to be done right.

TheQuantumPhysicist (approved):

common/src/chain/pow.rs
b-yap: updated; converted method into `const fn`

common/src/chain/config.rs
b-yap: i removed this, since getting the pow config is already a method under ChainConfig. I didn't expose the chain_type anymore.

PR #97, opened 2022-02-09
Peer management

altonen:
Implement the last missing pieces from peer management before the work on syncing can start

### DIFF
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -17,7 +17,8 @@
 #![cfg(not(loom))]
 
 use crate::{
-    event::{Event, PeerEvent},
+    error::P2pError,
+    event::{Event, PeerEvent, PeerEventType},
     net::NetworkService,
     peer::{Peer, PeerId, PeerRole},
 };
@@ -40,6 +41,28 @@ pub mod peer;
 pub mod proto;
 
 #[allow(unused)]
+#[derive(Debug, PartialEq, Eq)]
+enum PeerState {
+    /// Peer is handshaking
+    Handshaking,
+
+    /// Peer is ready for gossiping/syncing
+    Active,
+}
+
+#[allow(unused)]
+struct PeerContext {
+    /// Unique peer ID
+    id: PeerId,
+
+    /// Peer state
+    state: PeerState,
+
+    /// Channel for communication with the peer
+    tx: Sender<Event>,
+}
+
+#[allow(unused)]
 pub enum ConnectivityEvent<T>
 where
     T: NetworkService,
@@ -57,7 +80,7 @@ pub struct P2P<NetworkingBackend> {
     config: Arc<ChainConfig>,
 
     /// Hashmap for peer information
-    peers: HashMap<PeerId, Sender<Event>>,
+    peers: HashMap<PeerId, PeerContext>,
 
     /// Counter for getting unique peer IDs
     peer_cnt: AtomicU64,
@@ -102,9 +125,27 @@ where
     ///
     /// The event is wrapped in an `Option` because the peer might have ungracefully
     /// failed and reading from the closed channel might gives a `None` value, indicating
-    /// a protocol on error which should be handled accordingly.
+    /// a protocol error which should be handled accordingly.
     async fn on_peer_event(&mut self, event: Option<PeerEvent>) -> error::Result<()> {
-        todo!();
+        let event = event.ok_or(P2pError::ChannelClosed)?;
+
+        match event.event {
+            PeerEventType::HandshakeFailed => self
+                .peers
+                .remove(&event.peer_id)
+                .map(|_| ())
+                .ok_or_else(|| P2pError::Unknown("Peer does not exist".to_string())),
+            PeerEventType::HandshakeSucceeded => match self.peers.get_mut(&event.peer_id) {
+                Some(peer) => {
+                    (*peer).state = PeerState::Active;
+                    Ok(())
+                }
+                None => Err(P2pError::Unknown("Peer does not exist".to_string())),
+            },
+            PeerEventType::Disconnected | PeerEventType::Message(_) => {
+                todo!();
+            }
+        }
     }
 
     /// Handle a connectivity-related event
@@ -175,7 +216,14 @@ where
         let (tx, rx) = tokio::sync::mpsc::channel(self.peer_backlock);
 
         let peer_id: PeerId = self.peer_cnt.fetch_add(1, Ordering::Relaxed);
-        self.peers.insert(peer_id, tx);
+        self.peers.insert(
+            peer_id,
+            PeerContext {
+                id: peer_id,
+                state: PeerState::Handshaking,
+                tx,
+            },
+        );
 
         tokio::spawn(async move {
             Peer::<NetworkingBackend>::new(peer_id, role, config, socket, mgr_tx, rx)
@@ -225,5 +273,67 @@ mod tests {
             res,
             Err(P2pError::SocketError(std::io::ErrorKind::ConnectionRefused))
         );
+    }
+
+    // verify that if handshake succeeds, peer state is set to `Active`
+    #[tokio::test]
+    async fn test_on_peer_event_handshake_success() {
+        let config = Arc::new(config::create_mainnet());
+        let addr: SocketAddr = test_utils::make_address("[::1]:");
+        let mut p2p = P2P::<MockService>::new(256, 32, addr, Arc::clone(&config)).await.unwrap();
+        let (tx, _) = tokio::sync::mpsc::channel(16);
+
+        p2p.peers.insert(
+            1,
+            PeerContext {
+                id: 1,
+                state: PeerState::Handshaking,
+                tx: tx.clone(),
+            },
+        );
+
+        assert_eq!(p2p.peers.len(), 1);
+        assert_eq!(
+            p2p.on_peer_event(Some(PeerEvent {
+                peer_id: 1,
+                event: PeerEventType::HandshakeSucceeded,
+            }))
+            .await,
+            Ok(())
+        );
+        assert_eq!(p2p.peers.len(), 1);
+        match p2p.peers.get(&1) {
+            Some(peer) => assert_eq!(peer.state, PeerState::Active),
+            None => panic!("peer not found"),
+        }
+    }
+
+    // verify that if handshake fails, peer context is destroyed
+    #[tokio::test]
+    async fn test_on_peer_event_handshake_failure() {
+        let config = Arc::new(config::create_mainnet());
+        let addr: SocketAddr = test_utils::make_address("[::1]:");
+        let mut p2p = P2P::<MockService>::new(256, 32, addr, Arc::clone(&config)).await.unwrap();
+        let (tx, _) = tokio::sync::mpsc::channel(16);
+
+        p2p.peers.insert(
+            1,
+            PeerContext {
+                id: 1,
+                state: PeerState::Handshaking,
+                tx: tx.clone(),
+            },
+        );
+
+        assert_eq!(p2p.peers.len(), 1);
+        assert_eq!(
+            p2p.on_peer_event(Some(PeerEvent {
+                peer_id: 1,
+                event: PeerEventType::HandshakeFailed,
+            }))
+            .await,
+            Ok(())
+        );
+        assert_eq!(p2p.peers.len(), 0);
     }
 }
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -24,7 +24,10 @@ use crate::{
 };
 use common::chain::ChainConfig;
 use futures::FutureExt;
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::Arc,
+};
 use tokio::sync::mpsc::{Receiver, Sender};
 
 pub mod error;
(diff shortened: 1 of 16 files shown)


p2p/src/lib.rs
```diff
@@ -87,9 +133,9 @@ where
         Ok(Self {
             network: NetworkingBackend::new(addr, &[], &[]).await?,
             config,
-            peer_cnt: AtomicU64::default(),
             peer_backlock,
             peers: HashMap::new(),
```
muursh, line 137: Since we know how many peers there can be, we have a max number after all, it might be worth using `with_capacity()` with the peers hashmap so we can guarantee avoid the pain of reallocating memory as we add elements to it.

p2p/src/net/libp2p/mod.rs
```diff
@@ -70,6 +71,9 @@ pub struct Libp2pSocket {
     /// Multiaddress of the remote peer
     addr: Multiaddr,
 
+    /// Unique ID of the peer
+    pub id: PeerId,
```
TheQuantumPhysicist, line 75: pub?

p2p/src/net/libp2p/mod.rs
```diff
@@ -101,11 +105,41 @@ fn parse_discovered_addr(peer_id: PeerId, peer_addr: Multiaddr) -> Option<Multia
     }
 }
 
+/// Parse all discovered addresses and group them by PeerId
+fn parse_peers<T>(peers: Vec<(PeerId, Multiaddr)>) -> Vec<net::AddrInfo<T>>
```
TheQuantumPhysicist, line 109: Whenever I see `mut`, I feel allergic.

So this function seems to be doing the same thing for every peer, then putting the result in a Vec. Well, why isn't this isolated per peer, and then an `into_iter()` of the Vec is used to consume it to the desired result? Unless I'm missing something, this can be made easier on the eyes.

Since peers are independent, please write a function that processes each peer individually, then, iter() over these function calls and collect the result.

p2p/src/lib.rs
```diff
@@ -39,34 +37,82 @@ pub mod net;
 pub mod peer;
 pub mod proto;
 
+const MAX_ACTIVE_CONNECTIONS: usize = 32;
```
TheQuantumPhysicist, line 40: Btw, bitcoin can take up to 125, with 16 max as outbound.

p2p/src/lib.rs
```diff
@@ -117,24 +184,90 @@ where
...
+            assert!(!ip4.is_empty() || !ip6.is_empty());
+
+            // TODO: let user specify their preference?
+            let addr = if ip6.is_empty() {
+                Arc::clone(ip4.iter().next().unwrap())
+            } else {
+                Arc::clone(ip6.iter().next().unwrap())
+            };
+
+            peers.push((*peer_info.0, addr));
```
TheQuantumPhysicist, line 238: Please use `(0..npeers).into_iter().map(...)`

(I know there's a TODO, but this is simple enough to be done in this PR.)

On the other hand, this algorithm can be improved significantly. First, there has to be some pseudo-randomness in this. Otherwise, statistically speaking, a specific group of people will be bombarded with more connections than others. So what I suggest is, pick random references to `discovered`, put them in a set (to ensure there are no duplicates), then loop over them with a proper iterator. Using indexes here isn't really necessary and is basically the reason for that unnecessary `expect()`.

p2p/src/lib.rs
muursh: It should probably be command line option, default to whatever, but let people play with it

p2p/src/lib.rs
```diff
@@ -117,24 +184,90 @@ where
...
+            return Err(P2pError::NoPeers);
+        }
+
+        let npeers = std::cmp::min(
+            self.discovered.len(),
+            MAX_ACTIVE_CONNECTIONS - self.peers.len(),
+        );
+
+        // TODO: improve peer selection
+        let mut peers = Vec::new();
```
TheQuantumPhysicist, line 221: Whenever you create a Vec::new(), planning to push into it, think that this can be exchanged with a mapping as long as the elements don't depend on each other.

p2p/src/lib.rs
```diff
@@ -117,24 +184,90 @@ where
...
+        let mut iter = self.discovered.iter();
+
+        for _ in 0..npeers {
+            let peer_info = iter.next().expect("Peer to exist");
+            let (ip4, ip6) = match peer_info.1 {
+                PeerAddrInfo::Raw { ip4, ip6 } => (ip4, ip6),
+            };
+            assert!(!ip4.is_empty() || !ip6.is_empty());
+
+            // TODO: let user specify their preference?
```
TheQuantumPhysicist, line 231: Could you explain what this comment means?

p2p/src/lib.rs
```diff
@@ -117,24 +184,90 @@ where
...
+            let addr = if ip6.is_empty() {
+                Arc::clone(ip4.iter().next().unwrap())
```
TheQuantumPhysicist, line 233: Please avoid calling next() directly.

p2p/src/lib.rs
```diff
@@ -117,24 +184,90 @@ where
...
+    fn peer_discovered(&mut self, peers: &[net::AddrInfo<NetworkingBackend>]) -> error::Result<()> {
+        for info in peers.iter() {
+            if self.peers.contains_key(&info.id) {
+                continue;
+            }
+
+            match self.discovered.entry(info.id).or_insert_with(|| PeerAddrInfo::Raw {
+                ip4: HashSet::new(),
+                ip6: HashSet::new(),
+            }) {
```
TheQuantumPhysicist, line 259: Maybe there's something wrong with my IDE; though please format the file.

TheQuantumPhysicist:
Besides the few comments I made, everything else looks fine. After addressing them, I'll redo the review given that this PR is relatively big.

p2p/src/net/libp2p/mod.rs
altonen: I'll check if `pub` can be removed but basically our own peer ID is stored inside the the `Libp2pService` and having access to it is IMO cleaner than returning `(Libp2pService, PeerId)`. I also don't like the idea of "polluting" the `NetworkService` trait with getters.

p2p/src/lib.rs
altonen: It looks really ugly but the code is formatted. I believe if we used nightly, we'd have access to `PeerAddrInfo::Raw::default()` which would be much cleaner. But that alone probably doesn't warrant the switch

p2p/src/lib.rs
altonen: what should I call?

p2p/src/lib.rs
altonen: Mintlayer can fight the power and drive IPv6 adoption by preferring it over IPv4 :smiley: 

In all seriousness though, it's just something that we can decide for the user or the user can control if they so wish. It's extra work so might as well not give them the option. We know what's best for them.

p2p/src/net/libp2p/mod.rs
TheQuantumPhysicist: It's actually not pollution, I'd argue. Though I sympathize with what you're thinking. There are crates that add getters by just adding an attribute if you're worried about the lines of code. 

Using a getter basically gives you compile-time guarantees, including, for example, the inability to mutate the value even if the struct was mutable.

p2p/src/lib.rs
altonen: This current implementation of the peer selection algorithm is horrible, I fully admit that. Right now though, and probably for the next two months, there aren't going to be external peers so having more sophisticated mechanisms for selecting peers is not needed. I have some plans on how to improve the overall connectivity of our entire blockchain network but that requires that we have external peers (outside our local network) and a way to exchange peer information with them.

But I'll remove the iteration and replace it with `map()`

p2p/src/lib.rs
TheQuantumPhysicist: Depends on how many elements you expect in there and what you're really targeting. You seem to have an invariant that the set shouldn't be empty. But then do we specifically want to get the first element? A random element? Then based on that you can pick. This makes me wonder why we don't just accept all addresses, as is.

p2p/src/lib.rs
TheQuantumPhysicist: I get you... though it's not a big deal and can be easily implemented. Anyway, if you plan to keep it, please open an issue and reference this comment so that we don't forget.

p2p/src/net/libp2p/mod.rs
altonen: We get a vector of `(PeerId, MultiAddr)` tuples from libp2p and there is not other grouping so first we need to parse each address to verify that it's something Mintlayer supports and then group all the valid addresses under the given `PeerId`.

So basically this list (which is a valid list of discovered addresses):
```rust
vec![
    ("12D3KooWRn14SemPVxwzdQNg8e8Trythiww1FWrNfPbukYBmZEbJ", "/ip6/::1/tcp/9091"),
    ("12D3KooWE3kBRAnn6jxZMdK1JMWx1iHtR1NKzXSRv5HLTmfD9u9c", "/ip6/::1/tcp/5555"),
    ("12D3KooWE3kBRAnn6jxZMdK1JMWx1iHtR1NKzXSRv5HLTmfD9u9c", "/dns4/foo.com/tcp/80/http"),
    ("12D3KooWGK4RzvNeioS9aXdzmYXU3mgDrRPjQd8SVyXCkHNxLbWN", "/ip4/127.0.0.1/tcp/80/http"),
    ("12D3KooWRn14SemPVxwzdQNg8e8Trythiww1FWrNfPbukYBmZEbJ", "/dns4/foo.com/tcp/80/http"),
    ("12D3KooWGK4RzvNeioS9aXdzmYXU3mgDrRPjQd8SVyXCkHNxLbWN", "/ip4/127.0.0.1/tcp/8080"),
    ("12D3KooWE3kBRAnn6jxZMdK1JMWx1iHtR1NKzXSRv5HLTmfD9u9c", "/ip4/127.0.0.1/tcp/6666/quic"),
    ("12D3KooWE3kBRAnn6jxZMdK1JMWx1iHtR1NKzXSRv5HLTmfD9u9c", "/ip4/127.0.0.1/tcp/9090/quic"),
    ("12D3KooWRn14SemPVxwzdQNg8e8Trythiww1FWrNfPbukYBmZEbJ", "/ip4/127.0.0.1/tcp/7777"),
    ("12D3KooWGK4RzvNeioS9aXdzmYXU3mgDrRPjQd8SVyXCkHNxLbWN", "/ip6/::1/tcp/8081"),
    ("12D3KooWE3kBRAnn6jxZMdK1JMWx1iHtR1NKzXSRv5HLTmfD9u9c", "/dns6/foo.com/tcp/443/https"),
]
```
is converted into an address information list, like this:
```rust
vec![
    AddrInfo {
        id: "12D3KooWRn14SemPVxwzdQNg8e8Trythiww1FWrNfPbukYBmZEbJ",
        ip4: vec!["/ip4/127.0.0.1/tcp/7777"],
        ip6: vec!["/ip6/::1/tcp/9091"]
    },
    AddrInfo {
        id: "12D3KooWGK4RzvNeioS9aXdzmYXU3mgDrRPjQd8SVyXCkHNxLbWN",
        ip4: vec!["/ip4/127.0.0.1/tcp/8080"],
        ip6: vec!["/ip6/::1/tcp/8081"]
    }
    AddrInfo {
        id: "12D3KooWE3kBRAnn6jxZMdK1JMWx1iHtR1NKzXSRv5HLTmfD9u9c",
        ip4: vec![],
        ip6: vec!["/ip6/::1/tcp/5555"],
    }
]
```

So during this process there is no peer in the sense that P2P, or libp2p for that matter, would benefit from it but rather a unique ID and a bunch addresses. When this info is sent over the channel to P2P, libp2p has no use for these anymore so a creating temporary peer object doesn't make sense and creating the peer that P2P will use in libp2p would leak information to the network service provider which doesn't really concern in them in any way.

TheQuantumPhysicist: There's probably a misunderstanding. All I'm saying is, everything you're doing inside the loop should move into a separate function, and then you should map that function on the input vector and then collect into the result.

I may have more comments after doing this step. But from what it looks like now, it's very easy call a map instead of a mutable hashmap.

altonen: Yes I think there is a misunderstanding. I looked at the code and maybe I'm missing some Rust knowledge but what you're suggesting is not as trivial as it sounds and I'm not convinced the end result is cleaner or more maintainable than what the code currently is.

Could you show me how you would do it?

p2p/src/net/libp2p/mod.rs
altonen: Done, `pub` was not needed

p2p/src/net/libp2p/mod.rs
altonen: If it now looks OK, please approve and I'll squash the commit with the previous commit.

p2p/src/lib.rs
altonen: https://github.com/mintlayer/mintlayer-core/issues/98

p2p/src/net/libp2p/mod.rs
TheQuantumPhysicist: Looks perfect now!

TheQuantumPhysicist (approved):

muursh (approved):
I'm happy with this

PR #99, opened 2022-02-10
proof of work based on bitcoin.

b-yap:
continue from: https://github.com/mintlayer/mintlayer-core/pull/62

address comments: 
https://github.com/mintlayer/mintlayer-core/pull/62#discussion_r801266191
> I don't think the UpgradeVersion is necessary here.

https://github.com/mintlayer/mintlayer-core/pull/62#discussion_r801268578
> Whenever you take a link from github, make sure you use a permalink, which contains the commit hash. You can do this by clicking on the ... on the left of the line number. Why? Because the master branch can always change.

https://github.com/mintlayer/mintlayer-core/pull/62#discussion_r801270588
> This whole file should probably go into ChainConfig. 
TARGET_TIMESPAN_SECS should be in ChainConfig, and should be returned as rust Duration.
Uint256 version should be derived from that, and not hardcoded.
TARGET_SPACING should certainly be in ChainConfig.
DIFFICULTY_ADJUSTMENT_INTERVAL should be a function.
TIMESPAN_ADJUSTMENT_FACTOR: This should be called MAX_DIFFICULTY_ADJUSTMENT_PER_INTERVAL, which is basically the max increase/decrease in work that can happen. This protects from having huge swings in difficulty. Also should go into ChainConfig
Finally, the upper and lower things also should be functions.
Preferably, find a way to group them all under PoW in ChainConfig instead of having them scattered. Maybe just prefix? Maybe substruct? I don't know. Please propose something.

https://github.com/mintlayer/mintlayer-core/pull/62#discussion_r801282932
> The block index contains the height

https://github.com/mintlayer/mintlayer-core/pull/62#discussion_r801283521
> call it "interval" instead of 2016 blocks

https://github.com/mintlayer/mintlayer-core/pull/62#discussion_r801284291
> So why do we have a separate function for testnet if we do this check which is true anyway for testnet? Please combine both functions into one function and make them depend on ChainConfig.

~~Another TODO:~~ To be created in another PR.
https://github.com/mintlayer/mintlayer-core/pull/62#discussion_r801260163
> Sam: This isn't an operation I'm particularly happy about and I think we should discuss the reason why this is necessary. Since we're starting from scratch, let's try to make both H256 and Uint256 use the same order. No need to reverse things every time. This can get very confusing and error-prone very quickly.  

> Ben: As we briefly discussed earlier I think forking fixed_hash is the best way to approach this, my 30s peak makes it look like we should be dandy dependency wise so it's just a case of fixing the macros and tests and we should be good to go.

### DIFF
--- a/common/src/chain/block/consensus_data.rs
+++ b/common/src/chain/block/consensus_data.rs
@@ -1,3 +1,4 @@
+use crate::chain::TxOutput;
 use crate::primitives::Compact;
 use parity_scale_codec::{Decode, Encode};
 
@@ -13,11 +14,17 @@ pub enum ConsensusData {
 pub struct PoWData {
     bits: Compact,
     nonce: u128,
+    /// contains the block reward
+    outputs: Vec<TxOutput>,
 }
 
 impl PoWData {
-    pub fn new(bits: Compact, nonce: u128) -> Self {
-        PoWData { bits, nonce }
+    pub fn new(bits: Compact, nonce: u128, outputs: Vec<TxOutput>) -> Self {
+        PoWData {
+            bits,
+            nonce,
+            outputs,
+        }
     }
     pub fn bits(&self) -> Compact {
         self.bits
@@ -25,5 +32,13 @@ impl PoWData {
 
     pub fn nonce(&self) -> u128 {
         self.nonce
+    }
+
+    pub fn outputs(&self) -> &[TxOutput] {
+        &self.outputs
+    }
+
+    pub fn update_nonce(&mut self, nonce: u128) {
+        self.nonce = nonce;
     }
 }
--- a/consensus/src/pow/helpers.rs
+++ b/consensus/src/pow/helpers.rs
@@ -1,0 +1,110 @@
+use crate::pow::temp::BlockIndex;
+use crate::pow::Error;
+use common::primitives::{BlockHeight, Compact, H256};
+use common::Uint256;
+
+/// checks if retargeting is due for the provided block_height
+pub fn is_for_retarget(difficulty_adjustment_interval: u64, block_height: BlockHeight) -> bool {
+    block_height.inner() % difficulty_adjustment_interval == 0
+}
+
+/// The block time of the nth block, based on the difficulty adjustment interval,
+/// where nth block = height of given block - difficulty adjustment interval - 1 (off by one)
+pub fn retarget_block_time(difficulty_adjustment_interval: u64, block_index: &BlockIndex) -> u32 {
+    let retarget_height = {
+        // Go back by what we want to be 14 days worth of blocks (the last 2015 blocks)
+        let old_block_height = block_index.height.inner() - (difficulty_adjustment_interval - 1);
+        BlockHeight::new(old_block_height)
+    };
+
+    let retarget_block_index = block_index.get_ancestor(retarget_height);
+
+    retarget_block_index.get_block_time()
+}
+
+/// Returns a calculated new target as Compact datatype.
+/// See Bitcoin's Protocol rules of [Difficulty change](https://en.bitcoin.it/wiki/Protocol_rules)
+/// # Arguments
+/// `actual_timespan_of_last_interval` - the actual timespan or the difference between the current block
+/// and the 2016th block before it. This should be in seconds.
+/// `target_timespan` - found in the `PoWChainConfig`. This should be in seconds.
+/// `old_target` - Coming from the last block, this is the `bits` of the PoWData.
+/// `difficulty_limit` - found in the PoWChainConfig, as `limit`
+pub fn calculate_new_target(
+    actual_timespan_of_last_interval: u64,
+    target_timespan: u64,
+    old_target: Compact,
+    difficulty_limit: Uint256,
+) -> Result<Compact, Error> {
+    let actual_timespan = Uint256::from_u64(actual_timespan_of_last_interval).ok_or_else(|| {
+        Error::ConversionError(format!(
+            "conversion of actual timespan {:?} to Uint256 type failed.",
+            actual_timespan_of_last_interval
+        ))
+    })?;
+
+    let target_timespan = Uint256::from_u64(target_timespan).ok_or_else(|| {
+        Error::ConversionError(format!(
+            "conversion of target timespan {:?} to Uint256 type failed.",
+            target_timespan
+        ))
+    })?;
+
+    let old_target = Uint256::try_from(old_target).map_err(|e| {
+        Error::ConversionError(format!(
+            "conversion of bits {:?} to Uint256 type: {:?}",
+            old_target, e
+        ))
+    })?;
+
+    // new target is coputed by  multiplying the old target by ratio of the actual timespan / target timespan.
+    // see Bitcoin's Protocol rules of Difficulty change: https://en.bitcoin.it/wiki/Protocol_rules
+    let mut new_target = old_target * actual_timespan;
+    new_target = new_target / target_timespan;
+
+    new_target = if new_target > difficulty_limit {
+        difficulty_limit
+    } else {
+        new_target
+    };
+
+    Ok(Compact::from(new_target))
+}
+
+pub fn check_difficulty(block_hash: H256, difficulty: &Uint256) -> bool {
+    let id: Uint256 = block_hash.into(); //TODO: needs to be tested
+
+    id <= *difficulty
+}
+
+pub mod special_rules {
+    use super::*;
+
+    /// Checks if it took > 20 minutes to find a block
+    pub fn is_restart_difficulty(
+        target_spacing_in_secs: u64,
+        new_block_time: u32,
+        prev_block_time: u32,
+    ) -> bool {
+        new_block_time as u64 > (prev_block_time as u64 + (target_spacing_in_secs * 2))
+    }
+
+    pub fn last_non_special_min_difficulty(_block_index: &BlockIndex) -> Compact {
+        // Return the last non-special-min-difficulty-rules-block
+        // let mut ctr_index = block_index.clone();
+        // loop {
+        //     let block_bits = ctr_index.data.bits();
+        //     if ctr_index.height == BlockHeight::zero() {
+        //         return block_bits;
+        //     }
+        //
+        //     if is_for_retarget(pow_cfg, ctr_index.height) && block_bits == pow_cfg.limit() {
+        //         match ctr_index.prev() {
+        //             None => { return block_bits; }
+        //             Some(id) => {   }
+        //         }
+        //     }
+        // }
+        todo!()
+    }
+}
--- a/consensus/src/pow/helpers.rs
+++ b/consensus/src/pow/helpers.rs
@@ -4,7 +4,7 @@ use common::primitives::{BlockHeight, Compact};
 use common::Uint256;
 
 /// checks if retargeting is due for the provided block_height
-pub fn is_for_retarget(difficulty_adjustment_interval: u64, block_height: BlockHeight) -> bool {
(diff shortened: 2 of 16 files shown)


consensus/src/pow/work.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::pow::temp::BlockIndex;
+use crate::pow::{Error, PoW};
+use common::chain::block::Block;
+use common::primitives::consensus_data::{ConsensusData, PoWData};
+use common::primitives::{Compact, Idable};
+use common::Uint256;
+
+pub fn check_proof_of_work(hash: Uint256, bits: Compact) -> bool {
+    if let Ok(target) = Uint256::try_from(bits) {
+        hash > target
```
TheQuantumPhysicist, line 15: Two things:

1. This should be the other way around. The hash has to be less than the target.
2. Depending on how this is gonna be used (it's never used anywhere so far), if this is gonna be used once per block, then this is OK. If this is gonna be used for mining, then this is bad, because we don't want to reinterpret the bits every time we change the nonce.

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+impl PoW {
+    /// The difference (in block time) between the current block and 2016th block before the current one.
+    /// This difference should be inclusive between (2 weeks/4) and (2 weeks*4).
+    /// See Bitcoin's Protocol rules on [Difficulty change](https://en.bitcoin.it/wiki/Protocol_rules)
+    pub fn actual_timespan(&self, prev_block_blocktime: u32, retarget_blocktime: u32) -> u64 {
+        let actual_timespan = (prev_block_blocktime - retarget_blocktime) as u64;
+
+        // 2 weeks / 4
```
TheQuantumPhysicist, line 28: This is the wrong place for these comments. It should be where you put the numbers. Here you just explain the concept.

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,133 @@
...
+        let lower_bound = self.min_target_timespan_in_secs();
+        // 2 wees * 4
+        let upper_bound = self.max_target_timespan_in_secs();
+
+        if actual_timespan < lower_bound {
```
TheQuantumPhysicist, line 33: I think you're looking for `num::clamp` here.

consensus/src/pow/work.rs
TheQuantumPhysicist: Also, btw, this isn't 2 weeks in mintlayer. It's in bitcoin.

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        let current_height = prev_block_index
+            .height
+            .checked_add(1)
+            .ok_or_else(|| Error::OutofBounds("max block height has been reached.".to_string()))?;
+        let adjustment_interval = self.difficulty_adjustment_interval();
+
+        if is_for_retarget(adjustment_interval, current_height) {
+            let retarget_block_time = retarget_block_time(adjustment_interval, prev_block_index);
```
TheQuantumPhysicist, line 61: This doesn't seem to be correct. The block that you use to recalculate the new nBits is the first block in the 2016-block interval. Then, if that is longer than two weeks, the difficulty goes down. If it's less than two weeks, the difficulty goes up.

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,133 @@
...
+        )
+    }
+
+    fn next_work_required_for_min_difficulty(
+        &self,
+        new_block_time: u32,
+        prev_block_index: &BlockIndex,
+    ) -> Compact {
+        // If the new block's timestamp is more than 2 * 10 minutes
+        // then allow mining of a min-difficulty block.
```
TheQuantumPhysicist, line 96: I don't think bitcoin uses such a rule. Where is this coming from?

b-yap: https://github.com/bitcoin/bitcoin/blob/master/src/pow.cpp#L24

This is for testnet,

consensus/src/pow/work.rs
b-yap: I will rename the function to `get_starting_block_time`. 

The function is using the `prev_block_index` to call the `get_ancestor` method, based on the adjustment interval.

consensus/src/pow/work.rs
b-yap: Yes, that's wrong. :( I will change it asap.

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        // Return the last non-special-min-difficulty-rules-block
+        special_rules::last_non_special_min_difficulty(prev_block_index)
+    }
+}
+
+pub fn mine(block: &mut Block, max_nonce: u128, bits: Compact) -> Result<bool, Error> {
+    for nonce in 0..max_nonce {
+        //TODO: block reward is currently empty.
+        let data = PoWData::new(bits, nonce, vec![]);
```
TheQuantumPhysicist, line 113: Shouldn't the outputs be passed as a parameter to this function?

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        block.update_consensus_data(ConsensusData::PoW(data));
```
TheQuantumPhysicist, line 115: Please add a TODO to optimize this... this is very inefficient. We can mutate the binary form of the block directly to make this much more efficient. But this isn't a priority, even by a long shot.

consensus/src/pow/work.rs
```diff
@@ -0,0 +1,163 @@
...
+        let adjustment_interval = self.difficulty_adjustment_interval();
+
+        if is_for_retarget(adjustment_interval, current_height) {
+            let retarget_block_time =
+                get_starting_block_time(adjustment_interval, prev_block_index);
+            self.next_work_required(retarget_block_time, prev_block_index)
+        }
+        // special difficulty rules
+        else if self.allow_min_difficulty_blocks() {
+            Ok(self.next_work_required_for_min_difficulty(new_block_time, prev_block_index))
```
altonen, line 67: Shouldn't the special rule be checked first and if it's false, then adjust the difficulty if needed or am I missing something obvious?

consensus/src/pow/helpers.rs
```diff
@@ -0,0 +1,129 @@
+use crate::pow::temp::BlockIndex;
+use crate::pow::Error;
+use common::primitives::{BlockHeight, Compact};
+use common::Uint256;
+
+/// checks if retargeting is due for the provided block_height
+pub fn is_for_retarget(difficulty_adjustment_interval: u64, block_height: BlockHeight) -> bool {
```
TheQuantumPhysicist, line 7: Please call this `due_for_retarget`

consensus/src/pow/helpers.rs
```diff
@@ -0,0 +1,129 @@
...
+    };
+
+    Ok(Compact::from(new_target))
+}
+
+pub mod special_rules {
+    use super::*;
+
+    /// Checks if it took > 20 minutes to find a block
+    pub fn is_restart_difficulty(
```
TheQuantumPhysicist, line 81: Please call this `block_production_stalled`

consensus/src/pow/work.rs
TheQuantumPhysicist: I agree with you. In fact, thinking of this, I could argue that this could be a mistake in the implementation. Unless I'm missing something, retargeting has nothing to do with whether min_difficulty is allowed. Testnet also should be allowed to retarget. The way it's done now, it's EITHER testnet has constant difficulty OR a retarget happens, but not both. I'm 90% sure this is not the case in bitcoin. This is why I'd love to have all bitcoin tests here as well to ensure we're doing this right.

b-yap: i have updated the logic, if you guys can check again.

But i cannot do the `CalculationNextRequired` tests from Bitcoin.  
This is equivalent to the `fn next_work_required(...)` but the method is calling another method from a temporary BlockIndex with `todo!()` implementations.

I have added a `TODO` on the mod tests.

common/src/chain/block/consensus_data.rs
```diff
@@ -0,0 +1,44 @@
...
+pub enum ConsensusData {
+    #[codec(index = 0)]
+    None,
+    #[codec(index = 1)]
+    PoW(PoWData),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub struct PoWData {
+    bits: Compact,
```
iljakuklic, line 15: I know this would deviate from Bitcoin but the difficulty is determined by the time difference of the last two blocks that are at a height that is a multiple of 2016, so this field is not needed. Even if the field is included, we need to check it has the correct value if it's relied upon for the difficulty check. Have you thought about getting rid of it?

b-yap: The bits field is being used in the ff. ways:
1. When `no_retargeting` field in the config is set to `true`. That only happens for Regtest. https://github.com/mintlayer/mintlayer-core/blob/1e390f95555ed9a29c696174fd2f3781c6333761/common/src/chain/pow.rs#L59
2. if it's not due for retargeting yet, and `allow_min_dfficulty_blocks` is set to `false`, which is the Mainnet and Signet: https://github.com/mintlayer/mintlayer-core/blob/1e390f95555ed9a29c696174fd2f3781c6333761/common/src/chain/pow.rs#L65

Let's say the current block height is 8067, and it's not due for retarget (since retarget has happened at 8064). 
I am assuming that the difficulty is retrieved from the previous block, at height 8066. No calculation needed.
If we remove the `bits` field, we'll have to keep calculating the time diff of the blocks of height 8064 and 6048. 

As for checking whether its value is correct, Did you mean the `check_proof_of_work()`? Or do you mean checking whether the previous block's bits field is a correct value?

iljakuklic: > As for checking whether its value is correct, Did you mean the `check_proof_of_work()`? Or do you mean checking whether the previous block's bits field is a correct value?

What I mean is that the value of the `bits` field cannot be trusted to contain the difficulty anyway. I see the `get_work_required` method here but I can't see it being used anywhere. I.e. at any point, the following must pass:

```rust
assert_eq!(calculate_difficulty_from_the_scratch(...).try_into(), Just(this_block.bits))
```

Blocks that do not satisfy that condition must be rejected, since the `check_proof_of_work` function relies on the `bits` field to contain accurate difficulty.

b-yap: The `get_work_required()`(previously `check_for_work_required()`) will be used in the block production.
This was the earliest implementation : https://github.com/mintlayer/mintlayer-core/pull/62/commits/ed8db7d39d28167eadaef7a97fa9602b277c1113#diff-9a00856cffad13af6904b1cb9d4b5a4fd9e0591f10a35a67ddd805141890eeb6R91

But I removed it in this PR, because it's meant for the block production PR (which is currently closed, until Anton's or this is pushed).

Did you mean because the bits is a u32 representation, while the difficulty is 256, that's why the bits field can't be trusted? Or because the previous block can't be trusted, since the bits field can easily be just any number and not from a calculated value during the generation of the 2016th multiple block? 

I haven't checked in bitcoin where that is, but we can actually add that kind of validation.

I somehow do agree with you; removing the bits field is not so bad, since we can always calculate for the difficulty. Your `get_block_id_by_height` function will be useful (Though I think I have to access it through the `BlockIndex`).

For now I've been following how Bitcoin does it. But also I don't have problems diverging from Bitcoin. 
Although I don't know what Sam @TheQuantumPhysicist thinks, we can wait for his opinion.

PR #100, opened 2022-02-10
Adding Block as an outpoint source beside Transaction

zorvan:
Differentiating between outpoints coming from Coinbase and Normal transactions.

### DIFF
--- a/common/src/chain/transaction/input.rs
+++ b/common/src/chain/transaction/input.rs
@@ -5,25 +5,25 @@ use parity_scale_codec::{Decode, Encode};
 
 #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
 pub enum OutpointSource {
-    Transaction(Transaction),
-    Block(Block),
+    Transaction(Id<Transaction>),
+    BlockReward(Id<Block>),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
 pub struct OutPoint {
-    id: Id<OutpointSource>,
+    id: OutpointSource,
     index: u32,
 }
 
 impl OutPoint {
-    pub fn new(prev_tx_id: Id<OutpointSource>, output_index: u32) -> Self {
+    pub fn new(prev_tx_id: OutpointSource, output_index: u32) -> Self {
         OutPoint {
             id: prev_tx_id,
             index: output_index,
         }
     }
 
-    pub fn get_tx_id(&self) -> Id<OutpointSource> {
+    pub fn get_tx_id(&self) -> OutpointSource {
         self.id.clone()
     }
 
@@ -39,7 +39,7 @@ pub struct TxInput {
 }
 
 impl TxInput {
-    pub fn new(prev_tx_id: Id<OutpointSource>, output_index: u32, witness: Vec<u8>) -> Self {
+    pub fn new(prev_tx_id: OutpointSource, output_index: u32, witness: Vec<u8>) -> Self {
         TxInput {
             outpoint: OutPoint::new(prev_tx_id, output_index),
             witness,
--- a/common/src/chain/transaction/input.rs
+++ b/common/src/chain/transaction/input.rs
@@ -9,9 +9,22 @@ pub enum OutpointSource {
     BlockReward(Id<Block>),
 }
 
+impl From<Id<Transaction>> for OutpointSource {
+    fn from(tx: Id<Transaction>) -> Self {
+        OutpointSource::Transaction(tx)
+    }
+}
+
+impl From<Id<Block>> for OutpointSource {
+    fn from(blkrwd: Id<Block>) -> Self {
+        OutpointSource::BlockReward(blkrwd)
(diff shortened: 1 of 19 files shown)


common/src/chain/transaction/input.rs
```diff
@@ -1,22 +1,29 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub enum OutpointSource {
+    Transaction(Transaction),
+    Block(Block),
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
 pub struct OutPoint {
-    id: Id<Transaction>,
+    id: Id<OutpointSource>,
```
TheQuantumPhysicist, line 14: This is not correct. The Idable type is Transaction or Block. Not the enum. So the enum should be:

```
pub enum OutpointSource {
    Transaction(Id<Transaction>),
    Block(Id<Block>),
}
```

common/src/chain/transaction/input.rs
```diff
@@ -1,22 +1,29 @@
+use crate::chain::block::Block;
 use crate::chain::transaction::Transaction;
 use crate::primitives::Id;
 use parity_scale_codec::{Decode, Encode};
 
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub enum OutpointSource {
+    Transaction(Transaction),
+    Block(Block),
```
TheQuantumPhysicist, line 9: Please call this BlockReward

TheQuantumPhysicist:
`TxMainChainPosition` has to be updated to in order to store the UTXO set.

common/src/primitives/consensus_data.rs
```diff
@@ -8,7 +8,6 @@ pub enum ConsensusData {
     #[codec(index = 1)]
     PoW(PoWData),
 }
-
```
muursh, line 11: Is there any reason this blank line has been removed? It looked neater before IMO

common/src/chain/transaction/input.rs
zorvan: You're right. corrected!

common/src/primitives/consensus_data.rs
zorvan: You're right. There was a lot of code there which I had to remove, overlooked that line!

common/src/chain/transaction/input.rs
```diff
@@ -1,22 +1,29 @@
+use crate::chain::block::Block;
 use crate::chain::transaction::Transaction;
 use crate::primitives::Id;
 use parity_scale_codec::{Decode, Encode};
 
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub enum OutpointSource {
+    Transaction(Id<Transaction>),
+    BlockReward(Id<Block>),
+}
```
iljakuklic, line 10: Can we have `From<Id<Transaction>>` and `From<Id<Block>>` for this type?

common/src/chain/transaction/input.rs
```diff
@@ -1,22 +1,29 @@
...
+
 #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
 pub struct OutPoint {
-    id: Id<Transaction>,
+    id: OutpointSource,
     index: u32,
```
iljakuklic, line 15: While we're at it, can we make `index` compact? Having full 4 bytes every time seems a bit wasteful.

zorvan: I guess you mean u16 not Compact as primitive, because the latter is pub struct Compact(pub u32) is still 4 bytes.
U16 makes sense to me,

TheQuantumPhysicist: No. He's talking about scale's serialization. You can specify compact encoding in scale, just like bitcoin's encoding of a vector size.

zorvan:
I get this test failure :

`thread 'chain::transaction::tests::tx_id_collision' panicked at 'decoding outpoint: Error { cause: Some(Error { cause: None, desc: "Could not decode `OutpointSource`, variant doesn't exist" }), desc: "Could not decode `OutPoint::id`" }', common/src/chain/transaction.rs:128:49`

Lukas says :
The point of that test was to find two transactions that hash to the same thing, but that was since fixed so it is not simple to recreate the test in a useful way without temporarily reverting the fix first.
It would be possible to update the test to fit the new encoding but it's quite difficult and would have to be done every time transaction encoding changes, which would be a maintenance nightmare.

should we remove the test?

TheQuantumPhysicist:
That test is failing because of a decoding issue, since you changed the format of transactions. The more important question is, how did you end up with a working test before you did your change if Lukas says it's not possible to reproduce the test after having done the fix? Are you not on the latest master?

zorvan:
Yes, I am on latest master, There was no problem on first commit, I guess didn't check the tests for second commit, I noticed the problem when testing my changes on TxMainChainPosition .

PR #102, opened 2022-02-16
Update `construct_fixed_hash!` to use little-endian

b-yap:
Forking from [parity's fixed_hash](https://github.com/paritytech/parity-common/blob/bcb2e48a7f0e1594b97fdd9086973f046d6bdd5d/fixed-hash/src/hash.rs), only the **_debug_** is in _big-endian_. 

The value stored in the constructed structure (like the `[u8;32]` of `H256`)  follows little-endian format. 
This assumes that the 32-byte array parameter (especially when calling for example `H256::from(...)` or  initializing `H256(...)` ) is  in little-endian format.

To convert H256 to Uint256, I added the `from_le_bytes` and `to_le_bytes`.
Also updated test cases in the `transaction.rs`

### DIFF
--- a/common/src/fixed_hash.rs
+++ b/common/src/fixed_hash.rs
@@ -1,0 +1,805 @@
+// Copyright 2020 Parity Technologies
+//
+// Modified in 2022 by
+//   Carla Yap <carla.yap@rbblab.com>
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+/// Construct a fixed-size hash type.
+///
+/// # Examples
+///
+/// Create a public unformatted hash type with 32 bytes size.
+///
+/// ```
+/// use common::construct_fixed_hash;
+///
+/// construct_fixed_hash!{ pub struct H256(32); }
+/// assert_eq!(std::mem::size_of::<H256>(), 32);
+/// ```
+///
+/// With additional attributes and doc comments.
+///
+/// ```
+/// use common::construct_fixed_hash;
+/// construct_fixed_hash!{
+///     /// My unformatted 160 bytes sized hash type.
+///     #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
+///     pub struct H160(20);
+/// }
+/// assert_eq!(std::mem::size_of::<H160>(), 20);
+/// ```
+///
+/// The visibility modifier is optional and you can create a private hash type.
+///
+/// ```
+/// use common::construct_fixed_hash;
+/// construct_fixed_hash!{ struct H512(64); }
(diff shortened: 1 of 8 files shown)


common/src/uint/impls.rs
```diff
@@ -592,6 +628,57 @@ mod tests {
         );
     }
 
+    #[test]
+    pub fn uint_from_le_bytes() {
```
TheQuantumPhysicist, line 632: Let's discuss the presence of these le functions

common/src/uint/impls.rs
```diff
@@ -135,6 +135,42 @@ macro_rules! construct_uint {
                 res
             }
 
+            pub fn from_le_bytes(bytes: [u8; $n_words * 8]) -> $name {
```
TheQuantumPhysicist, line 138: Let's discuss the presence of these le functions

common/src/fixed_hash.rs
```diff
@@ -0,0 +1,704 @@
+// Copyright 2020 Parity Technologies
+//
+// Modified in 2022 by
+//   Carla Yap <carla.yap@rbblab.com>
```
muursh, line 4: Can you put your Mintlayer email rather than rbb please :)

b-yap: done,

common/src/uint/impls.rs
```diff
@@ -135,6 +135,16 @@ macro_rules! construct_uint {
                 res
             }
 
+            fn _from_le_slice(bytes: &[u8]) -> $name {
```
TheQuantumPhysicist, line 138: Two things
1. Please remove the `_`... this is not python :-)
2. Please restore the the functions you removed; just remove the `le` part to assert it's the default.

common/src/primitives/id.rs
```diff
@@ -29,6 +30,12 @@ impl From<GenericArray<u8, typenum::U32>> for H256 {
     }
 }
 
+impl From<H256> for Uint256 {
```
TheQuantumPhysicist, line 33: Do we have the conversion to the other way?

b-yap: Uint256 to H256? I didn't need it (yet?), so I didn't do one.
If it's needed, I'll do it

TheQuantumPhysicist: In the interest of having full tooling. Please provide both conversions and test them.

altonen: I'm actually with Carla on this one, even if it's a minor thing. I don't think we should have code that we don't need because it's just dead weight with no concrete benefit.

TheQuantumPhysicist: We will need it sooner or later. Even in bitcoin there's both ArithToUint256() and UintToArith256(). Assuming we won't need them is very optimistic. We're now in that mindset, so doing that now will reduce the risk of having to switch context later.

muursh: I'm with Sam here, having it ready to go on the off chance we need it in the future makes sense.

b-yap: added conversion Uint256to H256

TheQuantumPhysicist (approved):

PR #103, opened 2022-02-16
Add support for gossipsub

altonen:


p2p/src/net/mod.rs
```diff
@@ -50,10 +50,13 @@ where
...
     /// One one more peers have expired
     PeerExpired(Vec<AddrInfo<T>>),
+
+    /// Message received from a Gossipsub topic
+    MessageReceived(GossipsubTopic, message::Message),
 }
 
 #[derive(Debug)]
-pub enum GossipSubTopic {
+pub enum GossipsubTopic {
```
TheQuantumPhysicist, line 59: Why is it necessary to have a "topic"? Don't we have a message type already?

p2p/src/net/libp2p/behaviour.rs
```diff
@@ -0,0 +1,399 @@
...
+                    connection,
+                });
+            }
+            Poll::Pending => {}
+        }
+
+        let f: Poll<NetworkBehaviourAction<Self::OutEvent, Self::ProtocolsHandler>> = Poll::Pending;
+        f
+    }
+}
```
TheQuantumPhysicist, line 399: No tests here?

muursh:
I have no major concerns with anything here. `behaviour.rs` does seem to be lacking tests as has already been mentioned but all in all I have no issues

p2p/src/net/libp2p/behaviour.rs
altonen: I honestly don't think there's anything to test. The functions basically return nothing and just call internal behavior objects that have been tested elsewhere so testing this would just be filling up the code coverage quota because apart from one or two functions, I don't get any output from them and even with those functions I don't see any meaningful test cases that would actually benefit us or provide value.

That said, I'm not convinced we need gossipsub. The implementation shown here is kinda hackish and requires doing stuff that you should not be doing (like calling `inject_connected()` directly). The largest benefit of gossipsub is that it builds a mesh from the peers and thus makes the message propagation more intelligent compared to, e.g., `libp2p-floodsub`, which just sends the message to all the peers it's connected to. There's also some attack-resistance mechanisms involved which are not part of `libp2p-floodsub` but because we're going to implement peer profiling/reputation mechanism, we either start duplicating features at two different levels (P2P and libp2p/mock) or we make libp2p shoulder half the load which makes crucial part of the protocol a black box for us.

Gossipsub is objectively better out of the two but it's much more complex than floodsub and provides a lot of stuff we don't need or want. I think Bitcoin uses something very similar to `libp2p-floodsub` so you could make the argument that if it's good enough for them, it's good enough for us. Floodsub would also be much simpler so easier to maintain and also understand and it only does one thing (sending messages) and let's us worry about everything else. We also would not need this ugly `behaviour.rs` at all but could use the default behavior of libp2p.

What do you guys think? I've convinced myself that floodsub is better but there are some arguments in favor of gossipsub.

altonen: Btw I added the floodsub impl for comparison: https://github.com/mintlayer/mintlayer-core/pull/104

p2p/src/net/mod.rs
altonen: We need to be able to subscribe to a topic

altonen:
Had a chat with Sam. We decided it's for the best if we use floodsub instead of gossipsub

PR #105, opened 2022-02-18
Generalize spending and creation of outputs to include block rewards and stake kernels

TheQuantumPhysicist:


common/src/chain/transaction.rs
```diff
@@ -104,38 +104,3 @@ impl Transaction {
         }
     }
 }
-
```
TheQuantumPhysicist, line 107: @iljakuklic should we drop this test? Just double-checking.

common/src/chain/transaction/input.rs
```diff
@@ -1,22 +1,40 @@
-use crate::chain::transaction::Transaction;
+use crate::chain::{block::Block, transaction::Transaction};
 use crate::primitives::Id;
 use parity_scale_codec::{Decode, Encode};
 
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub enum OutPointSourceId {
+    RegularTransaction(Id<Transaction>),
```
altonen, line 7: What other kinds of transactions are we going to have?

muursh (approved):
If @iljakuklic is happy with the test going, it's good to go for me

blockchain_storage/src/store.rs
```diff
@@ -328,7 +320,7 @@ mod test {
     #[test]
     #[cfg(not(loom))]
     fn test_storage_manipulation() {
-        use common::primitives::H256;
+        use common::{chain::SepndablePosition, primitives::H256};
```
altonen, line 323: typo

common/src/chain/transaction.rs
iljakuklic: I'd drop it. The maintenance burden is going to greatly outweigh its usefulness in my view/

common/src/chain/transaction/transaction_index.rs
```diff
@@ -1,10 +1,33 @@
...
 use parity_scale_codec_derive::{Decode, Encode};
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Encode, Decode)]
+use super::Transaction;
+
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode)]
+pub enum Spender {
+    RegularInput(Id<Transaction>),
+    StakeKernel(Id<Block>),
+}
```
iljakuklic, line 13: What's the difference between this and `OutPointSourceId`? Could these be one type?

common/src/chain/transaction/transaction_index.rs
```diff
@@ -188,8 +230,14 @@ mod tests {
         );
         assert_eq!(tx_index.get_output_count(), 3);
 
+        let p = if let SpendablePosition::Transaction(ref p) = tx_index.position {
+            p
+        } else {
+            unreachable!();
+        };
```
iljakuklic, line 237: Very minor comment: I'd personally go for a `match` expression instead of `if let` here. But that's just a style preference

TheQuantumPhysicist: Yeah, it's just a test. That's why I don't care. Usually I prefer match as well.

common/src/chain/transaction/input.rs
TheQuantumPhysicist: Only this. But this is called regular as opposed to block reward, if this is what you're asking. I don't have a preference though.

common/src/chain/transaction/transaction_index.rs
TheQuantumPhysicist: While they look the same, the difference is very subtle. The question related here "is every outpoint type equivalent to every possible input type"? The answer is that outpoint types are a superset of input types, because proof of work, for example, doesn't have inputs (like stake kernels). So it's a design decision whether we ant to distinguish between those. I'm open to suggestions here.

TheQuantumPhysicist:
If there are no other comments/issues, then I'll be merging. If you're happy with my answers to your questions, please let me know so that I could resolve the comment. I'm aiming to get all concerns addressed before merging.

common/src/chain/transaction/input.rs
altonen: In my mind the word regular implied that we'd have irregular or special transactions in addition to regular transactions but in all honesty I don't care what's it called.

iljakuklic (approved):

PR #106, opened 2022-02-19
p2p: Fix libp2p mDNS peer discovery test

altonen:
The test assumed that mDNS would discover one IPv6 address for the
peer which proved to be obviously incorrect on some computers.


muursh (approved):

TheQuantumPhysicist (approved):
I wonder whether it's possible to have zero interfaces. Also, when I reported this issue to you, it was happening randomly on the same machine. Hopefully this is stable enough.

altonen:
You can manually remove all interfaces but I can't think of a single reason why you'd want to do that.

PR #107, opened 2022-02-22
storage: Separate mutable and immutable interfaces

iljakuklic:
* The blockchain_storage crate now takes advantage of the immutabl/mutable separation in the low-level storage framework
* Add traits defining an abstraction over mutiple storage backends
* Add traits abstracting over real and mock implementations of the application-level blockchain storage.
* It is now easier to write generic backend-agnostic and mockery-agnostic (?) code

Comments regarding the PR review process:
* Sorry for a big PR. It conceptually contains two changes but both touch the same bits of code and would take some effort to separate at this point.
* This is a result of many iterations on the design and trait definitions. There may be some residual artefacts from the previous iterations that I might have overlooked. Please have a look and point them out.

## API changes

There have been some changes to the way database transactions are used.

When using transaction objects, this:

```rust
let tx = storage.start_transaction_rw();
// stuff
tx.commit();
```

becomes:

```rust
let tx = storage.transaction_rw();
// stuff
tx.commit();
```

When using scoped transactions, this:

```rust
let result = storage.transaction_rw(|tx| {
    // stuff
    storage::commit(res)
});
```

becomes:

```rust
let result = storage.transaction_rw().run(|tx| {
    // stuff
    storage::commit(res)
});
```

Same applies to read-only transactions.

### DIFF
--- a/blockchain_storage/src/lib.rs
+++ b/blockchain_storage/src/lib.rs
@@ -3,6 +3,7 @@
 use common::chain::block::Block;
 use common::chain::transaction::{Transaction, TxMainChainIndex, TxMainChainPosition};
 use common::primitives::{BlockHeight, Id};
+use storage::traits;
 
 #[cfg(any(test, feature = "mock"))]
 pub mod mock;
@@ -26,22 +27,40 @@ impl From<storage::Error> for Error {
 /// Possibly failing result of blockchain storage query
 pub type Result<T> = core::result::Result<T, Error>;
 
-/// Operations on persistent blockchain data
-pub trait BlockchainStorage {
+/// Queries on persistent blockchain data
+pub trait BlockchainStorageRead {
     /// Get storage version
     fn get_storage_version(&self) -> crate::Result<u32>;
-
-    /// Set storage version
-    fn set_storage_version(&mut self, version: u32) -> crate::Result<()>;
 
     /// Get the hash of the best block
     fn get_best_block_id(&self) -> crate::Result<Option<Id<Block>>>;
 
-    /// Set the hash of the best block
-    fn set_best_block_id(&mut self, id: &Id<Block>) -> crate::Result<()>;
-
     /// Get block by its hash
     fn get_block(&self, id: Id<Block>) -> crate::Result<Option<Block>>;
+
+    /// Get outputs state for given transaction in the mainchain
+    fn get_mainchain_tx_index(
+        &self,
+        tx_id: &Id<Transaction>,
+    ) -> crate::Result<Option<TxMainChainIndex>>;
+
+    /// Get transaction by block ID and position
+    fn get_mainchain_tx_by_position(
+        &self,
+        tx_index: &TxMainChainPosition,
+    ) -> crate::Result<Option<Transaction>>;
+
+    /// Get mainchain block by its height
+    fn get_block_id_by_height(&self, height: &BlockHeight) -> crate::Result<Option<Id<Block>>>;
+}
+
+/// Modifying operations on persistent blockchain data
+pub trait BlockchainStorageWrite: BlockchainStorageRead {
+    /// Set storage version
+    fn set_storage_version(&mut self, version: u32) -> crate::Result<()>;
+
+    /// Set the hash of the best block
+    fn set_best_block_id(&mut self, id: &Id<Block>) -> crate::Result<()>;
 
     /// Add a new block into the database
     fn add_block(&mut self, block: &Block) -> crate::Result<()>;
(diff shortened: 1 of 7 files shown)


muursh (approved):
I have no concerns with any of this. There are a few names I could, maybe, grumble about but given my raw incompetence when it comes to naming things I won't be able to come up with anything better.

iljakuklic:
> I have no concerns with any of this. There are a few names I could, maybe, grumble about but given my raw incompetence when it comes to naming things I won't be able to come up with anything better.

Which names you think should be improved? Maybe we can come up with something better.

blockchain_storage/src/mock.rs
```diff
@@ -48,38 +51,60 @@ mockall::mock! {
         fn del_block_id_at_height(&mut self, height: &BlockHeight) -> crate::Result<()>;
     }
 
-    impl storage::traits::Transactional<'_> for Store {
-        type TransactionRo = MockStoreTx;
-        type TransactionRw = MockStoreTx;
-        fn start_transaction_ro(&self) -> MockStoreTx;
-        fn start_transaction_rw(&self) -> MockStoreTx;
+    impl<'tx> crate::Transactional<'tx> for Store {
```
muursh, line 54: My issue is `Transactional` now I don't really have a good reason to dislike it (maybe it sounds weird as it's an adjective?) but I do. But as I said before I can deal with it

altonen (approved):
Looks good but I can't claim that I fully understand all that's going on

blockchain_storage/src/lib.rs
```diff
@@ -26,23 +27,41 @@ impl From<storage::Error> for Error {
 /// Possibly failing result of blockchain storage query
 pub type Result<T> = core::result::Result<T, Error>;
 
-/// Operations on persistent blockchain data
-pub trait BlockchainStorage {
+/// Queries on persistent blockchain data
+pub trait BlockchainStorageRead {
```
TheQuantumPhysicist, line 31: Is this separation between read and write necessary? This is really unfortunate... Please explain why this is necessary. (I know there's an alias BlockchainStorage that seems to have replaced the old BlockchainStorage and seems to have taken over the same functionality of read/write, but still, is this really necessary?)

iljakuklic: These methods are on the `TransactionRo` and `TransactionRw` types which implement different subsets of these operations. Maybe we could differentiate between mutable and immutable operations using the `mut` anootation, but I don't think it would work on transaction directly. We would have to introduce an intermediate type and have the transaction types `Deref` and `DerefMut` into that type. I may explore this further but there seems to be a tradeoff: either there are two separate traits or there is an extra itermediate type.

TheQuantumPhysicist: Isn't this issue primarily caused by the fact that transactions are either Ro or Rw? I can imagine that write operations can't be attached to Ro because the object expects everything to be immutable.

storage/src/lib.rs
```diff
@@ -81,3 +81,39 @@ pub use transaction::{abort, commit};
 
 pub type Data = Vec<u8>;
 pub type Result<T> = std::result::Result<T, Error>;
+
+#[cfg(test)]
+mod test {
+    use crate::{schema, traits::*};
+
+    struct MyColumn;
+    impl schema::Column for MyColumn {
```
TheQuantumPhysicist, line 90: Based on the discussion yesterday, can we call "Column" a "DBIndex", then define it somewhere as clearly as `map<DBIndex, map<key, value>>`? To me, it's very important that people understand that none of this is even close to a relational database.

TheQuantumPhysicist: Btw, this doesn't have to be changed in this PR. Maybe once this is merged, you can create a subseuent PR with only this change to separate them.

storage/src/lib.rs
```diff
@@ -81,3 +81,39 @@ pub use transaction::{abort, commit};
...
+        const NAME: &'static str = "MyColumnV1";
+        type Kind = schema::Single;
+    }
+
+    type MySchema = (MyColumn, ());
+    type MyStore = crate::Store<MySchema>;
```
TheQuantumPhysicist, line 96: Also, the reason why I don't like the word "Store" specifically, is that it can be both a noun and a verb depending on the context. I have always preferred "Database". "Storage" is also OK. Both are clear and unambiguous. I would feel happier if you chose one of these.

TheQuantumPhysicist (approved):
Besides my comments, everything looks fine. I just think this whole schema thing is overcomplicating things. Maybe we should consider removing it in the future.

blockchain_storage/src/lib.rs
iljakuklic: The issue is (kind of) the fact that the `transaction_{ro,rw}` returns a fresh `Transaction{Ro,Rw}` that the caller takes ownership of. Therefore, the caller can make the transaction object mutable or immutable at will. The best we can do as far as I can see is to introduce an extra intermadiate type (I call it `Handle` until a better name shows up). This new type would implement the storage operations via a single trait that combines `BlockchainStorage{Read,Write}`. The transaction types would give access to it by implementing:
```rust
impl Deref for TranctionRo { target = Handle; ... }
impl Deref for TranctionRw { target = Handle; ... }
impl DerefMut for TranctionRw { target = Handle; ... }
```

I have not tried to implement this yet so may need some modifications to work. Also I am not convinced it is a big improvement over having the trait split in two parts. Any ideas / comments?

storage/src/lib.rs
iljakuklic: I think I am leaning towards removing the schema machinery and partially avoid the issue of coming up with a good name for this.

storage/src/lib.rs
iljakuklic: I wanted to call it `Storage` but then the module should also be called `storage`, which is also a name of the crate. That caused some name resolution clashes in some contexts and was occasionally annoying to work with. I am fine with `Database` but would rather do the change in a subsequent PR, this one is messy enough already I reckon.

blockchain_storage/src/lib.rs
iljakuklic: I experimented a bit with the extra type + `Deref` approach and it seems to me it's not really worth it. We can get rid of the `Read/Write` distinction but the extra complexity that crops up elsewhere basically negates the benefits in my view. We can revisit in the future if this causes issues.

PR #108, opened 2022-02-23
net/mock: Split mock interface into front and backend

altonen:
This is a change which was anyway going to happen (floodsub support) but introducing the split now makes it easier to review the upcoming changes to the network service provider trait as they split the functionality even further. Most of the tests (both unit and integration) are implemented for mock service so this allows to introduce the upcoming changes gradually (so no +1500,-1200 PR that touch 13 different files :smiley: )

### DIFF
--- a/p2p/src/net/mock/mod.rs
+++ b/p2p/src/net/mock/mod.rs
@@ -31,21 +31,25 @@ use std::{
 use tokio::{
     io::{AsyncReadExt, AsyncWriteExt},
     net::{TcpListener, TcpStream},
+    sync::{mpsc, oneshot},
 };
 
-/// This file provides a mock implementation of the network service.
-/// It implements the `NetworkService` trait on top of `tokio::net::TcpListener`
+pub mod backend;
+pub mod types;
 
 #[derive(Debug)]
 pub enum MockStrategy {}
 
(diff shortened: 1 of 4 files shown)


p2p/src/net/mock/mod.rs
```diff
@@ -87,26 +101,26 @@ impl NetworkService for MockService {
...
-                socket: TcpStream::connect(addr).await?,
-            },
-        ))
+        let (tx, rx) = oneshot::channel();
+        self.cmd_tx.send(types::Command::Connect { addr, response: tx }).await?;
+
+        let socket = rx
+            .await
+            .map_err(|e| e)? // channel closed
+            .map_err(|e| e)?; // command failure
```
iljakuklic, line 110: I don't get the `map_err` here. Seems like mapping an identity function should not do anything. Is there some funny thing with type casting going on?

iljakuklic (approved):
FIne with me

muursh (approved):

p2p/src/net/mock/mod.rs
altonen: I prefer this over `let socket = rx.await??;` because it documents the two different errors that can happen when `rx` is polled.

PR #109, opened 2022-02-25
Network event updates

altonen:
This PR splits the `net::Event` enum into two separate enums, `net::ConnectivityEvent` and `net::FloodsubEvent`, which is required for the implementation of separated connectivity and floodsub handling in application-level code. It also temporarily disables libp2p as a network service provider because it's easier to demonstrate the changes I wish to make to the network interface if they are first introduced only to the mock interface which is significantly simpler.

The reason why we want to split these two is that code becomes easier to maintain, especially in the future when all of the peer and syncing-related features have been implemented. Without the split they would all be in `src/lib.rs` which would grow into a very long and unmaintainable file.

### DIFF
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -380,8 +380,8 @@ mod tests {
     use super::*;
     use crate::error::P2pError;
     use common::chain::config;
-    use libp2p::Multiaddr;
-    use net::{libp2p::Libp2pService, mock::MockService};
+    // use libp2p::Multiaddr;
+    use net::mock::MockService;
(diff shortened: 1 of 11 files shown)


muursh (approved):
Seems sensible to me

iljakuklic:
What's the deal with all these commented out bits of code?

altonen:
> What's the deal with all these commented out bits of code?

Disabling libp2p so that the diffs for these PRs are slightly more readable

TheQuantumPhysicist (approved):

PR #113, opened 2022-03-01
feat/h256_displays: fix lowerhex and upperhex displays

b-yap:
The lowerhex and upperhex needs to be reversed too. 
Since the debug uses lowerhex, it's best to reverse it from there, rather from the debug trait.

^ Issue pointed out by @iljakuklic . 

added a test case to all displays.

### DIFF
--- a/common/src/fixed_hash.rs
+++ b/common/src/fixed_hash.rs
@@ -244,7 +244,7 @@ macro_rules! construct_fixed_hash {
 				}
 
 				let ctr = &self.0[0..];
-                for i in ctr.iter().rev() {
+				for i in ctr.iter().rev() {
 					core::write!(f, "{:02x}", i)?;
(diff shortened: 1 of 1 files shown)


common/src/fixed_hash.rs
```diff
@@ -244,7 +242,9 @@ macro_rules! construct_fixed_hash {
 				if f.alternate() {
 					core::write!(f, "0x")?;
 				}
-				for i in &self.0[..] {
+
+				let ctr = &self.0[0..];
+                for i in ctr.iter().rev() {
```
iljakuklic, line 247: The indentation looks weird here and below

iljakuklic (approved):
Looks good other than the indentation issue.

muursh (approved):

PR #114, opened 2022-03-03
Utxo view and its implementations.

b-yap:
This is not ready for review. This is only to make visible the changes so far.

TODO:

- [x] tests for `spend_utxo()`
- [x] tests for fresh, dirty, and spent flags, especially with a parent cache
- [x] tests for `have_utxo_in_cache()`
- [x] implementation of `batch_write`
- [x] tests for `UtxosView`'s methods
- [x] tests for stacked caches
- [ ] check for memory usage in bitcoin
- [x] check the coins_tests.cpp in bitcoin
- [x] flush functionality (rewrite it; do not use the best_block for batch write)


common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+
+    pub fn height(&self) -> BlockHeight {
+        self.height
+    }
+
+    pub fn output(&self) -> &TxOutput {
+        &self.output
+    }
+
+    pub fn block_reward(&mut self, value: bool) {
```
TheQuantumPhysicist, line 92: I don't think we need setters for this struct.

TheQuantumPhysicist: Also, even if we do, this should be called "set_block_reward"

common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// Retrieves utxo.
+    fn get_utxo(&self, outpoint: &OutPoint) -> Option<Utxo>;
+
+    /// Checks whether outpoint is unspent.
+    fn have_utxo(&self, outpoint: &OutPoint) -> bool;
+
+    /// Retrieves the block hash of the best block in this view
+    fn get_best_block_hash(&self) -> Option<H256>;
+
+    /// Estimated size of the whole view (0 if not implemented)
```
TheQuantumPhysicist, line 159: Please use an `Option`. No special values like "0 if not implemented".

common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+                }
+            }
+        }
+
+        self.current_block_hash = Some(block_hash);
+        Ok(())
+    }
+
+    fn derive_cache(&self) -> UtxosCache {
+        self.clone()
```
TheQuantumPhysicist, line 460: Clone? Why? We should create a new instant linked to the parent.

common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+    base: &mut T,
+) -> Result<(), Error> {
+    base.batch_write(cache.utxos, block_hash)
+}
+
+#[derive(Clone, Default)]
+pub struct UtxosCache<'a> {
+    parent: Option<&'a dyn UtxosView>,
+    current_block_hash: Option<H256>,
+    pub(in crate::chainstate) utxos: HashMap<OutPointKey, UtxoEntry>,
```
TheQuantumPhysicist, line 184: Pub... hmmmmm... not so happy about this. If this is used in the chainstate, it'll still be visible to everything in there.

common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+            return Some(res.clone());
+        }
+
+        // if utxo is not found in this view, use parent's `get_utxo`.
+        self.parent.map(|parent| {
+            match parent.get_utxo(outpoint) {
+                None => {
+                    // The parent only has an empty entry for this outpoint; we can consider our
+                    // version as fresh.
+                    UtxoEntry::new_spent(true, false)
```
TheQuantumPhysicist, line 203: I'm confused... trying to understand what this means. So we look in the parent for a utxo, and if it doesn't exist, we create a new one? At least the function name doesn't reflect this, and there's no explanation for why this is the behavior we need.

If we want to indicate that an output is spent, isn't it better to return an enum that says that it's spent?

I really hate same data meaning different things in rust... there's no need to use special values, like 0, to mean something other than the other numbers. So if we `get_utxo_entry` and it doesn't exist, we just say it doesn't exist. The caller can deal with that the way they find appropriate. Is there something I'm missing in this plan?

common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+
+        for (idx, output) in tx.get_outputs().iter().enumerate() {
+            let outpoint = OutPoint::new(id.clone(), idx as u32);
+
+            let overwrite = if check_for_overwrite {
+                self.have_utxo(&outpoint)
+            } else {
+                // TODO: a temporary return of false.
+                // utxobase transactions can always be overwritten, in order to correctly
+                // deal with the pre-BIP30 occurrences of duplicate utxobase transactions.
```
TheQuantumPhysicist, line 239: We don't have BIP30 issue in Mintlayer.

common/src/chainstate/utxo.rs
TheQuantumPhysicist: You can solve this problem by rearranging the modules, since you're making this public specifically for the tests
https://doc.rust-lang.org/reference/visibility-and-privacy.html

common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+pub struct UtxosCache<'a> {
+    parent: Option<&'a dyn UtxosView>,
+    current_block_hash: Option<H256>,
+    pub(in crate::chainstate) utxos: HashMap<OutPointKey, UtxoEntry>,
+    memory_usage: usize,
+}
+
+impl<'a> UtxosCache<'a> {
+    /// returns a copy of the UtxoEntry, given the outpoint.
+    fn get_utxo_entry(&self, outpoint: &OutPoint) -> Option<UtxoEntry> {
```
TheQuantumPhysicist, line 190: I'm really wondering whether there's a difference between this and `UtxosView::get_utxo()`.

common/src/chainstate/utxo.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    pub fn set_best_block(&mut self, block_hash: H256) {
+        self.current_block_hash = Some(block_hash);
+    }
+
+    pub fn add_utxos(
+        &mut self,
+        tx: &Transaction,
+        height: BlockHeight,
```
TheQuantumPhysicist, line 226: There's also a case for adding UTXOs for the mempool. Do we also use height == 0 for that? If yes, then please make the block height either an Option or even more preferably use an enum that specifies a block height OR just indicates that it's in the mempool.

common/src/chainstate/utxo.rs
b-yap: updated,

common/src/chainstate/utxo.rs
b-yap: the `get_utxo_entry` is needed to identify whether the utxo is: 
dirty -> if the utxo version is different from the parent.
fresh -> when the utxo only exists in that view.

the `get_utxo_entry` is needed for `spend_utxo`.

common/src/chainstate/utxo.rs
b-yap: removed comment about BIP30.

common/src/chainstate/utxo.rs
b-yap: I updated this, removing the `new_spent` function. Also created an enum: `UtxoStatus`.

common/src/chainstate/utxo.rs
b-yap: removed.

common/src/chainstate/utxo.rs
b-yap: option doesn't sit right with me... when it's asking for a size. That comment was copied from bitcoin, 
I updated the comment to "0 if empty".

common/src/chainstate/utxo.rs
b-yap: added en enum:
```
pub enum UtxoType {
    /// At which height this containing tx was included in the active block chain
    BlockChain(BlockHeight),
    MemPool,
}
```

The Utxo struct will now look like this:
```
/// The Unspent Transaction Output
#[derive(Debug, Clone, Eq, PartialEq, Encode, Decode)]
pub struct Utxo {
    output: TxOutput,
    is_block_reward: bool,
    /// identifies whether the utxo is for the blockchain or for mempool.
    tx_type: UtxoType,
}
```

common/src/chainstate/utxo.rs
TheQuantumPhysicist: This method is private and isn't for the user.

common/src/chain/block/mod.rs
```diff
@@ -118,6 +118,31 @@ impl Block {
         Ok(block)
     }
 
+    // this function is needed to avoid a circular dependency with storage
```
iljakuklic, line 121: We should think about how the code base is organised into crates to avoid stuff like this.

common/src/chain/transaction/input.rs
```diff
@@ -22,12 +22,50 @@ impl From<Id<Block>> for OutPointSourceId {
...
-#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
 pub struct OutPoint {
     id: OutPointSourceId,
     index: u32,
 }
 
+fn outpoint_source_id_as_monolithic_tuple(v: &OutPointSourceId) -> (u8, H256) {
+    let tx_out_index = 0;
+    let blk_reward_index = 1;
```
iljakuklic, line 33: Minor: these could be constants, i.e. `const TX_OUT_INDEX: u8 = 0;` etc.

iljakuklic: Even more minor: this could be a method `OutPointSourceId::as_monolithic_tuple(&self)`

iljakuklic (commented):
Added a bunch of comments but no serious objections overall.

blockchain_storage/src/store.rs
```diff
@@ -31,6 +36,7 @@ mod well_known {
 
     declare_entry!(StoreVersion: u32);
     declare_entry!(BestBlockId: Id<Block>);
+    declare_entry!(UtxosBestBlockId: Id<Block>);
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Because the utxo view may or may not update its internal state to reflect the true best block. Ideally, they would work in tandem. But in practice, they're not the same thing. In other words, the utxo storage should be able to check whether it's in sync with the best chain.

common/src/chain/block/mod.rs
TheQuantumPhysicist: Agreed. Though keep in mind that the primary reason for not solving this now, apart from that it's not a big deal right now, is that Roy is working on the consensus struct.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+                        }
+                    }
+                    // found entry in the parent cache
+                    Some(parent_entry) => {
+                        if entry.is_fresh() && !parent_entry.is_spent() {
+                            // The utxo was marked FRESH in the child cache, but the utxo
+                            // exists in the parent cache. If this ever happens, it means
+                            // the FRESH flag was misapplied and there is a logic error in
+                            // the calling code.
+                            return Err(Error::UtxoAlreadyExists);
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Maybe. I'm not sure. Though maybe we let the caller decide that.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+            // see bitcoin's Uncache.
+            if !entry.is_fresh && !entry.is_dirty {
+                //todo: decrement the memory usage
+                return self.utxos.remove(key);
+            }
+        }
+        None
+    }
+
+    fn consume(self) -> Result<ConsumedUtxoCache, Error> {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: This is a thinker. I'm really not sure about this one. While I don't really care whether it's finalize or consume, I feel like FinalUtxoCache doesn't really convey what's going on. Unlike crypto stream hasher, where a finalize is understandable to get the result of hashing. I'm happy to hear more opinions about this. I use the same terminology in the Consensus crate, btw for the reason with spent inputs cache.

utxo/src/utxo_impl/simulation.rs
```diff
@@ -0,0 +1,147 @@
...
+                    status: UtxoStatus::Spent,
+                    is_dirty: true,
+                    is_fresh: true,
+                },
+                _ => UtxoEntry {
+                    status: UtxoStatus::Spent,
+                    is_dirty: false,
+                    is_fresh: false,
+                },
+            };
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yeah, sure. Carla wrote this. It was taken from bitcoin as is. I honestly don't care.

utxo/src/utxo_impl/test.rs
```diff
@@ -0,0 +1,497 @@
...
+            assert!(res.is_ok());
+        }
+        Err(expected_e) => {
+            if let Err(actual_e) = res {
+                assert_eq!(expected_e, actual_e);
+            } else {
+                assert!(false);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

utxo/src/utxo_impl/mod.rs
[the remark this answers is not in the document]
iljakuklic: Ok I get it and don't have a strong opinion. It just seems to me that `consume` conveys more about implementation details rather than the essence of the operation.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+                    self.utxos.get_mut(key).and_then(|entry| entry.utxo_mut())
+                }
+            }
+        })
+    }
+
+    /// removes the utxo in the cache with the outpoint
+    pub(crate) fn uncache(&mut self, outpoint: &OutPoint) -> Option<UtxoEntry> {
+        let key = outpoint;
+        if let Some(entry) = self.utxos.get(key) {
```
altonen, line 379: This may be possible to simplified by combining `filter()` and `map()`, not sure if it's worth it though.

utxo/src/utxo_impl/simulation.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::{flush_to_base, UtxosCache, UtxosView};
+use crate::{UtxoEntry, UtxoStatus};
+use common::chain::OutPoint;
+use common::primitives::{Id, H256};
+use crypto::random::{make_pseudo_rng, Rng};
+
+fn random_bool() -> bool {
+    let rng = make_pseudo_rng().gen_range(0..2);
+    rng == 0
+}
```
altonen, line 13: `make_pseudo_rng().gen::<bool>()`?

utxo/src/utxo_impl/test.rs
```diff
@@ -0,0 +1,486 @@
...
+    assert_eq!(add_result, op_result);
+
+    if add_result.is_ok() {
+        let key = &outpoint;
+        let ret_value = cache.utxos.get(key);
+
+        check_flags(ret_value, result_flags, false);
+    }
+
+    println!("the value of the map: {:?}", cache);
```
altonen, line 47: `log::info!(..)`

utxo/src/utxo_impl/test.rs
```diff
@@ -0,0 +1,486 @@
...
+    parent_presence: Presence,
+    cache_presence: Presence,
+    cache_flags: Option<u8>,
+    spend_result: Result<(), Error>,
+    result_flags: Option<u8>,
+) {
+    // initialize the parent cache.
+    let mut parent = UtxosCache::default();
+    let (_, parent_outpoint) =
+        insert_single_entry(&mut parent, &parent_presence, Some(FRESH | DIRTY), None);
```
altonen, line 66: I think this `Some(FRESH | DIRTY)` is much nicer way of expressing the attributes

utxo/src/utxo_impl/test.rs
```diff
@@ -0,0 +1,486 @@
...
+                    is_dirty,
+                    is_fresh,
+                };
+                single_entry_map.insert(key.clone(), entry);
+            }
+        }
+    }
+
+    // perform batch write
+    let sinlge_entry_cache = ConsumedUtxoCache {
```
altonen, line 142: sinlge

utxo/src/utxo_impl/utxo_storage.rs
```diff
@@ -0,0 +1,527 @@
...
+        self.0.get_undo_data(id)
+    }
+}
+
+impl<'a, S: UtxosPersistentStorage> UtxosView for UtxoDB<'a, S> {
+    fn get_utxo(&self, outpoint: &OutPoint) -> Option<Utxo> {
+        match self.0.get_utxo(outpoint) {
+            Ok(res) => res,
+            Err(_) => {
+                //todo: handle errors
```
altonen, line 49: Should this return `Result<Option<Utxo>, Err>` instead?

utxo/src/utxo_impl/mod.rs
TheQuantumPhysicist: Yeah, many things are not optimum in the PR. It's not a priority though. We'll fix the little things we can.

TheQuantumPhysicist: The source of these things is that Carla copied bitcoin, btw.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+        Ok(())
+    }
+
+    /// Flags the utxo as "spent", given an outpoint.
+    /// Returns the Utxo if an update was performed.
+    pub fn spend_utxo(&mut self, outpoint: &OutPoint) -> Result<Utxo, Error> {
+        match self.get_utxo_entry(outpoint) {
+            None => Err(Error::NoUtxoFound),
+            Some(entry) => {
+                let key = outpoint;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Relics... I'll fix them when I see them.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+
+        Ok(())
+    }
+
+    /// Flags the utxo as "spent", given an outpoint.
+    /// Returns the Utxo if an update was performed.
+    pub fn spend_utxo(&mut self, outpoint: &OutPoint) -> Result<Utxo, Error> {
+        match self.get_utxo_entry(outpoint) {
+            None => Err(Error::NoUtxoFound),
+            Some(entry) => {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Fixed

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+                    !curr_entry.is_dirty || curr_entry.is_fresh
+                } else {
+                    // copy from the original entry
+                    curr_entry.is_fresh
+                }
+            }
+        };
+
+        // create a new entry
+        let new_entry = UtxoEntry::new(utxo, is_fresh, true);
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Dirty means it should update the parent; and this case we have to update the parent because the parent doesn't even have it. Fresh means we just created it, so if we spend it in the same round, no need to write it to the parent since the output is created and spent in the same block. Both are optimizations.

And to answer your second concern, ... even better... an enum.

But again, Carla copied bitcoin core... So you're seeing here C++ combined with bitcoin core practices literally translated to rust. 

We'll fix that later. It's not a priority. Once this works properly in tandem with the SpentInputsCache with a working node, we'll be able to clean it up at will.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+        &mut self,
+        utxo: Utxo,
+        outpoint: &OutPoint,
+        possible_overwrite: bool,
+    ) -> Result<(), Error> {
+        let key = outpoint;
+        let is_fresh = match self.utxos.get(key) {
+            None => {
+                // An insert can be done. This utxo doesn't exist yet, so it's fresh.
+                !possible_overwrite
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Generally speaking, a utxo can be overwritten if it's spent. This is motivated in bitcoin by BIP30, where coinbases can have identical hashes. Apart from that issues, there's no harm in overwriting utxos as long they're spent, which I don't think will ever happen. I find it unnecessary because the chance to have a collision is very low, it's there in bitcoin. It's a point of discussion whether we want to drop this feature. Otherwise it may be there for reasons that I still don't comprehend.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+    fn get_utxo(&self, outpoint: &OutPoint) -> Option<Utxo>;
+
+    /// Checks whether outpoint is unspent.
+    fn has_utxo(&self, outpoint: &OutPoint) -> bool;
+
+    /// Retrieves the block hash of the best block in this view
+    fn get_best_block_hash(&self) -> Option<Id<Block>>;
+
+    /// Estimated size of the whole view (None if not implemented)
+    fn estimated_size(&self) -> Option<usize>;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I'm not sure what you're referring to and why Some(0) is interesting at all. If this is about why we don't use 0 as None, it's because I wanna make sure the developer handles that case.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+        }
+    }
+
+    pub fn is_block_reward(&self) -> bool {
+        self.is_block_reward
+    }
+
+    pub fn height(&self) -> &UtxoSource {
+        &self.source
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I changed it to source_height. I think that's OK. I want something to indicate that this is about the height.

Option is not OK because it's not clear what None really means.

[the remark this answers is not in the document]
TheQuantumPhysicist: Resolve if you're happy with this.

utxo/src/utxo_impl/simulation.rs
TheQuantumPhysicist: Welcome to translating things literally from C++ xD

Fixed.

utxo/src/utxo_impl/test.rs
TheQuantumPhysicist: Are you suggesting a specific change? I'm not sure I understand the request or whether it's just a comment with no implicit request.

utxo/src/utxo_impl/utxo_storage.rs
TheQuantumPhysicist: Fair point. I never was told why it was done like this.

TheQuantumPhysicist: The thing is, while it makes sense for a database to do this, I can see that an error from the database is definitely irrecoverable. Besides, the UtxoView doesn't expect the storage to "fail". So a panic may be appropriate here.

utxo/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+#[derive(Debug, Eq, PartialEq)]
+pub enum Error {
+    OverwritingUtxo,
+    UtxoAlreadyExists,
+    UtxoAlreadySpent,
+    NoUtxoFound,
+    CacheWithoutBestBlock,
+    NoBlockchainHeightFound,
+    DBError(String),
+}
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I added thiserror. I'm not sure though I understand what you mean with anyhow errors. In fact, part of me thinks all database errors should be irrecoverable at this level.

utxo/src/utxo_impl/utxo_storage.rs
altonen: I'm fine with a panic if you think it's appropriate.

utxo/src/utxo_impl/test.rs
altonen: This was just a general comment referencing my earlier complaint about how IMO the booleans make the code unnecessarily hard to understand.

utxo/src/utxo_impl/mod.rs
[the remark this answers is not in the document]
altonen: I agree it's not a priority but something to keep in mind

utxo/src/utxo_impl/mod.rs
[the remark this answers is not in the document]
altonen: IIRC previously the value `0` was used to denote that it's not implemented which I preferred. I think a better question is that can the estimated size of the view be zero if it's implemented? If it can be and it's totally valid value for the view then there's no problem. If it can't be zero if an implementation exists, then `Some(0)` is an unnecessary special case we need to be mindful of. Or use `Option<NonZeroUsize>`.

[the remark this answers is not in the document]
TheQuantumPhysicist: This depends primarily on the goal of doing that. If your goal is to optimize for byte size and utilize all the space there is (like in MCUs), then sure, using just a usize will be sufficient. However, my goal is different. My goal is to force the developer to handle all cases, 100% at compile-time. When using an int for two meanings, it's very easy to overlook the case when it's not implemented. If None and 0 are equivalent for a certain caller, a developer can easily do a `unwrap_or(0)`.

utxo/src/utxo_impl/utxo_storage.rs
TheQuantumPhysicist: I'll keep this unresolved because I need to think about it.

utxo/src/utxo_impl/mod.rs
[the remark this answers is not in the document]
altonen: I'm not thinking about optimizations but I am confused because previously, if I remember correctly, `0` was used to denote that implementation doesn't exist which implies that when this estimation has been implemented it cannot return zero (because that would be an error?). Is this a correct assessment/was it ever true?

My point is that if an implementation cannot have a size of zero when it exists, wrapping the return value in an `Option<T>` doesn't make things easier because now you have to match `None`, `Some(0)` (two possible errors) and `Some(_)` which doesn't seem easier to me. I'm not talking about optimizations but I'm hung up on this zero return value.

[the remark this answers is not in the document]
TheQuantumPhysicist: > was it ever true

Yes it was, and it's like that in Bitcoin. I changed it.

> doesn't make things easier because now you have to match

That's exactly the point. "Easier" is not really the goal here. The goal here is "correct". How can we make it correct? We force the developer to handle all possible cases.

Some(0) is not really the interface's problem. It's an implementation problem. I think this is the cause of the confusion. You're mixing implementation issues and interface design. The right answer there is that you test and ensure that your implementation doesn't yield Some(0) through unit tests. But you can't guarantee that all developers who use this interface will handle the 0 case unless you build it in the interface.

This is just like Eq and PartialEq. There's a reason that distinction exists in the interface. PartialEq basically forces you to handle a corner case when NaN != NaN. Here also we want to force the dev to handle a corner case, unlike C/C++ where the dev can just omit this corner case and then when it happens, bad things happen.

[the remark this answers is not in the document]
altonen: Ok so in that case `Some(0)` is never a valid value which is my problem here. Why not use `Option<NonZeroUsize>` in that case? Then the interface makes more sense: either you get `Some(_)` and you know it's valid or you get `None` which means it's not implemented. That's also the safer option because now there's no way to ignore the zero return value because you never get that.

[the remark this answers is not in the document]
TheQuantumPhysicist: That's alright. We can change it to that. My only contention there is that we're probably using too much from the outside. But since it's std::num crate, it's not a big deal.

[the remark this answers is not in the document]
TheQuantumPhysicist: Actually I just noticed that the estimated size can be zero if the view is empty... so 0 is valid. I thought from the first look this is about a UtxoEntry, not a UtxoView.

utxo/src/utxo_impl/utxo_storage.rs
```diff
@@ -0,0 +1,531 @@
...
+    fn new() -> Self {
+        Self {
+            store: BTreeMap::new(),
+            undo_store: HashMap::new(),
+            best_block_id: None,
+        }
+    }
+}
+
+impl UtxosPersistentStorage for UtxoInMemoryDBImpl {
```
altonen, line 125: Unless I missed something, none of these functions ever return an error so it doesn't really make sense to return a `Result<>`, unless it's of course required by the database interface.

altonen (approved):
Left a few comments but LGTM

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,525 @@
...
+
+impl<'a> UtxosView for UtxosCache<'a> {
+    fn get_utxo(&self, outpoint: &OutPoint) -> Option<Utxo> {
+        let key = outpoint;
+        if let Some(res) = self.utxos.get(key) {
+            return res.utxo();
+        }
+
+        // if utxo is not found in this view, use parent's `get_utxo`.
+        self.parent.and_then(|parent| parent.get_utxo(outpoint))
```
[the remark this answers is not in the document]
TheQuantumPhysicist: It's not really recursive because it calls a parent object, not itself. It would be recursive if the parent was recursively created.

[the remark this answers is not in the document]
altonen: But if the parent fails to fetch the UTXO from its cache and calls its own parent, there may be a problem. I don't know if any of that is actually possible.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,525 @@
...
+        let status = self.get_utxo_entry(outpoint)?;
+        match status.status {
+            UtxoStatus::Spent => None,
+            UtxoStatus::Entry(utxo) => {
+                self.utxos.insert(
+                    outpoint.clone(),
+                    UtxoEntry::new(utxo, status.is_fresh, status.is_dirty),
+                );
+                //TODO: update the memory storage here
+                self.utxos.get_mut(outpoint).and_then(|entry| entry.utxo_mut())
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

common/src/chain/block/mod.rs
TheQuantumPhysicist: I'm resolving this because no action can be done right now and it's not a big deal.

utxo/src/lib.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: @iljakuklic if you're satisfied with what you're seeing here, please resolve the comment.

utxo/src/utxo_impl/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I'll keep this with an error message so that we can catch it if it ever happens. I'm not confident about it.

utxo/src/utxo_impl/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: The view is only created at one or two levels at most. So it'll ask the parent, and then the next parent, and then it'll stop. The creation of views is controlled by us, not by a block or a transaction. 

To clearify the purpose, we create a view to isolate the current state from the new state and apply it atomically. We apply the change, then flush it to the parent. So with this design, it's like mostly one parent, and maybe one more parent above that for the persistent storage layer.

Maybe I'm failing to understand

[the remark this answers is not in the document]
TheQuantumPhysicist: I unresolved this to make sure I'm not missing something.

utxo/src/utxo_impl/mod.rs
```diff
@@ -0,0 +1,533 @@
...
+
+        for (idx, output) in tx.get_outputs().iter().enumerate() {
+            let outpoint = OutPoint::new(id.clone(), idx as u32);
+
+            let overwrite = if check_for_overwrite {
+                self.has_utxo(&outpoint)
+            } else {
+                // by default no overwrite allowed.
+                false
+            };
```
iljakuklic, line 249: Minor: this could be simplified to:
```rust
let overwrite = check_for_overwrite && self.has_utxo(&outpoint);
```

iljakuklic (approved):

utxo/src/utxo_impl/mod.rs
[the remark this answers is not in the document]
altonen: This can be closed if you know the call depth is bounded. I was just looking at this function and seeing it call another function of the same name and as `self.parents` type is `Option<&'a dyn UtxosView>` (i.e., also a `UtxoView` implementation) so I was just wondering if there was a possibility of chaining these `get_utxo()` calls into stack overflow.

PR #116, opened 2022-03-04
Swarm management

altonen:
This PR introduces the `SwarmManager` object which takes the responsibility of connectivity and peer management. It's a big PR but most of the changes come from moving almost all code from `src/lib.rs` to `src/swarm.rs` so there is very little new code added.

### DIFF
--- a/p2p/src/swarm.rs
+++ b/p2p/src/swarm.rs
@@ -35,7 +35,6 @@
 use crate::{
     error::{self, P2pError},
     event,
-    event::PeerEventType,
     net::{self, ConnectivityService, NetworkService},
     peer::{Peer, PeerRole},
 };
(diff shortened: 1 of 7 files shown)


p2p/src/swarm.rs
```diff
@@ -0,0 +1,645 @@
...
+    Raw {
+        /// Hashset of IPv4 addresses
+        ip4: HashSet<Arc<T::Address>>,
+
+        /// Hashset of IPv6 addresses
+        ip6: HashSet<Arc<T::Address>>,
+    },
+}
+
+pub struct SwarmManager<T>
```
TheQuantumPhysicist, line 77: Isn't the concept of "swarm" libp2p specific, which means the backend object that controls the connections? Are we deciding here to call any backend object a swarm? The name swarm to me is very confusing... not sure this is the best label. While this isn't a big deal, maybe in the worst case, we should write a little comment explaining what a swarm is, since it's part of our interface.

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
Besides the naming issue, this looks like just moving code.

p2p/src/swarm.rs
altonen: I think the word swarm itself is a very good description for a collection of peers. It's not used only in libp2p but also, e.g., in BitTorrent and it basically just describes a set of peer we know about/are connected to. The problem I have with libp2p's use of the word swarm is that the object they call `Swarm` is not a swarm, but rather a libp2p top-level object that handles everything from swarm management to mDNS peer discovery and floodsub pub/sub functionality. So the word swarm is not descriptive of what it actually does but here our use of the word is IMO very descriptive because this is a manager object that handles all swarm-related functionality and that's all. I'll add a comment about swarm though.

PR #117, opened 2022-03-07
p2p: Do not use full/default features of tokio/libp2p

altonen:
before: `cargo check  442.80s user 48.29s system 917% cpu 53.500 total`
after:  `cargo check  355.23s user 38.61s system 830% cpu 47.424 total`


TheQuantumPhysicist (approved):
You have my blessing!

PR #122, opened 2022-03-23
Added utils crate with newtype macro

royster57:
A convenience macro for implementing the newtype pattern. I'm using it in the mempool and have seen other places where it could be relevant.

### DIFF
--- a/utils/src/newtype.rs
+++ b/utils/src/newtype.rs
@@ -1,0 +1,67 @@
+#[macro_export]
+macro_rules! newtype {
+    ($(#[$meta:meta])* $vis:vis struct $name:ident($wrapped:ty)) => {
+        $(#[$meta])*
+        $vis struct $name($wrapped);
+
+        impl From<$name> for $wrapped {
+            fn from(newtype_instance: $name) -> Self {
+                newtype_instance.0
+            }
+        }
+
+        impl From<$wrapped> for $name {
+            fn from(inner: $wrapped) -> Self {
+                Self(inner)
+            }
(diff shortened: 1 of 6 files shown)


muursh (approved):
I'm okay with this. A part of me wonders if it's worth deriving the obvious traits in `macro_rules! newtype_derive {}`

utils/src/newtype.rs
```diff
@@ -0,0 +1,31 @@
+#[macro_export]
+macro_rules! newtype {
+    ($vis:vis $name:ident($wrapped:ty)) => {
```
iljakuklic, line 3: I wonder if that would be more idiomatic to more closely follow the usual Rust syntax inside the macro. I.e. the syntax to match would be `$vis struct $name($wrapped)` (notice the struct keyword). Maybe even consider a (optional?) semicolon at the end.

iljakuklic: Btw does the `$vis:vis` pattern match empty (i.e. private) visibility?

utils/src/newtype.rs
```diff
@@ -0,0 +1,31 @@
...
+        $vis struct $name($wrapped);
```
iljakuklic, line 4: As per @muursh's comment, I think the macro rule should also match a list of attibutes and expand them before the struct declaration, so we can pass in `#[derive(Blah)]` and whatnot.

iljakuklic (approved):
I'm fine with this. Added some improvement suggestions as code comments.

PR #123, opened 2022-03-25
storage: Add a convenience macro for defining schema

iljakuklic:
This is to avoid having to manually write nested tuples and other boilerplate. Also, to provide a more friendly syntax independent of underlying encoding in terms of types and trait instances.

### DIFF
--- a/blockchain_storage/src/store.rs
+++ b/blockchain_storage/src/store.rs
@@ -31,38 +31,19 @@ mod well_known {
     declare_entry!(BestBlockId: Id<Block>);
 }
 
-// Type-level tags for individual key-value stores:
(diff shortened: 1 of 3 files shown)


TheQuantumPhysicist (approved):
Looks good to me. I'm just hoping we'd change the terminology from Column to DBIndex, for the reasons we discussed before: This isn't an RDBMS.

iljakuklic:
> Looks good to me. I'm just hoping we'd change the terminology from Column to DBIndex, for the reasons we discussed before: This isn't an RDBMS.

Will do that in a follow-up.

muursh (approved):

PR #125, opened 2022-03-28
storage: Rename Column to DBIndex

iljakuklic:
To make the terminology less confusing.

### DIFF
--- a/storage/src/basic.rs
+++ b/storage/src/basic.rs
@@ -12,7 +12,7 @@ pub enum StoreMap {
     Multi(StoreMapMulti),
 }
 
-// Set of store maps, one per column.
+// Set of store maps, one per db index.
 type StoreMapSet = BTreeMap<&'static str, StoreMap>;
 
 // These store changes to the data that could be commited or discarded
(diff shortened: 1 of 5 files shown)


storage/src/basic.rs
```diff
@@ -141,8 +141,8 @@ impl<'st, Sch: Schema> TransactionRo<'st, Sch> {
 impl<'st, 'm, Sch: Schema> crate::traits::GetMapRef<'m, Sch> for TransactionRo<'st, Sch> {
     type MapRef = SingleMapView<'m>;
 
-    fn get<'tx: 'm, Col: schema::Column, I>(&'tx self) -> Self::MapRef {
-        if let Some(StoreMap::Single(store)) = self.store.get(Col::NAME) {
+    fn get<'tx: 'm, DBIx: schema::DBIndex, I>(&'tx self) -> Self::MapRef {
```
TheQuantumPhysicist, line 144: Idx instead of Ix?

TheQuantumPhysicist (approved):

storage/src/basic.rs
muursh: Agreed that Idx is clearer

iljakuklic: Changed

muursh (approved):

PR #130, opened 2022-04-06
[application] Add a subsystem framework

iljakuklic:
TODO

- [x] Review what to write in logs
- [x] Add ability to name subsystems (to show up in logs and diagnostic messages)
- [x] Add more documentation and examples
- [x] Add shutdown timeouts
- [x] Add suport for async subsystem calls
- [x] Add some loom-enabled tests
- [x] Rename `Application` -> `Manager`, crate `application` -> `subsystem`


application/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "application"
+version = "0.1.0"
+edition = "2021"
+
```
muursh, line 5: You need to add a license here to shut cargo-deny up. You can also add `LicenseRef-webpki` to the `allow = []` field for cargo deny too as it's got a known funky but valid license that it's moaning about

application/src/lib.rs
```diff
@@ -0,0 +1,295 @@
+// Copyright (c) 2021 RBB S.r.l
```
muursh, line 1: I guess we should update this to 2021-2022?

application/Cargo.toml
iljakuklic: Added the license in `Cargo.toml`. I don't quite get the cargo-deny comment

muursh: Screenshot from 2022-04-06 18-14-53

I was talking about fixing that warning to make cargo deny cleaner. It's not important though.

muursh: I'd forget about it

application/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+impl<T: Send + 'static> Subsystem<T> {
+    /// Crate a new subsystem handle.
+    fn new(action_tx: mpsc::Sender<Action<T>>) -> Self {
+        Self { action_tx }
+    }
+
+    /// Dispatch a function call to the subsystem
+    pub async fn call_mut<R: Send + 'static>(
```
altonen, line 233: Is it possible to create non-blocking interface so, e.g., P2P could submit a bunch of blocks or maybe a block with PP so it takes non-trivial amount of computation (= time) to validate it and it wouldn't have to block itself until the validation is done but instead would receive like a receipt that the blocks were processed correctly/they were invalid and the connection should be closed?

application/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+        &self,
+        func: impl FnOnce(&mut T) -> R + Send + 'static,
+    ) -> R {
+        let (rtx, rrx) = oneshot::channel::<R>();
+
+        self.action_tx
+            .send(Box::new(move |subsys| {
+                let result = func(subsys);
+                rtx.send(result).ok().expect("Value return channel closed");
```
altonen, line 242: I'm not sure if panicking is the best way to handle a closed channel, isn't this akin to acquiring a poisoned mutex? I think it would be better if the subsystem that received the call and noticed that the calling subsystem has crashed or whatever, it would initiate a graceful shutdown of itself instead of panicking.

altonen (commented):
LGTM apart from calling panic. Also why `[application]` and not `application:`?

application/src/lib.rs
```diff
@@ -0,0 +1,331 @@
+// Copyright (c) 2022 RBB S.r.l
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Honestly, I don't like the name `application`. The `application`, or similar names, should be reserved for something like the `node` application, which already Aaro pointed out.

The `src` will contain the library components of the mintlayer.
The `node` will contain the executable of the node that imports everything it needs from src.

All the technical tools should remain in this crate, while the simple main and relevant imports should go to the `node` crate. This crate should be renamed to something simple... maybe `subsystem`?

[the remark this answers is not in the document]
muursh: Yeah I had this thought too but couldn't think of a name I actually liked

application/src/lib.rs
TheQuantumPhysicist: Maybe there's something I don't understand. But doesn't the `async` keyword do exactly that by returning a future of the result, acting like a receipt that you can poll later?

application/src/lib.rs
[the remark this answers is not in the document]
muursh: My issue with the name `subsystem` is we then have a "subsystem" called `subsystem` so discussing "subsystems" becomes, well as messy as this comment has been.

application/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+    // Signal a subsystem is shutting down. This is taken as an instruction for all subsystems to
+    // shut down. Shutdown completion is detected by all senders having closed this channel.
+    shutting_down_tx: mpsc::Sender<()>,
+    shutting_down_rx: mpsc::Receiver<()>,
+}
+
+impl Application {
+    /// Spin up a new application
+    pub fn new(name: &'static str) -> Self {
+        log::info!("Starting application {}", name);
```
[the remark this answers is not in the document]
iljakuklic: No, not really. Maybe `initialising`?

application/src/lib.rs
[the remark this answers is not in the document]
iljakuklic: I don't like the name either. I can call it `subsystem` but a prominent part of the crate is a top-level type that manages the subsystems so it would be slightly misleading.

application/src/lib.rs
TheQuantumPhysicist: I would rather have the caller decide what to do, instead of having the subsystem acting on its behalf. Generally this is the correct thing to do in all multi-threading based on the fork model. Even in C++, you pass exceptions back to the calling thread and then they're thrown when `std::future::get()` is called, because throwing exceptions (or panics) in threads can have unintended consequences.

I think we should tell the caller "we couldn't do this because the channel is closed", which is returned, maybe with an enum, and then the caller can panic if it wants. This will also make debugging much easier.

application/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+            self.value
+        }
+    }
+
+    #[tokio::test]
+    async fn test_passive() {
+        let logpath: Option<&std::path::Path> = None;
+        logging::init_logging(logpath);
+
+        let app = Application::new("app");
```
TheQuantumPhysicist, line 310: Can you please also add a test for non-passive things, where subsystems interact with one another?

application/src/lib.rs
iljakuklic: I think in this case, the channel outlives the calculation and sending of the return value, but I am not super confident about that. The issue is what happens if you request a subsystem call while other subsystem requests a shutdown. I will construct some tests doing it and run them through `loom`.

application/src/lib.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Maybe rename application to SubsystemManager? Then calling the crate subsystem wouldn't be so bad.

application/src/lib.rs
altonen: That's exactly what happens in P2P: propagate the "channel closed" error to the main task which currently does nothing. I can't think what else we'd do than shutdown the software because unless we implement restarting mechanisms for each subsystem, it's highly unlikely that the program is usable after one of its "microservices" has crashed. So even if P2P doesn't care that chainstate is down because it can still send TXs to mempool, in the grand scheme of things the software is in such a state that it might as well restarted, that's at least how I see. Anyway, whatever we choose to do, I still think panicking here is not the correct approach.

application/src/lib.rs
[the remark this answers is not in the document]
iljakuklic: > Maybe rename application to SubsystemManager? Then calling the crate subsystem wouldn't be so bad.

This sounds good to me. I think we can then call it just `Manager` and refer to it by `subsystem::Manager` from other crates.

application/src/lib.rs
TheQuantumPhysicist: > The issue is what happens if you request a subsystem call while other subsystem requests a shutdown

There should never be any race conditions in this design, otherwise we're doing it wrong. That's actually one of the design goals here. The receiver can only process one request at a time. So a shutdown request is received, all subsequent requests will simply be denied. The receiver will then will simply wait for the queue to become empty then shutdown.

application/src/lib.rs
altonen: Unless I'm reading the code incorrectly, that's not really what happens here. The code is asynchronous on "executor-level" but not on "application-level" meaning while my task is able to sleep and let somebody else run because I'm polling a future, I cannot progress my own state because I'm polling the future, if that makes sense. This is because calling the function (`mpsc::Sender::send()`) and fetching the return value (`oneshot::Receiver:recv()`) happen inside this function so I can't separate the "making a request" and "getting a response" in my application logic. If it would instead return the oneshot receiver, then at least I could decided myself when I want that to poll the future that gives me the return value but as you probably can imagine, it becomes really messy with multiple futures.

If there was instead an API like `submit_work(work_id: u64, work: vec![..])` and I would have an RX channel that would poll for `WorkResponse { work_id: u64, status: WorkStatus }`, then I could asynchronously submit work to mempool and chainstate and get the response back when they are ready, without having to poll the return value future in my application logic.

This is at least how I understood the current code.

TheQuantumPhysicist: Maybe I'm also ignorant enough to be saying nonsense. But here's what I understand:

When we call this function, it immediately returns and gives a future object. 

> I cannot progress my own state because I'm polling the future, if that makes sense

Once you have the future object, polling it isn't a blocking operation unless you call "await". You can keep asking it whenever you want "hey, buddy, are you ready?", and if not, you can do whatever other things you want to do. Once it's ready, it can be either an error or a result. At least this is how I did futures before. I'm failing to understand the problem, but maybe you know this better than me.

TheQuantumPhysicist: > I can change this to return the oneshot receiver to give the caller more flexibility if desired.

While I don't understand the real problem still, this seems to be the wrong abstraction for this problem.

TheQuantumPhysicist: I think we should have a call about this problem. I'm not happy about exposing the internals of a subsystem to the contents of other subsystems, or maybe that's not even what's being proposed here. For example, we should totally be able to change the channels' types to another kind of channels without having to touch any users of the subsystems. Let's have that call at some time and maybe you guys can make me understand the issue better.

iljakuklic: The idea was to make the approach easiest for the user (i.e. one that is the closest to the regular function calls) the default one. The P2P system is somewhat different as it is highly concurrent already. There is not an issue having both in the API here as far as I'm concerned.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+        name: &'static str,
+        subsystem: impl 'static + Send + FnOnce(CallRequest<T>, ShutdownRequest) -> F,
+    ) -> Subsystem<T> {
+        // App name
+        let appname = self.name;
+        // Shutdown-related channels
+        let shutting_down_tx = self.shutting_down_tx.clone();
+        let shutdown_rq = ShutdownRequest(self.shutdown_request_tx.subscribe());
+        // Call related channels
+        let (action_tx, action_rx) = mpsc::channel(64);
```
TheQuantumPhysicist, line 95: Can we make this 64 a parameter? I think there should be a configuration struct for everything related (?)

subsystem/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+            .await
+            .ok()
+            .expect("Target subsystem down");
+
+        rrx.await.expect("Target subsystem down")
+    }
+
+    /// Dispatch a function call to the subsystem (convenience method if subsystem is not mutated)
+    pub async fn call<R: Send + 'static>(&self, func: impl FnOnce(&T) -> R + Send + 'static) -> R {
+        self.call_mut(|this| func(this)).await
```
TheQuantumPhysicist, line 253: Right now they're both the same. Later, we need to allow read-only calls to be done in paralle, but read/write calls to be done uniquely. Just like an rw-lock. While this isn't a requirement right now, you're free to plan for it now.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        // Drop the unused shutdown request end
+        std::mem::drop(self.shutdown_request_rx);
+        // Signal the manager is shut down so it does not wait for itself
+        std::mem::drop(self.shutting_down_tx);
+
+        log::info!("Subsytem manager {} shutting down", self.name);
+
+        // Order all the remaining subsystems to shut down.
+        let _ = self.shutdown_request_tx.send(());
```
TheQuantumPhysicist, line 170: Can't we do this tx/rx of shutdown with a clean future without exposing the internal mechanism?

subsystem/src/lib.rs
iljakuklic: Aye this could be done

subsystem/src/lib.rs
iljakuklic: Not sure how this would be done with a plain future. Also I don't see how this exposes internal mechanisms in any significant way. The user does not have direct access to this channel.

application/src/lib.rs
iljakuklic: The conclusion from the call was to keep the interface to the `call` method as it is now. It's a responsibility of the client to spawn tasks appropriately if more concurrency is required. If that becomes too burdensome, we can expose the channel-based API too at some point.

subsystem/src/lib.rs
TheQuantumPhysicist: Maybe there's an angle I'm not seeing. But assuming that const-correctness is held, we can guarantee that no side effects will happen. At least that's one of my design goals.

iljakuklic: Looking at it now, I just realised: The subsystem is already in control what to pass into the function as the argument. I.e. in the code snippet `subsys.call(|this| this.foo(5))`, the subsystem decides what to pass in as `this`. Passing in a reference is a relic of the times when only passive subsystems were supported, where a mutable reference to the subsystem object was always passed. Currently, the subsystem can pass a mutable reference to a value of its own choosing. It makes sense to me to get rid of the reference and let the subsystem pass in a value of any type `T`. We can recover the original behaviour by setting `T = &mut MySubsystem`. This makes some parts of the code a bit less awkward and also removes the need for this distinction between `call` and `call_mut`.

subsystem/src/lib.rs
TheQuantumPhysicist: I was thinking of using that, but thought it's a little of an over kill... better than nothing though.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        self.action_tx
+            .send(Box::new(move |subsys| {
+                let result = func(subsys);
+                rtx.send(result).ok().expect("Value return channel closed");
+            }))
+            .await
+            .ok()
+            .expect("Target subsystem down");
+
+        rrx.await.expect("Target subsystem down")
```
altonen, line 309: You mentioned in the call that you were working on nested calls but I was wondering can this deadlock the blockchain? I don't know if it can realistically happen with how a blockchain generally works but if subsystem calls can themselves execute subsystem calls, is it possible that A->B->C->A call chain deadlocks the program? So subsystem A calls B which, before returning a value to A, calls C which in turn, before returning a value to B, calls A. Or maybe A calls B and as B is calling C, C independently calls A (as they are scheduled to execute in parallel).

Can any of this actually happen or is this just unwarranted paranoia?

altonen: I think if we could somehow rule out cyclic calls entirely that would be very nice so I would prefer option 1. I don't know if that is possible as we have two input sources running in parallel (RPC and P2P) so them independently interacting with the blockchain can create a cycle, maybe? Perhaps we could prove with TLA+ that it doesn't happen but I fear we don't have the time :smiley: 

I think the problem with 3 is that it can make refactoring a really scary process because even a hint of carelessness you can create a deadlock.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    /// Run the application main task.
+    ///
+    /// Completes when all the subsystems are fully shut down.
+    pub async fn main(mut self) {
+        // The main task just waits for the shutdown signal and coordinates the cleanup at the end.
+        // All other functionality is performed by subsystems.
+
+        // Install the Ctrl-C handler as a subsystem.
+        #[cfg(not(loom))]
```
royster57, line 97: Why do we need this annotation?

subsystem/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        self.start("ctrl-c", |mut call_rq, mut shutdown_rq| async move {
+            tokio::select! {
+                ctrl_c_signal = tokio::signal::ctrl_c() => {
```
royster57, line 100: What will happen when the user presses ctrl-c? Nothing?

subsystem/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+
+/// Shutdown request
+pub struct ShutdownRequest(broadcast::Receiver<()>);
+
+impl ShutdownRequest {
+    /// Receive a shutdown request.
+    pub async fn recv(&mut self) {
+        match self.0.recv().await {
+            Err(broadcast::error::RecvError::Lagged(_)) => {
+                panic!("Multiple shutdown broadcast requests issued")
```
royster57, line 260: Why is this a panic? Is it a logic error for several subsystems to independently signal shutdown? In general, what will make a subsystem want to shut down?

subsystem/src/lib.rs
iljakuklic: Loom does not emulate signals (or any other IO for that matter), so does not implement the `Ctrl-C` handling. Therefore, if you want to run something under loom, it can only spawn tasks that communicate through the synchronisation primitives provided by loom. No keyboard input, no network, no file IO, no signals etc.

iljakuklic: Now thinking about it, instead of disabling the subsystem under loom, we could make it an empty subsystem that immediately terminates. Then loom would try to schedule it at every possible point, effectively emulating trying to hit `Ctrl-C` at every point. Will give it a go and see what happens. Just note that loom does not currently handle this crate properly because the channels are not swapped out for the loom version when compiled for loom.

subsystem/src/lib.rs
royster57: Can I test this out? It seems that there is no actual handling of the signal. Where is the code that will cause the subsystem to terminate? As far as I can tell `tokio::signal::ctrl_c` is a future that is resolved when ctrl_c is pressed, but we don't do anything here once that happens (except for log if the future resolved to an error).

subsystem/src/lib.rs
royster57: Not sure I quite got it... It seems to me like the meaning of the annotation is that when running loom tests, the call to `start` should be skipped?
What would be the consequence of omitting the annotation?

iljakuklic: > Not sure I quite got it... It seems to me like the meaning of the annotation is that when running loom tests, the call to `start` should be skipped? What would be the consequence of omitting the annotation?

You would not be able to compile this with loom

subsystem/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    name: &'static str,
+
+    // Order all subsystems to shutdown
+    shutdown_request_tx: broadcast::Sender<()>,
+    shutdown_request_rx: broadcast::Receiver<()>,
+
+    // Signal a subsystem is shutting down. This is taken as an instruction for all subsystems to
+    // shut down. Shutdown completion is detected by all senders having closed this channel.
+    shutting_down_tx: mpsc::Sender<()>,
+    shutting_down_rx: mpsc::Receiver<()>,
```
royster57, line 42: It seems all we do with `shutting_down_request_rx` is drop it in main. Why do we need to keep it at all? What would happen if we were to not store this receiver handle in the `Manager`?

subsystem/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            log::info!("Subsystem {}/{} started", manager_name, subsys_name);
+
+            // Perform the subsystem task.
+            subsystem(call_rq, shutdown_rq).await;
+
+            // Signal the intent to shut down to the other parts of the application.
+            let _ = shutting_down_tx.send(()).await;
+
+            // Close the channel to signal the completion of the shutdown.
+            std::mem::drop(shutting_down_tx);
```
royster57, line 210: Won't this be dropped anyhow once it goes out of scope in a few more lines?

subsystem/src/lib.rs
iljakuklic: Maybe it could be removed, let me check. Dropping this immediately after channel creation would close the channel which is taken as a signal that the system has terminated, so not sure if this would trigger a premature shutdown.

subsystem/src/lib.rs
royster57: I removed the annotation and clippy and cargo build still succeed...

royster57: Never mind, the loom tests don't compile.

subsystem/src/lib.rs
iljakuklic: This seems to have some interesting effects. If the main task does not hang on to its end of the channel, the behaviour when there are no subsystem changes. Before it would just wait because it is waiting for a subsystem to signal termination. If there is no such subsystem there is nobody to tell the whole system to stop. With some code changes, we can make it so that the main thread just exits immediately if there are no subsystems, which seems like a cleaner solution.

iljakuklic: I have now implemented the cleaner shutdown semantics with respect to no subsystems scenario. Also the Ctrl-C handler now has to be explicitly enabled so it does not get in the way in tests etc.

subsystem/src/lib.rs
```diff
@@ -81,6 +98,33 @@ impl Manager {
         self.builder().with_name(name).start_passive(obj)
     }
 
+    /// Install termination signal handlers.
+    ///
+    /// This adds a subsystem that listens for the Ctrl-C signal and exists once it is received,
```
royster57, line 103: exists -> exits

subsystem/tests/helpers/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author(s): L. Kuklinek
+
+static INIT: std::sync::Once = std::sync::Once::new();
+
+pub fn init_test_runtime() -> tokio::runtime::Runtime {
```
royster57, line 20: This can be pub(crate)

royster57:
Looks good to me, I added another couple of minor comments.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+            // Perform the subsystem task.
+            subsystem(call_rq, shutdown_rq).await;
+
+            // Signal the intent to shut down to the other parts of the application.
+            shutting_down_tx.send(()).await.expect("Subsystem outlived the manager!?");
+
+            log::info!("Subsystem {}/{} terminated", manager_name, subsys_name);
+
+            // Close the channel to signal the completion of the shutdown.
+            std::mem::drop(shutting_down_tx);
```
royster57, line 243: Like I mentioned in our call, I personally prefer not have calls to `std::mem::drop` unless they are strictly necessary, since whenever I see such a call I wonder why it is there. I would remove the call and modify the comment to say something like "shutting_down_tx is dropped at the end of this scope, thus signalling the completion of the shutdown". I understand your desire to be explicit though, just stating my point of view.

subsystem/tests/helpers/mod.rs
iljakuklic: Frankly, I don't see the point. This is not a library so will not be "imported" elsewhere. In addition to that, it is in a private module so would not be accessible if you imported it as a library anyway.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+            Ok(()) => (),
+        }
+    }
+}
+
+/// Subsystem handle.
+///
+/// This allows the user to interact with the subsystem from the outside. Currently, it only
+/// supports calling functions on the subsystem.
+pub struct Subsystem<T> {
```
iljakuklic, line 305: While this type represents an interface to a subsystem, it is not in itself a subsystem. I am considering renaming it to `Handle` so it would be referred to as `subsystem::Handle` from other crates. Would you consider it an improvement?

application/src/lib.rs
iljakuklic: Calls now return a `Result` so the caller can detect if it's shouting at a zombie subsystem

subsystem/src/lib.rs
iljakuklic: There was no feedback so I decided to go ahead with this for the time of being. Any objections let me know.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,497 @@
...
+
+/// Defines hooks into a subsystem lifecycle.
+#[async_trait::async_trait]
+pub trait Subsystem: Sized {
+    /// Custom shutdown procedure.
+    async fn shutdown(self) {}
+}
+
+/// Manager configuration options.
+pub struct ManagerConfig {
```
TheQuantumPhysicist, line 54: May I suggest moving things around and making this design more modular? Maybe the manager can be moved to a submodule. The file is already 500 lines.

subsystem/src/lib.rs
```diff
@@ -0,0 +1,497 @@
...
+
+impl SubsystemConfig {
+    const DEFAULT_CALL_QUEUE_CAPACITY: usize = 64;
+    const DEFAULT_SUBSYSTEM_NAME: &'static str = "<unnamed>";
+
+    /// New configuration with given name, all other options are defaults.
+    fn named(subsystem_name: &'static str) -> Self {
+        Self {
+            subsystem_name,
+            ..Default::default()
```
TheQuantumPhysicist, line 338: I get that the intention of this function is to just define a configuration just by name. However, I'm allergic to these blanket defaults... I always imagine someone adding some other configuration field, and then it getting assigned by default and causing problems. I would rather get a compiler error, and have the developer attend to it and choose the correct value instead of a possible unintended default that could create problems.

My ask: Please change this to: `call_queue_capacity: Default::default()`.

subsystem/tests/helpers/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    INIT.call_once(|| logging::init_logging::<&std::path::Path>(None));
```
TheQuantumPhysicist, line 21: Does this reset for every individual test, give that it's declared globally static?

subsystem/src/lib.rs
iljakuklic: Yep very good suggestion. Will do. Although moving files around while a PR is open makes reviewing more difficult, so will only go ahead if you and others are fine with doing it now.

TheQuantumPhysicist (approved):
Besides the minor issues I mentioned, I think the PR looks fine.

subsystem/tests/helpers/mod.rs
iljakuklic: No it's only once per test binary run and one run of test binary may run multiple tests. Calling `init_logging` multiple times (e.g. for every test) causes a panic, that's why it's only initialised once using `std::sync::Once`.

subsystem/src/lib.rs
iljakuklic: Actually, I can use my time zone advantage and do it today in the evening, when others are unlikely to be looking at it.

subsystem/src/lib.rs
TheQuantumPhysicist: Then please set a default value in a different way.

iljakuklic: How about now?

subsystem/tests/helpers/mod.rs
TheQuantumPhysicist: Perfect!

PR #138, opened 2022-04-18
Initial syncing implementation

altonen:
This PR provides an initial implementation for syncing. Currently the very basics work so connecting to other peers, exchanging header information with them, downloading missing blocks and eventually importing them to the local block index. Here's a short list of stuff that still needs to be implemented:
- partial draining (now the import queue is drained only after everything has been downloaded)
- peer reputation and profiling
- memory reclamation and general optimizations to peer indices
- some other stuff that I have forgotten

There's also a bunch of minor TODOs that I need to fix but those are the big ones and none of them are critical blockers. Let me know what you think!


p2p/src/net/mock/floodsub.rs
```diff
@@ -0,0 +1,549 @@
...
+    ),
+
+    /// Channels for receiving messages from the sockets
+    msg_chan: (
+        mpsc::Sender<(SocketAddr, FloodsubMessage)>,
+        mpsc::Receiver<(SocketAddr, FloodsubMessage)>,
+    ),
+
+    /// All messages that have been received and forwarded to frontend
+    seen_messages: HashSet<u64>,
```
TheQuantumPhysicist, line 75: This is not necessary. In general there are much easier ways to ddos a node than sending the same message again and again.

p2p/src/sync/index.rs
```diff
@@ -0,0 +1,413 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author(s): A. Altonen
+use crate::{
+    error::P2pError,
+    sync::{
+        mock_consensus::{BlockHeader, BlockId},
```
TheQuantumPhysicist, line 20: There's already a BlockHeader defined in common

p2p/src/sync/index.rs
```diff
@@ -0,0 +1,413 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct InnerPeerIndex {
+    pub id: BlockId,
+    pub prev_id: Option<BlockId>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct PeerIndex {
```
TheQuantumPhysicist, line 45: Can you please document these structs and functions? It's really difficult to understand the context and purpose from the outside... so one has to dig deep to see what's going on.

Is this supposed to be the "PeerState"?

p2p/src/sync/peer.rs
```diff
@@ -0,0 +1,405 @@
...
+use tokio::sync::mpsc;
+
+/// State of the peer
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum PeerSyncState {
+    /// Peer state is unknown
+    Unknown,
+
+    /// Peer is uploading blocks to local node
+    UploadingBlocks(mock_consensus::BlockHeader),
```
TheQuantumPhysicist, line 36: Maybe I'm confused and don't understand how this works, but I think this is in the wrong place. The connection state should be irrelevant in the peer state. The problem there is that we're mixing connection state with blockchain state. Now I understand that for an async system this is difficult to manage because we have to know both so that not multiple sync processes execute at the same time, but also these are two different things and probably shouldn't be at the same level.

In other words: Whether "I'm currently doing something for this peer" shouldn't be at the same level of "this peer's tip is this/that block".

Again, maybe there's something I still don't understand.

p2p/src/sync/queue.rs
```diff
@@ -0,0 +1,707 @@
...
+
+    /// Element resolves a dependency and all elements can be queried
+    Resolved,
+
+    /// Element was attempted to be reinserted
+    Duplicate,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ImportQueue<T>
```
TheQuantumPhysicist, line 86: Please document what this is used for and its place in the sync process

p2p/src/sync/queue.rs
```diff
@@ -0,0 +1,707 @@
...
+            });
+            self.lookup.remove(data);
+
+            return Some(resolved);
+        }
+
+        None
+    }
+
+    /// Get all non-orphan chains from the import queue and clear all unresolved data
```
TheQuantumPhysicist, line 215: Been wondering whether this is about orphan blocks... take a look at the OrphansBlockPool container in Consensus. You may have reinvented the wheel.

p2p/src/sync/mod.rs
```diff
@@ -73,182 +127,2369 @@ where
...
                 }
             }
+
+            // advance the state of the manager after each event has been handled
+            self.advance_state().await.into_fatal()?;
         }
     }
 }
 
 #[cfg(test)]
```
TheQuantumPhysicist, line 629: Huge tests... maybe move them to their own file?

p2p/src/net/mock/floodsub.rs
altonen: This is actually not for DDoS prevention but just that I don't spam the front-end with messages that it has already seen because for the "happy path" it has no use for them. But thinking about it again made me realize we need to forward also the duplicate messages so we can implement proper peer reputation.

As an aside though, this is the mock interface so IMO there could be a WinNuke here for all I care and fixing it wouldn't have a high priority unless it made the upper-level protocol implementation incorrect. But yeah, I should strive for correctness either way.

p2p/src/sync/index.rs
altonen: I know. This is a bit of a can of worms because on one hand, having the mock block index implementation allows us to have self-contained P2P + syncing implementation that can be tested without connecting to other subsystem but if we want to make syncing generic over the chainstate it uses and chainstate generic over the P2P implementation it uses, it will be become complicated.

p2p/src/sync/index.rs
altonen: I think this piece of code is the one that'll evolve the most during the next few weeks so I wouldn't concentrate on this too much. This is the intermediary index that holds the peer's chain but at the moment it's not optimized and doesn't even serve much of a purpose so it'll be reworked quite a bit.

p2p/src/sync/mod.rs
altonen: Yeah some of them don't fall into the unit testing category so I'll see if I can move them somewhere else.

p2p/src/sync/queue.rs
altonen: I think it's different for two reasons. Firstly, all of the blocks have a block source different than `BlockSource::Local` so in `process_block()` error would be returned which is not something we want. Secondly the blocks are "orphan" only for a very brief moment in time and only because we download multiple blocks simultaneously and they may be received out of order. If we were to download them in order and one by one, there would be only valid blocks and actual orphan blocks in which case we could just directly import them to the chainstate. Btw, Substrate also uses an import queue for the block syncing, probably to solve this exact same problem.

altonen:
Here's a new TODO I'll be working on:

- download blocks for each peer sequentially (essentially remove the import queue and SyncManager's `active` queue)
- switch `mock_consensus.rs` to use structures defined in `common/`

How should we deal with floodsub events? Right now blocks received from the floodsub get pushed into the import queue which orders them correctly with other incoming blocks and they all get drained into the block index without problems. If we now remove the import queue and download blocks sequentially, how do we deal with floodsub? We either don't subscribe to the topics until we are fully synced but at that point we may already be receiving "orphan" blocks from the floodsub, depending on how long it took to sync our own node and we keep "re-syncing" until what exactly, or we subscribe to it as soon as we start syncing but then we are back to the ordering problem, which was already solved. Any thoughts?

altonen:
Here's btw a "cherry-picked" syncing implementation that only contains the important parts. Should be much easier to get the high-level picture of syncing and the missing parts are basically tests. Without comments it's like 570 lines. `self.p2p_handle` is the subsystems caller wrapper handle to chainstate.

https://gist.github.com/altonen/72eb9574a039589219d77a4899cb9d4f

TheQuantumPhysicist:
> Here's btw a "cherry-picked" syncing implementation that only contains the important parts. Should be much easier to get the high-level picture of syncing and the missing parts are basically tests. Without comments it's like 570 lines. `self.p2p_handle` is the subsystems caller wrapper handle to chainstate.
> 
> https://gist.github.com/altonen/72eb9574a039589219d77a4899cb9d4f

Is there a reason why this didn't go to the branch? I would love to see it in action... feel free to create another branch/PR.

altonen:
It can't go to the branch because it doesn't work on its own but essentially if I remove all P2P scaffolding/tests, that is what we're left with. I don't understand why that code would be in this branch or any other given that it doesn't work without help.

PR #139, opened 2022-04-18
Use env_logger everywhere

altonen:
P2P is not fixed in this commit as the prints were converted to use env_logger in the syncing PR

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -1,3 +1,3 @@
 fn main() {
-    println!("Hello, world!");
+    logging::log::info!("Hello, world!");
 }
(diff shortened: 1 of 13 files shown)


muursh (approved):

node/src/main.rs
```diff
@@ -1,3 +1,3 @@
 fn main() {
-    println!("Hello, world!");
+    logging::log::info!("Hello, world!");
```
TheQuantumPhysicist, line 2: 😄

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
Let me also fix something in the logger

PR #140, opened 2022-04-19
feat/undo: first iteration

b-yap:


utxo/src/undo.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn into_inner(self) -> Vec<Utxo> {
+        self.0
+    }
+}
+
+
+#[derive(Debug, Eq, PartialEq, Encode, Decode)]
+pub struct BlockUndo(Vec<TxUndo>);
+
+impl From<Vec<TxUndo>> for BlockUndo {
```
TheQuantumPhysicist, line 27: I'm not a big fan of this kind of meaningless conversion... `From` is meant to be kind of isomorphic or epimorphic conversion of data from one type to another (I hope I'm using these names right)... the point is, just because by sheer coincidence the current content of a type is another type, doesn't mean it represents it. To illustrate the issue, if we decide to have the block height baked into Undos in order to test blocks that are not in the main chain, this function will just break.

b-yap: changed to `fn new()`

utxo/src/undo.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    pub fn into_inner(self) -> Vec<Utxo> {
+        self.0
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Encode, Decode)]
+pub struct BlockUndo{
+    // determines at what height this undo file belongs to.
+    height:BlockHeight,
```
TheQuantumPhysicist, line 39: Please make sure to format all the files before merging this.

TheQuantumPhysicist (approved):

utxo/src/undo.rs
b-yap: done.

PR #145, opened 2022-04-22
testing: Make loom optional on per-package basis

iljakuklic:
For now, loom is only enabled for `storage` and `blockchain-storage`.

### DIFF
--- a/.github/workflows/loom.yml
+++ b/.github/workflows/loom.yml
@@ -28,4 +28,4 @@ jobs:
       with:
         submodules: recursive
     - name: Run permutation tests
-      run: cargo test --tests --release --workspace
+      run: cargo test --tests --release -p storage -p blockchain-storage
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -14,7 +14,6 @@
 // limitations under the License.
 //
 // Author(s): A. Altonen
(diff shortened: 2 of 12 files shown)


muursh (approved):

TheQuantumPhysicist (approved):
I would more advocate for everything to be explicitly taken off loom to ensure everything is tested; in fact, I would even advocate for working more on why they're not working, assuming loom doesn't have bugs. I know it's looking like a problem now for every new crate, but all there's to this is getting used to it. 

I'm approving, but let's keep restoring this option on the table.

iljakuklic:
> I would more advocate for everything to be explicitly taken off loom to ensure everything is tested; in fact, I would even advocate for working more on why they're not working, assuming loom doesn't have bugs. I know it's looking like a problem now for every new crate, but all there's to this is getting used to it.
> 
> I'm approving, but let's keep restoring this option on the table.

The main issue is that loom only supports "pure" concurrency, so anything to do with IO, timers, networking, etc. cannot be run under loom. This mainly applies to `p2p` and `rpc`. Then there is also the consideration that some crates have no concurrency (for exmaple `script` and `common`), so there is no benefit in running them using loom.

TheQuantumPhysicist:
Go ahead and merge it. We can discuss this later in detail.

PR #149, opened 2022-04-26
rpc: Update the RPC mechanism to work with subsystems

iljakuklic:
Some changes to the RPC module:
* Turn RPC into a subsystem
* Hide most of `jsonrpsee` so the RPC mechanism is usable without explicitly depending on `jsonrpsee`
* Update the executable example to work with subsystems.

This is a draft because it depends on the `subsystems` branch which has not been merged.


rpc/Cargo.toml
```diff
@@ -7,8 +7,14 @@ license = "MIT"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+
```
TheQuantumPhysicist, line 10: line

TheQuantumPhysicist (approved):

muursh (approved):

PR #150, opened 2022-04-27
Roy/bam

royster57:
This PR refactors the PoW module to use a real Block Index.


common/src/chain/block/block_index.rs
```diff
@@ -1,8 +1,25 @@
...
+    InvalidHeightForAncestor,
+    #[error("Block not found")]
+    BlockNotFound,
+}
+
+pub trait BlockIndexDBAccessor {
+    fn get_previous_block_index(
+        &self,
+        block_index: &BlockIndex,
+    ) -> Result<BlockIndex, BlockIndexError>;
```
TheQuantumPhysicist, line 20: Option<BlockIndex>

consensus/src/detail/pow/work.rs
```diff
@@ -84,24 +94,37 @@ impl PoW {
...
         prev_block_index: &BlockIndex,
+        db_accessor: &dyn BlockIndexDBAccessor,
     ) -> Result<Compact, Error> {
         // limit adjustment step
         let actual_timespan_of_last_interval =
             self.actual_timespan(prev_block_index.get_block_time(), retarget_block_time);
 
+        let prev_block_bits = {
+            if let ConsensusData::PoW(pow_data) = db_accessor
+                .get_block(prev_block_index.get_block_id().to_owned())?
```
TheQuantumPhysicist, line 115: Reading the whole block is crazy. This information should be in the block index.

PR #151, opened 2022-04-27
Cryptographic signature support for transactions

TheQuantumPhysicist:


common/src/chain/transaction/signature/inputsig.rs
```diff
@@ -0,0 +1,170 @@
...
+            Destination::Address(addr) => {
+                let sig_components = AuthorizedAddressSpend::from_data(&self.raw_signature)?;
+                verify_address_spending(chain_config, addr, &sig_components, sighash)?
+            }
+            Destination::PublicKey(pubkey) => {
+                let sig_components = AuthorizedPublicKeySpend::from_data(&self.raw_signature)?;
+                verify_public_key_spending(pubkey, &sig_components, sighash)?
+            }
+            Destination::ScriptHash(_) => return Err(TransactionSigError::Unsupported),
+        }
```
iljakuklic, line 92: I think output/input pair verification could be done with something like:
```rust
match (destination, witness) {
    (Destination::Pubkey(pk), Witness::Signature(sig)) => verify_pubkey_spending(pk, sig, sighash),
    (Destination::ScriptHash(sh), Witness::Script(blah)) => verify_script(sh, blah, ....);
    /* (... etc ...) */
    (_dest, _wit) => Err(TransactionError::BadDestinationWitnessPair),
}
```
This is quite nice way of ensuring we have witness type corresponding to the destination type.

Another possibility would be to keep witness as a `Vec<u8>` and decode it on demand once the destination is known into the appropriate type.

iljakuklic (commented):
Adding a bunch of comments. The ones discussing what should be committed to in the signature I consider fairly important.

common/src/chain/transaction/signature/inputsig.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    pub fn get_sighash_type(&self) -> SigHashType {
+        self.sighash_type
+    }
+
+    pub fn from_data(raw_data: &Vec<u8>) -> Result<Self, TransactionSigError> {
+        let decoded_sig = StandardInputSignature::decode(&mut raw_data.as_slice())
+            .map_err(|_| TransactionSigError::DecodingWitnessFailed)?;
+        Ok(decoded_sig)
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I typically do this to make debugging easier... thanks to lldb it doesn't show return values

common/src/chain/transaction/signature/inputsig.rs
```diff
@@ -0,0 +1,170 @@
...
+    signature_hash, TransactionSigError,
+};
+
+#[derive(Debug, Encode, Decode, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub enum InputWitness {
+    #[codec(index = 0)]
+    NoSignature(Option<Vec<u8>>),
+    #[codec(index = 1)]
+    Standard(StandardInputSignature),
+}
```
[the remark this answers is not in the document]
TheQuantumPhysicist: The destination types go inside the StandardInputSignature. The other one is for tests and genesis. Keep in mind that in the future we may want to support different types, including smart contract ones.

common/src/chain/transaction/output.rs
```diff
@@ -4,8 +4,11 @@ use script::Script;
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
 pub enum Destination {
+    #[codec(index = 0)]
     Address(crate::address::Address), // Address type to be added
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yes. I think there's a misunderstanding. An address cannot be converted back to a public key, which is why there's two kinds of transactions. Same in bitcoin, P2PKH and P2PK. Hence, you cannot deduce the public key given an address, because it's generally a hash of the public key.

crypto/src/key/signature/mod.rs
```diff
@@ -56,6 +56,11 @@ impl Decode for Signature {
 }
 
 impl Signature {
+    pub fn from_data(data: &Vec<u8>) -> Result<Self, parity_scale_codec::Error> {
+        let decoded_sig = Signature::decode(&mut data.as_slice())?;
+        Ok(decoded_sig)
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

common/src/chain/transaction/signature/inputsig.rs
```diff
@@ -0,0 +1,170 @@
...
+
+impl StandardInputSignature {
+    pub fn new(sighash_type: sighashtype::SigHashType, raw_signature: Vec<u8>) -> Self {
+        Self {
+            sighash_type,
+            raw_signature,
+        }
+    }
+
+    pub fn get_sighash_type(&self) -> SigHashType {
```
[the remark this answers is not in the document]
iljakuklic: Rust is able to disambiguate these so you are allowed to have a member and a method with the same name.

common/src/chain/transaction/signature/inputsig/authorize_address_spend.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2021 RBB S.r.l
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Not really. They're not the same. Paying to addresses have a different way for verification than public keys.

common/src/chain/transaction/signature/inputsig.rs
TheQuantumPhysicist: Btw, the second option "Another possibility would be to keep witness as a Vec<u8> and decode it on demand once the destination is known into the appropriate type" is what I'm doing here anyway.

common/src/chain/transaction/output.rs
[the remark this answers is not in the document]
iljakuklic: The contention I have is that the `crate::address::Address` type contains a `bech32`-encoded string, which is not something I would expect to be included in blocks.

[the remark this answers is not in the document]
TheQuantumPhysicist: That's fair. But regardless of whether we include the string itself or some other pre-bech32 form, there has to be distinction between addresses and public keys. I'll see what I can do.

common/src/chain/config.rs
```diff
@@ -72,7 +74,7 @@ impl ChainConfig {
     }
 }
 
-const MAINNET_ADDRESS_PREFIX: &str = "mlt";
+const MAINNET_ADDRESS_PREFIX: &str = "mt";
```
muursh, line 77: There's a conversation on slack about this. `mt` has a few issues when it comes to the testnet prefix

common/src/chain/config.rs
```diff
@@ -82,17 +84,19 @@ fn create_mainnet_genesis() -> Block {
...
-    let genesis_mint_receiver = Address::new_with_hrp(MAINNET_ADDRESS_PREFIX, [])
-        .expect("Failed to create genesis mint address");
     let input = TxInput::new(
         Id::<Transaction>::new(&H256::zero()).into(),
         0,
-        genesis_message,
+        InputWitness::NoSignature(Some(genesis_message)),
     );
     let output = TxOutput::new(
         Amount::from_atoms(100000000000000),
```
muursh, line 98: We should probably add a comment about this being temporary and not the final amount

common/src/chain/config.rs
altonen: Generally speaking I can understand the issue with `l` in an address but I wonder is it an actual issue as a prefix for *Mintlayer's* bech32 addresses. I'm looking at `mt` and thinking where did the `l` go. I don't feel strongly about this but I think it's extremely unlikely anybody would typo/read the prefix incorrectly.

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,166 @@
...
+    match mode.inputs_mode() {
+        sighashtype::InputsMode::CommitWhoPays => {
+            for input in tx.get_inputs() {
+                let encoded = input.get_outpoint().encode();
+                stream.write(encoded);
+            }
+        }
+        sighashtype::InputsMode::AnyoneCanPay => {
+            let encoded = target_input.get_outpoint().encode();
+            stream.write(encoded);
```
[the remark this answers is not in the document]
TheQuantumPhysicist: The witness is never included. That's why I'm encoding only the outpoint. Is that an oversight from you or am I doing something wrong?

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,166 @@
...
+        .ok_or_else(|| TransactionSigError::InvalidInputIndex(input_num, tx.get_inputs().len()))?;
+
+    stream.write(tx.get_flags().encode());
+
+    match mode.inputs_mode() {
+        sighashtype::InputsMode::CommitWhoPays => {
+            for input in tx.get_inputs() {
+                let encoded = input.get_outpoint().encode();
+                stream.write(encoded);
+            }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Same comment for witness from below codeeeeeeeeeeee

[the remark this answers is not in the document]
TheQuantumPhysicist: I don't think we need to commit to the number. Committing to the individual hashes is equivalent to committing to the number of inputs too.

common/src/chain/transaction/signature.rs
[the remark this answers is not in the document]
iljakuklic: Ah probably oversight on my part, apologies.

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,166 @@
...
+            let output = tx.get_outputs().get(input_num).ok_or_else(|| {
+                TransactionSigError::InvalidInputIndex(input_num, tx.get_outputs().len())
+            })?;
+            let encoded = output.encode();
+            stream.write(encoded);
+        }
+    }
+
+    stream.write(tx.get_lock_time().encode());
+
```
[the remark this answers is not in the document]
TheQuantumPhysicist: - I added sighash
- I added tx format number
- Amounts are already part of outputs
- OP_CODESEPARATOR is a bad idea
- I hope I didn't miss anything... let's double-check.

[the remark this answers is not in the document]
iljakuklic: > * OP_CODESEPARATOR is a bad idea

Agreed. It is, however, a part of the script language. So we should either somehow include it in sighash or disable it in scripts.

common/src/primitives/encoding/bech32m.rs
```diff
@@ -0,0 +1,52 @@
...
+                return Err(Bech32Error::UnsupportedVariant);
+            }
+
+            // ------- this checking is only for BITCOIN: Witness Programs
+            // if hrp == "bc" && ( s.len() < 2 || s.len() > 40 ) {
+            //     return Err(Bech32Error::InvalidLength);
+            // }
+            // ------- EOL
+
+            let data = base32.iter().map(|x| x.to_u8()).collect();
```
TheQuantumPhysicist, line 25: This is a crime. There's no relation between a u5 array and a u8 array without a proper bit conversion. This is like convert a binary 1100110 to a decimal 1100110.

common/src/primitives/encoding/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+        &self.hrp
+    }
+
+    pub fn get_data(&self) -> &[u8] {
+        &self.data
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DecodedBase32FromBech32(Decoded);
```
TheQuantumPhysicist, line 25: A decoding process with target Base32 should have a `Vec<u5>`, not a `Vec<u8>`.

TheQuantumPhysicist: Keep in mind that we're not required to expose this. This can be private.

common/src/primitives/encoding/tests.rs
```diff
@@ -0,0 +1,264 @@
...
+            ("tb1q0sqzfp3zj42u0perxr6jahhu4y03uw4dypk6sc", Bech32Error::UnsupportedVariant),
+            ("tb1q9jv4qnawnuevqaeadn47gkq05ev78m4qg3zqejykdr9u0cm7yutq6gu5dj", Bech32Error::UnsupportedVariant),
+            ("bc1qz377zwe5awr68dnggengqx9vrjt05k98q3sw2n", Bech32Error::UnsupportedVariant),
+            ("tb1qgk665m2auw09rc7pqyf7aulcuhmatz9xqtr5mxew7zuysacaascqs9v0vn", Bech32Error::FailedChecksum)
+        ).iter().for_each(|(s,b_err)| {
+            match super::decode(*s) {
+                Ok(_) => { panic!("Should be invalid: {:?}", s) }
+                Err(e) => { assert_eq!(*b_err,e) }
+            }
+        });
```
TheQuantumPhysicist, line 263: Please create a tests that:

1. Takes arbitrary u8 data, converts it to bech32, and then converts it back
2. Converts selected results from bech32 to expected values of arbitrary data, back and forth.

common/src/chain/transaction/signature.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: It's now legacy in segwit and is ignored.

common/src/primitives/encoding/mod.rs
sinitcin: Fixed!

common/src/primitives/encoding/bech32m.rs
sinitcin: Gotcha! Fixed!

common/src/chain/transaction/signature.rs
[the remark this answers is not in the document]
iljakuklic: OK, added an issue to deal with it on Script side, most likely just by disabling the instruction. #171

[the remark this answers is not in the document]
TheQuantumPhysicist: I'm still not sure what to do with code separator. Even the reference I used to find out that it's disabled in segwit seems shaky. I'll add a code-separator option which will do nothing in normal transactions. Let's investigate it later in more detail.

common/src/chain/transaction/signature.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: If you're happy with this, please resolve it.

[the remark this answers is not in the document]
TheQuantumPhysicist: Tell you what, I changed my mind. I'll add the size anyway.

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,254 @@
...
+        Ok(tx)
+    }
+
+    fn sign_tx(
+        tx: &mut Transaction,
+        private_key: &PrivateKey,
+        sighash_type: SigHashType,
+        dest: Destination,
+    ) {
+        for i in 0..tx.get_inputs().len() {
```
TheQuantumPhysicist, line 209: Let's stop using `len()` for loops to avoid off-by-one errors, and start always using `.iter().enumerate()`

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,254 @@
...
+        let sighash_type = SigHashType::try_from(SigHashType::ALL).unwrap();
+        let mut tx = generate_tx(private_key.clone(), sighash_type, outpoint_dest.clone()).unwrap();
+        sign_tx(&mut tx, &private_key, sighash_type, outpoint_dest.clone());
+        assert!(verify_signature(&outpoint_dest, &tx, 0).is_ok());
+        assert!(verify_signature(&outpoint_dest, &tx, 1).is_ok());
+        // ALL
+        let sighash_type = SigHashType::try_from(SigHashType::ALL).unwrap();
+        let tx = generate_tx(private_key.clone(), sighash_type, outpoint_dest.clone()).unwrap();
+        assert!(verify_signature(&outpoint_dest, &tx, 0).is_ok());
+        assert!(verify_signature(&outpoint_dest, &tx, 1).is_err());
```
TheQuantumPhysicist, line 236: Add a comment above every test that fails and say why it should fails. In this case, something like "input index out of range"

Also, please don't just use `is_err()`. Please equate to the exact error expected.

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,254 @@
...
+        let sighash_type = SigHashType::try_from(SigHashType::SINGLE).unwrap();
+        let tx = generate_tx(private_key.clone(), sighash_type, outpoint_dest.clone()).unwrap();
+        assert!(verify_signature(&outpoint_dest, &tx, 0).is_ok());
+        assert!(verify_signature(&outpoint_dest, &tx, 1).is_err());
+        // ANYONECANPAY
+        let sighash_type =
+            SigHashType::try_from(SigHashType::ALL | SigHashType::ANYONECANPAY).unwrap();
+        let tx = generate_tx(private_key.clone(), sighash_type, outpoint_dest.clone()).unwrap();
+        assert!(verify_signature(&outpoint_dest, &tx, 0).is_ok());
+        assert!(verify_signature(&outpoint_dest, &tx, 1).is_err());
```
TheQuantumPhysicist, line 252: There should also be tests that mutate:
1. The signature
2. The data behind the signature, one by one

So you try to manipulate the signature by botching a byte (or a bit) in it. Then you ensure that verification will fail. Be careful in such tests as they can become flaky if done incorrectly. For example, don't set a single byte to a specific value you give, like `'x'` always, because then if the byte value is `'x'`, the test will basically change nothing. A good solution is using a `wrapping_add()` on a random byte` in the signature.

Then, you have multiple components that we signed. We want to test that manipulating them isn't really possible. For every type. For example, you change the amounts, number of outputs, etc. See what signature hash is hashing, and try to change everything that a specific signature tries to seal, and modify that.

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,254 @@
...
+                0,
+            )
+            .unwrap();
+            tx.update_witness(i, InputWitness::Standard(input_sign)).unwrap();
+        }
+    }
+
+    #[test]
+    #[allow(clippy::eq_op)]
+    fn sign_and_verify() {
```
TheQuantumPhysicist, line 224: Test names should be meaningful. Please call this something like `sign_and_verify_different_sighash_types()`.

sinitcin: Gotcha! Fixed!

common/src/chain/transaction/signature.rs
sinitcin: If we use `.iter().enumerate()` here, then hidden immutable borrow occurs here for `tx`. Meanwhile, in cycle, we update `tx` that occurs mutable borrow. So, I'm not entirely sure how here possible to use enumerate.

TheQuantumPhysicist: I see. That won't work. Ignore this for now.

common/src/chain/transaction/signature.rs
sinitcin: Look at the code please, is it enough?

common/src/chain/transaction/signature.rs
sinitcin: Fixed

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,512 @@
...
+        sign_tx(
+            &mut original_tx,
+            &private_key,
+            sighash_type,
+            outpoint_dest.clone(),
+        );
+        assert_eq!(verify_sign_tx(&original_tx, &outpoint_dest), Ok(()));
+
+        // Should failed due to changed flags
+        let mut tx_updater = TransactionUpdater::try_from(&original_tx).unwrap();
```
TheQuantumPhysicist, line 428: I'm having trouble understanding the benefit of `TransactionUpdater`... if I understand correctly, it's just a fancy `clone()`... why not just `original_tx.clone()` to a mutable and then modify it?!

common/src/chain/transaction/signature.rs
TheQuantumPhysicist: Things are getting better, but, not everything is covered yet.

For every sighash type, please list in a comment the contents, and then modify them one by one. I see only SighashType::ALL. This certainly isn't enough.

TheQuantumPhysicist:
I want to push this PR to be merged since tests are taking a long time to be produced and the most basic tests are there. The TODO for tests will remain. If anyone would like to review this for production value, please go ahead. I changed removed the draft status off from the PR.

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,512 @@
...
+        }
+    }
+
+    hash_encoded_to(&tx.get_lock_time(), &mut stream);
+
+    // TODO: for P2SH add OP_CODESEPARATOR position
+    hash_encoded_to(&u32::MAX, &mut stream);
+
+    let result = stream.finalize().into();
+    Ok(result)
```
iljakuklic, line 113: I guess incorporating spent output amounts into sighash is going to be a follow-up?

iljakuklic (approved):
Looks good. Just a couple of fairly minor comments.

common/src/chain/transaction/signature.rs
TheQuantumPhysicist: Isn't that part of outputs?

common/src/chain/transaction/signature.rs
```diff
@@ -0,0 +1,512 @@
...
+}
+
+fn verify_standard_input_signature(
```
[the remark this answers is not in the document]
TheQuantumPhysicist: :D

Well, you have to choose your battles, I guess xD

common/src/primitives/encoding/decoded.rs
```diff
@@ -0,0 +1,59 @@
+use super::*;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct Decoded<T> {
+    hrp: String,
+    data: Vec<T>,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I'm OK with all kinds of optimizations. But all in good time.

common/src/primitives/encoding/decoded.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    pub fn get_data(&self) -> &[u8] {
+        self.0.get_data()
+    }
+
+    pub fn encode(&self) -> Result<String, Bech32Error> {
+        bech32m::arbitrary_data_to_bech32m(self.get_hrp(), self.get_data())
+    }
+}
```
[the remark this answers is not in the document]
TheQuantumPhysicist: That was kind of the plan. I told Anton to make the Base32 components private (which is why it was dead-code, and I resolved that comment to explain here), but he did this anyway... and given that I don't really care about this right now, I let it slide. On the other hand, you have no idea how much confusion this Bech32 stuff has caused, starting from Carla, where people consistently thought that Bech32 is decodable to data directly, which is at best not how bitcoin does it, which made the tests break. Having said all that, I agree with everything you said, but honestly I'm fine with it now with long names and everything. Later we can see what we can do about it.

The get_ stuff I'll fix though.

common/src/chain/transaction/signature.rs
sinitcin: So, let me try to explain. We can't mutate `Transaction` or `TransactionV1` due to they have private fields. We have two ways to solve it:
1. We can add functions that will mutate inner state (tx.set_flag \ tx.set_inputs \ etc)
2. We can add a type with public fields. We load inner data into this type, mutable it, and then create a new tx. For me, this way looks better.

common/src/chain/transaction/signature.rs
TheQuantumPhysicist: It just occurs to me. If you mean including the amount of inputs, it turned out it's a bad idea and there's an attack related to that that nullifies its benefit. The most secure way is, unfortuntaely, just to download the input transactions and their merkle branch. So we'll probably just not add the amount of the inputs.

TheQuantumPhysicist: If you're happy with this answer, please resolve this.

common/src/primitives/encoding/decoded.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I fixed the get. If you're happy with the current state, please resolve the comment.

common/src/chain/transaction/signature.rs
TheQuantumPhysicist: I see. Alright. Keep that for now. Just add a comment above `TransactionUpdater` saying "this is required because we can't access private fields of a the `Transaction` class". Also please call it `MutableTransaction`.

common/src/chain/transaction/signature.rs
iljakuklic: OK, resolving. I'd be interested in learning more details about the attack.

iljakuklic (approved):
All good as far as I'm concerned

common/src/chain/transaction/signature.rs
TheQuantumPhysicist: The attack is simply that attempting to use an output amount without verification of the output basically makes it possible for compromised software to trick the user with a hardware wallet into signing multiple outputs under the guise of "something went wrong, let's try again", given that the user can know nothing about the outputs being spent, which can result in burning a huge amount as a transaction fee.

PR #153, opened 2022-04-27
Fix drop checks

iljakuklic:
Signal to the compiler that `Id<T>` and `storage::basic::Store<Sch>` do not drop any `T`s or `Sch`s respectively. This makes the compiler less restrictive when using these types if the generic parameters are types that implement `Drop`.


iljakuklic:
@altonen I have seen some `PhantomData<T>` uses in `p2p` as well. Could you please check if a similar change would be appropriate there too?

TheQuantumPhysicist (approved):
Quite surprising that the compiler doesn't understand that right through `PhantomData` itself...

altonen (approved):
Yes they seem appropriate, I added a commit to my local branch fixing these. Thanks!

muursh (approved):

PR #155, opened 2022-04-28
utils: Introduce the `ensure!` macro

iljakuklic:
Resurrect `ensure!` form Substrate. Used for the common pattern where function exits early with an error if some condition is not satisfied.

<s>The `func_style` submodule is an altenative design that does not use macros. Its advantage is less magic and no unusual behaviour (like interrupting function flow). The disadvantage is that it requires more functions and creating error is not lazy by default, so a bit more clumsy to use in certain cases.</s>

Any preferences between the two? Either the macro or the functions in the `func_style` module will be deleted before finalising this PR. **Going for the macro version**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3252,6 +3252,7 @@ dependencies = [
  "parity-scale-codec",
  "parity-scale-codec-derive",
  "proptest",
+ "utils",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


TheQuantumPhysicist (approved):

muursh (approved):
LGTM. If I had to vote I would vote macro

altonen (approved):
I prefer the macro version

PR #157, opened 2022-05-02
consensus: Make OrphanBlocksPool usable in async code

iljakuklic:
For `consensus` to be usable as a subsystem, it has to be able to oparate in async constexts. In multithreaded runtimes, it means the type has to implement `Send`.

There are many ways to achieve this. This PR goes with the most straightforward approach. Alternative suggestions are welcome.

### DIFF
--- a/consensus/src/detail/orphan_blocks.rs
+++ b/consensus/src/detail/orphan_blocks.rs
@@ -17,19 +17,18 @@
 
 use common::chain::block::Block;
 use common::primitives::{Idable, H256};
-use rand::prelude::ThreadRng;
 use rand::seq::SliceRandom;
 use std::collections::BTreeMap;
-use std::rc::Rc;
+use std::sync::Arc;
 
 pub const DEFAULT_MAX_ORPHAN_BLOCKS: usize = 512;
 
+// FIXME: The Arc here is unnecessary: https://github.com/mintlayer/mintlayer-core/issues/164
 pub struct OrphanBlocksPool {
     orphan_ids: Vec<H256>,
-    orphan_by_id: BTreeMap<H256, Rc<Block>>,
-    orphan_by_prev_id: BTreeMap<H256, Vec<Rc<Block>>>,
+    orphan_by_id: BTreeMap<H256, Arc<Block>>,
+    orphan_by_prev_id: BTreeMap<H256, Vec<Arc<Block>>>,
     max_orphans: usize,
-    rng: ThreadRng,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


consensus/src/detail/orphan_blocks.rs
```diff
@@ -45,7 +45,7 @@ impl OrphanBlocksPool {
             orphan_by_id: BTreeMap::new(),
             orphan_by_prev_id: BTreeMap::new(),
             max_orphans: DEFAULT_MAX_ORPHAN_BLOCKS,
-            rng: rand::thread_rng(),
+            rng: StdRng::from_entropy(),
```
muursh, line 48: Is there a specific reason we're not using the centralised randomness source in the crypto crate for randomness?

muursh, line 48: Is there a specific reason we're not using the centralised randomness source in the crypto crate for randomness?

iljakuklic: I am not sure what the original reason for using `thread_rng` is. Having a centralised rng would require some form of synchronisation, most likely via a mutex. The advantage of having the rng local to the task is the lack of need for synchronisation like that. I am not quite sure what the implications are in terms of security.

iljakuklic: I misunderstood the comment. I thought you were asking why it's not using one big global prng (which may be an interesting question as well). Will have a look into going through our crypto crate instead of using `rand` directly tomorrow.

iljakuklic: At first glance, the `crypto::random` module `make_*_rng` functions return `impl rand::Rng + ...`. The return type is not a concrete named type, which means it can't be used in contexts where the rng type has to be mentioned in code. Here, it has to be specified in the struct definition.

muursh: A single global prng scares me slightly. All I see are weeks and weeks of stats showing we haven't broken anything.

iljakuklic: So I just removed the local RNG from the `OrphanBlocksPool` struct and invoke `crypto::random::make_pseudo_rng` every time. In single-threaded context, this is the same we had before because it just grabs a thread-local RNG state. SInce the RNG no longer crosses any `await` boundaries, it makes `OrphaBlocksPool` fine to be used in async code. If the consensus task happens to be migrated between threads, it may use a different thread-local RNG every time. Is that an issue?

Also, since the `make_pseudo_rng` just returns a reference to an existing thread-local RNG, the name is a bit misleading. Should we rename the function or make it clear this property should not be relied on?

iljakuklic: Resolving as a duplicate

altonen (approved):
Maybe `Arc` is not such a big deal in orphan pool. Can't really comment on the crypto stuff but otherwise LGTM.

muursh (approved):
I'm happy with the `Arc` changes. I'm also fine with the randomness changes for now although I'll sit and think about it a bit this evening.

PR #158, opened 2022-05-02
node: Initital Mintlayer node executable

iljakuklic:
A skeleton node binary. It is able to respond on RPC requests to get the current best block ID.


common/src/primitives/height.rs
```diff
@@ -5,7 +5,19 @@ use std::ops::{Add, Sub};
...
+    Debug,
+    Copy,
+    Clone,
+    PartialOrd,
+    Ord,
+    PartialEq,
+    Eq,
+    Encode,
+    Decode,
+    serde::Serialize,
```
TheQuantumPhysicist, line 18: This is the first file I see in this PR, and my internal reaction was: "Nooooooooooooooooooooooooooooooooooooo!"

https://www.youtube.com/watch?v=yEMimY0HyUk

consensus/src/rpc.rs
```diff
@@ -1 +1,41 @@
...
+        // TODO better way of converting Id<Foo> to Json
+        let res = self.call(move |this| this.get_block_id_from_height(&height)).await;
+        handle_error(res)
+    }
+}
+
+fn handle_error<T>(e: Result<Result<T, ConsensusError>, CallError>) -> rpc::Result<T> {
+    e.map_err(rpc::Error::to_call_error)
+        .and_then(|r| r.map_err(rpc::Error::to_call_error))
+}
```
TheQuantumPhysicist, line 41: No tests?

common/src/primitives/height.rs
TheQuantumPhysicist: On a serious note, this should have a custom implementation, probably to avoid having the tuple affect the json representation. I'm not sure whether serde will map the tuple to json or will do something reasonable to flatten the data in it.

node/src/initialise.rs
```diff
@@ -0,0 +1,30 @@
+//! Node initialisation routine.
+
+use crate::options::Options;
+use consensus::rpc::ConsensusRpcServer;
+
+pub async fn initialise(opts: Options) -> anyhow::Result<subsystem::Manager> {
+    // Initialise storage and chain configuration
+    let storage = blockchain_storage::Store::new_empty()?;
+    let chain_config = common::chain::config::create_mainnet();
```
TheQuantumPhysicist, line 9: This should come from Options; we should be able to create any type of net we need: `--net=mainnet` (which is the default), `--net=testnet`, `--net=regtest`, etc. The `net` thing is up for discussion if anyone doesn't like it.

We should also support flags, like `--testnet` or `--regtest`, just to make it like bitcoin.

iljakuklic: Sure. At the moment, the chain config only exports `create_mainnet` so I did not bother with it for now.

node/src/options.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use std::net::SocketAddr;
+use std::path::PathBuf;
+
+/// Mintlayer node binary
+#[derive(clap::Parser, Debug)]
+#[structopt(name = "mintlayer-node")]
+pub struct Options {
+    /// Where to write logs
+    #[clap(long, name = "PATH")]
```
TheQuantumPhysicist, line 11: Is there a reason why PATH is all caps?

node/src/options.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub log_path: Option<PathBuf>,
+
+    /// Address to bind RPC to
+    #[clap(long, name = "ADDR", default_value = "127.0.0.1:3030")]
```
TheQuantumPhysicist, line 15: Keep in mind that there are two bind addresses, rpc and p2p (hence, 1. I'm opposed to calling it addr as it should have the word bind in it, and 2. it should indicate what this is binding)

TheQuantumPhysicist:
Besides the lack of tests and the few comments I made, the principle in this PR looks good to me.

common/src/primitives/id.rs
```diff
@@ -42,9 +42,16 @@ impl From<Uint256> for H256 {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+impl serde::Serialize for H256 {
```
TheQuantumPhysicist, line 45: We probably also need deserialize here + tests for both

node/src/options.rs
TheQuantumPhysicist: Ah, I see. That's OK. Keep it caps then.

consensus/src/rpc.rs
iljakuklic: WIll have a look if there's a way to test the RPC interface without starting a fully-fledged RPC server. Just passing some json in and checking json we get out.

common/src/primitives/height.rs
iljakuklic: A unit tuple struct is represented as the contained element by `serde` in json. So in json, `BlockHeight` will be just a number. That seems sensible to me, do we want something else?

common/src/primitives/id.rs
iljakuklic: Sure will add

common/src/primitives/height.rs
TheQuantumPhysicist: Perfect. I thought it would try something like `{ "0": 1234 }`.

node/Cargo.toml
```diff
@@ -7,4 +7,15 @@ license = "MIT"
...
+common = { path = "../common/" }
+consensus = { path = "../consensus/" }
 logging = { path = "../logging/" }
+rpc = { path = "../rpc/" }
+subsystem = { path = "../subsystem/" }
+
+# External dependencies
+anyhow = "1.0"
+clap = { version = "3.1", features = [ "derive" ] }
+tokio = "1.17"
```
altonen, line 21: do you need everything from tokio?

altonen (approved):

node/Cargo.toml
iljakuklic: Good point, will check. Probably could be narrowed down somewhat.

common/src/primitives/id.rs
```diff
@@ -181,10 +210,36 @@ mod tests {
...
+        SAMPLE_HASHES.iter().cloned().for_each(check)
+    }
+
+    #[test]
+    fn id_json() {
+        fn check(hex: &'static str) {
+            use serde_test::Token::{Str, Struct, StructEnd};
+            use serde_json::Value;
+            let id: Id<()> = Id::new(&hex.parse().unwrap());
+            // An id serializes into {"id": "<hex_hash>"}
```
TheQuantumPhysicist, line 234: I would suggest not making this elaborate and just making it map to a string of the hash. This can have bad consequences when requesting things by id from rpc. For example, if we request the block index of a block by hash, the input from the user should be `{"id": "<hex_hash>"}`, which is some unnecessary pain. Just make it `"<hex_hash>"`.

consensus/src/rpc.rs
TheQuantumPhysicist: Maybe these tests can be somehow mocked instead of fully instantiating an RPC server?

TheQuantumPhysicist:
Looking much better. Please address the few comments I made before merging.

TheQuantumPhysicist (approved):

common/src/primitives/id.rs
iljakuklic: I actually wanted to ask about this. Will change that.

node/Cargo.toml
iljakuklic: It seems that by default, tokio does not pull in any functionality other than bare minimum. Added `default-features = false` to the manifest anyway just to be sure

iljakuklic:
Added `clap`, `heck`, `strsim`, `textwrap` to `cargo.deny` whitelist. Does that sound acceptable?

muursh (approved):

consensus/src/rpc.rs
iljakuklic: Managed to add some tests. The methods can be invoked without starting a server fairly easily.

iljakuklic:
For RPC tests to work properly, I had to implement panic handling in subsystem manager. If a subsystem now panics, it tears down the whole manager, not just the subsystem task.

altonen (approved):

TheQuantumPhysicist (approved):

PR #160, opened 2022-05-04
Implement consensus-related syncing functions

altonen:
For now I've just exported `block_v1::BlockHeader` and used that, it probably should be done some other way. I'm also not entirely sure how optimal the database accesses are so any tips on how to improve them are appreciated.


consensus/src/detail/mod.rs
```diff
@@ -202,6 +203,34 @@ impl Consensus {
...
         Ok(block)
     }
+
+    pub fn get_header_from_height(
+        &self,
+        height: &BlockHeight,
+    ) -> Result<Option<BlockHeader>, BlockError> {
+        let id = self.get_block_id_from_height(height)?.ok_or(BlockError::NotFound)?;
+        Ok(Some(
+            self.get_block(id)?.ok_or(BlockError::NotFound)?.header().clone(),
```
iljakuklic, line 213: I think we can come up with a way of only returning headers directly from the db. Similarly as we do for transactions, just the offset is always 1 and decode the appropriate bit.

altonen: I'll try that, thanks

TheQuantumPhysicist: Use the block index

altonen: @TheQuantumPhysicist could you elaborate?

I think the blockheaders are variable-sized because they contain `Option<T>` fields so isn't it true we'd have to store the serialized size of them somewhere in order to query them from the storage? For now I just read the entire block and only return header which is a small API improvement.

consensus/src/detail/mod.rs
```diff
@@ -202,6 +203,34 @@ impl Consensus {
...
+        ))
+    }
+
+    pub fn get_locator(&self) -> Result<Vec<BlockHeader>, BlockError> {
+        let id = self.get_best_block_id()?.ok_or(BlockError::NotFound)?;
```
sinitcin, line 218: Best block must be existed, otherwise we see a database corruption. It's legal to use here `expect("...")`

consensus/src/detail/mod.rs
TheQuantumPhysicist: The block index should either contain the header size or the whole header. In bitcoin it contains the whole header. Reading the entire block just for the header is crazy.

consensus/src/detail/test.rs
```diff
@@ -1423,3 +1423,51 @@ fn test_events_orphan_block() {
...
+            block_index = new_block_index;
+            prev_block = new_block;
+        }
+        let locator = consensus.get_locator().unwrap();
+
+        // only genesis
+        if limit == 0 {
+            assert_eq!(locator.len(), 1);
+        } else {
+            assert_eq!(locator.len(), (limit as f64).log2().floor() as usize + 2);
```
sinitcin, line 1459: Out of curiosity, why do we need log2 here?

consensus/src/detail/mod.rs
altonen: Yeah it's idiotic, I'll try storing the header to block index and see how well that works. Thanks

consensus/src/detail/mod.rs
altonen: Thanks!

consensus/src/detail/mod.rs
TheQuantumPhysicist: Btw, I think Roy is doing the same as well. I thought it's in master already, but it's not.

consensus/src/detail/test.rs
altonen: The locator is like a function `fn(n) = get_block_at(2^n)` so it collects the blocks 1,2,4,8,16,32... Taking `floor(log2(our_height))` should give the length of the locator (+ 2) if everything works as it should/I've understood locators correctly.

consensus/src/detail/mod.rs
altonen: Sweet, I'll rebase on top of his branch then. Thanks again

TheQuantumPhysicist: I'm not sure when he's gonna merge. If you're in hurry, it's OK to beat him up to it to master.

altonen: Nope, not in a rush

altonen:
I'll mark this is as ready for review. It depends on Roy's changes and it won't be merged to master until they are in master but I think the code can still be reviewed.

altonen:
I'll close this as the functionality this PR implements is now part of #181

PR #161, opened 2022-05-04
Update for consensus tests

sinitcin:


consensus/src/detail/tests/double_spend_tests.rs
```diff
@@ -0,0 +1,314 @@
+// Copyright (c) 2021 RBB S.r.l
```
muursh, line 1: 2022 :)

consensus/src/detail/tests/double_spend_tests.rs
```diff
@@ -0,0 +1,314 @@
...
+                .expect(ERR_STORAGE_FAIL),
+            consensus.chain_config.genesis_block().get_id()
+        );
+    });
+}
+
+#[test]
+fn double_spend_tx_in_another_block() {
+    common::concurrency::model(|| {
+        // Check is it correctly spend when a couple of transactions in a different blocks pointing on one output
```
muursh, line 227: I'm not sure I understand this comment

consensus/src/detail/tests/processing_tests.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+#[test]
+fn test_straight_chain() {
+    common::concurrency::model(|| {
+        const COUNT_BLOCKS: usize = 255;
+        // In this test, processing a few correct blocks in a single chain
+        let config = create_mainnet();
+        let storage = Store::new_empty().unwrap();
+        let mut consensus = Consensus::new_no_genesis(config, storage).unwrap();
```
altonen, line 185: Oh, it's because we can have an empty consensus? What is the use case for that?

consensus/src/detail/tests/test_framework.rs
```diff
@@ -0,0 +1,302 @@
+use crate::detail::tests::*;
```
altonen, line 1: copyright header missing, other files seem to have it

altonen (approved):

consensus/src/detail/tests/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+
+fn create_new_outputs(config: &ChainConfig, tx: &Transaction) -> Vec<(TxInput, TxOutput)> {
+    tx.get_outputs()
+        .iter()
+        .enumerate()
+        .filter_map(move |(index, output)| create_utxo_data(config, &tx.get_id(), index, output))
+        .collect::<Vec<(TxInput, TxOutput)>>()
+}
+
+fn wait_for_threadpool(consensus: &mut Consensus) {
```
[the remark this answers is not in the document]
iljakuklic: We will add a generic broadcasting mechanism to the subsystem framework eventually. I'd regard this as temporary. Anyone has a different view?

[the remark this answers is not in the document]
TheQuantumPhysicist: We start simple, and then we discuss changes. For me, I wanna guarantee two things with what I did:

1. Consensus can only be called by one thread at a time
2. There will never be a dead-lock
3. (bonus): simplicity

There's no need to go to details now. Let's get the node to work first, then we see what we can do. Please don't go for big changes just for beauty right now.

> eventually

Perfect!

consensus/src/detail/tests/processing_tests.rs
```diff
@@ -0,0 +1,242 @@
...
+    common::concurrency::model(|| {
+        // This test process only Genesis block
+        let config = create_mainnet();
+        let storage = Store::new_empty().unwrap();
+        let mut consensus = Consensus::new_no_genesis(config, storage).unwrap();
+
+        // process the genesis block
+        let block_source = BlockSource::Local;
+        let block_index = consensus
+            .process_block(consensus.chain_config.genesis_block().clone(), block_source)
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Because it has to be processed just like any other block. It has to be analyzed, its outputs should become available to the UTXO set, and it should be written to the database as a valid block, and should have a valid block index. We can either do the same work twice, or make it a valid block for processing.

consensus/src/detail/tests/processing_tests.rs
TheQuantumPhysicist: For testing + we will have different genesis blocks for different chain configs (testnet genesis shouldn't be the same as mainnet genesis).

consensus/src/detail/tests/double_spend_tests.rs
```diff
@@ -0,0 +1,314 @@
...
+        //
+        // Genesis -> b1 -> b2 where
```
TheQuantumPhysicist, line 229: Comment incomplete?

consensus/src/detail/tests/double_spend_tests.rs
```diff
@@ -0,0 +1,314 @@
...
+        assert_eq!(
+            consensus
+                .blockchain_storage
+                .get_best_block_id()
+                .expect(ERR_BEST_BLOCK_NOT_FOUND)
+                .expect(ERR_STORAGE_FAIL),
+            first_block_id
+        );
+    });
+}
```
TheQuantumPhysicist, line 314: For all these tests, if there are forks, please show them in the little ASCII art you're using. Better use something like bitcoin does in feature_blocks.py

consensus/src/detail/tests/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+    // We continue execution when previous threads finished, or cause panic in 3 secs
+    let handle = consensus.events_broadcaster.spawn_handle(|| {});
```
TheQuantumPhysicist, line 219: I don't understand this. A thread pool should have a queue to handle incoming events. Why is this necessary?

TheQuantumPhysicist: Alrighty. If I understand this, you push this to wait for everything to finish in the queue? That's OK. Just please call the function `wait_for_threadpool_to_finish()`

consensus/src/detail/tests/indices_tests.rs
```diff
@@ -0,0 +1,52 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author(s): S. Afach
+
+use crate::detail::tests::*;
+
+#[test]
+fn test_indices_calculations() {
```
TheQuantumPhysicist, line 21: This test should probably be moved to where `calculate_tx_index_from_block()` is

consensus/src/detail/tests/reorgs_tests.rs
```diff
@@ -0,0 +1,411 @@
...
+                .blockchain_storage
+                .get_best_block_id()
+                .expect(ERR_BEST_BLOCK_NOT_FOUND),
+            new_id
+        );
+    });
+}
+
+#[test]
+fn test_very_long_reorgs() {
```
TheQuantumPhysicist, line 100: I don't see a very long reorg test. A very long reorg implies that you create something like 2000 blocks, than you create another chain that's 2001 blocks longer, and then you make sure that it's the new state.

consensus/src/detail/mod.rs
```diff
@@ -643,4 +641,4 @@ impl<'a> ConsensusRef<'a> {
 }
 
 #[cfg(test)]
-mod test;
+mod tests;
```
TheQuantumPhysicist, line 644: To get proper unit-tests visibility to all components, even private ones, it should be `test`, not `tests`:

https://doc.rust-lang.org/reference/visibility-and-privacy.html

iljakuklic: > To get proper unit-tests visibility to all components, even private ones, it should be `test`, not `tests`:
> 
> https://doc.rust-lang.org/reference/visibility-and-privacy.html

I think the name does not matter. Modules always have access to private fields in parent scope, regardless of the name, there is no special case for modules called `test`. The docs just point out that if you add `mod test` as a submodule of code under test, you get access to private fields of stuff in ancestor modules. That is a consequence of the parent-child module relationship, not module name. I am not 100% certain about this but I think I'm right.

TheQuantumPhysicist: You basically elaborated my point.

consensus/src/detail/tests/double_spend_tests.rs
sinitcin: Fixed :)

consensus/src/detail/tests/double_spend_tests.rs
```diff
@@ -0,0 +1,314 @@
...
+            assert!(consensus.process_block(block, BlockSource::Local).is_ok());
+            assert_eq!(
+                consensus
+                    .blockchain_storage
+                    .get_best_block_id()
+                    .expect(ERR_BEST_BLOCK_NOT_FOUND),
+                Some(block_id)
+            );
+        }
+        // The case is invalid. Transsactions should be in order
```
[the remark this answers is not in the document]
sinitcin: Fair enough

consensus/src/detail/tests/double_spend_tests.rs
sinitcin: I would rephrase: A few different blocks use an exactly one output. Check whether it's might spend

sinitcin: Is it works for you? :)

consensus/src/detail/tests/test_framework.rs
sinitcin: Fixed

consensus/src/detail/tests/double_spend_tests.rs
sinitcin: Misprint. Removed

consensus/src/detail/tests/mod.rs
sinitcin: Yep, renamed

consensus/src/detail/mod.rs
sinitcin: @iljakuklic Yes, you are right! Anyway, if `test` looks better, I wouldn't mind renaming it.

consensus/src/detail/tests/double_spend_tests.rs
sinitcin: In this file, no forks so far.

consensus/src/detail/tests/indices_tests.rs
sinitcin: So, then we should make importable functions `generate_random_invalid_block` \ `generate_random_invalid_transaction` \ etc.

consensus/src/detail/tests/reorgs_tests.rs
TheQuantumPhysicist: OK. Based on the call, we'll do this later.

consensus/src/detail/tests/double_spend_tests.rs
TheQuantumPhysicist: Fixed

consensus/src/detail/tests/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I'm resolving this as there doesn't seem to be any request out of it.

consensus/src/detail/tests/reorgs_tests.rs
sinitcin: https://github.com/mintlayer/mintlayer-core/issues/173 I made the issue for it

consensus/src/detail/tests/indices_tests.rs
sinitcin: Fixed!

PR #162, opened 2022-05-05
Add a test to ensure our sk is zeroed correctly

muursh:
This test is somewhat superfluous while we rely on `Tari` for our sks but on the assumption it's viable (likely?) that someone will refactor this to work with an "in house" implementation at some point in time it makes me happier to have tests to catch these annoying little things rather than just the raw functionality of the sigs like we currently have. 
I've started thinking about the key storage so there may be a few more of these over the next few days adding small tests to make life "easier" in the long run.

### DIFF
--- a/crypto/src/key/rschnorr/mod.rs
+++ b/crypto/src/key/rschnorr/mod.rs
@@ -290,4 +290,19 @@ mod test {
         let sig = sk.sign_message(&mut rng, &msg).unwrap();
         assert!(pk.verify_message(&sig, &msg));
     }
+
+    #[test]
+    fn sk_zeroed() {
+        use std::slice;
+        let mut rng = rand::rngs::StdRng::from_entropy();
+        let zero_sk = &vec![0u8; 32][..];
+        unsafe {
+            let hldr;
+            {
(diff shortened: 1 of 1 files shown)


crypto/src/key/rschnorr/mod.rs
```diff
@@ -290,4 +290,19 @@ mod test {
...
+        let mut rng = rand::rngs::StdRng::from_entropy();
+        let zero_sk = &vec![0u8; 32][..];
+        unsafe {
+            let hldr;
+            {
+                let (sk, _pk) = MLRistrettoPrivateKey::new(&mut rng);
+                hldr = sk.as_bytes().as_ptr();
+            }
+            assert_eq!(slice::from_raw_parts(hldr, 32), zero_sk);
+        }
```
iljakuklic, line 306: Which parts are actually unsafe? It's only `from_raw_parts` as far as I can see. Usually, the practice is for the scope of `unsafe` to be minimized.

iljakuklic (approved):

altonen (approved):

crypto/src/key/rschnorr/mod.rs
muursh: Yep it's just `from_raw_parts`

PR #165, opened 2022-05-09
Only start subsystems when `Manager::main` is invoked

iljakuklic:
Previously, subsystems started immediately as soon as they were registered with the `Manager`. With this change, the subsystems are collected by the `Manager` when registered and started when the `main` method on `Manager` is invoked. This is a less surprising behaviour, giving a cleaner separation between initialisation and startup stages. It will also give the opportunity to subsystems to set up all event handlers and communication channels before startup, avoiding certain race conditions.

### DIFF
--- a/crypto/src/key/rschnorr/mod.rs
+++ b/crypto/src/key/rschnorr/mod.rs
@@ -196,9 +196,9 @@ impl std::ops::Add for &MLRistrettoPublicKey {
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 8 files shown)


iljakuklic:
> The `Manager::start` method name is now somewhat misleading. Will change it to `Manager::register` or `Manager::add_subsystem`. Any suggestions?

muursh (approved):
`add_subsytem` is fine by me

PR #166, opened 2022-05-09
serialization: Introduce an abstraction layer over SCALE

iljakuklic:
Make crates go through the `serialization` crate to access the binary encoding and decoding functionality.

This is not an airtight abstraction. SCALE codec is not explicitly mentioned anywhere in the source files of crates, it is accessed through the `serialization` crate. However, the crates still have to pull in `parity_scale_codec` as a dependency. That is because the derive macros still expand that refers to `parity_scale_codec` items.

I think this leaky abstraction is still preferable to using the scale codec directly. Will allow us to extend functionality of the serialization framework while presenting a consistent interface. Support for types that decode based on length and support for serialization modes where some fields are omitted comes to mind.


TheQuantumPhysicist:
We had that discussion back then when we decided to use SCALE. The fact that we had to include SCALE in all library was the primary motivation to use cargo-deny to protect from different versions of SCALE riddling the project. And to be clear, we follow the pattern suggested in this PR in all other crates, such as in crypto and logging.

I'm actually not fan of this change. The reason is that it gives the developer the illusion that they can isolate serialization, which is not true. I want a developer to see SCALE all over the place, and come and ask us why, instead of someone not seeing that and making the assumption it's encapsulated properly while it's leaking.

There's no reason why we still don't put our generic sub-solutions in the serialization crate. But anything that's directly related to SCALE should still be taken from SCALE until this derive issue is resolved.

iljakuklic:
> I'm actually not fan of this change. The reason is that it gives the developer the illusion that they can isolate serialization, which is not true. I want a developer to see SCALE all over the place, and come and ask us why, instead of someone not seeing that and making the assumption it's encapsulated properly while it's leaking.

What is the danger of that? It seems to me that not having scale anywhere in code except for `Cargo.toml` is reasonably strong (although not perfect) isolation. What problem does having SCALE explicitly everywhere solve?

TheQuantumPhysicist:
> What is the danger of that?

It's less of a danger and more of a half-solution, and I don't like half-solutions. The "danger", if we wanna call it that, is a developer not understanding what's going on, spending a day trying to inherit "derive" stuff from serialization, only to realize that it doesn't work for the same reason we discovered months ago.

Things are either isolated or they're not. Especially when the cost is just another line in Cargo.toml. One or two lines... not much of a difference to me. Equal maintenance cost. 

And I'd like to emphasize that I'm more than happy to see some specific functionality come from the serialization crate. We still can do that, while the same assumptions apply: We have scale crates in all of our cargo files... until the derive issue is resolved by rust maintainers.

iljakuklic:
> > What is the danger of that?
> 
> It's less of a danger and more of a half-solution, and I don't like half-solutions. The "danger", if we wanna call it that, is a developer not understanding what's going on, spending a day trying to inherit "derive" stuff from serialization, only to realize that it doesn't work for the same reason we discovered months ago.

Using SCALE directly is also a half-solution. We are picking between two suboptimal half-solutions here sadly.

> And I'd like to emphasize that I'm more than happy to see some specific functionality come from the serialization crate. We still can do that, while the same assumptions apply: We have scale crates in all of our cargo files... until the derive issue is resolved by rust maintainers.

The motivation is to present a consistent interface. When doing stuff with serialization, you go through `serialization::` every time, rather than `parity_scale_codec::` sometimes and `serialization::` sometimes. The differentiation seems like we'd expose an unnecessary amount of implementation detail to other crates. This will become increasingly awkward as we add custom serialization tools. The implementation detail is still exposed in `Cargo.toml`, however that is a fairly limited exposure compared to what we have now.

TheQuantumPhysicist:
I do see your point, and both are suboptimal solutions. And this is why it's difficult to agree on this as you'll always find that my priority is to block developers from doing bad things in these cases. Having said that, if you insist, and other people agree with this change, then please go ahead and merge it. 

@muursh @altonen @royster57 @sinitcin , if you feel you have an opinion here, please throw it in.

muursh (approved):
I'll approve as I have no real qualms with the content. As to whether we should merge or not I'm easy to be honest, I have no strong feelings one way or another. If there are strong views from one side or the other then I'm happy to go with the majority opinion.

TheQuantumPhysicist:
I will approve this. I'm considering giving this a shot. Hopefully we won't be overwhelmed with combinations of imports in every different case.

TheQuantumPhysicist (approved):

PR #168, opened 2022-05-10
Roy/bam c

royster57:


common/src/chain/config.rs
```diff
@@ -117,6 +117,21 @@ pub fn create_mainnet() -> ChainConfig {
     }
 }
 
+pub fn create_mainnet_with_netupgrades(net_upgrades: NetUpgrades<UpgradeVersion>) -> ChainConfig {
```
TheQuantumPhysicist, line 120: This is very dangerous. Mainnet parameters should not be parametrized in any way and should be 100% determined at compile-time. If this is for testing, please create a separate net for unit tests like we did today.

royster57: Removed it.

consensus/src/detail/consensus_validator.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn get_ancestor(
+        &self,
+        block_index: &BlockIndex,
+        ancestor_height: BlockHeight,
+    ) -> Result<BlockIndex, BlockError>;
+}
+
+pub(crate) fn validate_consensus(
+    chain_config: &ChainConfig,
+    block: &Block,
```
TheQuantumPhysicist, line 46: The header should be enough, not the whole block.

consensus/src/detail/consensus_validator.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+fn do_validate(
+    chain_config: &ChainConfig,
+    block: &Block,
+    block_consensus_data: &ConsensusData,
+    consensus_status: &ConsensusStatus,
+    block_index_handle: &dyn BlockIndexHandle,
+) -> Result<(), BlockError> {
+    match (block_consensus_data, consensus_status) {
```
TheQuantumPhysicist, line 80: Nope. Please don't match multiple things together. Please match one, then pass the result to another function that does the mapping to an error or a valid result.

Also, requiring no consensus and providing consensus is invalid.

consensus/src/detail/consensus_validator.rs
```diff
@@ -0,0 +1,90 @@
...
+        block_index: &BlockIndex,
+        ancestor_height: BlockHeight,
+    ) -> Result<BlockIndex, BlockError>;
+}
+
+pub(crate) fn validate_consensus(
+    chain_config: &ChainConfig,
+    block: &Block,
+    block_index_handle: &dyn BlockIndexHandle,
+) -> Result<(), BlockError> {
```
TheQuantumPhysicist, line 48: We also need an accessor for the utxo set. This is not a priority and we can postpone this whole thing. But feel free to do it if you have nothing else to do.

The accessor should have the functions:

- Check if a utxo is spent
- Check if a utxo is spendable

Currently this information can be retrieved from the transaction index.

common/src/chain/upgrades/netupgrade.rs
```diff
@@ -25,17 +35,55 @@ pub trait Activate {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd)]
+pub enum ConsensusUpgrade {
+    PoW { initial_difficulty: Compact },
+    PoS,
+    DSA,
+    IgnoreConsensus,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd)]
+pub enum ConsensusStatus {
```
TheQuantumPhysicist, line 51: I would call this `RequiredConsensus`

consensus/src/detail/error.rs
```diff
@@ -81,7 +90,15 @@ pub enum BlockError {
...
     #[error("Unexpected numeric type conversion error `{0:?}`")]
     InternalNumTypeConversionError(Id<Block>),
-    // To be expanded
+    #[error("Conversion failed: `{0:?}`")]
+    Conversion(String),
+    #[error("No PoW data for block")]
+    NoPowData,
+    #[error("Block consensus data is {block_consensus_data:?}, current config is {chain_consensus_status:?}")]
+    ConsensusTypeMismatch {
+        block_consensus_data: ConsensusData,
```
TheQuantumPhysicist, line 99: Consensus data is relatively large. You can't print it in an error. You can make a string value out of it that mentions the type.

consensus/src/detail/consensus_validator.rs
```diff
@@ -0,0 +1,90 @@
...
+        &consensus_status,
+        block_index_handle,
+    )?;
+    Ok(())
+}
+
+fn do_validate(
+    chain_config: &ChainConfig,
+    block: &Block,
+    block_consensus_data: &ConsensusData,
```
TheQuantumPhysicist, line 76: Isn't it redundant to pass both the block and its consensus data? You can already extract the consensus data from the block itself.

common/src/chain/config.rs
```diff
@@ -117,6 +116,21 @@ pub fn create_mainnet() -> ChainConfig {
     }
 }
 
+pub fn unit_test_config() -> ChainConfig {
```
TheQuantumPhysicist, line 119: `create_unit_test_config()`. Let's follow the same convention.

consensus/src/detail/mod.rs
```diff
@@ -567,9 +608,7 @@ impl<'a> ConsensusRef<'a> {
     }
 
     fn check_consensus(&self, block: &Block) -> Result<(), BlockError> {
```
TheQuantumPhysicist, line 610: This function can be eliminated. No need for one liners in this context.

TheQuantumPhysicist:
Overall this is starting to look like what I envisioned. I like the isolation you created. Please continue and see the comments I made.

consensus/src/detail/consensus_validator.rs
royster57: Done.

consensus/src/detail/consensus_validator.rs
royster57: The IgnoreConsensus variant is so tests don't trip up for not having consensus data... That's why we return Ok when we encounter this variatn.

common/src/chain/upgrades/netupgrade.rs
royster57: Done.

consensus/src/detail/error.rs
royster57: Done.

consensus/src/detail/consensus_validator.rs
TheQuantumPhysicist: I think there's a misunderstanding here. I meant, IF netupgrades says no consensus is required, then adding consensus is a wrong.

common/src/chain/config.rs
royster57: Done.

consensus/src/detail/mod.rs
royster57: Done.

consensus/src/detail/consensus_validator.rs
royster57: Changed this around.

consensus/src/detail/consensus_validator.rs
royster57: Gotcha.

common/src/primitives/compact.rs
```diff
@@ -129,4 +129,19 @@ mod tests {
         err_conversion(0x01fedcba);
         err_conversion(!0x00800000); // overflow
     }
+    /*
```
TheQuantumPhysicist, line 132: ?

consensus/src/detail/consensus_validator.rs
TheQuantumPhysicist: I'm resolving this as it's gonna be done for PoS at some point later.

TheQuantumPhysicist (approved):
Looking good

consensus/src/detail/pow/work.rs
```diff
@@ -17,35 +17,83 @@
...
+    pow_status: &PoWStatus,
+    block_index_handle: &dyn BlockIndexHandle,
+) -> Result<Compact, BlockError> {
+    match pow_status {
+        PoWStatus::Threshold { initial_difficulty } => Ok(*initial_difficulty),
+        PoWStatus::Ongoing => {
+            let prev_block_id = header
+                .get_prev_block_id()
+                .clone()
+                .expect("If PoWStatus is `Onging` then we cannot be at genesis");
```
altonen, line 81: Onging

altonen (approved):

muursh (approved):

PR #169, opened 2022-05-10
Subsystem wrapper testing

altonen:


p2p/src/lib.rs
```diff
@@ -101,3 +104,60 @@ where
...
+            pubsub_mgr.run(call_rq, shut_rq).await
+        });
+
+        let mut sync_mgr = SyncManager::<Libp2pService>::new(sync, rx2, swarm_handle);
+        let sync = manager.add_raw_subsystem("sync", |call_rq, shut_rq| async move {
+            sync_mgr.run(call_rq, shut_rq).await
+        });
+
+        manager.main().await;
+    }
```
altonen, line 162: @iljakuklic 

This is where I hit a bit of roadblock. The managers would need handles passed to each other but I think it's still not possible and I'm not sure if it will ever be because maybe it's too hard to/not worth the effort to decouple the handle creation from spawning the subsystem function.

Channels work since I can create them before I create the managers and I can implement some nice `Handle<T>`-style interface for them with macros.

I'll think about the relationship between them a little bit and maybe some pattern will emerge that doesn't require passing the handle to all of them. Right now the "worst case" (that handles can't be used between P2P objects) is not the end of the world because channels work just fine.

iljakuklic: I'd like to extend the subsystem framework with an event publish-subscribe mechanism at some point. Maybe that could be used? Depends what the interaction between subsystems looks like.

altonen: I had a chat with Sam and it seems that using subsystems wrapper for individual manager objects inside P2P is not correct behavior so you can ignore this PR.

PR #170, opened 2022-05-12
Upgrade scale to latest version: 3.1.2

TheQuantumPhysicist:


muursh (approved):
lgtm

blockchain_storage/src/store.rs
```diff
@@ -197,8 +197,9 @@ impl<Tx: for<'a> traits::GetMapRef<'a, Schema>> BlockchainStorageRead for StoreT
             Ok(Some(block)) => {
                 let begin = tx_index.get_byte_offset_in_block() as usize;
                 let end = begin + tx_index.get_serialized_size() as usize;
-                let tx = block.get(begin..end).expect("Transaction outside of block range");
-                let tx = Transaction::decode_all(tx).expect("Invalid tx encoding in DB");
+                let mut tx = block.get(begin..end).expect("Transaction outside of block range");
+                let decode_ptr = &mut tx;
+                let tx = Transaction::decode_all(decode_ptr).expect("Invalid tx encoding in DB");
```
iljakuklic, line 202: Why the temporary ptr? The `tx` variable is immediately shadowed anyway so this could be just:
```rust
let mut tx = block.get(begin..end).expect("Transaction outside of block range");
let tx = Transaction::decode_all(&mut tx).expect("Invalid tx encoding in DB");
```
(maybe rename the first `tx` to `encoded_tx` or something like that for clarity)

blockchain_storage/src/store.rs
```diff
@@ -265,7 +266,10 @@ impl<'a, Tx: traits::GetMapRef<'a, Schema>> StoreTx<Tx> {
     {
         let col = self.0.get::<DBIdx, I>();
         let data = col.get(key).map_err(crate::Error::from)?;
-        Ok(data.map(|d| T::decode_all(d).expect("Cannot decode a database value")))
+        Ok(data.map(|d| {
+            let mut decode_ptr = d;
+            T::decode_all(&mut decode_ptr).expect("Cannot decode a database value")
+        }))
```
iljakuklic, line 272: Also works:
```rust
Ok(data.map(|mut d| T::decode_all(&mut d).expect("Cannot decode a database value")))
```

iljakuklic (approved):
Looks good. There will, however, be some conflicts with current master I think.

blockchain_storage/src/store.rs
TheQuantumPhysicist: `mut d` is bad. I found another solution though.

blockchain_storage/src/store.rs
TheQuantumPhysicist: `&mut tx` also doens't work. The function needs a mutable pointer, not a mutable object.

iljakuklic (approved):

PR #174, opened 2022-05-17
Include block version in header and use header for block id and some little refactoring

TheQuantumPhysicist:
There are two issues here I try to address:

1. Scale doesn't seem to provide a way to extract the enum index byte value from an enum object
2. The header is the source of a block hash, yet there's no way for it to know the enum index number

My solutions are probably not the best. But I'm hoping I created something consistent that prevents both malicious behavior and mistakes in the future.

Please take a look, and if you can find a better solution, let's talk.

In addition to this, I'll be (somehow) moving the idable implementation to the header. I still am working on that.

### DIFF
--- a/common/src/chain/block/block_v1.rs
+++ b/common/src/chain/block/block_v1.rs
@@ -1,6 +1,8 @@
 use crate::chain::block::Block;
 use crate::chain::block::ConsensusData;
 use crate::chain::transaction::Transaction;
+use crate::primitives::id;
+use crate::primitives::id::Idable;
 use crate::primitives::{Id, H256};
 
 use serialization::{Decode, Encode};
@@ -27,6 +29,24 @@ pub struct BlockV1 {
 
 impl BlockVersion for BlockV1 {
     const BLOCK_VERSION: u32 = 1;
+}
+
+impl Idable<BlockHeader> for BlockHeader {
+    fn get_id(&self) -> Id<Self> {
+        Id::new(&id::hash_encoded(self))
+    }
+}
+
+impl From<&Id<BlockHeader>> for Id<Block> {
+    fn from(id: &Id<BlockHeader>) -> Self {
+        Id::new(&id.get())
+    }
+}
+
+impl From<Id<BlockHeader>> for Id<Block> {
+    fn from(id: Id<BlockHeader>) -> Self {
+        Id::new(&id.get())
+    }
 }
 
 impl BlockV1 {
--- a/common/src/chain/block/block_v1.rs
+++ b/common/src/chain/block/block_v1.rs
@@ -31,7 +31,8 @@ impl BlockVersion for BlockV1 {
     const BLOCK_VERSION: u32 = 1;
 }
 
-impl Idable<Block> for BlockHeader {
+impl Idable for BlockHeader {
+    type Tag = Block;
     fn get_id(&self) -> Id<Block> {
         Id::new(&id::hash_encoded(self))
     }
--- a/common/src/chain/block/block_v1.rs
+++ b/common/src/chain/block/block_v1.rs
@@ -32,7 +32,7 @@ impl BlockVersion for BlockV1 {
 impl BlockV1 {
     pub fn check_version(&self) -> Result<(), super::BlockConsistencyError> {
         let a = self.header.block_version;
-        let b = <Self as BlockVersion>::static_version(&self);
+        let b = <Self as BlockVersion>::static_version(self);
         if a != b {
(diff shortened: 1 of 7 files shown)


common/src/chain/block/block_v1.rs
```diff
@@ -20,7 +27,42 @@ pub struct BlockV1 {
...
+
 impl BlockV1 {
+    pub fn check_version(&self) -> Result<(), super::BlockConsistencyError> {
+        let a = self.header.block_version;
+        let b = <Self as BlockVersion>::static_version(self);
+        if a != b {
+            return Err(super::BlockConsistencyError::VersionMismatch(a, b));
+        }
+        Ok(())
+    }
```
iljakuklic, line 60: Hmm wondering if we could just have the version in one place rather than having header version and block version and requiring them to be in agreement.

common/src/chain/block/block_v1.rs
```diff
@@ -20,7 +27,42 @@ pub struct BlockV1 {
...
 
+impl BlockVersion for BlockV1 {
+    const BLOCK_VERSION: u32 = 1;
+}
+
+impl Idable<BlockHeader> for BlockHeader {
+    fn get_id(&self) -> Id<Self> {
+        Id::new(&id::hash_encoded(self))
+    }
+}
```
iljakuklic, line 38: It is somewhat suboptimal IMHO for `block_header.get_id()` to return `Id<BlockHeader>` rather than `Id<Block>`, since it is essentially the same thing. We can decouple the annotation from the type by changing the `Idable` trait:
```rust
trait Idable: ?Sized {
    type Tag: ?Sized;
    fn get_id(&self) -> Id<Self::Tag>;
}
```

The `Tag` type would be almost always the same as `Self` but it would allow us to return `Id<Block>` form `block_header.get_id()`. It would also allow us to provide implementations for `Idable` on various wrapper types like `Box<T>` with result `Id<T>` rather than `Id<Box<T>>`.

iljakuklic: Also I noticed `Idable` currently has a superfluous type argument
```rust
pub trait Idable<T: ?Sized>
```
should be:
```rust
pub trait Idable
```

consensus/src/detail/error.rs
```diff
@@ -81,6 +84,8 @@ pub enum BlockError {
     SerializationInvariantError(Id<Block>),
     #[error("Unexpected numeric type conversion error `{0:?}`")]
     InternalNumTypeConversionError(Id<Block>),
+    #[error("Internal block representation is invalid `{0:?}`")]
```
iljakuklic, line 87: I think it's preferable to use `{0}` here rather than `{0:?}`. That would give a proper error message instead of strings in Rust syntax.

iljakuklic (commented):
Added a couple of high-level discussion points, trying to make the `Idbale` trait more usable.

common/src/chain/block/block_v1.rs
TheQuantumPhysicist: That's exactly the struggle. Given that we want a separate Header + we want this to be handled with enums, it's not really possible. In other words, we cannot use the version byte from the enum in the hash because the header alone doesn't know about the enum. The only solution there is to get rid of the enum and then process the block contents completely manually.

common/src/chain/block/block_v1.rs
TheQuantumPhysicist: I think a better solution is to keep `T` and use it as the tag. I'll push a commit soon that does that.

iljakuklic: > I think a better solution is to keep `T` and use it as the tag. I'll push a commit soon that does that.

That is also a possibility. Just keep in mind there's a difference between using `T` as tag and having Tag associated type as I presented above. With `Idable<T>` you can have *multiple* `Idable` implementations on each type. E.g. you can have `Idable<Foo> for Block` and `Idable<Bar> for Block`, giving you `Id<Foo>` and `Id<Bar>`, respectively. With associated type, there is one canonical `Idable` for each type implementing it.

TheQuantumPhysicist: Alrighty. I'll go for the tag. It fits better.

consensus/src/detail/error.rs
TheQuantumPhysicist: Done

iljakuklic (approved):
Let's resolve the block + header versioning later. Other than that, I'm happy with this.

PR #177, opened 2022-05-17
ci: Clippy configuration

iljakuklic:
Pick clippy lints to enable. Not all currently pass.


TheQuantumPhysicist:
> Pick clippy lints to enable. Not all currently pass.

In certain situations I can imagine, fixing the code wouldn't be too bad. We were introduced to this and then we minimally changed the code to get it to work.

iljakuklic:
Looks like the newly enabled lints now pass with #179

PR #172, opened 2022-05-16
P2P interface

altonen:
Implements a skeleton interfaces for P2P and P2P RCP and adds a small PubSub implementation

Not sure should I disable SyncManager here and implement it in another PR or rebase this PR on top of #159 and implement SyncManager in this PR. Or rebase #159 on top of this PR and implement syncing using the subsystem manager.

PR #181, opened 2022-05-18
Syncing + consensus integration

altonen:
This PR replaces #159 and #172 by merging them and rewriting the syncing tests using subsystems wrapper. It depends on #160. CI won't pass because one of the commits implements a temporary hack to `blockchain-storage` to get consensus working with P2P


p2p/src/event.rs
```diff
@@ -16,25 +16,23 @@
 // Author(s): A. Altonen
```
TheQuantumPhysicist, line 16: As per the call, please delete this file

p2p/src/event.rs
```diff
@@ -43,14 +41,34 @@ where
...
 
     /// Peer disconnected
-    Disconnected {
-        /// Unique peer ID
-        peer_id: T::PeerId,
+    Disconnected(T::PeerId),
+}
+
+#[derive(Debug, Handle)]
+pub enum P2pEvent {
```
TheQuantumPhysicist, line 51: As per call, please remove this enum

p2p/src/sync/mock_consensus.rs
```diff
@@ -0,0 +1,869 @@
+// Copyright (c) 2022 RBB S.r.l
```
TheQuantumPhysicist, line 1: Please delete this file

consensus/src/detail/mod.rs
```diff
@@ -200,6 +208,119 @@ impl Consensus {
...
+            .take_while(|height| height <= &limit)
+            .map(|height| self.get_header_from_height(&height))
+            .filter(|res| res.is_ok())
+            .filter_map(|res| res.unwrap())
+            .collect::<Vec<_>>();
+
+        Ok(headers)
+    }
+
+    pub fn get_uniq_headers(
```
altonen, line 290: Having looked at this function again, I'd argue it is useful and worth keeping. The reason is because the locator object is inherently inaccurate which means that the response we receive is most likely going to contain headers we already know. For example, if we are on chain `a - b - c - d - e - f - g - h - i - j - k - l - m`, our locator is  `m - l - k - i - e`.

The node we connect to is tracking chain `a - b - c - d - e - f - g -  h - N1 - N2 - N3`, but because the locator they received attached to their chain at `e`, they send us `f - g -  h - N1 - N2 - N3` which attaches to our chain but contains blocks that we already know.

What this function did was verify that the headers we received in response to `GetHeaders` attached to our chain and that they were in order. While verifying the order, it would select all headers we don't know about and report those to syncing.

I'd argue this function is useful and we should have it in the consensus interface.

consensus/src/detail/mod.rs
```diff
@@ -200,6 +208,119 @@ impl Consensus {
...
+            Some(id) => Ok(id == *block_index.get_block_id()),
+            None => Ok(false),
+        }
+    }
+
+    pub fn get_headers(&self, locator: Vec<BlockHeader>) -> Result<Vec<BlockHeader>, BlockError> {
+        // use genesis block if no common ancestor with better block height is found
+        let mut best = BlockHeight::new(0);
+
+        for header in locator.iter() {
```
altonen, line 262: Based on our discussion yesterday it was decided that this is not needed but I'd ask you to take a look again. The `get_ancestor()` function uses block height to find the ancestor block index (meaning the height must be known) whereas this function tries to find the height of the block index so it can collect the correct headers from that point on until it reaches the tip. I don't see a straight-forward way of using the `get_ancestor()` function to advantage here.

consensus/src/detail/mod.rs
TheQuantumPhysicist: The problem is that this function is not written correctly in a way that solves the minimal problem it's meant to be solving (the problem you described in this comment). This is the primary issue. So, I'm gonna make a few suggestions on how to make this function useful and minimal.

1. The function name should be `filter_already_existing_blocks()`.
2. The last loop is mostly pointless and redundant. The chainstate cannot have any blocks that don't have a previous block in the blockchain, which means, you should just break once the block is not found.
3. You should check that blocks are a "chain" in P2P, not in consensus. No reason to delegate this call.

Meaning, this whole function could be changed to the following:
1. Check that we have the previous block at headers[0] (please remove the check for empty, and then use get(0) which will incorporate that
2. Loop over all blocks and check that they exist. Once you hit a block we don't know, break the loop, erase everything from the beginning to that point, and return the new headers.
3. In P2P, before calling this function, make sure that all headers are consecutive. This property can be a `debug_assert!()` in this function.

consensus/src/detail/mod.rs
TheQuantumPhysicist: We should have a function here that gets the common ancestor. I seem to have confused these two. Unfortunately, this function doesn't exist right now. Put a TODO to change this function in the future to use the `last_common_ancestor()` implementation when it exists.

TheQuantumPhysicist: I unresolved this because we have an update. Roy has implemented `last_common_ancestor()`. Please use it. No need for a TODO.

iljakuklic (commented):
A bunch of comments added

p2p/src/sync/mod.rs
```diff
@@ -55,137 +94,465 @@ where
...
+            }
             net::SyncingMessage::Response {
                 peer_id,
-                request_id: _,
+                request_id,
                 response:
                     Message {
                         msg: MessageType::Syncing(SyncingMessage::Response(message)),
-                        ..
+                        magic: _,
```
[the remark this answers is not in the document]
altonen: It's to suppress the unused variable warning

consensus/src/detail/mod.rs
```diff
@@ -206,6 +214,103 @@ impl Consensus {
...
+
+        let headers = itertools::iterate(0, |&i| if i == 0 { 1 } else { i * 2 })
+            .take_while(|i| (height - BlockDistance::new(*i)).is_some())
+            .map(|i| {
+                self.get_header_from_height(
+                    &(height - BlockDistance::new(i)).expect("distance to be valid"),
+                )
+            });
+
+        itertools::process_results(headers, |iter| iter.flatten().collect::<Vec<_>>())
```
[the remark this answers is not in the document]
altonen: Yeah. I tried a couple of different things but it doesn't know how to convert `Flatten` into `Vec<BlockHeader>`

consensus/src/detail/mod.rs
```diff
@@ -206,6 +214,103 @@ impl Consensus {
...
+    // TODO: use `ConsensusRef::is_block_in_main_chain()` implementation instead, how?
+    fn is_block_in_main_chain(&self, block_index: &BlockIndex) -> Result<bool, BlockError> {
+        let consensus_ref = self.make_ro_db_tx();
+        let height = block_index.get_block_height();
+        let id_at_height =
+            consensus_ref.db_tx.get_block_id_by_height(&height).map_err(BlockError::from)?;
+        match id_at_height {
+            Some(id) => Ok(id == *block_index.get_block_id()),
+            None => Ok(false),
+        }
```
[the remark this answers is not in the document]
altonen: I like it, fixed.

consensus/src/detail/mod.rs
```diff
@@ -206,6 +214,103 @@ impl Consensus {
...
+        let limit = std::cmp::min(
+            (best + HEADER_LIMIT).ok_or(BlockError::Unknown)?,
+            self.get_block_height_in_main_chain(
+                &self.get_best_block_id()?.expect("best block to exist"),
+            )?
+            .expect("best block's height to exist"),
+        );
+
+        let headers = itertools::iterate(best.next_height(), |iter| iter.next_height())
+            .take_while(|height| height <= &limit)
```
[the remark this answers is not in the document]
altonen: I'm not entirely sure how to do this. Before I settled on using `process_results()` I was playing around with a normal iteration but that required to implement `std::iter::Step` for `BlockHeight` which is a nightly feature so I'm not certain if what you're asking for here is possible to do with a stable compiler. I'll leave this as unresolved for now.

common/src/chain/config.rs
```diff
@@ -261,7 +261,11 @@ mod tests {
         assert_eq!(&ChainType::Mainnet.to_string(), "mainnet");
         assert_eq!(&ChainType::Testnet.to_string(), "testnet");
 
-        assert_eq!(ChainType::VARIANTS.len(), 4, "Unexpected number of chain types");
+        assert_eq!(
+            ChainType::VARIANTS.len(),
+            4,
+            "Unexpected number of chain types"
+        );
```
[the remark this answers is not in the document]
muursh: I agree that this can be removed.

consensus/src/detail/mod.rs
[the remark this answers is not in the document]
iljakuklic: Feel free to resolve it if you don't want to deal with it, it's a tiny minor issue. I thought it would be possible to implement `Iterator` directly on the range but that might cause issues with overlapping trait impls so not sure.

common/src/chain/config.rs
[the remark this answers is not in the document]
altonen: I think Sam already removed the test. After rebase it's gone

TheQuantumPhysicist (approved):

muursh (approved):

PR #184, opened 2022-05-19
Improvements to node CLI

iljakuklic:
See commit messages for the list of changes


common/src/chain/config.rs
```diff
@@ -199,4 +200,18 @@ mod tests {
...
     }
+
+    #[test]
+    fn chain_type_names() {
+        use strum::VariantNames;
+
+        assert_eq!(&ChainType::Mainnet.to_string(), "mainnet");
+        assert_eq!(&ChainType::Testnet.to_string(), "testnet");
+
+        assert_eq!(ChainType::VARIANTS.len(), 4, "Unexpected number of chain types");
```
TheQuantumPhysicist, line 211: Is there a reason this is necessary?

TheQuantumPhysicist (approved):

common/src/chain/config.rs
iljakuklic: Just to check the external crate generates the variants, probably too paranoid on my part

PR #186, opened 2022-05-23
Signatures tests

sinitcin:
To avoid blocking people, additional tests for brunch `feat/tx_signatures` will be added here


common/src/chain/transaction/signature/inputsig/authorize_pubkey_spend.rs
```diff
@@ -66,4 +66,221 @@ pub fn sign_pubkey_spending(
...
+                sighashtype::SigHashType,
+                signature_hash,
+            },
+            Destination, Transaction, TransactionCreationError, TxInput, TxOutput,
+        },
+        primitives::{Amount, Id},
+    };
+    use crypto::key::{KeyKind, PrivateKey, PublicKey};
+
+    fn generate_unsigned_tx(
```
TheQuantumPhysicist, line 86: When you generate a transaction with just one output, it's not really possible to test the difference between ALL and SINGLE, because they'll be equivalent.

common/src/chain/transaction/signature/inputsig/authorize_pubkey_spend.rs
```diff
@@ -66,4 +66,221 @@ pub fn sign_pubkey_spending(
     Ok(AuthorizedPublicKeySpend::new(signature))
 }
 
-// TODO: tests
+#[cfg(test)]
```
TheQuantumPhysicist, line 69: No mutation tests

common/src/chain/transaction/signature/inputsig/authorize_pubkeyhash_spend.rs
```diff
@@ -77,4 +77,223 @@ pub fn sign_address_spending(
...
+        use std::mem::discriminant;
+        // Destination = Address, but we try to use here AuthorizedPublicKeyHashSpend
+        let sighash_type = SigHashType::try_from(SigHashType::ALL).unwrap();
+        let witness = prepare_data_for_wrong_destination(sighash_type);
+        // Compare Err(TransactionSigError) without inner message
+        assert_eq!(
+            discriminant(&AuthorizedPublicKeyHashSpend::from_data(
+                witness.get_raw_signature()
+            )),
+            discriminant(&Err(TransactionSigError::AddressAuthDecodingFailed(
```
TheQuantumPhysicist, line 162: Please don't use discriminant. In case you're forced to compare enum arms, use `matches!` macro instead. Something like this:

```
        matches!(
            AuthorizedPublicKeyHashSpend::from_data(witness.get_raw_signature()).unwrap_err(),
            TransactionSigError::AddressAuthDecodingFailed(_)
        );
```

common/src/chain/transaction/signature/inputsig/authorize_pubkeyhash_spend.rs
```diff
@@ -77,4 +77,223 @@ pub fn sign_address_spending(
...
+                String::new()
+            )))
+        );
+        let sighash_type =
```
TheQuantumPhysicist, line 166: When you test the same thing multiple times in one test, it's extremely helpful for the reader and for safety to isolate these tests. Usually this is done by having multiple unit tests, but you're not required to do that specifically. An easier solution would be to just wrap every case in its own scope with `{}`. So here, start with `{` at line 153 and close with `}` at line 161.

The reason for this is two fold:

1. Like I said before, easier for the reader to understand the scope of the test.
2. Safer, since variables can't collide and shadowing of variables won't be necessary. Carrying the state of a variable through different tests and accidently using it is a dangerous things. So having a separate scope for tests is a good idea to prevent things like that from happening.

common/src/chain/transaction/signature/inputsig/authorize_pubkeyhash_spend.rs
```diff
@@ -77,4 +77,223 @@ pub fn sign_address_spending(
...
+        let sighash_type = SigHashType::try_from(SigHashType::SINGLE).unwrap();
+        let (private_key, public_key, sighash) = prepare_data_for_sign(sighash_type);
+        let _ = sign_address_spending(&private_key, &public_key, &sighash).unwrap();
+
+        let sighash_type =
+            SigHashType::try_from(SigHashType::SINGLE | SigHashType::ANYONECANPAY).unwrap();
+        let (private_key, public_key, sighash) = prepare_data_for_sign(sighash_type);
+        let _ = sign_address_spending(&private_key, &public_key, &sighash).unwrap();
+    }
+}
```
TheQuantumPhysicist, line 299: Missing mutation tests

p2p/src/net/libp2p/backend.rs
```diff
@@ -18,6 +18,8 @@
 
 // TODO: think about connection management
 
+// TODO: think about connection management
```
TheQuantumPhysicist, line 21: ?

sinitcin: It's appear after rebase. I didn't add it

common/src/chain/transaction/signature/inputsig/authorize_pubkeyhash_spend.rs
sinitcin: I didn't know that matches! can do it. Thanks! Fixed!

common/src/chain/transaction/signature/tests/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+        let signature = match tx_updater.inputs[0].get_witness() {
+            InputWitness::Standard(signature) => {
+                // Let's change around 20ish last bytes, it's also avoided SCALE errors
+                let mut raw_signature = (&signature.get_raw_signature()[0..60]).to_vec();
+                let body_signature: Vec<u8> = signature
+                    .get_raw_signature()
+                    .iter()
+                    .skip(60)
+                    .map(|item| {
+                        if item < &u8::MAX {
```
TheQuantumPhysicist, line 276: Can you please fix this unnecessary if condition?

common/src/chain/transaction/signature/tests/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    let mut tx_updater = MutableTransaction::from(original_tx);
+    for (input_num, _) in original_tx.get_inputs().iter().enumerate() {
+        let signature = match tx_updater.inputs[0].get_witness() {
+            InputWitness::Standard(signature) => {
+                // Let's change around 20ish last bytes, it's also avoided SCALE errors
+                let mut raw_signature = (&signature.get_raw_signature()[0..60]).to_vec();
+                let body_signature: Vec<u8> = signature
+                    .get_raw_signature()
+                    .iter()
+                    .skip(60)
```
TheQuantumPhysicist, line 274: 60?

common/src/chain/transaction/signature/tests/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    let (_, pub_key) = PrivateKey::new(KeyKind::RistrettoSchnorr);
+    tx_updater.outputs.push(TxOutput::new(
+        Amount::from_atoms(1234567890),
+        Destination::PublicKey(pub_key),
+    ));
+    let tx = tx_updater.generate_tx().unwrap();
+    assert_verify_signature(destination, &tx, should_fail);
+}
+
+fn check_change_output(original_tx: &Transaction, destination: &Destination, should_fail: bool) {
```
TheQuantumPhysicist, line 309: mutate

common/src/chain/transaction/signature/tests/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    let mut tx_updater = MutableTransaction::from(original_tx);
+    tx_updater.outputs[0] = TxOutput::new(
+        (tx_updater.outputs[0].get_value() + Amount::from_atoms(100)).unwrap(),
+        tx_updater.outputs[0].get_destination().clone(),
+    );
+    let tx = tx_updater.generate_tx().unwrap();
+    assert_verify_signature(destination, &tx, should_fail);
+}
+
+fn check_change_input(original_tx: &Transaction, destination: &Destination, should_fail: bool) {
```
TheQuantumPhysicist, line 320: mutate

common/src/chain/transaction/signature/tests/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    tx_updater.inputs[0] = TxInput::new(
+        OutPointSourceId::Transaction(Id::<Transaction>::from(H256::random())),
+        9999,
+        tx_updater.inputs[0].get_witness().clone(),
+    );
+    let tx = tx_updater.generate_tx().unwrap();
+    assert_verify_signature(destination, &tx, should_fail);
+}
+
+fn assert_verify_signature(outpoint: &Destination, tx: &Transaction, should_fail: bool) {
```
TheQuantumPhysicist, line 332: This function can be removed. It isn't necessary and just complicates reading the code. Just use an unwrap with an assert.

common/src/chain/transaction/signature/tests/sign_and_mutate.rs
```diff
@@ -0,0 +1,509 @@
...
+        SigHashType::try_from(SigHashType::SINGLE | SigHashType::ANYONECANPAY).unwrap();
+    let tx =
+        generate_and_sigh_tx(&destination, INPUTS, OUTPUTS, &private_key, sighash_type).unwrap();
+
+    let mutations = [add_input, remove_last_input, add_output, remove_last_output];
+    for mutate in mutations.into_iter() {
+        let tx = mutate(&tx);
+        let inputs = tx.inputs().unwrap().len();
+
+        for input in 0..inputs - 1 {
```
TheQuantumPhysicist, line 350: why inputs - 1? Is this an off-by-one error? If not, please explain it. Either in a comment, or use an iterator with `take` so that the reader can see that it's deliberate.

common/src/chain/transaction/signature/tests/sign_and_mutate.rs
```diff
@@ -0,0 +1,509 @@
...
+    let mutations = [modify_input, modify_output];
+    for mutate in mutations.into_iter() {
+        let tx = mutate(&tx);
+        let inputs = tx.inputs().unwrap().len();
+
+        assert_eq!(
+            verify_signature(&destination, &tx, 0),
+            Err(TransactionSigError::SignatureVerificationFailed),
+        );
+        for input in 1..inputs - 1 {
```
TheQuantumPhysicist, line 372: These random ranges are basically playing with fire... we should use something more imperative than something that can be wrong by coincidence.

common/src/chain/transaction/signature/tests/sign_and_verify.rs
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+}
+
+// Same as `sign_and_verify_all_and_none` but for `SINGLE` and `SINGLE | ANYONECANPAY` signature
+// hash types.
+#[test]
+fn sign_and_verify_single() {
+    let (private_key, public_key) = PrivateKey::new(KeyKind::RistrettoSchnorr);
+    let test_data = [
```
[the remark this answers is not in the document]
TheQuantumPhysicist: You know... never mind. I think it's fine... it's not the best, but I think we can live with it.

p2p/src/net/libp2p/backend.rs
```diff
@@ -18,6 +18,8 @@
...
+
```
TheQuantumPhysicist, line 22: Please remove this

TheQuantumPhysicist (commented):
Very good. I have minor comments here and there.

Please fix the issues, rebase on master, and request another review.

common/src/chain/transaction/signature/tests/sign_and_mutate.rs
stanislav-tkach: I agree that is somewhat controversial, but I still think it is better than checking for arbitrary inputs as before.

TheQuantumPhysicist (approved):
I think this is good enough in its current shape. Good job!

PR #187, opened 2022-05-23
Signature verification in consensus

TheQuantumPhysicist:


consensus/src/detail/spend_cache.rs
```diff
@@ -165,13 +176,86 @@ impl<'a> CachedInputs<'a> {
...
     }
 
+    fn verify_signatures(&self, tx: &Transaction) -> Result<(), BlockError> {
+        for (input_idx, input) in tx.get_inputs().iter().enumerate() {
+            let outpoint = input.get_outpoint();
+            let prev_tx_index_op = self.get_from_cached(outpoint)?;
+
+            let tx_index = prev_tx_index_op
+                .get_tx_index()
+                .ok_or(BlockError::PreviouslyCachedInputNotFound)?;
```
iljakuklic, line 186: Is this a verification error or a code invarinat? If the latter, maybe an `expect` would be more appropriate?

consensus/src/detail/spend_cache.rs
```diff
@@ -165,13 +176,86 @@ impl<'a> CachedInputs<'a> {
...
+
+            use common::chain::SpendablePosition;
```
iljakuklic, line 188: There is a clippy lint about putting `use` items in the middle of a function. We can disable it if preferred.

iljakuklic (approved):
Looks good overall. Some minor comments. I like the error cleanup.

common/src/chain/block/block_v1.rs
```diff
@@ -42,6 +43,13 @@ impl BlockHeader {
     pub fn block_time(&self) -> u32 {
         self.time
     }
+
+    pub fn block_reward_destinations(&self) -> Option<&[TxOutput]> {
+        match &self.consensus_data {
+            ConsensusData::None => None,
+            ConsensusData::PoW(pow_data) => Some(pow_data.outputs()),
+        }
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: None is when the ConsensusData inherently doesn't provide a mechanism for rewards. &[] is when the user chose to renounce their block reward.

consensus/src/detail/spend_cache.rs
TheQuantumPhysicist: In CI clippy seems to be happy.

consensus/src/detail/spend_cache.rs
TheQuantumPhysicist: I'm torn in these because I still want the final form to be available before making such a commitment. I'm always afraid we'd mess this up and change some order of things and then things break and crash the node on demand by an attacker.

consensus/src/detail/spend_cache.rs
iljakuklic: I see it here on Github:

```
GitHub Actions
/ clippy
adding items after statements is confusing, since items exist from the start of the scope
warning: adding items after statements is confusing, since items exist from the start of the scope
   --> consensus/src/detail/spend_cache.rs:188:13
    |
188 |             use common::chain::SpendablePosition;
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: requested on the command line with `-W clippy::items-after-statements`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#items_after_statements
```

TheQuantumPhysicist: Alrighty. I'll move it.

common/src/chain/transaction/signature/inputsig.rs
```diff
@@ -88,6 +88,12 @@ impl StandardInputSignature {
                 verify_public_key_spending(pubkey, &sig_components, sighash)?
             }
             Destination::ScriptHash(_) => return Err(TransactionSigError::Unsupported),
+            Destination::AnyoneCanSpend => {
+                // AnyoneCanSpend must use InputWitness::NoSignature, so this is unreachable
```
altonen, line 92: wouldn't `unreachable!()` more appropriate in that case?

common/src/chain/transaction/signature/inputsig.rs
```diff
@@ -110,6 +116,11 @@ impl StandardInputSignature {
                 sig.encode()
             }
             Destination::ScriptHash(_) => return Err(TransactionSigError::Unsupported),
+
+            Destination::AnyoneCanSpend => {
+                // AnyoneCanSpend must use InputWitness::NoSignature, so this is unreachable
```
altonen, line 121: same here

consensus/src/detail/spend_cache.rs
```diff
@@ -165,13 +176,86 @@ impl<'a> CachedInputs<'a> {
...
+    fn apply_spend(
+        &mut self,
+        tx: &Transaction,
+        spend_height: &BlockHeight,
+        blockreward_maturity: &BlockDistance,
+    ) -> Result<(), BlockError> {
+        for input in tx.get_inputs() {
+            let outpoint = input.get_outpoint();
+
+            if let OutPointSourceId::BlockReward(block_id) = outpoint.get_tx_id() {
```
altonen, line 228: `match` instead of `if let`?

altonen (approved):

common/src/chain/transaction/signature/inputsig.rs
TheQuantumPhysicist: Lukas asked the same question about something else. I'm not confident someone won't by mistake call these for some reason. I'm afraid that using expect and unreachable in certain complex places can lead to attack vectors.

iljakuklic: > Lukas asked the same question about something else. I'm not confident someone won't by mistake call these for some reason. I'm afraid that using expect and unreachable in certain complex places can lead to attack vectors.

While I understand the concern, the tradeoff is that it will be more difficult to detect bugs, which is not great either. Seems there is no good option here.

TheQuantumPhysicist: Well, detecting bugs can be done through proper logging after all. It's not ideal, but it's the best we got.

PR #191, opened 2022-05-24
test: Add a functional test runner integrated with Cargo

iljakuklic:
To run functional tests:

`cargo test -p mintlayer-test --test functional`

With extra command line arguments (same as `test_runner.py`):

`cargo test -p mintlayer-test --test functional -- extra args here`

Add a convenient alias to `.cargo/config.toml`:

```toml
[alias]
functional-tests = ['test', '-pmintlayer-test', '--test=functional', '--']
```

And run with:

`cargo functional-tests [extra_args]`

The functional tests run as a part of `cargo test --workspace`.


test/functional/test_framework/test_node.py
```diff
@@ -85,24 +85,32 @@ def __init__(self, i, datadir, *, chain, rpchost, timewait, timeout_factor, bitc
...
         # spam debug.log.
         self.args = [
             self.binary,
-            "-datadir=" + self.datadir,
-            "-logtimemicros",
-            "-debug",
-            "-debugexclude=libevent",
-            "-debugexclude=leveldb",
-            "-uacomment=testnode%d" % i,
+            "--net=mainnet",
```
TheQuantumPhysicist, line 105: While this command line parameter should be OK, the default should be mainnet anyway. This has to be replaced with regtest sometime in the future.

test/functional/test_framework/test_node.py
```diff
@@ -277,7 +286,7 @@ def wait_for_rpc_connection(self):
                     pass  # Port not yet open?
                 else:
                     raise  # unknown OS error
-            except ValueError as e:  # cookie file not found and no rpcuser or rpcpassword; bitcoind is still starting
+            except ValueError:  # cookie file not found and no rpcuser or rpcpassword; bitcoind is still starting
                 if "No RPC credentials" not in str(e):
```
TheQuantumPhysicist, line 290: This will throw an error because `e` is not defined. For some reason you seem to have removed it from the `except` clause.

test/functional/test_framework/test_node.py
```diff
@@ -333,9 +342,9 @@ def stop_node(self, expected_stderr='', *, wait=0, wait_until_stopped=True):
         try:
             # Do not use wait argument when testing older nodes, e.g. in feature_backwards_compatibility.py
             if self.version_is_at_least(180000):
-                self.stop(wait=wait)
+                self.node_exit(wait=wait)
```
TheQuantumPhysicist, line 345: Kind of nitty, but I'd rather call it `node_shutdown` or `node_stop`.

test/functional/test_framework/util.py
```diff
@@ -339,7 +340,9 @@ def rpc_url(datadir, i, chain, rpchost):
             host, port = parts
         else:
             host = rpchost
-    return "http://%s:%s@%s:%d" % (rpc_u, rpc_p, host, int(port))
+    # TODO re-enable authentication
+    #return "http://%s:%s@%s:%d" % (rpc_u, rpc_p, host, int(port))
+    return "http://%s:%d" % (host, int(port))
```
TheQuantumPhysicist, line 345: This is a topic we should have a discussion about. There has to be a username/password for accessing the node, probably. Also an authentication cookie, like bitcoin, would be a very good idea. Is authentication provided by `jsonrpsee`? I wonder.

TheQuantumPhysicist (approved):
Besides the few little things I mentioned, I think this is a good start.

test/functional/test_framework/test_node.py
iljakuklic: Changed

test/functional/test_framework/test_node.py
iljakuklic: I added a todo comment saying this should be changed to regtest but left the option in for completeness. Also, made `mainnet` default so the command line option is not strictly necessary.

test/functional/test_framework/util.py
iljakuklic: Added an issue about this #192

test/functional/test_framework/test_framework.py
```diff
@@ -233,22 +233,20 @@ def setup(self):
...
+        fname_mintlayer_cli = os.path.join(
             config["environment"]["BUILDDIR"],
-            "src",
-            "bitcoin-cli" + config["environment"]["EXEEXT"],
+            "mintlayer-cli" + config["environment"]["EXEEXT"],
         )
-        self.options.bitcoind = os.getenv("BITCOIND", default=fname_bitcoind)
-        self.options.bitcoincli = os.getenv("BITCOINCLI", default=fname_bitcoincli)
+        self.options.bitcoind = os.getenv("MINTLAYER_NODE", default=fname_mintlayer_node)
+        self.options.bitcoincli = os.getenv("MINTLAYER_CLI", default=fname_mintlayer_cli)
```
altonen, line 245: these should probably be `self.options.node` and `self.options.mintlayercli`

test/functional/test_framework/test_node.py
```diff
@@ -226,17 +235,19 @@ def wait_for_rpc_connection(self):
...
-                    rpc_url(self.datadir, self.index, self.chain, self.rpchost),
+                    self.init_rpc_url,
                     self.index,
                     timeout=self.rpc_timeout // 2,  # Shorter timeout to allow for one retry in case of ETIMEDOUT
                     coveragedir=self.coverage_dir,
                 )
-                rpc.getblockcount()
+                rpc.consensus_best_block_id()
                 # If the call to getblockcount() succeeds then the RPC connection is up
                 if self.version_is_at_least(190000):
```
altonen, line 245: I think some of this code could be simplified if you removed these version checks. I don't think `190000` is a meaningful version number for `mintlayer-core` at this time so IMO it's just adding unnecessary noise. I would add it back when we have a version `190000` of the node and it needs to do this thing with mempool.

test/functional/test_framework/test_node.py
```diff
@@ -85,24 +85,33 @@ def __init__(self, i, datadir, *, chain, rpchost, timewait, timeout_factor, bitc
...
+            "--rpc-addr={}".format(rpc_addr),
+            #"-X",
+            #"-datadir=" + self.datadir,
+            #"-logtimemicros",
+            #"-debug",
+            #"-debugexclude=libevent",
+            #"-debugexclude=leveldb",
+            #"-uacomment=testnode%d" % i,
         ]
         if use_valgrind:
```
altonen, line 116: I think this can be removed. I don't think we need to have valgrind code in the testing script, not at least right now.

altonen (approved):
I think the python scripts contain a lot of stuff that we don't need (and may never need) so I'd rather see them go. If we ever need them, we have a great reference but us having to maintain that code right now seems unnecessary

test/functional/test_framework/test_node.py
iljakuklic: So the amount of version checks is not all that massive after all. Removed.

node/src/runner.rs
```diff
@@ -56,9 +56,9 @@ pub async fn run(opts: Options) -> anyhow::Result<()> {
 
 #[rpc::rpc(server, namespace = "node")]
 trait NodeRpc {
-    /// Order the node to exit
-    #[method(name = "exit")]
-    fn exit(&self) -> rpc::Result<()>;
+    /// Order the node to shut down
```
muursh, line 59: `shutdown`

test/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "mintlayer-test"
+version = "0.1.0"
+edition = "2021"
+homepage = "https://github.com/mintlayer/core/issues"
```
muursh, line 5: Should the homepage be issues?

muursh (approved):

test/Cargo.toml
iljakuklic: This is only for the `test` crate. It's presented when tests fail, and is easy to get from Cargo into the system. Could be hard-coded in the code instead though. Would that be preferable?

muursh: Actually that's a fair point, it's probably a good place to throw at people. I'm fine with this

altonen: I think the URL should be updated to point to this repository

iljakuklic: @altonen So the homepage would be this repo, and the url to point people to report bugs would be hardcoded in code or also this repo? I thought it was quite appropriate for homepage for tests to be the issues page but admittedly it's probably not how the homepage key was intended to be used.

TheQuantumPhysicist: I'm not sure whether there's a misunderstanding going on. But I think Aaro means to use `mintlayer-core` instead of `core`.

iljakuklic: Right, thanks for pointing that out!

sinitcin (approved):

PR #193, opened 2022-05-25
Functional test utils

altonen:
This PR fixes the `SwarmControlEvent::Connect` to a working state so connections to other peers can be established with the RPC command `connect`. It also adds a new RPC command called `get_bind_address` which allows the caller to get the full bind address of the local node (including the peer ID).

In functional tests, establishing connection with other node would be something like:
```python
node1 = create_node()
node2 = create_node()

node1_addr = rpc.call(node1, "get_bind_address")
rpc.call(node2, "connect", node1_addr)
```
Probably some RPC command has to be implemented so that we can verify that both peers have a connection. I'll play around with it when both this and #191 are in master

### DIFF
--- a/p2p/src/swarm/mod.rs
+++ b/p2p/src/swarm/mod.rs
@@ -124,7 +124,7 @@ where
                                 response
                                     .send(Err(P2pError::PeerExists))
                                     .map_err(|_| P2pError::ChannelClosed)
-                            },
+                            }
                             None => {
                                 log::warn!("peer count: {:?}", self.peers.len());
                                 log::info!(
@@ -145,6 +145,7 @@ where
                     }
                 }
             }
+            event::SwarmEvent::Disconnect(peer_id) => self.disconnect_peer(peer_id).await,
             event::SwarmEvent::GetPeerCount(response) => {
                 response.send(self.peers.len()).map_err(|_| P2pError::ChannelClosed)
             }
@@ -261,7 +262,7 @@ where
     }
 
     /// Destroy peer information and close all connections to it
-    async fn destroy_peer(&mut self, peer_id: T::PeerId) -> error::Result<()> {
+    async fn disconnect_peer(&mut self, peer_id: T::PeerId) -> error::Result<()> {
         log::debug!("destroying peer {:?}", peer_id);
 
         self.tx_sync
@@ -285,14 +286,14 @@ where
 
                 if self.peers.get(&peer_id).is_some() {
                     log::error!("peer {:?} re-established connection", peer_id);
-                    return self.destroy_peer(peer_id).await;
+                    return self.disconnect_peer(peer_id).await;
                 }
 
                 if self.peers.len() == MAX_ACTIVE_CONNECTIONS {
                     log::warn!("maximum number of connections reached, close new connection with peer {:?}", peer_id);
                     // TODO: save peer information for later?
                     // TODO: i.e., consider this a peer discovery event?
-                    return self.destroy_peer(peer_id).await;
+                    return self.disconnect_peer(peer_id).await;
                 }
 
                 if peer_info.magic_bytes != *self.config.magic_bytes() {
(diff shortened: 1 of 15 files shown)


p2p/src/swarm/mod.rs
```diff
@@ -340,10 +362,10 @@ where
...
                 event = self.rx_swarm.recv().fuse() => {
-                    self.on_swarm_control_event(event).await?;
+                    self.on_swarm_control_event(event).await.map_fatal_err()?;
                 }
                 event = self.handle.poll_next() => match event {
-                    Ok(event) => self.on_network_event(event).await?,
+                    Ok(event) => self.on_network_event(event).await.map_fatal_err()?,
                     Err(e) => {
                         log::error!("failed to read network event: {:?}", e);
                         return Err(e);
```
iljakuklic, line 371: Unrelated to this PR but I noticed that the `run` function returns a `Result<(), Error>` but the `Ok` variant is never returned as far as I can see. It would be more precise to express it in the type with something like `Result<Void, Error>` (which is basically equivalent to just `Error`). Not sure if the `!` type is stable yet, but if it is, that would be even better.

p2p/src/sync/mod.rs
```diff
@@ -47,6 +47,10 @@ const HEADER_LIMIT: usize = 2000;
 // TODO: split syncing into separate files
 // TODO: match against error in `run()` and deal with `ProtocolError`
 // TODO: use ensure
+// TODO: create better api for request/response codec
+// TODO: create helper function for creating requests/responses
+// TODO: cache locator and invalidate it when `NewTip` event is received
+// TODO: simplify code
```
TheQuantumPhysicist, line 53: Maybe more details on "simplify code"... we tend to forget the reason for thinking the code is "complex".

p2p/src/swarm/mod.rs
altonen: Yeah, I'll fix that in a future PR because it also needs to be fixed in all P2P event loops.

p2p/src/sync/mod.rs
altonen: I'll expand the comment but it's impossible for me to forget that it means because you just need to take a look at the current code and see how bad it is to know what to simplify.

TheQuantumPhysicist (approved):

p2p/src/sync/mod.rs
```diff
@@ -72,6 +76,16 @@ impl<T> FatalError for error::Result<T> {
     }
 }
 
+enum RequestType {
+    GetHeaders,
+    GetBlocks(Id<Block>),
```
iljakuklic, line 81: Should it be `GetBlock`? It only takes one block ID.

p2p/src/sync/mod.rs
```diff
@@ -502,10 +574,40 @@ where
...
+                        RequestType::GetHeaders => {
+                            let locator =
+                                self.consensus_handle.call(|this| this.get_locator()).await??;
+                            self.send_header_request(peer_id, locator).await?;
+                        }
+                        RequestType::GetBlocks(block_id) => {
+                            self.send_block_request(peer_id, block_id).await?;
+                        }
+                    }
+                }
```
iljakuklic, line 601: Is there a limit on the number of resend attempts?

iljakuklic (commented):
Added some comments. I am a bit hesitant to approve since I don't understand very well how the p2p code fits together yet but don't have any objections from what I can see.

p2p/src/sync/mod.rs
altonen: No but I think there should be. Maybe 3?

p2p/src/sync/mod.rs
altonen: I'll rename it. It's called `GetBlocks` because we have a message called `GetBlocks` which takes a vector (right now the vector can only contain one element). It's a mess

TheQuantumPhysicist: Actually this should be a `Vec<Id<Block>>`. But it was set to this temporarily, IIUC.

altonen: I'll use a vector

TheQuantumPhysicist: Don't forget to enforce a limit on the size of this vector.

p2p/src/sync/mod.rs
TheQuantumPhysicist: Usually, if communication fails, the connection is just dropped.

altonen: I don't think one failed request automatically means a failed communication. We've also specified that only after three failed PING/PONG retries is the remote peer considered unresponsive. Especially for public nodes during peak traffic there will certainly come a time when it must drop an incoming message because its own queues are full, meaning the request will timeout. But IMO that doesn't mean that after the first dropped message the communication has failed, just that they were not able to service us at that time and we should retry. But I agree with Lukas that there should be a limit for how many times we retry before closing the connection.

PR #195, opened 2022-05-25
Node connectivity for functional tests

altonen:
Implement node connectivity and block syncing for functional tests

### DIFF
--- a/test/functional/example_test.py
+++ b/test/functional/example_test.py
@@ -81,10 +81,7 @@ class ExampleTest(BitcoinTestFramework):
         # block.
         self.setup_clean_chain = True
         self.num_nodes = 3
-        # Use self.extra_args to change command-line arguments for the nodes
         self.extra_args = [[], [], []]
-
-        # self.log.info("I've finished set_test_params")  # Oops! Can't run self.log before run_test()
 
     # Use skip_test_if_missing_module() to skip the test if your test requires certain modules to be present.
     # This test uses generate which requires wallet to be compiled
@@ -114,12 +111,12 @@ class ExampleTest(BitcoinTestFramework):
         them to self.nodes, connect them and then sync."""
 
         self.setup_nodes()
+        self.connect_nodes(0, 1)
 
         # In this test, we're not connecting node2 to node0 or node1. Calls to
         # sync_all() should not include node2, since we're not expecting it to
         # sync.
         # TODO: Re-enable connecting nodes
-        #self.connect_nodes(0, 1)
         #self.sync_all(self.nodes[0:2])
 
     # Use setup_nodes() to customize the node start behaviour (for example if
@@ -140,15 +137,17 @@ class ExampleTest(BitcoinTestFramework):
         """Main test logic"""
 
         # Get current tip hash
-        tip_hash = self.nodes[0].consensus_best_block_id()
-        self.log.info("bestblk: " + tip_hash)
+        tip_hash_0 = self.nodes[0].consensus_best_block_id()
+        tip_hash_1 = self.nodes[1].consensus_best_block_id()
+        self.log.info("bestblk: " + tip_hash_0)
 
         # Get genesis hash (hash of block at height 0)
         genesis_hash = self.nodes[0].consensus_block_id_at_height(0)
         self.log.info("genesis: " + genesis_hash)
 
         # Check they match
-        assert_equal(tip_hash, genesis_hash)
+        assert_equal(tip_hash_0, genesis_hash)
+        assert_equal(tip_hash_0, tip_hash_1)
 
         return
         # TODO: Re-enable the follwing:
--- a/test/functional/example_test.py
+++ b/test/functional/example_test.py
@@ -111,13 +111,12 @@ class ExampleTest(BitcoinTestFramework):
         them to self.nodes, connect them and then sync."""
 
         self.setup_nodes()
-        self.connect_nodes(0, 1)
 
         # In this test, we're not connecting node2 to node0 or node1. Calls to
         # sync_all() should not include node2, since we're not expecting it to
         # sync.
-        # TODO: Re-enable connecting nodes
-        #self.sync_all(self.nodes[0:2])
+        self.connect_nodes(0, 1)
+        self.log.info(self.nodes[0:2])
 
     # Use setup_nodes() to customize the node start behaviour (for example if
     # you don't want to start all nodes at the start of the test).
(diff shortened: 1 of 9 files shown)


test/functional/test_framework/test_framework.py
```diff
@@ -558,19 +558,22 @@ def wait_for_node_exit(self, i, timeout):
...
-        # poll until version handshake complete to avoid race conditions
-        # with transaction relaying
-        # See comments in net_processing:
-        # * Must have a version message before anything else
-        # * Must have a verack message before anything else
-        wait_until_helper(lambda: all(peer['version'] != 0 for peer in from_connection.getpeerinfo()))
-        wait_until_helper(lambda: all(peer['version'] != 0 for peer in to_connection.getpeerinfo()))
-        wait_until_helper(lambda: all(peer['bytesrecv_per_msg'].pop('verack', 0) == 24 for peer in from_connection.getpeerinfo()))
-        wait_until_helper(lambda: all(peer['bytesrecv_per_msg'].pop('verack', 0) == 24 for peer in to_connection.getpeerinfo()))
+        addr_a = self.nodes[a].p2p_get_bind_address()
```
TheQuantumPhysicist, line 561: Actually using the bind address to get the destination address is incorrect. The bind address can be `0.0.0.0` in order to allow connection to all network interfaces in the OS (albeit insecure, but the OS won't complain). What you want to do is specifically connect to `127.0.0.1` (or any loopback of your choosing). What you can learn from the bind address is the bind port, but then there should be a function that returns that separately, I'd say.

In fact, this is exactly why I consistenly insisted on calling this a "bind address" everywhere, because this is not really an address... it's a way of telling the OS what network interface(s) you're interested in getting connected through.

TheQuantumPhysicist: Also I noticed that you're not using `p2p_port`. Keep in mind that `p2p_port` stores the port that was used to initialize the binding. Meaning, you don't even need the `p2p_get_bind_address` to get the correct port. I think Lukas already has used `p2p_port` to initialize the node? I'm not sure.

test/functional/example_test.py
```diff
@@ -80,11 +80,11 @@ def set_test_params(self):
...
         # block.
         self.setup_clean_chain = True
-        self.num_nodes = 3
-        # Use self.extra_args to change command-line arguments for the nodes
-        self.extra_args = [[], [], []]
-
-        # self.log.info("I've finished set_test_params")  # Oops! Can't run self.log before run_test()
+        self.num_nodes = 2
+        self.extra_args = [
+            ["--p2p-addr=/ip6/::1/tcp/3031"],
```
TheQuantumPhysicist, line 85: OK. Now I see the issue with ports on another level. These parameters should be passed in the initial configuration of launching a node. Not in individual tests, through the `p2p_port` variable.

So the work flow should be:

1. You launch the node, giving it the parameter `p2p_port` + the node number
2. Now the node obeys and uses the given port number
3. The `connect_node` function uses `p2p_port` to reproduce the same port number for the same node number
4. In the test, you just say specify the connection topology, like `connect_nodes(0,1)`, and the rest is done automatically through the internal configuration.

test/functional/example_test.py
```diff
@@ -80,11 +80,11 @@ def set_test_params(self):
         # Set setup_clean_chain to True to skip this and start from the Genesis
         # block.
         self.setup_clean_chain = True
-        self.num_nodes = 3
-        # Use self.extra_args to change command-line arguments for the nodes
-        self.extra_args = [[], [], []]
-
-        # self.log.info("I've finished set_test_params")  # Oops! Can't run self.log before run_test()
+        self.num_nodes = 2
```
TheQuantumPhysicist, line 83: Can we make this 3, just like the default?

test/functional/test_framework/test_framework.py
altonen: Localhost is the exception where connecting to `0.0.0.0`/`::1` works because kernel is able to figure out where to send the packets. But I realized that outside localhost that call is not useful, what we're actually interested in is the public-facing address which might be possible to query from libp2p. Anyway, I can change it to connect localhost, I didn't like that because I have to craft the address manually.

test/functional/test_framework/test_framework.py
```diff
@@ -558,19 +559,21 @@ def wait_for_node_exit(self, i, timeout):
...
-        # * Must have a verack message before anything else
-        wait_until_helper(lambda: all(peer['version'] != 0 for peer in from_connection.getpeerinfo()))
-        wait_until_helper(lambda: all(peer['version'] != 0 for peer in to_connection.getpeerinfo()))
-        wait_until_helper(lambda: all(peer['bytesrecv_per_msg'].pop('verack', 0) == 24 for peer in from_connection.getpeerinfo()))
-        wait_until_helper(lambda: all(peer['bytesrecv_per_msg'].pop('verack', 0) == 24 for peer in to_connection.getpeerinfo()))
+        id_a = self.nodes[a].p2p_get_peer_id()
+        id_b = self.nodes[b].p2p_get_peer_id()
+
+        addr_a = p2p_url(a) + "/p2p/" + id_a
+        ret = self.nodes[b].p2p_connect(addr_a)
```
TheQuantumPhysicist, line 566: The connect call is synchronous, right? So it'll be definitely connected when the function returns. Or is it async such that it will attempt to connect?

I can imagine it's synchonous now. But in the future, connecting to peers should be done asynchronously, where requesting to connect basically means "we add the node to the list of nodes we want to connect to". However, this will require having different peer types, which I think isn't implemented right now.

TheQuantumPhysicist: Btw, once things are async, we have to use the `wait_until_helper` from above.

test/functional/example_test.py
```diff
@@ -114,13 +111,13 @@ def setup_network(self):
         them to self.nodes, connect them and then sync."""
 
         self.setup_nodes()
+        self.connect_nodes(0, 1)
```
TheQuantumPhysicist, line 114: Nitty: Is there a reason for moving these up here? Can you please move them back to where they were by default?

TheQuantumPhysicist (approved):
Besides the few comments I made, this looks good.

test/functional/test_framework/test_framework.py
altonen: Yeah its synchronous. Either an error is thrown or the communication succeeds.

PR #196, opened 2022-05-25
Proper test harness for functional tests

iljakuklic:
* Use `libtest-mimic` for functional test runner harness
* Mark functional tests `ignored` so they do not run by default
* Breaking change: Require another `--` to separate harness args from `test_runner.py` args
* Add README
* Add extra commands to run functional tests from CI since `cargo test` no longer runs them by default
* Also fixes some issues when run using `cargo nextest`


iljakuklic:
Please note the command line to the Cargo-based test runner has changed. See the new README for details. If you use the cargo alias, it has to be updated too.

.github/workflows/build.yml
```diff
@@ -40,6 +40,8 @@ jobs:
       run: cargo build --release
     - name: Run tests
       run: cargo test --release --workspace
+    - name: Run functional tests
+      run: cargo test --release -p mintlayer-test --test functional -- --ignored
```
muursh, line 44: Having this in the CI takes the "completion" time to ~20 mins for the CI. Is that a good idea? I genuinely don't know.

muursh: I like the idea of having it in the CI to catch issues

muursh (approved):

.github/workflows/build.yml
TheQuantumPhysicist: There's no question that running functional tests in CI is required. In fact, this is the right place to run them, before merging and breaking the node because of some new code.

TheQuantumPhysicist:
While we're at it, can we change the `tmp` dir name to something more reasonable? Something like `functional_test_config` is much more meaningful.

On the other hand, I just noticed that the tests' temp dir is inside of target. Is there a particular reason why we're not just doing it in `/tmp/` just like bitcoin?

iljakuklic:
> On the other hand, I just noticed that the tests' temp dir is inside of target. Is there a particular reason why we're not just doing it in `/tmp/` just like bitcoin?

When running from Cargo, I set it to the temp dir specifically for test temporary files provided by Cargo. In my experience `/tmp` is often on a small partition (including on my machine) and fills up quickly, resulting in test failures and annoyance.

TheQuantumPhysicist (approved):

altonen:
Is anyone else able to reproduce the CI error locally? I ran the test 1000 times and it worked every time. Furthermore, I can't understand how peer B has established a connection with peer A (meaning they handshaked and all) but peer A still doesn't know about peer B. It could be that by the time the RPC call returns, peer A's peer manager hasn't yet added the peer to its own set of peers. I'll see if I can reproduce it locally.

iljakuklic:
Will try to re-run to see if the CI reproduces the issue reliably.

Backing up the failed test result here: https://github.com/mintlayer/mintlayer-core/runs/6607339201?check_suite_focus=true

TheQuantumPhysicist:
> Will try to re-run to see if the CI reproduces the issue reliably.
> 
> Backing up the failed test result here: https://github.com/mintlayer/mintlayer-core/runs/6607339201?check_suite_focus=true

I've discussed the issue with Aaro. It might be some synchronicity issue due to CI being single-threaded. He's working on a fix in his own PR.

iljakuklic:
The second CI run passed without changes. It is pretty unlikely for the issue to be caused by changes in this PR. Any objections merging it?

TheQuantumPhysicist:
> The second CI run passed without changes. It is pretty unlikely for the issue to be caused by changes in this PR. Any objections merging it?

I think we should just to merge it and consider the other one as improvement. It's a PR anyway.

PR #200, opened 2022-05-26
Add functional test for syncing and block announcements

altonen:


common/src/chain/config.rs
```diff
@@ -101,6 +101,8 @@ impl ChainConfig {
 const MAINNET_ADDRESS_PREFIX: &str = "mtc";
 #[allow(dead_code)]
 const TESTNET_ADDRESS_PREFIX: &str = "tmt";
+#[allow(dead_code)]
+const REGTEST_ADDRESS_PREFIX: &str = "todo";
```
TheQuantumPhysicist, line 105: `rmt`

p2p/src/sync/mod.rs
```diff
@@ -345,11 +345,12 @@ where
...
-            "initialize peer {:?} state, headers: {:#?}",
+        log::debug!(
+            "initialize peer {:?} state, number of headers: {}",
             peer_id,
-            headers
+            headers.len(),
         );
+        log::trace!("headers: {:#?}", headers);
 
         if headers.len() > 2000 {
```
TheQuantumPhysicist, line 355: Constant

test/functional/test_runner.py
```diff
@@ -100,7 +100,8 @@ class UnicodeOnWindowsError(ValueError):
     # vv Tests less than 60s vv
 
     # vv Tests less than 30s vv
-    'example_test.py',
+    # 'example_test.py',
```
TheQuantumPhysicist, line 103: Why comment this one?

test/functional/p2p_syncing_test.py
```diff
@@ -0,0 +1,190 @@
...
+
+        self.log.info("Running custom_method")
+
+    def run_test(self):
+        """Main test logic"""
+
+        # Get current tip hash
+        node0_tip = self.nodes[0].consensus_best_block_id()
+        node1_tip = self.nodes[1].consensus_best_block_id()
+        assert_equal(node0_tip, node1_tip)
```
TheQuantumPhysicist, line 142: Maybe also get the height and ensure it's 0, then progressively check the heights. Better consistency checks.

test/functional/p2p_syncing_test.py
```diff
@@ -0,0 +1,190 @@
...
+            "010401a1b22556e351745689d0a24dce7c13f9ba5b9d09ed56a0d939117a04f7aa6b1d01ab0841549c1fba4345c9b219b0d8de30dbcf602be4aaa87c52ae48e6e532457801e0a83175c089a671d45b657b22c555034f2ceed61480f5684ee2da34b75066d703cb998f6201ffff7f2003000000000000000000000000000000000401000000000400ce56881b009ba213a7ecf315600366f126ccf5f6a8d6168c1c89b17caba2f7530000000000018d01560a194260515b3b33213735411e2838161c3a24105d1547065803445009572531324d4c225f3e0d2f2b5c631f12025e462d2740613029344a5936070e0f0b14111b4f5420451a0501233d132c43494e17551d0c5253083c5a3926483f2a6204184b2e040b2f23aeea58010300000000"
+        ]
+
+        self.nodes[0].consensus_submit_block(blocks[0])
+        self.nodes[0].consensus_submit_block(blocks[1])
```
TheQuantumPhysicist, line 166: Typically after this point one checks that the tip is of hash of `blocks[1]`. But I understand that we don't have a mechanism to calculate the hashes here. So, checking height will suffice for now (with a TODO entry about checking hashes)

TheQuantumPhysicist (approved):
Great!

test/functional/test_runner.py
altonen: There is some problem with node start up if I enable both tests. I'll need to look into it some more.

common/src/chain/config.rs
```diff
@@ -192,6 +194,37 @@ pub fn create_mainnet() -> ChainConfig {
...
+        ),
+        (
+            BlockHeight::new(1),
+            UpgradeVersion::ConsensusUpgrade(ConsensusUpgrade::PoW {
+                initial_difficulty: pow_config.limit().into(),
+            }),
+        ),
+    ];
+
+    ChainConfig {
```
muursh, line 214: Is there a specific reason the data in here is the mainnet but the fn is called regtest? Should `address_prefix` not be `REGTEST_ADDRESS_PREFIX` for example?

altonen: It should be, I don't know how I missed that. But apparently I don't know the alphabets either so there's that.

TheQuantumPhysicist: A typo by Aaro. I'll see if I can fix it. I think it won't break anything.

altonen:
Functional tests were failing when multiple tests were enabled because the test script would call `_initialize_chains()` which passed unrecognized parameters and tried to create a 199 block long chain.

PR #203, opened 2022-05-27
Fix waiting_for_events and do it safely

TheQuantumPhysicist:


consensus/src/detail/mod.rs
```diff
@@ -67,6 +75,17 @@ pub enum BlockSource {
 }
 
 impl Consensus {
+    pub fn wait_for_all_events(&self) {
+        while self.events_in_progress.load(std::sync::atomic::Ordering::Acquire) > 0 {
+            std::thread::yield_now();
+            std::thread::sleep(Duration::from_millis(10));
+        }
```
iljakuklic, line 82: Would it be useful to have a limit on the number of iterations here and hard exit when some sort of upper bound is reached?

iljakuklic: Also does `sleep` not yield implicitly?

iljakuklic (commented):
I am not sure how mixing synchronous code and top-level async used by the subsystem stuff would interact, that should be given more thought. Perhaps @altonen has better insight in that regard. Also `tokio` has a doc for it: https://tokio.rs/tokio/topics/bridging

Also it would be nicer if the `events_in_progress` counter was abstracted in its own wrapper (say, `EventWaitBlocker`, better name needed) type and moved to `util` alongside `CountTracker` and do the waiting inside the destructor of `EventWaitBlocker`. Consensus would wait for events just by having an instance of `EventWaitBlocker` as a member. That way the mechanism could be made reusable.

consensus/src/detail/mod.rs
TheQuantumPhysicist: > Would it be useful to have a limit on the number of iterations here and hard exit when some sort of upper bound is reached?

Probably not. This should never fail if we know what we're doing.

> Also does sleep not yield implicitly?

It may or may not yield depending on the OS. Usually it does, but I don't see the harm.

TheQuantumPhysicist:
> I am not sure how mixing synchronous code and top-level async used by the subsystem stuff would interact, that should be given more thought. Perhaps @altonen has better insight in that regard. Also tokio has a doc for it: https://tokio.rs/tokio/topics/bridging

This is the concern I have with making everything asynchronous with tokio. Everything becomes a soup and it feels unsafe to do things, which is the alarm that triggered in my head when we change Rc to Arc in the orphans object. However, let's go back to the fundamentals of the design: Mutable operations should strictly run in serial, and immutable operations are allowed to be run in parallel. If we follow this recipe, we shouldn't face any issue. This is what I'm basing my design on. If there's more considerations, let's have a discussion about them. If this is a concern, we should isolate Consensus further and only bridge to it with operations that respect these rules. I'm hoping the the subsystem crate takes care of all these problems because it's the gate to any access to consensus. And to be clear, if these rules are broken, I can imagine we'll face bigger problems, like database inconsistency issues, because reads and writes could run in parallel, and with this design, even the database doesn't have to be multi-threaded, even though our initial design was planned to be multithreaded. This was something else coming up soon to make the database single-threaded. 

> Also it would be nicer if the events_in_progress counter was abstracted in its own wrapper (say, EventWaitBlocker, better name needed) type and moved to util alongside CountTracker and do the waiting inside the destructor of EventWaitBlocker. Consensus would wait for events just by having an instance of EventWaitBlocker as a member. That way the mechanism could be made reusable.

I'll see what I can do there.

consensus/src/detail/mod.rs
TheQuantumPhysicist: > Also it would be nicer if the events_in_progress counter was abstracted in its own wrapper (say, EventWaitBlocker, better name needed) type and moved to util alongside CountTracker and do the waiting inside the destructor of EventWaitBlocker. Consensus would wait for events just by having an instance of EventWaitBlocker as a member. That way the mechanism could be made reusable

Done. I create a struct that blocks automatically on drop. It's now in utils.

TheQuantumPhysicist:
> Also it would be nicer if the events_in_progress counter was abstracted in its own wrapper (say, EventWaitBlocker, better name needed) type and moved to util alongside CountTracker and do the waiting inside the destructor of EventWaitBlocker. Consensus would wait for events just by having an instance of EventWaitBlocker as a member. That way the mechanism could be made reusable

Done. I create a struct that blocks automatically on drop. It's now in utils.

utils/src/blockuntilzero.rs
```diff
@@ -0,0 +1,110 @@
+use atomic_traits::{Atomic, NumOps};
+use num_traits::{One, Zero};
+use std::{
+    sync::{atomic::Ordering, Arc},
+    time::Duration,
+};
+
+use crate::counttracker::CountTracker;
+
+pub struct BlockUntilZero<T>
```
altonen, line 10: Why are `BlockUntilZero` and `CountTracker` generic over the atomic data type? They already have a very specific task so generalizing them over the underlying data type doesn't make sense to me. Furthermore `usize`/`isize` is a natural choice for storing a size/count of something so I think it would make much more sense for these objects to use `AtomicUsize`/`AtomicIsize` as the type. Now the generic argument requires us to add two new crates while the value proposition of `BlockUntilZero<AtomicU8>` or `BlockUntilZero<AtomicI64>` is unclear to me.

utils/src/blockuntilzero.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl<T> Drop for BlockUntilZero<T>
+where
+    T: Atomic + NumOps,
+    <T as Atomic>::Type: One,
+    <T as Atomic>::Type: Zero,
+    <T as Atomic>::Type: Ord,
+{
+    fn drop(&mut self) {
+        self.wait_for_zero()
```
altonen, line 70: I don't think this is a good idea. I've already raised my concern regarding us mixing sync and async code elsewhere so we don't need to have that discussion here but having a "hidden" synchronous blocking operation at the end of scope while non-trivial part of our blockchain is async sounds very suspicious to me. If we are going to go this route of both sync and async, I'd much rather see this wait being as explicit as possible.

altonen (commented):
Left some comments. Honestly I'd prefer seeing a generic event implementation in the `subsystem` create which integrates to whole flow better instead of this ad-hoc implementation for consensus.

utils/src/blockuntilzero.rs
TheQuantumPhysicist: Honestly the thing that pissed me off and made me go with generic initially was the lack of `int_fast32_t` alias in rust. And given that I didn't want to think about it and that we want to isolate this to a library for various purposes (which Lukas emphasized even more that I created the BlockUntilZero), I went with a generic to avoid having the discussion "why did you choose 32-bit, and not 16? We'll never have more than 16k events in parallel"... I just don't care about the size thing. I'm sorry; I don't have a good answer for you.

utils/src/blockuntilzero.rs
TheQuantumPhysicist: I'll see if I can wrap it even further.

About combining sync and async... quite frankly I don't know what to do. Since we started with the subsystem thing, I'm not happy that consensus will get into that soup. Like I told Lukas last week, there are certain guarantees and that's all I care about. I've never heard anyone state in C++ "mixing async and sync code" except for worries of deadlock. Quite frankly I'm failing to understand the issue.

Let me see if I can wrap it even further to make it self-contained and serve events. Though I can't promise a solution for the async/sync thing.

altonen: Right now I think the complexity is manageable but IMO the subsystem wrapper should act as the bridge, if at all possible. I think when we start adding more complicated features like maybe the free market for gas or programmable pools the complexity will grow and at that point having one central place for the bridge seems like a good idea to me. Like for example right now in P2P I need to convert the synchronous event into an async event so it can be polled by `tokio::select!`. If this was handled by the subsystem wrapper through some async interface, that would be perfect

TheQuantumPhysicist: I'm still undecided what the best generic solution for events should be. But I think this should do for a good start with a complete solution until the time for that discussion comes.

iljakuklic (approved):
I am fine with this for now. But we should think about how we do it in a generic way that integrates nicely with the rest of the stuff, like the subsystem mechanism. Ideally utilise concurrency primitives Rust already implements. Also I would say it would be preferable to use Tokio for the events distribution system too. While it is possible to use both async and blocking concurrency in one application, it is usually done when there is not good alternative, e.g. using a library that only exposes sync API in an async app.

PR #206, opened 2022-05-30
Add attempts to process blocks

TheQuantumPhysicist:


chainstate/src/detail/mod.rs
```diff
@@ -167,11 +167,33 @@ impl Chainstate {
...
-    pub fn process_block(
+    fn process_db_commit_error(
+        &mut self,
+        db_error: blockchain_storage::Error,
+        block: Block,
+        block_source: BlockSource,
+        attempt_number: usize,
+    ) -> Result<Option<BlockIndex>, BlockError> {
+        // TODO: move to a configuration object that loads from command line arguments
+        const MAX_DB_COMMIT_COUNT: usize = 10;
```
altonen, line 178: Doesn't it now try to retry the operation irrespective of what the actual commit error was? Like if the database was full on previous try and we retry immediately again (the function has exclusive access throughout this process), I don't see how this operation is going to perform any differently unless we start implementing some healing mechanisms. Shouldn't it only retry if the error is deemed recoverable? If the error is always recoverable then this is not an issue.

TheQuantumPhysicist: Retrying is restricted to recoverable errors. Unrecoverable error will just panic at a lower level. The error that's targeted here is a multi-threaded transaction serialization failure. As an example to explain it, if you start transaction T1, and then transaction T2, then T1 modifies value A, then T2 modifies the same value A, then T2 commits A, and finally T1 attempts to commit A. In this case, the value A was not isolated enough for A to be updated. The correct solution in this case is to clear the transaction and start over with a new assumption of the state of A.

muursh (approved):
I have no issues with anything here

PR #207, opened 2022-05-30
Add block reward destination

sinitcin:


chainstate/src/detail/tests/processing_tests.rs
```diff
@@ -531,11 +533,15 @@ fn test_consensus_type() {
...
-        assert!(btf.add_special_block(mined_block).is_ok());
+        let (_, pub_key) = PrivateKey::new(KeyKind::RistrettoSchnorr);
+        assert!(crate::detail::pow::work::mine(
+            &mut mined_block,
+            u128::MAX,
+            bits,
+            vec![TxOutput::new(Amount::from_atoms(10), Destination::PublicKey(pub_key))]
+        )
+        .expect("Unexpected conversion error"));
+        assert!(btf.add_special_block(mined_block).unwrap().is_some());
```
TheQuantumPhysicist, line 544: Please just unwrap. Do not test for is_some or is_ok.

chainstate/src/detail/tests/processing_tests.rs
```diff
@@ -28,6 +28,8 @@ use common::chain::OutputSpentState;
 use common::chain::UpgradeVersion;
 use common::primitives::Compact;
 use common::Uint256;
+use crypto::key::KeyKind;
+use crypto::key::PrivateKey;
```
TheQuantumPhysicist, line 32: Not a big deal, but both can be in one `use`

chainstate/src/detail/spend_cache.rs
```diff
@@ -259,27 +306,56 @@ impl<'a> CachedInputs<'a> {
...
+                            self.check_inputs_amounts(inputs, outputs)?;
+                        }
+                    }
+                    None => {}
+                }
 
-        // spend inputs of this transaction
-        self.apply_spend(tx, spend_height, blockreward_maturity)?;
+                // verify input signatures
+                self.verify_all_signatures(block)?;
```
TheQuantumPhysicist, line 348: I'm having trouble understanding this. So this block reward attempts to spend some output. And to check whether the signature is correct, hence, the spend is authorized, we check all the signatures in the block except for the one we're checking?! This is definitely incorrect.

chainstate/src/detail/spend_cache.rs
```diff
@@ -201,28 +235,40 @@ impl<'a> CachedInputs<'a> {
...
-                SpendablePosition::BlockReward(_reward_pos) => {
+                SpendablePosition::BlockReward(reward_pos) => {
                     // TODO(Roy): fill this with the block reward that the user now is spending
-                    todo!()
+
+                    let block = self
+                        .db_tx
+                        .get_block(reward_pos.clone())
+                        .map_err(BlockError::from)?
+                        .ok_or(BlockError::NotFound)?;
```
TheQuantumPhysicist, line 248: Please. Do not use generic error codes. This is extremely dangerous. Errors hould have a context, not just "NotFound" or "Orphan" or something similar. When someone submits a block for processing, returning an error must mean something. What does it mean to P2P when you tell them that the error for the block submitted by a peer is "NotFound"? It means nothing. I got rid of all the Unknown errors which meant nothing, and now we're having another dimension of meaningless errors appearing. This is basically not a NotFound. This is basically an InvalidBlock of some sorts, because the creator of the block referenced a block reward that could not be found in the history. Please give error codes more thought because they're very essential in block processing, or otherwise we'll be open to all kinds of attacks.

chainstate/src/detail/spend_cache.rs
```diff
@@ -201,28 +235,40 @@ impl<'a> CachedInputs<'a> {
...
+                        .map_err(|_| BlockError::SignatureVerificationFailed(tx.get_id()))?;
                 }
-                SpendablePosition::BlockReward(_reward_pos) => {
+                SpendablePosition::BlockReward(reward_pos) => {
                     // TODO(Roy): fill this with the block reward that the user now is spending
-                    todo!()
+
+                    let block = self
+                        .db_tx
+                        .get_block(reward_pos.clone())
```
TheQuantumPhysicist, line 246: To get a block reward in history, you do not need a whole block. You only need the header. And the header can be found in the block index.

chainstate/src/detail/spend_cache.rs
```diff
@@ -259,27 +306,56 @@ impl<'a> CachedInputs<'a> {
...
+                self.verify_signatures(tx)?;
 
-        // verify input signatures
-        self.verify_signatures(tx)?;
+                // spend inputs of this transaction
+                let spender = Spender::from(tx.get_id());
+                self.apply_spend(tx.get_inputs(), spend_height, blockreward_maturity, spender)?;
+            }
+            SpendKind::BlockReward => {
+                match block.header().block_production_inputs() {
```
TheQuantumPhysicist, line 332: Please add: TODO: test spending block rewards from chains outside the mainchain

chainstate/src/detail/tests/processing_tests.rs
sinitcin: Ok, fixed!

chainstate/src/detail/tests/processing_tests.rs
sinitcin: Ok, I wouldn't mind combining them in one `use`.

chainstate/src/detail/spend_cache.rs
sinitcin: Thanks!

chainstate/src/detail/spend_cache.rs
sinitcin: Done!

common/src/chain/config.rs
```diff
@@ -47,6 +55,9 @@ pub struct ChainConfig {
     genesis_block_id: Id<Block>,
     blockreward_maturity: BlockDistance,
     version: SemVer,
+    target_block_spacing: Duration,
+    coin_decimals: u8,
+    emission_schedule: Vec<(BlockHeight, Amount)>,
```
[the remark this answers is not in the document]
iljakuklic: Hmm we need total emission as well, maybe it could be a function that gives the total supply at given block height and the reward would be calculated as total supply at current block minus total supply at previous block.

common/src/chain/config.rs
```diff
@@ -169,6 +200,36 @@ fn create_unit_test_genesis(premine_destination: Destination) -> Block {
         .expect("Error creating genesis block")
 }
 
+pub fn make_mainnet_emission_schedule(
+    target_block_spacing: &Duration,
+    coin_decimals: u8,
+) -> Vec<(BlockHeight, Amount)> {
+    let year_in_blocks = (365 * 24 * 60 * 60) / target_block_spacing.as_secs();
+    let rewards_per_block_in_year_n_str =
+        ["202", "151", "113", "85", "64", "48", "36", "27", "20", "15", "0"];
```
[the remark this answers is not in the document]
TheQuantumPhysicist: That depends. It's a design question more than being a technical impossibility. My inherent fear in all these integer vs string is someone interpreting an atom as a coin and vice versa. To me, the convention is, at the lowest level, a string is a coin and an integer is an atom. We can try to run around this, but no matter what we do, the amount has to be entered some how as a number, and we spin back to the first question, how can we do this in an unambiguous way.

Open question. My opinion is, we have bigger fish to fry.

If you're satisfied, please resolve the comment.

common/src/chain/config.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Unfortunately, while the function idea is programmatically a good idea and I do like it, it doesn't fulfill the requirement for a generic mapping that we can convert from a string. In this paradigm, the requirement is to be able to parse the schedule from a string. This helps in creating a functional test that tests the schedule without having to mine years worth of blocks.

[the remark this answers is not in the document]
TheQuantumPhysicist: About the function with the total supply; we can't do that here. The correct way to do this (which will come up later) is whenever we accept a block in the chain, there will be a total supply entry in the block index that tells us the total supply at that block. This is, unfortunately, not achievable abstractly here because miners/stakers are free to renounce their rewards. Mining a block with zero reward is totally acceptable.

TheQuantumPhysicist:
> Have you considered separating out the sighash serialization from the actual hashing?

It occurred to me. But the performance hit is not really worth it. No need to do extra allocations. Unless I misunderstand what you're envisioning.

common/src/chain/config.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: To avoid off-by-one errors, I added lots of tooling and used pairs instead. Check it out.

TheQuantumPhysicist (approved):

PR #208, opened 2022-05-31
p2p: Stop using #[allow(unused)]

altonen:


p2p/src/net/libp2p/proto/connectivity.rs
```diff
@@ -286,7 +285,7 @@ mod tests {
 
     #[tokio::test]
     async fn outgoing_error_inbound_exists() {
-        let (mut backend, _, _, _, _) = util::make_libp2p(
+        let (_, _, _, _, _) = util::make_libp2p(
```
TheQuantumPhysicist, line 288: This look is pretty much why returning big tuples is an anti-pattern. Please consider changing this to a struct some time in the future.

p2p/src/net/libp2p/proto/mdns.rs
```diff
@@ -209,7 +201,7 @@ mod tests {
                     Some(_) => {}
                     None => panic!("got None"),
                 },
-                event = backend2.swarm.next() => {}
+                _ = backend2.swarm.next() => {}
```
TheQuantumPhysicist, line 204: It's preferable to just prefix variables with _ instead of just using _ as variable name. This helps in the future if someone needs to extend something and use the result. This is not particular to this line, but I think lots of _ where meaningful information could've been conveyed.

p2p/src/net/libp2p/proto/sync.rs
```diff
@@ -137,8 +140,6 @@ impl Backend {
 
 #[cfg(test)]
 mod tests {
```
TheQuantumPhysicist, line 142: Maybe either remove this or add a TODO and explain what tests have to be added later

p2p/src/net/libp2p/proto/util.rs
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 //
 // Author(s): A. Altonen
+#![allow(unused)]
```
TheQuantumPhysicist, line 17: I thought this PR is about removing these :smile:

p2p/src/net/libp2p/backend.rs
```diff
@@ -264,12 +252,18 @@ impl Backend {
...
                     result,
-                );
-                response.send(Ok(())).map_err(|_| P2pError::ChannelClosed)
+                ) {
+                    Ok(_) => response.send(Ok(())).map_err(|_| P2pError::ChannelClosed),
+                    Err(e) => response
+                        .send(Err(P2pError::Libp2pError(Libp2pError::PublishError(
+                            e.to_string(),
+                        ))))
+                        .map_err(|_| P2pError::ChannelClosed),
```
TheQuantumPhysicist, line 265: As I'm seeing "ChannelClosed" all over the place, I'm wondering whether this error can be more expressive by containing the reason of that so that it can go to logs with that information.

p2p/src/sync/mod.rs
```diff
@@ -516,15 +513,15 @@ where
...
+                        _peer_id: peer_id,
                         request_type: RequestType::GetHeaders,
                         retry_count: 0,
                     },
                 );
                 peer.set_state(peer::PeerSyncState::UploadingHeaders);
                 Ok(())
             }
-            Err(e) => {
+            Err(_) => {
```
TheQuantumPhysicist, line 524: Since you're logging anyway, why not print that error? Our errors in general implement the Display trait or at least the Debug trait.

TheQuantumPhysicist (approved):
Looks OK. Please look into the few comments I made.

p2p/src/sync/mod.rs
altonen: Fixed

iljakuklic (approved):

p2p/src/net/libp2p/backend.rs
altonen: There isn't a multitude of reasons for why the channel was closed though. It's that the the sender/receiver end was dropped, meaning of the manager shut down for whatever reason.

p2p/src/net/libp2p/proto/mdns.rs
iljakuklic: I agree with @TheQuantumPhysicist in general. However, be aware that there is a subtle difference. Single `_` is a wildcard pattern [1] while `_foo` is an identifier pattern [2]. They differ in when is the value bound to it dropped. Adding a descriptive name can unnecessarily extend the lifetime of the value if the type implements `Drop`.

[1] https://doc.rust-lang.org/stable/reference/patterns.html#wildcard-pattern
[2] https://doc.rust-lang.org/stable/reference/patterns.html#identifier-patterns

muursh (approved):

PR #209, opened 2022-05-31
test: Add helper function for generating blocks for p2p_syncing_test.py

altonen:


chainstate/src/detail/tests/mod.rs
```diff
@@ -162,3 +164,27 @@ fn create_new_outputs(tx: &Transaction) -> Vec<(TxInput, TxOutput)> {
         .filter_map(move |(index, output)| create_utxo_data(&tx.get_id(), index, output))
         .collect::<Vec<(TxInput, TxOutput)>>()
 }
+
+// generate 5 regtest blocks and print them in hex
+#[ignore]
+#[test]
+fn generate_blocks_for_functional_tests() {
```
TheQuantumPhysicist, line 171: Please add a comment that this should be removed when block production is complete and is usable through RPC

TheQuantumPhysicist (approved):

chainstate/src/detail/tests/mod.rs
TheQuantumPhysicist: Another TODO :-)

altonen: For each item I remove the TODO list, I add 1-3 more. I need to start working *really* fast :smiley:

iljakuklic (approved):
Could this be an example rather than an ignored test? Seems a better fit for this use case to me but an example would not have the access to private methods.

altonen:
I like the idea of making it an example but it uses private modules so I don't know if we want to expose them just for this example.

muursh (approved):

PR #210, opened 2022-05-31
Refactor P2P error code

altonen:


p2p/src/sync/peer.rs
```diff
@@ -92,12 +95,14 @@ where
...
                         header
                     );
-                    return Err(P2pError::InvalidData);
+                    // TODO: this is a protocol error
+                    return Err(P2pError::ProtocolError(ProtocolError::InvalidMessage));
                 }
             }
             _ => {
+                // TODO: this is a protocol error
                 log::error!("received a header from peer while not expecting it");
```
TheQuantumPhysicist, line 104: Please try to add useful information to error messages. Mentioning the peer information, like address, is useful. If this message appears in the logs, it'd be completely useless.

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
-    #[error("PeerExists")]
-    PeerExists,
+    #[error("Peer-related error: `{0:?}`")]
+    PeerError(PeerError),
+    #[error("Invalid data: `{0}`")]
+    InvalidData(&'static str),
     #[error("SubsystemFailure")]
     SubsystemFailure,
     #[error("ConsensusError: `{0:?}`")]
     ChainstateError(chainstate::ChainstateError),
```
TheQuantumPhysicist, line 122: ChainstateError implements the Display trait. No need to use `:?`.

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
     #[error("DatabaseFailure")]
     DatabaseFailure,
-    #[error("InvalidPeerId")]
-    InvalidPeerId,
+    #[error("Failed to convert data `{0}`")]
+    ConversionError(&'static str),
+    #[error("Other: `{0:?}`")]
```
TheQuantumPhysicist, line 127: This is also just an `&str`. No need for `:?`

TheQuantumPhysicist: There are many others. Please try to avoid `:?` as it makes reading logs much harder.

p2p/src/lib.rs
```diff
@@ -147,13 +153,15 @@ where
...
+    ) -> crate::Result<Self>
     where
         <T as NetworkingService>::Address: FromStr,
         <<T as NetworkingService>::Address as FromStr>::Err: Debug,
     {
         let (conn, pubsub, sync) = T::start(
-            bind_addr.parse::<T::Address>().map_err(|_| P2pError::InvalidAddress)?,
+            bind_addr
+                .parse::<T::Address>()
+                .map_err(|_| P2pError::ConversionError("Invalid address"))?,
```
TheQuantumPhysicist, line 164: Including the address here is useful to see it in the logs.

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum PublishError {
+    #[error("Message has already been published")]
+    Duplicate,
+    #[error("Failed to sign message")]
+    SigningFailed,
+    #[error("Not enough peers in topic")]
+    InsufficientPeers,
+    #[error("Message is too large")]
+    MessageTooLarge,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I would even include the limit that was crossed.

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
+    swarm::{handler::ConnectionHandlerUpgrErr, DialError::*},
+};
 use thiserror::Error;
 
-// TODO: think about which errors should be returned and when
-// TODO: store peerid where appropriate!
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum ProtocolError {
+    #[error("Peer is in different network")]
     DifferentNetwork,
```
TheQuantumPhysicist, line 26: Also useful here is to include the "difference". I imagine it's magic numbers, right? Why not include both?

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
+    #[error("Peer is incompatible")] // TODO: remove?
     Incompatible,
+    #[error("Peer is unresponsive")]
     Unresponsive,
+    #[error("Peer uses an invalid protocol")] // TODO: remove?
     InvalidProtocol,
+    #[error("Peer is an unknown network")] // TODO: remove?
     UnknownNetwork,
+    #[error("Peer is in an invalid state to perform this operation")]
     InvalidState,
```
TheQuantumPhysicist, line 40: How about including the expected state vs the current state?

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
 pub enum ProtocolError {
+    #[error("Peer is in different network")]
     DifferentNetwork,
+    #[error("Peer has an unsupported version")]
     InvalidVersion,
+    #[error("Peer sent an invalid message")]
     InvalidMessage,
+    #[error("Peer is incompatible")] // TODO: remove?
     Incompatible,
+    #[error("Peer is unresponsive")]
```
TheQuantumPhysicist, line 33: Perhaps add the definition of Unresponsive in the error text

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
     Unresponsive,
+    #[error("Peer uses an invalid protocol")] // TODO: remove?
     InvalidProtocol,
+    #[error("Peer is an unknown network")] // TODO: remove?
```
TheQuantumPhysicist, line 37: Expected vs current?

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
-    IdentifyError(String),
+pub enum PeerError {
+    #[error("Peer disconnected")]
+    PeerDisconnected,
+    #[error("No peers")]
+    NoPeers,
+    #[error("Peer doesn't exist")]
+    PeerDoesntExist,
+    #[error("Peer already exists")]
+    PeerExists,
```
TheQuantumPhysicist, line 52: `PeerAlreadyExists`?

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
+    #[error("Failed to compress the message")]
+    TransformFailed,
+}
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum DialError {
+    #[error("Peer is banned")]
+    Banned,
+    #[error("Limit for outgoing connections reached")]
+    ConnectionLimit,
```
TheQuantumPhysicist, line 74: Maybe include the connection limit

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
+    #[error("Tried to dial local node")]
+    LocalPeerId,
```
TheQuantumPhysicist, line 76: This doesn't sound like a crime unless one is call themselves. Functional tests do call local nodes. Why is this an error? Did you mean `AttemptedToDialToSelf`?

p2p/src/sync/peer.rs
altonen: The peer ID would be logged by the PeerManager but I can log it here too

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
 use thiserror::Error;
 
-// TODO: think about which errors should be returned and when
-// TODO: store peerid where appropriate!
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum ProtocolError {
+    #[error("Peer is in different network")]
     DifferentNetwork,
+    #[error("Peer has an unsupported version")]
     InvalidVersion,
```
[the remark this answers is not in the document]
altonen: The error is not used at the moment. I'll add the version(s) and whatever else is needed to the error message when the functionality is implemented.

p2p/src/error.rs
altonen: In libp2p, local refers to self so if the dial target is local peer ID, it means that the we're trying to dial "self". But I can change our error to say self

p2p/src/error.rs
altonen: I removed this error. The condition is covered by `DifferentNetwork`

p2p/src/error.rs
altonen: I removed the error, it's not used anywhere. I'll add it back if it's needed

p2p/src/error.rs
```diff
@@ -14,124 +14,133 @@
...
-    #[error("InvalidAddress")]
-    InvalidAddress,
-    #[error("InvalidData")]
-    InvalidData,
-    #[error("PeerExists")]
-    PeerExists,
+    #[error("Peer-related error: `{0:?}`")]
+    PeerError(PeerError),
+    #[error("Invalid data: `{0}`")]
+    InvalidData(&'static str),
```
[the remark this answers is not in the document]
altonen: I actually hate seeing that `to_string()` so much that I just added a new subcategory for conversion errors so that I don't have to use it. `P2pError::Other` is the only category that uses `&'static str` and it's reserved for errors that are fatal but don't belong to any other category and don't have any meaningful parameters to add to the error message. It's like the `DispatchError::Other` from Substrate.

`InvalidData` is bad error because you could argue that almost all errors are `InvalidData` as long as you define `Invalid` properly. It's gone.

TheQuantumPhysicist (approved):

PR #212, opened 2022-05-31
Prototype: Size-dependent encoding support

iljakuklic:
The idea is to utilise this as a building block to implement support for:
* Encoding/decoding size-dependent stuff (like public keys in Bitcoin, mostly for compatibility with bitcoin)
* Data structures that can be extended using soft forks (old nodes have to be able to skip unrecognized parts)

Fairly incomplete at this point but basic things work. There are some design decisions to be made still. There are instances for size-dependent coding for a two-element tuple and for `Either`, which readily generalise to structs and enums. I.e. derive macros which follow the same behaviour could be implemented.

Other building blocks to be implemented for full upgradability support:
1. Introduce a type that differentiates either a decoded value or a byte sequence reserved for future extension.
2. Structs extensible by adding extra fields
3. Enums extensible by adding extra variants

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3496,6 +3496,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "serialization-size-dependent"
+version = "0.1.0"
+dependencies = [
+ "either",
+ "hex-literal",
+ "parity-scale-codec",
+ "serialization-core",
+]
+
+[[package]]
 name = "serialization-tagged"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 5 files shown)


iljakuklic:
> How complete is this? Can it be merged?

I changed this to be a separate crate, similarly to other serialization "extensions". Looks like there is some missing stuff. Test coverage is not sufficient at the moment and some things would be nicer as derive macros. In other words, there is still some work to be done before I'd consider this to be something worth merging.

TheQuantumPhysicist:
We're closing this temporarily and we'll pick up working on it later since it's not a priority.

PR #213, opened 2022-06-01
Improve network interface

altonen:
Split the interface into two files: services (`net/mod.rs`) and types (`net/types.rs`) and improve the documentation for the interface

### DIFF
--- a/p2p/src/net/mod.rs
+++ b/p2p/src/net/mod.rs
@@ -23,7 +23,7 @@ pub mod libp2p;
 pub mod mock;
 pub mod types;
 
-/// `NetworkingService` provides the low-level network interface
+/// [NetworkingService] provides the low-level network interface
 /// that each network service provider must implement
 #[async_trait]
 pub trait NetworkingService {
@@ -35,13 +35,13 @@ pub trait NetworkingService {
     ///
     /// For an implementation built on libp2p, the address format is:
     ///     `/ip4/0.0.0.0/tcp/8888/p2p/<peer ID>`
-    type Address: Send + Sync + Debug + PartialEq + Eq + Hash + Clone + ToString;
+    type Address: Clone + Debug + Eq + Hash + Send + Sync + ToString;
 
     /// Unique ID assigned to a peer on the network
-    type PeerId: Send + Copy + PartialEq + Eq + Hash + Debug + Sync + ToString;
+    type PeerId: Copy + Debug + Eq + Hash + Send + Sync + ToString;
 
-	/// Unique ID assigned to each received request
-    type RequestId: Send + Debug + Eq + Hash + Sync;
+    /// Unique ID assigned to each received request
+    type RequestId: Debug + Eq + Hash + Send + Sync;
 
     /// Enum of different peer discovery strategies that the implementation provides
     type DiscoveryStrategy;
@@ -55,18 +55,23 @@ pub trait NetworkingService {
     /// Handle for sending/receiving pubsub-related events
     type PubSubHandle: Send;
 
-	/// Handle for sending/receiving request-response events
+    /// Handle for sending/receiving request-response events
     type SyncingCodecHandle: Send;
 
     /// Unique ID assigned to each pubsub message
-    type MessageId: Send + Clone + Debug;
+    type MessageId: Clone + Debug + Send;
 
     /// Initialize the network service provider
     ///
     /// # Arguments
     /// `bind_addr` - socket address for incoming P2P traffic
+    ///
     /// `strategies` - list of strategies that are used for peer discovery
+    ///
     /// `topics` - list of pubsub topics that the implementation should subscribe to
+    ///
+    /// `chain_config` - chain config of the node
+    ///
     /// `timeout` - timeout for outbound connections
     async fn start(
         bind_addr: Self::Address,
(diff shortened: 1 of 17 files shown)


p2p/tests/sync.rs
```diff
@@ -25,8 +25,8 @@ use p2p::{
     event::{PubSubControlEvent, SwarmEvent, SyncControlEvent},
```
TheQuantumPhysicist, line 25: Lots of unwrap in this file. Can we change it to expect with a reasonable message that explains why it should never fail?

p2p/src/pubsub.rs
```diff
@@ -164,9 +168,9 @@ where
...
                                 .await?
                             {
-                                Ok(_) => net::ValidationResult::Accept,
+                                Ok(_) => net::types::ValidationResult::Accept,
                                 Err(ProcessBlockError(BlockError::BlockAlreadyExists(_id))) =>
-                                    net::ValidationResult::Accept, // TODO: ignore?
+                                    net::types::ValidationResult::Accept, // TODO: ignore?
                                 Err(err) => {
                                     // TODO: report misbehaviour to swarm manager and close connection
                                     log::error!(
```
TheQuantumPhysicist, line 176: There's an overuse of `:?` all over the place. Please keep in mind that this makes logs much bigger and unreadable. Perhaps adding the Display trait to many things may be a good solution. Something like the PeerId, which is just a number, doesn't need to be printed as Debug trait. Maybe also MessageId? Even the errors too! You can postpone this to be done later. But we don't want our logs to be flooded with code symbols as much as possible.

p2p/src/net/mod.rs
```diff
@@ -196,43 +35,48 @@ pub trait NetworkingService {
...
-    type RequestId: Send + Debug + Eq + Hash + Sync;
+    /// Unique ID assigned to each received request
+    type RequestId: Debug + Eq + Hash + Send + Sync;
 
     /// Enum of different peer discovery strategies that the implementation provides
     type DiscoveryStrategy;
 
     /// Id that identifies a protocol
-    type ProtocolId: Debug + Send + Clone + PartialEq;
+    type ProtocolId: Clone + Debug + PartialEq + Send;
```
TheQuantumPhysicist, line 50: There's absolutely no reason I'm aware of to require PartialEq and not Eq unless we're dealing with floating point numbers. Either both or nothing, please.

TheQuantumPhysicist: Btw, this problem exists in many places, not just here.

TheQuantumPhysicist (approved):
Besides the few comments, all good.

p2p/tests/sync.rs
altonen: These are tests and after taking a quick look most of them were about `ChannelClosed`/subsystem wrapper failure. I'm not sure if converting them to `expect()` is worth the effort.

p2p/src/pubsub.rs
altonen: I'll convert them to use `Display` as this maintenance journey takes me through P2P. I want to remind though that almost none of the errors of P2P had implement `Display` until two days ago so it's not that surprising they are printed using `{:?}`.

p2p/src/net/mod.rs
altonen: I'm not sure what you mean. I removed `PartialEq` from `PeerId` and `Address` as it was not used but `ProtocolId` needs it. It doesn't need `Eq` though so I'm not entirely sure what you're saying.

p2p/tests/sync.rs
TheQuantumPhysicist: Apologies, I kind of missed that this is just tests.

iljakuklic: I still see a bunch of clippy warnings regarding `unwrap` in the `p2p/test-utils` crate. I think we can just `allow` them in that crate.

altonen: I can fix the unwrap issue in another PR when I'm modifying those files so we don't have to make exceptions.

p2p/src/net/mod.rs
iljakuklic: As far as I can tell, this is only used in ad-hoc `PartialEq` instance for `PeerInfo` in libp2p tests.

A couple of observations:
1. The `PeerInfo` impl could be restricted locally using a `where` clause, instead of restricting `ProtocolId` globally
2. It would be cleaner for `PeerInfo` to provide `Eq` as well provided corresponding `NetworkingService::ProtocolId` implements `Eq`, but I am not too bothered by that since it is only in tests
3. The impl is required for `assert_eq!`. While this strictly only requires `PartialEq`, it is reasonable to expect total equality here. I don't see a simple way of adding the additional restriction and also not probably not wort hit.
4. Trait impls are visible "globally", you just need to import the trait and have a value of the corresponding type. This applies even if the trait impl itself is buried in a submodule. So it's better to get trait impls right.
5. However, the `PeerInfo` impl is in test module and as such is only compiled in when testing, so (4) only becomes an issue when someone moves the impl out of the test module without realizing full implications
6. Having laid out all the possible issues here, I still think this is a very low-risk issue.

iljakuklic (approved):

p2p/src/net/mod.rs
altonen: IIRC the reason why `PeerInfo` manually implemented `PartialEq` was because it contained `Arc<T>`. I'm right now in the process of trying to remove that because hopefully it's not needed. I agree that `PartialEq` implementation should be visible globally and I will make it visible, if the `ProtocolId`/`PeerInfo` in its current form survive the maintenance period. We can revisit this issue when the `PeerManager` implementation is in better condition and it's more clear what is needed.

PR #216, opened 2022-06-07
Unify read/write and read-only chainstate interface to avoid code duplication

TheQuantumPhysicist:


chainstate/src/detail/mod.rs
```diff
@@ -374,23 +366,14 @@ impl Chainstate {
...
 
-pub(crate) struct ChainstateRef<'a> {
+pub(crate) struct ChainstateRef<'a, S> {
     chain_config: &'a ChainConfig,
-    // TODO: make this generic over Rw and Ro
-    db_tx: TxRw<'a>,
-    orphan_blocks: &'a mut OrphanBlocksPool,
+    db_tx: S,
+    // TODO: get rid of the Option<>. The Option is here because mutability abstraction wasn't done for orphans while it was done for db transaction
+    orphan_blocks: Option<&'a mut OrphanBlocksPool>,
```
iljakuklic, line 373: Any ideas regarding how and when to do this?

iljakuklic (approved):
While there is the ugly workaround for `OrphanBlocksBool`, this is an improvement on balance.

chainstate/src/detail/mod.rs
TheQuantumPhysicist: Well, we'll have to play the same game and define traits for read and write. I don't see a way around it. That, assuming we don't get rid of orphans altogether in the future.

Nevertheless, given that orphans are only relevant in rw, it's not important right now.

PR #217, opened 2022-06-07
Isolate ChainstateRef and make its internals only accessible with a db transaction

TheQuantumPhysicist:


TheQuantumPhysicist:
This PR isn't really that big. But it looks big because it's mostly moving stuff around.

iljakuklic (approved):

PR #218, opened 2022-06-07
Fix the utxo crate description

stanislav-tkach:
The description of the `utxo` crate is probably copy-paste from the precious line (`utils`). I have tried to give a better description. I'm just getting acquainted with the project, so it probably can be improved.

### DIFF
--- a/utxo/src/utxo_impl/simulation.rs
+++ b/utxo/src/utxo_impl/simulation.rs
@@ -112,7 +112,8 @@ fn populate_cache<'a>(
     (cache, outps)
 }
 
-// #[test]
+#[ignore]
+#[test]
 fn stack_flush_test() {
     let mut outps: Vec<OutPoint> = vec![];
 
--- a/utxo/src/utxo_impl/simulation.rs
+++ b/utxo/src/utxo_impl/simulation.rs
@@ -112,6 +112,7 @@ fn populate_cache<'a>(
     (cache, outps)
 }
 
+// TODO: Fix the test (https://github.com/mintlayer/mintlayer-core/issues/219).
 #[ignore]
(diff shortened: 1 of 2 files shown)


utxo/src/utxo_impl/simulation.rs
```diff
@@ -112,7 +112,8 @@ fn populate_cache<'a>(
     (cache, outps)
 }
 
-// #[test]
+#[ignore]
+#[test]
```
stanislav-tkach, line 116: Perhaps I’m missing something, but it seems better to me to mark the test as ignored instead of commenting it.

TheQuantumPhysicist (approved):

utxo/src/utxo_impl/simulation.rs
TheQuantumPhysicist: Actually this may be a reminder for looking into this again later because this crate isn't used yet. Marking it ignore shows it's intentional. It's OK to mark it as ignored but please add a TODO to reconsider enabling this test.

stanislav-tkach: Unfortunately, the test is failing now, so I have created a separate issue for that: https://github.com/mintlayer/mintlayer-core/issues/219
I can try to fix it, but I will probably need some guidance.

muursh (approved):

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
Thank you for fixing this. Btw, @stanislav-tkach please add a PGP key and sign your commits that you push to this repo. Git can do that automatically for you. It's a requirement for this repo. I had to force the merge because the commits are not signed.

PR #221, opened 2022-06-08
Replace the staging branch with master

stanislav-tkach:
As far as I can see, there is no `staging` branch.


TheQuantumPhysicist:
I'm not sure we want to do this change. We will have staging soon enough when the software is useful for users. What do you think, @muursh ?

stanislav-tkach:
@TheQuantumPhysicist I have added a PGP key, but there is still no "verified" badge near my commit. I'm going to recheck everything.

muursh:
You need to sign your commits too `git commit -S -m "commit message"`

muursh:
And yep I'm with @TheQuantumPhysicist here. I'd say leave this. The README is technically wrong now and it was copied across from an older repo where we did use staging to base development from. As it stands we don't have a truly stable release to protect so we're basing everything from master right now. If we merged this we'd have to revert it fairly soon I'd guess, probably in a month or so we'll reintroduce the staging branch and the old dev work flow

PR #222, opened 2022-06-09
Introduce ban score

TheQuantumPhysicist:
This PR basically maps every error from updating the chainstate to a ban score. I encourage scrutinizing the decisions I made as much as possible. Also please understand that we prefer to be more conservative and hostile towards errors. Some times it's difficult to make a decision on whether to consider something malicious. In that case, we just consider it malicious.

A ban score follows the bitcoin definition. A peer has a maximum ban score of 100. We ban peers that reach 100. Nodes are free to define their own max, which is why we don't define a "max" and assign it to all values. Nodes can choose to tolerate malicious behavior.

### DIFF
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -5,6 +5,8 @@ use super::{
     CheckBlockTransactionsError, ConsensusVerificationError, OrphanCheckError,
 };
 
+// TODO: use a ban_score macro in a form similar to thiserror in order to define the ban score value of an error
+
 pub trait BanScore {
     fn ban_score(&self) -> u32;
 }
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -5,7 +5,7 @@ use super::{
     CheckBlockTransactionsError, ConsensusVerificationError, OrphanCheckError,
 };
 
-trait BanScore {
+pub trait BanScore {
     fn ban_score(&self) -> u32;
 }
 
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -5,7 +5,8 @@ use super::{
     CheckBlockTransactionsError, ConsensusVerificationError, OrphanCheckError,
 };
 
-// TODO: use a ban_score macro in a form similar to thiserror in order to define the ban score value of an error
+// TODO: use a ban_score macro in a form similar to thiserror::Error in order to define the ban score
+//       value of an error on the error enum arms instead of separately like in this file
 
 pub trait BanScore {
     fn ban_score(&self) -> u32;
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -1,0 +1,143 @@
+use crate::BlockError;
+
+use super::{
+    pow::error::ConsensusPoWError, spend_cache::error::StateUpdateError, CheckBlockError,
+    CheckBlockTransactionsError, ConsensusVerificationError, OrphanCheckError,
+};
+
+trait BanScoreFromError {
+    fn ban_score(&self) -> u32;
+}
+
+impl BanScoreFromError for BlockError {
+    fn ban_score(&self) -> u32 {
+        match self {
+            BlockError::StorageError(_) => 0,
+            BlockError::OrphanCheckFailed(err) => err.ban_score(),
+            BlockError::CheckBlockFailed(err) => err.ban_score(),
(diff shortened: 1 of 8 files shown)


chainstate/src/detail/ban_score.rs
```diff
@@ -0,0 +1,143 @@
...
+            BlockError::OrphanCheckFailed(err) => err.ban_score(),
+            BlockError::CheckBlockFailed(err) => err.ban_score(),
+            BlockError::StateUpdateFailed(err) => err.ban_score(),
+            BlockError::BestBlockLoadError(_) => 0,
+            BlockError::InvariantErrorFailedToFindNewChainPath(_, _, _) => 0,
+            BlockError::InvariantErrorInvalidTip => 0,
+            BlockError::InvariantErrorPrevBlockNotFound => 0,
+            // Even though this should've been caught by orphans check, its mere presence means a peer sent a block they're not supposed to send
+            BlockError::PrevBlockNotFound => 100,
+            BlockError::InvalidBlockSource => 100,
```
iljakuklic, line 25: Is `100` the threshold at which a peer gets banned? Maybe there could be a constant for that.

TheQuantumPhysicist: Yes. As described in the PR top message, the purpose for not having this be a constant is because the node operator will have the choice to change this to another value. It's a fair question what's the best way to write this number, but honestly I'm not too concerned about that point right now. It's just a constant being thrown around. I'm even considering making something like `thiserror` where we define the value in the error itself. Not the time for niceties right now though.

muursh (approved):
I'm good with this. Erring on the side of banning if there's even a slight doubt makes me happy. As you mention in the pr this needs testing but yeah looks good to me

chainstate/src/detail/ban_score.rs
iljakuklic: Sure I see your points. I still think a named constant would be nice to establish some sort of reference value at which a peer gets banned by default, even if the numeric value is completely arbitrary

iljakuklic (approved):

chainstate/src/detail/ban_score.rs
muursh: > I still think a named constant would be nice

I'd probably vote for this over the magic number too in all honesty

muursh: I'm fine with updating this later

chainstate/src/detail/ban_score.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use super::{
+    pow::error::ConsensusPoWError, spend_cache::error::StateUpdateError, CheckBlockError,
+    CheckBlockTransactionsError, ConsensusVerificationError, OrphanCheckError,
+};
+
+// TODO: use a ban_score macro in a form similar to thiserror::Error in order to define the ban score
+//       value of an error on the error enum arms instead of separately like in this file
+
+pub trait BanScore {
```
altonen, line 11: I don't think `chainstate` is the right place for this trait, if anything it should be in P2P but I can understand how that might create a weird dependency. I'd rather see this in `common` because P2P and maybe some other subsystems also use this and I think ban score is not really a "detail of chainstate" but rather a core part of swarm management.

chainstate/src/detail/ban_score.rs
```diff
@@ -0,0 +1,145 @@
...
+        match self {
+            CheckBlockError::StorageError(_) => 0,
+            CheckBlockError::MerkleRootMismatch => 100,
+            CheckBlockError::WitnessMerkleRootMismatch => 100,
+            CheckBlockError::BlockConsistencyError(_) => 100,
+            CheckBlockError::InvalidBlockNoPrevBlock => 100,
+            // even though this may be an invariant error, we treat it strictly
+            CheckBlockError::PrevBlockNotFound(_, _) => 100,
+            CheckBlockError::BlockTimeOrderInvalid => 100,
+            CheckBlockError::BlockFromTheFuture => 100,
```
altonen, line 99: This sounds quite harsh, considering the clocks are not going to be perfectly in sync. I think we should agree on an acceptable clock disparity and handle this error more gracefully.

chainstate/src/detail/ban_score.rs
TheQuantumPhysicist: As discussed on slack, this whole structure will change when we attempt to use a semi-derive macro to define the scores.

chainstate/src/detail/ban_score.rs
TheQuantumPhysicist: The current implementation is incorrect. This should be a few hours into the future for this to be an error.

PR #223, opened 2022-06-09
serialization: Tagless enum encoding + decoding

iljakuklic:
Add support for encoding mode where enum does not attach an extra byte
to determine which variant to take. Instead, the variant is determined
by the first byte specified by the contained type.

There are multiple components:
* The `Tagged` trait that marks types with a costant initial byte.
* A mechanism to automatically derive the `Tagged` trait.
* The `DirectEncode` and `DirectDecode` derive macros that implement `Encode` and `Decode` without placing the initial byte first. The `Tagged` trait is used to determine the correct enum variant.

The code is not yet integrated into the main code but it is in a pretty good shape and tested. The integration could be done as a follow-up


serialization/tests/derive.rs
```diff
@@ -0,0 +1,229 @@
...
+
+fn check_initial_byte<T: Tagged + Encode>(x: &T) {
+    x.using_encoded(|encoded| {
+        assert_eq!(encoded[0], T::TAG, "tag mismatch");
+    });
+}
+
+fn check_codec<T: Eq + std::fmt::Debug + Encode + Decode>(original: &T) {
+    let encoded = original.encode();
+    let decoded = T::decode(&mut &encoded[..]).ok();
```
TheQuantumPhysicist, line 100: In tests, I always prefer to use unwrap() instead of compared with Some() or None. The benefit is that when you unwrap, an error will print the error for you.

serialization/src/tagged.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// Derive [Decode] for decoding without the initial discriminant byte
+pub use serialization_derive::DirectDecode;
+
+use crate::{Decode, Encode, Input};
+
+/// Types whose encoding starts with a fixed byte (tag)
+pub trait Tagged {
+    /// Initial byte in encoding of `Self`
+    const TAG: u8;
```
TheQuantumPhysicist, line 141: Can the tag be of a custom type? Nothing crazy, but other stack-based types, like u32 or whatever other size we need?

iljakuklic: I don't currently see a simple way to introduce the option to make this type customisable. It's takes a part in a bunch of const generics. Even if we figure out how to do it, I don't think it would be worth the extra complexity.

We could pick a different fixed type but I decided for `u8` for the following reasons:
* Compatibility with how are enums represented in SCALE (e.g. porting the existing Transaction enum to this would not change the encoding)
* I think it is unlikely that more than 256 versions/variants will be needed
* Even if over 256 variants are needed for some stuff
  * that is easily worked around by having nested versions when the need arises
  * the types where over 256 variants are not needed would have a needlessly wide tag representation

serialization/tests/derive.rs
iljakuklic: I changed this to compare to `Ok`. That way the error is not lost and we get both the error and expected value printed out on failure. I am not sure why I converted to `Option` in the first place, probably some sort of error that I assumed would be solved by it so it's an oversight on my part.

common/src/chain/block/block_v1.rs
```diff
@@ -4,17 +4,17 @@ use crate::chain::transaction::Transaction;
...
 use super::consensus_data::BlockRewardTransactable;
-use super::BlockVersion;
 
-#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode, serialization::Tagged)]
 pub struct BlockHeader {
+    pub(super) version: Version<1>,
     #[codec(compact)]
-    pub(super) block_version: u32,
+    pub(super) _reserved: u32,
```
TheQuantumPhysicist, line 17: Could you please explain how this could help?

iljakuklic: I just changed the original `block_version` to something else so that the encoding stays the same and tests that hard-code block binary data don't break without updating them for now. Can be removed later.

iljakuklic: Removed this

iljakuklic:
Marked this as ready for review
* Fixed the test fauilures
* In addition to `Block` I have also ported `Transaction` to the direct encoding.
* Review note: `serialization/src/tagged.rs` is a good start where the top-level doc comment explains the goals and the design

serialization/src/tagged.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+    };
+}
+// Up to 8 elements for now, can be expanded if needed
+impl_tagged_for_tuple!(T0,);
+impl_tagged_for_tuple!(T0, T1);
+impl_tagged_for_tuple!(T0, T1, T2);
+impl_tagged_for_tuple!(T0, T1, T2, T3);
+impl_tagged_for_tuple!(T0, T1, T2, T3, T4);
+impl_tagged_for_tuple!(T0, T1, T2, T3, T4, T6);
```
muursh, line 169: Where is T5?

serialization/Cargo.toml
```diff
@@ -7,9 +7,13 @@ license = "MIT"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+serialization-derive = { path = 'derive' }
+
 arraytools = "0.1.5"
 hex-literal = "0.3.4"
 parity-scale-codec = {version = "3.1.2", features = ["derive", "chain-error"]}
+rand = "0.8.5"
```
muursh, line 15: I guess this should be moved to the crypto subsystem randomness source

common/src/chain/block/block_v1.rs
```diff
@@ -4,17 +4,15 @@ use crate::chain::transaction::Transaction;
...
 
 use super::consensus_data::BlockRewardTransactable;
-use super::BlockVersion;
 
-#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode, serialization::Tagged)]
 pub struct BlockHeader {
-    #[codec(compact)]
-    pub(super) block_version: u32,
+    pub(super) version: Version<1>,
```
TheQuantumPhysicist, line 15: If I understand correctly how this works, the value of the version for this block is 0x01 (maybe we should choose 0x00), attempting to deserialize data that leads to some other value, say 0x02, in the version will cause a deserialization error.

common/src/primitives/version.rs
```diff
@@ -14,7 +14,17 @@
 // limitations under the License.
 //
 // Author(s): A. Altonen
-use serialization::{Decode, Encode};
+use serialization::{Decode, Encode, Tagged};
+
+/// Version tag for SCALE-encoded values
+#[derive(Debug, Ord, PartialOrd, PartialEq, Eq, Encode, Decode, Copy, Clone, Tagged)]
+pub struct Version<const V: u8>(serialization::Tag<V>);
```
TheQuantumPhysicist, line 21: As it stands now, I think this may be confusing. Can you please rename this to TagVersion, and move it to its own file tag_version.rs

The purpose here is to define a good name for this struct with its properties. Just calling it version will make discussions harder. We need a good name for it.

Later we'll also move SemVer to semver.rs

common/src/primitives/version.rs
```diff
@@ -65,6 +75,11 @@ impl TryFrom<String> for SemVer {
 mod tests {
     use super::*;
 
+    #[test]
+    fn version1_encodes_to_1() {
+        assert_eq!(Version::<1>::default().encode(), vec![1]);
```
TheQuantumPhysicist, line 80: I know this is trivial... but a few values would be good :smile:

serialization/src/derive_support.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub fn new(inner: &'a mut I) -> Self {
+        Self { init: None, inner }
+    }
+
+    /// Peek the next byte
+    pub fn peek(&mut self) -> Result<u8, crate::Error> {
+        match self.init {
+            Some(b) => Ok(b),
+            None => {
+                let b = self.inner.read_byte()?;
```
TheQuantumPhysicist, line 42: I'm confused on what "peek" means. In this context, I thought it means reading a stream without changing the state of the stream. But:

- We're using read_byte(), which consumes a byte
- inner is mutable.

Either "peek" is the wrong word for this or I'm misunderstanding how this works.

serialization/src/derive_support.rs
```diff
@@ -0,0 +1,125 @@
...
+
+//! Implementation tools used by code generated by derive macros
+#![doc(hidden)]
+
+pub use static_assertions as sa;
+
+use crate::Input;
+
+/// Input byte stream with a one-byte lookahead
+pub struct Peekable<'a, I> {
```
TheQuantumPhysicist, line 26: Isn't it better to just make this a trait implemented for I instead of a struct that takes a reference THEN does the same thing to it? From what I'm seeing, the reference wrapper doesn't serve any purpose. Maybe I'm missing something.

common/src/chain/block/block_v1.rs
iljakuklic: Correct. Serializing `Version<1>` generates one byte of 0x01, deserializing consumes 0x01 and fails if it's not there

iljakuklic: I chose 1 for compatibility. In block it does not matter because there are more changes for block. In Transaction, it means the encoding is same as it was before

TheQuantumPhysicist: Let's change it some time later.

common/src/primitives/version.rs
iljakuklic: Not sure about `TagVersion`. Maybe `VersionByte`? I find `VersionTag` slightly better if you feel tag should be there

TheQuantumPhysicist: VersionTag is OK. I'd rather not put type information in the name.

serialization/src/tagged.rs
iljakuklic: Hmm maybe I should go back to elementary school. Fixing.

serialization/Cargo.toml
TheQuantumPhysicist: Note that this is here because serialization cannot depend on the crypto crate because the crypto crate depends on serialization. If we want to brute-force the randomness dependency, we should move random out of crypto to its own crate.

TheQuantumPhysicist (approved):
The PR looks OK besides the few comments I made. I'm mostly concerned about Peekable and the comments around it. I feel it can be stripped down heavily.

serialization/src/derive_support.rs
iljakuklic: It refers to how it behaves from the point of view of the user of the `Peekable` type. Peeking may (or may not) advance the `inner` input but `Peekable` still behaves as if it was not consumed, i.e. peeking and reading a byte afterwards gives the same value.

serialization/Cargo.toml
muursh: Gotcha. Makes sense to keep it imported directly then.

serialization/src/derive_support.rs
iljakuklic: I don't think it is doable the way you describe. `Decode::decode` has to work for any type that implements `Input`. The new trait would have to be implemented for all types that implement `Input`.

PR #229, opened 2022-06-14
Introduce expanded block size checking that limits size based on transaction type

TheQuantumPhysicist:


common/src/chain/config.rs
```diff
@@ -147,7 +162,9 @@ const REGTEST_ADDRESS_PREFIX: &str = "rmt";
 // If block time is 2 minutes (which is my goal eventually), then 500 is equivalent to 100 in bitcoin's 10 minutes.
 const MAINNET_BLOCKREWARD_MATURITY: BlockDistance = BlockDistance::new(500);
 // DSA allows us to have blocks up to 1mb
-pub const MAX_BLOCK_WEIGHT: usize = 1_048_576;
+const MAX_BLOCK_HEADER_SIZE: usize = 1024;
+const MAX_BLOCK_TXS_SIZE: usize = 524_288;
+const MAX_BLOCK_CONSTRACTS_SIZE: usize = 524_288;
```
muursh, line 167: Do we want this 50/50? I'm fine with this for now but I think we should revisit this and consider having more space for TXs.

chainstate/src/detail/chainstateref.rs
```diff
@@ -307,8 +304,25 @@ impl<'a, S: BlockchainStorageRead> ChainstateRef<'a, S> {
...
+
+    fn check_block_size(&self, block: &Block) -> Result<(), CheckBlockError> {
+        let block_size = block.block_size();
+
+        if block_size.size_from_header() > self.chain_config.max_block_header_size() {
+            return Err(CheckBlockError::BlockTooLarge);
+        }
+
+        if block_size.size_from_txs() > self.chain_config.max_block_size_from_txs() {
+            return Err(CheckBlockError::BlockTooLarge);
```
muursh, line 320: Is it worth splitting `BlockTooLarge` so that too many tx or too many contracts are separate issues?

TheQuantumPhysicist: I'm still undecided, TBH. This occurred to me. Torn between over engineering and good tracing.

muursh: I'd vote for splitting it.

TheQuantumPhysicist: Alrighty. I'll go for it.

common/src/chain/config.rs
iljakuklic, line 167: typo: `CONSTRACTS`

iljakuklic: There is also the possibility of having total size (say 1MB) and a limit on contract space (0.5 MB for example). Implying TXs have 0.5MB reserved and could compete with contract stuff for the other 0.5MB so it could be up to full 1MB TXs given enough demand.

chainstate/src/detail/chainstateref.rs
```diff
@@ -307,8 +304,25 @@ impl<'a, S: BlockchainStorageRead> ChainstateRef<'a, S> {
...
-        if block.encoded_size() > MAX_BLOCK_WEIGHT {
+        self.check_block_size(block)?;
+
+        Ok(())
+    }
+
+    fn check_block_size(&self, block: &Block) -> Result<(), CheckBlockError> {
+        let block_size = block.block_size();
+
+        if block_size.size_from_header() > self.chain_config.max_block_header_size() {
```
altonen, line 315: I think using `ensure!()` here would be slightly nicer

altonen (approved):

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: Done

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: Done

muursh (approved):

PR #232, opened 2022-06-15
Create custom time getter option for Chainstate for unit tests

TheQuantumPhysicist:


altonen (approved):

chainstate/src/detail/chainstateref.rs
```diff
@@ -285,16 +283,15 @@ impl<'a, S: BlockchainStorageRead> ChainstateRef<'a, S> {
 
         match &block.prev_block_id() {
             Some(prev_block_id) => {
-                let median_time_past = self.calculate_median_time_past(prev_block_id);
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

chainstate/src/detail/chainstateref.rs
```diff
@@ -275,34 +274,36 @@ impl<'a, S: BlockchainStorageRead> ChainstateRef<'a, S> {
...
-                if i64::from(block.block_time()) > time::get() + max_future_offset.as_secs() as i64
-                {
-                    // TODO: test submitting a block that fails this
-                    return Err(CheckBlockError::BlockFromTheFuture);
-                }
+                let median_time_past = calculate_median_time_past(self, prev_block_id);
+                ensure!(
+                    block.block_time() >= median_time_past,
+                    CheckBlockError::BlockTimeOrderInvalid
+                );
```
iljakuklic, line 291: A very minor nitpick:

You can add a tailing comma at the end, to make it consistent with how function calls are usually formatted:
```rust
                ensure!(
                    block.block_time() >= median_time_past,
                    CheckBlockError::BlockTimeOrderInvalid,
                );
```

Sadly rustfmt cannot make any assumptions on whether the macro is able to deal with this or not so it conservatively leaves the original punctuation there.

chainstate/src/detail/mod.rs
```diff
@@ -120,13 +130,15 @@ impl Chainstate {
...
         custom_orphan_error_hook: Option<Arc<OrphanErrorHandler>>,
+        custom_time_getter: Option<Arc<TimeGetter>>,
     ) -> Result<Self, crate::ChainstateError> {
         let cons = Self {
             chain_config,
             blockchain_storage,
             orphan_blocks: OrphanBlocksPool::new_default(),
             custom_orphan_error_hook,
             events_controller: EventsController::new(),
+            time_getter: custom_time_getter.unwrap_or_else(|| Arc::new(time::get)),
```
iljakuklic, line 141: If `TimeGetter` was a type wrapper over `Arc<dyn Fn() -> i64 + Send + Sync>`, you could implement `Default` for it and use that (instead of `Some`/`None`) to signify whether the default or a custom time getter is requested.

iljakuklic (approved):
One more minor suggestion. LGTM otherwise

chainstate/src/detail/mod.rs
TheQuantumPhysicist: Actually I consider that playing with fire. A default for the `Arc<dyn Fn() -> i64 + Send + Sync>` type doesn't semantically mean anything... by coincidence we have an alias for that type, but by sheer chance it's related to time. On one hand, consider the confusion by some outsider reading the code. Also consider someone importing the wrong Default because of their auto-complete functions in their IDE. Leave alone the case where we, again, by coincidence, we need another function that has the same signature for something else... and guess what default it'll get.

I like the idea of Default in rust, but I consider it one of those slippery slopes that can easily delete the benefit of type-safety. Same issue with From. If a type has a single `Vec<H256>` (think of locators, for example) in it, it doesn't mean that every `Vec<H256>` should be convertible to a locator. Carla almost did that mistake with BlockUndo. That's dangerous territory.

TheQuantumPhysicist: Ah, I see. Yeah, that's not bad at all. Let me look into that. Apologies, I misunderstood.

TheQuantumPhysicist: I implemented this change. Take a look if you'd like to critique it.

chainstate/src/detail/mod.rs
```diff
@@ -130,15 +133,15 @@ impl Chainstate {
         chain_config: Arc<ChainConfig>,
         blockchain_storage: blockchain_storage::Store,
         custom_orphan_error_hook: Option<Arc<OrphanErrorHandler>>,
-        custom_time_getter: Option<Arc<TimeGetter>>,
+        custom_time_getter: TimeGetter,
```
iljakuklic, line 136: This can be renamed to just `time_getter`

iljakuklic (approved):
Looks good

chainstate/src/detail/mod.rs
TheQuantumPhysicist: Done

muursh (approved):

PR #233, opened 2022-06-15
Refactor swarm code

altonen:
This is one of many swarm PRs that implement the missing functionality. This PR is big but it mainly splits the implementation into `PeerManager` and `PeerDb`, splits larger functions into smaller functions, adds some validations for connections/peer information and adds a bunch of documentation so it shouldn't be too bad.

It doesn't implement any new functionality but converts the old auto connect implementation into this new `PeerManager::heartbeat()` function that manages the overall connectivity state.


common/src/primitives/version.rs
```diff
@@ -61,6 +61,12 @@ impl TryFrom<String> for SemVer {
     }
 }
 
+impl std::fmt::Display for SemVer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "v{}.{}.{}", self.major, self.minor, self.patch)
```
TheQuantumPhysicist, line 66: Perhaps it's better to move that v out of here. I don't see the benefit in forcing everyone to format their version with a v prefix.

p2p/src/swarm/mod.rs
```diff
@@ -96,314 +86,395 @@ where
...
-            }
-            net::types::ConnectivityEvent::ConnectionAccepted { peer_info } => {
-                let peer_id = peer_info.peer_id;
-                log::debug!("outbound connection accepted by peer {:?}", peer_id);
+    /// Maintain the swarm state
+    ///
+    /// `PeerManager::heartbeat()` is called every time a network/control event is received
+    /// or the heartbeat timer of the event loop expires. In other words, the swarm state
+    /// is checked and updated at least once every 30 seconds. In high-traffic scenarios the
+    /// update interval is clamped to a sensible lower bound.
```
TheQuantumPhysicist, line 297: Maybe briefly explain how it's lowered (what part does it)

p2p/src/swarm/mod.rs
```diff
@@ -14,46 +14,33 @@
...
-// TODO: store other discovered addresses
-#[derive(Debug)]
-struct PeerContext<T>
-where
-    T: NetworkingService,
-{
-    _info: net::types::PeerInfo<T>,
-}
+/// Maximum number of connections the [`PeerManager`] is allowed to have open
+const MAX_ACTIVE_CONNECTIONS: usize = 128;
```
TheQuantumPhysicist, line 40: I may have mentioned this before, but this should be a command line argument. Not necessarily right now to be done, but a TODO wouldn't be bad.

p2p/src/swarm/mod.rs
```diff
@@ -96,314 +86,395 @@ where
...
+            };
+
+            // handle error, exit on fatal errors and finally update peer manager state
+            self.handle_error(result)?;
+            self.heartbeat().await?;
         }
     }
 }
 
 #[cfg(test)]
```
TheQuantumPhysicist, line 470: Perhaps move the tests to a submodule? The file is becoming huge.

p2p/src/swarm/peerdb.rs
```diff
@@ -0,0 +1,267 @@
...
+        /// Hashset of IPv4 addresses
+        ip4: HashSet<T::Address>,
+
+        /// Hashset of IPv6 addresses
+        ip6: HashSet<T::Address>,
+    },
+}
+
+// TODO: improve how peers are stored, order by reputation?
+pub struct PeerDb<T: NetworkingService> {
```
TheQuantumPhysicist, line 56: It's also necessary to store those on disk. But that's probably something can be done later. Bitcoin uses a custom format for storing peers in the data directory. If you want to go through with that, please make sure it's mockable.

p2p/src/swarm/mod.rs
```diff
@@ -14,46 +14,33 @@
...
+
+#![allow(rustdoc::private_intra_doc_links)]
 use crate::{
-    error::{FatalError, P2pError, PeerError, ProtocolError},
+    error::{P2pError, PeerError, ProtocolError},
     event,
     net::{self, ConnectivityService, NetworkingService},
 };
-use common::chain::ChainConfig;
+use common::{chain::ChainConfig, primitives::version};
```
iljakuklic, line 30: We should probably rename `primitives::version` to `primitives::semver`, so it is not confused with the newly introduced `primitives::version_tag`

iljakuklic (commented):
A bunch of minor comments. As usual, I still don't understand `p2p` well enough to be confident to approve but looks fine from what I can tell.

p2p/src/swarm/mod.rs
altonen: Done

p2p/src/swarm/mod.rs
altonen: Yeah it's on the TODO list but not with a high priority. I'll add configs to P2P at some point

p2p/src/swarm/mod.rs
altonen: If you're talking about the lower bound then my thought was just to clamp it to like 5 seconds or something. Each time `heartbeat()` is called it checks when it was last called (when it executed the whole function) and if it's more than 5 seconds (or whatever the lower bound is configured to be), it exits early.

p2p/src/swarm/peerdb.rs
altonen: How about SCALE-encoded binary blobs, wouldn't that be lovely. I created an issue, #241 and I'll resolve this for now.

p2p/src/swarm/mod.rs
```diff
@@ -96,314 +86,395 @@ where
...
+    /// - listening to network events
+    /// - updating internal state
+    ///
+    /// After handling an event from one of the aforementioned sources, the event loop
+    /// handles the error (if any) and runs the [`PeerManager::heartbeat()`] function
+    /// to perform swarm maintenance. If the `PeerManager` doesn't receive any events,
+    /// [`PEER_MGR_HEARTBEAT_INTERVAL`] defines how often the heartbeat function is called.
+    /// This is done to prevent the `PeerManager` from stalling in case the network doesn't
+    /// have any events.
     pub async fn run(&mut self) -> crate::Result<()> {
```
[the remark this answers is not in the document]
altonen: Fixed. This needs to be fixed in other subsystems as well but I added them on my TODO list and I'll fix them when I start working on them.

p2p/src/swarm/mod.rs
TheQuantumPhysicist: I meant in the comment mention where that mechanism is done :)

altonen: Done

PR #236, opened 2022-06-16
Use Duration instead of integer for mocked time to minimize unit conversion errors

TheQuantumPhysicist:


chainstate/src/detail/block_index_history_iter.rs
```diff
@@ -64,17 +64,18 @@ mod tests {
...
-            let mut chainstate = Chainstate::new(chain_config.clone(), storage, None).unwrap();
+            let mut chainstate =
+                Chainstate::new(chain_config.clone(), storage, None, None).unwrap();
 
             // put three blocks in a chain after genesis
             let block1 = Block::new(
                 vec![],
                 Some(Id::new(&chainstate.chain_config.genesis_block_id().get())),
-                time::get() as u32,
+                time::get().as_secs() as u32,
```
stanislav-tkach, line 78: It isn't directly related to this PR,  but maybe we want for `Block::new` to accept `Duration` instead of `u32` in order to get rid of all these `time::get().as_secs() as u32`?

TheQuantumPhysicist: That's going to be the next step. This is just the beginning. There will be a BlockTimestamp type that will solve this problem.

One step at a time ;-)

muursh (approved):
No issues from me

common/src/primitives/time.rs
```diff
@@ -17,48 +17,52 @@
 #![allow(unused, dead_code)]
 use lazy_static::lazy_static;
 use logging::log;
-use std::sync::atomic::{AtomicI64, Ordering};
-use std::time::SystemTime;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::{Duration, SystemTime};
+
+fn duration_to_int(d: &Duration) -> Result<u64, Box<dyn std::error::Error>> {
```
iljakuklic, line 23: Out of curiosity: why a boxed error?

iljakuklic (approved):

common/src/primitives/time.rs
TheQuantumPhysicist: Btw, this is the idiomatic way to make generic errors in rust. I'm not sure this idea is already on the table before answering before.

PR #238, opened 2022-06-16
all: Change some `unwrap()` calls to `expect()`

iljakuklic:


common/src/chain/config.rs
```diff
@@ -442,34 +431,7 @@ pub fn create_unit_test_config() -> ChainConfig {
...
-        (10, "0"),
-    ];
-
-    let emission_schedule = emission_schedule_from_yearly(
-        &subsidy_per_block_in_year_n_str,
-        &target_block_spacing,
-        coin_decimals,
-        Some(expected_total_block_subsidy),
-    );
+    let (target_block_spacing, coin_decimals, emission_schedule) = mainnet_emission_schedule();
```
TheQuantumPhysicist, line 434: This needs to be overhauled in a different way. There has to be a module for the emission schedule and all its details (I mentioned that in a comment somewhere in this file). I don't mind making this change now, but this is just temporary.

TheQuantumPhysicist:
If you'd like to clean this up, there are like 4 different instantiation methods of nets. We should probably have less. One for mainnet, which is constant and takes only as parameters things that we'll parametrize in command line arguments, and a builder for other testing stuff.

TheQuantumPhysicist:
Btw, the unwrap here I didn't care about because this is the first thing to be executed in the program and it can be tracked easily. I was hoping to make the code as concise as possible. I don't mind moving everything to expect, though.

iljakuklic:
> Btw, the unwrap here I didn't care about because this is the first thing to be executed in the program and it can be tracked easily. I was hoping to make the code as concise as possible. I don't mind moving everything to expect, though.

Was getting annoying clippy warnings so wanted to address this in some way. I don't mind to just selectively disable the unwrap linter in some of these cases, using `#[allow()]`. I was going by the heuristic of `unwrap` being in testing code only

iljakuklic:
> > Btw, the unwrap here I didn't care about because this is the first thing to be executed in the program and it can be tracked easily. I was hoping to make the code as concise as possible. I don't mind moving everything to expect, though.
> 
> Was getting annoying clippy warnings so wanted to address this in some way. I don't mind to just selectively disable the unwrap linter in some of these cases, using `#[allow()]`. I was going by the heuristic of `unwrap` being in testing code only

Moved some of them back to `unwrap` with `allow` attribute and explanation. Code was getting a bit convoluted with the expect strings there.

iljakuklic:
> If you'd like to clean this up, there are like 4 different instantiation methods of nets. We should probably have less. One for mainnet, which is constant and takes only as parameters things that we'll parametrize in command line arguments, and a builder for other testing stuff.

I can have a look into this. As a separate PR preferably

stanislav-tkach (approved):

PR #242, opened 2022-06-17
Replace iter_tool dependency with itertools

stanislav-tkach:
The `iter_tools` library simply reexports `itertools` and we already have the second one, so it makes sense to remove the first one.

### DIFF
--- a/utxo/Cargo.toml
+++ b/utxo/Cargo.toml
@@ -13,5 +13,5 @@ logging = { path = "../logging/" }
 serialization = { path = "../serialization" }
 
 parity-scale-codec = { version = "3.1.2", features = ["chain-error"] }
-iter_tools = "0.1.0"
+itertools = "0.10"
 thiserror = "1.0.30"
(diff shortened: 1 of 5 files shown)


muursh (approved):

utxo/Cargo.toml
```diff
@@ -13,5 +13,5 @@ logging = { path = "../logging/" }
 serialization = { path = "../serialization" }
 
 parity-scale-codec = { version = "3.1.2", features = ["chain-error"] }
-iter_tools = "0.1.0"
+itertools = "0.10"
```
muursh, line 16: why `0.10` over the newest version?

iljakuklic: The `0.10` version is anything `0.10.0 <= x < 0.11`, so this will match the newest version `0.10.3`, which is nailed down in `Cargo.lock`. Some [docs here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio).

muursh: You know what I completely misread that. Ignore me entirely

iljakuklic (approved):

TheQuantumPhysicist (approved):

utxo/Cargo.toml
muursh: > Without a patch version specified the dependency resolver will take the latest one (or a version compatible to the specified for other crates). We already have the Cargo.lock file committed in the repository, so one less reason to stick to full versions.

Yeah I agree. Just ignore me. I completely misread it in the first place

PR #244, opened 2022-06-19
Redesign libp2p backend

altonen:
This PR refactors the previous `Backend` construct into event-specific libp2p handlers. It doesn't add new features but changes how the events are handled and makes the libp2p more extensible for future extensions.

Next steps for libp2p are to implement proper connection manager (now the management is scattered around the backend code), better abstraction for requests and responses and possibly a better discovery behavior abstraction. These will be implemented in future PRs and they are not high-priority tasks.


p2p/src/net/libp2p/behaviour.rs
```diff
@@ -0,0 +1,481 @@
...
+    task::{Context, Poll, Waker},
+};
+
+#[derive(NetworkBehaviour)]
+#[behaviour(
+    out_event = "Libp2pBehaviourEvent",
+    event_process = true,
+    poll_method = "poll"
+)]
+pub struct Libp2pBehaviour {
```
TheQuantumPhysicist, line 60: Could you please document this a little bit, explaining what "behavior" really means with some examples?

p2p/src/net/libp2p/constants.rs
```diff
@@ -0,0 +1,38 @@
...
+//! Libp2p constants
+//!
+//! See protocol specification for more details
+
+use std::time::Duration;
+
+/// Gossipsub configuration
+// TODO: config or spec?
+pub const GOSSIPSUB_HEARTBEAT: Duration = Duration::from_secs(10);
+pub const GOSSIPSUB_MAX_TRANSMIT_SIZE: usize = 2 * 1024 * 1024;
```
TheQuantumPhysicist, line 27: I think this should be determined in a more consistent way to be compatible with block size. I don't know how yet though.

p2p/src/net/libp2p/behaviour.rs
TheQuantumPhysicist: This seems to be a part of how libp2p works. So, never mind.

p2p/src/net/libp2p/backend.rs
```diff
@@ -62,57 +49,163 @@ pub struct Backend {
...
+    pub async fn on_connection_established(
+        &mut self,
+        peer_id: PeerId,
+        endpoint: ConnectedPoint,
+    ) -> crate::Result<()> {
+        match endpoint {
+            ConnectedPoint::Dialer { .. } => {
+                log::trace!("connection established (dialer), peer id {:?}", peer_id);
+
+                match self.swarm.behaviour_mut().pending_conns.remove(&peer_id) {
```
TheQuantumPhysicist, line 80: This kind of code style is almost C programming. Everything is mutable and we have a soup of mutations everywhere. Can this be done in such a way where behavior is not mutated by other objects without proper state verification? Can you please rethink the design a little here so that we don't have to do things like this?

Is a method "remove_pending_conns" gonna make me happier? Maybe a little. Depends on whether you just call remove in that function; that won't solve this problem. But basically I'd like a system where all invariants are held during the execution of the program, more than just mutating continuously things like C code does.

p2p/src/net/libp2p/backend.rs
```diff
@@ -212,7 +256,7 @@ impl Backend {
                 match self.swarm.disconnect_peer_id(peer_id) {
                     Ok(_) => {
                         log::trace!("peer {:?} disconnected", peer_id);
-                        self.established_conns.remove(&peer_id);
+                        self.swarm.behaviour_mut().established_conns.remove(&peer_id);
```
TheQuantumPhysicist, line 259: here too...

I think this fiasco was started by separating objects but not separating behavior.

TheQuantumPhysicist (commented):
I think this PR needs more discussion as I need to understand the full plan, if any. Separating objects without separating behavior can lead to big problems in maintaining a valid state of the program during execution.

p2p/src/net/libp2p/backend.rs
altonen: Yeah this is very much work in progress. The idea is to implement `ConnectionManager` which deals with low-level connectivity events. It implements the `NetworkBehaviour` trait so it perfectly integrates with the changes introduced in this PR. That's the goal but implementing it in this PR would've made this PR even larger than it already is and it would've introduced a lot of new changes which I wanted to avoid in this kind of "moving stuff around" PR.

So in closing, this is bad design and there absolutely is a plan to fix this which integrates with the libp2p `Swarm`/`NetworkBehaviour` design much more nicely. I just haven't started to work on that yet.

TheQuantumPhysicist: Alrighty. I'll approve based on that.

altonen:
As you can see only the connectivity-events are mutating the state outside the handlers because fixing that would require introducing a lot of new code and make this PR even harder to review.

I've outlined the plan in one comment already but to recap, the `behaviour_mut().pending_conns` will be removed in favor of a connection manager that handles the incoming connections, tracks their handshake state, and then finally reports to the front-end. The quite significant refactor relates to the discovery mechanism but they're not relevant here.

TheQuantumPhysicist (approved):

p2p/src/net/libp2p/constants.rs
altonen: We could have a global constants where this and the block size could both be. Making this easily configurable is IMO asking for trouble because you'll be banned immediately if you send too a message that's too large. Can this be resolved and/or should we move this discussion to a Github issue?

TheQuantumPhysicist: Yes, please. Create an issue for this.

altonen: https://github.com/mintlayer/mintlayer-core/issues/250

PR #246, opened 2022-06-21
Refactor emission schedule

iljakuklic:
Part 1 of chain config refactoring, mostly to do with emission schedule

* Move stuff around a bit and separate into modules
* Introduce `Mlt` type with separate constructors to specify amount of MLTs or atom units
  * Also eliminates string conversions at various points
* Different emission schedule representation
  * It is represented as a function from block height to total supply
  * Block reward is calculated as difference between the previous supply and current supply

The part 2 (in a separate PR) will be the introduction of builder pattern for constructing `ChainConfig`.


iljakuklic:
This does not change behaviour. I checked by leaving the old implementation around temporarily and exhaustively checking block rewards for the whole reward emission period.

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,336 @@
...
+        Self(f.into())
+    }
+
+    /// Construct an emission schedule from yearly block award table
+    ///
+    /// This constructor takes:
+    ///  * The initial supply (premine amount)
+    ///  * Yearly per-block rewards, as a vector.
+    ///    In years outside of the range of the vector, the reward is assumed to be zero.
+    ///  * The target block interval
```
stanislav-tkach, line 99: Should there be a mention of the precondition that an interval cannot contain sub-seconds?

stanislav-tkach (approved):

common/src/chain/config/emission_schedule.rs
iljakuklic: Good point. Added.

altonen (approved):

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,336 @@
...
+        Self(Amount::from_atoms(n as u128 * Mlt::ATOMS_PER_MLT))
+    }
+
+    /// Convert the number of atoms to Amount
+    pub const fn to_amount_atoms(self) -> Amount {
+        self.0
+    }
+
+    /// Multiplication, saturated at [Mlt::MAX]
+    pub fn saturating_mul(self, rhs: u128) -> Self {
```
TheQuantumPhysicist, line 38: I'm sorry, but this is not allowed. We're not allowed to use any saturating operations because it could lead to security vulnerabilities and it's a slippery slope. As boring and explicit as it sounds, every mathematical operation should return an Option or Result and be evaluated manually.

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// Maximum representable amount of MLTs
+    pub const MAX: Self = Self(Amount::MAX);
+
+    /// Construct from the number atomic units
+    pub const fn from_atoms(n: u128) -> Self {
+        Self(Amount::from_atoms(n))
+    }
+
+    /// Construct from the number of MLTs
+    pub const fn from_mlt(n: u64) -> Self {
```
TheQuantumPhysicist, line 25: While I understand why you're using u64 here, but it should basically be the int type of Amount (it's public, btw). Then the result should be an Option or Result.

Related to that, I understand that you're using const fn, but this is wrong as it breaks the requirements. Remember, when you brought this up the first time, I mentioned the issue with moving from strings. Even though the Mlt type is a good idea, there has to be a way to read this from strings (or string -> u64 -> Amount, whichever you prefer), because we will want to establish a schedule in the functional tests as a command line argument. If you're writing this as a generic MLT struct, then there has to be a way to initialize MLT at runtime with MLT amount. This function seems appropriate for that. It's a nice touch to have this work with static assert, but this has to always work at runtime as well, despite it not having a specific use right now.

To summarize: Either add another function for runtime, or make this one do that. I think this being a const fn is unnecessary. Just do the Amount multiplication and return an Option and expect/unwrap on usage.

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,336 @@
...
+/// Internal emission schedule representation
+pub type EmissionScheduleFn = dyn Fn(BlockHeight) -> Mlt + Sync + Send + 'static;
+
+/// Emission schedule, characterized by function from block height to total supply at that point.
+///
+/// The function has to be a monotonic non-decreasing function. The function also has to
+/// eventually flatten out to a constant function if [Self::final_supply] is required to make any
+/// sense. The point at which the function becomes constant is where the block rewards run out.
+#[derive(Clone)]
+pub struct EmissionSchedule(std::sync::Arc<EmissionScheduleFn>);
```
TheQuantumPhysicist, line 78: If you want to keep this as a function, that's OK. Just please present a way to read this from a string so that it can be passed using functional tests.

TheQuantumPhysicist: I think the easy solution to this is to just make another caller for `from_yearly_table` that accepts somekind of string input and parses it.

TheQuantumPhysicist (commented):
Good work. This is what I wanted to do eventually. Just little touches here and there and this would be complete for our future plans. Mainly the ability to determine the schedule in functional tests and pass the table as a string.

common/src/chain/config/emission_schedule.rs
iljakuklic: Regarding `const`-ness here, `const fn` works at runtime too. Annotating the function `const` just places some extra restrictions on it so it can be used in `const` contexts (like initializing some constants below) but executing it at runtime is still possible.

common/src/chain/config/emission_schedule.rs
iljakuklic: Ok, will add a string representation tomorrow.

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,336 @@
...
+
+impl std::fmt::Debug for EmissionSchedule {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "EmissionSchedule(<function>)")
+    }
+}
+
+// Emission schedule for mainnet
+
+pub const DEFAULT_TARGET_BLOCK_SPACING: Duration = Duration::from_secs(120);
```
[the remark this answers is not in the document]
iljakuklic: Agreed. Would one level up, in `common::chain::config` be a more appropriate place? Any other suggestions?

common/src/chain/config/emission_schedule.rs
iljakuklic: The static assert only refers to constants and types, not to the input value. It could well be outside of this function but I decided to put it inside because it is related to the argument made in the comment. It does not hinder the ability to run the function at runtime in any way.

common/src/chain/config/emission_schedule.rs
iljakuklic: Thinking about this once again, there is a slight inconvenience in doing it at the level of `EmissionSchedule`. It is because `from_yearly_table` also takes the time interval between blocks as an argument, which is used also in other places. That means the user would have to specify the block interval in multiple places if the emission schedule was supposed to be constructed entirely from one string.

I think it would be an improvement to specify the reward drop period in number of blocks rather than making it yearly. I quite like this one since it does not introduce the notion of wall clock time into emission schedule, which otherwise works exclusively on blocks. Also allows user to specify different reward period, not just yearly. Conversion from year to number of blocks would be done elsewhere (or using a helper method here).

All in all the `from_yearly_table` (or equivalent) method requires three pieces of information:

1. The initial MLT supply
2. Per-block reward in each reward period (by default number of blocks corresponding to 1 year)
3. The reward drop period in the number of blocks (either passed in directly or calculated from block interval and period time)

When parsing these from the command line, I think these should be separate command line arguments. The reward table (2) could be a comma-separated list of decimal numbers. However due to way the drop period is linked with the block interval, I think it may best be done outside of this module.

My proposal is:
1. Add some tools to support table parsing to this PR:
  a. Change the reward block period to block count
  b. Add parsing of the per-period reward table from string (but not the whole emission schedule)
2. Integrate these things together in the subsequent PR (ChainConfig builder refactor):
  a. Can calculate number of blocks per year because ChainConfig has access to block interval
  b. Let user override the reward drop period (as block count or real time)
  c. Let user override the initial supply
3. In yet another PR
  a. Add the ability to build ChainConfig from the command line by exposing most of its methods

Cmments?

iljakuklic: > I'm not super-bullish on the `from_yearly_table`. It's only valuable because in Mintlayer we have yearly schedule for the coins on mainnet. When we use this function we're, by definition, not gonna use it for anything else that doesn't conform with the yearly schedule. The format I prefer is (as you saw before) a list of pair of reward/block-height in the form of "this is the reward from now on".

Ok. I think the representation deserves some thought so we can enforce as many of the invariants as possible by construction. That is monotonocity, totality and finiteness of supply. I think it is possible with minor modifications.

> There's no simpler format for this without making assumptions (the drop from initial reward works too, but it's O(N) in finding the current value and I feel it just complicates things).

If you mean starting with some initial reward and going down with the reward by a constant amount each block, that becomes a quadratic formula for total supply at given block height. That would be a pretty elegant supply schedule IMHO. I have a feeling that it's not what you had in mind though.

> The reason why I oppose the idea of making a function for the yearly assumption is that the tests will kind of become useless. I want everything to go through a single path and be tested from there by only varying the block height vs reward.
> 
> Right now I feel like we're dancing around the `from_yearly_table` function or something with a fixed period... I don't think we need to do this.

Ok will come up with something more general. There is the function representation that is about as general as it gets but that is not easily passed from command line as a string (unless we want to embed some sort of interpreter).

> I'm not opposed to the refactoring you did in general. But the yearly/periodic thing looks like a bad idea. Besides of the generic format I suggested, any function that will calculate the schedule and validate it should be generic enough to work for any kind of schedule, whether it's the one from mainnet or an arbitrary one coming from functional tests.

Well you could still set the period to one block and provide the reward for each block as a huge table explicitly but that's not very elegant.

> Basically, the required condition for this PR to look complete is just to be able to provide the block rewards arbitrarily as a string (whether that string function exists or not). I'm afraid with this PR, as is, we're cutting that off and accepting assumptions that aren't always true.

iljakuklic:
Added an extra type that represents piecewise-linear emission schedule as a table with the option to load from / write to a string. There are some tests but will expand on them slightly tomorrow.

iljakuklic:
Added some extra tests, docs and the string format examples. Please re-review if interested

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,611 @@
...
+        Self {
+            initial_supply,
+            initial_reward,
+            periods,
+        }
+    }
+
+    /// All reward periods, starting at block height 0
+    pub fn reward_periods(&self) -> impl Iterator<Item = (BlockHeight, Mlt)> + '_ {
+        std::iter::once((BlockHeight::zero(), self.initial_reward))
```
TheQuantumPhysicist, line 164: Yep... you're manually chaining the initial reward. I don't really see the benefit in complicating the process. Just make element zero the value.

I know you may be doing this to enforce a compile-time check for this, but then there's this trade-off between more code/complexity vs a runtime check. I don't know which is worse.

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,611 @@
...
+    ///  * Reward periods. These are mappings from block heights to per-block rewards from that
+    ///    point onwards, until the next period. The periods are implicitly sorted by BTreeMap.
+    pub fn new(
+        initial_supply: Mlt,
+        initial_reward: Mlt,
+        periods: BTreeMap<BlockHeight, Mlt>,
+    ) -> Self {
+        Self {
+            initial_supply,
+            initial_reward,
```
TheQuantumPhysicist, line 157: We try to distinguish between reward and subsidies by saying that rewards are subsidy + transaction fees.

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,611 @@
...
+            self.initial_supply,
+            self.initial_reward,
+            BlockHeight::zero(),
+        );
+
+        // Pre-calculate a table with starting supply and per-block rewards for each reward period.
+        let table: BTreeMap<BlockHeight, (Mlt, Mlt)> = self
+            .periods
+            .iter()
+            .scan(
```
TheQuantumPhysicist, line 197: Well, I'm allergic to scan because I'm allergic to mutable... I would've used rfold for this, but sure. If this works...

common/Cargo.toml
```diff
@@ -11,6 +11,7 @@ script = { path = '../script'}
 crypto = { path = '../crypto'}
 serialization = { path = "../serialization" }
 logging = { path = "../logging/" }
+utils = { path = "../utils/" }
```
TheQuantumPhysicist, line 14: I have a feeling this isn't used anywhere?

TheQuantumPhysicist (approved):
A few minor comments were made. Please make sure to label everything here a subsidy instead of reward. Let's try to be consistent about this.

common/src/chain/config/emission_schedule.rs
```diff
@@ -0,0 +1,611 @@
...
+        Self(std::sync::Arc::new(f))
+    }
+
+    /// Construct an emission schedule from a function. See also [Self::from_fn]
+    pub fn from_dyn_fn(f: Box<EmissionScheduleFn>) -> Self {
+        Self(f.into())
+    }
+
+    /// Get supply at given block height
+    pub fn supply_at(&self, ht: BlockHeight) -> Mlt {
```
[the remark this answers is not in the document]
iljakuklic: Well the total supply is a rather elusive concept. There are also lost keys and coins burned by sending to a random pubkey, which are not detectable. Total amount emitted up to point X is at least a simple well-defined concept. Will try to rename this to something more accurate. Maybe we should have equivalents of M1, M2, etc :D.

common/src/chain/config/emission_schedule.rs
iljakuklic: Basically removes a bunch of `expect` and extra invariants placed on the map. The invariant here is not self-contained by this module, so it's basically to ensure that the user of the module cannot violate it by construction.

common/src/chain/config/emission_schedule.rs
iljakuklic: Good point, will change that. Thanks for pointing this out

common/Cargo.toml
iljakuklic: Well spotted. I used `ensure` somewhere, removed it later but forgot about the dep.

common/src/chain/config/emission_schedule.rs
iljakuklic: Well conceptually scan is the right operation to use here. I took a brief look into `itertools` to see whether there is a version of `scan` that lets you return the updated values rather than mutating them (which I would prefer too) but unfortunately did not find anything. Anyway all mutation is contained in the closure body in the next 5 lines of code so hopefully this is not so bad.

PR #248, opened 2022-06-22
Add symmetric encryption support

TheQuantumPhysicist:


crypto/src/symkey/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+        assert_eq!(decrypted, message);
+    }
+
+    #[test]
+    fn select_text_external() {
+        let test_vectors = read_test_vectors_file(
+            std::path::Path::new("src")
+                .join("symkey")
+                .join("chacha20poly1305")
+                .join("XCHACHA20_POLY1305_TEST_VECTORS_QIG_CRL_2019_10_17.tv"),
```
muursh, line 301: I'm tempted to say we rename this to something take makes more sense for us. The file name won't mean a lot to most people, we could probably strip the header in the file too since it just refers to a bunch of people that nobody knows

crypto/src/symkey/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+use chacha20poly1305::{XChaCha20Poly1305, XNonce};
+use parity_scale_codec::{Decode, Encode};
+
+pub const NONCE_LEN: usize = 24;
+pub const KEY_LEN: usize = 32;
+
+#[must_use]
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Chacha20poly1305Key {
+    key_data: chacha20poly1305::Key,
```
muursh, line 13: One thing here is the key isn't zeroed on drop. As in, in the lib they don't do it

muursh: https://docs.rs/aead/0.4.3/src/aead/lib.rs.html#79

crypto/src/symkey/chacha20poly1305/mod.rs
muursh: Yeah he more I think about it cleaning it up is probably worth the 10 mins it will take

muursh (commented):
All in all no real issues from my side

altonen:
> > The `'net::libp2p::proto::gossipsub::tests::test_unsubscribed` test is failed on macos. Is it OK?..
> 
> It's flakey tests. @altonen will be looking into them soon enough.

yeah sorry about that, the test wasn't written in the most reliable way and in all honesty, didn't make much sense either.

crypto/src/symkey/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,308 @@
+use crate::random::{CryptoRng, Rng};
+use crate::symkey::Error;
+use chacha20poly1305::aead::{AeadInPlace, NewAead};
+use chacha20poly1305::{XChaCha20Poly1305, XNonce};
+use parity_scale_codec::{Decode, Encode};
```
altonen, line 5: `serialization` crate

crypto/src/symkey/chacha20poly1305/mod.rs
altonen: I have some questions. What do you guys think about having `Clone + Debug` implemented for the private key? Are there any security concerns? I don't know if `Clone` can be harmful in any way but maybe malicious actor can somehow trigger the key to be printed to log at which point zeroing the memory is not really as effective.

This maybe useful for zeroing memory on drop: https://docs.rs/zeroize/latest/zeroize/#custom-derive-support

crypto/src/symkey/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+            .map_err(|e| Error::DecryptionError(e.to_string()))?;
+        Ok(message)
+    }
+
+    pub fn decrypt<T: AsRef<[u8]>>(
+        &self,
+        cipher_text_in: T,
+        associated_data: Option<&[u8]>,
+    ) -> Result<Vec<u8>, Error> {
+        if cipher_text_in.as_ref().len() < NONCE_LEN {
```
altonen, line 77: `ensure!()`

crypto/src/symkey/chacha20poly1305/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+        KEY_LEN
+    }
+}
+
+impl Decode for Chacha20poly1305Key {
+    fn encoded_fixed_size() -> Option<usize> {
+        Some(KEY_LEN)
+    }
+
+    fn decode<I: parity_scale_codec::Input>(
```
altonen, line 108: `serialization` crate?

crypto/src/symkey/mod.rs
```diff
@@ -0,0 +1,124 @@
+use crate::random::{CryptoRng, Rng};
+
+mod chacha20poly1305;
+
+use parity_scale_codec::{Decode, Encode};
```
altonen, line 5: `serialization` crate

crypto/src/symkey/chacha20poly1305/mod.rs
TheQuantumPhysicist: I know these limitations, but there's almost nothing we can do currently as this is a whole campaign on its own. The right thing to do is to add mlock support, but it's POSIX specific and I found one esoteric crate that uses it and I'm not sure how it works on Windows. Windows has its its own VirtualLock... so imagine the nightmare of implementing platform specific things. 

On one hand, Copy can be an issue, but clone is necessary for the struct to be useful; after all it's just an array, and the wallet may require to clone the key. The problem with copy, AFAIU, is that it triggers intermediate objects all over the place. Clone is strict and will only create a copy when asked to. The clones will have the same characteristics. 

Now about Debug, maybe we can remove it. But we have to keep something in mind: If a computer is compromised by a "malicious actor" to the point of controlling the program, then everything is already compromised. The actor can already take a copy of the virtual memory of the program and look into it. In other words, we want to be careful of overdoing things.

TheQuantumPhysicist: @muursh I think you quoted an older version of the crate. The current version does zeroize:

https://github.com/RustCrypto/AEADs/blob/d93b4e098cd7dfaa72518286060ad0d6ab94f346/chacha20poly1305/src/lib.rs#L311

TheQuantumPhysicist: Unresolve the comment if I'm wrong

muursh: So it does. Not sure how I managed that

muursh (approved):

PR #249, opened 2022-06-23
sync: Refactor request/response handling

altonen:
Separate functions that send requests/response from the main syncing
code, reduce code duplication and document the existing code more.

This PR doesn't add any new functionality but just cleans up the previous implementation. Next steps (in different PRs) are to improve error handling and change how the sync state is detected (detect when initial block download is done).

### DIFF
--- a/p2p/tests/sync.rs
+++ b/p2p/tests/sync.rs
@@ -193,8 +193,49 @@ where
     T: NetworkingService,
     T::SyncingCodecHandle: SyncingCodecService<T>,
 {
-    let event = mgr.handle_mut().poll_next().await.unwrap();
-    mgr.on_syncing_event(event).await.unwrap();
+    match mgr.handle_mut().poll_next().await.unwrap() {
+        net::types::SyncingEvent::Request {
+            peer_id,
+            request_id,
+            request:
+                Message {
+                    msg: MessageType::Syncing(SyncingMessage::Request(message)),
+                    magic: _,
+                },
+        } => match message {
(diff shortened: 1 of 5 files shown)


p2p/tests/sync.rs
```diff
@@ -193,8 +193,49 @@ where
     T: NetworkingService,
     T::SyncingCodecHandle: SyncingCodecService<T>,
 {
-    let event = mgr.handle_mut().poll_next().await.unwrap();
-    mgr.on_syncing_event(event).await.unwrap();
+    match mgr.handle_mut().poll_next().await.unwrap() {
```
TheQuantumPhysicist, line 196: Can we replace that unwrap() with expect? Or maybe some better error handling?

p2p/tests/sync.rs
```diff
@@ -193,8 +193,49 @@ where
...
+            }
+        },
+        net::types::SyncingEvent::Error {
+            peer_id,
+            request_id,
+            error,
+        } => {
+            mgr.process_error(peer_id, request_id, error).await?;
+        }
+        _ => {}
```
TheQuantumPhysicist, line 236: Is this necessary? I thought the arms are exhaustive.

TheQuantumPhysicist (approved):

p2p/tests/sync.rs
altonen: It was not, the invalid messages are ignored for now. If I write more comprehensive tests in `tests/sync.rs` I'll match against those too but I'll leave this as is for now.

PR #252, opened 2022-06-23
Block proof calculation is now with the correct units and is done properly from the ConsensusData (at least for PoW)

TheQuantumPhysicist:


iljakuklic (approved):
Nice to see these `unwrap`s go.

common/src/chain/block/consensus_data.rs
```diff
@@ -105,4 +114,15 @@ impl PoWData {
     pub fn update_nonce(&mut self, nonce: u128) {
         self.nonce = nonce;
     }
+
+    pub fn get_block_proof(&self) -> Option<Uint256> {
```
altonen, line 118: should these be without the `get_` prefix?

common/src/uint/impls.rs
```diff
@@ -505,6 +521,30 @@ macro_rules! construct_uint {
 construct_uint!(Uint256, 4);
 construct_uint!(Uint128, 2);
 
+impl Encode for Uint256 {
+    fn size_hint(&self) -> usize {
+        32
+    }
+
+    fn encode_to<T: parity_scale_codec::Output + ?Sized>(&self, dest: &mut T) {
```
altonen, line 529: `serialization`?

common/src/uint/impls.rs
```diff
@@ -505,6 +521,30 @@ macro_rules! construct_uint {
...
+        v.encode_to(dest)
+    }
+
+    fn encoded_size(&self) -> usize {
+        32
+    }
+}
+
+impl Decode for Uint256 {
+    fn decode<I: parity_scale_codec::Input>(
```
altonen, line 540: same comment here

altonen (approved):

common/src/chain/block/consensus_data.rs
TheQuantumPhysicist: get_ is OK if there's a calculation involved underneath.

PR #253, opened 2022-06-23
Minor fix for filter blocks

TheQuantumPhysicist:
Is this worth it? What do you guys think?

Unfortuantely, I couldn't find a try_skip_while function.


altonen (approved):
I like it, thanks for fixing the TODO

iljakuklic (approved):
I think it's an improvement over what was there before

muursh (approved):
I like it

PR #254, opened 2022-06-23
Add previous block skipper for searching for ancestors faster

TheQuantumPhysicist:
Use the bitcoin's way to search for ancestors faster. The get_skip_height function is taken literally from bitcoin, and also the get_ancestor as well. It's not idiomatic for that reason.


chainstate/src/detail/chainstateref.rs
```diff
@@ -682,6 +703,16 @@ impl<'a, S: BlockchainStorageWrite, O: OrphanBlocksMut> ChainstateRef<'a, S, O>
             prev_block_index.block_height().next_height()
         });
 
+        let some_ancestor = match prev_block_index {
+            Some(ref prev_bi) => Some(
+                self.get_ancestor(prev_bi, get_skip_height(height))
+                    .map_err(|_| BlockError::AncestorRetrievalFailed(block.get_id()))?
+                    .block_id()
+                    .clone(),
```
altonen, line 711: Could you explain why is the peer banned if ancestor retrieval fails? I'm not sure if I fully understand this change because to me it seems we're calculating the height based on the data we have locally available (on line 702) but when the skip height is calculated and that ancestor is not found, it's a peer error. It sounds more like a database inconsistency to me so what am I missing?

TheQuantumPhysicist: Yep... it's a database consistency or an implementation issue. You're right. But this is, again, the issue of fearing having done something wrong and causing an attack vector.

TheQuantumPhysicist: After rethinking this, I changed it to an expect and added more aggressive tests to ensure all corner cases are covered. Let's hope I missed nothing.

stanislav-tkach (approved):

chainstate/src/detail/chainstateref.rs
```diff
@@ -136,25 +137,54 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                ancestor_height: target_height,
             });
         }
 
+        let step_to_prev_block = |block_index_walk: &mut BlockIndex,
+                                  height_walk: &mut BlockHeight|
+         -> Result<(), PropertyQueryError> {
+            *block_index_walk = self.get_previous_block_index(block_index_walk)?;
+            *height_walk = (*height_walk - BlockDistance::from(1))
+                .expect("height_walk is greater than height");
```
iljakuklic, line 157: I added `BlockHeight::prev_height` but it's a part of a live PR #246, not merged yet. Slightly more convenient and works in more cases.

common/src/chain/block/height_skip.rs
```diff
@@ -0,0 +1,1032 @@
+use crate::primitives::BlockHeight;
+
+fn invert_lowest_one(n: u64) -> u64 {
+    n & n.wrapping_sub(1)
+}
+
+pub fn get_skip_height(h: BlockHeight) -> BlockHeight {
```
iljakuklic, line 7: Would this be more ergonomic/idiomatic as a method of `BlockHeight`?

common/src/chain/block/block_index.rs
```diff
@@ -12,6 +12,7 @@ use super::timestamp::BlockTimestamp;
 pub struct BlockIndex {
     block_id: Id<Block>,
     block_header: BlockHeader,
+    skip: Option<Id<Block>>,
```
iljakuklic, line 15: When is this `None`? Is it only in genesis or are there other cases?

iljakuklic (approved):
Looks good, just added a couple of minor comments.

common/src/chain/block/height_skip.rs
TheQuantumPhysicist: I don't prefer that. It's unrealted to the block height concept.

common/src/chain/block/block_index.rs
TheQuantumPhysicist: Just genesis

common/src/chain/block/height_skip.rs
iljakuklic: Fair enough

PR #255, opened 2022-06-24
Make opt-level=2 only in script crate tests and get rid of recursion to disconnect blocks

TheQuantumPhysicist:


Cargo.toml
```diff
@@ -65,8 +65,11 @@ utxo = { path = "utxo"}
 [profile.release]
 debug-assertions = true
 overflow-checks = true
+lto = true
```
stanislav-tkach, line 68: Should we also use `codegen-units = 1`?

stanislav-tkach (approved):

Cargo.toml
stanislav-tkach: https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units

Cargo.toml
```diff
@@ -65,8 +65,11 @@ utxo = { path = "utxo"}
...
 
 [profile.test]
-opt-level = 2
 debug-assertions = true
 overflow-checks = true
+
+[profile.test.package.script]
+opt-level = 2
```
iljakuklic, line 75: Does this inherit the other properties of the "global" test profile?

iljakuklic (approved):
LGTM

Cargo.toml
TheQuantumPhysicist: I think so. I only tested that script now is release.

Cargo.toml
TheQuantumPhysicist: Yeah, I'm not sure... builds may become release slow...

Cargo.toml
iljakuklic: Ok docs seem to suggest so, although it's not super explicit. Let's wait and see how long the CI takes.

Cargo.toml
stanislav-tkach: OK, I didn't see that we're building everything in the release mode on CI.

muursh (approved):

stanislav-tkach:
I found it really strange that enabling LTO breaks our code.

TheQuantumPhysicist:
Related: https://github.com/mintlayer/mintlayer-core/issues/258

PR #256, opened 2022-06-24
Refactor `PubSubMessageHandler`

altonen:
Remove code duplication, add documentation, improve error handling, and subscribe to pubsub topics only after the initial block download.

Let me know your opinions on what are the appropriate responses to various errors the message handler might receive from chainstate.

### DIFF
--- a/p2p/src/pubsub.rs
+++ b/p2p/src/pubsub.rs
@@ -14,14 +14,27 @@
 // limitations under the License.
 //
 // Author(s): A. Altonen
+
+//! Publish-subscribe message/event handling
+
 use crate::{
-    error::{P2pError, ProtocolError},
+    error::{P2pError, ProtocolError, PublishError},
     event,
     message::{self, Message, MessageType, PubSubMessage},
-    net::{self, NetworkingService, PubSubService},
+    net::{
+        types::{PubSubEvent, ValidationResult},
+        NetworkingService, PubSubService,
+    },
 };
-use chainstate::{chainstate_interface, BlockError, ChainstateError::ProcessBlockError};
-use common::chain::ChainConfig;
+use chainstate::{
+    ban_score::BanScore,
+    chainstate_interface, BlockError,
+    ChainstateError::{FailedToInitializeChainstate, FailedToReadProperty, ProcessBlockError},
+};
+use common::{
+    chain::{block::Block, ChainConfig},
+    primitives::Id,
+};
 use futures::FutureExt;
 use logging::log;
 use std::sync::Arc;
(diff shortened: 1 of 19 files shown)


p2p/src/pubsub.rs
```diff
@@ -149,90 +111,142 @@ where
...
+        block: Block,
+    ) -> crate::Result<()> {
+        let result = self
+            .chainstate_handle
+            .call_mut(move |this| this.process_block(block, chainstate::BlockSource::Peer))
+            .await?;
+
+        let (validation_result, score) = match result {
+            Ok(_) => (ValidationResult::Accept, 0),
+            Err(ProcessBlockError(ref block_error)) => match block_error {
```
TheQuantumPhysicist, line 132: I'm confused. Why don't you directly call ban_score on the BlockError object? Why match all these error cases? You shouldn't care. Just get the ban score and add it to the total. If it's higher than the max, ban the dude. The max should come from command line arguments, btw. That's still not implemented. Currently 100 is fine for default.

p2p/src/pubsub.rs
```diff
@@ -149,90 +111,142 @@ where
...
+    /// Process block announcement from the network
+    async fn process_block_announcement(
+        &mut self,
+        peer_id: T::PeerId,
+        message_id: T::MessageId,
+        block: Block,
+    ) -> crate::Result<()> {
+        let result = self
+            .chainstate_handle
+            .call_mut(move |this| this.process_block(block, chainstate::BlockSource::Peer))
```
TheQuantumPhysicist, line 127: Please call preliminary_block_check() before calling process_block() (separately, not in the same lambda). The reason is that the precheck is immutable, which means it can run in parallel and not block the queue of chainstate.

altonen: I think this would require making a clone of each block, regardless of whether there would be contention with calls to chainstate. Seems awfully wasteful

p2p/src/pubsub.rs
altonen: I need to match `BlockAlreadyExists` either way but I think `ValidationResult` here is more interesting than the ban score. The easiest way to handle is that if we have an error and ban score is 0, the block is ignored, if it's more than 0, the block is rejected and peer is banned. I'm not opposed to switching to that. But some of these errors might be totally our fault in which case we shouldn't ignore the block. I don't know

p2p/src/pubsub.rs
TheQuantumPhysicist: That's fair. Maybe we should use Arc. Another solution is to ping-pong Block back and forth. So it can be returned in the return value of preliminary_block_check().

p2p/src/pubsub.rs
TheQuantumPhysicist: Let's revisit this issue later when the implementation is complete to see how best to tackle this. At best, I'd suggest separating the score calculation from the match. It seems too convoluted while you can simply just call a method on the error to get that number.

p2p/src/pubsub.rs
altonen: I like the latter idea better, I'll use that.

PR #257, opened 2022-06-24
Add rustfmt as its own check in CI

stanislav-tkach:
As far as I can see, we are not using rustfmt on CI, so there is no reason to install it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,24 @@ jobs:
       with:
         arguments: --all-features --workspace
 
+  rustfmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install rust stable
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt
+          default: true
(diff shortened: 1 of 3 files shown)


iljakuklic:
Would it make sense to instead use the `--check` mode of rustfmt to verify formatting as a part of the CI process?

stanislav-tkach:
@iljakuklic Highly agree! I have updated the pull request.

iljakuklic:
Should this be a hard error that blocks the PR? Maybe just some sort of "warning" is good enough. Not sure how to make github CI understand the distinction.

muursh:
@iljakuklic by default it will be a "warning" rather than a blocker. We have requirement rules for the 3 OS builds but that's it. As long as the code builds everywhere in theory anything else can be ignored.

muursh:
@stanislav-tkach renamed the PR to match what it actually does now

muursh (approved):
I'm happy with this

TheQuantumPhysicist (approved):
While I'm OK with this, keep in mind that clippy auto fixing ruins formatting. It'll be a little annoying.

muursh:
We can always remove it if it gets too annoying

iljakuklic (approved):

PR #259, opened 2022-06-27
Refactor error handling in `sync/mod.rs`

altonen:


p2p/src/sync/peer.rs
```diff
@@ -47,33 +47,33 @@ where
...
         Self {
-            peer_id,
-            locator,
+            _peer_id,
             state: PeerSyncState::Unknown,
             work: VecDeque::new(),
         }
     }
 
+    pub fn new_with_locator(_peer_id: T::PeerId, locator: Vec<BlockHeader>) -> Self {
```
TheQuantumPhysicist, line 69: I think there should be a separate type for locators... maybe later.

TheQuantumPhysicist (approved):

p2p/src/sync/peer.rs
altonen: I'll add an entry to my TODO list

PR #262, opened 2022-06-27
Ignore the 'mixed_sighash_types' test

stanislav-tkach:
The `mixed_sighash_types` runs for more than 9 minutes on my (weak) laptop, so marking it as ignored can save a lot of time for everyone, but we still have it on CI.

I am worried that we have a performance regression in some of the latest commits because previously this test took about a minute (which is probably also a little bit too long).

By the way, I can add a lightweight version of the test to run every time if it makes sense.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,6 +57,9 @@ jobs:
       run: cargo build --release
     - name: Run tests
       run: cargo test --release --workspace
+    # This test is ignored, so it needs to run separately.
+    - name: Run mixed_sighash_types test
+      run: cargo test mixed_sighash_types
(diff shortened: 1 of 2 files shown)


TheQuantumPhysicist:
Make it run in CI

stanislav-tkach:
@TheQuantumPhysicist 
> Make it run in CI

I thought that we were already running ignored tests on CI, but now I see that it is only true for the functional tests.

stanislav-tkach:
If someone knows a better way to run all normal tests plus some of the ignored ones please tell me.

TheQuantumPhysicist (approved):

muursh (approved):

PR #264, opened 2022-06-28
p2p: Create new message format

altonen:
Remove global message type and use task-specific types that prevent
publish-subscribe messages to be received by `SyncManager` or header
requests to be received from one of the Gossipsub topics.

The magic numbers of the nodes are exchanged during handshaking, meaning
if the connection is not closed during handshaking, the magic numbers
are going to match for the entire duration of the session.

### DIFF
--- a/p2p/src/message.rs
+++ b/p2p/src/message.rs
@@ -21,48 +21,99 @@ use common::{
 use serialization::{Decode, Encode};
 
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
-pub enum SyncingRequest {
-    #[codec(index = 0)]
-    GetHeaders { locator: Vec<BlockHeader> },
-    #[codec(index = 1)]
-    GetBlocks { block_ids: Vec<Id<Block>> },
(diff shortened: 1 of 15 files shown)


p2p/src/message.rs
```diff
@@ -21,48 +21,43 @@ use common::{
 use serialization::{Decode, Encode};
 
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
-pub enum SyncingRequest {
-    #[codec(index = 0)]
-    GetHeaders { locator: Vec<BlockHeader> },
-    #[codec(index = 1)]
-    GetBlocks { block_ids: Vec<Id<Block>> },
+pub struct HeaderRequest {
```
TheQuantumPhysicist, line 24: Why aren't these types encapsulated with proper factor functions and private field(s)? This looks more like a C struct.

altonen: Fixed

TheQuantumPhysicist (approved):

PR #265, opened 2022-06-28
common: Builder for ChainConfig

iljakuklic:
Use the builder pattern to construct ChainConfig. Improvement suggestions welcome.

### DIFF
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -1,6 +1,8 @@
+mod builder;
 pub mod emission_schedule;
 
-use emission_schedule::{EmissionSchedule, EmissionScheduleTabular, Mlt};
+pub use builder::Builder;
+pub use emission_schedule::{EmissionSchedule, EmissionScheduleTabular, Mlt};
 
 use hex::FromHex;
 
@@ -10,13 +12,11 @@ use crate::chain::block::ConsensusData;
 use crate::chain::signature::inputsig::InputWitness;
 use crate::chain::transaction::Destination;
 use crate::chain::transaction::Transaction;
-use crate::chain::upgrades::ConsensusUpgrade;
 use crate::chain::upgrades::NetUpgrades;
 use crate::chain::{PoWChainConfig, UpgradeVersion};
 use crate::primitives::id::{Id, H256};
 use crate::primitives::Amount;
 use crate::primitives::BlockDistance;
-use crate::primitives::Idable;
 use crate::primitives::{semver::SemVer, BlockHeight};
 use std::collections::BTreeMap;
 use std::time::Duration;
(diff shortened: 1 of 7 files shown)


stanislav-tkach (approved):

common/src/chain/config/mod.rs
```diff
@@ -44,6 +44,260 @@ pub enum ChainType {
...
+            ChainType::Regtest => NetUpgrades::unit_tests(),
+            ChainType::Signet => NetUpgrades::unit_tests(),
+        }
+    }
+}
+
+// Builder support types
+
+#[derive(Clone)]
+enum EmissionScheduleInit {
```
TheQuantumPhysicist, line 103: All these are equivalent. I don't think it's a good idea to make them different arms of an enum that are processed differently, because this would increase the maintenance burden and we'll have different execution paths for different cases. Basically, Mainnet will have an execution path that will only be touched in mainnet... we're killing the benefit of functional tests (even though they're possibly built from the same tool, still). Whether you prefer Table or Fn, pick one, and convert everything to that in ChainConfig, and I honestly prefer Table because reading that as a variable in the debugger is possible.

common/src/chain/config/mod.rs
```diff
@@ -44,6 +44,260 @@ pub enum ChainType {
...
+    Mainnet,
+    Table(emission_schedule::EmissionScheduleTabular),
+    Fn(std::sync::Arc<emission_schedule::EmissionScheduleFn>),
+}
+
+#[derive(Clone)]
+enum GenesisInit {
```
TheQuantumPhysicist, line 110: GenesisBlock, I'd call it. Genesis in general can mean many things.

common/src/chain/config/mod.rs
iljakuklic: It all ends up as the `EmissionSchedule` type, which internally contains `Fn`. These are just different ways to initialise it with different tradeoffs that all get converted into it when you `.build()`. The real motivation here is `EmissionScheduleInit::Mainnet` which is a per-year emission schedule. In order to calculate the number of blocks per year, we need to know the final decision regarding the block interval and that is only available at the point `.build()` is called.

TheQuantumPhysicist: I just realized this is going on only for the builder. I thought the enum is for the ChainConfig. My bad.

On another note, I'm not really happy about EmissionSchedule being an Fn... really... debugging that is going to be really nasty.

common/src/chain/config/mod.rs
TheQuantumPhysicist: To be clear, GenesisBlockInit, if you want to keep the init.

common/src/chain/config/mod.rs
```diff
@@ -44,6 +44,260 @@ pub enum ChainType {
...
+            genesis_block_id: genesis_block.get_id(),
+            genesis_block,
+            height_checkpoint_data: BTreeMap::new(),
+            emission_schedule,
+            net_upgrades,
+        }
+    }
+}
+
+macro_rules! builder_method {
```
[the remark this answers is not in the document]
iljakuklic: There is https://crates.io/crates/derive_builder but not sure it's worth pulling a new dependecy for this.

TheQuantumPhysicist (approved):
Looks good. Though it's probably a good idea to move the builder to its own submodule.

altonen (approved):

common/src/chain/config/mod.rs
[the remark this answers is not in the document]
stanislav-tkach: Makes sense.

PR #266, opened 2022-06-28
Chainstate tests improvements

stanislav-tkach:
Just a little cleanup for now, will hopefully add more test cases soon.


chainstate-types/src/block_index.rs
```diff
@@ -7,7 +7,7 @@ use serialization::{Decode, Encode};
 
 use common::chain::block::timestamp::BlockTimestamp;
 
-#[derive(Debug, Clone, Encode, Decode)]
+#[derive(Debug, Clone, Encode, Decode, Eq, PartialEq)]
```
TheQuantumPhysicist, line 10: Please remove Eq and PartialEq inheritance. Block indexes should be strictly compared by block id, not values. Even implementing Eq manually won't give the correct outcome because blockindex_a == blockindex_b is ambiguous and can mean either that we're checking for equal fields or same block. In case of ambiguity we just delegate the implementation to the dev using it.

chainstate/src/chainstate_interface.rs
```diff
@@ -26,8 +26,22 @@ pub trait ChainstateInterface: Send {
         height: &BlockHeight,
     ) -> Result<Option<Id<Block>>, ChainstateError>;
     fn get_block(&self, block_id: Id<Block>) -> Result<Option<Block>, ChainstateError>;
+
+    /// Returns an exponential sequence of block headers.
```
TheQuantumPhysicist, line 30: I would call it "a list of block hashes whose heights distances increase exponentially starting from the current tip"

chainstate-types/src/block_index.rs
stanislav-tkach: Now I see that `Eq` implementation was missing for a reason... Fixed (reverted). I can use the `maches!` macro in tests anyway.

TheQuantumPhysicist: You can see that all the other tests compare using the block id

stanislav-tkach: Yes, I see that. The issue arises when I try to use `assert_eq!` for `process_block` function (even when an error is expected), because it returns `Result<Option<BlockIndex>, BlockError>`. But it can be worked around by using `matches!` or `unwrap_err`.

TheQuantumPhysicist: `matches!` should always be your last resort. Unwrapping is always better. This is because in case of failure, unwrapping will tell you what the current value is. `matches!` will just tell you a good/bad result.

chainstate/src/detail/tests/syncing_tests.rs
```diff
@@ -14,292 +14,268 @@
 // limitations under the License.
 //
 // Author(s): A. Altonen
```
altonen, line 16: I think you  should add S. Tkach here

altonen (approved):

chainstate/src/detail/tests/double_spend_tests.rs
```diff
@@ -15,231 +15,139 @@
...
+use common::{
+    chain::{
+        block::{timestamp::BlockTimestamp, Block, ConsensusData},
+        OutPointSourceId, Spender, Transaction, TxInput, TxOutput,
+    },
+    primitives::{time, Amount, Id},
+};
+use crypto::random::{self, Rng};
+
+// Process a block where the second transaction uses the first one as input.
```
TheQuantumPhysicist, line 29: Isn't it idiomatic to use /// instead of // for documentation? Or it doesn't matter?

chainstate/src/detail/tests/double_spend_tests.rs
```diff
@@ -15,231 +15,139 @@
...
+// | +-------tx-1--------+ |
+// | |input = prev_block | |
+// | +-------------------+ |
+// |                       |
+// | +-------tx-2--------+ |
+// | |input = tx1        | |
+// | +-------------------+ |
+// +-----------------------+
 #[test]
 fn spend_tx_in_the_same_block() {
```
TheQuantumPhysicist, line 42: spend output from the same block

chainstate/src/detail/tests/double_spend_tests.rs
```diff
@@ -15,231 +15,139 @@
...
+// | +-------------------+ |
+// |                       |
+// | +-------tx-2--------+ |
+// | |input = prev_block | |
+// | +-------------------+ |
+// +-----------------------+
 #[test]
-#[allow(clippy::eq_op)]
-fn double_spend_tx_in_the_same_block() {
+fn spend_tx_in_the_same_block_invalid_order() {
```
TheQuantumPhysicist, line 83: same

chainstate/src/detail/tests/double_spend_tests.rs
```diff
@@ -15,231 +15,139 @@
...
+        let block_id = block.get_id();
+        assert_eq!(
+            chainstate.process_block(block, BlockSource::Local).unwrap_err(),
+            BlockError::CheckBlockFailed(CheckBlockError::CheckTransactionFailed(
+                CheckBlockTransactionsError::DuplicateInputInBlock(block_id)
+            ))
+        );
         assert_eq!(
             chainstate
                 .chainstate_storage
```
TheQuantumPhysicist, line 153: The expect below seem to be in the wrong order, btw.

chainstate/src/detail/tests/double_spend_tests.rs
```diff
@@ -311,31 +198,21 @@ fn double_spend_tx_in_another_block() {
...
-        assert!(chainstate.process_block(second_block, BlockSource::Local).is_err());
+        assert_eq!(
+            chainstate.process_block(second_block, BlockSource::Local).unwrap_err(),
+            BlockError::StateUpdateFailed(StateUpdateError::DoubleSpendAttempt(
+                Spender::RegularInput(first_tx.get_id())
+            ))
+        );
         assert_eq!(
             chainstate
                 .chainstate_storage
```
TheQuantumPhysicist, line 218: Same order issue

TheQuantumPhysicist (approved):
Besides the minor comments, it looks fine. Let Anton also review it as they were his tests originally.

chainstate/src/detail/tests/syncing_tests.rs
stanislav-tkach: Is that mandatory? It seems kind of pointless to me. Should I add myself to every file where I change one line?.. We already have commits history for that. Perhaps it is needed for some legal reasons?

chainstate/src/detail/tests/double_spend_tests.rs
stanislav-tkach: Usually I prefer to use documentation comments even for private functions (a documentation can still be generated for them), but not for tests. I guess I don't have solid reasons for that, so I can change these comments if you like.

TheQuantumPhysicist: Not really... but I was wondering since you went for documentation above the function why it's not done "the rust way"'

sinitcin (approved):
I added a bunch of minor comments, but I still feel that these tests aren't enough to cover all degree of freedom.

chainstate/src/detail/tests/double_spend_tests.rs
stanislav-tkach: I have checked: the documentation isn't generated for tests even with the `--document-private-items` flag. It can be worked around by replacing `#[cfg(test)]` with `#[cfg(any(test, doctest))]`, but I doubt that we need that.

chainstate/src/detail/tests/syncing_tests.rs
altonen: I believe the idea was that if you've made non-trivial changes to the file, you should add your name to the list of authors so people have an idea of who to ask to review the code changes.

PR #268, opened 2022-06-30
Introduce `ConnectionManager`

altonen:
Implement `ConnectionManager` which handles all connectivity-related events of the libp2p backend and reports all relevant connectivity-events (accepted inbound/outbound and closed connections to the `PeerManager` object).


p2p/src/net/libp2p/connectivity/types.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Entry for an established and active connections
+pub struct Connection {
+    /// Active address of the remote peer
+    addr: Multiaddr,
+
+    /// Connection state of the remote peer
+    state: ConnectionState,
+
+    /// `IdentifyInfo` of the remote peer
+    peer_info: Option<identify::IdentifyInfo>,
```
TheQuantumPhysicist, line 141: Why is this optional?

p2p/src/net/libp2p/connectivity/types.rs
```diff
@@ -0,0 +1,209 @@
...
+            && $a.protocol_version == $b.protocol_version
+            && $a.agent_version == $b.agent_version
+            && $a.listen_addrs == $b.listen_addrs
+            && $a.protocols == $b.protocols
+            && $a.observed_addr == $b.observed_addr
+    };
+}
+
+impl PartialEq for BehaviourEvent {
+    fn eq(&self, other: &Self) -> bool {
```
TheQuantumPhysicist, line 64: Why implementing this manually and not just inheriting it?

TheQuantumPhysicist (approved):
Looks good. We'll have to discuss everything when you're done in more detail.

p2p/src/net/libp2p/connectivity/types.rs
altonen: Because it's not available yet when the peer is dialing/waiting for the information to be received from remote.

p2p/src/net/libp2p/connectivity/types.rs
altonen: I can't inherit it because `identify::IdentifyInfo` doesn't implement it.

p2p/src/net/libp2p/connectivity/types.rs
TheQuantumPhysicist: This is a concerning part for me in designing communication logic in general as a complex state machine. Can't we instead define two structs and the correct transition over state update over an enum instead of updating the mutable state? You don't have to do this right now, but think about it.

altonen: That was the original version of the code (both the one that now exists and the first version of this code) but the resulting state machine was IMO even more complicated than this because state transitions made the code to peek into both connection storages and match against their respective states. Having peer information as `None` while the peer is dialing was in my mind a small price to pay to reduce the complexity. But we can revisit this later if this is not acceptable. I'll add a note to my todo list

p2p/src/net/libp2p/connectivity/types.rs
TheQuantumPhysicist: Instead of manually implementing the eq for this whole thing, create a simple wrapper of IdentifyInfo in which you define equality, then use that instead, then you can inherit Eq. Usually defining equality like this is a landmine from my experience, given that this is the only way to do it in C++.

altonen: I don't see how it's a landmine but I'll try adding a wrapper for `IdentifyInfo`

TheQuantumPhysicist: The issue usually happens when you change the struct and forget to update the Eq impl. As a rule of thumb, any primitive that requires updating two or more places to keep the code consistent is a bad idea. That's why they invented the "spaceship operator" in C++, which solved part of the problem, but not the whole problem. Rust's inheritance solution is wonderful.

muursh (approved):

p2p/src/net/libp2p/behaviour.rs
```diff
@@ -470,3 +421,33 @@ impl NetworkBehaviourEventProcess<mdns::MdnsEvent> for Libp2pBehaviour {
         }
     }
 }
+
+impl NetworkBehaviourEventProcess<connectivity::types::ConnectionManagerEvent> for Libp2pBehaviour {
+    fn inject_event(&mut self, event: connectivity::types::ConnectionManagerEvent) {
+        match event {
+            connectivity::types::ConnectionManagerEvent::Behaviour(event) => match event {
```
stanislav-tkach, line 428: What do you think about adding some usings in this function, so that match arms would be a little shorter? Something like that:
```rust
fn inject_event(&mut self, event: connectivity::types::ConnectionManagerEvent) {
    use connectivity::types::{ConnectionManagerEvent, BehaviourEvent};
    
    match event {
        ConnectionManagerEvent::Behaviour(event) => match event {
            BehaviourEvent::InboundAccepted { ... } => ...
        }
    }
}
```
Perhaps even wildcard import or something like `use connectivity::types::BehaviourEvent as BE` can be used here.

It is just a nitpick so feel free to ignore this comment. 🙃

altonen: Yes good idea, the code is quite verbose. I'll fix them

PR #269, opened 2022-06-30
Adding normal tokens

sinitcin:
# Architecture 

Each transaction output must have a `data` field. This field is the designation of the purpose of the transaction, we must highlight the following at the moment:

* `TokenTransferV1` - Transfer Tokens or NFT
* `TokenIssuanceV1` - Token Issuance
* `TokenBurnV1` - Burning token
* `NftIssuanceV1` - NFT creation

Below, a `data` field representation in TxOutput: 

```rust
    pub struct TxOutput {
        value: OutputValue,
        purpose: OutputPurpose,
    }

    pub enum OutputValue {
        Coin(Amount), 
        Asset(AssetData)
    }

    pub enum AssetData {
        // TokenTransfer data to another user. If it is a token, then the token data must also be transferred to the recipient.
        #[codec(index = 1)]
        TokenTransferV1 { token_id: TokenId, amount: Amount },
        // A new token creation
        #[codec(index = 2)]
        TokenIssuanceV1 {
            token_ticker: Vec<u8>,
            amount_to_issue: Amount,
            // Should be not more than 18 numbers
            number_of_decimals: u8,
            metadata_uri: Vec<u8>,
        },
        // Burning a token or NFT
        #[codec(index = 3)]
        TokenBurnV1 {
            token_id: TokenId,
            amount_to_burn: Amount,
        },
        // Increase amount of tokens 
        #[codec(index = 4)]
        TokenReissueV1 {
            token_id: TokenId,
            amount_to_issue: Amount,
        },
        // A new NFT creation
        #[codec(index = 5)]
        NftIssuanceV1 {
            data_hash: NftDataHash,
            metadata_uri: Vec<u8>,
        },
    }

```

## Token Issuance 

For issuing a new token, we have to create a new transaction with correct filled up `Some(TxOutput.data)`. 

* `token_ticker` - A short name of the issueing token.
* `amount_to_issue` - Amount of new tokens.
* `number_of_decimals` - Token might be integer or float. There is possible to use any number of digits after decimal point, if it not more than 18. 
* `metadata_uri` - A distinct Uniform Resource Identifier (URI) for a given token. There is might be additional information about token (Full name, description, image, array of custom  properties). 

Inputs should have enough amount of MLT to pay creation fee.

After signing and spending the transaction, the token will be created. To operate it, you should use `token_id` that essentially calculate from hash of all inputs of the issue transaction.

There is a possibility to fetch information about TX that created the token by `token_id`. 

Obviously, ownership provided by UTXO architecture. No need to add additional storage data.

[TODO] The URI MUST point to a JSON file that conforms to the "ERC-1155 Metadata URI JSON Schema"? Or we can develop our format? 
[TODO] `token_id` is a hash of all inputs, or it's a hash of the first input? Or, it's hash of TX that created the token?  

## Token Mint

Sometimes, we have to increase amount of issued tokens. It is possible to do only if reissue transaction spend by token creator. Otherwise, tx must be rejected.

* `token_id` - ID of the token (see Token Issuance).
* `amount_to_issue` - It's amount of the tokens that will be issue on creator destination. 

## Token burn

For decrease amount of tokens. Token creator can execute a tx with data:

* `token_id` - ID of the token (see Token Issuance).
* `amount_to_burn` - It's amount of the tokens that will be burn. 

If amount in output is less than in the input then should exist at least one output for returning the funds change. In this case, you can burn any desired number of tokens. After this operation, you can use UTXO for the remaining amount of tokens.

[TODO] Which is destination for output must be?

## NFT creation

For minting a new NFT, we have to create a new transaction with correct filled up `Some(TxOutput.data)`: 

* `data_hash` - The field essentially is a hash of external data, which should be taken from the digital asset.
* `metadata_uri` - A distinct Uniform Resource Identifier (URI) for a given NFT. There is might be additional information about NFT (Full name, description, image, array of custom  properties). 

Inputs should have enough amount of MLT to pay creation fee.

[TODO] What kind of hash we should use?  
[TODO] Might be worth if we research JSON formats for `metadata_uri` that usually use NFT marketplaces, but anyway, this is area of user responsibility.  

## Transfer normal token or NFT

Transfer assets to another person in a given UTXO.

* `token_id` - ID of the token or NFT (see Token Issuance and NFT creation).
* `amount` - Desirable amount of tokens. For NFT this field must be ignored.

## Versioning 

Adding a new version of data is in fact adding a new field to the enum, if the names match, add the version number at the end, for example:

```
TokenTransferV1
TokenTransferV2
etc
```

What happened if someone accidentally changed in `OutputData` arm order or fields order? Nothing good. It's 100% entail incompatibility with other versions of software.

There is only one way to protect fields order - make unit tests and snapshot tests.


iljakuklic:
A couple of observations here:

Why are transfers of MLTs separate from other tokens? Could we just use `AssetData::TokenTransfer` for MLTs and treat MLTs and other stuff uniformly? The extra `OutputValue` enum would become redundant, using `AssetData` directly.

Burning a token could be just sending it to an unspendable destination (we can have a separate destination for this to be more explicit), regardless whether it's a token or a coin. That would also allow us to treat tokens and MLTs more uniformly. A destination seems more appropriate for burning tokens than

Another related idea is that we don't have to mention the big TokenIDs (hashes) directly in transactions. The UTXO set could track the full token IDs, and the transaction could just refer to some input index that identifies the token being used, which could be a `u8` or a `u16` rather than full hash. Would this open some attacks? It seems to me that with some sighash types, the sighash would have to include the full token IDs or some other mitigation, so the small local IDs are guaranteed to refer to the correct global IDs even if the transaction is not signed in full. Any other comments?

TheQuantumPhysicist:
> Why are transfers of MLTs separate from other tokens? Could we just use AssetData::TokenTransfer for MLTs and treat MLTs and other stuff uniformly? The extra OutputValue enum would become redundant, using AssetData directly.

Because they are different types of assets. One is the native token and one is just arbitrary data that we make rules for modification. I can see how they can be unified, but maybe later after we understand the outcome better.

> Burning a token could be just sending it to an unspendable destination (we can have a separate destination for this to be more explicit), regardless whether it's a token or a coin. That would also allow us to treat tokens and MLTs more uniformly. A destination seems more appropriate for burning tokens than

This was a design decision that we made back then in the substrate days. The problem here is that many things have to be consolidated together, one of which is how we treat tokens that are in inputs but not outputs. We basically will prevent that at the protocol level, and the user has to explicitly specify the burn. A specific burn command is always better as it provides clarity.

> Another related idea is that we don't have to mention the big TokenIDs (hashes) directly in transactions. The UTXO set could track the full token IDs, and the transaction could just refer to some input index that identifies the token being used, which could be a u8 or a u16 rather than full hash. Would this open some attacks? It seems to me that with some sighash types, the sighash would have to include the full token IDs or some other mitigation, so the small local IDs are guaranteed to refer to the correct global IDs even if the transaction is not signed in full. Any other comments?

The only problem with that is that transaction validation will not be possible without a full node. It's another design decision. In other words, if someone sends you a transaction with tokens that does tracking like you say, there will be no way to tell whether it's valid without checking the UTXO set, which is very inconvenient. Whether we can use a shorter ID is an open question. The primary issue there is that token IDs will be malleable and a reorg can change the token id, which is a big risk. Perhaps there's a design that solves that problem... not sure.

common/src/chain/tokens/mod.rs
```diff
@@ -0,0 +1,46 @@
+use parity_scale_codec::{Decode, Encode};
```
stanislav-tkach, line 1: Nitpick: the license header is missing. 🙃

sinitcin: Thanks, but this is the draft pull request, and not ready for review yet. When it will be done, I mark this PR as "ready for review". Have a look on github UI, there is a special label where you can see a status. Btw, if you have any suggestions about tokens architecture, welcome to tell us here.

sinitcin:
As Sam mentioned today on the daily, NFT we add in another branch later

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -241,7 +248,7 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
         let inputs = tx.inputs();
         let outputs = tx.outputs();
 
-        let inputs_total = self.calculate_total_inputs(inputs)?;
+        let inputs_total = self.calculate_mlt_total_inputs(inputs)?;
```
TheQuantumPhysicist, line 251: We don't use "mlt" for talking about coins. We called them coins, so we stick to that.

chainstate/src/detail/chainstateref.rs
```diff
@@ -422,46 +425,404 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+        token_ticker: &Vec<u8>,
+        amount_to_issue: &Amount,
+        number_of_decimals: &u8,
+        metadata_uri: &Vec<u8>,
+        tx_id: Id<Transaction>,
+        block_id: Id<Block>,
+    ) -> Result<(), CheckBlockTransactionsError> {
+        const MAX_URI_LEN: usize = 1024;
+        const MAX_DEC_COUNT: u8 = 18;
+        const MAX_TICKER_LEN: usize = 5;
```
stanislav-tkach, line 480: Should these constants be in come config? For example, `ChainstateConfig`.

chainstate/src/detail/chainstateref.rs
```diff
@@ -422,46 +425,404 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                } => {
+                    let token_id = token_id(&prev_tx)?;
+                    (token_id, *amount_to_issue)
+                }
+                AssetData::TokenBurnV1 {
+                    token_id: _,
+                    amount_to_burn: _,
+                } => {
+                    /* Token have burned and can't be transfered */
+                    None?
```
stanislav-tkach, line 554: Nitpick: I guess the question mark isn't necessary here?..

common/Cargo.toml
```diff
@@ -38,6 +38,7 @@ serde_test = "1.0"
 serde_json = "1.0"
 rand = "0.8.5"
 itertools = "0.10"
+chainstate = { path = "../chainstate"}
```
TheQuantumPhysicist, line 42: This is wrong. Storage should not depend on chainstate, because chainstate typically depends on storage.

chainstate/src/detail/error.rs
```diff
@@ -104,8 +104,28 @@ pub enum CheckBlockTransactionsError {
     DuplicateInputInTransaction(Id<Transaction>, Id<Block>),
     #[error("Duplicate input in block")]
     DuplicateInputInBlock(Id<Block>),
-    #[error("Duplicate transaction found in block")]
+    #[error("Duplicate transaction {0} found in block {1}")]
     DuplicatedTransactionInBlock(Id<Transaction>, Id<Block>),
+    #[error("Incorrect issue transaction {0} in block {1}")]
+    TokenIssueFail(Id<Transaction>, Id<Block>),
+    #[error("Incorrect transfer transaction {0} in block {1}")]
+    TokenTransferFail(Id<Transaction>, Id<Block>),
```
TheQuantumPhysicist, line 112: Another meaningless title for an error

chainstate/src/detail/chainstateref.rs
```diff
@@ -525,6 +886,44 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                AssetData::TokenIssuanceV1 {
+                    token_ticker: _,
+                    amount_to_issue: _,
+                    number_of_decimals: _,
+                    metadata_uri: _,
+                } => None,
+                AssetData::TokenBurnV1 {
+                    token_id: _,
+                    amount_to_burn: _,
+                } => None,
```
stanislav-tkach, line 917: @TheQuantumPhysicist probably wouldn't like that, but I would write it this way:
```rust
AssetData::TokenIssuanceV1 { .. } | AssetData::TokenBurnV1 { .. } => None,
```

chainstate/src/detail/error.rs
```diff
@@ -104,8 +104,28 @@ pub enum CheckBlockTransactionsError {
...
+    #[error("Too many token issues in transaction {0} in block {1}")]
+    TooManyTokenIssues(Id<Transaction>, Id<Block>),
```
stanislav-tkach, line 114: Is that only me, or is it a little bit weird to have the "issue" verb in plural?.. I'm not a native speaker, so it somewhat strange for me to complain. 🙃 I have checked [here](https://www.britannica.com/dictionary/issue) and it says that it should be singular:
>  [singular] : the act of officially making something available or giving something to people to be used : the act of issuing something 

Anyway, personally I would say `TooManyTokensIssued`.

https://www.britannica.com/dictionary/issue

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,419 @@
...
+            $block_index,
+            Err(BlockError::CheckBlockFailed(
+                CheckBlockError::CheckTransactionFailed(CheckBlockTransactionsError::$err_name(
+                    _,
+                    _
+                ))
+            ))
+        ));
+    };
+}
```
stanislav-tkach, line 47: Am I right that you are using macro here to preserve the exact line of the (potential) failure? In that case a function with the `#[track_caller]` attribute can be used, too. Personally, I prefer to fallback to a macro only when the same thing cannot be expressed without a boilerplate by other language tools. I'm not sure about this case, though.

test-utils/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        let result = 2 + 2;
+        assert_eq!(result, 4);
+    }
+}
```
stanislav-tkach, line 8: The license header is missing. Additionally, I suppose something else is expected to be placed in this file. Can we remove it if it isn't needed yet?

stanislav-tkach (commented):
I have read through the pull request mostly to understand it, so no comments on design (yet), just some minor style-nitpicks. Feel free to ignore them if you disagree.

common/Cargo.toml
stanislav-tkach: It is only a `dev-dependency`, so it is OK because technically tests is another crate.

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: They should be in ChainConfig. I was about to comment about that.

Anything consensus related (meaning, if it changed it'll cause a fork) should go to chainconfig.

chainstate/src/detail/error.rs
TheQuantumPhysicist: I would even prefer to call it `MultipleTokenIssuanceInTransaction`.

common/Cargo.toml
TheQuantumPhysicist: I noticed it's a `dev-dependency`, but it's just wrong. The only purpose of `dev-dependency` is to reduce compilation time and isolate test dependencies. It should not be used to circumvent the dependency tree.

common/src/chain/config/mod.rs
```diff
@@ -39,6 +39,7 @@ use std::time::Duration;
 
 const DEFAULT_MAX_FUTURE_BLOCK_TIME_OFFSET: Duration = Duration::from_secs(60 * 60);
 pub const DEFAULT_TARGET_BLOCK_SPACING: Duration = Duration::from_secs(120);
+pub const MIN_TOKEN_ISSUANCE_FEE: Amount = Amount::from_atoms(10_000_000_000_000);
```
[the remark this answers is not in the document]
TheQuantumPhysicist: ChainConfig is the right place. It's a consensus property.

common/Cargo.toml
stanislav-tkach: I would like to argue. 🙃 But I have checked and was unable to found how the chainstate depencency is used. Perhaps I'm missing something, but probably it can simply be removed.

chainstate/src/detail/chainstateref.rs
sinitcin: I can't just remove the question mark. If I remove the question mark, then I have to add in other arms `Some`. Not sure, is it better than have question mark or not. What do you think?

test-utils/src/lib.rs
sinitcin: My bad! Forget to remove it!

chainstate/src/detail/chainstateref.rs
sinitcin: Yeah, I miss that, agree with you

chainstate/src/detail/error.rs
sinitcin: Fixed

chainstate/src/detail/chainstateref.rs
sinitcin: Moved these to ChainCoing, and I changed names of consts. Please let me know if you will happy with that.

Cargo.toml
```diff
@@ -28,7 +28,7 @@ members = [
   "wallet",        # wallet executable
   "utils",         # various utilities
   "utxo",          # utxo and related utilities (cache, undo, etc.)
-  "test",          # integration tests
+  "test"           # integration tests
```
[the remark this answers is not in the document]
sinitcin: Done

common/Cargo.toml
sinitcin: Removed

chainstate/src/detail/spend_cache/mod.rs
sinitcin: Fixed! Thanks!

chainstate/src/detail/chainstateref.rs
```diff
@@ -422,46 +425,404 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+            let issuance_count = Self::get_issuance_count(tx);
+            if issuance_count > MAX_ISSUANCE_ALLOWED {
+                return Err(CheckBlockTransactionsError::TooManyTokenIssues(
+                    tx.get_id(),
+                    block.get_id(),
+                ));
+            }
+
+            // Calc MLT and check assets
+            let mut mlt_amount_in_outputs = Amount::from_atoms(0);
```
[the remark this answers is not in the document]
sinitcin: Have a look, please. If now you are happy, I will resolve this conversation

chainstate/src/detail/chainstateref.rs
sinitcin: Why not?! Let's leave here `return None`

chainstate/src/detail/chainstateref.rs
```diff
@@ -422,46 +427,393 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                tx.get_id(),
+                block_id.clone(),
+            ));
+        }
+        if amount_to_burn == &Amount::from_atoms(0) {
+            return Err(CheckBlockTransactionsError::BurnZeroTokens(
+                tx.get_id(),
+                block_id.clone(),
+            ));
+        }
```
altonen, line 761: `ensure!()`?

chainstate/src/detail/error.rs
muursh: "Too many token issues in transaction {0} in block {1}"
The issue exists here too. "issued" should be "issued" and "token" should be "tokens"

sinitcin: @muursh `Too many tokens issued in transaction {0} in block {1}` ?

chainstate/src/detail/chainstateref.rs
sinitcin: Thanks!

chainstate/src/detail/error.rs
muursh: Yep

common/src/chain/tokens/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author(s): A. Sinitsyn
+
+use parity_scale_codec::{Decode, Encode};
+
```
TheQuantumPhysicist, line 19: Use serialization crate instead of parity_scale_codec

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: Using `..` should be avoided. I don't mind using `|`. The idea is to force you to handle any change you do to data types.

chainstate/src/detail/chainstateref.rs
```diff
@@ -422,46 +427,402 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                    output.value(),
+                    OutputValue::Asset(AssetData::TokenIssuanceV1 {
+                        token_ticker: _,
+                        amount_to_issue: _,
+                        number_of_decimals: _,
+                        metadata_uri: _,
+                    })
+                ) {
+                    true => Some(()),
+                    false => None,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Please don't use `matches!` here. Use `match`, and keep the irrelevant arms empty. I'd like to make it clear that using `matches!` in production code is discouraged. It's OK to use it in tests, but in production code, typically a full match with ignored arms is the right way to go.

Notice the pattern: With pattern matching, using `_`, `..`, `matches!` and `if let`, all have the same effect: They make you ignore other arms which makes you fail to handle new cases. All discouraged, all shouldn't be used. Explicitly handling arms is how we do it.

chainstate/src/detail/chainstateref.rs
stanislav-tkach: Don't we always want to return `None` here even if new fields are added?.. I agree that it is very nice for the language to force handling of new enum variants, but I'm not so sure about fields.

chainstate/src/detail/chainstateref.rs
```diff
@@ -422,46 +427,393 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                    .db_tx
+                    .get_mainchain_tx_by_position(tx_pos)
+                    .ok()
+                    .flatten()
+                    .ok_or(CheckBlockTransactionsError::FetchFail)?;
+
+                Ok(tx)
+            }
+            common::chain::SpendablePosition::BlockReward(_) => {
+                todo!("Block reward in tokens is not allowed")
```
TheQuantumPhysicist, line 525: Please don't use `todo!` unless it's something that will be done by the end of the PR, but return a proper error.

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: What if it's TransferV2?

stanislav-tkach: I'm not sure I understand the comment about the potential new enum variant. If [the new variant is added](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3b142204c49cbcfc5698ab6fac09336f), then it only needs to be handled in the match expression. So `TokenIssuanceV1` and `TokenBurnV1` can stay as they are. Perhaps I'm missing something?.. Are you talking about the situation when we need to introduce a new transfer version which will require to change the logic in token issuance and token burn handling without introducing the corresponding V2 versions?

stanislav-tkach: OK, if you are really insisting I'm not going to complain any longer and will always follow this rule, but I think that it should be mentioned somewhere in the `CONTRIBUTING.md`.

chainstate/src/detail/error.rs
sinitcin: @muursh thanks, fixed!

chainstate/src/detail/ban_score.rs
```diff
@@ -126,6 +126,16 @@ impl BanScore for CheckBlockTransactionsError {
             CheckBlockTransactionsError::DuplicateInputInTransaction(_, _) => 100,
             CheckBlockTransactionsError::DuplicateInputInBlock(_) => 100,
             CheckBlockTransactionsError::DuplicatedTransactionInBlock(_, _) => 100,
+            CheckBlockTransactionsError::TokenIssueFail(_, _) => 100,
+            CheckBlockTransactionsError::TokenTransferFail(_, _) => 100,
```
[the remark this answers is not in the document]
sinitcin: Thanks! I've got you guys.

chainstate/src/detail/chainstateref.rs
sinitcin: I see, thanks!

chainstate/src/detail/error.rs
sinitcin: Removed! Thanks!

chainstate/src/detail/error.rs
```diff
@@ -104,8 +104,28 @@ pub enum CheckBlockTransactionsError {
     DuplicateInputInTransaction(Id<Transaction>, Id<Block>),
     #[error("Duplicate input in block")]
     DuplicateInputInBlock(Id<Block>),
-    #[error("Duplicate transaction found in block")]
+    #[error("Duplicate transaction {0} found in block {1}")]
     DuplicatedTransactionInBlock(Id<Transaction>, Id<Block>),
+    #[error("Incorrect issue transaction {0} in block {1}")]
+    TokenIssueFail(Id<Transaction>, Id<Block>),
```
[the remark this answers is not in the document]
sinitcin: Agree with you. When I wrote it, I felt uncomfortable with these errors, but didn't see why. Thank you for explanation. Can you have a look? If you will be happy, we can resolve it.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Removed assert_token

chainstate/src/detail/chainstateref.rs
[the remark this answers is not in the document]
sinitcin: Yep, `matches!` here was completely unnecessary. I made this place more clean due to using `filter_map` and `fold`. Have a look, and if you happy, I'll resolve this.

chainstate/src/detail/chainstateref.rs
sinitcin: Unfortunately, on some places like `check_issue_data` it's inconvenient, because cause errors to borrow-checker. I think, there is better to use `if` and `return` rather a bunch of `clone` on `tx_id` and `block_id`

sinitcin: In other places, I started to use `ensure!`. If you are happy with that, I resolve conversation

common/src/chain/tokens/mod.rs
sinitcin: Fixed

chainstate/src/detail/chainstateref.rs
```diff
@@ -422,46 +427,393 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+            && metadata_uri.len() <= TOKEN_MAX_URI_LEN
+            && String::from_utf8_lossy(metadata_uri).is_ascii();
+
+        match is_correct_data {
+            true => Ok(()),
+            false => Err(CheckBlockTransactionsError::TokenIssueFail(tx_id, block_id)),
+        }
+    }
+
+    fn fetch(&self, outpoint: &OutPoint) -> Result<Transaction, CheckBlockTransactionsError> {
```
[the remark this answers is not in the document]
sinitcin: Would be better to use `fetch_tx_by_outpoint`?

chainstate/src/detail/chainstateref.rs
```diff
@@ -438,46 +443,179 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                        block.get_id(),
+                    ))
+                }
+                None => txs_ids.insert(tx_id),
+            };
+        }
+        Ok(())
+    }
+
+    fn check_issue_data(
```
TheQuantumPhysicist, line 487: check_token_issuance_data

chainstate/src/detail/chainstateref.rs
```diff
@@ -438,46 +443,179 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+        // Check URI
+        if metadata_uri.len() > TOKEN_MAX_URI_LEN
+            || !String::from_utf8_lossy(metadata_uri).is_ascii()
+        {
+            return Err(TokensError::IssueErrorIncorrectMetadataURI(tx_id, block_id));
+        }
+        Ok(())
+    }
+
+    fn check_transfer_data(
```
TheQuantumPhysicist, line 525: check_token_transfer_data

chainstate/src/detail/chainstateref.rs
```diff
@@ -438,46 +443,179 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+        &self,
+        block_id: Id<Block>,
+        tx: &Transaction,
+        _token_id: &TokenId,
+        amount: &Amount,
+    ) -> Result<(), TokensError> {
+        // Check amount
+        ensure!(
+            amount > &Amount::from_atoms(0),
+            TokensError::TransferZeroTokens(tx.get_id(), block_id,)
```
TheQuantumPhysicist, line 535: Extra comma

chainstate/src/detail/chainstateref.rs
```diff
@@ -438,46 +443,179 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+    ) -> Result<(), TokensError> {
+        // Check amount
+        ensure!(
+            amount_to_burn != &Amount::from_atoms(0),
+            TokensError::BurnZeroTokens(tx.get_id(), *block_id)
+        );
+        Ok(())
+    }
+
+    fn check_asset(
```
TheQuantumPhysicist, line 580: You know, let's call everything tokens and get rid of "asset". Replace all "asset" instances with "token". We can't have both.

chainstate/src/detail/chainstateref.rs
```diff
@@ -438,46 +443,179 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                    OutputValue::Coin(_) => None,
+                    OutputValue::Asset(asset) => Some(asset),
+                })
+                .try_for_each(|asset| self.check_asset(asset, tx, block))
+                .map_err(CheckBlockTransactionsError::CheckTokensError)?;
+        }
+        Ok(())
+    }
+
+    fn check_burn_data(
```
TheQuantumPhysicist, line 565: All these checks for tokens should go to a separate submodule. No need to mix them all up in the chainstateref

sinitcin: Ok

TheQuantumPhysicist (commented):
You seem to have forgotten something. You need some storage in chainstate storage for tokens on issuance so they can be consulted when knowing about the token. Currently you're just checking utxos.

Also you need an rpc function that queries the token information. See how other chainstate rpc functions are created.

I'm not done reviewing, but this is an important part that I noticed to be missing.

sinitcin:
> You seem to have forgotten something. You need some storage in chainstate storage for tokens on issuance so they can be consulted when knowing about the token. Currently you're just checking utxos.

OK, I haven't added storage part, because thought it needs only for re-issuance, that we decide to do later.

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -205,15 +209,105 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
+    }
+
+    pub fn calculate_assets_total_inputs(
+        &self,
+        inputs: &[TxInput],
+    ) -> Result<BTreeMap<TokenId, Amount>, StateUpdateError> {
+        let mut total_assets: BTreeMap<TokenId, Amount> = BTreeMap::new();
+        for (_input_idx, input) in inputs.iter().enumerate() {
+            let outpoint = input.outpoint();
+            let tx_index = match self.inputs.get(&outpoint.tx_id()) {
```
TheQuantumPhysicist, line 257: Since this call is done multiple times around, abstract it away with a function call and use `?` to return an error, if any.

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -498,6 +602,191 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
+        token_id: &TokenId,
+        amount: &Amount,
+    ) -> Result<(), TokensError> {
+        // Collect token inputs
+        let total_value_tokens = self
+            .calculate_assets_total_inputs(tx.inputs())
+            .map_err(|_err| TokensError::NoTokenInInputs(tx.get_id(), block_id))?;
+
+        // Is token exist in inputs?
+        let (_, origin_amount) = total_value_tokens
```
TheQuantumPhysicist, line 618: This implementation is incorrect.

To prove this, create a test that has one input of token X with value 100. Then create two or more outputs with value 100 + 100 + 100 + 100... You basically can print as much tokens as you want with this bug by having more outputs that spend the same input.

The mistake here is that you're not removing the tokens you marked as spent from the map. Since we're doing things functionally and we dislike using mutable variables, the better way is to create a map with the "spent tokens", similar to the total_value_tokens map, but then every output you "consume" adds to that map (here in this line of the code). Then, your comparison should be with the value of the token in the map total_value_tokens minus the one in the map that have the "consumed" tokens.

TheQuantumPhysicist: A simpler solution is perhaps to just create another map for the total in outputs, and compare.

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -498,6 +602,191 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
+        // Check amount
+        ensure!(
+            origin_amount >= amount,
+            TokensError::InsuffienceTokenValueInInputs(tx.get_id(), block_id,)
+        );
+
+        Ok(())
+    }
+
+    fn check_connected_issue_data(
```
TheQuantumPhysicist, line 632: So the token can be infinitely long?

Please check the rules that we used in substrate and implement them here.

sinitcin: These checks are done in `check_block`.

chainstate/src/detail/chainstateref.rs
sinitcin: Done

chainstate/src/detail/chainstateref.rs
sinitcin: Done

chainstate/src/detail/spend_cache/mod.rs
sinitcin: Fixed!

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -201,32 +207,157 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
+                    amount_to_burn: _,
+                } => {
+                    /* Token have burned and can't be transfered */
+                    return None;
+                }
+            }),
+        }
+    }
+
+    pub fn calculate_transfered_and_burn_tokens(
```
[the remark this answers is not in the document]
sinitcin: Fixed

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -201,32 +207,157 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
             tx_id: Some(spender_id),
             source_output_index: output_index,
         })?;
         Ok(output.value())
     }
 
-    fn calculate_total_inputs(&self, inputs: &[TxInput]) -> Result<Amount, StateUpdateError> {
-        let mut total = Amount::from_atoms(0);
+    // Get TokenId and Amount in input
+    fn filter_transfered_and_issued_amounts(
```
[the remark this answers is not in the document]
sinitcin: Fixed

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+    primitives::{Amount, Id, Idable},
+};
+use utils::ensure;
+
+use crate::detail::TokensError;
+
+pub fn check_token_transfer_data(
+    block_id: Id<Block>,
+    tx: &Transaction,
+    _token_id: &TokenId,
```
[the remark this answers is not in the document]
sinitcin: For now, it's unnecessary. I think we can remove it.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,109 @@
+use common::{
```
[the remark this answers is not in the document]
sinitcin: Fixed

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -498,6 +641,196 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
+    // Calc not burned tokens that were placed in OutputValue::TokenTransfer after partial burn
+    fn get_change_amount(tx: &Transaction, burn_token_id: &TokenId) -> Result<Amount, TokensError> {
+        let change_amount = tx
+            .outputs()
+            .iter()
+            .filter_map(|x| match x.value() {
+                OutputValue::Coin(_) => None,
+                OutputValue::Token(asset) => match asset {
+                    TokenData::TokenTransferV1 { token_id, amount } => {
+                        if token_id == burn_token_id {
```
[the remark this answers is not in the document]
sinitcin: Perfect!

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +621,193 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                })
+                .try_for_each(|token_data| self.check_connected_tokens(token_data, tx, block))
+                .map_err(ConnectTransactionError::TokensError)?;
+
+            // If it is not a genesis and in tx issuance
+            if self
+                .db_tx
+                .get_best_block_id()
+                .map_err(ConnectTransactionError::StorageError)?
+                .is_some()
```
altonen, line 790: does this check make sense? I've understood that best block should always be available

sinitcin: I think so. I guess, in the future, some tokens might be connected through genesis. If it is the genesis block, we won't check fees for issuance and issuance count. Correct me if I'm wrong @TheQuantumPhysicist  

However, I'm still working on it, and next commits make the code clearer.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) 2022 RBB S.r.l
```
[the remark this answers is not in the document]
sinitcin: Probably, I miss something, but seems all not mine files include this license header, but without `Authors` field. May I ask you where I can get a new one?

chainstate/src/detail/mod.rs
```diff
@@ -409,6 +412,90 @@ impl<S: BlockchainStorage> Chainstate<S> {
...
+
+        // Find issuance transaction id by token_id
+        let creation_tx_id = chainstate_ref
+            .get_token_tx(token_id)?
+            .ok_or(TokensError::TokensNotRegistered(token_id))
+            .map_err(PropertyQueryError::TokensError)?;
+        let tx_index = chainstate_ref
+            .get_mainchain_tx_index(&OutPointSourceId::Transaction(creation_tx_id))?
+            .ok_or(TokensError::TokensNotRegistered(token_id))
+            .map_err(PropertyQueryError::TokensError)?;
```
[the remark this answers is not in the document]
sinitcin: Skip that, It must be removed

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +621,193 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+    fn check_connected_burn_data(
+        &self,
+        tx: &Transaction,
+        block_id: &Id<Block>,
+        burn_token_id: &TokenId,
+        amount_to_burn: &Amount,
+    ) -> Result<(), TokensError> {
+        // Collect token inputs
+        let total_value_tokens = self
+            .calculate_tokens_total_inputs(tx.inputs())
```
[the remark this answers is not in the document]
sinitcin: That's a good point, will upgrade that

chainstate/src/detail/tokens/mod.rs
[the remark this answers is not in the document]
altonen: The authors field and the URL to the license were changed, here's the new license: https://github.com/mintlayer/mintlayer-core/blob/master/chainstate-storage/src/lib.rs#L1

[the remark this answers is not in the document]
sinitcin: Thank you, Aaro!

chainstate/src/detail/transaction_verifier/mod.rs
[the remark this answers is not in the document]
sinitcin: Done

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +665,206 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                && get_tokens_issuance_count(tx) == TOKEN_MAX_ISSUANCE_ALLOWED
+            {
+                // check is fee enough for issuance
+                ensure!(
+                    self.check_transferred_amounts_and_get_fee(tx).map_err(|_| {
+                        ConnectTransactionError::TokensError(TokensError::InsuffienceTokenFees(
+                            tx.get_id(),
+                            block.get_id(),
+                        ))
+                    })? >= TOKEN_MIN_ISSUANCE_FEE,
```
sinitcin, line 762: BTW, `TOKEN_MIN_ISSUANCE_FEE` are equal to 10_000_000_000_000. This value was used in Substrate, not sure is it correct for now or not. Correct me if I'm wrong @TheQuantumPhysicist

.github/workflows/coverage.yml
```diff
@@ -13,7 +13,8 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-03-22
+          #toolchain: nightly-2022-03-22
+          toolchain: nightly
```
TheQuantumPhysicist, line 17: I don't know if there's been a discussion about this, but this is probably a bad idea. Nightly sometimes have bugs and things aren't stable. We should pick a date and stick to it. If there's a feature that doesnt' work with an old nightly, just switch to last night's nightly.

common/src/chain/tokens/rpc.rs
```diff
@@ -0,0 +1,65 @@
...
+// use crate::primitives::VersionTag;
+use crate::{
+    chain::{Block, Transaction},
+    primitives::{Amount, Id},
+};
+
+use super::TokenId;
+
+#[derive(Debug, Clone, Encode, Decode, serde::Serialize, serde::Deserialize)]
+pub struct RPCTokenInfoV1 {
```
TheQuantumPhysicist, line 28: Versioning is not arbitrary. We only version things that are part of consensus. The reason for versioning is to be able to support protocol upgrades. If I'm not mistaken, this struct is not used in anything protocol-related. This is just used as an abstraction layer for token info. Meaning, if we change it, it doesn't break the blockchain by failing to deserialize old things, it just changes the printed information in RPC... well, who cares? If that's the case, please remove the V1 from the name.

common/src/chain/tokens/rpc.rs
```diff
@@ -0,0 +1,65 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::Decode;
+use serialization::Encode;
+
+// use crate::primitives::VersionTag;
```
TheQuantumPhysicist, line 19: ?

common/src/chain/tokens/rpc.rs
```diff
@@ -0,0 +1,65 @@
...
+    chain::{Block, Transaction},
+    primitives::{Amount, Id},
+};
+
+use super::TokenId;
+
+#[derive(Debug, Clone, Encode, Decode, serde::Serialize, serde::Deserialize)]
+pub struct RPCTokenInfoV1 {
+    // TODO: Should we use here VersionTag?
+    pub version: u32,
```
TheQuantumPhysicist, line 30: What does this version represent?

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+
+pub fn check_tokens_issuance_data(
+    token_ticker: &Vec<u8>,
+    amount_to_issue: &Amount,
+    number_of_decimals: &u8,
+    metadata_uri: &Vec<u8>,
+    tx_id: Id<Transaction>,
+    block_id: Id<Block>,
+) -> Result<(), TokensError> {
+    //TODO: Shall we have a check for unique token name?
```
TheQuantumPhysicist, line 61: 1. Did we decide we want to enforce that back then with substrate?
2. If yes, then that enforcement can only be done when connecting transactions, because it depends on blockchain history.

.github/workflows/coverage.yml
sinitcin: Unfortunately, some features wasn't stable on `nightly-2022-03-22`, for example `bool_to_option` feature that provide `then_some` function.  

I understand risks, but looks the current nightly works fine, also the current stable works fine too. Btw, the stable version have more features than `nightly-2022-03-22`. So, I see here a couple of solutions:
1. We can use the latest stable, that might confront with previous discussions. But `nightly-2022-03-22` more outdated than the current stable. 
2. Stick with the current nightly. 

I agree with both variants, because a nightly build that released almost 6 months ago, it's quite outdated.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+    if metadata_uri.len() > TOKEN_MAX_URI_LEN || !String::from_utf8_lossy(metadata_uri).is_ascii() {
+        return Err(TokensError::IssueErrorIncorrectMetadataURI(tx_id, block_id));
+    }
+    Ok(())
+}
+
+pub fn check_tokens_data(
+    token_data: &TokenData,
+    tx: &Transaction,
+    block: &Block,
```
TheQuantumPhysicist, line 91: If you only need the block id, then just pass the block id. Maybe call it `source_block_id`

common/src/chain/tokens/rpc.rs
sinitcin: Ok, no problem. Thanks for explanation

.github/workflows/coverage.yml
TheQuantumPhysicist: I think you misunderstood what I'm saying. I'm not saying to stick to 22-03-2022. I'm saying use the one from yesterday, 16-08-2022, instead of keeping it unbounded with just `nightly`.

If it works with stable, that's even better. I don't really know why we're using nightly, but obviously I imagine there's something not ready for coverage on stable.

common/src/chain/tokens/rpc.rs
sinitcin: As you explain in the previous comment that there is no need to add versioning, I'm going to remove this field. 

Before that, I thought that might be we have to use versioning, and was doubts about should we use VersionTag like it does in BlockHeader, or like I did with token data. This is why I left here the todo.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+            amount_to_burn,
+        } => {
+            check_tokens_burn_data(tx, &block.get_id(), amount_to_burn)?;
+        }
+    }
+    Ok(())
+}
+
+// Get TokenId and Amount in input
+pub fn filter_transferred_and_issued_amounts(
```
TheQuantumPhysicist, line 126: This is wrong.

When you calculate the total number of tokens in outputs you should also add the burned outputs, because you can't burn outputs you don't have.

Please create a test that proves this bug, where you create an asset/token, transfer 10 tokens in the inputs, but burn 10000000 in the outputs.

.github/workflows/coverage.yml
sinitcin: I understood you, no worries. Ok, let's leave yesterday nightly version, and when we have free hands to inspect is coverage ready for the stable, then change it on in another branch. 

p.s. As far as I remember, we started to use nightly due to substrate, it doesn't build on stable. And then we just kept it as legacy.

common/src/chain/tokens/rpc.rs
sinitcin: Removed versioning

chainstate/src/detail/tokens/mod.rs
sinitcin: 1. We didn't. But we spoke a lot about cases like when someone can send some token (USDT for example), but it wouldn't be original USDT. Also, NEAR blockchain require unique names for tokens. However, Ethereum allows using the same names for different tokens, and the only thing that makes your token unique is its address. As far as I remember, we left this decision for future time. 

2. Probably, for now we can create an issue, and just left a comment. What do you think?

chainstate/src/detail/chainstateref.rs
```diff
@@ -452,46 +459,75 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                .try_for_each(|token| tokens::check_tokens_data(token, tx, block))
+                .map_err(CheckBlockTransactionsError::TokensError)?;
         }
+        Ok(())
+    }
 
+    fn check_transactions(&self, block: &Block) -> Result<(), CheckBlockTransactionsError> {
+        self.check_duplicate_inputs(block)?;
+        self.check_duplicate_txs(block)?;
+        self.check_tokens_txs(block)?;
```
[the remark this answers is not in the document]
sinitcin: I think, there is a question about balance of readable, and efficient. In previous iterations of the review, people told me that we have to split functions to a small one with more clear code. However, over here, two of three functions, use `BTreeSet` as temporary mut value for intermediate computations. If we use through the one cycle, we have to expose mut parameters out of functions. That probably a bad idea. Or we should combine all functions into big one, that seems not a better solution too. Probably, @TheQuantumPhysicist can be pointing us to the right solution.

chainstate/src/detail/tokens/mod.rs
sinitcin: Yep, sounds good, I fixed it

chainstate/src/detail/chainstateref.rs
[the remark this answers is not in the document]
azarovh: You are right that readability is important here. I would just try to move loop out of functions and leave these 3 function as subroutines doing their checks on a single tx. But again this is more of a suggestion.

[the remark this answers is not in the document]
TheQuantumPhysicist: It doesn't really matter if you loop one or 10 times... every loop iteration costs one extra integer comparison... that's literally nothing even for a raspberry pi. It's better to keep things as independent as possible so that they can be individually tested (if we come to want to do it).

azarovh (approved):

chainstate/src/detail/tokens/mod.rs
TheQuantumPhysicist: Yes, create an issue for it.

common/src/chain/config/mod.rs
```diff
@@ -37,6 +36,11 @@ use std::time::Duration;
 
 const DEFAULT_MAX_FUTURE_BLOCK_TIME_OFFSET: Duration = Duration::from_secs(60 * 60);
 pub const DEFAULT_TARGET_BLOCK_SPACING: Duration = Duration::from_secs(120);
+pub const TOKEN_MIN_ISSUANCE_FEE: Amount = Amount::from_atoms(10_000_000_000_000);
+pub const TOKEN_MAX_URI_LEN: usize = 1024;
+pub const TOKEN_MAX_DEC_COUNT: u8 = 18;
+pub const TOKEN_MAX_TICKER_LEN: usize = 5;
+pub const TOKEN_MAX_ISSUANCE_ALLOWED: usize = 1;
```
stanislav-tkach, line 43: Should these values be in `ChainConfig`?..

stanislav-tkach, line 43: Should these values be in `ChainConfig`?..

common/src/chain/tokens/mod.rs
```diff
@@ -15,14 +15,54 @@
...
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
 pub enum OutputValue {
     Coin(Amount),
+    Token(TokenData),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
 pub enum TokenData {
-    // Add todo in the code, if you should be processing tokens
+    // TokenTransfer data to another user. If it is a token, then the token data must also be transferred to the recipient.
```
stanislav-tkach, line 38: Nitpick: I feel like these can be doc-comments.

stanislav-tkach (approved):

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+            } => {
+                let token_id = token_id(prev_tx)?;
+                (token_id, *amount_to_issue)
+            }
+            TokenData::TokenBurnV1 {
+                token_id: _,
+                amount_to_burn: _,
+            } => {
+                /* Token have burned and can't be transferred */
+                return None;
```
TheQuantumPhysicist, line 160: So someone is attempting to spend burnt coins, and we're OK with that?
This function should return a result with an error when trying to spend burned tokens. This is an invalid operation.

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: Please put this value in the chain_config

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +577,212 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                token_id,
+                amount_to_burn,
+            } => {
+                self.check_connected_burn_data(tx, &block.get_id(), token_id, amount_to_burn)?;
+            }
+        }
+        Ok(())
+    }
+
+    fn check_tokens_values(&mut self, block: &Block) -> Result<(), ConnectTransactionError> {
```
TheQuantumPhysicist, line 742: Please move this and all other token-related stuff to a submodule. No need to make this a bigger pan of soup of unrelated things. The file is already 1000 lines long.

common/src/chain/tokens/errors.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[error("Too many decimals in issue transaction {0} in block {1}")]
+    IssueErrorTooManyDecimals(Id<Transaction>, Id<Block>),
+    #[error("Incorrect metadata URI in issue transaction {0} in block {1}")]
+    IssueErrorIncorrectMetadataURI(Id<Transaction>, Id<Block>),
+    #[error("Too many tokens issued in transaction {0} in block {1}")]
+    MultipleTokenIssuanceInTransaction(Id<Transaction>, Id<Block>),
+    #[error("Coin or token overflow")]
+    CoinOrTokenOverflow,
+    #[error("Tokens fees insuffience in transaction {0} in block {1}")]
+    InsuffienceTokenFees(Id<Transaction>, Id<Block>),
```
TheQuantumPhysicist, line 38: InsufficientTokenFees

common/src/chain/tokens/errors.rs
```diff
@@ -0,0 +1,57 @@
...
+    #[error("Tokens value in inputs insuffience in transaction {0} in block {1}")]
+    InsuffienceTokenValueInInputs(Id<Transaction>, Id<Block>),
```
TheQuantumPhysicist, line 40: InsufficientTokensInInputs

Please also fix the str descriptions

common/src/chain/config/mod.rs
TheQuantumPhysicist: Yes

chainstate/src/detail/tests/mod.rs
```diff
@@ -74,6 +75,35 @@ fn create_utxo_data(
...
+                } => TxOutput::new(
+                    OutputValue::Token(asset.clone()),
+                    OutputPurpose::Transfer(anyonecanspend_address()),
+                ),
+                TokenData::TokenIssuanceV1 {
+                    token_ticker: _,
+                    amount_to_issue: _,
+                    number_of_decimals: _,
+                    metadata_uri: _,
+                } => todo!("Add token_id here and use bellow code"),
```
muursh, line 91: below

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,837 @@
...
+        assert!(matches!(
+            process_token(&mut test_framework, ParentBlock::BestBlock, values),
+            Err(BlockError::CheckBlockFailed(
+                CheckBlockError::CheckTransactionFailed(CheckBlockTransactionsError::TokensError(
+                    TokensError::IssueErrorIncorrectTicker(_, _)
+                ))
+            ))
+        ));
+
+        // Doesn't exist ticker
```
muursh, line 141: "Ticker doesn't exist"

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,837 @@
...
+
+#[test]
+fn token_issue_test() {
+    common::concurrency::model(|| {
+        // Process token without errors
+
+        let mut test_framework = TestFramework::default();
+        // let mut chainstate = setup_chainstate();
+        let values = vec![OutputValue::Token(TokenData::TokenIssuanceV1 {
+            token_ticker: b"USDC".to_vec(),
```
muursh, line 113: Change USDC to ABCD. I'm 99.99% sure we won't run in to issues here and sure it's a test but given it's probably a trademark (that I'm too lazy to check for) I'd rather err on the side of caution

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,837 @@
...
+        assert!(matches!(
+            process_token(&mut test_framework, ParentBlock::BestBlock, values),
+            Err(BlockError::CheckBlockFailed(
+                CheckBlockError::CheckTransactionFailed(CheckBlockTransactionsError::TokensError(
+                    TokensError::IssueErrorIncorrectTicker(_, _)
+                ))
+            ))
+        ));
+
+        // Ticker contain not alpha-numeric byte
```
muursh, line 157: not -> non

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,837 @@
...
+            Err(BlockError::CheckBlockFailed(
+                CheckBlockError::CheckTransactionFailed(CheckBlockTransactionsError::TokensError(
+                    TokensError::IssueErrorIncorrectTicker(_, _)
+                ))
+            ))
+        ));
+
+        // Issue amount is too low
+        let values = vec![OutputValue::Token(TokenData::TokenIssuanceV1 {
+            token_ticker: "USDT".as_bytes().to_vec(),
```
muursh, line 175: Same...

muursh: Okay there's a bunch of these. It applied to all.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +577,212 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+        for tx in block.transactions() {
```
TheQuantumPhysicist, line 743: So this function is called for every `connect_transaction()` call, where this checks every transaction in a block. Which means, if we have a block with 1000 transactions in it, for every transaction in that block we'll recheck every transaction 1000000 times.

sinitcin: Good catch! I guess, we can rename `check_tokens_values` to `connect_tokens`, and use it like that: 

```rust

    fn make_cache_with_connected_transactions(
        &self,
        block: &Block,
        spend_height: &BlockHeight,
    ...
        let mut tx_verifier = TransactionVerifier::new(&self.db_tx, self.chain_config);
        tx_verifier.connect_transaction(
            BlockTransactableRef::BlockReward(block),
            spend_height,
            &median_time_past,
            blockreward_maturity,
        )?;
        tx_verifier.connect_tokens(block)?;
    ...
    ```
`check_tokens_values` use the whole block, we can't make tokens calculations on some tx apart from others txs.

File: chainstate/src/detail/transaction_verifier/mod.rs [hunk 59]
```diff
@@ -489,6 +577,212 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                    }
+                    TokenData::TokenIssuanceV1 {
+                        token_ticker: _,
+                        amount_to_issue: _,
+                        number_of_decimals: _,
+                        metadata_uri: _,
+                    }
+                    | TokenData::TokenBurnV1 {
+                        token_id: _,
+                        amount_to_burn: _,
```
TheQuantumPhysicist, line 674: This doesn't need to be mutable.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Ok, I rid off all USDT and USDC mentions. Thanks!

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +577,212 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                        block.get_id()
+                    ),)
+                );
+            }
+        }
+        Ok(())
+    }
+
+    fn is_issuance_in_custom_block(
+        &mut self,
```
TheQuantumPhysicist, line 774: This doesn't need to be mutable

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Fixed

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: Nope. Connect implies marking inputs as spent. Simply use BlockTransactableRef instead of using a whole block.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Just left everywhere - `SOME` ticker, I think this is enough

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Fixed

chainstate/src/detail/tests/mod.rs
sinitcin: Yeah, that's true

common/src/chain/tokens/errors.rs
sinitcin: Done! Should I change `InsuffienceTokenFees` too?

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: Or just use the transaction itself. It was extracted in a prior step

common/src/chain/tokens/errors.rs
sinitcin: Renamed

common/src/chain/tokens/errors.rs
sinitcin: Fixed

chainstate/src/detail/chainstateref.rs
```diff
@@ -452,46 +459,75 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                    block_inputs.insert(input.outpoint()),
+                    CheckBlockTransactionsError::DuplicateInputInBlock(block.get_id())
+                );
             }
         }
+        Ok(())
+    }
+
+    fn check_duplicate_txs(&self, block: &Block) -> Result<(), CheckBlockTransactionsError> {
+        // check duplicate transactions
```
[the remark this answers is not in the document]
sinitcin: Yeah, that was a part of another function and I just move this code block that includes the comment, into a function. 

Fixed now

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +577,212 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+            } => {
+                self.check_connected_issue_data(
+                    token_ticker,
+                    amount_to_issue,
+                    number_of_decimals,
+                    metadata_uri,
+                    tx.get_id(),
+                    block.get_id(),
+                )?;
+                self.register_issuance(tx)?;
```
TheQuantumPhysicist, line 730: This function is "check" yet it's mutable and "registers" issuance mutably. Please move issuance registration and separate it from "checking". Mutable functions must be isolated and as small as possible.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +577,212 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                        tx.get_id(),
+                        block.get_id()
+                    ),)
+                );
+            }
+        }
+        Ok(())
+    }
+
+    fn is_issuance_in_custom_block(
```
TheQuantumPhysicist, line 773: What does "custom" mean? Please refrain from introducing arbitrary terms. If what you mean here is "non_genesis", then call it that. 

In fact, I'm really having trouble understanding the purpose of this. Why does it matter if the current best is genesis or not?

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +577,212 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+    fn check_tokens_values(&mut self, block: &Block) -> Result<(), ConnectTransactionError> {
+        for tx in block.transactions() {
+            // Check tokens before connect tx
+            tx.outputs()
+                .iter()
+                .filter_map(|output| match output.value() {
+                    OutputValue::Coin(_) => None,
+                    OutputValue::Token(token_data) => Some(token_data),
+                })
+                .try_for_each(|token_data| self.check_connected_tokens(token_data, tx, block))
```
TheQuantumPhysicist, line 751: This is very suboptimal. For every output, you recheck all the token amounts of a token id. Meaning, if a 5 outputs use the same token id, the checking will be done 5 times, again and again and again. This is a bad design.

The only reason this is not a performance disaster is because inputs are already precached at this point.

I'm not satisfied with how this is written. It's too complex, and very non-linear. This has to be completely revamped to check token amounts exactly once. Not as many outputs we have. I will write a comment on the PR explaining more in detail.

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: Also, "check" can never be mutable.

chainstate/src/detail/chainstateref.rs
```diff
@@ -452,46 +459,75 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+            let already_in_tx_id = txs_ids.get(&tx_id);
+            match already_in_tx_id {
+                Some(_) => {
+                    return Err(CheckBlockTransactionsError::DuplicatedTransactionInBlock(
+                        tx_id,
+                        block.get_id(),
+                    ))
+                }
+                None => txs_ids.insert(tx_id),
+            };
```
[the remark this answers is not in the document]
sinitcin: Fixed!

TheQuantumPhysicist:
I'm sorry. This implementation is not good enough to be merged. Let me suggest the way to start doing this right.

The functions `calculate_coins_total_inputs()` and `calculate_coins_total_outputs()` should return a `BTreeMap<CoinOrTokenId, Amount>`. The CoinOrTokenId should be an enum that's either a token id or a coin.

The check functions must be 100% immutable, and obviously shouldn't depend on history or be in tx_verifier (which you mostly have already, except for the immutability). Meaning: While connecting a transaction, we check everything in that transaction (we double-check that the token data is valid). But if we don't connect it, it just gets checked once in check_block().

You can create a separate function in `transaction_verifier::connect_transaction()` that will register issuance. I'm not sure how this should be done, but you shouldn't make registering issuance a part of checking. Checking is by definition immutable.

In summary, when you connect a transaction, you check the token data, THEN check the total inputs/outputs, THEN register issuance. If you can find a cleaner way to do this, I don't mind. But in its current form, it's too complex.

Finally, all token-specific operations/functions should be moved to separate submodules.

TheQuantumPhysicist:
I'm wondering whether we should close this PR. I suggest we close this PR because it got really big with way too many comments. Then, once you're done and have addressed all the issues in it, create a new one. Let me know, guys, if this makes sense.

muursh (changes requested):
I'm with @TheQuantumPhysicist here. A clean PR seems like the way to go here. This is getting cluttered and hard to follow.

chainstate/src/detail/mod.rs
```diff
@@ -13,24 +13,19 @@
...
 mod median_time;
 mod orphan_blocks;
 mod transaction_verifier;
-
-use std::sync::Arc;
-
-use itertools::Itertools;
-
-use chainstate_storage::{BlockchainStorage, Transactional};
+pub use chainstate_types::Locator;
```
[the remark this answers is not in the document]
sinitcin: To be honest, I use `cargo-fmt` that sort `pub mod` and `pub use` automatically. 

@TheQuantumPhysicist do we need to stick to the Stanislav's rule?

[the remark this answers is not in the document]
TheQuantumPhysicist: cargo-fmt won't reorder such things for you. But it's a good-to-have to get things together. Why not.

chainstate/src/detail/tokens/mod.rs
sinitcin: Pretty nice catch! 

I fixed it, and found that burn tests are failed. After the short inspections, I figured out that test was contained issues. Fixed it too.

sinitcin:
Sadly, but make sense! Thank you for a review, guys. See you during the next attempt to merge it. 

@TheQuantumPhysicist @muursh @stanislav-tkach @azarovh

TheQuantumPhysicist:
> Sadly, but make sense! Thank you for a review, guys. See you during the next attempt to merge it.
> 
> @TheQuantumPhysicist @muursh @stanislav-tkach @azarovh

Please don't forget to address all the comments in the PR.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Depends on, will we have any tokens in genesis or not?

sinitcin: If we don't plan to have tokens in genesis, then we can simplify this check

chainstate/src/detail/mod.rs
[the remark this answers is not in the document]
sinitcin: Ok. No problem. Will stick to it.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: I ridded off from mut in all check functions

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: We can't have tokens on genesis because there are no inputs, meaning, there's no systematic way to calculate any token id.

chainstate/src/detail/spend_cache/mod.rs
sinitcin: Done

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Fixed, this check was removed according to Sam's comments.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Done

chainstate/src/detail/tokens/mod.rs
sinitcin: Done!

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Now all coin and tokens values checks in `check_transferred_amounts_and_get_fee`

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Now, token data checks in `check_block`. When we connect a transaction, we check transferred amounts of coin and tokens. Function `register_tokens` isolated from checks.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Fixed, removed unnecessary check.

PR #271, opened 2022-07-01
Fix new Clippy lints (Rust 1.62)

stanislav-tkach:
Actually, new lints aren't triggered for the code-base, but some of the old ones were improved, too.

### DIFF
--- a/common/src/uint/impls.rs
+++ b/common/src/uint/impls.rs
@@ -272,21 +272,21 @@ macro_rules! construct_uint {
             }
         }
 
-        impl<'a> From<u64> for $name {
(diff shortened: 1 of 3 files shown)


TheQuantumPhysicist (approved):
@altonen please check whether the error handling makes sense. I preferred the mild commented errors.

stanislav-tkach:
I also feel like the previous comments were clearer. On the other hand, personally I'm used to have `x.await??` in the code-base, so extensive commenting feels a little bit too much, especially if exactly the same comment is used in several places.

altonen (approved):
LGTM

PR #273, opened 2022-07-04
Create a verifiable random function for proof of stake

TheQuantumPhysicist:


crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[must_use]
+pub struct SchnorrkelPublicKey {
+    key: schnorrkel::PublicKey,
+}
+
+impl Encode for SchnorrkelPublicKey {
+    fn encode(&self) -> Vec<u8> {
+        self.key.to_bytes().to_vec()
+    }
```
iljakuklic, line 15: It's probably more appropriate to implement `using_encoded` instead of `encode` for types that can be cheaply converted to `&[u8]` encoded representation. That avoids allocating an extra `Vec` in many cases and you get efficient `encoded_size` for free.

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,158 @@
+use crate::random::{CryptoRng, Rng};
+use parity_scale_codec::{Decode, Encode};
```
iljakuklic, line 2: Is there an issue depending on `serialization` rather than `parity_scale_codec`? (Few other places below too)

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+
+const EXPECTED_PUBKEY_LEN: usize = 32;
```
iljakuklic, line 4: I'd personally just call this `PUBKEY_LEN`

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl SchnorrkelPrivateKey {
+    #[allow(dead_code)]
+    pub fn new<R: Rng + CryptoRng>(rng: &mut R) -> (SchnorrkelPrivateKey, SchnorrkelPublicKey) {
+        let sk = schnorrkel::SecretKey::generate_with(rng);
+        let pk = sk.to_public();
+        let sk = Self { key: sk };
+        let pk = SchnorrkelPublicKey { key: pk };
+        (sk, pk)
+    }
```
iljakuklic, line 53: It may be useful to expose `to_public` as a stand-alone function as well. Or are there reasons to not do it?

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+const EXPECTED_PRIVKEY_LEN: usize = 64;
+
+impl Encode for SchnorrkelPrivateKey {
+    fn encode(&self) -> Vec<u8> {
+        self.key.to_bytes().to_vec()
+    }
```
iljakuklic, line 61: Same comment as for `SchnorrkelPublicKey`

iljakuklic (commented):
A bunch of minor comments

crypto/src/vrf/schnorrkel/mod.rs
TheQuantumPhysicist: Alrighty

crypto/src/vrf/schnorrkel/mod.rs
TheQuantumPhysicist: Alrighty

crypto/src/vrf/schnorrkel/mod.rs
TheQuantumPhysicist: This is still incomplete

crypto/src/vrf/mod.rs
```diff
@@ -0,0 +1,117 @@
+use serialization::{Decode, Encode};
```
stanislav-tkach, line 1: The license header is missing. 🙃

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,185 @@
+use crate::random::{CryptoRng, Rng};
```
stanislav-tkach, line 1: Same here.

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+        &self,
+        message: Transcript,
+        vrf_data: &VRFReturn,
+    ) -> Result<(), VRFError> {
+        match vrf_data {
+            VRFReturn::Schnorrkel(d) => self.verify_vrf(message, d),
+        }
+    }
+
+    pub fn verify_vrf(
```
muursh: This name feels as generic as the "generic" one above in all honesty

TheQuantumPhysicist: I have very little confidence about the naming scheme... if you can think of a different naming scheme (and also the design), please go ahead.

muursh (approved):
Design looks fine, implementation looks fine. My only concern lays in the naming but I don't have better proposals for the few bits I don't like the names of and my concerns aren't strong. 

The key thing now is going to be actually using this and making sure we use it correctly and safely.

All in all a job well done :)

altonen (approved):

crypto/src/vrf/transcript.rs
```diff
@@ -0,0 +1,108 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TranscriptAssembler {
+    label: &'static [u8],
+    components: Vec<(&'static [u8], TranscriptComponent)>,
+}
+
+// A wrapper that makes it unnecessary to directly use the merlin dependency
+#[derive(Clone)]
+pub struct WrappedTranscript(Transcript);
```
iljakuklic, line 34: Maybe call this just `Transacript` and refer to the `Transcript` type from the upstream crate by the full name, i.e. `merlin::Transcript`. I'd prefer our code base to be free of filler words like `Wrapper`.

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    pub fn verify_generic_vrf_data(
+        &self,
+        message: Transcript,
+        vrf_data: &VRFReturn,
+    ) -> Result<(), VRFError> {
+        match vrf_data {
+            VRFReturn::Schnorrkel(d) => self.verify_vrf_data(message, d),
+        }
```
iljakuklic, line 52: What is this supposed to do if another enum variant is added to `VRFReturn`?

iljakuklic (approved):
No major objections, just a few comments on design and naming.

crypto/src/vrf/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+pub enum VRFKeyKind {
+    #[codec(index = 0)]
+    Schnorrkel,
+}
+
+#[must_use]
+#[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
+pub struct VRFPrivateKey {
+    key: VRFPrivateKeyHolder,
+}
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Because the variants are not public. The initialization should be done structly by key-type, not by value.

crypto/src/vrf/transcript.rs
```diff
@@ -0,0 +1,108 @@
...
+        Self {
+            label,
+            components: Vec::new(),
+        }
+    }
+
+    pub fn attach(self, label: &'static [u8], value: TranscriptComponent) -> Self {
+        let mut result = self;
+        result.components.push((label, value));
+        result
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yeah, I know... I don't like mut self though for my own (possibly stupid) biases... it doesn't make any sense. The value is being moved... why does it matter if it's gonna be changed in the interface? This is like in C++ doing a const string&&... that doesn't make sense. A moved value should always be considered mutable for a reader of an interface.

About the additional methods, I'd like to keep the touchable interface as small as possible to avoid having to test more. This is sensitive stuff.

crypto/src/vrf/transcript.rs
TheQuantumPhysicist: I've had a similar discussion with stanislav today about `Config` vs `ChainConfig` vs `ChainstateConfig`. The problem is that IDEs automatically add `use` clauses and we're gonna have to depend on the discretion of the programmer to make the reader not have two different Transcript or Config contexts in the same file in a readable way for symbol names that touch multiple crates. I'd rather be a little more verbose than ambiguous for the reader. After all, the rust code is written for humans, not for the machine. The machine couldn't care less.

crypto/src/vrf/schnorrkel/mod.rs
TheQuantumPhysicist: Just to be clear, are you proposing doing something specific? So far everything is covered because of match.

TheQuantumPhysicist:
I'll merge this now so that I can continue with PoS. If we have to change something, I'll create another PR.

crypto/src/vrf/schnorrkel/mod.rs
iljakuklic: I know you don't like matching on tuples and wildcards but I'd find something like the following pattern quite clear:
```rust
fn verify_vrf(vrf_pk: PublicKey, msg: Transcript, vrf_return: VRFReturn) -> Result<(), Error> {
  match (vrf_pk, vrf_return) {
    (PublicKey::Schnorrkel(pk), VRFReturn::Schnorrkel(ret)) => pk.verify_vrf_data(msg, ret),
    (PublicKey::SomeOther(pk), VRFReturn::SomeOther(ret)) => pk.verify_vrf_data(msg, ret),
    (pk, ret) => {
      assert_ne!(pk.kind(), ret.kind(), "VRF kind {:?} not covered", pk.kind());
      Err(Error::VRFKindMismatch(pk.kind(), ret.kind()))
    }
  }
}
```
(`vrf_pk` could be `self` to make it a method on PublicKey)

crypto/src/vrf/transcript.rs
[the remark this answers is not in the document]
iljakuklic: > Yeah, I know... I don't like mut self though for my own (possibly stupid) biases... it doesn't make any sense. The value is being moved... why does it matter if it's gonna be changed in the interface?

I see. I don't like that `mut` shows up in code like this. It is not, however, considered part of the interface. The tooling understand this: if you generate rustdoc or ask IDE for type of the method, it does not display the `mut` annotation.

crypto/src/vrf/schnorrkel/mod.rs
TheQuantumPhysicist: > I know you don't like matching on tuples and wildcards

How well you know me 😄 

I get the appeal, but I would rather have the compiler yell at me if I add a new type and forget to cover it instead of having this deceiving beauty that can lead to bugs in the future.

crypto/src/vrf/transcript.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I really don't know which is worse... I have to think more about this. Though for what it's worth, I like that tooling understands the distinction... unlike in C++ where gmock wouldn't compile at times if you pass by const value in an interface.

PR #276, opened 2022-07-05
Add the copyright header everywhere

stanislav-tkach:
Copyright headers are missing in some files. I have used a simple tool (built into the IDE) to add them, so the result isn't perfect: the year is fixed, so the date of file creation isn't taken into account and the authors are missing. I can probably add that information manually if it is really needed.


TheQuantumPhysicist (approved):
I hate that we have to do this... if I were to make the decision, I'd just remove them all.

stanislav-tkach:
I also think that is somewhat useless, but I understand that we probably need to have these headers because of legal reasons. 🙃

muursh:
It was requested by legal early on

muursh (approved):

PR #277, opened 2022-07-05
Raw prototype of subsystems configuration

stanislav-tkach:
Work in progress...

### DIFF
--- a/chainstate/src/config.rs
+++ b/chainstate/src/config.rs
@@ -1,0 +1,20 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://spdx.org/licenses/MIT
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::Args;
+
+/// The chainstate subsystem configuration.
+#[derive(Args, Debug)]
+pub struct Config {}
--- a/node/src/options.rs
+++ b/node/src/options.rs
@@ -15,9 +15,8 @@
 
 //! Node configuration options
 
-use std::ffi::OsString;
-use std::net::SocketAddr;
-use std::path::PathBuf;
+use std::{ffi::OsString, path::PathBuf};
+
 use strum::VariantNames;
 
 use common::chain::config::ChainType;
@@ -30,17 +29,18 @@ pub struct Options {
     #[clap(long, value_name = "PATH")]
     pub log_path: Option<PathBuf>,
 
-    /// Address to bind RPC to
-    #[clap(long, value_name = "ADDR", default_value = "127.0.0.1:3030")]
-    pub rpc_addr: SocketAddr,
-
     /// Blockchain type
     #[clap(long, possible_values = ChainType::VARIANTS, default_value = "mainnet")]
     pub net: ChainType,
 
-    /// Address to bind P2P to
-    #[clap(long, value_name = "ADDR", default_value = "/ip6/::1/tcp/3031")]
-    pub p2p_addr: String,
+    #[clap(flatten)]
+    pub chainstate_config: chainstate::Config,
+
+    #[clap(flatten)]
+    pub p2p_config: p2p::Config,
+
+    #[clap(flatten)]
+    pub rpc_config: rpc::Config,
 }
 
 impl Options {
--- a/p2p/src/config.rs
+++ b/p2p/src/config.rs
@@ -1,0 +1,24 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://spdx.org/licenses/MIT
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::Args;
+
+/// The p2p subsystem configuration.
+#[derive(Args, Debug)]
+pub struct Config {
+    /// Address to bind P2P to.
+    #[clap(long, value_name = "ADDR", default_value = "/ip6/::1/tcp/3031")]
+    pub p2p_addr: String,
+}
--- a/rpc/src/config.rs
+++ b/rpc/src/config.rs
@@ -1,0 +1,26 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://spdx.org/licenses/MIT
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::net::SocketAddr;
+
+use clap::Args;
+
+/// The rpc subsystem configuration.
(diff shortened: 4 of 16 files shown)


p2p/src/config.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use common::{chain::config::ChainType, primitives::semver::SemVer};
+
+/// The p2p subsystem configuration.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Config {
+    /// Address to bind P2P to.
+    pub address: String,
+    pub version: SemVer,
+    pub magic_bytes: [u8; 4],
```
altonen, line 26: This is still a work in progress so feel free to ignore this comment but I'm not sure if this is correct. I believe the idea is to have `SemVer` as some kind of constant so user cannot specify it on the command line and same goes for the magic bytes, I don't think it's supposed to be controlled via command line.

Even if they were configurable, I'd still think chainstate configuration would be the right place and not P2P.

p2p/src/config.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl Config {
+    /// Creates a new p2p configuration instance.
+    pub fn new(net: ChainType) -> Self {
+        Self {
+            address: "/ip/::1/tcp/3031".into(),
```
altonen, line 33: I believe it should be `/ip6/`

p2p/src/config.rs
stanislav-tkach: Thank you for pointing this out! I'm almost sure you are right. I think I was too eager to move everything in the configuration.

stanislav-tkach:
I'm still not sure about what approach would be better here: split the current `ChainConfig` structure or not, but here is the overview of the current state:
- The node now takes two sub-commands: `create-config` and `run`.
- Subsystems need both `ChainConfig` and a config specific to that subsystem for initialization.
- The version becomes constant.

p2p/src/config.rs
```diff
@@ -0,0 +1,38 @@
...
+use serde::{Deserialize, Serialize};
+
+/// The p2p subsystem configuration.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct P2pConfig {
+    /// Address to bind P2P to.
+    pub address: String,
+    /// The score threshold after which a peer is baned.
+    pub ban_threshold: u32,
+    /// The timeout value in seconds.
```
TheQuantumPhysicist, line 25: Timeout of what? Connection timeout for a peer? If yes, then probably `peer_conn_timeout` is better. Let's try to avoid generic words that can mean many things.

TheQuantumPhysicist (commented):
The PR is OK. There are two things that I feel can be improved:
1. There's no configuration reading for ChainConfig. There has to be a way to configure ChainConfig itself for functional tests. But this can be made in another PR.
2. The initialization of the config can be improved. All these `if let`s are very ugly. One solution is to create a wrapper type with a generic trait for a NodeConfigEntry. Then you implement that for all types that can be taken from config (ints, strings, etc). Finally, you can make a method there to load values from `Option<T>` if a value  exists. Maybe you can come up with something else that's simpler.

chainstate/src/config.rs
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serde::{Deserialize, Serialize};
+
+/// The chainstate subsystem configuration.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ChainstateConfig {
+    /// The number of maximum attempts to process a block.
+    pub max_db_commit_attempts: usize,
```
muursh, line 22: I'm not convinced commit is the right nomenclature here

muursh: I know it's keeping the old name but still not a fan

stanislav-tkach: It was called slightly differently before: `MAX_DB_COMMIT_COUNT` and I thought that I was improved this name enough. 😅 Would `max_attempts_to_process_block` be better?

stanislav-tkach:
@TheQuantumPhysicist 
1. We have a separate test-node, so I see no issues in reading the chain configuration only in functional tests. And yes, I would prefer to do that in a separate pull request.
2. Ideally it can be fixed in `clap` itself (https://github.com/clap-rs/clap/issues/748 and https://github.com/clap-rs/clap/issues/3113), but I don't expect that it will be done in the nearest future. I don't like `if let` boilerplate either mostly because it can easily lead to an error (reading a wrong value or forgetting to override one of the options). I'm going to think about a better solution.

chainstate/src/config.rs
TheQuantumPhysicist: I'm not sure whether you have an alternative word in mind, Ben, but "commit" is the word used in the context of ACID compliant databases. For example, see the option "synchronous_commit" in postgresql: https://postgresqlco.nf/doc/en/param/synchronous_commit/

stanislav-tkach: Discussed: this name will do for now.

rpc/src/config.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::{net::SocketAddr, str::FromStr};
+
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+/// The rpc subsystem configuration.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RpcConfig {
+    /// Address to bind RPC to.
+    pub address: SocketAddr,
```
TheQuantumPhysicist, line 25: bind_address

node/src/config.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Reads a configuration from the path specified in options and overrides the provided
+    /// parameters.
+    pub fn read(options: &RunOptions) -> Result<Self> {
+        let config = fs::read_to_string(&options.config_path)
+            .with_context(|| format!("Failed to read '{:?}' config", options.config_path))?;
+        let NodeConfig {
+            chainstate,
+            p2p,
+            rpc,
+        } = toml::from_str(&config).context("Failed to deserialize config")?;
```
TheQuantumPhysicist, line 60: I would not call it deserialize. Perhaps "parse" is a better word.

p2p/src/config.rs
```diff
@@ -0,0 +1,38 @@
...
+// limitations under the License.
+
+use serde::{Deserialize, Serialize};
+
+/// The p2p subsystem configuration.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct P2pConfig {
+    /// Address to bind P2P to.
+    pub bind_address: String,
+    /// The score threshold after which a peer is baned.
```
TheQuantumPhysicist, line 23: banned

TheQuantumPhysicist (approved):
I think this looks good now, besides the minor comments I mentioned.

altonen (approved):

chainstate/src/config.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// The maximum capacity of the orphan blocks pool.
+    pub max_orphan_blocks: usize,
```
iljakuklic, line 24: (Not for this PR.) Maybe there should be a limit on the total size of orphan blocks in bytes in addition to this.

chainstate/src/detail/block_index_history_iter.rs
```diff
@@ -75,17 +75,24 @@ mod tests {
...
-    use crate::{BlockSource, Chainstate};
+    use crate::{BlockSource, Chainstate, ChainstateConfig};
 
     use super::*;
 
     #[test]
     fn history_iteration() {
         common::concurrency::model(|| {
             let chain_config = Arc::new(create_unit_test_config());
+            let chainstte_config = ChainstateConfig::new();
```
iljakuklic, line 86: Typo: `chainstte` -> `chainstate`

iljakuklic (approved):
Looks good. Rather unfortunate there are separate mechanisms for dealing with options coming from a config file and options coming from the command line with this manual translation, but given that, the solution presented here looks rather nice.

node/src/options.rs
```diff
@@ -13,34 +13,75 @@
...
-    /// Address to bind P2P to
-    #[clap(long, value_name = "ADDR", default_value = "/ip6/::1/tcp/3031")]
-    pub p2p_addr: String,
+    /// Blockchain type.
+    #[clap(long)]
+    pub max_db_commit_attempts: Option<usize>,
+
+    /// The maximum capacity of the orphan blocks pool.
+    #[clap(long)]
+    pub max_orphan_blocks: Option<usize>,
```
iljakuklic, line 68: The comment will show up in the help message. It should clarify the capacity is specified in the number of blocks. People may think it's a limit on the number of bytes or something like that and open themselves up to a DoS attack.

chainstate/src/config.rs
TheQuantumPhysicist: The total size should come as a consequence of the total number since the size of a single block is capped already.

node/src/options.rs
stanislav-tkach: I agree that a documentation, especially that is directly user-facing should be as clear as possible.

And thank you for pointing out the typos too!

PR #278, opened 2022-07-06
Remove unused dependencies

stanislav-tkach:
I intended to add the unused dependencies check on the CI, but it gives false positives for the `node` crate because it contains both `lib.rs` and `main.rs`.


TheQuantumPhysicist (approved):
I don't mind doing things like this, but this isn't a priority right now.

altonen (approved):

PR #281, opened 2022-07-08
Fix the get_headers_genesis test

stanislav-tkach:
A fix of the regression introduced in the https://github.com/mintlayer/mintlayer-core/pull/266 pull request.


iljakuklic:
Why was it failing before?

stanislav-tkach:
@iljakuklic Previously the difference between chain lengths was bigger. I suppose that after the refactoring chains can have a very similar length, so locators had the same length sometimes.

iljakuklic (approved):

PR #282, opened 2022-07-10
sync: Add timeouts for request/response tests

altonen:
The test is a little unreliable on Windows but it's really infrequent and thus hard to debug so these changes will hopefully reveal something about the issue.

### DIFF
--- a/p2p/src/sync/tests/request_response.rs
+++ b/p2p/src/sync/tests/request_response.rs
@@ -16,7 +16,8 @@
 // Author(s): A. Altonen
 use super::*;
 use crate::message::*;
-use std::collections::HashSet;
+use std::{collections::HashSet, time::Duration};
+use tokio::time::timeout;
 
 #[tokio::test]
 async fn test_request_response() {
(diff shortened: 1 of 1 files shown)


stanislav-tkach (approved):

p2p/src/sync/tests/request_response.rs
```diff
@@ -90,34 +90,36 @@ async fn test_multiple_requests_and_responses() {
...
+                            Response::HeaderResponse(HeaderResponse::new(vec![])),
+                        )
+                        .await
+                        .unwrap();
+                }
+                _ => panic!("invalid event received {:?}, expected `Request`", event),
+            },
+            _ = tokio::time::sleep(std::time::Duration::from_secs(15)) => {
+                panic!("did not receive `Request in time`");
+            }
```
iljakuklic, line 108: It seems to me `tokio::time::timeout` could be used in place of this select pattern. Would that be an improvement?

iljakuklic (approved):

p2p/src/sync/tests/request_response.rs
altonen: It would be, thanks!

muursh (approved):

PR #283, opened 2022-07-11
Implement peer score

altonen:


p2p/src/net/libp2p/connectivity/tests.rs
```diff
@@ -430,3 +430,26 @@ async fn close_connection() {
...
 }
+
+#[tokio::test]
+async fn banned_peer() {
+    let mut connmgr = ConnectionManager::new();
+
+    // known connection
+    let peer_id = PeerId::random();
+
+    connmgr.connections.insert(
```
TheQuantumPhysicist, line 441: What makes this peer banned in this test? Where is this set?

p2p/src/net/libp2p/backend.rs
```diff
@@ -73,6 +73,9 @@ impl Backend {
                     SwarmEvent::NewListenAddr { address, .. } => {
                         log::trace!("new listen address {:?}", address);
                     }
+                    SwarmEvent::BannedPeer { peer_id, .. } => {
```
TheQuantumPhysicist, line 76: Please avoid using `..`. This is for the same reason why we don't use catch-all `_` in match. You *want* the compiler to complain at every point these structs/arms are used if you change something about them. This will basically silence the compiler.

TheQuantumPhysicist (commented):
Please change behaviour to behavior. We have to stick to one kind of English in the source code, which was earlier decided to be American English.

p2p/src/net/libp2p/mod.rs
```diff
@@ -368,6 +368,19 @@ where
         &self.peer_id
     }
 
+    async fn ban_peer(&mut self, peer_id: T::PeerId) -> crate::Result<()> {
+        log::debug!("ban peer {}", peer_id);
```
TheQuantumPhysicist, line 372: Perhaps it's helpful to print more information. At least the ip address and the reason for banning. The peer_id is meaningless without context.

altonen:
> Please change behaviour to behavior. We have to stick to one kind of English in the source code, which was earlier decided to be American English.

We've already talked about this in a daily. Libp2p uses the word behaviour so back then it was decided that we use behaviour in p2p even though it's British

p2p/src/net/types.rs
```diff
@@ -142,13 +142,13 @@ pub enum ConnectivityEvent<T: NetworkingService> {
         error: error::P2pError,
     },
 
-    /// Peer misbehaved
-    Misbehaved {
+    /// Protocol violation
+    ProtocolViolation {
```
TheQuantumPhysicist, line 146: Is there a reason for renaming Misbehaved to ProtocolViolation? I'm not a big fan of this change given that it's well-known what misbehaving means in bitcoin.

p2p/src/swarm/mod.rs
```diff
@@ -253,6 +266,27 @@ where
...
+        log::debug!("adjusting score for peer {}, adjustment {}", peer_id, score);
+
+        let score = if let Some(entry) = self.peers.get_mut(&peer_id) {
+            entry.score = entry.score.saturating_add(score);
+            entry.score
+        } else {
+            score
+        };
+
+        if score >= 100 {
```
TheQuantumPhysicist, line 282: This should come from a configuration parameter, not a constant.

TheQuantumPhysicist (approved):
Besides the few comments, I think this looks good.

stanislav-tkach (approved):

p2p/src/net/libp2p/connectivity/tests.rs
altonen: This is just testing how `ConnectionManager` handles banned peers. It's relevant to it what the peer did to get banned.

p2p/src/net/libp2p/backend.rs
altonen: Fixed, although in this case the front-end code doesn't need extra information in order to ban the peer so I don't really see the harm in ignoring rest of the fields.

p2p/src/swarm/mod.rs
altonen: I'll add a todo

p2p/src/net/libp2p/mod.rs
altonen: I'll log that information in `PeerManager`

PR #286, opened 2022-07-13
swarm: Close connections which are rejected by the `PeerManager`

altonen:


stanislav-tkach (approved):

muursh (approved):

p2p/src/swarm/mod.rs
```diff
@@ -472,10 +472,23 @@ where
...
-                            // TODO: report rejection to networking backend
                             let peer_id = peer_info.peer_id;
-                            let res = self.accept_inbound_connection(addr, peer_info).await;
-                            self.handle_result(Some(peer_id), res).await?;
+
+                            match self.accept_inbound_connection(addr, peer_info).await {
+                                Ok(_) => {},
+                                Err(P2pError::ChannelClosed) => return Err(P2pError::ChannelClosed),
+                                Err(P2pError::PeerError(err)) => {
+                                    log::warn!("peer error for peer {}: {}", peer_id, err);
```
TheQuantumPhysicist, line 481: Again, same issue with logging information. Just the peer_id alone without context is meaningless; at least the address of the peer should appear as well. I think this issue is spread all over p2p.

TheQuantumPhysicist (approved):

p2p/src/swarm/mod.rs
altonen: The `PeerManager`/`PeerDb` implementations are still under work but were not critical in terms of getting the P2P working so printing proper peer information when the connection closes hasn't been implemented yet. I can work on that after this PR and fixing the mDNS issues

PR #288, opened 2022-07-16
Add timelock to output purpose

TheQuantumPhysicist:
For staking (and also vesting) we need to be able to make outputs unspendable for a given period of time. This similar to bitcoin timelocks, but it's not implemented at the script level.


common/src/chain/transaction/output/timelock.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::{chain::block::timestamp::BlockTimestamp, primitives::BlockHeight};
+
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Encode, Decode)]
+pub enum OutputTimeLock {
+    #[codec(index = 0)]
+    UntilHeight(BlockHeight),
+    #[codec(index = 1)]
+    UntilTime(BlockTimestamp),
+    #[codec(index = 2)]
+    ForBlockCount(#[codec(compact)] u64),
```
iljakuklic, line 12: How about making this the `BlockDistance` type instead of `u64`? Semantically it seems a good match but there might be an issue with negative distances. Maybe `BlockDistance` should be unsigned after all, since it is used in a way where we know which height is supposed to come first.

common/src/chain/transaction/output/timelock.rs
```diff
@@ -0,0 +1,15 @@
+use serialization::{Decode, Encode};
+
+use crate::{chain::block::timestamp::BlockTimestamp, primitives::BlockHeight};
+
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Encode, Decode)]
+pub enum OutputTimeLock {
```
iljakuklic, line 6: Why is this `OutputTimeLock`? Any other time lock kinds planned? It seems to me that it could be useful in script-enforced input time locks too.

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -308,7 +311,52 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
+            }
+            common::chain::timelock::OutputTimeLock::ForSeconds(dt) => {
+                spending_time.as_duration_since_epoch()
+                    >= source_block_time.as_duration_since_epoch() + Duration::from_secs(*dt)
+            }
+        };
+
+        if !past_lock {
+            return Err(StateUpdateError::TimelockViolation);
+        }
```
iljakuklic, line 349: This could be
```rust
ensure!(past_lock, StateUpdateError::TimelockViolation);
```

iljakuklic (commented):
Added some comments.

In addition to that, I think it would be nice to see this mechanism used for block reward maturity too. Just let consensus rules enforce that the reward outputs have the appropriate time lock on them. It seems like a nice design to have the same mechanism deal with all these time-locky things. However this would deviate from Bitcoin somewhat so may not be exactly what the devs expect.

common/src/chain/transaction/output/timelock.rs
TheQuantumPhysicist: Funny enough, if you look into the history, you'll see that it was BlockDistance. The issue though is that compact encoding isn't supported for signed types. In the big picture, this isn't a concern now. We can beautify this later since there's a technical issue behind it.

chainstate/src/detail/spend_cache/mod.rs
TheQuantumPhysicist: Will do.

common/src/chain/transaction/output/timelock.rs
TheQuantumPhysicist: I'm personally not familiar with all timelock types, which is why I asked for your review. As far as I know, the problem with script timelocks is that they're bound to the integer time lock in the transaction... that's incompatible with this enum. Honestly I don't know how to reconcile all those. You're more familiar with this. If this is conclusively determined and doesn't conflict with anything, I'll be happy to change it to just Timelock. If my choice is not crazy (in the sense that we can update it in the near future), we can keep it. If there's an obvious mistake/conflict, let me know. I'm not sure after all.

TheQuantumPhysicist:
> Added some comments.
> 
> In addition to that, I think it would be nice to see this mechanism used for block reward maturity too. Just let consensus rules enforce that the reward outputs have the appropriate time lock on them. It seems like a nice design to have the same mechanism deal with all these time-locky things. However this would deviate from Bitcoin somewhat so may not be exactly what the devs expect.

It's part of the plan. All in good time. We'll have to asses all options once PoS works.

common/src/chain/transaction/output/timelock.rs
iljakuklic: The issue with script-based time locks it that the absolute time lock is enforced via transaction-level time lock field and relative time lock is enforced via input-level time lock field (nSequence). If we follow that design, this `OutputTimeLock` type would not work so neatly, so let's leave it as it is at the moment and have a more in-depth discussion at some point.

iljakuklic (approved):

chainstate/src/detail/tests/output_timelock.rs
```diff
@@ -0,0 +1,1226 @@
...
+    TxInput::new(
+        OutPointSourceId::Transaction(block.transactions().get(0).unwrap().get_id()),
+        1,
+        InputWitness::NoSignature(None),
+    )
+}
+
+fn median_block_time(times: &[u64]) -> u64 {
+    // Only the last 11 blocks are used for calculating the median time.
+    *times.iter().take(11).nth(times.len() / 2).unwrap()
```
TheQuantumPhysicist, line 1225: Why not use the median time function that already calculates this?

stanislav-tkach:
I have merged master to this branch. Everything vent smoothly and I only needed to fix tests after that, but please take a look anyway.

chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -363,6 +414,24 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
+                        let block_index = self
+                            .db_tx
+                            .get_block_index(tx_pos.block_id())
+                            .map_err(StateUpdateError::from)?
+                            .ok_or_else(|| {
+                                StateUpdateError::InvariantErrorHeaderCouldNotBeLoaded(
+                                    *tx_pos.block_id(),
+                                )
+                            })?;
+                        self.check_timelock(
```
altonen, line 427: I think the name `verify_signatures()` is a little confusing because I wouldn't start looking for the time lock check from that function.

altonen (approved):

chainstate/src/detail/spend_cache/mod.rs
TheQuantumPhysicist: I've been thinking about that for a while... I came up with the name `verify_authorization`... not sure yet. But I'm not too worried now about that. Renaming stuff is easy.

iljakuklic (approved):
Looks good

p2p/Cargo.toml
```diff
@@ -17,6 +17,7 @@ serde = { version = "1", features = ["derive"] }
 # local dependencies
 common = { path = "../common/" }
 chainstate = { path = "../chainstate/" }
+chainstate-types = { path = "../chainstate-types/" }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Nope.

PR #290, opened 2022-07-18
Add mDNS configuration

altonen:


node/src/options.rs
```diff
@@ -79,6 +79,14 @@ pub struct RunOptions {
     #[clap(long)]
     pub p2p_outbound_connection_timeout: Option<u64>,
 
+    /// Enable multicast DNS peer discovery
+    #[clap(long)]
+    pub enable_mdns: Option<bool>,
```
stanislav-tkach, line 84: I would rather use `p2p_enable_mdns` here. It isn't ideal, but as the command line options aren't hierarchical it makes sense to use prefixes for subsystems-related options.

node/src/options.rs
```diff
@@ -79,6 +79,14 @@ pub struct RunOptions {
...
+
+    /// Interval (in milliseconds) at which to poll the network for new peers.
+    #[clap(long)]
+    pub mdns_query_interval: Option<u64>,
```
stanislav-tkach, line 88: Same here.

stanislav-tkach (approved):

p2p/src/net/libp2p/discovery/mdns.rs
```diff
@@ -38,9 +38,15 @@ pub enum Mdns {
 }
 
 impl Mdns {
-    pub async fn new(enabled: bool) -> Self {
+    pub async fn new(enabled: bool, query_interval: u64) -> Self {
```
TheQuantumPhysicist, line 41: It's probably better to pass the argument here as a Duration, so that no ambiguity about units happen. I understand that the argument is a number, but the argument has to identify the unit, and then there should be a proper conversion there to this function as a Duration. That's probably the cleanest solution to time issues.

p2p/src/net/libp2p/discovery/mdns.rs
```diff
@@ -38,9 +38,15 @@ pub enum Mdns {
...
         if enabled {
-            match mdns::Mdns::new(Default::default()).await {
+            match mdns::Mdns::new(mdns::MdnsConfig {
+                ttl: Default::default(),
+                query_interval: std::time::Duration::from_millis(query_interval),
+                enable_ipv6: Default::default(),
```
TheQuantumPhysicist, line 46: IPv6 should also be optional

p2p/src/net/libp2p/discovery/mdns.rs
```diff
@@ -96,7 +102,7 @@ mod tests {
 
     #[tokio::test]
     async fn mdns_disabled() {
-        let mdns = Mdns::new(false).await;
+        let mdns = Mdns::new(false, 0).await;
```
TheQuantumPhysicist, line 105: What does 0 mean? If I understand correctly, 0 has no effect because mdns is disabled. I would rather then see an `Option<Duration>` or `Option<Settings>` instead of values with no effect.

p2p/src/net/libp2p/tests/frontend.rs
```diff
@@ -307,9 +291,11 @@ async fn test_connect_with_timeout() {
     let config = Arc::new(common::chain::config::create_mainnet());
     let (mut service, _, _) = Libp2pService::start(
         test_utils::make_libp2p_addr(),
-        &[],
         config,
-        Duration::from_secs(2),
+        Arc::new(config::P2pConfig {
+            outbound_connection_timeout: 2,
+            ..Default::default()
```
TheQuantumPhysicist, line 297: No `..` please.

p2p/src/net/libp2p/discovery/mdns.rs
TheQuantumPhysicist: Btw, a separate enum would be better than Option.

TheQuantumPhysicist (approved):
Looks good except for the minor stuff I mentioned.

node/src/options.rs
altonen: I was debating between whether the p2p prefix should or should not be here. It's really inconvenient if it has to be added for each flag even if there will be no conflict, just like ike in this case. For example, `max_orphan_blocks` is unambiguous just like `mdns_query_interval` whereas `addr` isnt.

p2p/src/net/libp2p/discovery/mdns.rs
altonen: `std::time::Duration` will convey as little information about the units used when reading the code but now the conversion is not happening here but at the call site (or probably when the argument is parsed).

p2p/src/net/libp2p/tests/frontend.rs
altonen: these are tests

p2p/src/net/libp2p/tests/mdns.rs
```diff
@@ -27,17 +30,25 @@ use test_utils::make_libp2p_addr;
 async fn test_discovered_and_expired() {
     let (mut backend1, _, _conn_rx, _, _) = make_libp2p(
         common::chain::config::create_mainnet(),
+        Arc::new(config::P2pConfig {
+            enable_mdns: true,
+            mdns_query_interval: 200,
+            ..Default::default()
```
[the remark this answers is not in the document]
altonen: these too

p2p/src/net/libp2p/discovery/mdns.rs
altonen: 0 is what `Default::default()` gives for `std::time::Duration` which is no timeout. I want to preserve the default behavior of libp2p's mDNS so I need to think about this.

p2p/src/net/libp2p/tests/frontend.rs
TheQuantumPhysicist: I know they're tests... but still not very happy about them. If you insist on keeping them like this, I'm fine with that. But I think the right thing to do is not use `..` so that you even check the tests when a change is done.

p2p/src/net/libp2p/discovery/mdns.rs
altonen: It now accepts an `Option` for the query interval but I'm not sure if it's any better if I'm honest

p2p/src/net/libp2p/discovery/mdns.rs
```diff
@@ -38,9 +39,15 @@ pub enum Mdns {
 }
 
 impl Mdns {
-    pub async fn new(enabled: bool) -> Self {
-        if enabled {
-            match mdns::Mdns::new(Default::default()).await {
+    pub async fn new(config: &config::MdnsConfig) -> Self {
+        if config.enable_mdns {
```
TheQuantumPhysicist, line 43: This can still be improved.

How about an enum that looks like this?

```
enum Mdns {
    Enabled(MdnsConfig),
    Disabled,
}
```

p2p/src/net/libp2p/discovery/mdns.rs
altonen: I think I don't understand your proposal for fixing this, could you explain it in more detail? Just changing the parameter type doesn't fix the issue of the unit being ambiguous as I explained in my previous comment so I'm not entirely sure how you want this information displayed.

p2p/src/net/libp2p/discovery/mdns.rs
altonen: It doesn't make sense to me, how would you implement this? The reason why it's implemented like it is right now is that it's transparent to the `DiscoveryManager`, meaning the interface for this discovery behavior and for all others is the same: they all implement `poll()` which the `DiscoveryManager` calls irrespective of whether the discovery is actually enabled or not. That information is relayed to the the specified behavior via the config and it internally sets its state as either `Enabled` or `Disabled`.

I'm trying to imagine the implementation for what you're proposing and at least I am having trouble seeing a clean solution.

node/src/options.rs
stanislav-tkach: Personally I prefer consistent naming, but I'm not going to insist. 🙃

altonen:
Enums serialized quite nicely, let me know what you think

p2p/src/config.rs
```diff
@@ -15,6 +15,49 @@
...
+        enable_ipv6: Option<bool>,
+    ) -> Self {
+        if enable_mdns {
+            MdnsConfig::Enabled {
+                query_interval: query_interval.unwrap_or(0),
+                enable_ipv6: enable_ipv6.unwrap_or(false),
+            }
+        } else {
+            assert!(
+                query_interval.is_none(),
```
TheQuantumPhysicist, line 48: I wonder whether this can be done through a macro on the struct/enum somehow. Maybe throw in a TODO there describing that these checks shouldn't be done manually.

TheQuantumPhysicist:
> Enums serialized quite nicely, let me know what you think

Looking good! I'm glad we tried.

node/src/options.rs
stanislav-tkach: Sorry for bringing this up again, but as far as I can see, the options are sorted alphabetically. So without prefixes the order will be somewhat confusing.

p2p/src/config.rs
```diff
@@ -15,6 +15,50 @@
...
+    }
+
+    pub fn from_options(
+        enable_mdns: bool,
+        query_interval: Option<u64>,
+        enable_ipv6_mdns_discovery: Option<bool>,
+    ) -> Self {
+        if enable_mdns {
+            MdnsConfig::Enabled {
+                query_interval: query_interval.unwrap_or(0),
```
TheQuantumPhysicist, line 43: Can you please put the default values of the configuration somewhere isolated, instead of directly injecting them to options?

test/functional/test_framework/.mintlayer/config.toml
```diff
@@ -9,5 +9,8 @@ bind_address = "/ip6/::1/tcp/3031"
 ban_threshold = 100
 outbound_connection_timeout = 10
 
+[p2p.mdns_config]
+type = "Disabled"
```
TheQuantumPhysicist, line 13: Is there a way to change the `type` to something else? I wonder. I think that's the default for enums, but there probably is a way to name it whatever we want.

stanislav-tkach: Perhaps `serde(untagged)` can allow to write something like `p2p.mdns_config = Disabled`.

node/src/options.rs
altonen: I'm not sure what's the best way to fix them. I change the order of these new options to be in the correct place, that is at least what Substrate seems to be doing but I take it you'd rather have the p2p prefix?

test/functional/test_framework/.mintlayer/config.toml
altonen: `type` can be changed to something else but at least I couldn't get `untagged` to work after playing with it briefly.

altonen: Renamed it to `state`

node/src/options.rs
stanislav-tkach: Perhaps I'm missing something, but it looks like the order of the filed in a struct isn't important and clap outputs them in the alphabetical order. So yes, I'm talking about the prefix again. 🙃 Maybe I'm exaggerating this issue, but for me it seems better to have some structure here, so a user can easily see which options are related to some subsystem. Still it is just my opinion and I'm not going to insist (too much).

altonen: I see your point and I like consistency too but for me it feels really awkward to type `--p2p-mdns-query-interval`.

Maybe @TheQuantumPhysicist or @muursh can make the decision since we're in a stalemate :smiley:

TheQuantumPhysicist: We either use a prefix for everything or for nothing. Half and half is confusing.

stanislav-tkach: I think that we are forced to used prefix for some parameters that have generic name such as "address" .

altonen: But `max_db_commit_attempts` and `max_orphan_blocks` don't have a `chainstate_` prefix. IMO P2P/RPC addresses are a special case which shouldn't dictate the entire naming scheme for command line parameters

stanislav-tkach: > But `max_db_commit_attempts` and `max_orphan_blocks` don't have a chainstate_ prefix.

It is my oversight. 🙃

iljakuklic: Nobody asked for my opinion but I'm going to throw it in anyway :laughing:. Whatever prefixes end up being there, I personally don't think they need or should be based on subsystem. How is the node broken down into subsystems internally I prefer to consider an implementation detail. Although we already use subsystem-based prefixes in RPC and config file too I believe.

p2p/src/config.rs
```diff
@@ -15,6 +15,54 @@
 
 use serde::{Deserialize, Serialize};
 
+pub const MDNS_DEFAULT_QUERY_INTERVAL: u64 = 0;
+pub const MDNS_IPV6_DEFAULT_IPV6_STATE: bool = false;
```
muursh, line 19: Do we really need `IPV6` twice?

p2p/src/config.rs
```diff
@@ -15,6 +15,54 @@
...
+    }
+
+    pub fn from_options(
+        enable_mdns: bool,
+        query_interval: Option<u64>,
+        enable_ipv6_mdns_discovery: Option<bool>,
+    ) -> Self {
+        if enable_mdns {
+            MdnsConfig::Enabled {
+                query_interval: query_interval.unwrap_or(MDNS_DEFAULT_QUERY_INTERVAL),
```
muursh, line 46: If mDNS is enabled the default query time is 0?

muursh (approved):
No real complaints from me

p2p/src/config.rs
altonen: Yes, it comes from libp2p

PR #291, opened 2022-07-18
Replace path to the config by the data-dir command line option

stanislav-tkach:


node/src/options.rs
```diff
@@ -85,7 +88,31 @@ pub struct RunOptions {
 }
 
 impl Options {
-    pub fn from_args<A: Into<OsString> + Clone>(args: impl IntoIterator<Item = A>) -> Self {
-        clap::Parser::parse_from(args)
+    /// Constructs an instance by parsing the given arguments.
+    ///
+    /// The data directory is created as a side-effect of the invocation.
+    pub fn from_args<A: Into<OsString> + Clone>(args: impl IntoIterator<Item = A>) -> Result<Self> {
+        let mut options: Options = clap::Parser::parse_from(args);
```
stanislav-tkach, line 95: @TheQuantumPhysicist Should I avoid mutable variables in such cases too?

node/src/options.rs
```diff
@@ -15,13 +15,23 @@
...
+};
+
+use anyhow::{anyhow, Context, Result};
 use clap::{Args, Parser, Subcommand};
+use directories::UserDirs;
 use strum::VariantNames;
 
 use common::chain::config::ChainType;
 
+const DEFAULT_DATA_DIR: &str = "$HOME/.mintlayer";
```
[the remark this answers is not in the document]
stanislav-tkach: I have worked this out by using the `default_value_os_t` attribute.

node/src/options.rs
stanislav-tkach: This is no longer relevant. 🙃

stanislav-tkach:
I have created a separate issue for a more thorough testing (https://github.com/mintlayer/mintlayer-core/issues/292).

iljakuklic (approved):
Looks good code-wise. Wondering if it would still make sense to have the config file and the data dir separate. Config file is written by humans and is read by the node, other datadir files would be written by the node, and there are often separate standard locations for these two use cases, e.g. `XDG_DATA_HOME` vs. `XDG_CONFIG_HOME`.

TheQuantumPhysicist:
> Looks good code-wise. Wondering if it would still make sense to have the config file and the data dir separate. Config file is written by humans and is read by the node, other datadir files would be written by the node, and there are often separate standard locations for these two use cases, e.g. `XDG_DATA_HOME` vs. `XDG_CONFIG_HOME`.

See how bitcoin.conf works.

stanislav-tkach:
I have checked the Bitcoin behavior and updated the pull request correspondingly. So the logic now is the following:
- Both `--datadir` and `--conf` options can be specified.
- The data directory location is stored in the config.
- The precedence is the following:
  - Command line options.
  - Values from the config.
  - Default values.

I think that these are somewhat significant changes, so I'm going to request a re-review.

test/functional/test_framework/test_node.py
```diff
@@ -99,15 +99,15 @@ def __init__(self, i, datadir, *, chain, rpchost, timewait, timeout_factor, bitc
         rpc_addr = self.init_rpc_url.split("http://")[-1].split('@')[-1]
         p2p_addr = p2p_url(self.index)
 
-        config_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "test_config.toml")
+        datadir_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "./.mintlayer")
```
TheQuantumPhysicist, line 102: You don't have to use `./.mintlayer` in join. Just `.mintlayer` suffices. This is actually confusing.

TheQuantumPhysicist (approved):
Looking good

muursh (approved):
Fine by me

PR #293, opened 2022-07-19
Implement missing functionality for `PeerDb`

altonen:
Previously there was an awkward split between `PeerManager` and `PeerDb` as the implementation for the latter was mostly missing but it was heading towards the right direction so implementing, e.g., peer info logging for banned peers would've resulted in an ugly and temporary hack. This PR implements the needed functionality to make both `PeerManager` and `PeerDb` deal with tasks that they're supposed to be concerned with.

### DIFF
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -6,13 +6,13 @@ license = "MIT"
 
 [dependencies]
 async-trait = "0.1"
+either = "1.7"
 futures = "0.3"
 itertools = "0.10"
 parity-scale-codec = "3.1"
 sscanf = "0.2"
 thiserror = "1.0"
 void = "1.0"
-serde = { version = "1", features = ["derive"] }
 
 # local dependencies
 common = { path = "../common/" }
@@ -31,6 +31,10 @@ features = ["macros"]
 version = "0.45"
 default-features = false
 features = ["gossipsub", "identify", "mdns", "mplex", "noise", "ping", "request-response", "tcp-async-io"]
+
+[dependencies.serde]
+version = "1"
+features = ["derive"]
 
 [dependencies.tokio]
 version = "1"
--- a/p2p/src/swarm/mod.rs
+++ b/p2p/src/swarm/mod.rs
@@ -181,7 +181,7 @@ where
         info: net::types::PeerInfo<T>,
     ) -> crate::Result<()> {
         let peer_id = info.peer_id;
-        log::debug!("peer {} connected, address {}, {}", peer_id, address, info);
+        log::debug!("peer {peer_id} connected, address {address}, {info}");
 
         ensure!(
             info.magic_bytes == *self.chain_config.magic_bytes(),
@@ -227,7 +227,7 @@ where
         address: T::Address,
         info: net::types::PeerInfo<T>,
     ) -> crate::Result<()> {
-        log::debug!("validate inbound connection, inbound address {}", address);
+        log::debug!("validate inbound connection, inbound address {address}");
 
         ensure!(
             !self.peerdb.is_active_peer(&info.peer_id),
@@ -260,7 +260,7 @@ where
     /// or by the [`PeerManager::heartbeat()`] function which has decided to cull
     /// this connection in favor of another potential connection.
     async fn close_connection(&mut self, peer_id: T::PeerId) -> crate::Result<()> {
-        log::debug!("connection closed for peer {}", peer_id);
+        log::debug!("connection closed for peer {peer_id}");
 
         self.tx_sync.send(event::SyncControlEvent::Disconnected(peer_id)).await?;
         self.peerdb.peer_disconnected(&peer_id);
@@ -273,7 +273,7 @@ where
     /// which makes the `PeerDb` mark is banned and prevents any further connections with the peer
     /// and also bans the peer in the networking backend.
     async fn adjust_peer_score(&mut self, peer_id: T::PeerId, score: u32) -> crate::Result<()> {
-        log::debug!("adjusting score for peer {}, adjustment {}", peer_id, score);
+        log::debug!("adjusting score for peer {peer_id}, adjustment {score}");
 
         if self.peerdb.adjust_peer_score(&peer_id, score) {
             return self.handle.ban_peer(peer_id).await;
@@ -350,7 +350,7 @@ where
         );
 
         for _ in 0..npeers {
-            if let Some(addr) = self.peerdb.best_peer_addr()? {
+            if let Some(addr) = self.peerdb.take_best_peer_addr()? {
                 match self.connect(addr.clone()).await {
                     Ok(_) => {
                         self.pending.insert(addr, None);
@@ -392,10 +392,10 @@ where
                 result
             }
             Err(err) => {
-                log::warn!("non-fatal error occurred: {}", err);
+                log::warn!("non-fatal error occurred: {err}");
 
                 if let Some(peer_id) = peer_id {
-                    log::info!("adjust peer score for peer {}, reason {}", peer_id, err);
+                    log::info!("adjust peer score for peer {peer_id}, reason {err}");
                     return self.adjust_peer_score(peer_id, err.ban_score()).await;
                 }
 
@@ -423,10 +423,7 @@ where
             tokio::select! {
                 event = self.rx_swarm.recv().fuse() => match event.ok_or(P2pError::ChannelClosed)? {
                     event::SwarmEvent::Connect(addr, response) => {
-                        log::debug!(
-                            "try to establish outbound connection to peer at address {:?}",
-                            addr
-                        );
+                        log::debug!("try to establish outbound connection to peer at address {addr}");
 
                         let res = self.connect(addr.clone()).await.map(|_| {
                             self.pending.insert(addr.clone(), Some(response));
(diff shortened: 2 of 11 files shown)


p2p/src/swarm/peerdb.rs
```diff
@@ -30,86 +30,119 @@
...
-        // TODO: improve peer selection
-        let key = match self.discovered.keys().next() {
-            Some(key) => *key,
-            None => return None,
-        };
+    /// Get the number of active peers
+    pub fn active_peer_count(&self) -> usize {
+        self.peers
+            .len()
+            .saturating_sub(self.pending.len())
```
TheQuantumPhysicist, line 113: I'm a big opponent of using saturating sub. It basically hides bugs. Is there a way we can do this in a better, consistent way that reveals errors?

p2p/src/swarm/peerdb.rs
```diff
@@ -122,25 +155,64 @@ impl<T: NetworkingService> PeerDb<T> {
...
+    }
 
-        for info in peers.iter() {
-            // TODO: update peer stats
-            if self.peers.contains_key(&info.id) {
-                continue;
-            }
+    /// Get socket address of the next best peer (TODO: in terms of peer score)
+    // TODO: rewrite all of this
+    pub fn best_peer_addr(&mut self) -> crate::Result<Option<T::Address>> {
```
TheQuantumPhysicist, line 165: Shouldn't this function be called "take_..."? Usually we use "take" or "pop" for things that erase from the container and return the result.

p2p/src/swarm/peerdb.rs
```diff
@@ -122,25 +155,64 @@ impl<T: NetworkingService> PeerDb<T> {
...
+                Ok(Some(addr))
+            }),
+            None => Err(P2pError::DatabaseFailure),
+        }
+    }
 
-                    ip4.extend(info.ip4.clone());
-                    ip6.extend(info.ip6.clone());
+    /// Discover new peer addresses
+    pub fn discover_peers(&mut self, peers: &[types::AddrInfo<T>]) {
```
TheQuantumPhysicist, line 191: I think this function can be broken down to become more meaningful. Usually functions that take lists and loop over their whole operation can be reduced to functions that take individual elements, which then you use with a map. As I see from the implementation, there's no relation between any of the peers. Maybe it's better to make this function `fn discover_peer(&mut self, peer: types::AddrInfo<T>) -> Result<()>`. Then, if you wish, you can use a helper function, though I'd prefer to see the caller just use a map or for_each with all the peers.

p2p/src/swarm/peerdb.rs
```diff
@@ -30,86 +30,119 @@
...
 
-                if ip6.is_empty() {
-                    ip4.iter().next().expect("ip4 empty").clone()
-                } else {
-                    ip6.iter().next().expect("ip6 empty").clone()
-                }
-            }
-        };
+    /// Get mutable reference to the peer store
+    pub fn peers(&mut self) -> &mut HashMap<T::PeerId, Peer<T>> {
```
TheQuantumPhysicist, line 129: Why do all these functions have to be mutable like this? This is quite dangerous... the struct should manage what access rules are allowed.

p2p/src/swarm/peerdb.rs
iljakuklic: I believe this `saturating_sub` is basically equivalent to `max(0, peers_len - pending_len)`, but without conversion and underflow issues, it's a well defined operation. In fact, there is a lint in clippy that suggests `saturating_sub` as a better behaved alternative to the `max` expression above.

TheQuantumPhysicist: If it's being as an alternative to max, I get it. That's OK. But I think in many cases developers use it just to get rid of underflow/overflow errors and its handling, which might lead to inconsistencies.

iljakuklic: To be more specific, I'm fine with strongly discouraging `T::saturating_add` and  `S::saturating_sub`, where `T` is any integer type and `S` is a signed integer type. I'd like to keep `U::saturating_sub` for unsigned types `U`, which is basically [monus](https://en.wikipedia.org/wiki/Monus).

p2p/src/swarm/peerdb.rs
iljakuklic: Agreed with @TheQuantumPhysicist. If you really want a mutable reference, I'd prefer the method to be called `peers_mut` so it's more more clear what's going on in use sites.

altonen: I can probably get rid of these

TheQuantumPhysicist: Even `peers_mut` is so invasive. IMHO, I think the right way is to get use getters/setters that limit that possible changes to this struct so that invariants can remain correct.

p2p/src/swarm/peerdb.rs
TheQuantumPhysicist: Guys, I'm still having trouble understanding this. Help me understand what's going on.

If the number of planets in the solar system is 8, and we want to know the number of non-red and non-blue planets in the system, we do `total_planet_count - total_red - total_blue`, so `8 - 1 - 1` which yields `6` (assuming earth is blue and mars is red). Good. Now, how does it help to make it `total_planet_count.saturating_sub(total_red).saturating_sub(total_blue)`? Yes, if all the inputs are correct, then this will still yield seven. But if you miscounted something (broke the invariants of your system), then the true result can be 5, or 3, or 0... but then you'll never catch anything less than zero (an underflow) because it's a saturating sub. Meaning, a bug will always give you wrong results.

How does using `max(0, total_planet_count - total_red)` help in this case, based on Lukas's assumption that this is what we're doing?

altonen: The most robust way to calculate number of active peers is to iterate over the collection and count the number of active peers.

TheQuantumPhysicist: I think I'm not making my point clear. I'll try again. Let's consider the equation (and please tell me if the equation is incorrect):

`all_peers = active + banned + available + pending`.

And if I understand correctly, all those types are orthogonal. Meaning, a peer can be only one of these at a time. Right?

In your variables that represents that model, you don't have active, and if I understand correctly, active is basically:

`active = all_peers - banned - available - pending`,

which is what you're doing in this subtraction chain. Am I getting this right? If yes, then this equation is an invariant. There's absolutely no way this will change unless something is broken. IF something is broken, then using saturating_sub will only cause hiding bugs.

The only way saturating_sub is acceptable is if those types aren't orthogonal.

altonen: I don't mind changing it to `expect()` and removing `saturating_sub()`

TheQuantumPhysicist: If my understanding is correct, then using expect is the right way to go. What do you think @iljakuklic ?

p2p/src/swarm/tests/peerdb.rs
```diff
@@ -0,0 +1,559 @@
...
+    // lower threshold, ban for more minor offense
+    {
+        let mut config = config::P2pConfig::new();
+        config.ban_threshold = 20;
+        let mut peerdb = PeerDb::<Libp2pService>::new(Arc::new(config));
+
+        let id = add_active_peer(&mut peerdb);
+        assert!(peerdb.adjust_peer_score(&id, 30));
+        assert!(peerdb.banned().contains(&id));
+    }
```
iljakuklic, line 169: I'd like to see these test cases each having its own test function. Makes it easier to see which fails and re-run them on individual basis. I appreciate coming up with names is annoying but you already have comments there to take inspiration from.

p2p/Cargo.toml
```diff
@@ -32,6 +32,10 @@ version = "0.45"
 default-features = false
 features = ["gossipsub", "identify", "mdns", "mplex", "noise", "ping", "request-response", "tcp-async-io"]
 
+[dependencies.serde]
```
stanislav-tkach, line 35: Just curious: do we have some agreement on the format of dependencies? I have seen one line form in other crates even  if both the version and the features are specified.

p2p/src/swarm/mod.rs
```diff
@@ -177,8 +175,13 @@ where
...
     /// by the node as result of swarm maintenance.
-    async fn accept_connection(&mut self, info: net::types::PeerInfo<T>) -> crate::Result<()> {
-        log::debug!("{}", info);
+    async fn accept_connection(
+        &mut self,
+        address: T::Address,
+        info: net::types::PeerInfo<T>,
+    ) -> crate::Result<()> {
+        let peer_id = info.peer_id;
+        log::debug!("peer {} connected, address {}, {}", peer_id, address, info);
```
stanislav-tkach, line 184:
```suggestion
        log::debug!("peer {peer_id} connected, address {address}, {info}");
```
Nitpick: you don't like captured identifiers? 🙃

stanislav-tkach (approved):

p2p/src/swarm/mod.rs
altonen: I'll update it, thanks!

p2p/Cargo.toml
altonen: I don't think we have any agreements but I've used the multi-line format in P2P when the dependency has more than one item because I find the curly bracket syntax harder to read and maintain.

p2p/src/swarm/peerdb.rs
```diff
@@ -30,86 +30,126 @@
...
-        let peer_info = self.discovered.remove(&key).expect("peer to exist");
-        let addr = match peer_info {
-            PeerAddrInfo::Raw { ref ip4, ref ip6 } => {
-                assert!(!(ip4.is_empty() && ip6.is_empty()));
+    pub fn active_peers(&self) -> Vec<(&T::PeerId, &PeerContext<T>)> {
+        self.peers
+            .iter()
+            .filter_map(|(id, info)| match info {
+                Peer::Active(inner) => Some((id, inner)),
+                _ => None,
```
TheQuantumPhysicist, line 130: Please, no wild-card in matches, no matter what.

TheQuantumPhysicist (approved):

PR #297, opened 2022-07-21
Get rid of all thread-related testing in time

TheQuantumPhysicist:
Given that rust tests run in parallel, testing time with a global state in multiple threads is basically undefined behavior. No more multi-threading.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,6 +681,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_test",
+ "serial_test",
  "serialization",
  "sscanf",
  "static_assertions",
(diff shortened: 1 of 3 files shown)


altonen (approved):

iljakuklic:
Are there any important situations that have been tested with the previous thread-based approach that are not tested now? If so, would it be hard to recover the lost coverage?

TheQuantumPhysicist:
> Are there any important situations that have been tested with the previous thread-based approach that are not tested now? If so, would it be hard to recover the lost coverage?

I don't think anything is lost... it's just a set and a get and it was being done from other threads as a fancy gymnastics. After all, if you set it, and it's set, we're done.

stanislav-tkach:
In case we still need to run some tests sequentially without limiting all tests execution to one thread, there is the `serial_test` crate.

PR #298, opened 2022-07-21
Change BlockTimestamp internal integer to u64

TheQuantumPhysicist:
Since we're using compact encoding, there's zero harm in using u64 for block timestamps. It's even beneficial so that we don't worry about running out of time in the future. It also simplify Duration usage since seconds there are also u64.

### DIFF
--- a/common/src/chain/block/timestamp.rs
+++ b/common/src/chain/block/timestamp.rs
@@ -15,7 +15,6 @@
 
 use parity_scale_codec::{Decode, Encode};
 use std::time::Duration;
-use thiserror::Error;
 
 pub type BlockTimestampInternalType = u64;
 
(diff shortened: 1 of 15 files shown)


common/src/chain/block/timestamp.rs
```diff
@@ -15,12 +15,13 @@
 
 use parity_scale_codec::{Decode, Encode};
 use std::time::Duration;
-use thiserror::Error;
+
+pub type BlockTimestampInternalType = u64;
```
altonen, line 19: Why introduce a new type?

altonen (approved):

common/src/chain/block/timestamp.rs
TheQuantumPhysicist: Just for consistency in the file... in case we think of changing it, it'll change everything automatically. I hate repeating the same thing multiple times.

stanislav-tkach (approved):

common/src/chain/block/timestamp.rs
stanislav-tkach: By the way, we have slightly differently named type aliases in other places. For example, `HeightIntType` or `DistanceIntType`. Should this also be `BlockTimestampIntType`? Or do we want to emphasize that it is an internal type?

iljakuklic (approved):

PR #299, opened 2022-07-21
common: Separate out `Block` and `Genesis` types

iljakuklic:
Introduce a type-level distinction between `Genesis` and proper `Block`. There is an additional abstraction called `GenBlock` representing something that can be either a proper `Block` or `Genesis`.

This is analogous to a linked list:
 * List roughly corresponds to GenBlock
 * Nil corresponds to Genesis
 * Cons cell roughly corresponds to a Block

Or in a pseudocode:

```
List     = Nil     | Cons  { head: Item,      tail: List       }
GenBlock = Genesis | Block { data: BlockData, parent: GenBlock }
```

Note: In the code, the parent link is actually part of the block (block header to be more precise), but the high-level structure still holds.

The idea is to be more precise about what could be expected where, and communicate the information in types:

* If only regular block or only genesis is possible in particular case, the type reflects that, and the code does not need to be concerned with the "other" case.
* If either is possible, the type is one of `GenBlock` and the code is required to handle both cases by using appropriate abstraction or by explicitly matching and handling both possibilities.

This is still somewhat rough around the edges and subsequent cleanups are expected. These include adding some helper methods to clean up repeating patterns, adding extra conversion and comparison methods, etc. These would hopefully be either more self-contained or mechanical changes. So hopefully less disruptive.


chainstate/src/detail/gen_block_index.rs
```diff
@@ -0,0 +1,57 @@
...
+use common::chain::{ChainConfig, GenBlock};
+use common::primitives::{BlockHeight, Id};
+use common::Uint256;
+
+/// Generalized block index
+#[derive(Clone, Debug)]
+#[allow(clippy::large_enum_variant)]
+pub enum GenBlockIndex<'a> {
+    Block(BlockIndex),
+    Genesis(&'a ChainConfig),
```
TheQuantumPhysicist, line 12: Why not include the genesis block object directly, instead of complicating it with a reference to ChainConfig?

common/src/chain/gen_block.rs
```diff
@@ -0,0 +1,80 @@
...
+    Genesis(Id<Genesis>),
+    Block(Id<Block>),
+}
+
+impl GenBlockId {
+    pub fn is_genesis(&self) -> bool {
+        matches!(self, GenBlockId::Genesis(_))
+    }
+
+    pub fn block_id(self) -> Option<Id<Block>> {
```
TheQuantumPhysicist, line 74: This is confusing. Why does Genesis return None? At least the function name is not appropriate. Something like `block_id_if_not_genesis` is more appropriate. I know it's a mouthful, but the current name is... confusing.

chainstate/src/detail/gen_block_index.rs
iljakuklic: * `ChainConfig` has a pre-calculated genesis ID, so when you use `GenBlockIndex::block_id`, it does not have to serialize and hash it.
* It's a reference just to avoid copying. Works out quite

I'm sure this could be improved, suggestions welcome. The question is how much of this should be done now and how much should be a follow up since this is now blocking other changes.

common/src/chain/gen_block.rs
iljakuklic: Ok. The idea was to turn it into an option, so you get access to all the goodies `Option` has to offer. Regarding the name maybe `proper_block_id`? I already use the "proper block" terminology when I want to be explicit about it in some places I think.

chainstate/src/detail/pow/work.rs
```diff
@@ -74,7 +74,8 @@ fn calculate_work_required<H: BlockIndexHandle>(
         PoWStatus::Ongoing => {
             let prev_block_id = header
                 .prev_block_id()
-                .clone()
+                .classify(chain_config)
+                .block_id()
                 .expect("If PoWStatus is `Onging` then we cannot be at genesis");
```
TheQuantumPhysicist, line 79: I think this is one of the examples we'd like to avoid. The better way of dealing with this, IMHO, is to just get the id, as is, and use it as if it's definitly not genesis. If it's not genesis, it won't be found. Hence, never causing a crash. The other day Aaro bumped, by coincidence, by a case where such an expect triggers if the wrong block is submitted.

chainstate/src/detail/gen_block_index.rs
iljakuklic: It could just store a reference to genesis and the pre-calculated ID

common/src/chain/gen_block.rs
TheQuantumPhysicist: `chain_block_id()` should be OK, I'd say.

chainstate/src/detail/gen_block_index.rs
TheQuantumPhysicist: Quite frankly, if we store the whole genesis block in it, that'd be fine. Or the id alone. But you're right, this doesn't matter currently. I was just wondering why this looks like this. Feel free to skip it.

iljakuklic: > Quite frankly, if we store the whole genesis block in it, that'd be fine. Or the id alone. But you're right, this doesn't matter currently. I was just wondering why this looks like this. Feel free to skip it.

Genesis block contains the initial UTXO set, which may get quite massive, especially with all the pre-mine payouts :smile:

TheQuantumPhysicist: Having said that you can skip this, premine in genesis will be very simple. It's gonna be one output to one address. Then the distribution is a whole other problem that can be done later. No need to complicate genesis block.

chainstate/src/detail/tests/test_framework.rs
```diff
@@ -97,40 +99,123 @@ impl BlockTestFramework {
...
+                match spent_status {
+                    Some(OutputSpentState::Unspent) => println!("\t\t+Spend: Unspent"),
+                    Some(OutputSpentState::SpentBy(spender)) => {
+                        println!("\t\t+Spend: {:?}", spender)
+                    }
+                    None => println!("\t\t+Spend: Not in mainchain"),
+                }
+            }
+        }
+    }
```
iljakuklic, line 199: Looks like this has been re-introduced by accident. Let me fix that.

azarovh:
Was a bit confused with a name `GenBlock` because it looks like a concatenation of Genesis and Block. Maybe something like `AnyBlock` or `GeneralBlock` would be more expressive?

TheQuantumPhysicist:
> Was a bit confused with a name `GenBlock` because it looks like a concatenation of Genesis and Block. Maybe something like `AnyBlock` or `GeneralBlock` would be more expressive?

It's meant to be `GeneralizedBlock`. But Lukas likes concise naming... I'm not happy with it either, but we can change it later after we agree on a better scheme.

chainstate/src/detail/mod.rs
```diff
@@ -268,13 +260,35 @@ impl Chainstate {
...
+        // Gather information about genesis.
+        let genesis = self.chain_config.genesis_block();
+        let genesis_id = self.chain_config.genesis_block_id();
+        let utxo_count = genesis.utxos().len() as u32;
+        let genesis_index =
+            common::chain::TxMainChainIndex::new(genesis_id.clone().into(), utxo_count)
+                .expect("Genesis not constructed correctly");
+
+        // Initialize storage with given info
+        let mut tx = self.chainstate_storage.transaction_rw();
```
TheQuantumPhysicist, line 276: I know it's nothing... but please call it `db_tx`... we wanna try to always distinguish between a value transaction and a database transaction that way.

chainstate/src/detail/mod.rs
```diff
@@ -268,13 +260,35 @@ impl Chainstate {
...
+        tx.set_best_block_id(&genesis_id).map_err(BlockError::StorageError)?;
+        tx.set_block_id_at_height(&BlockHeight::zero(), &genesis_id)
+            .map_err(BlockError::StorageError)?;
+        tx.set_mainchain_tx_index(&genesis_id.into(), &genesis_index)
+            .map_err(BlockError::StorageError)?;
+        tx.commit().map_err(BlockError::StorageError)
```
TheQuantumPhysicist, line 282: This probably deserves an expect?

common/src/chain/gen_block.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use super::{Block, Genesis};
+use crate::primitives::Id;
+
+/// Generalized block that's either [genesis::Genesis] or [block::Block].
+///
+/// Does not contain any data, only used as `Id<GenBlock>` to signify given Id can refer to either
+/// genesis or proper block.
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub enum GenBlock {}
```
TheQuantumPhysicist, line 26: I'm wondering whether this can be completely removed and replaced with GenBlockId...

chainstate/src/detail/chainstateref.rs
```diff
@@ -292,17 +303,22 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         let id = self
             .get_block_id_by_height(height)?
             .ok_or(PropertyQueryError::BlockForHeightNotFound(*height))?;
+        let id = id.classify(self.chain_config).block_id().expect("Genesis has no header");
```
TheQuantumPhysicist, line 306: The reason why I'm excited for separating genesis block from chain blocks is to avoid these expects as much as possible... we have to be really careful. If p2p, for example, tries to get the header from a height 0 for any reason because a peer sends some fishy information that they can manipulate, the result should not be crashing the node. Generally speaking, in the cases of property querying, it's always preferred to return `Ok(None)` in such cases unless it's provable that no one will ever call this function with an invalid height.

chainstate/src/detail/chainstateref.rs
```diff
@@ -266,23 +278,22 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
         }
 
-        while first_block_index.block_id() != second_block_index.block_id()
-            && !first_block_index.is_genesis(self.chain_config)
-            && !second_block_index.is_genesis(self.chain_config)
-        {
-            first_block_index = self.get_previous_block_index(&first_block_index)?;
-            second_block_index = self.get_previous_block_index(&second_block_index)?;
+        loop {
+            match (&first_block_index, &second_block_index) {
```
TheQuantumPhysicist, line 282: Isn't it better to be able to compare `Id<GenBlock>` and then that automatically figures out whether it's the same instead of handling cases here?

chainstate/src/detail/tests/processing_tests.rs
```diff
@@ -147,9 +119,9 @@ fn test_orphans_chains() {
     });
 }
 
-// Create the chainstate without a genesis block.
 #[test]
-fn empty_chainstate() {
+#[should_panic(expected = "Best block ID not initialized")]
+fn test_empty_chainstate_no_genesis() {
```
stanislav-tkach, line 124: I think that the "test" prefix is needless. There is the "`#[test]` attribute right here and the output is also pretty obvious: "running tests... test <test_name>". I suppose that `chainstate_no_genesis` is enough here.

chainstate/src/detail/tests/test_framework.rs
```diff
@@ -65,11 +60,18 @@ impl BlockTestFramework {
         Self::with_chainstate(chainstate)
     }
 
-    pub fn random_block(&self, parent_block: &Block, params: Option<&[TestBlockParams]>) -> Block {
-        let (mut inputs, outputs): (Vec<TxInput>, Vec<TxOutput>) =
-            parent_block.transactions().iter().flat_map(create_new_outputs).unzip();
+    pub(in crate::detail::tests) fn random_block(
```
stanislav-tkach, line 63: Is the `pub(in crate::detail::tests)` part really needed? Isn't `BlockTestFramework` itself only visible in these tests?

chainstate/src/detail/tests/test_framework.rs
```diff
@@ -97,40 +99,123 @@ impl BlockTestFramework {
...
-            Some(prev_block_hash),
+            prev_block_hash,
             BlockTimestamp::from_duration_since_epoch(time::get()),
             ConsensusData::None,
         )
         .expect(ERR_CREATE_BLOCK_FAIL)
     }
 
-    pub fn genesis(&self) -> &Block {
+    pub(in crate::detail::tests) fn genesis(&self) -> &Genesis {
```
stanislav-tkach, line 109: Same here, but I suppose there is a reason behind this. 🙃

stanislav-tkach (approved):
I went through the pull request and wasn’t able even to nitpick much. 🙃  By the way, I really like the idea of using specific types as much as possible.

chainstate/src/detail/tests/test_framework.rs
iljakuklic: I don't mind this being just `pub`

common/src/chain/gen_block.rs
iljakuklic: Done

chainstate/src/detail/mod.rs
iljakuklic: Only the last one or all of them? If genesis initialization fails at any point we're kind of screwed anyway...

chainstate/src/detail/pow/work.rs
iljakuklic: I am not sure I understand the comment completely. Let me just point out that this checks parent's ID and whether it is genesis, not whether the current block is a genesis. This kind of reasoning about block's parents/ancestors, while desirable, is out of scope of the present PR.

Similar issue comes up in orphan pool: A block with genesis as a parent (transitively) has no business being in orphan pool, so some `expect`s remain there. That should be addressed separately, although it would be nice to have too.

common/src/chain/gen_block.rs
iljakuklic: I don't like having extra bytes attached to the hashes, although it would simplify the code somewhat.

chainstate/src/detail/chainstateref.rs
iljakuklic: Good points. The point of this function is to get block header from height. Returning `Ok(None)`, we would lose the distinction between asking about block header at height 0 and asking about block header at height beyond the current tip. Do we care about that? Probably not at the moment, since the expect is not hit in the tests but in general I'm not sure.

On a bit more high-level note, I think one of the reasons for bunch of remaining `expect`s is that not all types currently reflect the distinction between the genesis and a chain block. The prime offender currently being `BlockHeight`. Genesis corresponds to block height 0, chain block corresponds to a non-zero block height, and in general block height may refer to either genesis or a chain block.

So similar thing could be done to it as to other types in this PR, i.e. we could have a `BlockHeight` and a `NonZeroBlockHeight`, with some conversion functions:
```rust
BlockHeight::non_zero(Self) -> Option<NonZeroBlockHeight>
NonZeroBlockHeight::into(Self) -> BlockHeight
```

Then, there could be methods like:
```rust
get_block_id_by_height(NonZeroBlockHeight) -> Result<Id<Block>>
get_gen_block_id_by_height(BlockHeight) -> Result<Id<GenBlock>>
get_block_by_height(NonZeroBlockHeight) -> Result<Block>
// This function could then have signature, that makes it callable only with non-zero height:
get_header_from_height(NonZeroBlockHeight) -> Result<BlockHeader>
```

With this we would not have to choose between an expect and an error condition that the caller has to deal with.

chainstate/src/detail/pow/work.rs
TheQuantumPhysicist: > Let me just point out that this checks parent's ID and whether it is genesis

A peer can set the parent id to whatever they want. That's the issue.

If it's in the orphan's pool, that's not good, but remember that nothing goes into the orphan's pool unless it's from local source.

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: > Returning Ok(None), we would lose the distinction between asking about block header at height 0 and asking about block header at height beyond the current tip

Easy. Just return an error if the value is zero. That's alright too. 

> So similar thing could be done to it as to other types in this PR, i.e. we could have a BlockHeight and a NonZeroBlockHeight, with some conversion functions:

Then we probably have to expect on some things... I don't see this as a bad idea to be tried out, but maybe not in this PR. Maybe after this one you can give it a shot. I think this PR is already huge and experimenting further could lead to this becoming a long standing PR, which we want to avoid because this is currently blocking everything else.

common/src/chain/gen_block.rs
TheQuantumPhysicist: I bet it'll simplify it a lot... but maybe, again, this can go to another PR (please slap a TODO on it)

chainstate/src/chainstate_interface.rs
```diff
@@ -29,17 +29,17 @@ pub trait ChainstateInterface: Send {
...
         &self,
-        block_id: &Id<Block>,
+        block_id: &Id<GenBlock>,
     ) -> Result<Option<BlockHeight>, ChainstateError>;
     fn get_best_block_height(&self) -> Result<BlockHeight, ChainstateError>;
     fn get_block_id_from_height(
         &self,
         height: &BlockHeight,
-    ) -> Result<Option<Id<Block>>, ChainstateError>;
+    ) -> Result<Option<Id<GenBlock>>, ChainstateError>;
```
iljakuklic, line 42: I kept the chainstate interface changes rather minimal, but this could be extended to eliminate some repeating patterns (e.g. match on block id to figure out if it's genesis, call something if it's not, etc)

chainstate/src/detail/chainstateref.rs
```diff
@@ -127,8 +133,11 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         (self.time_getter)()
     }
 
-    pub fn get_best_block_id(&self) -> Result<Option<Id<Block>>, PropertyQueryError> {
-        self.db_tx.get_best_block_id().map_err(PropertyQueryError::from)
+    pub fn get_best_block_id(&self) -> Result<Id<GenBlock>, PropertyQueryError> {
+        self.db_tx
+            .get_best_block_id()
+            .map_err(PropertyQueryError::from)
+            .map(|bid| bid.expect("Best block ID not initialized"))
```
iljakuklic, line 140: This is not strictly related to genesis/block separation. Just asserts that after chainstate has been initialised, the best block ID should always be available.

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: Btw: https://doc.rust-lang.org/std/num/struct.NonZeroU64.html

chainstate/src/detail/chainstateref.rs
```diff
@@ -755,38 +720,25 @@ impl<'a, S: BlockchainStorageWrite, O: OrphanBlocksMut> ChainstateRef<'a, S, O>
...
-            std::cmp::max(prev_block_index.chain_timestamps_max(), block.timestamp())
-        });
+        let time_max = std::cmp::max(prev_block_index.chain_timestamps_max(), block.timestamp());
 
         // Set Chain Trust
-        let prev_chain_trust = prev_block_index.map_or(Uint256::from_u64(0), |prev_block_index| {
-            *prev_block_index.chain_trust()
-        });
-        let chain_trust = prev_chain_trust + self.get_block_proof(block)?;
+        let chain_trust = *prev_block_index.chain_trust() + self.get_block_proof(block)?;
```
iljakuklic, line 741: This function is a fair bit cleaner after the refactor.

chainstate/src/detail/mod.rs
```diff
@@ -268,13 +260,37 @@ impl Chainstate {
...
+        let mut db_tx = self.chainstate_storage.transaction_rw();
+        db_tx.set_best_block_id(&genesis_id).map_err(BlockError::StorageError)?;
+        db_tx
+            .set_block_id_at_height(&BlockHeight::zero(), &genesis_id)
+            .map_err(BlockError::StorageError)?;
+        db_tx
+            .set_mainchain_tx_index(&genesis_id.into(), &genesis_index)
+            .map_err(BlockError::StorageError)?;
+        db_tx.commit().map_err(BlockError::StorageError)
+    }
```
iljakuklic, line 285: This is where the genesis block gets initialised and info about it is written in the storage.

chainstate/src/detail/tests/test_framework.rs
```diff
@@ -244,4 +238,9 @@ impl BlockTestFramework {
     pub fn chainstate(&mut self) -> &mut Chainstate {
         &mut self.chainstate
     }
+
+    pub fn index_at(&self, at: usize) -> &BlockIndex {
+        assert!(at > 0, "No block index for genesis");
+        &self.block_indexes[at - 1]
+    }
```
iljakuklic, line 245: Only "proper" block indices (not genesis) are stored in the `block_indexes` vector. That introduces an off-by-one array access issue, so it is better accessed through this function

common/src/chain/gen_block.rs
```diff
@@ -0,0 +1,80 @@
...
+//
+// http://spdx.org/licenses/MIT
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Generalized block, or [GenBlock]
```
iljakuklic, line 16: This is where the abstraction over IDs that can be either genesis or chain block gets introduced.

common/src/chain/transaction/input.rs
```diff
@@ -24,7 +25,7 @@ pub enum OutPointSourceId {
     #[codec(index = 0)]
     Transaction(Id<Transaction>),
     #[codec(index = 1)]
-    BlockReward(Id<Block>),
+    BlockReward(Id<GenBlock>),
```
iljakuklic, line 28: Genesis block UTXOs are addressed in a way similar to block rewards rather than by transaction IDs (genesis no longer contains tranctions, just the intial set of unspent outputs)

iljakuklic (commented):
Not really a review. Just pointing out a couple of places of interest below.

chainstate/src/detail/chainstateref.rs
iljakuklic: > Then we probably have to expect on some things... I don't see this as a bad idea to be tried out, but maybe not in this PR. Maybe after this one you can give it a shot.

But hopefully there would be fewer of them... and possibly in less tricky places

> I think this PR is already huge and experimenting further could lead to this becoming a long standing PR, which we want to avoid because this is currently blocking everything else.

Agreed

chainstate/src/detail/tests/test_framework.rs
TheQuantumPhysicist: To be honest, the block_indexes vector shouldn't exist. The chainstate object should be used, specifically.

common/src/chain/gen_block.rs
iljakuklic: I think a sort of compromise solution could be to use `Id<GenBlock>` in data structures but `GenBlockId` could be accepted by more methods, so they have immediate access to the information whether it is a genesis id or not, without having to query chain config. We can experiment with that later I think.

chainstate/src/detail/gen_block_index.rs
iljakuklic: Can this be resolved for the current PR?

TheQuantumPhysicist: Just slap TODOs around if you feel the change is big. No need to do it right now.

chainstate/src/detail/chainstateref.rs
iljakuklic: Will have a look into this

chainstate/src/detail/tests/test_framework.rs
TheQuantumPhysicist: Yeah, sure... we can look into this later. Don't worry about it.

chainstate/src/detail/tests/mod.rs
```diff
@@ -100,38 +92,78 @@ fn chainstate_with_config(
     .unwrap()
 }
 
-fn produce_test_block(prev_block: &Block, orphan: bool) -> Block {
-    produce_test_block_with_consensus_data(prev_block, orphan, ConsensusData::None)
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TestBlockInfo {
```
TheQuantumPhysicist, line 96: What problem does this solve?

altonen (approved):

common/src/chain/transaction/transaction_index/mod.rs
```diff
@@ -26,6 +26,8 @@ pub enum Spender {
     RegularInput(Id<Transaction>),
     #[codec(index = 1)]
     StakeKernel(Id<Block>),
+    #[codec(index = 2)]
+    Premine(Id<Genesis>),
```
TheQuantumPhysicist, line 30: This is an example of what we've discussed. There's no reason to distinguish between genesis and other kinds of blocks at the output level.

TheQuantumPhysicist (approved):
Good job. I have many reservations as I mentioned in chat and in the commetns, but I think we're going in the right direction with this. Treating genesis block as a special block seems to simplify things at times.

I hope I haven't missed anything. The PR is huge.

Please tag me if you need an answer from me about something.

chainstate/src/detail/tests/mod.rs
iljakuklic: The `generate_block` function takes the parent block as a parameter. However, the parent can be genesis as well so it can't just take the block directly. This struct collects just bits and pieces that are needed to generate the next block (i.e. parent's ID and outputs), and can be constructed from normal blocks or from genesis (or manually if desired).

It would be much nicer to have proper UTXO set abstraction in the code, which would abstract the user away from the consideration where the output comes from (tx output, block reward, genesis premine). I believe it would make some parts of the node code nicer too. This is just hacked-up poor man's version of it for testing.

chainstate/src/detail/tests/test_framework.rs
iljakuklic: Unresolved, so it's still visible to others

chainstate/src/chainstate_interface_impl.rs
```diff
@@ -80,7 +77,8 @@ impl ChainstateInterface for ChainstateInterfaceImpl {
     fn get_block_id_from_height(
         &self,
         height: &BlockHeight,
-    ) -> Result<Option<Id<Block>>, ChainstateError> {
+    ) -> Result<Option<Id<GenBlock>>, ChainstateError> {
+        // TODO
```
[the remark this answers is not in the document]
iljakuklic: Leftover from the middle of refactoring, will remove

chainstate/src/detail/pow/work.rs
iljakuklic: I'm resolving this as it's not a new issue introduced in this PR. The expect was there before, and I think the resolution is out of scope for the present PR.

chainstate/src/detail/chainstateref.rs
iljakuklic: Added a new error for this for the time of being

common/src/chain/transaction/transaction_index/mod.rs
iljakuklic: Adding a new variant here seemed like a path of least resistance at the time. I don't see much harm in how it's done currently, but I think it could be changed to just a variant for `Transaction` and a variant for `GenBlock` with some effort.

chainstate/src/detail/chainstateref.rs
iljakuklic: I think some matching is still necessary to get the parent (have to check whether genesis has been reached)

common/src/chain/transaction/transaction_index/mod.rs
iljakuklic: Ok, I introduced just one variant with `Id<GenBlock>` and it just works.

TheQuantumPhysicist (approved):
I hope I didn't miss anything, but I think we're good to go.

PR #300, opened 2022-07-22
Run CI build action on push to master

azarovh:
There was no check that after merging feature branches the result in master is build and tested.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,10 +1,11 @@
-
 name: build
 
-on: [pull_request, push]
-  branches:
-    - master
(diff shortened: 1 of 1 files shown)


TheQuantumPhysicist (approved):

stanislav-tkach (approved):

.github/workflows/build.yml
```diff
@@ -1,7 +1,7 @@
-
 name: build
 
 on:
+  push:
```
azarovh, line 4: That's incorrect, will fix

muursh (approved):

PR #301, opened 2022-07-22
Allow to override chainstate options for regtest

stanislav-tkach:


test/functional/test_framework/test_node.py
```diff
@@ -107,7 +107,7 @@ def __init__(self, i, datadir, *, chain, rpchost, timewait, timeout_factor, bitc
         self.args = [
             self.binary,
             "--datadir={}".format(datadir_path),
-            "run",
+            "regtest",
```
iljakuklic, line 110: Cool I kind of like the command-based interface.

node/src/runner.rs
```diff
@@ -105,15 +97,81 @@ pub async fn run(options: Options) -> Result<()> {
...
+    }
+    if let Some(max_block_header_size) = chain_max_block_header_size {
+        builder = builder.max_block_header_size(*max_block_header_size);
+    }
+    if let Some(max_block_size_with_standard_txs) = chain_max_block_size_with_standard_txs {
+        builder = builder.max_block_size_with_standard_txs(*max_block_size_with_standard_txs);
+    }
+    if let Some(max_block_size_with_smart_contracts) = chain_max_block_size_with_smart_contracts {
+        builder = builder.max_block_size_with_smart_contracts(*max_block_size_with_smart_contracts);
+    }
```
iljakuklic, line 172: Could the repeating pattern here be abstracted into a helper closure? Maybe it would help if `Builder` used a mutable reference instead of taking ownership and producing a new instance, otherwise some `std::mem::take` trickery may be needed.

iljakuklic (commented):
Adding some initial observations, will have a proper look next week

node/src/runner.rs
stanislav-tkach: I'm even thinking about a macro, because both builder method and options field have the same name. I'm going to experiment with it.

test/functional/test_framework/test_node.py
stanislav-tkach: Glad to hear! I was afraid that no one would like this approach.

node/src/runner.rs
```diff
@@ -105,17 +98,87 @@ pub async fn run(options: Options) -> Result<()> {
...
+            }
+        };
+    }
+
+    update_builder!(address_prefix);
+    update_builder!(blockreward_maturity, BlockDistance::new);
+    update_builder!(max_future_block_time_offset, Duration::from_secs);
+    update_builder!(version, SemVer::try_from, map_err);
+    update_builder!(target_block_spacing, Duration::from_secs);
+    update_builder!(coin_decimals);
```
iljakuklic, line 174: I am not sure configuring `coin_decimals` will work correctly. It should be treated as a read-only constant or removed form chain_config for the time of being.

node/src/runner.rs
```diff
@@ -105,17 +98,87 @@ pub async fn run(options: Options) -> Result<()> {
...
+                }
+            }
+        };
+        ($field: ident, $converter: stmt) => {
+            paste! {
+                if let Some(val) = [<chain_ $field>] {
+                    builder = builder.$field($converter(val.to_owned()));
+                }
+            }
+        };
```
iljakuklic, line 159: Could these two rules just delegate to the last (most general) one, with `std::convert::identity` used as the conversion function? Less code duplication at source level but it would expand into uglier code.

iljakuklic (commented):
Adding some more comments. Approving now but I think the `coin_decimals` one should be addressed or at least discussed.

iljakuklic (approved):

altonen (approved):

node/src/runner.rs
TheQuantumPhysicist: Even though coin_decimals is "too constant" to be treated like a variable, but I don't see an issue there. The only issue I see happening is with the reward schedule. What else could happen?

node/src/runner.rs
stanislav-tkach: The first branch can be delegated to the second one and it is a great idea, but I don't see how to delegate to the third branch: the types are different (`Result<T>` and `T`). Perhaps I'm missing something?

iljakuklic: Ok

node/src/runner.rs
stanislav-tkach: I have just removed this option, but I can bring it back based on the result of this discussion. 🙃

iljakuklic: It does not really have any effect in reward schedule, so I think it might be a rather misleading setting with poorly defined effects.

iljakuklic: Also it seems to me to be more of a UI concern rather than something node should be concerned with.

TheQuantumPhysicist: Remember that we want to test everything. Later, with rpc, we want to test how the wallet works. All this has to be tested together. It's not an option for normal ChainConfig. This is just for tests.

I'm OK with removing this option now and being able to bring it later if we have to. But I prefer to have the option than not. Having it and not using it is better than needing it and not having it.

node/src/options.rs
```diff
@@ -86,6 +80,50 @@ pub struct RunOptions {
     pub rpc_addr: Option<SocketAddr>,
 }
 
+/// Run the regtest node.
+#[derive(Args, Debug)]
```
TheQuantumPhysicist, line 84: Perhaps this can be moved to its own rs file. No need to mix it with the others here.

node/src/options.rs
```diff
@@ -51,16 +48,13 @@ pub struct Options {
 pub enum Command {
     /// Create a configuration file.
     CreateConfig,
-    Run(RunOptions),
+    Mainnet(RunOptions),
+    Regtest(RegtestOptions),
```
TheQuantumPhysicist, line 52: What about testnet? I think the model is incomplete. There can be nets other than mainnet that are not regtest.

node/src/options.rs
```diff
@@ -86,6 +80,50 @@ pub struct RunOptions {
...
+#[derive(Args, Debug)]
+pub struct RegtestOptions {
+    #[clap(flatten)]
+    pub run_options: RunOptions,
+    #[clap(flatten)]
+    pub chain_config: ChainConfigOptions,
+}
+
+#[derive(Args, Debug)]
+pub struct ChainConfigOptions {
```
TheQuantumPhysicist, line 93: One thing you can add here is the block subsidy. This is one thing we need to test in regtests. The idea is that it should be possible to submit block subsidy values in ChainConfig. There's a translation mechanism for that, where strings can be converted to a subsidy schedule. You don't have to do this in this PR, but I don't see why not. It's your call whether you want to make a separate PR for that if you find it more complex than this.

TheQuantumPhysicist: Btw, there's more stuff. This is just the beginning. NetUpdates, checkpoints and more should all be passed here. But they're still not parsable as strings. That comes later.

node/src/runner.rs
```diff
@@ -105,17 +98,81 @@ pub async fn run(options: Options) -> Result<()> {
...
 }
 
+async fn start(
+    config_path: &Path,
+    run_options: &RunOptions,
+    chain_config: ChainConfig,
+) -> Result<()> {
+    let node_config =
+        NodeConfig::read(config_path, run_options).context("Failed to initialize config")?;
+    log::trace!("Starting with the following config\n: {node_config:#?}");
```
TheQuantumPhysicist, line 124: I would make this info, not trace.

TheQuantumPhysicist (commented):
The minimal requirement to get this PR ready is to support testnet configuration too. 

I also don't see tests for ChainConfig settings either, which I'm not sure about. Do we want to have tests for that?

node/src/runner.rs
iljakuklic: I see a newline in the log message. How does that render in the output?

stanislav-tkach: @iljakuklic Actually there are many new lines here, because the `{:#?}` formatting produces a "pretty" output by splitting it into multiple lines:
```
> RUST_LOG=info ./node -d ./eee mainnet
[2022-07-25T14:58:36Z INFO  node::runner] Starting with the following config:
     NodeConfig {
        datadir: "./eee",
        chainstate: ChainstateConfig {
            max_db_commit_attempts: 10,
            max_orphan_blocks: 512,
        },
        p2p: P2pConfig {
            bind_address: "/ip6/::1/tcp/3031",
            ban_threshold: 100,
            outbound_connection_timeout: 10,
        },
        rpc: RpcConfig {
            bind_address: 127.0.0.1:3030,
        },
    }
[2022-07-25T14:58:36Z INFO  subsystem::manager] Initializing subsystem manager mintlayer
...
```
What do you think? Should I remove the new lines from the output?

iljakuklic: Not sure. I was mainly concerned it would continue at the start of the line. However I see it's indented so should be fine. Still makes grepping a bit annoying. So at this point it's mainly aesthetic preference and I don't mind too much either way. Let's see what others think.

TheQuantumPhysicist:
> @TheQuantumPhysicist I'm not sure about the tests. Ideally I would prefer these changes to be tested in two ways:
> 
> * Simple `assert_cmd` tests. This currently isn't trivial because `assert_cmd` don't allows to pass signals to the binary and we need that to stop the node.
> * Indirectly via functional tests. As far as I can see, the functional tests need a lot of work, so this is outside the scope of this pull request.

Actually with functional tests should happen sooner or later. If you think you can do it (in this PR or another one), I'm all for that. Again, I think this is a good opportunity to practice a little with functional tests. After all, there has to be some form of testing for these command line arguments.

stanislav-tkach:
I suppose I can do that, but I would prefer to work on functional tests in a separate pull request.

node/src/runner.rs
TheQuantumPhysicist: Would be nice if we could convert this to a table... it's a typical thing in command line arguments to print a table on start. I can live with this for now though.

TheQuantumPhysicist (approved):

PR #302, opened 2022-07-22
Make `Id` copyable

iljakuklic:
* Having to clone IDs manually is really annoying, especially since the underlying `H256` type already is copyable
* This builds on #299 and touches many of the same lines of code, so better swipe it in one go
* All changes of substance are in `common/src/primitives/id.rs`, the rest are fixes to make it work, mostly auto-applied `clippy::clone_on_copy` lint suggestions

### DIFF
--- a/chainstate-storage/src/mock.rs
+++ b/chainstate-storage/src/mock.rs
@@ -334,7 +334,7 @@ mod tests {
             let mut tx = MockStoreTxRw::new();
             tx.expect_get_best_block_id().return_const(Ok(Some(block0.get_id().into())));
             tx.expect_add_block().return_const(Ok(()));
-            let expected_id: Id<GenBlock> = block1_id.clone().into();
+            let expected_id: Id<GenBlock> = block1_id.into();
             tx.expect_set_best_block_id()
                 .with(mockall::predicate::eq(expected_id))
                 .return_const(Ok(()));
@@ -368,7 +368,7 @@ mod tests {
         let mut store = MockStore::new();
         store.expect_transaction_rw().returning(move || {
             let mut tx = MockStoreTxRw::new();
-            tx.expect_get_best_block_id().return_const(Ok(Some(top_id.clone())));
+            tx.expect_get_best_block_id().return_const(Ok(Some(top_id)));
             tx.expect_abort().return_const(Ok(()));
             tx
         });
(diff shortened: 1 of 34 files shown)


TheQuantumPhysicist:
I really am opposed to this. I'm even planning on making Uint256 non-copyable. The problem is that in rust making types copyable makes them very fluid and can affect performance. Does it affect performance? We don't know. But I think manually cloning and being careful where we throw in these 32 bytes is the right thing to do. 

In other words, no CPU supports 32-byte (256-bit) instructions... why should we have 32-byte words? That doesn't make sense.

We can benchmark things when the node is complete. But right now, I'm so sorry... I think this is just too optimistic.

iljakuklic:
> 

CPUs and compilers are much better at moving 32 bytes around than at reasoning about pointers.

TheQuantumPhysicist:
> CPUs and compilers are much better at moving 32 bytes around than at reasoning about pointers.

How do we make that decision? I'm not completely opposed to making Id copyable, but I'd rather have some metrics to make that decision.

iljakuklic:
> > CPUs and compilers are much better at moving 32 bytes around than at reasoning about pointers.
> 
> How do we make that decision? I'm not completely opposed to making Id copyable, but I'd rather have some metrics to make that decision.

1. You are already making a decision to use pointers rather than values. It may seem like a more conservative decision but I don't believe it is (see point 3 below)
2. The decision IMHO should not be based on performance alone, developer productivity and ergonomics should be take into account too.
3. Passing things by value does not necessarily constitute copying the value at each stack frame. You are basically offloading the decision whether to copy the value or pass a pointer to the compiler writers and ABI designers. They have one significant advantage over us when it comes to making the decision: they know the target architecture. At that point, the information about the availability of registers (including SIMD registers), efficient copying instructions, penalties induced by cache misses etc are all available, so different path could be taken depending on target architecture. In addition to that, you make alias analysis trivial, which is rather difficult in presence of pointers. Using references (which translate to pointers) ties the hands of the compiler somewhat in this respect. The rules for ABI lowering are bypassed too, forcing the use of a pointer (unless the function in question gets inlined).
4. I'd rather focus resources dedicated to benchmarking and optimizing to high-level architectural changes and clever data structures, rather than things compilers and CPUs are already pretty good at.

TheQuantumPhysicist (approved):
I have responses to your points, and I'm not entirely convinced, but I'm gonna make this concession because it seems to be causing you a lot of pain and I'm not sure it's harming performance. If this is gonna make development more convenient for you, I'm OK with this.

muursh (approved):
I'm fine with this

stanislav-tkach (approved):
I also find this logical, though it would be really interesting to see the benchmarking results.

PR #304, opened 2022-07-25
Assorted clippy fixes

iljakuklic:
Some simple changes suggested by clippy. <s>Mostly replacing `my_bool.then(|| something_trivial)` with `by_bool.then_some(something_trivial)` but</s> few other changes have been included as well.

### DIFF
--- a/common/src/chain/block/mod.rs
+++ b/common/src/chain/block/mod.rs
@@ -214,7 +214,7 @@ mod tests {
 
     fn check_block_tag(block: &Block) {
         let encoded_block = block.encode();
-        let first_byte = *encoded_block.get(0).unwrap();
+        let first_byte = *encoded_block.first().unwrap();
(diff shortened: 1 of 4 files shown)


TheQuantumPhysicist (approved):

muursh (approved):

altonen (approved):

iljakuklic:
Hmm may be an issue with me using too recent clippy, which suggests things that are still unstable in toolchains used by github CI. Will try to narrow this down to stuff that works there.

TheQuantumPhysicist:
> Hmm may be an issue with me using too recent clippy, which suggests things that are still unstable in toolchains used by github CI. Will try to narrow this down to stuff that works there.

Maybe we should use a toolchain file so that these issues don't happen?

stanislav-tkach (approved):

PR #305, opened 2022-07-25
Check wrapped crate usage

iljakuklic:
* Hack together a simple script to detect the direct use of `parity_scale_codec` and `jsonrpsee` in code
* Add a CI job for it
* Fix stuff it found

### DIFF
--- a/.github/scripts/codegrep.sh
+++ b/.github/scripts/codegrep.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+# Some simple custom code lints, mostly implemented by means of grepping code
+
+# Find rust source files, check if there are matching lines
+match_rs() {
+    grep --extended-regexp --with-filename --line-number \
+        --recursive --exclude-dir=.git --exclude-dir=target --include='*.rs' \
+        "$@"
+}
+
+disallow_except_in() {
+    echo
+    echo "==== Checking code does not contain '$1' ===="
+    match_rs --exclude-dir="$2" -e "$1"
+}
+
+FAIL=0
+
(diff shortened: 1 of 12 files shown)


TheQuantumPhysicist (approved):

altonen (approved):
Should this include direct use of `rand`/other crypto primitives too?

iljakuklic:
> Should this include direct use of `rand`/other crypto primitives too?

I think for `rand`, we can have something better than this. This is just a simple grepping of source code and would not detect some things. The reason is that the `serialization::{Encode, Decode}` macros as well as some macros re-exported from the RPC crate generate code that directly refer to `parity_scale_codec` and `jsonrpsee`, respectively.

If we want to disallow the direct usage of some methods completely, whether in source or inside macro, there is [this Clippy lint](https://rust-lang.github.io/rust-clippy/master/#disallowed_methods) that can be configured to do it. That should be somewhat more reliable than hacking together a shell script. But maybe has other drawbacks?

stanislav-tkach (approved):

.github/scripts/codegrep.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
TheQuantumPhysicist, line 1: Btw, while this works for now, I really dislike shell scripts for these kinds of jobs. We should use Python (or something more readable) for this.

iljakuklic: Agreed. Portability is not great either, in case someone wants to run it locally. I just made sure it runs in the CI.

PR #306, opened 2022-07-25
Introduce `WithId` wrapper and use it in ChainConfig for Genesis

iljakuklic:
* Introduce `WithId<T>` which behaves largely like `T` but also stores the pre-calculated ID. Calling `Idable::get_id` just returns the stored result instead of serializing and hashing the object.
* Use it in `ChainConfig` for Genesis (to store genesis itself and its pre-calculated ID)
* Also adjust `GenBlockIndex` to use it
* Can be used with any `Idable` type, e.g. the likes of `WithId<Transaction>` and `WithId<Block>` should also work, but that's not taken advantage of at the moment.

### DIFF
--- a/common/src/primitives/id.rs
+++ b/common/src/primitives/id.rs
@@ -143,8 +143,8 @@ impl<T: Idable> Idable for &T {
 /// An object together with its pre-calculated ID.
 ///
 /// This only allows immutable access to the underlying object to prevent it from going out of sync
-/// with the ID, which is calculated for its contents. Getting an ID is nearly cost-free.
-#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+/// with the ID, which is calculated for its contents. Getting an ID just returns the stored one.
+#[derive(Clone, Debug)]
 pub struct WithId<T: Idable> {
     id: Id<T::Tag>,
     object: T,
@@ -152,13 +152,18 @@ pub struct WithId<T: Idable> {
 
 impl<T: Idable> WithId<T> {
     /// Get a reference to the underlying object
-    pub fn get(&self) -> &T {
-        &self.object
+    pub fn get(this: &Self) -> &T {
+        &this.object
     }
 
     /// Get the pre-calucated object ID
-    pub fn id(&self) -> Id<T::Tag> {
-        self.id
+    pub fn id(this: &Self) -> Id<T::Tag> {
+        this.id
+    }
+
+    /// Take ownership of the underlying object
+    pub fn take(this: Self) -> T {
+        this.object
     }
 }
 
@@ -172,7 +177,7 @@ impl<T: Idable> WithId<T> {
 impl<T: Idable> Idable for WithId<T> {
     type Tag = T::Tag;
     fn get_id(&self) -> Id<Self::Tag> {
-        self.id()
+        Self::id(self)
     }
 }
 
@@ -180,7 +185,7 @@ impl<T: Idable> Idable for WithId<T> {
 impl<T: Idable> std::ops::Deref for WithId<T> {
     type Target = T;
     fn deref(&self) -> &T {
-        self.get()
+        Self::get(self)
     }
 }
 
--- a/common/src/primitives/id.rs
+++ b/common/src/primitives/id.rs
@@ -133,6 +133,57 @@ pub trait Idable {
     fn get_id(&self) -> Id<Self::Tag>;
 }
 
+impl<T: Idable> Idable for &T {
+    type Tag = T::Tag;
+    fn get_id(&self) -> Id<Self::Tag> {
+        (*self).get_id()
+    }
+}
+
+/// An object together with its pre-calculated ID.
+///
+/// This only allows immutable access to the underlying object to prevent it from going out of sync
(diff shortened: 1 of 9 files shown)


common/src/primitives/id.rs
```diff
@@ -133,6 +133,57 @@ pub trait Idable {
...
+    type Tag = T::Tag;
+    fn get_id(&self) -> Id<Self::Tag> {
+        (*self).get_id()
+    }
+}
+
+/// An object together with its pre-calculated ID.
+///
+/// This only allows immutable access to the underlying object to prevent it from going out of sync
+/// with the ID, which is calculated for its contents. Getting an ID is nearly cost-free.
```
stanislav-tkach, line 146: I think that the "near" word here is somewhat misleading. I have to immediately check if there is still some computation present. 🙃

stanislav-tkach (approved):

common/src/primitives/id.rs
```diff
@@ -133,6 +133,57 @@ pub trait Idable {
...
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub struct WithId<T: Idable> {
```
TheQuantumPhysicist, line 148: I was planning to create exactly this as a solution to id caching, so that we don't calculate it multiple times, especially during block processing. Thanks for going through with it.

Anyway, can you move this to its own rs file? I'm even OK with an `id` submodule.

common/src/primitives/id.rs
```diff
@@ -133,6 +133,57 @@ pub trait Idable {
...
+
+impl<T: Idable> Idable for WithId<T> {
+    type Tag = T::Tag;
+    fn get_id(&self) -> Id<Self::Tag> {
+        self.id()
+    }
+}
+
+// Implement Deref to the wrapped type but not DerefMut to prevent it from being modified.
+impl<T: Idable> std::ops::Deref for WithId<T> {
```
TheQuantumPhysicist, line 180: I know this simplifies calls a lot... by allowing references of WithId to act like T... but don't you think this is playing with fire, and can cause confusion? I usually lean towards being explicit to avoid unintended mistakes.

TheQuantumPhysicist (approved):
Thanks for doing this. You created this for a different purpose than what I had in mind. I was gonna create this for processing blocks as a caching mechanism.

common/src/primitives/id.rs
iljakuklic: The idea is to make `WithId<T>` a drop-in replacement for `T` for most part (unless `T` is being modified which will break its ID and so is not allowed). So you use the object with id caching in the same way as you would the normal one, just `get_id` becomes faster.

Regarding how dangerous it is, my initial instinct is that method resolution seems quite well-behaved in Rust. However, I will do some more research into it to see how likely are potential surprises. Do you have any particular potential sources of confusion in mind @TheQuantumPhysicist?

common/src/primitives/id.rs
iljakuklic: WIll move it.

iljakuklic:
> Thanks for doing this. You created this for a different purpose than what I had in mind. I was gonna create this for processing blocks as a caching mechanism.

Yea it is intended to be fairly generic, so it can be used in multiple places. Calculating IDs is pretty common.

common/src/primitives/id.rs
TheQuantumPhysicist: > Do you have any particular potential sources of confusion in mind @TheQuantumPhysicist?
Not really. It's instinct from experience. For example, in C++, it's a rule of thumb to not override the unary `operator*` and `operator->` unless we're dereferencing a pointer, as in creating a custom smart-pointer-like things. For something like our case, I usually create a `get()` method.

iljakuklic: Had a look around the standard library to see how they deal with types that implement `Deref`. These are broadly in two categories: data structures (e.g. `String` derefs into `&str`, `Vec<T>` into `&[T]`) and smart-pointer-like things (`Box<T>`, `MutexGuard`, `Rc`, etc.). The latter are more interesting when it comes to addressing potential method resolution issues because (like the `WIthId` type) they tend to be generic. Some observations:


They seem to be fairly generous implementing traits for these types, especially things like `AsRef<T>`, `Borrow<T>` and similar. Quite often, traits that just delegate to the underlying type are implemented too.

When it comes to methods, name collision is an issue. If there is a method called `get` on `T` and on `Wrapper<T>`, it may not be immediately clear which one is called. There seems to be a strong emerging convention to not implement inherent methods for wrappers like that, and use associated functions instead. So you have to write `Rc::strong_count(&rc)` rather than `rc.strong_count()`, which looks ambiguous if the underlying type also has a method called `strong_count`. Note that the `Rc` shows up in the method invocation so there is no confusion what is being called. I have changed `WithId` to use this convention too.

TheQuantumPhysicist: That sounds like a reasonable solution, albeit weird. We now know that all methods belong to the underlying T.

TheQuantumPhysicist (approved):

TheQuantumPhysicist (approved):

PR #308, opened 2022-07-26
Fix: p2p suggestions

TheQuantumPhysicist:
Some suggested changes to P2P after a discussion


altonen (approved):
LGTM but someone else should probably take a look too

p2p/src/message.rs
```diff
@@ -86,11 +86,11 @@ impl HeaderResponse {
 }
 
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
-pub struct BlockResponse {
+pub struct BlocksResponse {
```
muursh, line 89: Am I the only one who thinks the pluralisation of this sounds strange? `BlocksResponse` doesn't really make sense

altonen: No, I don't like it either but it was decided that we go with this

TheQuantumPhysicist: It's plural because it's about multiple blocks.

iljakuklic: Could it be something like `BlockListResponse` or `BlockSetResponse`? I find `BlocksResponse` a bit weird too.

TheQuantumPhysicist: I don't mind. Something like BlockListResponse is equally OK for me.

muursh: Yeah fine by me

muursh: I'd suggest the same for `HeadersResponse`

stanislav-tkach: I suppose we should also change the names of the corresponding requests (`BlockRequest` and `HeaderRequest`).

p2p/src/net/libp2p/service/pubsub.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        assert_eq!(1 + 1, 2);
+    }
+}
```
stanislav-tkach, line 141: I'm going to remove that. 🙃

stanislav-tkach (approved):

TheQuantumPhysicist:
Let's get this merged, as soon as possible, please.

altonen:
I think this is ready, are you guys OK with squash-merging this into master?

TheQuantumPhysicist:
> I think this is ready, are you guys OK with squash-merging this into master?

Quite frankly, I don't care, squash or not. Let's see if someone else has a different opinion. We talk latest about this during the daily.

muursh (approved):
I'm happy to merge this.

PR #309, opened 2022-07-26
storage: Separate into multiple crates

iljakuklic:
Now we have:
* `storage-core` with the backend-agnostic abstractions
* `storage-inmemory` with the in-memory implementation
* `storage` that collects and re-exports the above two

When using the top-level `storage`, the interface remains unchanged.


TheQuantumPhysicist (approved):
I would argue that storage-core doesn't need to be public. But that's OK too. Maybe we'll need it for tests? No idea.

stanislav-tkach (approved):
👍 

I not sure about the "core" name, but I'm unable to propose anything better.

iljakuklic:
> I would argue that storage-core doesn't need to be public. But that's OK too. Maybe we'll need it for tests? No idea.

The current idea `storage-core` crate contains the traits and interfaces that are used to write code in backend-agnostic way. WIll narrow the exports down if that changes as the exact design crystallises a bit more. For this PR, what is visible in the top-level `storage` crate is the same as it was before with the monolitinc `storage` crate.

Issue #182, opened 2022-05-18
Create a global block test framework


altonen:
Subsystems that communicate with consensus need to be able to create new blocks and chains. Currently this is implemented inside consensus using the `BlockTestFrameWork`. This framework should be exposed, maybe through a subsystem, so other subsystems could use it to create new blocks and test their own functionality.

iljakuklic:
I think @stanislav-tkach is currently working towards this, is that correct?

stanislav-tkach:
I haven't seen this issue before, but Sam mentioned to me that it would be nice to reuse this "test framework" in the p2p code. It seems logical to me, but as a second step. Right now I'm concentrating solely on the chainstate tests. But I would like to tackle this issue too.

stanislav-tkach:
Closed with https://github.com/mintlayer/mintlayer-core/pull/720.

PR #310, opened 2022-07-26
"Block test framework" refactoring

stanislav-tkach:
The goals of this pull request are the following:
- Cleanup the chainstate test framework slightly.
- Extent its api, so writing tests would be easier.
- Update the existing tests.

The next step is to fix the https://github.com/mintlayer/mintlayer-core/issues/182 issue.

### DIFF
--- a/chainstate/src/detail/tests/test_framework/block_builder.rs
+++ b/chainstate/src/detail/tests/test_framework/block_builder.rs
@@ -22,12 +22,13 @@ use common::{
 };
 
 use crate::{
-    detail::{tests::test_framework::BlockTestFramework, BlockIndex},
+    detail::{tests::test_framework::TestFramework, BlockIndex},
     Block, BlockError, BlockSource, GenBlock,
 };
 
-pub struct ProcessBlockBuilder<'f> {
-    framework: &'f mut BlockTestFramework,
+/// The block builder that allows construction and processing of a block.
+pub struct BlockBuilder<'f> {
+    framework: &'f mut TestFramework,
     transactions: Vec<Transaction>,
     prev_block_hash: Id<GenBlock>,
     timestamp: BlockTimestamp,
@@ -35,8 +36,8 @@ pub struct ProcessBlockBuilder<'f> {
     block_source: BlockSource,
 }
 
-impl<'f> ProcessBlockBuilder<'f> {
-    pub fn new(framework: &'f mut BlockTestFramework) -> Self {
+impl<'f> BlockBuilder<'f> {
+    pub fn new(framework: &'f mut TestFramework) -> Self {
         let transactions = Vec::new();
         let prev_block_hash = framework.chainstate.get_best_block_id().unwrap();
         let timestamp = BlockTimestamp::from_duration_since_epoch(time::get());
@@ -61,6 +62,16 @@ impl<'f> ProcessBlockBuilder<'f> {
     pub fn with_prev_block_hash(mut self, prev_block_hash: Id<GenBlock>) -> Self {
         self.prev_block_hash = prev_block_hash;
         self
+    }
+
+    pub fn build(self) -> Block {
+        Block::new(
+            self.transactions,
+            self.prev_block_hash,
+            self.timestamp,
+            self.consensus_data,
+        )
+        .unwrap()
     }
 
     pub fn process(self) -> Result<Option<BlockIndex>, BlockError> {
--- a/chainstate/src/detail/tests/test_framework/block_builder.rs
+++ b/chainstate/src/detail/tests/test_framework/block_builder.rs
@@ -81,13 +81,22 @@ impl<'f> BlockBuilder<'f> {
 
     /// Same as `add_test_transaction`, but with a custom parent.
     pub fn add_test_transaction_with_parent(
-        mut self,
+        self,
         parent: Id<GenBlock>,
         rng: &mut impl Rng,
     ) -> Self {
-        let (inputs, outputs): (Vec<_>, Vec<_>) = self.make_test_inputs_outputs(parent, rng);
-        self.transactions.push(Transaction::new(0, inputs, outputs, 0).unwrap());
-        self
+        let (inputs, outputs): (Vec<_>, Vec<_>) = self.make_test_inputs_outputs(
+            TestBlockInfo::from_id(&self.framework.chainstate, parent),
+            rng,
+        );
+        self.add_transaction(Transaction::new(0, inputs, outputs, 0).unwrap())
+    }
+
+    /// Same as `add_test_transaction_with_parent`, but uses reference to a block.
+    pub fn add_test_transaction_from_block(self, parent: &Block, rng: &mut impl Rng) -> Self {
+        let (inputs, outputs): (Vec<_>, Vec<_>) =
+            self.make_test_inputs_outputs(TestBlockInfo::from_block(parent), rng);
+        self.add_transaction(Transaction::new(0, inputs, outputs, 0).unwrap())
     }
 
     /// Adds a transaction that tries to spend the already spent output from the specified block.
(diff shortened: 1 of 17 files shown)


chainstate/src/detail/tests/test_framework/framework.rs
```diff
@@ -15,15 +15,156 @@
...
+}
+
+impl BlockTestFramework {
+    /// Creates a new `BlockTestFramework` instance using a builder api.
+    pub fn new() -> BlockTestFrameworkBuilder {
+        BlockTestFrameworkBuilder::new()
+    }
+
+    /// Processes a new block with the parameters specified using `ProcessBlockBuilder`.
+    pub fn process_block(&mut self) -> ProcessBlockBuilder {
```
iljakuklic, line 58: I think the name here is somewhat misleading, since no block processing actually happens upon calling this function. Maybe `process_block_builder`, or just `block_builder`?

chainstate/src/detail/tests/test_framework/framework.rs
```diff
@@ -15,15 +15,156 @@
...
+    pub chainstate: Chainstate,
+    // TODO: FIXME: Remove?..
+    pub block_indexes: Vec<BlockIndex>,
+}
+
+impl BlockTestFramework {
+    /// Creates a new `BlockTestFramework` instance using a builder api.
+    pub fn new() -> BlockTestFrameworkBuilder {
+        BlockTestFrameworkBuilder::new()
+    }
```
iljakuklic, line 55: Clippy is moaning about `new` not returning `Self`, just call the method `builder`, should be good enough

chainstate/src/detail/tests/test_framework/framework.rs
```diff
@@ -15,15 +15,156 @@
...
+        parent_block_id: &Id<GenBlock>,
+        count_blocks: usize,
+        rng: &mut impl Rng,
+    ) -> Result<Id<GenBlock>, BlockError> {
+        let mut prev_block = TestBlockInfo::from_id(&self.chainstate, *parent_block_id);
+
+        for _ in 0..count_blocks {
+            // TODO: FIXME:
+            let block = produce_test_block(prev_block, rng);
+            prev_block = TestBlockInfo::from_block(&block);
```
iljakuklic, line 74: The `TestBlockInfo` is rather hacked together to make the genesis/block split work. I don't like the fact that you always have to do this after creating a new block to get info for the next iteration. If you can think of a way of making this better or at least streamline the interface, that would be nice.

chainstate/src/detail/tests/test_framework/framework_builder.rs
```diff
@@ -0,0 +1,92 @@
...
+            self.custom_orphan_error_hook,
+            self.time_getter,
+        )
+        .unwrap();
+
+        BlockTestFramework {
+            chainstate,
+            block_indexes: Vec::new(),
+        }
+    }
```
iljakuklic, line 91: This looks more like `ChainstateBuilder`, only mentions `BlockTestFramework` as a last step when built. Maybe if the chainstate stuff was separated out, it might be useful in normal code too, not just tests.

chainstate/src/detail/tests/test_framework/process_block_builder.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub fn process(self) -> Result<Option<BlockIndex>, BlockError> {
+        let block = Block::new(
+            self.transactions,
+            self.prev_block_hash,
+            self.timestamp,
+            self.consensus_data,
+        )
+        .unwrap();
+        self.framework.chainstate.process_block(block, self.block_source)
+    }
```
iljakuklic, line 75: I might be useful to factor out the block creation part from here so it can be used on its own.

iljakuklic (commented):
Adding some initial unsolicited observations :stuck_out_tongue:

chainstate/src/detail/tests/test_framework/process_block_builder.rs
stanislav-tkach: Currently I have changed the `process_block` function to accept a block and the blocks construction is done by a block builder, but I'm still not sure. A block source is almost always `Local` in the tests, but without the "process block" builder (and optional arguments) we need to specify it every time.

stanislav-tkach:
@iljakuklic The opinions are always welcome!

chainstate/src/detail/tests/test_framework/process_block_builder.rs
iljakuklic: I was thinking of having a `build()` method that just returns the block *in addition* to the `process()` which would just call build and run chainstate. Would that be an option?

chainstate/src/detail/tests/test_framework/framework_builder.rs
iljakuklic: In principle, there could still be test framework builder that delegates to chainstate builder but that is probably not worth the extra boilerplate. There's also some merit in forcing the creator of chainstate to explicitly initialise everything in production code.

chainstate/src/detail/tests/test_framework/process_block_builder.rs
stanislav-tkach: I wrote a wall of text explaining that it would be a bad idea, but now I actually like that. 🙃

chainstate/src/detail/tests/test_framework/framework.rs
stanislav-tkach: Yes, we can have the genesis block only on the first iteration, but so far I was unable to find a solution that doesn't make the loop uglier... I'm going to try a little bit more, though.

iljakuklic: Some context (but not much) is here: https://github.com/mintlayer/mintlayer-core/pull/299#discussion_r927465928

stanislav-tkach: For now I'm going to add TODO notes, but I'm going to think about it.
(https://github.com/mintlayer/mintlayer-core/issues/312)

By the way, the link works for me.

stanislav-tkach:
There are still some rough edges that I want to fix, but it is mostly ready for review.

chainstate/src/detail/tests/test_framework/framework.rs
```diff
@@ -0,0 +1,197 @@
...
+        },
+        BlockIndex, GenBlockIndex, TimeGetter,
+    },
+    BlockError, BlockHeight, BlockSource, Chainstate, ChainstateConfig,
+};
+
+/// The `Chainstate` wrapper that simplifies operations and checks in the tests.
+pub struct TestFramework {
+    pub chainstate: Chainstate,
+    // TODO: FIXME: Remove?..
```
stanislav-tkach, line 42: 👀

stanislav-tkach: I have made several attempts to remove this field, but it is heavily used in the reorg tests and sometimes it will require to almost completely rewrite them. At the moment I feel like doing that isn’t worth it.

chainstate/src/detail/tests/test_framework/block_builder.rs
```diff
@@ -0,0 +1,180 @@
...
+            self.transactions,
+            self.prev_block_hash,
+            self.timestamp,
+            self.consensus_data,
+        )
+        .unwrap()
+    }
+
+    /// Constructs a block and processes it by the chainstate.
+    pub fn process(self) -> Result<Option<BlockIndex>, BlockError> {
```
TheQuantumPhysicist, line 157: I would call this build_and_process()

chainstate/src/detail/tests/test_framework/framework.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+impl TestFramework {
+    /// Creates a new test framework instance using a builder api.
+    pub fn builder() -> TestFrameworkBuilder {
+        TestFrameworkBuilder::new()
+    }
+
+    /// Returns a block builder instance that can be used for a block construction and processing.
+    pub fn block_builder(&mut self) -> BlockBuilder {
```
TheQuantumPhysicist, line 50: make_block_builder()? maybe new_block_builder()? or better, derive_block_builder()?

TheQuantumPhysicist (approved):

azarovh (approved):

PR #311, opened 2022-07-27
net: Clean up `ConnectivityEvent` enum

altonen:
Rename some events to be more logical, use consistent name for address,
and remove an old event that was not used for anything.

### DIFF
--- a/p2p/src/net/libp2p/types.rs
+++ b/p2p/src/net/libp2p/types.rs
@@ -133,20 +133,20 @@ pub enum Command {
 #[derive(Debug)]
 pub enum ConnectivityEvent {
     /// Outbound connection accepted by remote
-    ConnectionAccepted {
-        addr: Multiaddr,
+    OutboundAccepted {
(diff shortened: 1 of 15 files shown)


p2p/src/net/libp2p/types.rs
```diff
@@ -159,9 +159,6 @@ pub enum ConnectivityEvent {
     /// One or more peers that were previously discovered have expired
     Expired { peers: Vec<(PeerId, Multiaddr)> },
```
TheQuantumPhysicist, line 161: So now there's no event for when a peer gets disconnected? Or are we gonna use the same ConnectionClosed for that?

TheQuantumPhysicist (approved):

p2p/src/net/libp2p/types.rs
altonen: `ConnectionClosed`

muursh (approved):

PR #315, opened 2022-07-28
Change genesis block index to Arc to solve ownership issue when isolating chainstate interfaces

TheQuantumPhysicist:
Using a reference for genesis creates an issue with traits that are supposed to isolate chainstate interface from block index accesses. Using an Arc seems to be a reasonable solution.

### DIFF
--- a/chainstate/src/detail/spend_cache/mod.rs
+++ b/chainstate/src/detail/spend_cache/mod.rs
@@ -15,7 +15,10 @@
 //
 // Author(s): S. Afach
 
-use std::collections::{btree_map::Entry, BTreeMap};
+use std::{
+    collections::{btree_map::Entry, BTreeMap},
+    sync::Arc,
+};
 
 use super::gen_block_index::GenBlockIndex;
 use chainstate_storage::{BlockchainStorageRead, BlockchainStorageWrite};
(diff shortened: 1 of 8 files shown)


chainstate/src/detail/spend_cache/mod.rs
```diff
@@ -114,11 +117,11 @@ impl<'a, S: BlockchainStorageRead> CachedInputs<'a, S> {
...
         &self,
         block_id: &Id<GenBlock>,
-    ) -> Result<Option<GenBlockIndex<'a>>, StateUpdateError> {
+    ) -> Result<Option<GenBlockIndex>, StateUpdateError> {
         match block_id.classify(self.chain_config) {
-            GenBlockId::Genesis(_id) => Ok(Some(GenBlockIndex::Genesis(
+            GenBlockId::Genesis(_id) => Ok(Some(GenBlockIndex::Genesis(Arc::clone(
                 self.chain_config.genesis_block(),
-            ))),
+            )))),
```
iljakuklic, line 124: Almost thought this was LISP :laughing:

iljakuklic (approved):
Looks good. While this introduces some small amount of overhead that is not strictly necessary, I think the simpler interface and simpler types are worth it.

PR #319, opened 2022-07-28
Fix libp2p tests

altonen:
Fix flaky GossipSub test and increase the retry count for `connect_swarms()` to 3

### DIFF
--- a/p2p/src/net/libp2p/tests/gossipsub.rs
+++ b/p2p/src/net/libp2p/tests/gossipsub.rs
@@ -123,6 +123,9 @@ async fn test_gossipsub_not_supported() {
                 ) => {
                     break;
                 }
+                SwarmEvent::ConnectionClosed { .. } => {
+                    break;
+                }
                 _event => {},
             },
             _event = swarm.next() => {},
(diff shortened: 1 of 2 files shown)


TheQuantumPhysicist (approved):

muursh (approved):

iljakuklic (approved):

stanislav-tkach (approved):

stanislav-tkach:
Looks like we have issues with another test now:
> ---- net::libp2p::tests::identify::test_identify_not_supported stderr ----
thread 'main' panicked at 'didn't receive ConnectionEstablished event in time', p2p/src/net/libp2p/tests/mod.rs:278:17

altonen:
I'm actually a little disturbed by that error because it means there is something wrong on the libp2p side. Anyway, I pushed a change that hopefully fixes this issue in our CI.

On a related note, I would like to start using Prometheus/some other tracing framework in P2P because sometimes it's challenging to reason about all the changes that are made, especially on the libp2p side, and how they affect the functionality so it would be nice to have metrics about the incoming data and whether something is lost along the way to the front-end of the P2P subsystem.

Thoughts?

PR #320, opened 2022-07-29
chainstate-storage: Make Storage generic

iljakuklic:
Just another couple of tiny wee baby steps towards the support of multiple storage backends.

* Makes the implementation more agnostic to storage backend by making it generic
* The type that other crates use to interface storage is still just an alias to the in-memory storage engine. Abstracting that will be a follow-up PR.
* Also shuffles some module names around

### DIFF
--- a/chainstate-storage/src/internal/mod.rs
+++ b/chainstate-storage/src/internal/mod.rs
@@ -54,7 +54,7 @@ mod well_known {
 
 storage::decl_schema! {
     // Database schema for blockchain storage
-    Schema {
+    pub(crate) Schema {
         // Storage for individual values.
         pub DBValue: Single,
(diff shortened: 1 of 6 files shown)


chainstate-storage/src/internal/mod.rs
```diff
@@ -417,10 +416,12 @@ pub(crate) mod test {
     use test_utils::random::{make_seedable_rng, Seed};
     use utxo::{BlockUndo, TxUndo};
 
+    type TestStore = Store<storage::inmemory::Store<Schema>>;
```
TheQuantumPhysicist, line 419: Please call it InMemoryStore. This is not necessarily just for tests.

PR #321, opened 2022-07-29
Move modules around a bit in chainstate-storage

iljakuklic:
* Move tests into a separate file/module
* Move `utxo_db` into the internal part of the module hierarchy. It makes more sense because it builds on top of the other part of the storage implementation. Also the test submodule does not need to be `pub(crate)` this way.


TheQuantumPhysicist (approved):

chainstate-storage/src/internal/test.rs
```diff
@@ -0,0 +1,329 @@
+use super::*;
```
stanislav-tkach, line 1: The license header is missing. 🙃

stanislav-tkach (approved):

muursh (approved):

sinitcin (approved):

PR #322, opened 2022-07-31
A few fixes related to utxo in preparation to use it in chainstate

TheQuantumPhysicist:


chainstate-types/src/storage_result.rs
```diff
@@ -0,0 +1,15 @@
+/// Blockchain storage error
```
azarovh, line 1: copyright?

utxo/src/utxo_impl/utxo_storage/in_memory.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::{BTreeMap, HashMap};
```
azarovh, line 1: copyright and in other files. Btw can we automate this routine?

azarovh (approved):

utxo/src/utxo_impl/utxo_storage/in_memory.rs
```diff
@@ -0,0 +1,71 @@
...
+use common::{
+    chain::{Block, GenBlock, OutPoint},
+    primitives::Id,
+};
+
+#[derive(Clone)]
+pub struct UtxosDBInMemoryImpl {
+    store: BTreeMap<OutPoint, Utxo>,
+    undo_store: BTreeMap<Id<Block>, BlockUndo>,
+    best_block_id: Option<Id<GenBlock>>,
```
iljakuklic, line 15: Why is this an `Option`? Would it be appropriate to just initialise best block ID upon creation and be sure one exists from that point on? Also how is this different from best block ID in regular chainstate?

utxo/src/utxo_impl/utxo_storage/test.rs
```diff
@@ -0,0 +1,356 @@
...
+    // distribute the inputs on the number of the transactions specified.
+    let input_size = inputs.len() / num_of_txs;
+
+    // create the multiple transactions based on the inputs.
+    inputs
+        .chunks(input_size)
+        .into_iter()
+        .map(|inputs| {
+            let outputs = if max_num_of_outputs > 1 {
+                let rnd = make_pseudo_rng().gen_range(1..max_num_of_outputs);
```
iljakuklic, line 32: Maybe the new seedable RNG mechanism should be used here.

utxo/src/utxo_impl/utxo_storage/view_impls.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub fn utxo<S: UtxosStorageRead>(db: &S, outpoint: &OutPoint) -> Option<Utxo> {
+        match db.get_utxo(outpoint) {
+            Ok(res) => res,
+            Err(e) => {
+                panic!(
+                    "Database error while attempting to retrieve utxo from the database: {}",
+                    e
+                );
+            }
+        }
```
iljakuklic, line 28: Not sure how much better it is but I think the pattern here could be replaced with something like;
```rust
db.get_utxo(outpoint).unwrap_or_else(|e| panic!(...))
```

utxo/src/utxo_impl/test.rs
```diff
@@ -420,10 +420,10 @@ fn derive_cache_test() {
     let mut cache = UtxosCache::default();
 
     let (utxo, outpoint_1) = create_utxo(10);
-    assert!(cache.add_utxo(utxo, &outpoint_1, false).is_ok());
+    assert!(cache.add_utxo(&outpoint_1, utxo, false).is_ok());
```
iljakuklic, line 423: It should be considered changing the last argument of `add_utxo` from a bool to a two-state enum so it's clearer what's going on in the call sites.

iljakuklic (approved):
Added a couple of minor comments. All in all, good improvement.

stanislav-tkach (approved):

utxo/src/utxo_impl/utxo_storage/in_memory.rs
TheQuantumPhysicist: I'll give it a shot to remove the Option.

Btw, you asked the question of why it's not the same best block before. The answer is that the "utxo set" representation is isolated and separate from the rest of the database. The best block id, as part of the UtxosView is a representation of "what utxo set is represented here". Now, when we combine that with the rest of the node, then we can say that the whole BlockchainStorage is represented by the same best block. So to make them one, this should be done at the BlockchainStorage implementation level, but then if we do that, then this could get the utxo set wrapper (UtxosDB) to replace the BlockchainStorage value that's given by the block connection process, which is called when activating best chain. So, we either merge everything utxo related to the BlockchainStorage and get rid of this interface (which will kill the beautiful hierarchy created with UtxosCache), or we live with having two best blocks because both represent the same thing in two different contexts. Maybe there's a better solution I'm not seeing. I don't know.

utxo/src/utxo_impl/test.rs
TheQuantumPhysicist: I added a TODO on that.

utxo/src/utxo_impl/utxo_storage/test.rs
TheQuantumPhysicist: Another TODO

PR #324, opened 2022-08-01
Add unit tests and small code enhancements

azarovh:
Collected the changes I made while reading the code.

Nothing special, the only interesting change is I swapped the order of processing `DuplicateInputInTransaction` and `DuplicateInputInBlock` because previously one was shadowed by another and was effectively unreachable. Noticed that while writing a test.

### DIFF
--- a/chainstate/src/detail/chainstateref.rs
+++ b/chainstate/src/detail/chainstateref.rs
@@ -445,14 +445,14 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
             for tx in block.transactions() {
                 let mut tx_inputs = BTreeSet::new();
                 for input in tx.inputs() {
-                    if !block_inputs.insert(input.outpoint()) {
-                        return Err(CheckBlockTransactionsError::DuplicateInputInBlock(
-                            block.get_id(),
-                        ));
-                    }
                     if !tx_inputs.insert(input.outpoint()) {
                         return Err(CheckBlockTransactionsError::DuplicateInputInTransaction(
                             tx.get_id(),
+                            block.get_id(),
+                        ));
+                    }
+                    if !block_inputs.insert(input.outpoint()) {
+                        return Err(CheckBlockTransactionsError::DuplicateInputInBlock(
                             block.get_id(),
                         ));
                     }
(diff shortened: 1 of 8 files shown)


stanislav-tkach (approved):

chainstate/src/detail/chainstateref.rs
```diff
@@ -445,17 +445,17 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
                         return Err(CheckBlockTransactionsError::DuplicateInputInTransaction(
                             tx.get_id(),
                             block.get_id(),
                         ));
                     }
+                    if !block_inputs.insert(input.outpoint()) {
+                        return Err(CheckBlockTransactionsError::DuplicateInputInBlock(
+                            block.get_id(),
+                        ));
+                    }
```
iljakuklic, line 458: `ensure!` could be used here

chainstate/src/detail/tests/double_spend_tests.rs
```diff
@@ -220,6 +220,97 @@ fn spend_bigger_output_in_the_same_block(#[case] seed: Seed) {
...
+        let mut tf = TestFramework::default();
+
+        let mut rng = make_seedable_rng(seed);
+        let tx1_output_value = rng.gen_range(100_000..200_000);
+        let first_tx = tx_from_genesis(tf.genesis(), &mut rng, tx1_output_value);
+
+        let input = TxInput::new(first_tx.get_id().into(), 0, InputWitness::NoSignature(None));
+        let second_tx = TransactionBuilder::new()
+            .add_input(input.clone())
+            .add_input(input)
```
iljakuklic, line 250: Maybe also test the spend of the same outpoint twice, i.e. using two different witnesses. Not sure how easy will it be to find an example of that currently, is there a way to tinker with the witness? You can put some data into `InputWitness::NoSIgnature` but not sure what's that for.

iljakuklic (approved):

chainstate/src/detail/tests/double_spend_tests.rs
azarovh: Optional data is exactly for that purpose

TheQuantumPhysicist (approved):

PR #332, opened 2022-08-02
Move the block reward from the header to the block itself

stanislav-tkach:


stanislav-tkach:
The tests passed, so assuming that we had enough coverage before, all important invariants aren't broken. I don't mind adding more tests if you point out to me what exactly should be tested, though.

common/src/chain/block/block_v1.rs
```diff
@@ -13,21 +13,25 @@
...
+    },
+};
+
+use super::{block_header::BlockHeader, timestamp::BlockTimestamp};
 
 #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, serialization::Tagged)]
 pub struct BlockV1 {
     pub(super) header: BlockHeader,
+    pub(super) reward: BlockReward,
     pub(super) transactions: Vec<Transaction>,
```
iljakuklic, line 35: Would it make sense to introduce a `BlockBody` type that couples together reward and transactions?

iljakuklic (commented):
I see some relatively major issues with this as it stands.

First, we might want to impose some limit on the rewards output size. It is still subject to the total block size limit as far as I can tell. However previously, it was subject to block header size limit which is much tighter. This should be enforced by consensus (and maybe some limit should be in p2p as well to avoid sending a ton of data just to be rejected).

Second, and more importantly, I don't see the block rewards vector being committed to in the header. It includes a merkle root of transactions but not rewards. Do we want a separate merkle tree for rewards or do we want to include them somehow in existing ones (would require a new abstraction). Maybe just a plain hash of the rewards vector would be good enough? That would be the simplest way to start with. Either way, the rewards vector has to be committed to somehow and included in the header and the integrity has to be checked.

Also, it would be nice to test the issues mentioned above first. For example, you can generate a block, check it verifies, modify the rewards vector somehow (decrease the amount for example) and check it no longer verifies. In general, I'd say the scenarios where a block is supposed to be rejected are somewhat undertested but no less important than processing a well-formed block.

TheQuantumPhysicist:
> I see some relatively major issues with this as it stands.
> 
> First, we might want to impose some limit on the rewards output size. It is still subject to the total block size limit as far as I can tell. However previously, it was subject to block header size limit which is much tighter. This should be enforced by consensus (and maybe some limit should be in p2p as well to avoid sending a ton of data just to be rejected).
> 
> Second, and more importantly, I don't see the block rewards vector being committed to in the header. It includes a merkle root of transactions but not rewards. Do we want a separate merkle tree for rewards or do we want to include them somehow in existing ones (would require a new abstraction). Maybe just a plain hash of the rewards vector would be good enough? That would be the simplest way to start with. Either way, the rewards vector has to be committed to somehow and included in the header and the integrity has to be checked.
> 
> Also, it would be nice to test the issues mentioned above first. For example, you can generate a block, check it verifies, modify the rewards vector somehow (decrease the amount for example) and check it no longer verifies. In general, I'd say the scenarios where a block is supposed to be rejected are somewhat undertested but no less important than processing a well-formed block.

Merkle roots in the header should not be optional anymore, and in the absence of transactions, the hash of the reward should go in them.

There's no need to imposed a specific size for the reward because that's already done in the total block size.

iljakuklic:
> Merkle roots in the header should not be optional anymore, and in the absence of transactions, the hash of the reward should go in them.

I think this requires a bit more thought. do we treat rewards + transactions as one vector the merkle tree is superimposed upon? If so, an abstraction that captures the either reward or transaction concept has to be introduced, even if it's as simple as just hashing the two with the same hash function.

> There's no need to imposed a specific size for the reward because that's already done in the total block size.

Ok, just pointing out that the limit has been implicitly relaxed by making this change. If block rewards taking up the whole block is fine, we can keep it like that.

stanislav-tkach:
> Merkle roots in the header should not be optional anymore, and in the absence of transactions, the hash of the reward should go in them.

Does that mean that we shouldn't allow blocks without the reward?

TheQuantumPhysicist:
> Does that mean that we shouldn't allow blocks without the reward?

Nope. Just hash the serialized block reward.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -643,6 +632,29 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+            GenBlockId::Genesis(_) => Ok(self
+                .chain_config
+                .genesis_block()
+                .block_reward_transactable()
+                .outputs()
+                .unwrap_or(&[])
+                .to_vec()),
+            GenBlockId::Block(id) => Ok(self
+                .db_tx
+                .get_block(id)?
```
TheQuantumPhysicist, line 649: Getting the whole block just for the reward... not the best. Please throw in a TODO there to investigate how to improve this later.

TheQuantumPhysicist (approved):
From my relatively shallow review from my smartphone, this seems OK. I hope I didn't miss anything. I encourage others to review as well.

chainstate/src/detail/transaction_verifier/mod.rs
stanislav-tkach: https://github.com/mintlayer/mintlayer-core/issues/344

iljakuklic (approved):

azarovh (approved):

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -613,7 +602,7 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
                 }
             }
             BlockTransactableRef::BlockReward(block) => {
-                let reward_transactable = block.header().block_reward_transactable();
+                let reward_transactable = block.block_reward_transactable();
```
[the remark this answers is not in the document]
stanislav-tkach: https://github.com/mintlayer/mintlayer-core/issues/344

common/src/chain/block/block_reward.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://spdx.org/licenses/MIT
```
muursh, line 8: We've changed the url for this

muursh (approved):

common/src/chain/block/block_reward.rs
stanislav-tkach: https://github.com/mintlayer/mintlayer-core/pull/351#discussion_r940640078

Issue #219, opened 2022-06-07
Fix the stack_flush_test test


stanislav-tkach:
The `stack_flush_test` (utxo/src/utxo_impl/simulation.rs) test is currently ignored.

PR #333, opened 2022-08-03
Fix stack_flush_test

azarovh:
Fix for #219 
In order to fix the test, I had to rework the code in the crate to use seedable ranges so that the test became reproducible.


utxo/src/utxo_impl/simulation.rs
```diff
@@ -15,86 +15,69 @@
...
+        if rng.gen::<bool>() && existing_outpoints.len() > 1 {
             // just call the `spend_utxo`. Does not matter if it removes the outpoint entirely,
             // or just mark it as `spent`,
-            let outp_idx = make_pseudo_rng().gen_range(0..existing_outpoints.len());
+            let outp_idx = rng.gen_range(0..existing_outpoints.len());
             let to_spend = &existing_outpoints[outp_idx];
-            assert!(cache.spend_utxo(to_spend).is_ok());
-
-            //println!("child, spend: {:?}, removed", to_spend);
+            let _ = cache.spend_utxo(to_spend);
```
azarovh, line 72: The fix is basically the removal of this assertion. I see no value in panicing if we try to spend a utxo but it was already spent. This happens pretty frequently because utxo is chosen randomly from existing ones that could already be spent. Moreover, the purpose of this function is to "populate cache" - to make some random chaotic actions.

stanislav-tkach (approved):

azarovh:
this fix is not enough. Need to investigate further

azarovh:
I reworked stack_flush_test almost completely mimicking the coins_cache_simulation_test from bitcoin

stanislav-tkach (approved):

utxo/src/utxo_impl/test.rs
```diff
@@ -252,177 +272,192 @@ fn check_get_mut_utxo(
...
+#[trace]
+#[case(Seed::from_entropy())]
 #[rustfmt::skip]
-fn add_utxo_test() {
+fn add_utxo_test(#[case] seed: Seed) {
+    let mut rng = make_seedable_rng(seed);
     /*
-                CACHE PRESENCE CACHE Flags      Possible    RESULT flags       RESULT of `add_utxo` method
-                                                Overwrite
+                             CACHE      CACHE Flags      Possible    RESULT flags       RESULT of `add_utxo` method
```
muursh, line 282: I quite like aligning this over the fields below

muursh (approved):
All fine by me

TheQuantumPhysicist (approved):

PR #337, opened 2022-08-03
Output value enum

sinitcin:
This PR provides `TxOutput.value` as the enum. Now, `value` might contain a coin or a token. 

Tokens do not support at this PR, it's just changing fields of `TxOutput` that add fewer conflicts.


altonen (approved):
LGTM

stanislav-tkach (approved):

azarovh (approved):

common/src/chain/tokens/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+// Author(s): A. Sinitsyn
+
+use serialization::{Decode, Encode};
+
+use crate::primitives::Amount;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
+pub enum OutputValue {
+    Coin(Amount),
+    Token(TokenData),
```
TheQuantumPhysicist, line 25: I think it's better not to have this here. We want to see all the modifications related to this when it's added later.

p2p/p2p-test-utils/src/lib.rs
```diff
@@ -112,13 +113,20 @@ fn create_utxo_data(
     index: usize,
     output: &TxOutput,
 ) -> Option<(TxInput, TxOutput)> {
-    let new_value = (output.value() - Amount::from_atoms(1)).unwrap();
+    let output_value = match output.value() {
+        OutputValue::Coin(coin) => *coin,
+        OutputValue::Token(_) => return None,
```
TheQuantumPhysicist, line 118: Like here...

TheQuantumPhysicist (approved):
Besides removing the token arm, I think this is fine.

common/src/chain/tokens/mod.rs
iljakuklic: Agreed. Let's have just one variant here for now, and 1-way matches where `OutputValue` is being analysed. When you introduce a new variant later on, the places that need updating will be flagged up by the compiler.

iljakuklic (approved):

PR #338, opened 2022-08-03
ci: Rewrite custom code checks in Python and add a license header check

iljakuklic:
* Hopefully more robust and portable
* <s>Will fix the missing license headers once I see it fail in CI to check it works</s> DONE


TheQuantumPhysicist (approved):
Thank you for moving from bash to python!

muursh (approved):

.github/scripts/codecheck.py
```diff
@@ -0,0 +1,81 @@
...
+def run_checks():
+    ok = True
+    ok = ok and disallow(SCALECODEC_RE, exclude = ['serialization/core'])
+    ok = ok and disallow(JSONRPSEE_RE, exclude = ['rpc'])
+    ok = ok and check_licenses()
+    return ok
+
+if __name__ == '__main__':
+    if not run_checks():
+        sys.exit(1)
```
TheQuantumPhysicist, line 81: perhaps it's better to raise exceptions on errors. That will show the error on spot and exit the program.

stanislav-tkach (approved):

.github/scripts/codecheck.py
iljakuklic: I wanted to let the program list all code problems it finds during the run, without stopping at first one.

TheQuantumPhysicist: Alrighty. That's fine.

PR #339, opened 2022-08-04
A few license fixes

muursh:
Does a few small things:

- Changes the URL in the license header to our actual license, because why not
- The license headers had slightly different formatting sometimes. Some license URLs had an extra space before, some didn't. 
- Changed the regex for checking if the license header is valid. In theory someone could contribute code that wouldn't have the copyright assigned to RBB. Now we just check it's assigned somewhere. Could be a person, could be a company. We don't care. As long as there's something there then we're happy.


stanislav-tkach (approved):

altonen (approved):
Should we drop the "Author(s)" part from the licenses as well as now we sometimes have it and sometimes don't and it's not strictly part of the MIT license?

muursh:
> Should we drop the "Author(s)" part from the licenses as well as now we sometimes have it and sometimes don't and it's not strictly part of the MIT license?

This came up before. I think @stanislav-tkach mentioned it. I'm more than happy to scrap it, it doesn't really add a lot right now other than telling us who to prod when something breaks but git can tell us that anyway and we can always change how we use code owners to give ownership of subsystems to specific people etc... If nobody has any issues with that proposal I'll remove it in this PR.

.github/scripts/codecheck.py
```diff
@@ -9,14 +9,14 @@
...
-    r'// Copyright \(c\) 202[0-9](-202[0-9])? RBB S\.r\.l',
+    r'// Copyright \(c\) 202[0-9](-202[0-9])? .+',
     r'// opensource@mintlayer\.org',
     r'// SPDX-License-Identifier: MIT',
     r'// Licensed under the MIT License;',
     r'// you may not use this file except in compliance with the License\.',
     r'// You may obtain a copy of the License at',
     r'//',
-    r'//\s+http://spdx\.org/licenses/MIT',
+    r'//\s+https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE',
```
iljakuklic, line 19: Should this regex be tightened up to require one space exactly rather than some whitespace before the URL to enforce more consistency?

iljakuklic (approved):
Looks good. I am in favour of removing the authors section as long as there are no legal issues with it. It gets stale pretty quickly anyway.

.github/scripts/codecheck.py
iljakuklic, line 19: One more point about the URL. This points to the LICENSE file in the `master` branch. In case re-licensing is needed for some reason, the URL might start pointing to a different license than what the header mentions. Should it point to particular tag or commit instead? I am not sure how likely this scenario is or what should happen if it comes up, so maybe I am off the mark completely here.

muursh: Yeah it's probably worth it. I'll fix it up

muursh: If we re-license we'll have a load of other things to change too, `r'// SPDX-License-Identifier: MIT',` for example. What you're saying is true but given how many other things will require changing it probably saves a few minutes of time at most.

PR #340, opened 2022-08-04
Use WithId in TxMempoolEntry

royster57:
Small refactoring of the `TxMempoolEntry` struct to make use of the `WithId` caching mechanism.

### DIFF
--- a/common/src/primitives/id/with_id.rs
+++ b/common/src/primitives/id/with_id.rs
@@ -21,7 +21,7 @@ use super::{Id, Idable};
 ///
 /// This only allows immutable access to the underlying object to prevent it from going out of sync
 /// with the ID, which is calculated for its contents. Getting an ID just returns the stored one.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct WithId<T: Idable> {
     id: Id<T::Tag>,
     object: T,
--- a/mempool/src/pool.rs
+++ b/mempool/src/pool.rs
@@ -14,6 +14,7 @@ use common::chain::transaction::Transaction;
 use common::chain::transaction::TxInput;
 use common::chain::OutPoint;
 use common::primitives::amount::Amount;
+use common::primitives::id::WithId;
 use common::primitives::Id;
 use common::primitives::Idable;
 use common::primitives::H256;
@@ -140,7 +141,7 @@ newtype! {
 
 #[derive(Debug, Eq, Clone)]
 struct TxMempoolEntry {
-    tx: Transaction,
+    tx: WithId<Transaction>,
     fee: Amount,
     parents: BTreeSet<Id<Transaction>>,
     children: BTreeSet<Id<Transaction>>,
@@ -157,7 +158,7 @@ impl TxMempoolEntry {
         creation_time: Time,
     ) -> TxMempoolEntry {
         Self {
-            tx,
+            tx: WithId::new(tx),
             fee,
             parents,
             children: BTreeSet::default(),
@@ -172,7 +173,7 @@ impl TxMempoolEntry {
     }
 
     fn tx_id(&self) -> Id<Transaction> {
-        self.tx.get_id()
+        WithId::id(&self.tx)
     }
 
     fn unconfirmed_parents(&self) -> impl Iterator<Item = &Id<Transaction>> {
@@ -1109,7 +1110,7 @@ where
             .txs_by_descendant_score
             .values()
             .flatten()
-            .map(|id| &self.store.get_entry(id).expect("entry").tx)
+            .map(|id| WithId::get(&self.store.get_entry(id).expect("entry").tx))
             .collect()
     }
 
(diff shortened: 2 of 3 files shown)


common/src/primitives/id/with_id.rs
```diff
@@ -29,6 +29,26 @@ pub struct WithId<T: Idable> {
...
+where
+    T: Idable,
+    T: PartialEq,
+    <T as Idable>::Tag: PartialEq,
+{
+    fn eq(&self, other: &WithId<T>) -> bool
+where {
+        self.id == other.id
+    }
+}
```
iljakuklic, line 42: I don't think it's a good idea to introduce the instance with this particular implementation. Ideally, the `WithId` type behaves completely transparently, i.e. `(WithId::new(x) == WIthId::new(y)) == (x == y)` for all `x` and `y`. In other words, the `==` operator should behave the same on `WithId` as on the original value. Sadly in `Transaction` some fields are excluded from ID, so the property is violated.

Conclusions:
* If you want an `Eq + PartialEq` instance on `WithId`, just let it generate the auto-derived version
* If you want to compare by IDs, use something like `x.get_id() == y.get_id()`, it's more explicit about what exactly is being compared.

mempool/src/pool.rs
```diff
@@ -172,7 +173,7 @@ impl TxMempoolEntry {
     }
 
     fn tx_id(&self) -> Id<Transaction> {
-        self.tx.get_id()
+        WithId::id(&self.tx)
```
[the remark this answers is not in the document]
iljakuklic: On the other hand, there is an advantage to using `WithId::id` in that it gets flagged up by the compiler if `WithId` is no longer used here, potentially making the ID calculation more expensive. I leave the decision to your judgement.

iljakuklic (approved):

stanislav-tkach (approved):

muursh (approved):
Fine by me

common/src/primitives/id/with_id.rs
```diff
@@ -21,7 +21,7 @@ use super::{Id, Idable};
 ///
 /// This only allows immutable access to the underlying object to prevent it from going out of sync
 /// with the ID, which is calculated for its contents. Getting an ID just returns the stored one.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
```
TheQuantumPhysicist, line 24: Perhaps we can manually implement PartialEq and make it only compared the id.

mempool/src/pool.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I don't really get what's wrong with using get_id

TheQuantumPhysicist (approved):

common/src/primitives/id/with_id.rs
royster57: This would seem to conflict with Lukas's comment above, that we want `(x == y) == (WithId(x) == WithId(y))`. unless I'm missing something...

TheQuantumPhysicist: They should be the same, since idable by definition is implemented for things that can have a unique id. What am I missing, @iljakuklic ?

iljakuklic: I think Roy is referring to my earlier comment here (modifying a bit to add some context):

> Ideally, the WithId type behaves completely transparently, i.e. (WithId::new(x) == WIthId::new(y)) == (x == y) for all x and y. In other words, the == operator should behave the same on WithId as on the original value. Sadly in Transaction some fields are excluded from ID, so the property is violated [if we only compare IDs].
> 
> Conclusions:
>
> * If you want an Eq + PartialEq instance on WithId, just let it generate the auto-derived version
> * If you want to compare by IDs, use something like x.get_id() == y.get_id(), it's more explicit about what exactly is being compared.

Or did I get this wrong and `Idable` for `Transaction` does include the witness?

TheQuantumPhysicist: Even if it doesn't include the witness, it's basically the same transaction. That's my point. The only reason we omit the witness is because it's malleable. 

This makes me think that we should treat this like we treated the BlockIndex. There's no equality there and equality should come by comparing IDs.

So now we have two options:

1. We either accept the id as a unique and identifiable thing that defines an object, and hence use id for equality by default
2. We drop the equality trait because it makes things more expensive and ambiguous, and the caller decides whether they want to compare the inner objects by getting the reference, or just use the id.

I slightly lean towards the second option because I like things being explicit. Either way, that derive has to go in my opinion. It's a bad idea.

iljakuklic: > So now we have two options:
> 
> 1. We either accept the id as a unique and identifiable thing that defines an object, and hence use id for equality by default
> 2. We drop the equality trait because it makes things more expensive and ambiguous, and the caller decides whether they want to compare the inner objects by getting the reference, or just use the id.
> 
> I slightly lean towards the second option because I like things being explicit. Either way, that derive has to go in my opinion. It's a bad idea.

Of these two, I'd be in favour of the second option too. We have to check to what extent is the comparison used in tests but that will be pointed out by the compiler once it's removed. The larger point is that operations for `WithId<T>` should be consistent with operations on `T` directly.

PR #342, opened 2022-08-04
Add syncing support for mock interface

altonen:
Add request-response support for the mock interface and rewrite all syncing-related tests to be executed both with mock interface and libp2p.

### DIFF
--- a/p2p/src/net/mock/request_manager.rs
+++ b/p2p/src/net/mock/request_manager.rs
@@ -32,41 +32,18 @@ use crate::{
 use std::collections::{hash_map::Entry, HashMap, HashSet};
 
 #[derive(Debug, Default)]
-struct PeerContext {
-    ephemerals: HashSet<types::MockRequestId>,
-    request_id: u16,
-    base: u64,
-}
-
-impl PeerContext {
-    pub fn new(base: u64) -> Self {
-        Self {
-            base,
-            request_id: 0u16,
-            ephemerals: Default::default(),
-        }
-    }
-
-    pub fn allocate_request_id(&mut self) -> types::MockRequestId {
-        let id = self.base + self.request_id as u64;
-        self.request_id = self.request_id.overflowing_add(1).0;
-        types::MockRequestId::new(id)
-    }
-}
-
-#[derive(Debug, Default)]
 pub struct RequestManager {
-    /// Registered peers
-    peers: HashMap<types::MockPeerId, PeerContext>,
+    /// Active ephemeral IDs
+    ephemerals: HashMap<types::MockPeerId, HashSet<types::MockRequestId>>,
 
     /// Pending outbound requests (TODO: timeouts)
     _pending: HashMap<types::MockRequestId, types::MockRequestId>,
 
-    /// Request ID zone for allocating unique IDs for peers
-    req_id_zone: u64,
+    // Next ID for outbound request
+    next_request_id: types::MockRequestId,
 
     /// Next ephemeral request ID
-    next_ephemeral: types::MockRequestId,
+    next_ephemeral_id: types::MockRequestId,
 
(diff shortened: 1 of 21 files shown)


p2p/src/net/mock/types.rs
```diff
@@ -68,7 +94,28 @@ pub enum PubSubEvent {
...
+    }
+}
+
+impl std::ops::Deref for MockRequestId {
+    type Target = u64;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
iljakuklic, line 112: This will probably expose some operations that don't make sense on IDs. How many u64 operations are actually nneded, besides comparisons which are auto-derived anyway? Possibly just exposing required operations one by one would not be that much more code.

p2p/src/net/mock/request_manager.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// Register peer to the request manager
+    ///
+    /// Initialize peer context and allocate request ID slice for the peer
+    pub fn register_peer(&mut self, peer_id: types::MockPeerId) -> crate::Result<()> {
+        match self.peers.entry(peer_id) {
+            Entry::Occupied(_) => Err(P2pError::PeerError(PeerError::PeerAlreadyExists)),
+            Entry::Vacant(entry) => {
+                entry.insert(PeerContext::new(self.req_id_zone));
+                self.req_id_zone =
+                    self.req_id_zone.checked_add(0x10000).expect("number of peers to be lower");
```
iljakuklic, line 89: Since these IDs are u64 (correct me if I'm wrong), how about having 2<spu>32</sup> zones with 2<sup>32</sup> IDs each to support the occasional odd test that needs many requests. Also is the possibility of one peers request IDs overflowing into the next zone checked somewhere so it's detected when it happens?

iljakuklic (approved):

p2p/src/net/mock/request_manager.rs
altonen: I'm not entirely opposed to changing the allocation scheme to what you're proposing but I find it more likely that we'll have a test that needs more than `232` zones rather than a test that has `2^32` pending requests. The request ID's are most likely going to require some work either when request timeouts are implemented. I'd prefer doing that work later as it's not in the same rush as pubsub or fixing connectivity issues.

I'm not sure if it's possible for a peer's IDs to overflow into another peer's zone since the size of the zone is capped at `u16::MAX` and zones starts at a multiple of `0x10000`. There is currently one test checking that the allocation/overflow doesn't break at least for the normal case.

iljakuklic: That's fair

stanislav-tkach (approved):

p2p/src/net/mock/types.rs
altonen: Done

Also simplified the request ID allocation to just use one common counter for all outbound requests and increment it once for each request. No need to deal with peer-specific zones, it was part of previous design.

iljakuklic (approved):

TheQuantumPhysicist (approved):

TheQuantumPhysicist (commented):
Any reason why this isn't getting merged?

altonen:
I think somebody resrun the tests but last time everything was broken so I was hesitant about merging this.

muursh (approved):

altonen:
Once the CI succeeds, I think I'll merge this unless there are objections. I can continue debugging the functional tests in some other branch

TheQuantumPhysicist:
> Once the CI succeeds, I think I'll merge this unless there are objections. I can continue debugging the functional tests in some other branch

Go for it. I hope we won't regret this.

PR #346, opened 2022-08-08
Fix WithId equality issue

TheQuantumPhysicist:
As per our earlier discussion, a generic WithId should not have equality derived, but the developer should decide whether they want to compare the id or the object by getting its reference. Since the equality of WithId of transactions is used in mempool entries, I implemented equality for WithId of transactions.

All done from my smart phone on a hotel bed 😅... I feel the urgency for this change because I don't want anyone to build upon it.

### DIFF
--- a/common/src/chain/transaction.rs
+++ b/common/src/chain/transaction.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::primitives::{Id, Idable};
+use crate::primitives::{id::WithId, Id, Idable};
 use serialization::{DirectDecode, DirectEncode, Encode};
 use thiserror::Error;
 
@@ -52,6 +52,14 @@ impl Idable for Transaction {
         }
     }
 }
+
+impl PartialEq for WithId<Transaction> {
+    fn eq(&self, other: &Self) -> bool {
+        WithId::get(self) == WithId::get(other)
+    }
+}
+
+impl Eq for WithId<Transaction> {}
 
 #[derive(Error, Debug, Clone)]
 pub enum TransactionCreationError {
--- a/common/src/primitives/id/with_id.rs
+++ b/common/src/primitives/id/with_id.rs
@@ -21,7 +21,7 @@ use super::{Id, Idable};
 ///
 /// This only allows immutable access to the underlying object to prevent it from going out of sync
 /// with the ID, which is calculated for its contents. Getting an ID just returns the stored one.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct WithId<T: Idable> {
     id: Id<T::Tag>,
     object: T,


iljakuklic:
I think it's fine to derive equality here. The auto-generated instance looks something like
```rust
impl<T: PartialEq> PartialEq for WithId<T> { ... }
impl<T: Eq> Eq for WithId<T> { ... }
```

I.e. the operations only exist if the underlying object is `Eq`/`ParitailEq`. It's up to the underlying object to decide if and how equality is implemented, `WithId` only inherits the behaviour.

TheQuantumPhysicist:
> I think it's fine to derive equality here. The auto-generated instance looks something like
> 
> ```rust
> impl<T: PartialEq> PartialEq for WithId<T> { ... }
> impl<T: Eq> Eq for WithId<T> { ... }
> ```
> 
> I.e. the operations only exist if the underlying object is `Eq`/`ParitailEq`. It's up to the underlying object to decide if and how equality is implemented, `WithId` only inherits the behaviour.

This is not about things not working, but about inefficiency and it being wrong from an implementation point of view. Think of it as a cache holder. When you compare caching objects, you don't just compare the contents. Typically the cache is ignored. This is the same. Why compare the id AND the object? Either this or that. Comparing both is wasteful, from a performance point of view and wrong from a design point of view.

For transactions, it makes sense to compare the object alone because of the witness thing. For blocks, comparing just the id is good enough. Why compare everything for everything? I don't see that as beneficial on any level.

iljakuklic:
> This is not about things not working, but about inefficiency and it being wrong from an implementation point of view. Think of it as a cache holder. When you compare caching objects, you don't just compare the contents. Typically the cache is ignored. This is the same. Why compare the id AND the object? Either this or that. Comparing both is wasteful, from a performance point of view and wrong from a design point of view.

I don't think performance should be a massive issue here but there is a potential performance benefit to comparing both. The `id` part in this case acts as a quick rejection filter in case the objects are not equal in some part that is included in the ID hash. Comparing the object contents is skipped in that case. For exmaple, when iterating through a long list to find an element, this can be quite a boost. If the objects are equal, this ends up being a bit slower. Overall it all depends on how often the ids don't match.

> For transactions, it makes sense to compare the object alone because of the witness thing. For blocks, comparing just the id is good enough. Why compare everything for everything? I don't see that as beneficial on any level.

What I don't like about this PR is the special casing for `Transaction`. Comparing the underlying object for everything would be correct but would lose the benefit of comparing IDs only in cases where it's sufficient such as blocks.

This whole thing points to the deeper issue that sometimes the ID determines the object entirely and sometimes not. It also comes up when calculating the ID itself, in signatures, etc., where we have to carefully avoid including witnesses but make sure to include everything else. It creates a bunch of mess. We could improve on it with some more refactoring, perhaps, where we separate witnesses and transactions a bit more, i.e. something like (heavily simplified):

```rust
struct BlockBody {
    rewards: Vec<TxOutput>,
    witnesses: Vec<Witness>,
    transactions: Vec<Transaction>,
}
```

The `Transaction` type would not include a witness. Then `Id<Transaction>` would fully determine the transaction, equality comparison could be done by comparing `Id` just like with everything else. Another type would be introduced, say `SignedTransaction`, which pairs up `Transaction` with the corresponding subsequence of witnesses. Then `Id<SignedTransaction>` would signify an id of transaction with particular set of witnesses. There are multiple valid `SignedTransaction`s for each `Transaction` because of malleability. The two Merkle roots would be derived one from a seqiuence of `Id<Transaction>` and the other one from a sequence of `Id<SignedTransaction>`. The distinction between transactions with or without witnesses and IDs thereof becomes explicit.

This means, however, departing even more from Bitcoin, although segwit is quite similar with having witnesses completely separate. Another possibility is to do it at `Transaction` level (same reasoning about IDs as above applies):

```rust
struct Transaction {
    inputs: Vec<TxInput>, // Does not contain a Witness
    outputs: Vec<TxOutput>,
    // other stuff
}
struct SignedTransaction {
   tx: Transaction,
   witnesses: Vec<Witness>, // Exactly one per tx.inputs
}
```

Or something similar. That would address a bunch of issues where we have to special case `Transaction`:
* Calculating ID of objects would reduce to just serializing and hashing
* Comparing by ID is sufficient to fully determine equality of the object
* Perhaps some more I've forgotten, I recall a couple of instances where this is the underlying issue.

Comments / thoughts / counter-proposals?

TheQuantumPhysicist:
> This whole thing points to the deeper issue that sometimes the ID determines the object entirely and sometimes not.

I agree. This is basically my point. It should not be automated at the generic level, and the comparison should be done at case-by-case basis. For transactions, it's just the object.

Now, the witness issue.

The SignedTransaction struct is something I suggested in the engineering channel a few months ago... but I never got to it as other things with higher priority got in the way. It's not simple though because it'll create issues with the enum of versions of Transaction. Creating an interface for that is difficult. Feel free to give it a shot.

On the other hand, I don't think the SignedTransaction thing will help alleviate any of your concerns. We still have to specialize for SignedTransaction anyway, which will become the mempool transaction type because unsigned transactions don't make sense in the mempool. So we're back at square one. How are you gonna implement Eq for `WithId<SignedTransaction>`? 

Separate transactions at the block level won't work, btw, because we need to bundle the witness with transactions somehow for the mempool. So we'll have SignedTransaction anyway.

I'm OK with changing the Transaction comparison to also compare also with the id first. Will that be a satisfying compromise for you?

iljakuklic:
> I'm OK with changing the Transaction comparison to also compare also with the id first. Will that be a satisfying compromise for you?

I don't exactly understand your suggestion here. At high level, my preference is for `WithId` to behave consistently with the underlying object, preferably without special casing for `WithId<Transaction>` and the like.

These options are fine as far as I am concerned:
* No `Eq` impl on `WithId`, be explicit every time
* Use `Eq` that delegates to `Eq` of the underlying object (no direct ID comparison)
* Use auto-derived `Eq`, comparing ID and if that checks out, comparing the whole thing
* [in the future, not quite possible now but best option] Compare only IDs if it is guaranteed to fully determine the object contents

TheQuantumPhysicist:
> I don't exactly understand your suggestion here.

I mean add obj.id comparison to that PartialEq manual trait impl. This way you get your quick shortcut in case they're not equal.

> No Eq impl on WithId, be explicit every time

This is currently not possible because, like I said in the PR message, mempool entries require Eq for that. See the struct. So it's either for WithId or for Transactions specifically like I did. Pick your poison.

Please understand that deriving Eq for WithId is horrible for something like blocks. You're gonna potentially compare like 1 MB of data while a 32-byte fingerprint suffices. And yes, I know this is only in case they're equal, but still it's lots of comparison operations (remember this is tons of function calls for every equality for every subtype, not like comparing serialized data altogether that can be optimized with SIMD or other vector instructions). 

A workaround could be to just create another type that wraps Transaction just for mempool entry. But I find that an overkill.

iljakuklic:
> The SignedTransaction struct is something I suggested in the engineering channel a few months ago... but I never got to it as other things with higher priority got in the way. It's not simple though because it'll create issues with the enum of versions of Transaction. Creating an interface for that is difficult.

True, it's not going to be trivial to come up with a nice representation for this.

> On the other hand, I don't think the SignedTransaction thing will help alleviate any of your concerns. We still have to specialize for SignedTransaction anyway, which will become the mempool transaction type because unsigned transactions don't make sense in the mempool.

Well the maps would look like something like `BTreeMap<Id<Transaction>, SignedTransaction>` where `key == value.transaction().get_id()`, i.e. signed transactions addressed by transaction ID. Notice how everything is completely explicit and visible in the code. This forces you to think about what happens when a transaction with the same ID but different signature arrives, since the distinction between plain transaction and (transaction + witness) is now explicit. In general, I'd say, 

> So we're back at square one. How are you gonna implement Eq for `WithId<SignedTransaction>`?

Ideally everything would behave in a consistent, predictable manner. For most objects `x, y: T` that form the blockchain, the following should hold:

1. `(x.encode() == y.encode()) = (x == y)`
2. `WithId::new(x).get_id() = x.get_id()`
3. `(x == y) = (x.get_id() == y.get_id())`

Today, (1) and (3) don't hold for all types, although they still hold asymetrically as implications.

There would not be any special `Eq` instance for `WithId<SignedTransaction>`. You are comparing signed transactions, so the result is the same as comparing signed transactions directly, i.e. everything including witnesses. If comparison without witnesses is desired, compare `Transaction`s instead. This holds regardless the use of `WithId`. The `WithId` impl would just compare the IDs, as for all other types.

iljakuklic:
I summarised the the discussion about extracting witnesses in a new issue #348, we can continue there. Let's focus here on the more immediate-term issue

iljakuklic:
> This is currently not possible because, like I said in the PR message, mempool entries require Eq for that. See the struct. So it's either for WithId or for Transactions specifically like I did. Pick your poison.

Ah I see. I guess we can roll with this for now, if there's no workaround and a wrapper is deemed too invasive. Maybe just move the impl closer to `Transaction` definition, rather than where `WithId` is defined. It seems to work out better dependency-wise. `WithId` only depends on `Id` and `Idable` and such, without knowing anything about particular `Idable` types. On the other hand, `Transaction` already depends on `primitives::id` so it seems more natural to put the impl there.

> Please understand that deriving Eq for WithId is horrible for something like blocks. You're gonna potentially compare like 1 MB of data while a 32-byte fingerprint suffices. And yes, I know this is only in case they're equal, but still it's lots of comparison operations (remember this is tons of function calls for every equality for every subtype, not like comparing serialized data altogether that can be optimized with SIMD or other vector instructions).

True, but still probably no worse than what we have today. Do we compare whole blocks often? We can try removing the `Eq`
impl on Block and see how much it complains. An exception would be tests where we might compare things just to check against expected output. Maybe comparing the whole thing is fine or even preferable there. Although I think tests still compare `Block` and not `WithId<Block>` so the issue does not quite apply.

TheQuantumPhysicist:
> Ah I see. I guess we can roll with this for now, if there's no workaround and a wrapper is deemed too invasive. Maybe just move the impl closer to `Transaction` definition, rather than where `WithId` is defined.

It's already under the Transaction definition.

> We can try removing the `Eq` impl on Block and see how much it complains.

Agreed. This occurred to me after writing my last reply. It's worth a shot. But again, I'm on my phone in Subway. I may try to do this when I'm back at the hotel.

Are you now OK with the PR that the impl is besides Transaction and the reasoning is justified? Or am I missing more requirements?

iljakuklic (approved):

iljakuklic:
> It's already under the Transaction definition.

Sorry, missed that somehow. Approved

PR #347, opened 2022-08-08
Add timeout for connection establishment in P2P tests

altonen:
Couple all `poll_next()` calls with `tokio::time::timeout` so tests in CI don't hang in case the connection establishment fails


muursh:
Is this only needed for CI?

altonen:
Yes, I just want to see which event causes the CI to hang. It should also prevent CI getting stuck for 6 hours and fail immediately instead.

muursh (approved):

PR #350, opened 2022-08-08
chainstate: Make generic w.r.t. the storage backend

iljakuklic:
Not that many changes were required in the end.

* Tests use type aliases that refer to the in-memory backend.
* The node initialisation code specifies the in-memory backend since it's the only one currently available

### DIFF
--- a/chainstate/src/lib.rs
+++ b/chainstate/src/lib.rs
@@ -54,10 +54,10 @@ impl subsystem::Subsystem for Box<dyn ChainstateInterface> {}
 
 type ChainstateHandle = subsystem::Handle<Box<dyn ChainstateInterface>>;
 
-pub fn make_chainstate(
+pub fn make_chainstate<S: chainstate_storage::BlockchainStorage + 'static>(
     chain_config: Arc<ChainConfig>,
     chainstate_config: ChainstateConfig,
-    chainstate_storage: chainstate_storage::Store,
+    chainstate_storage: S,
     custom_orphan_error_hook: Option<Arc<detail::OrphanErrorHandler>>,
     time_getter: TimeGetter,
 ) -> Result<Box<dyn ChainstateInterface>, ChainstateError> {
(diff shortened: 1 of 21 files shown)


chainstate/src/lib.rs
```diff
@@ -54,10 +54,10 @@ impl subsystem::Subsystem for Box<dyn ChainstateInterface> {}
 
 type ChainstateHandle = subsystem::Handle<Box<dyn ChainstateInterface>>;
 
-pub fn make_chainstate(
+pub fn make_chainstate<S: chainstate_storage::BlockchainStorage + 'static>(
```
TheQuantumPhysicist, line 57: Is there a way to make this proper lifetime instead of static?

TheQuantumPhysicist (approved):
Besides the static thing, looks good. I'm glad we're still maintaining the self-consumption.

chainstate/src/lib.rs
stanislav-tkach: I think that the `static` lifetime isn't an issue here, because we don't use references anyway: chainstate owns a storage.

stanislav-tkach (approved):

chainstate/src/lib.rs
TheQuantumPhysicist: Why can't we relate the lifetime to chainstate lifetime instead of static?

iljakuklic: The `'static` thing here declares that `S` does not have any references outside of stuff it owns, i.e. it is self-contained. In some sense, having `'static` here is maximally limiting to the storage implementation (it can't have references to non-static things) but maximally liberating to clients, that is chainstate and anything that uses it (its allowed to take references to it and move it with the least amount of restriction). I think this is a good starting point but we can adjust it if the need arises.

> Why can't we relate the lifetime to chainstate lifetime instead of static?

How do you think they should relate? It's should only be needed if there are going to be references between the two. Since chainstate owns storage, I don't see what would it help with.

stanislav-tkach (approved):

muursh (approved):
lgtm

sinitcin (approved):

PR #351, opened 2022-08-08
Move the consensus related logic from chainstate to a separate crate

stanislav-tkach:
There should be no changes in the logic, only moving some modules and just a bit of refactoring.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,29 +11,29 @@ edition = "2021"
 
 [workspace]
 members = [
-  "common",        # everything else, until it's moved to another crate
-  "crypto",        # cryptographic primitives and their interfaces
-  "consensus",     # Consensus related logic.
-  "chainstate",    # code on chainstate of blocks and transactions
-  "script",        # bitcoin script and its interfaces
-  "logging",       # logging engine and its interfaces
-  "mempool",       # mempool interface and implementation
-  "p2p",           # p2p communication interfaces and protocols
-  "rpc",           # rpc abstraction and implementation
-  "serialization", # full featured serialization interfaces and implementations
-  "serialization/core",    # serialization core tools
-  "serialization/tagged",  # serialization for direct/tagged encoding style
-  "serialization/tagged/derive",  # direct/tagged encoding style derive macros
-  "storage",       # storage abstraction layer and its implementation
-  "storage/core",  # Core backend-agnostic storage abstraction
-  "storage/inmemory",      # In-memory storage backend implementation
-  "subsystem",     # Utilities for working with concurrent subsystems
-  "node",          # node executable
-  "wallet",        # wallet executable
-  "utils",         # various utilities
-  "utxo",          # utxo and related utilities (cache, undo, etc.)
-  "test",          # integration tests
-  "test-utils",    # various utilities for tests
+  "common",                       # Everything else, until it's moved to another crate.
+  "crypto",                       # Cryptographic primitives and their interfaces.
+  "consensus",                    # Consensus related logic.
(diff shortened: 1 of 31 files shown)


chainstate-types/src/error.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use thiserror::Error;
+
+use common::{
+    chain::{Block, GenBlock},
+    primitives::{BlockHeight, Id},
+};
+
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
+pub enum PropertyQueryError {
```
stanislav-tkach, line 24: I need to reuse this error in the new consensus crate, so it was moved from chainstate itself to `chainstate_types`.

chainstate-types/src/lib.rs
```diff
@@ -13,7 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub mod block_index;
-pub mod height_skip;
-pub mod locator;
 pub mod storage_result;
+
+pub use crate::{
+    block_index::BlockIndex, error::PropertyQueryError, gen_block_index::GenBlockIndex,
```
stanislav-tkach, line 19: In my opinion it is really inconvenient to always import these types like this:
```
use chainstate_types::{block_index::BlockIndex, gen_block_index::GenBlockIndex};
```
These modules contain only one type each and type names are already contains all needed context, so I see no harm in that.

common/src/chain/block/block_reward.rs
```diff
@@ -5,7 +5,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-// http://spdx.org/licenses/MIT
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
```
stanislav-tkach, line 8: https://github.com/mintlayer/mintlayer-core/pull/332#discussion_r940043447

iljakuklic (approved):
Looks good

azarovh (approved):

Cargo.toml
```diff
@@ -13,6 +13,7 @@ edition = "2021"
 members = [
   "common",        # everything else, until it's moved to another crate
   "crypto",        # cryptographic primitives and their interfaces
+  "consensus",     # Consensus related logic.
```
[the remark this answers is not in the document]
stanislav-tkach: I like your attention to the details! Though `storage/core` and `storage/inmemory` descriptions also start with a capital letter. 🙃 Personally I prefer to use capital letters everywhere.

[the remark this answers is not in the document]
iljakuklic: Since this is a pretty minor change, I suggest @stanislav-tkach to change all the comments here in accordance with his aesthetic sensibilities as a part of this PR (or a separate one if preferred).

iljakuklic (approved):

muursh (approved):

stanislav-tkach:
I'm still not sure that moving out `PropertyQueryError` to the `chainstate_types` crate is a bad idea. This error is contained in the public `ChainstateError` type. In the new `consensus` crate it is also used as part of the `ConsensusVerificationError:PrevBlockLoadError` error.

Additionally, the implementations of the `TransactionIndexHandle` and `BlockIndexHandle` traits heavily rely on chainstate methods that return `PropertyQueryError`.

TheQuantumPhysicist:
Looks good. Please fix the PoS branch after merging this to master because it has pos dir beside pow, which is now moved.

TheQuantumPhysicist (approved):

PR #353, opened 2022-08-10
Utxo crate cleanup

azarovh:
Note: this PR is based on #333 which is not merged yet. 

Decided to clean up a crate a bit before introducing any new changes. I think these changes will increase readability and make the code cleaner.

- Changed the layout of the crate
- Reworked `pub` declarations and `use` directives
- removed redundant dead_code and dead_imports with useless todos.


utxo/src/utxo_entry.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    pub fn is_fresh(&self) -> bool {
+        self.is_fresh
+    }
+
+    pub fn is_spent(&self) -> bool {
+        self.status == UtxoStatus::Spent
+    }
+
+    pub fn utxo(&self) -> Option<&Utxo> {
```
azarovh, line 63: pay attention that this getter is now returning a ref instead of implicitly cloning every time

utxo/src/cache.rs
```diff
@@ -351,16 +176,12 @@ impl<'a> UtxosCache<'a> {
...
             // This is only available in this view. Remove immediately.
             self.utxos.remove(outpoint);
         } else {
             // mark this as 'spent'
-            let new_entry = UtxoEntry {
-                status: UtxoStatus::Spent,
-                is_dirty: true,
-                is_fresh: false,
-            };
+            let new_entry = UtxoEntry::new(None, false, true);
```
azarovh, line 184: I don't like these booleans - they are easy to misuse. Will do something about them

utxo/src/cache.rs
```diff
@@ -13,202 +13,35 @@
...
+    pub(crate) best_block: Id<GenBlock>,
 }
 
 #[derive(Clone)]
 pub struct UtxosCache<'a> {
     parent: Option<&'a dyn UtxosView>,
     current_block_hash: Id<GenBlock>,
-    utxos: BTreeMap<OutPoint, UtxoEntry>,
-    //TODO: do we need this?
+    pub(crate) utxos: BTreeMap<OutPoint, UtxoEntry>,
```
stanislav-tkach, line 40: I think that it is worth putting a comment (maybe even todo?..) here that this `pub(crate)` is only needed for tests.

utxo/src/cache.rs
```diff
@@ -13,202 +13,35 @@
...
+    #[allow(dead_code)]
     memory_usage: usize,
```
stanislav-tkach, line 42: Do we really need this?.. As far as I can see, there are several TODO-comments already. Perhaps it makes sense to create an issue and leave a reference to it here.

utxo/src/cache.rs
```diff
@@ -13,202 +13,35 @@
...
-        match &mut self.status {
-            UtxoStatus::Spent => None,
-            UtxoStatus::Entry(utxo) => Some(utxo),
-        }
-    }
-}
-
 impl<'a> UtxosCache<'a> {
     #[cfg(test)]
     pub(crate) fn new_for_test(best_block: Id<GenBlock>) -> Self {
```
stanislav-tkach, line 47: Nitpick: currently the `#[cfg(test)]` attribute makes no sense together with `pub(crate)`.
(https://github.com/rust-lang/cargo/issues/8379)

utxo/src/storage/in_memory.rs
```diff
@@ -31,14 +30,16 @@ pub struct UtxosDBInMemoryImpl {
 }
 
 impl UtxosDBInMemoryImpl {
-    pub fn new(best_block: Id<GenBlock>, initial_utxos: BTreeMap<OutPoint, Utxo>) -> Self {
+    #[allow(dead_code)]
```
stanislav-tkach, line 33: When I see this I am always tempted to try to remove the attribute in order to check if it is really needed and why. 🙃 I think a comment explaining it would be nice.

stanislav-tkach (approved):

utxo/src/utxo_entry.rs
TheQuantumPhysicist: It sounds like it should've been like that from the beginning. It's the right thing to do.

utxo/src/cache.rs
TheQuantumPhysicist: Flags are not good either. They're OK for storage, but they're not OK for function calls and interface operations. We need to find a solution that's type-safe. In fact, two enums or some combination structs with good overrides may solve this problem. Please think of something.

utxo/src/cache.rs
TheQuantumPhysicist: Please do not eliminate everything related to memory, yet. It'll make the process of introducing it much harder. Please restore everythign related to memory, and keep them at value zero with proper TODOs, so that the person who has to add them later doesn't have to rethink everything.

utxo/src/storage/in_memory.rs
```diff
@@ -31,15 +30,15 @@ pub struct UtxosDBInMemoryImpl {
 }
 
 impl UtxosDBInMemoryImpl {
-    pub fn new(best_block: Id<GenBlock>, initial_utxos: BTreeMap<OutPoint, Utxo>) -> Self {
+    pub fn new(best_block: Id<GenBlock>, _initial_utxos: BTreeMap<OutPoint, Utxo>) -> Self {
         Self {
             store: BTreeMap::new(),
```
TheQuantumPhysicist, line 35: This seems to be a mistake. The initial Utxos should go into the map.

utxo/src/storage/in_memory.rs
```diff
@@ -31,15 +30,15 @@ pub struct UtxosDBInMemoryImpl {
...
             undo_store: BTreeMap::new(),
             best_block_id: best_block,
         }
     }
 
-    pub(crate) fn internal_store(&mut self) -> &BTreeMap<OutPoint, Utxo> {
+    pub fn internal_store(&mut self) -> &BTreeMap<OutPoint, Utxo> {
```
TheQuantumPhysicist, line 41: Why did you replace pub(crate) with pub? This shouldn't be part of the public interface.

TheQuantumPhysicist:
This looks mostly fine. Just a few things have to be thought through, especially the flags vs booleans thing. Booleans are a MUCH better solution than flags, but there's certainly solutions better than booleans, which we postponed for priority reasons. When I first read that you changed them to flags before looking into the PR, I was happy, because I thought we created some rusty solution that's type-safe... but it seems like we're back at C/C++ interface stuff with dangerously ambiguous non-type-safe solutions. We can certainly do better! If we don't want to do this now, then we better go back to booleans. At least booleans won't accept the value 0x56 at compile-time.

iljakuklic:
> This looks mostly fine. Just a few things have to be thought through, especially the flags vs booleans thing. Booleans are a MUCH better solution than flags, but there's certainly solutions better than booleans, which we postponed for priority reasons. When I first read that you changed them to flags before looking into the PR, I was happy, because I thought we created some rusty solution that's type-safe... but it seems like we're back at C/C++ interface stuff with dangerously ambiguous non-type-safe solutions. We can certainly do better! If we don't want to do this now, then we better go back to booleans. At least booleans won't accept the value 0x56 at compile-time.

I think a modest improvement over flags is a custom enum for each of these, like:
```rust
enum IsDirty { Yes, No }
```

Similarly for other bool-y things. These are all still basically bools but at least it's nicely explicit and it does not allow you to pass one in place of another without an explicit conversion of some sort.

utxo/src/storage/test.rs
```diff
@@ -100,7 +101,7 @@ fn create_utxo_entries(rng: &mut impl Rng, num_of_utxos: u8) -> BTreeMap<OutPoin
     let mut map = BTreeMap::new();
     for _ in 0..num_of_utxos {
         let (utxo, outpoint) = create_utxo(rng, 0);
-        let entry = UtxoEntry::new(utxo.clone(), true, true);
+        let entry = UtxoEntry::new(Some(utxo), FRESH | DIRTY);
```
iljakuklic, line 104: See my reply to @TheQuantumPhysicist. Maybe a bunch of custom enums is better.

utxo/src/utxo.rs
```diff
@@ -0,0 +1,98 @@
...
+use common::{chain::TxOutput, primitives::BlockHeight};
+use serialization::{Decode, Encode};
+use std::fmt::Debug;
+
+/// Determines whether the utxo is for the blockchain of for mempool
+#[derive(Debug, Clone, Eq, PartialEq, Encode, Decode)]
+pub enum UtxoSource {
+    /// At which height this containing tx was included in the active block chain
+    BlockChain(BlockHeight),
+    MemPool,
```
iljakuklic, line 26: Can we rename this `MemPool` -> `Mempool`?

iljakuklic (commented):
A couple of minor comments. Please address @TheQuantumPhysicist's comment on `UtxosDBInMemoryImpl` initialisation

TheQuantumPhysicist:
> > This looks mostly fine. Just a few things have to be thought through, especially the flags vs booleans thing. Booleans are a MUCH better solution than flags, but there's certainly solutions better than booleans, which we postponed for priority reasons. When I first read that you changed them to flags before looking into the PR, I was happy, because I thought we created some rusty solution that's type-safe... but it seems like we're back at C/C++ interface stuff with dangerously ambiguous non-type-safe solutions. We can certainly do better! If we don't want to do this now, then we better go back to booleans. At least booleans won't accept the value 0x56 at compile-time.
> 
> I think a modest improvement over bools is a custom enum for each of these, like:
> 
> ```rust
> enum IsDirty { Yes, No }
> ```
> 
> Similarly for other bool-y things. These are all still basically bools but at least it's nicely explicit and it does not allow you to pass one in place of another without an explicit conversion of some sort.

Like I mentioned in the daily, a type-safe flags, where we wrap u8 with a struct that has associated constant values is also OK from my end. So either that enum from Lukas or this type-safe flags thing. Whichever is fine for me.

utxo/src/storage/in_memory.rs
azarovh: in_memory mod is not exposed to the public and is kept only for tests. So this seems like the right thing to do just to keep it clean.

TheQuantumPhysicist: I get that, but in-memory should be exposable to the public if we feel the need for it. Having pub(crate) signals that this is a special method that shouldn't be used normally. Even better, if you can change this to pub(super) or something more restrictive, I'm OK with that..

utxo/src/cache.rs
azarovh: Enums were also in my mind, but I thought they are too verbose. Though now I think it might be an optimal solution.

Will go with enums

utxo/src/utxo.rs
azarovh: Honestly, I don't like it either. And if no one objects I'll rename `BlockChain` to `Blockchain` as well

TheQuantumPhysicist: You have my blessing too.

utxo/src/cache.rs
```diff
@@ -351,16 +200,12 @@ impl<'a> UtxosCache<'a> {
...
             // This is only available in this view. Remove immediately.
             self.utxos.remove(outpoint);
         } else {
             // mark this as 'spent'
-            let new_entry = UtxoEntry {
-                status: UtxoStatus::Spent,
-                is_dirty: true,
-                is_fresh: false,
-            };
+            let new_entry = UtxoEntry::new(None, IsFresh::No, IsDirty::Yes);
```
iljakuklic, line 208: One thing just occurred to me. Are all the combinations possible here? E.g. is it possible to have UTXO freash and dirty simultaneously? If not, maybe one enum (like `UtxoState` or something) would be better so only valid states can be constructed.

azarovh: Good observation! Actually out of these 2 fields every combination is possible, but if we take 3rd field `status` into account then only 5 out of 8 is valid. I'll add assertions into constructor

iljakuklic: Would it be possible to reorganize the enums so the invalid states cannot be constructed? That would move the run-time checking (assertions) to compilation time, which is preferable unless it makes the code too cryptic.

utxo/src/storage/in_memory.rs
TheQuantumPhysicist: You know, even better, maybe just add `#[cfg(test)]` above it. Then you can keep it as pub if you wish.

TheQuantumPhysicist (approved):

utxo/src/cache.rs
azarovh: yeah the problem is that code starts looking weird. As a compromise `debug_assert` can be used. `UtxoEntry` is private so no one outside the crate will construct it and all combinations can be checked by tests.

iljakuklic: Ok. Let's keep `assert` then

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #355, opened 2022-08-15
Avoid reading the whole block to get the block reward

TheQuantumPhysicist:


chainstate-types/src/error.rs
```diff
@@ -30,6 +30,8 @@ pub enum PropertyQueryError {
     BestBlockIndexNotFound,
     #[error("Block not found {0}")]
     BlockNotFound(Id<Block>),
+    #[error("Block reward not found for block with id {0}")]
+    BlockRewardNotFound(Id<Block>),
```
stanislav-tkach, line 34: Am I missing something or this error isn't used anywhere?

stanislav-tkach (approved):

PR #356, opened 2022-08-16
p2p: Start using `tokio::unbounded_channel`

altonen:


TheQuantumPhysicist (approved):
All good. I wonder whether there's a way to monitor some indicator for the number of channels being used so that we notice when they're growing uncontrollably due to a bug. Unlikely, but worth thinking about. Perhaps open an issue for it. What do you think?

altonen:
It is on my todo list to create some tracing mechanism for channels so that we could observe what kinds of events are in the system and how they move from one service to another. I am working on this abstraction for channels and having tracing part of that (either Prometheus or something ad-hoc for now) wouldn't be too much work, let's see

PR #357, opened 2022-08-16
Remove transaction size from transaction index

TheQuantumPhysicist:
We don't really need it... it was added as part of being paranoid, but maybe that isn't necessary. Let's see what the fine engineers here think.


muursh (approved):
To be honest I'm fine with removing this.

iljakuklic (approved):

PR #359, opened 2022-08-16
p2p: Add function call abstraction for events

altonen:
Add abstraction over channels

```rust
let (tx, rx) = oneshot::channel();
self.tx_swarm
    .send(event::SwarmEvent::AdjustPeerScore(peer_id, score, tx))
    .await
    .map_err(P2pError::from)?;
let _ = rx.await.map_err(P2pError::from)?;
```
becomes
```rust
match swarm_sender.adjust_peer_score(1, 100).await? {
    Ok(res) => {},
    Err(err) => println!("failed to adjust score: {:?}", err),
}
```

TODO:
- add tracing(?)

### DIFF
--- a/p2p/src/util/src/lib.rs
+++ b/p2p/src/util/src/lib.rs
@@ -1,0 +1,181 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use convert_case::{Case, Casing};
+use proc_macro::{self, TokenStream};
+use proc_macro2::Span;
+use quote::{quote, ToTokens};
+use syn::{
+    parse::{Parse, ParseStream},
+    parse2, parse_macro_input, Attribute, DeriveInput, Fields, Ident, Result, Token, Type,
+};
+
+struct Retval {
+    _oneshot_tok: Ident,
+    _sep1: Token![::],
+    _sender_tok: Ident,
+    _sep2: Token![<],
+    inner: Type,
+    _sep3: Token![>],
+}
+
+impl Parse for Retval {
+    fn parse(input: ParseStream) -> Result<Self> {
+        Ok(Retval {
+            _oneshot_tok: input.parse()?,
+            _sep1: input.parse()?,
+            _sender_tok: input.parse()?,
+            _sep2: input.parse()?,
+            inner: input.parse()?,
+            _sep3: input.parse()?,
+        })
+    }
+}
+
+// try to parse real name for the function if it was supplied using `#[real_name = "..."]`
+fn parse_real_name(attributes: &[Attribute]) -> Option<proc_macro2::TokenStream> {
+    if let Some(attr) = attributes.iter().find(|attr| attr.path.is_ident("name")) {
+        let ident = if let Ok(syn::Meta::NameValue(syn::MetaNameValue {
+            lit: syn::Lit::Str(litstr),
+            ..
+        })) = attr.parse_meta()
+        {
+            syn::Ident::new(litstr.value().as_str(), Span::call_site())
+        } else {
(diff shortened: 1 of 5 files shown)


p2p/src/util/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+                    let ident = &field.ident;
+                    let ty = &field.ty;
+
+                    (!field
+                        .attrs
+                        .iter()
+                        .any(|attr| attr.path.is_ident("return_value")))
+                    .then_some((
+                        quote! {
+                            , #ident: #ty
```
iljakuklic, line 144: This is a minor style issue but I'd prefer the fragments here to just expand to `#ident: #ty` (without the comma) and generate the commas when it's spliced into the code later `#(#func_args)*` -> `#(, #func_args)*`.

p2p/src/util/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+        })
+        .collect::<Vec<_>>();
+
+    quote! {
+        #[derive(Debug, Clone)]
+        pub struct #sender {
+            tx: mpsc::UnboundedSender<#orig>,
+        }
+
+        impl #sender{
```
iljakuklic, line 182: Minor: space before `{`

p2p/src/util/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+                }
+            } else {
+                quote! {
+                    pub async fn #real_name(&mut self #(#func_args)*) -> core::result::Result<(), P2pError> {
+                        self.tx
+                            .send(#orig::#name { #(#chan_args,)* })
+                            .map_err(|_| P2pError::ChannelClosed)
+                    }
+                }
+            }
```
iljakuklic, line 172: I am somewhat uncomfortable that the presence or absence of return value determines whether the call blocks the current task. If someone decides to add or remove the return value, the behaviour here changes silently.

Not sure what a good approach may be here. Maybe prefix the method name with `send_` for fire-and-forget or with `call_` when waiting for the result is requested. Not sure if the `send_` version should only be generated for things with no return type or for everything. Ergonomics may suffer a bit but at least it's readily obvious which one is used by looking at the call. Any thoughts?

iljakuklic (commented):
Adding some feedbackz.

p2p/src/util/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+                        rx.await.map_err(|_| P2pError::ChannelClosed)
+                    }
+                }
+            } else {
+                quote! {
+                    pub async fn #real_name(&mut self #(#func_args)*) -> core::result::Result<(), P2pError> {
+                        self.tx
+                            .send(#orig::#name { #(#chan_args,)* })
+                            .map_err(|_| P2pError::ChannelClosed)
+                    }
```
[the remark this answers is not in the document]
altonen: It doesn't

p2p/src/util/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    let orig = data.ident.clone();
+    let sender = syn::Ident::new(format!("{}Sender", data.ident).as_str(), Span::call_site());
+
+    let filtered = if let syn::Data::Enum(syn::DataEnum { variants, .. }) = data.data {
+        variants
+            .into_pairs()
+            .filter_map(|x| match x {
+                Punctuated(entry, _) => Some(entry),
+                _ => None,
+            })
```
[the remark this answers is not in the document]
altonen: Yes, thank you

p2p/src/util/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+            .collect::<Vec<_>>()
```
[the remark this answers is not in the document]
altonen: Fixed

p2p/src/util/src/lib.rs
altonen: Now that I removed the unnecessary `async`, the split is actually quite obvious. Fire-and-forget function calls are synchronous whereas functions with return value require you to `await` on them. Do you think this is clear enough distinction? I personally don't like the idea of prefixing them with something but if others also think it's better to have a prefix I don't mind adding one too much.

iljakuklic: > Now that I removed the unnecessary `async`, the split is actually quite obvious. Fire-and-forget function calls are synchronous whereas functions with return value require you to `await` on them. Do you think this is clear enough distinction? I personally don't like the idea of prefixing them with something but if others also think it's better to have a prefix I don't mind adding one too much.

Yea presence or absence of `.await` is a clear enough distinction I think.

TheQuantumPhysicist:
This seems to be dead. Any benefit from pursuing this?

stanislav-tkach:
I'm thinking that the proc-macro is an overkill here.

PR #360, opened 2022-08-16
Use the 'tap' library as a replacement for the unstable inspect_err function

stanislav-tkach:
While implementing some functionality I really missed the `inspect_err` api that is still unstable and as far as I know there are no recent plans to stabilize it. Additionally we have a TODO-comment in our code-base that suggests using `inspect_err`. There is the `tap` crate that provides the missing functionality. Sometimes it leads to a more concise and clear code. On the other hand, it isn't that important and adding one more dependency for such a small thing is controversial. Therefore I would like to hear you opinions.

### DIFF
--- a/p2p/src/net/libp2p/backend.rs
+++ b/p2p/src/net/libp2p/backend.rs
@@ -16,17 +16,6 @@
 
 //! Libp2p backend service
 
-use crate::{
-    error::{P2pError, PeerError},
-    net::{
-        self,
-        libp2p::{
-            behaviour,
-            types::{self, ControlEvent, Libp2pBehaviourEvent},
-        },
-    },
-};
-use behaviour::sync_codec::message_types::{SyncRequest, SyncResponse};
 use futures::StreamExt;
 use libp2p::{
     gossipsub::{IdentTopic, MessageAcceptance, MessageId},
@@ -34,8 +23,24 @@ use libp2p::{
     swarm::{Swarm, SwarmEvent},
     Multiaddr, PeerId,
 };
-use logging::log;
+use tap::TapFallible;
 use tokio::sync::{mpsc, oneshot};
+
+use logging::log;
+
+use crate::{
+    error::{P2pError, PeerError},
+    net::{
+        self,
+        libp2p::{
+            behaviour::{
+                self,
+                sync_codec::message_types::{SyncRequest, SyncResponse},
+            },
+            types::{self, ControlEvent, Libp2pBehaviourEvent},
+        },
+    },
+};
 
 pub struct Libp2pBackend {
     /// Created libp2p swarm object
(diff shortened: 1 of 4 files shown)


iljakuklic:
Maybe intead of pulling a new dep, we could just shove some hacked together thing into `utils`

```rust
trait LogErr {
   fn log_err(&self, prefix: &str);
}

impl<T, E: Display> LogErr for Result<T, E> {
    fn log_err(&self, prefix: &str) {
        if let Err(err) = self {
            log::error!("{prefix}: {err}");
        }
    }
}
```
And then just use it like
```rust
        .log_err("PeerManager failed")
```


Thoughts?

stanislav-tkach:
I like the idea, though in some (probably rare) cases we want to log some parameters in addition to an error.

By the way, I can leave without this dependency. In some case the common `if let Err(e)` pattern is even better because it uses the result (that has the `must_use` attribute). Still some call chains can benefit from this or a similar api.

TheQuantumPhysicist (approved):
I like the idea in general. In fact, I've been looking for a solution to this problem and I've been meaning to do it myself. We need this for chainstate because currently, when using the try operator "`?`", there's basically no way to hierarchically report errors at certain spots, and I found the tracing crate to be an overkill.

Now about Lukas's concern, adding this dependency isn't really a big issue for me. However, we can write our own "copy" of this. If you see the implementations in the source code, they're quite simplistic. I like the idea in general. In fact, I've been looking for a solution to this problem and I've been meaning to do it myself. We need this for chainstate because currently, when using the try operator "`?`", there's basically no way to hierarchically report errors at certain spots, and I found the tracing crate to be an overkill.

Now about Lukas's concern, adding this dependency isn't really a big issue for me. However, we can write our own "copy" of this. If you see the implementations in the source code, they're quite simplistic. What Lukas did is basically what the crate does, except that it takes functions instead of &str. But I really don't care that much. This crate is too simple for me to bother about it being another dependency and/or waste time rewriting it.

muursh (approved):
I'm good with this too, and have no issues with the new dependency

p2p/src/net/libp2p/backend.rs
```diff
@@ -101,12 +101,8 @@ impl Libp2pBackend {
...
                         ControlEvent::CloseConnection { peer_id })
                     ) => {
-                        // TODO: `inspect_err`
-                        match self.swarm.disconnect_peer_id(peer_id) {
-                            Ok(_) => {}
-                            Err(err) => {
-                                log::error!("Failed to disconnect peer {}: {:?}", peer_id, err);
-                            }
+                        if let Err(e) =  self.swarm.disconnect_peer_id(peer_id) {
+                            log::error!("Failed to disconnect peer {peer_id}: {e:?}");
```
altonen, line 105: why are you not using `tap_err()` here?

altonen (approved):

iljakuklic (commented):
External dependency fine too.

iljakuklic (approved):

p2p/src/net/libp2p/backend.rs
stanislav-tkach: It is somewhat ironic, because the compiler complains about the unused result. It still can be used this way:
```rust
let _ = self.swarm.disconnect_peer_id(peer_id).tap_err(...);
```
But in that case I prefer to use the `if let Err` construction.

TheQuantumPhysicist: Actually I prefer the `let _`, because that signals that the developer deliberately wants to ignore the result, instead of it just being a side-effect.

PR #361, opened 2022-08-17
Introduce SignedAmount

TheQuantumPhysicist:
Since we're gonna deal with some accounting (like in PoS pool balances), it helps to have a type that represents a signed amount, or an amount delta. This type isn't meant to be stored unless we find a good reason for that (hence no Encode/Decode derives). No bit operations are allowed with it as well. This is meant to support certain operations with `Amount` and be a temporary to simplify accounting.

### DIFF
--- a/common/src/primitives/amount.rs
+++ b/common/src/primitives/amount.rs
@@ -567,6 +567,14 @@ mod tests {
         assert_eq!(Amount::from_fixedpoint_str("21987654321.000000", 8).unwrap(), Amount { val: 2198765432100000000 });
         assert_eq!(Amount::from_fixedpoint_str("21987654321.0000000", 8).unwrap(), Amount { val: 2198765432100000000 });
         assert_eq!(Amount::from_fixedpoint_str("21987654321.00000000", 8).unwrap(), Amount { val: 2198765432100000000 });
+        assert_eq!(Amount::from_fixedpoint_str(".2", 8).unwrap(), Amount { val: 20000000 });
+        assert_eq!(Amount::from_fixedpoint_str(".23", 8).unwrap(), Amount { val: 23000000 });
+        assert_eq!(Amount::from_fixedpoint_str(".234", 8).unwrap(), Amount { val: 23400000 });
+        assert_eq!(Amount::from_fixedpoint_str(".2345", 8).unwrap(), Amount { val: 23450000 });
+        assert_eq!(Amount::from_fixedpoint_str(".23456", 8).unwrap(), Amount { val: 23456000 });
+        assert_eq!(Amount::from_fixedpoint_str(".234567", 8).unwrap(), Amount { val: 23456700 });
+        assert_eq!(Amount::from_fixedpoint_str(".2345678", 8).unwrap(), Amount { val: 23456780 });
+        assert_eq!(Amount::from_fixedpoint_str(".23456789", 8).unwrap(), Amount { val: 23456789 });
         assert!(Amount::from_fixedpoint_str("", 8).is_none());
         assert!(Amount::from_fixedpoint_str(" ", 8).is_none());
         assert!(Amount::from_fixedpoint_str("21987654321.000000000", 8).is_none());
@@ -677,6 +685,7 @@ mod tests {
         assert_eq!(Amount::from_fixedpoint_str("21", 1).unwrap(), Amount { val: 210 });
         assert_eq!(Amount::from_fixedpoint_str("1", 1).unwrap(), Amount { val: 10 });
         assert_eq!(Amount::from_fixedpoint_str("1.2", 1).unwrap(), Amount { val: 12 });
+        assert_eq!(Amount::from_fixedpoint_str(".2", 1).unwrap(), Amount { val: 2 });
         assert!(Amount::from_fixedpoint_str("1.23", 1).is_none());
         assert!(Amount::from_fixedpoint_str("1.234", 1).is_none());
         assert!(Amount::from_fixedpoint_str("1.2345", 1).is_none());
--- a/common/src/primitives/amount.rs
+++ b/common/src/primitives/amount.rs
@@ -781,7 +781,6 @@ mod tests {
         assert_eq!(Amount { val: 1234567890100 }.into_fixedpoint_str(0), "1234567890100");
         assert_eq!(Amount { val: 12345678901200 }.into_fixedpoint_str(0), "12345678901200");
         assert_eq!(Amount { val: 123456789012300 }.into_fixedpoint_str(0), "123456789012300");
-
     }
 
     #[rustfmt::skip]
--- a/common/src/primitives/amount.rs
+++ b/common/src/primitives/amount.rs
@@ -18,6 +18,8 @@
 use serialization::{Decode, Encode};
 use std::iter::Sum;
 
+use super::signed_amount::SignedAmount;
+
 // Copyright (c) 2021 RBB S.r.l
 // opensource@mintlayer.org
 // SPDX-License-Identifier: MIT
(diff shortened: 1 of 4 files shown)


common/src/primitives/amount.rs
```diff
@@ -56,16 +58,28 @@ fn remove_right_most_zeros_and_decimal_point(s: String) -> String {
...
+    pub const fn from_atoms(v: UnsignedIntType) -> Self {
         Amount { val: v }
     }
 
-    pub fn into_atoms(&self) -> IntType {
+    pub fn into_atoms(&self) -> UnsignedIntType {
         self.val
     }
 
+    pub fn from_signed(amount: SignedAmount) -> Option<Self> {
```
stanislav-tkach, line 71: Should we add simple tests for this and `into_signed` methods?

common/src/primitives/signed_amount.rs
```diff
@@ -0,0 +1,1030 @@
...
+use std::iter::Sum;
+
+use super::Amount;
+
+// use only unsigned types
+// if you need a signed amount, we should create a separate type for it and implement proper conversion
+pub type SignedIntType = i128;
+
+/// An unsigned fixed-point type for amounts
+/// The smallest unit of count is called an atom
```
azarovh, line 27: comments should be updated

stanislav-tkach (approved):

common/src/primitives/amount.rs
TheQuantumPhysicist: Will do.

common/src/primitives/signed_amount.rs
```diff
@@ -0,0 +1,1030 @@
...
+            let unit = ten.pow(decimals as u32);
+            let whole = self.val.abs() / unit;
+            let fraction = self.val.abs() % unit;
+            let result = format!("{sign}{whole}.{fraction:00$}", decimals as usize);
+
+            remove_right_most_zeros_and_decimal_point(result)
+        }
+    }
+
+    pub fn from_fixedpoint_str(amount_str: &str, decimals: u8) -> Option<Self> {
```
azarovh, line 86: Are conversions from strings like "0.8" or even better ".8" legit? I believe these cases are not uncovered by tests

common/src/primitives/signed_amount.rs
```diff
@@ -0,0 +1,1030 @@
...
+            None
+        } else if amount_str.contains('-') && !amount_str.starts_with('-') {
+            None
+        } else if amount_str.matches('.').count() == 0 {
+            // if there is no decimal point, then just add N zeros to the right and we're done
+            let zeros = "0".repeat(decimals);
+            let amount_str = amount_str + &zeros;
+
+            amount_str.parse::<SignedIntType>().ok().map(|v| SignedAmount { val: v })
+        } else {
```
altonen, line 122: `.` and `-` are parsed into `None` but `-.` is parsed into `Some(SignedAmount { val: 0 })`. Is this correct?

azarovh (approved):

common/src/primitives/signed_amount.rs
altonen: That is, when the number of decimals is more than zero

TheQuantumPhysicist: I updated the implementation and added the tests. I think that's not the case anymore.

common/src/primitives/signed_amount.rs
TheQuantumPhysicist: Done

common/src/primitives/signed_amount.rs
```diff
@@ -0,0 +1,1030 @@
...
+// limitations under the License.
+
+#![allow(clippy::eq_op)]
+
+use std::iter::Sum;
+
+use super::Amount;
+
+// use only unsigned types
+// if you need a signed amount, we should create a separate type for it and implement proper conversion
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Fixed

common/src/primitives/signed_amount.rs
TheQuantumPhysicist: Done

common/src/primitives/signed_amount.rs
```diff
@@ -0,0 +1,1039 @@
...
+        Some(Self::from_atoms(atoms))
+    }
+
+    pub fn into_unsigned(self) -> Option<Amount> {
+        let atoms = self.val;
+        let unsigned_atoms: super::amount::UnsignedIntType = atoms.try_into().ok()?;
+        Some(Amount::from_atoms(unsigned_atoms))
+    }
+
+    pub fn into_fixedpoint_str(self, decimals: u8) -> String {
```
azarovh, line 64: I like the idea of reusing `Amount::from_fixedpoint_str`. Can we do the same for `into_fixedpoint_str`? Not not only we reuse the logic but also can get rid of duplicated `remove_right_most_zeros_and_decimal_point` in this file

azarovh (approved):

altonen (approved):

common/src/primitives/signed_amount.rs
TheQuantumPhysicist: Done

muursh (approved):
lgtm

common/src/primitives/signed_amount.rs
```diff
@@ -0,0 +1,1039 @@
...
+        }
+
+        let signed_amount = unsigned_amount.into_signed()?;
+        let signed_atoms = if negative {
+            -signed_amount.into_atoms()
+        } else {
+            signed_amount.into_atoms()
+        };
+        Some(Self::from_atoms(signed_atoms))
+    }
```
iljakuklic, line 87: Seems to me a fair portion of parsing and printing code could be reused between this and the unsigned amount

iljakuklic (approved):

common/src/primitives/signed_amount.rs
TheQuantumPhysicist: Already done that using Amount. Not sure what you mean.

common/src/primitives/amount.rs
```diff
@@ -56,16 +58,28 @@ fn remove_right_most_zeros_and_decimal_point(s: String) -> String {
...
+        let signed_atoms = amount.into_atoms();
+        let atoms: UnsignedIntType = signed_atoms.try_into().ok()?;
+        Some(Self::from_atoms(atoms))
+    }
+
+    pub fn into_signed(self) -> Option<SignedAmount> {
+        let atoms = self.val;
+        let signed_atoms: super::signed_amount::SignedIntType = atoms.try_into().ok()?;
+        Some(SignedAmount::from_atoms(signed_atoms))
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Because TryFrom requires an Error type and I'm just not into that right now... If there was a way to put Option into it, I'd use it.

common/src/primitives/signed_amount.rs
iljakuklic: Was looking at outdated version

common/src/primitives/amount.rs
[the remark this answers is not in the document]
iljakuklic: I guess we could just take the one built-in types use for numeric conversion errors in `TryFrom`
https://doc.rust-lang.org/std/num/struct.TryFromIntError.html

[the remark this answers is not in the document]
iljakuklic: Hmm I don't see a way to construct it, please ignore the last comment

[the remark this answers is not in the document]
TheQuantumPhysicist: OK, so I guess you abandoned this idea. I'm resolving this.

PR #362, opened 2022-08-18
Replace string ProtocolId with enum

stanislav-tkach:


p2p/src/net/types/protocol.rs
```diff
@@ -0,0 +1,148 @@
...
+                return None;
+            }
+        };
+
+        let version = SemVer::try_from(version)
+            .tap_err(|e| log::trace!("Ignoring protocol with malformed version: {val} ({e:?})"))
+            .ok()?;
+
+        Some(Protocol::new(protocol, version))
+    }
```
iljakuklic, line 74: I think you can turn this into an impl of `std::str::FromStr`, and get the `.parse()` method for free.

p2p/src/net/mock/peer.rs
```diff
@@ -13,18 +13,24 @@
...
 // limitations under the License.
 
+use std::sync::Arc;
+
+use futures::FutureExt;
+use tokio::sync::mpsc;
+
+use common::chain::ChainConfig;
+use logging::log;
+
```
altonen, line 23: why are there empty lines between these?

p2p/src/net/mock/peer.rs
```diff
@@ -97,10 +103,9 @@ impl Peer {
                             peer_id: self.local_peer_id,
                             version: *self.config.version(),
                             network: *self.config.magic_bytes(),
-                            protocols: vec![
-                                types::Protocol::new("floodsub", *self.config.version()),
-                                types::Protocol::new("ping", *self.config.version()),
-                            ],
+                            protocols: parse_protocols(["floodsub", "/ipfs/ping/1.0.0"])
```
altonen, line 106: I think this is incorrect. The ping protocol used by the mock interface is not called `/ipfs/ping/1.0.0` because the mock interface has no relation to IPFS. If you want to use a libp2p-like protocol name and not just `ping` as it was before, `/mintlayer/mock/ping/0.1.0` would be an appropriate name.

p2p/src/net/mock/peer.rs
```diff
@@ -127,10 +132,9 @@ impl Peer {
                         peer_id: self.local_peer_id,
                         version: *self.config.version(),
                         network: *self.config.magic_bytes(),
-                        protocols: vec![
-                            types::Protocol::new("floodsub", *self.config.version()),
-                            types::Protocol::new("ping", *self.config.version()),
-                        ],
+                        protocols: parse_protocols(["floodsub", "/ipfs/ping/1.0.0"])
```
altonen, line 135: Previous comment applies here as well

p2p/src/net/types/protocol.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Constructs a new protocol instance with given type and version.
+    pub const fn new(protocol: ProtocolType, version: SemVer) -> Self {
+        Self { protocol, version }
+    }
+
+    /// Parses a protocol from the given string.
+    ///
+    /// A string must contain the name of a protocol and a version after the last slash. For
+    /// example: "/ipfs/ping/1.0.0" or "/mintlayer/sync/0.1.0".
+    pub fn parse(val: &str) -> Option<Self> {
```
altonen, line 56: I think this is in the wrong place. Parsing the backend-specific protocol string into `Protocol` should be part of the backend code. So instead of this global `parse()` there should be a `TryFrom<String> for Protocol` inside `src/net/libp2p/service/connectivity.rs` which parses the libp2p-specific protocol strings into this global and implementation-agnostic protocol type. Same thing for mock interface as well.

p2p/src/net/types/protocol.rs
```diff
@@ -0,0 +1,148 @@
...
+pub struct Protocol {
+    version: SemVer,
+    protocol: ProtocolType,
+}
+
+/// Protocol type.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Encode, Decode)]
+pub enum ProtocolType {
+    /// The publish/subscription protocol.
+    Pubsub,
```
altonen, line 38: I think this is inconsistent naming because at least `src/pubsub.rs` uses `PubSub`. I don't mind changing `PubSub` to `Pubsub` but I think it would be better if everything was named the same way.

p2p/src/peer_manager/mod.rs
```diff
@@ -115,50 +129,36 @@ where
...
-            "/meshsub/1.1.0",
-            "/meshsub/1.0.0",
-            "/ipfs/ping/1.0.0",
-            "/ipfs/id/1.0.0",
-            "/ipfs/id/push/1.0.0",
-            "/mintlayer/sync/0.1.0",
+    fn validate_supported_protocols(&self, protocols: &HashSet<Protocol>) -> bool {
+        const REQUIRED: &[Protocol] = &[
+            Protocol::new(ProtocolType::Pubsub, SemVer::new(1, 0, 0)),
+            Protocol::new(ProtocolType::Pubsub, SemVer::new(1, 1, 0)),
```
altonen, line 150: It is a bit of an unfortunate detail of libp2p that they advertise both `/meshsub/1.1.0` and `/meshsub/1.0.0` but it is what it is. I think the correct way, as long as they keep advertising both, is to either parse the combination of these two into `Protocol::new(ProtocolType::Pubsub, SemVer::new(1, 1, 0))` (so there's only one PubSub protocol reported to `PeerManager`) or the check for required protocols is adjusted to verify that there exists at least one PubSub protocol that is compatible with our implementation.

The reason is that the mock interface won't have two PubSub implementations for now at least so requiring two versions of the protocol from `MockService` doesn't make sense.

p2p/src/peer_manager/tests/ban.rs
```diff
@@ -417,3 +384,16 @@ async fn inbound_connection_invalid_magic_libp2p() {
 async fn inbound_connection_invalid_magic_mock() {
     inbound_connection_invalid_magic::<MockService>(make_mock_addr(), make_mock_addr()).await;
 }
+
+// TODO: Move to `p2p-test-utils`?
```
altonen, line 388: It should probably go there and it should be different for mock and libp2p. That is because mock interface doesn't know what IPFS or meshsub are so while sharing the protocol names is convenient, it's also leaking implementation details of libp2p into mock.

altonen (commented):
Left some comments. Overall it looks good but the protocol strings should be parsed inside the networking backend into `Protocol` types because they are different for networking backend implementations and identify different, low-level protocol implementations.

p2p/src/net/mock/peer.rs
stanislav-tkach: Personally I prefer to separate imports into a several groups: std, third-party crates, workspace crates and imports from the current crate. We don't have such convention, so I usually try not to introduce unnecessary changes, but I got carried away with this pull request. I can revert these changes if you don't like them.

p2p/src/net/mock/peer.rs
stanislav-tkach: I have reverted the ping protocol name, but with the current `parse_protocols` implementation it produces an empty set. Are we OK with that?

p2p/src/peer_manager/tests/ban.rs
stanislav-tkach: I have tried to do that, but I got the weird error:
```
error[E0308]: mismatched types
   --> p2p/src/peer_manager/tests/ban.rs:198:28
    |
198 |                 protocols: p2p_test_utils::default_protocols(),
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `net::types::protocol::Protocol`, found struct `p2p::net::types::protocol::Protocol`
    |
    = note: expected struct `std::collections::HashSet<net::types::protocol::Protocol>`
               found struct `std::collections::HashSet<p2p::net::types::protocol::Protocol>
```
I have only one idea: the compiler considers the types to be different because in one case it is reexported by the `p2p-test-utils` crate.

p2p/src/net/types/protocol.rs
stanislav-tkach: I thought about that, but I don't want to introduce an error type required by the `FromStr` trait, because it is ignored anyway. Now fixed. 🙃

p2p/src/net/types/protocol.rs
stanislav-tkach: Makes sens, though trait implementations are kind of global.

Do I understand correctly that mocks should have an independent implementation of protocols parsing?

stanislav-tkach:
I think that I fixed all comments, but I didn't resolved some of them in case I got something wrong.

p2p/src/net/types/mod.rs
```diff
@@ -49,9 +55,8 @@ pub struct PeerInfo<T: NetworkingService> {
     /// User agent of the peer
     pub agent: Option<String>,
 
-    // TODO: protocolid must not generic!
-    /// List of supported protocols
-    pub protocols: Vec<T::ProtocolId>,
+    /// A set of supported protocols.
+    pub protocols: HashSet<Protocol>,
```
TheQuantumPhysicist, line 59: Please use a BTreeSet. Usually the number of supported protocols is very small, hence, no need for a hash set. Generally, we try to avoid HashMap and HashSet unless we definitely need them because that typically causes confusion in hash function uses since we're doing blockchain and hashing is everywhere.

p2p/src/net/mock/peer.rs
altonen: yes it looks good

common/src/primitives/semver.rs
```diff
@@ -15,15 +15,15 @@
 
 use serialization::{Decode, Encode};
 
-#[derive(Debug, PartialEq, Eq, Encode, Decode, Copy, Clone)]
+#[derive(Debug, PartialEq, Eq, Encode, Decode, Copy, Clone, Hash)]
```
TheQuantumPhysicist, line 18: Replace this with Ord and PartialOrd. At least, having a correct and defined order for SemVer is a win anyway.

TheQuantumPhysicist: Btw, from what I'm seeing in the implementation, the ordering will be correct by just deriving. Please add a test or two to ensure that this will remain to be the case.

p2p/src/net/types/protocol.rs
```diff
@@ -0,0 +1,61 @@
...
+    protocol: ProtocolType,
+}
+
+/// Protocol type.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Encode, Decode)]
+pub enum ProtocolType {
+    /// The publish/subscription protocol.
+    PubSub,
+    /// The ping protocol can be used as a simple application-layer health check. Corresponds to
+    /// the "/ipfs/ping" protocol.
```
altonen, line 34: It doesn't necessarily correspond to the `/ipfs/ping` protocol

p2p/src/peer_manager/mod.rs
```diff
@@ -115,50 +129,45 @@ where
...
+    /// Checks that the given set of protocols contains the following protocols with the exact
+    /// versions:
     ///
-    /// - `/meshsub/1.1.0`
-    /// - `/meshsub/1.0.0`
-    /// - `/ipfs/ping/1.0.0`
-    /// - `/ipfs/id/1.0.0`
-    /// - `/ipfs/id/push/1.0.0`
-    /// - `/mintlayer/sync/0.1.0`
+    /// - PubSub 1.1.0
```
altonen, line 137: `PubSub 1.1.0` or `PubSub 1.0.0`

p2p/src/net/types/protocol.rs
```diff
@@ -0,0 +1,61 @@
...
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Encode, Decode)]
+pub struct Protocol {
+    version: SemVer,
+    protocol: ProtocolType,
+}
+
+/// Protocol type.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Encode, Decode)]
+pub enum ProtocolType {
+    /// The publish/subscription protocol.
```
altonen, line 31:
```suggestion
    /// The publish/subscribe protocol.
```

p2p/src/peer_manager/tests/ban.rs
altonen: But this function actually could be moved to `src/peer_manager/tests/mod.rs` so all `PeerManager` tests could use the function

altonen (approved):
Looks good, only had some nitpicking about documentation

p2p/src/net/types/protocol.rs
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+use std::fmt::{self, Display, Formatter};
+
+use common::primitives::semver::SemVer;
+use serialization::{Decode, Encode};
+
+/// Protocol type and version.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Encode, Decode)]
+pub struct Protocol {
```
TheQuantumPhysicist, line 23: I think calling this name "Protocol" will be a ripe source of confusion. Perhaps "SupportedProtocol"? Though also supported protocol has its problems that I'll mention in another comment.

p2p/src/peer_manager/mod.rs
```diff
@@ -115,50 +129,45 @@ where
...
-    // TODO: define better protocol id type
-    fn validate_supported_protocols(&self, protocols: &[T::ProtocolId]) -> bool {
-        const REQUIRED: &[&str] = &[
-            "/meshsub/1.1.0",
-            "/meshsub/1.0.0",
-            "/ipfs/ping/1.0.0",
-            "/ipfs/id/1.0.0",
-            "/ipfs/id/push/1.0.0",
-            "/mintlayer/sync/0.1.0",
+    fn validate_supported_protocols(&self, protocols: &HashSet<Protocol>) -> bool {
```
TheQuantumPhysicist, line 147: So. Here's the other problem I wanted to talk about. Usually protocol comparison is upgrade-based. Meaning, the general case is expected compatibility, not incompatibility. Let's say we support 5 protocols. All have 5 names and 5 versions. Now the idea that we have 5 protocol versions is a big issue already. There should be one "Mintlayer protocol version". If we upgrade one of the five protocols, we shouldn't check for the compatibility of each individual protocol, but only for the Mintlayer upgrade. So if we upgrade from version 0.1.0 to 0.2.0, we check that the peer has `0.1.0 <= version <= 0.2.0`. That's much more maintainable that having 5 different versions.

Now I understand that this whole 5 different versions thing is inherent from the Behaviour thing in libp2p... and I don't know how to solve this problem. But maybe we should have a discussion to see how this should work. This whole substream thing is becoming messy very quickly.

TheQuantumPhysicist (approved):
Besides the minor things I said that can be fixed (the hash thing and the Protocol struct name), I'm happy to merge this. The issue of protocol versioning is a big mess that we can probably deal with later, but we should discuss ASAP before Aaro leaves.

p2p/src/net/mod.rs
```diff
@@ -46,9 +49,6 @@ pub trait NetworkingService {
     /// Unique ID assigned to each received request from a peer
     type SyncingPeerRequestId: Debug + Eq + Hash + Send + Sync;
 
-    /// Id that identifies a protocol
-    type ProtocolId: Clone + Debug + Display + Eq + PartialEq + Send;
-
```
iljakuklic, line 51: Out of curiosity, why did you decide to remove this associated type completely instead specifying the newly introduced enum as the `ProtocolId` type in `impl`s?

iljakuklic (approved):

p2p/src/net/mod.rs
stanislav-tkach: There was a TODO-comment in the `p2p/src/net/types.rs` [file](https://github.com/mintlayer/mintlayer-core/pull/362/files#diff-df7dc133e821c16663d902bb26b75e37f8e19cf19f2d318452ab4aaee4fb224cL52) that said that a protocol id shouldn't be generic. Perhaps I misinterpret it, but I still don't see any benefits in making generic a thing that will probably never be changed.

p2p/src/net/types/protocol.rs
stanislav-tkach: I agree with you about the naming, but I don't like `SupportedProtocol` either. I think that the emphasis should be on what this protocol is or at least on its "level". I mean that we want to abstract libp2p "raw" p2p protocols and introduce our higher level protocol. I don't have better ideas than "application protocol" for now, that I also don't like.

stanislav-tkach: https://github.com/mintlayer/mintlayer-core/issues/371

PR #363, opened 2022-08-18
Implement chainstate box delegation

TheQuantumPhysicist:
We have faced an issue yesterday where `Box<dyn ChainstateInterface>` doesn't implement `ChainstateInterface`. So this is supposed to be the solution to this problem. What do you guys think?

### DIFF
--- a/chainstate/src/chainstate_interface_impl_delegation.rs
+++ b/chainstate/src/chainstate_interface_impl_delegation.rs
@@ -108,7 +108,7 @@ mod tests {
         make_chainstate, ChainstateConfig,
     };
 
-    fn test_box_interface_ref<C: ChainstateInterface>(chainstate: &C, chain_config: &ChainConfig) {
+    fn test_interface_ref<C: ChainstateInterface>(chainstate: &C, chain_config: &ChainConfig) {
         assert_eq!(
             chainstate.get_best_block_id().unwrap(),
             chain_config.genesis_block_id()
@@ -119,7 +119,7 @@ mod tests {
         );
     }
 
-    fn test_box_interface<C: ChainstateInterface>(chainstate: C, chain_config: &ChainConfig) {
+    fn test_interface<C: ChainstateInterface>(chainstate: C, chain_config: &ChainConfig) {
         assert_eq!(
             chainstate.get_best_block_id().unwrap(),
             chain_config.genesis_block_id()
@@ -149,8 +149,8 @@ mod tests {
             )
             .unwrap();
 
-            test_box_interface_ref(&boxed_chainstate, &*chain_config);
-            test_box_interface(boxed_chainstate, &*chain_config);
+            test_interface_ref(&boxed_chainstate, &*chain_config);
+            test_interface(boxed_chainstate, &*chain_config);
         });
     }
 }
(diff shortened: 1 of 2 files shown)


stanislav-tkach (approved):
I think it is OK, except for the missing license header. 🙃

stanislav-tkach:
I would check if there are crates that simplify such implementation. I have tried [ambassador](https://crates.io/crates/ambassador) and [delegate](https://crates.io/crates/delegate). I wasn't entirely happy with them, but it was long ago. For now I think that the manual implementation is OK.

chainstate/src/chainstate_interface_impl_delegation.rs
```diff
@@ -0,0 +1,96 @@
...
+        locator: Locator,
+    ) -> Result<Vec<common::chain::block::BlockHeader>, ChainstateError> {
+        self.deref().get_headers(locator)
+    }
+
+    fn filter_already_existing_blocks(
+        &self,
+        headers: Vec<common::chain::block::BlockHeader>,
+    ) -> Result<Vec<common::chain::block::BlockHeader>, ChainstateError> {
+        self.deref().filter_already_existing_blocks(headers)
```
iljakuklic, line 94: Have you checked whet all the `deref` and `deref_mut` calls resolve to? There is an impl of `Deref` for `&T` and `&mut T`. Since `self` is either a reference or a mut reference, we should try to call these methods to see if this does not end up being an infinite recursion.

TheQuantumPhysicist: Yesterday I did a quick test of that and made sure it works. In fact the compiler does warn you of recursion if you don't explicitly call deref/deref_mut. However, to make sure that this works indefinitely, I added a test. Check the test out.

iljakuklic (approved):

muursh (approved):
I'm happy with this

altonen (approved):

chainstate/src/chainstate_interface_impl_delegation.rs
```diff
@@ -0,0 +1,156 @@
...
+        &self,
+        locator: Locator,
+    ) -> Result<Vec<common::chain::block::BlockHeader>, ChainstateError> {
+        self.deref().get_headers(locator)
+    }
+
+    fn filter_already_existing_blocks(
+        &self,
+        headers: Vec<common::chain::block::BlockHeader>,
+    ) -> Result<Vec<common::chain::block::BlockHeader>, ChainstateError> {
```
azarovh, line 91: use common::chain::block::BlockHeader at the top?

chainstate/src/lib.rs
```diff
@@ -14,6 +14,7 @@
 // limitations under the License.
 
 pub mod chainstate_interface;
+pub mod chainstate_interface_impl_delegation;
```
azarovh, line 17: why not chainstate_interface_delegation?

azarovh (approved):

chainstate/src/lib.rs
TheQuantumPhysicist: Because we're delegating the implementation, not the interface.

chainstate/src/chainstate_interface_impl_delegation.rs
TheQuantumPhysicist: Doesn't matter... if I find another issue and push again, I'll change it.

PR #364, opened 2022-08-18
Add utxo simulation test with undo

azarovh:
- Added another simulation test that focuses on creating utxo from block reward, spending utxo from tx and undoing utxo spending.
- Added function that creates utxo from block reward
- Cleaned up some tests in utxo/storage

P.S. simulation_with_undo.rs is almost copy/paste of simulation.rs, but generalisation of this code is imho an overkill and would affect readability. Besides we don't plan to add more tests like these.

### DIFF
--- a/utxo/src/cache.rs
+++ b/utxo/src/cache.rs
@@ -87,7 +87,7 @@ impl<'a> UtxosCache<'a> {
         self.current_block_hash = block_hash;
     }
 
-    /// Given a block reward adds its outputs to the utxo set
+    /// Given a block reward add its outputs to the utxo set
     pub fn add_utxos_from_block_reward(
         &mut self,
         reward: &BlockReward,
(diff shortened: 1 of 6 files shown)


TheQuantumPhysicist (approved):

stanislav-tkach (approved):

utxo/src/cache.rs
```diff
@@ -87,7 +87,31 @@ impl<'a> UtxosCache<'a> {
         self.current_block_hash = block_hash;
     }
 
-    pub fn add_utxos(
+    /// Given a block reward adds its outputs to the utxo set
```
muursh, line 90: `adds` -> `add`

utxo/src/cache.rs
```diff
@@ -87,7 +87,31 @@ impl<'a> UtxosCache<'a> {
...
+                true
+            };
+            let utxo = Utxo::new(output.clone(), true, source.clone());
+
+            self.add_utxo(&outpoint, utxo, overwrite)?;
+        }
+        Ok(())
+    }
+
+    /// Given a transaction adds its outputs to the utxo set
```
muursh, line 113: Same plural issue in the comment here too

muursh (approved):

PR #366, opened 2022-08-19
Update crates

TheQuantumPhysicist:
So this one turned out to be a little bigger than usual, especially for libp2p and libtest-mimic. Lots of breaking changes.


stanislav-tkach (approved):

deny.toml
```diff
@@ -52,6 +52,12 @@ skip = [
     {name = "windows_x86_64_msvc" },
     {name = "curve25519-dalek"},
     {name = "hashbrown"},
+    {name = "universal-hash"},
```
stanislav-tkach, line 55: Does it make sense to deny multiple versions if we usually add a problematic crate to the list of exceptions anyway?

TheQuantumPhysicist: I'm not sure what you mean with "problematic crate". But just to tell you my opinion in general, I'm not happy that we have to keep doing this. It kind of defeats the purpose. The only reason I'm putting up with this nonsense is because we want to ensure that scale serialization has a consistent version. To me, everything else kind of doesn't matter.

stanislav-tkach: As far as I can see, there is no way to deny multiple versions of one specific crate, so I suppose we don't have a choice and will continue to use `multiple-versions = "deny"`.

altonen (approved):

deny.toml
```diff
@@ -52,6 +52,12 @@ skip = [
...
+    {name = "poly1305"},
+    {name = "cipher"},
+    {name = "chacha20poly1305"},
+    {name = "chacha20"},
+    {name = "aead"},
```
iljakuklic, line 60: It is possible some item can now be removed from this list and other cargo deny "exception" lists

iljakuklic (approved):
Good to go

deny.toml
TheQuantumPhysicist: Same idea occurred to me... but honestly I don't care right now. Maybe eventually we can right a script to check that.

PR #368, opened 2022-08-22
Specify version for skipped crates in the cargo deny bans check

stanislav-tkach:
It is recommended in the [documentation](https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-skip-field-optional) to specify the version for skipped crates and it makes sense because in that case the linter will produce a warning if a suppression can be removed.

Additionally I have removed several crates that we no longer have several versions of.


TheQuantumPhysicist (approved):
Thank you. I'm all for that.

deny.toml
```diff
@@ -8,56 +8,49 @@ multiple-versions = "deny" #catch mutliple versions of a crate
...
+    {name = "windows_i686_gnu", version = "=0.34.0"},
+    {name = "windows_i686_msvc", version = "=0.32.0"},
+    {name = "windows_i686_msvc", version = "=0.34.0"},
+    {name = "windows_x86_64_gnu", version = "=0.32.0"},
+    {name = "windows_x86_64_gnu", version = "=0.34.0"},
+    {name = "windows_x86_64_msvc", version = "=0.32.0"},
+    {name = "windows_x86_64_msvc", version = "=0.34.0"},
+    {name = "curve25519-dalek", version = "=3.2.0"},
+    {name = "universal-hash", version = "=0.4.1"},
+    {name = "poly1305", version = "0.7.2"},
```
muursh, line 49: Why doesn't this have the "=" inside the quotation marks too? Am I missing something obvious?

muursh (commented):
This seems to have broken cargo deny too

deny.toml
stanislav-tkach: By default the Cargo (and it seems that cargo deny uses the same behavior) version requirements are relaxed, for example:
```
1.2.3  :=  >=1.2.3, <2.0.0
1.2    :=  >=1.2.0, <2.0.0
1      :=  >=1.0.0, <2.0.0
```
So "=" means "use exactly this version".

muursh: Yeah I get the idea. My question here perhaps a little more general, as to why we care for some of these and not others

muursh: Is there a reason we treat the symmetric crypto stuff differently here?

stanislav-tkach:
> This seems to have broken cargo deny too

It is strange, because I have no errors locally. I'm going to investigate.

deny.toml
stanislav-tkach: Not really, I have just forgot to put "=" to some crates. 🙃  Fixed.

muursh (approved):

iljakuklic (approved):

Issue #318, opened 2022-07-28
Introduce Mempool validity function in tests


royster57:
Introduce a new method to `MempoolStore`, called say `assert_valid`, that is enabled for testing (does nothing otherwise) and contains a number of potentially expensive integrity checks:

* Every transaction ID form `txs_by_id` occurs in `txs_by_descendant_score` exactly once
* For each entry, each child has the entry as a parent
* And so on, there are many more of these I can imagine

Then sprinkle calls to it throughout the tests or into implementation of methods that are expected to preserve these invariants.

_Originally posted by @iljakuklic in https://github.com/mintlayer/mintlayer-core/pull/43#discussion_r931312677_

PR #376, opened 2022-08-23
Roy/mempool consistency

royster57:
Addresses #318

### DIFF
--- a/mempool/src/pool/store.rs
+++ b/mempool/src/pool/store.rs
@@ -123,6 +123,32 @@ impl MempoolStore {
         self.txs_by_id.get(&id.get())
     }
 
+    pub(super) fn assert_valid(&self) {
+        #[cfg(test)]
+        self.assert_valid_inner()
+    }
+
+    #[cfg(test)]
+    pub(super) fn assert_valid_inner(&self) {
+        let entries = self.txs_by_descendant_score.values().flatten().collect::<Vec<_>>();
+        for id in self.txs_by_id.keys() {
+            assert_eq!(
+                entries.iter().filter(|entry_id| ***entry_id == Id::new(*id)).count(),
+                1
+            )
+        }
+        for entry in self.txs_by_id.values() {
+            for child in &entry.children {
+                assert!(self
+                    .txs_by_id
+                    .get(&child.get())
+                    .expect("child")
+                    .parents
+                    .contains(&entry.tx_id()))
+            }
+        }
+    }
+
     fn append_to_parents(&mut self, entry: &TxMempoolEntry) {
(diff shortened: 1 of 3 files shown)


TheQuantumPhysicist (approved):

mempool/src/pool/store.rs
```diff
@@ -123,6 +123,32 @@ impl MempoolStore {
         self.txs_by_id.get(&id.get())
     }
 
+    pub(super) fn assert_valid(&self) {
```
stanislav-tkach, line 126: I think it would be nice to give this function a name that emphasizes that it only performs some checks when the `test` feature is defined, though I cannot suggest a better name.

stanislav-tkach (approved):

mempool/src/pool/store.rs
```diff
@@ -123,6 +123,32 @@ impl MempoolStore {
...
+        #[cfg(test)]
+        self.assert_valid_inner()
+    }
+
+    #[cfg(test)]
+    pub(super) fn assert_valid_inner(&self) {
```
iljakuklic, line 132: I think this can just be private if I'm not mistaken.

mempool/src/pool/store.rs
```diff
@@ -123,6 +123,32 @@ impl MempoolStore {
...
+        for entry in self.txs_by_id.values() {
+            for child in &entry.children {
+                assert!(self
+                    .txs_by_id
+                    .get(&child.get())
+                    .expect("child")
+                    .parents
+                    .contains(&entry.tx_id()))
+            }
+        }
```
iljakuklic, line 149: Cool stuff. Do you plan to extend this further? I guess there are still more consistency invariants not covered by these checks.

iljakuklic (approved):

mempool/src/pool/store.rs
royster57: Good catch, thanks.

mempool/src/pool/store.rs
royster57: There's plenty more than can be checked I'm sure, haven't given it much thought yet but at least wanted the two invariants that you mentioned in the original PR to be covered :)
If anything else comes to mind feel free to suggest and we'll add it.
For example, a similar check as was done for txs_by_descendant_score can be done for txs_by_creation_time.

mempool/src/pool/store.rs
royster57: I see what you're saying, just can't seem to think of a name either... Something like `assert_valid_test_only` would make me cringe :grimacing:

TheQuantumPhysicist:
If there's nothing blocking this, please go ahead and merge it.

PR #378, opened 2022-08-24
Integration of utxo crate in chainstate

azarovh:
~~This is initial draft integration just to check that I'm on the right track. 
The code still doesn't fully compile and I didn't touch tests at all.~~

The main idea is that I added `utxo` into `transactino_verifier` side by side with `tx_index_cache`. Transaction index that was responsible for the logic of tracking spent outputs is left almost without changes. It should be modified in the next separate PR to clean up the code and avoid duplicated logic.

### DIFF
--- a/chainstate-storage/src/internal/test.rs
+++ b/chainstate-storage/src/internal/test.rs
@@ -302,7 +302,7 @@ pub fn create_rand_block_undo(
         tx_undo.push(TxUndo::new(tx_utxos));
     }
 
-    BlockUndo::new(reward_undo, tx_undo)
+    BlockUndo::new(Some(reward_undo), tx_undo)
 }
 
 #[cfg(not(loom))]
--- a/chainstate-storage/src/internal/test.rs
+++ b/chainstate-storage/src/internal/test.rs
@@ -21,7 +21,7 @@ use crypto::key::{KeyKind, PrivateKey};
 use crypto::random::Rng;
 use rstest::rstest;
 use test_utils::random::{make_seedable_rng, Seed};
-use utxo::{BlockUndo, TxUndo};
+use utxo::{BlockRewardUndo, BlockUndo, TxUndo};
 
 type TestStore = crate::inmemory::Store;
 
@@ -277,27 +277,32 @@ pub fn create_rand_block_undo(
     rng: &mut impl Rng,
     max_lim_of_utxos: u8,
     max_lim_of_tx_undos: u8,
-    block_height: BlockHeight,
 ) -> BlockUndo {
     let mut counter: u64 = 0;
 
-    let mut block_undo: Vec<TxUndo> = vec![];
+    let mut reward_utxos: Vec<Utxo> = vec![];
+    let utxo_rng = rng.gen_range(1..max_lim_of_utxos);
+    for i in 0..utxo_rng {
+        counter += u64::from(i);
+        reward_utxos.push(create_rand_utxo(rng, counter));
+    }
+    let reward_undo = BlockRewardUndo::new(reward_utxos);
 
+    let mut tx_undo = vec![];
     let undo_rng = rng.gen_range(1..max_lim_of_tx_undos);
     for _ in 0..undo_rng {
-        let mut tx_undo = vec![];
-
+        let mut tx_utxos = vec![];
         let utxo_rng = rng.gen_range(1..max_lim_of_utxos);
         for i in 0..utxo_rng {
             counter += u64::from(i);
 
-            tx_undo.push(create_rand_utxo(rng, counter));
+            tx_utxos.push(create_rand_utxo(rng, counter));
         }
 
-        block_undo.push(TxUndo::new(tx_undo));
+        tx_undo.push(TxUndo::new(tx_utxos));
     }
 
-    BlockUndo::new(block_undo, block_height)
+    BlockUndo::new(reward_undo, tx_undo)
 }
 
 #[cfg(not(loom))]
@@ -306,7 +311,7 @@ pub fn create_rand_block_undo(
 #[case(Seed::from_entropy())]
 fn undo_test(#[case] seed: Seed) {
     let mut rng = make_seedable_rng(seed);
-    let block_undo0 = create_rand_block_undo(&mut rng, 10, 5, BlockHeight::new(1));
+    let block_undo0 = create_rand_block_undo(&mut rng, 10, 5);
     // create id:
     let id0: Id<Block> = Id::new(H256::random());
 
@@ -325,7 +330,7 @@ fn undo_test(#[case] seed: Seed) {
 
     // insert, remove, and reinsert the next block_undo
 
-    let block_undo1 = create_rand_block_undo(&mut rng, 5, 10, BlockHeight::new(2));
+    let block_undo1 = create_rand_block_undo(&mut rng, 5, 10);
     // create id:
     let id1: Id<Block> = Id::new(H256::random());
 
--- a/chainstate-storage/src/internal/test.rs
+++ b/chainstate-storage/src/internal/test.rs
@@ -278,26 +278,23 @@ pub fn create_rand_block_undo(
     max_lim_of_utxos: u8,
     max_lim_of_tx_undos: u8,
 ) -> BlockUndo {
-    let mut counter: u64 = 0;
-
-    let mut reward_utxos: Vec<Utxo> = vec![];
     let utxo_rng = rng.gen_range(1..max_lim_of_utxos);
-    for i in 0..utxo_rng {
-        counter += u64::from(i);
-        reward_utxos.push(create_rand_utxo(rng, counter));
-    }
+    let reward_utxos = (0..utxo_rng)
+        .into_iter()
+        .enumerate()
+        .map(|(i, _)| create_rand_utxo(rng, i as u64))
+        .collect();
     let reward_undo = BlockRewardUndo::new(reward_utxos);
 
     let mut tx_undo = vec![];
     let undo_rng = rng.gen_range(1..max_lim_of_tx_undos);
     for _ in 0..undo_rng {
-        let mut tx_utxos = vec![];
         let utxo_rng = rng.gen_range(1..max_lim_of_utxos);
-        for i in 0..utxo_rng {
-            counter += u64::from(i);
-
-            tx_utxos.push(create_rand_utxo(rng, counter));
-        }
+        let tx_utxos = (0..utxo_rng)
+            .into_iter()
+            .enumerate()
+            .map(|(i, _)| create_rand_utxo(rng, i as u64))
+            .collect();
 
         tx_undo.push(TxUndo::new(tx_utxos));
     }
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -102,7 +102,7 @@ impl BanScore for ConnectTransactionError {
             &ConnectTransactionError::MissingBlockUndo(_) => 0,
             &ConnectTransactionError::MissingBlockRewardUndo(_) => 0,
             &ConnectTransactionError::MissingTxUndo(_, _) => 0,
-            &ConnectTransactionError::UtxoInvariantBroken(_) => 0,
+            &ConnectTransactionError::UtxoInvariantBroken => 0,
         }
     }
 }
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -100,6 +100,7 @@ impl BanScore for ConnectTransactionError {
             ConnectTransactionError::SerializationInvariantError(_) => 100,
             ConnectTransactionError::TimeLockViolation => 100,
             &ConnectTransactionError::MissingBlockUndo(_) => 0,
+            &ConnectTransactionError::MissingBlockRewardUndo(_) => 0,
             &ConnectTransactionError::MissingTxUndo(_, _) => 0,
             &ConnectTransactionError::UtxoInvariantBroken(_) => 0,
         }
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -101,8 +101,8 @@ impl BanScore for ConnectTransactionError {
             ConnectTransactionError::TimeLockViolation => 100,
             &ConnectTransactionError::MissingBlockUndo(_) => 0,
             &ConnectTransactionError::MissingBlockRewardUndo(_) => 0,
-            &ConnectTransactionError::MissingTxUndo(_, _) => 0,
-            &ConnectTransactionError::UtxoInvariantBroken => 0,
+            ConnectTransactionError::MissingTxUndo(_, _) => 0,
+            ConnectTransactionError::UtxoError(err) => err.ban_score(),
         }
     }
 }
@@ -168,6 +168,20 @@ impl BanScore for BlockSizeError {
             BlockSizeError::Header(_, _) => 100,
             BlockSizeError::SizeOfTxs(_, _) => 100,
             BlockSizeError::SizeOfSmartContracts(_, _) => 100,
+        }
+    }
+}
+
+impl BanScore for utxo::Error {
+    fn ban_score(&self) -> u32 {
+        match self {
+            utxo::Error::OverwritingUtxo => 0,
+            utxo::Error::FreshUtxoAlreadyExists => 0,
+            utxo::Error::UtxoAlreadySpent(_) => 100,
+            utxo::Error::NoUtxoFound => 0,
+            utxo::Error::NoBlockchainHeightFound => 0,
+            utxo::Error::MissingBlockRewardUndo(_) => 0,
+            utxo::Error::DBError(_) => 0,
         }
     }
 }
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -99,6 +99,9 @@ impl BanScore for ConnectTransactionError {
             // Even though this is an invariant, we consider it a violation to be overly cautious
             ConnectTransactionError::SerializationInvariantError(_) => 100,
             ConnectTransactionError::TimeLockViolation => 100,
+            &ConnectTransactionError::MissingBlockUndo(_) => 0, //???
+            &ConnectTransactionError::MissingTxUndo(_, _) => 0, //???
+            &ConnectTransactionError::UtxoInvariantBroken(_) => 0, //???
         }
     }
 }
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -99,9 +99,9 @@ impl BanScore for ConnectTransactionError {
             // Even though this is an invariant, we consider it a violation to be overly cautious
             ConnectTransactionError::SerializationInvariantError(_) => 100,
             ConnectTransactionError::TimeLockViolation => 100,
-            &ConnectTransactionError::MissingBlockUndo(_) => 0, //???
-            &ConnectTransactionError::MissingTxUndo(_, _) => 0, //???
-            &ConnectTransactionError::UtxoInvariantBroken(_) => 0, //???
+            &ConnectTransactionError::MissingBlockUndo(_) => 0,
+            &ConnectTransactionError::MissingTxUndo(_, _) => 0,
+            &ConnectTransactionError::UtxoInvariantBroken(_) => 0,
         }
     }
 }
--- a/chainstate/src/detail/ban_score.rs
+++ b/chainstate/src/detail/ban_score.rs
@@ -99,8 +99,8 @@ impl BanScore for ConnectTransactionError {
             // Even though this is an invariant, we consider it a violation to be overly cautious
             ConnectTransactionError::SerializationInvariantError(_) => 100,
             ConnectTransactionError::TimeLockViolation => 100,
-            &ConnectTransactionError::MissingBlockUndo(_) => 0,
-            &ConnectTransactionError::MissingBlockRewardUndo(_) => 0,
+            ConnectTransactionError::MissingBlockUndo(_) => 0,
+            ConnectTransactionError::MissingBlockRewardUndo(_) => 0,
             ConnectTransactionError::MissingTxUndo(_, _) => 0,
             ConnectTransactionError::UtxoError(err) => err.ban_score(),
         }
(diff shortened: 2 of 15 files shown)


chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -181,97 +185,27 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
-                    Self::get_output_value(&outputs, output_index, (*block_id).into()).cloned()?
-                }
-            };
-            match output_amount {
-                OutputValue::Coin(output_amount) => {
-                    total = (total + output_amount)
-                        .ok_or(ConnectTransactionError::InputAdditionError)?
-                }
-            }
+        if let Entry::Vacant(entry) = self.tx_index_cache.entry(outpoint.tx_id()) {
```
TheQuantumPhysicist, line 200: I have this deep grudge against if let because it's heavily overused. I use it only in very specific situations. I prefer match because it lays out the possibilities clearly for the developer, and it doesn't allow you to miss new arms added in the future (irrelevant for this case). There's no practical benefit of the change you did here. If you want to get rid of the reference returned, you can return () in the other arm, but I still find this unnecessary as it'll be optimized out by the compiler.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -522,6 +459,7 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
         spend_height: &BlockHeight,
         median_time_past: &BlockTimestamp,
         blockreward_maturity: &BlockDistance,
+        utxo_block_undo: &mut BlockUndo,
```
TheQuantumPhysicist, line 462: This is heavily frawned upon. We don't pass references of mutable variables because it can easily lead to undefined behavior.

The simplest solution here is to return the tx_undo, and then you can construct the undo at the called level.

utxo/src/undo.rs
```diff
@@ -14,10 +14,22 @@
...
 use crate::Utxo;
-use common::primitives::BlockHeight;
 use serialization::{Decode, Encode};
 
-#[derive(Debug, Clone, Eq, PartialEq, Encode, Decode)]
+#[derive(Default, Debug, Clone, Eq, PartialEq, Encode, Decode)]
+pub struct BlockRewardUndo(Vec<Utxo>);
+
+impl BlockRewardUndo {
+    pub fn new(utxos: Vec<Utxo>) -> Self {
```
TheQuantumPhysicist, line 23: I don't think this is sufficient. You need to also undo kernel spends in ConsensusData.

chainstate/src/detail/chainstateref.rs
```diff
@@ -639,19 +651,37 @@ impl<'a, S: BlockchainStorageWrite, O: OrphanBlocksMut> ChainstateRef<'a, S, O>
         spend_height: &BlockHeight,
         blockreward_maturity: &BlockDistance,
     ) -> Result<(), BlockError> {
-        let connected_txs =
-            self.make_cache_with_connected_transactions(block, spend_height, blockreward_maturity)?;
-        let consumed = connected_txs.consume()?;
+        let mut utxo_db = UtxosDB::new(&mut self.db_tx);
+        let mut block_undo: BlockUndo = Default::default();
```
TheQuantumPhysicist, line 655: Why not make this a member of TransactionVerifier? Doesn't it get flushed with everything else on success?

TheQuantumPhysicist (commented):
Very close to what we need. Moving in the right direction. Please ensure you take the comments I made into account.

chainstate/src/detail/transaction_verifier/mod.rs
azarovh: The reason I changed match to if let here is that it is more clean. I'm new to the code and things like unused variables and empty branches (that returns useless refs) are very distractive.

chainstate/src/detail/chainstateref.rs
azarovh: `transaction_verifier` doesn't work on the block level but with separate transactions. It lacks the context on how it is used - to connect or disconnect txs and whether it should set or del BlockUndo. So I decided to keep this logic here

utxo/src/undo.rs
azarovh: Aren't kernel spends just inputs for BlockRewardTransactable?

chainstate/src/detail/transaction_verifier/mod.rs
azarovh: agree

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: Still, when you're connecting transactions, you're doing them at the block level. Think and see if you can make this cleaner.

utxo/src/undo.rs
TheQuantumPhysicist: Yep. They're being spent to create a block, and the block rewards reflects their value + reward. So, they get marked "spent", and then we have to unspend them when disconnecting.

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: > The reason I changed match to if let here is that it is more clean

That's quite subjective. To me, I call match cleaner because it's easier to maintain and clearer to the developer what the options are.

Please restore it. I understand it's not ideal for your taste, but please bear with me in such issues as I have goals that pertain to security more than things being nicer/uglier.

utxo/src/undo.rs
azarovh: Right, `BlockRewardUndo` stores this inputs and the `unspend_utxos_from_block_transactable` will add them to utxo set as unspent. I thought that I covered this, am I missing something?

chainstate/src/detail/transaction_verifier/mod.rs
azarovh: ok

utxo/src/undo.rs
TheQuantumPhysicist: I'm not sure. Please make sure they're being unspent when disconnecting a block. I may have misunderstood how this is meant to be used. Please keep in mind that we don't have tests for this in chainstate for this yet.

chainstate-storage/src/internal/test.rs
```diff
@@ -277,27 +277,32 @@ pub fn create_rand_block_undo(
...
     max_lim_of_tx_undos: u8,
-    block_height: BlockHeight,
 ) -> BlockUndo {
     let mut counter: u64 = 0;
 
-    let mut block_undo: Vec<TxUndo> = vec![];
+    let mut reward_utxos: Vec<Utxo> = vec![];
+    let utxo_rng = rng.gen_range(1..max_lim_of_utxos);
+    for i in 0..utxo_rng {
+        counter += u64::from(i);
```
TheQuantumPhysicist, line 286: Is this like a series sum, from 0 to n-1, yielding n(n-1)/2? Why?

Or is this a mistake and you just need a counter, in which case you could just use i?

Unlikely the former, so in case it's the latter, this can be simplified with a generator function pattern: `(0..utxo_rng).into_iter().map(|_|create_rand_utxo(rng, counter))` or something similar.

chainstate/src/detail/transaction_verifier/error.rs
```diff
@@ -86,6 +97,8 @@ pub enum ConnectTransactionError {
     SerializationInvariantError(Id<Block>),
     #[error("Timelock rules violated")]
     TimeLockViolation,
+    #[error("Utxo invariant broken: `{0}`")]
+    UtxoInvariantBroken(String),
```
TheQuantumPhysicist, line 101: Please, no string errors in this library. Every error has to have a different type so that it can be handled by a receiving library. Even though it's an invariant error.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -50,22 +57,28 @@ pub enum BlockTransactableRef<'a> {
...
     db_tx: &'a S,
     tx_index_cache: BTreeMap<OutPointSourceId, CachedInputsOperation>,
+    utxo_cache: UtxosCache<'a>,
+    utxo_block_undo: BTreeMap<Id<Block>, BlockUndoEntry>,
     chain_config: &'a ChainConfig,
 }
 
-impl<'a, S> TransactionVerifier<'a, S> {
-    pub fn new(db_tx: &'a S, chain_config: &'a ChainConfig) -> Self {
+impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
```
TheQuantumPhysicist, line 74: I don't think this bound that you added here is necessary. There are no operations related to it in this impl block.

chainstate-storage/src/internal/test.rs
azarovh: That's just the way the rest of the test was written. I'm not sure about the reasoning, but I agree that it makes sense to simplify.

utxo/src/undo.rs
```diff
@@ -46,26 +58,34 @@ impl TxUndo {
...
+    pub fn push_tx_undo(&mut self, tx_undo: TxUndo) {
+        self.tx_undos.push(tx_undo);
+    }
+
+    pub fn block_reward_undo(&self) -> Option<&BlockRewardUndo> {
+        self.reward_undo.as_ref()
+    }
+
+    pub fn set_block_reward_undo(&mut self, reward_undo: BlockRewardUndo) {
+        self.reward_undo = Some(reward_undo);
```
TheQuantumPhysicist, line 88: Maybe add an assert here that it wasn't set before? Not sure.

utxo/src/cache.rs
```diff
@@ -145,6 +153,73 @@ impl<'a> UtxosCache<'a> {
...
+        for (tx_in, utxo) in tx.inputs().iter().zip(tx_undo.inner().iter()) {
+            self.add_utxo(tx_in.outpoint(), utxo.clone(), false)?;
+        }
+        Ok(())
+    }
+
+    /// Marks the inputs of a transactable block reward as 'spent', adds outputs to the utxo set.
+    /// If BlockRewardTransactable has no inputs then just adds outputs to utxo set.
+    /// Returns a BlockRewardUndo if function is a success or an error if the input cannot be spent.
+    pub fn spend_utxos_from_block_transactable(
```
TheQuantumPhysicist, line 176: I think the right terminology here is "connect_block_transactable". Spending is the act of marking an input as spent and/or removing it from the utxo set. Here you're doing both, spending inputs and adding outputs, we call that connecting.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -538,35 +566,58 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
-                // spend inputs of this transaction
+                // mark tx index as spend
                 let spender = tx.get_id().into();
-                self.spend(tx.inputs(), spend_height, blockreward_maturity, spender)?;
+                self.spend_tx_index(tx.inputs(), spender)?;
+
+                //spend utxos
+                let tx_undo = self
+                    .utxo_cache
+                    .spend_utxos_from_tx(tx, *spend_height)
```
TheQuantumPhysicist, line 576: Same here. This should be probably called connect, not spend.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -516,6 +512,37 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
         Ok(())
     }
 
+    fn fetch_block_undo<'b>(
+        db_tx: &'a S,
+        utxo_block_undo: &'b mut BTreeMap<Id<Block>, BlockUndoEntry>,
```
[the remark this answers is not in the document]
azarovh: borrows checker is complaining that the utxo_cache field is borrowed here (though it is not used) and in the next line after this function call

TheQuantumPhysicist (approved):
I think this is fine, besides the few comments I made. Please wait for others to review it in case I missed something.

chainstate/src/detail/ban_score.rs
```diff
@@ -99,6 +99,10 @@ impl BanScore for ConnectTransactionError {
             // Even though this is an invariant, we consider it a violation to be overly cautious
             ConnectTransactionError::SerializationInvariantError(_) => 100,
             ConnectTransactionError::TimeLockViolation => 100,
+            &ConnectTransactionError::MissingBlockUndo(_) => 0,
+            &ConnectTransactionError::MissingBlockRewardUndo(_) => 0,
+            &ConnectTransactionError::MissingTxUndo(_, _) => 0,
+            &ConnectTransactionError::UtxoInvariantBroken(_) => 0,
```
stanislav-tkach, line 105: Could you please explain to me why these errors have zero ban score?

stanislav-tkach (approved):

chainstate/src/detail/ban_score.rs
TheQuantumPhysicist: Because it indicates something broken in the node and its invariants and not maliciousness of a peer.

chainstate/src/detail/chainstateref.rs
```diff
@@ -547,10 +551,12 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
     }
 
     fn make_cache_with_disconnected_transactions(
-        &self,
+        &'a self,
+        utxo_view: &'a impl UtxosView,
```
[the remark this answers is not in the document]
azarovh: Because a ref to the utxo_view is stored in tx_verifier which is returned from this function by value. So utxo_view should live longer than that.

chainstate/src/detail/transaction_verifier/mod.rs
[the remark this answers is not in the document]
azarovh: The idea was that I return a ref to an entry in the map to avoid copying or additional lookup. I don't see a design issue here, besides that such style with refs is avoided in rust. 
If you believe that I overthink the problem and I should just return by value I can rewrite this.

chainstate/src/detail/ban_score.rs
stanislav-tkach: @TheQuantumPhysicist Makes sense, but we have the "Even though this is an invariant, we consider it a violation to be overly cautious" comment there and and `SerializationInvariantError` has 100 ban score. I'm just trying to understand the logic behind that.

TheQuantumPhysicist: Look at the comments there. We can have a call some time about it if the comments aren't enough. I need to revisit them to remember the details.

The simple version is that some times we act cautiously because there could be bugs we don't anticipate that can create an attack vector. When in doubt, go for 100.

chainstate/src/detail/transaction_verifier/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: The better way is to pop the entries you're disconnecting one by one by value since you're disconnecting them one by one. Logically, if you're gonna use TxUndo entries and put them back into the utxo set, then taking them by value only makes sense. Why take their reference and then use the self reference to do clone inside it? This was my guess of what you were doing.

With a little digging, I noticed that inside the disconnect_transaction function, you're cloning the utxo! The utxo from TxUndo! This is exactly the design issue rust is trying to protect you from. If you're going to do a clone eventually, maybe it makes more sense to pass things by value, or at least do the cloning before using multiple mutable and immutable references together. That way, you don't have complex references all around the place.

chainstate/src/detail/chainstateref.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: In that case, how about we put the UtxosDB inside of transaction verifier, since it's just a thin wrapper around db_tx, which is already in the transaction verifier?

chainstate/src/detail/chainstateref.rs
```diff
@@ -547,10 +551,12 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
         block: &Block,
     ) -> Result<TransactionVerifier<S>, BlockError> {
-        let mut tx_verifier = TransactionVerifier::new(&self.db_tx, self.chain_config);
+        let mut tx_verifier =
+            TransactionVerifier::new(&self.db_tx, utxo_view.derive_cache(), self.chain_config);
         block.transactions().iter().enumerate().try_for_each(|(tx_num, _tx)| {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Can you please add a simple test that tests this and fails if the order is wrong? You can do it after this PR if you wish.

chainstate/src/detail/transaction_verifier/error.rs
```diff
@@ -86,6 +97,8 @@ pub enum ConnectTransactionError {
     SerializationInvariantError(Id<Block>),
     #[error("Timelock rules violated")]
     TimeLockViolation,
+    #[error("Utxo invariant broken")]
+    UtxoInvariantBroken,
```
TheQuantumPhysicist, line 101: Instead of removing the string and keeping it bare, please add the parent error type so that it can be see. Strings are not good enough because they can only be found with inspection. Hiding errors is not good because it can never be found unless some good logging was done. See for example how CheckBlockError works.

chainstate/src/detail/chainstateref.rs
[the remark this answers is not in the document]
azarovh: The problem with that is that I'm not sure it's possible in rust to initialize struct fields where one is a reference to another. And in this case utxo_cache has a reference to utxo_db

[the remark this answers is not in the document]
TheQuantumPhysicist: As long as the lifetime is proper and you're borrowing only immutable references, there won't be an issue.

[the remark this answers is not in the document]
azarovh: I mean I literally don't know how to do that syntactically:
```rust
impl<'a, S: UtxosStorageRead> TransactionVerifier<'a, S> {
    pub fn new(db_tx: &'a S, chain_config: &'a ChainConfig) -> Self {
        Self {
            db_tx,
            chain_config,
            tx_index_cache: BTreeMap::new(),
            utxo_db: UtxosDB::new(db_tx),
            utxo_cache: utxo_db???.derive_cache(), //self is not ready and utxo_db field is not accessable
            utxo_block_undo: BTreeMap::new(),
        }
    }
}
```

something like this doesn't work either:
```rust
impl<'a, S: UtxosStorageRead> TransactionVerifier<'a, S> {
    pub fn new(db_tx: &'a S, chain_config: &'a ChainConfig) -> Self {
        let utxo_db = UtxosDB::new(db_tx);
        let utxo_cache = utxo_db.derive_cache(); //borrows var that is moved
        Self {
            db_tx,
            chain_config,
            tx_index_cache: BTreeMap::new(),
            utxo_db,
            utxo_cache,
            utxo_block_undo: BTreeMap::new(),
        }
    }
}
```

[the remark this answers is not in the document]
azarovh: discussed, leaving a TODO to investigate later

chainstate/src/detail/chainstateref.rs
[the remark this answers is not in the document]
azarovh: tbh I don't see a way how to implement that test.  There are no mocks to track function calls or any side effects that can be checked (like the indexes that are accessed in the db). Added a todo for now

utxo/src/undo.rs
```diff
@@ -46,26 +62,47 @@ impl TxUndo {
...
+
+    pub fn push_tx_undo(&mut self, tx_undo: TxUndo) {
+        self.tx_undos.push(tx_undo);
     }
 
-    pub fn height(&self) -> BlockHeight {
-        self.height
+    pub fn take_tx_undo(&mut self, tx_num: usize) -> Option<TxUndo> {
+        if tx_num < self.tx_undos.len() {
+            Some(self.tx_undos.remove(tx_num))
```
TheQuantumPhysicist, line 89: Ignoring the performance issues here, this can easily become a fire hazard. If you remove a transaction at an index other than the end, it basically means you've modified all the indices that come after it. Now we can look at this in two different ways. We either want to prevent this at the API level and return an error with `Result<Option<TxUndo>, Error>`, or consider this an invariant and should never happen, in which case we panic (aka, assert). For now, either way is fine for me.

TheQuantumPhysicist (approved):
There's only the undo order issue left, I think. I would appreciate others reviewing this as well.

iljakuklic (commented):
Adding some comments

chainstate/src/detail/ban_score.rs
```diff
@@ -99,6 +99,10 @@ impl BanScore for ConnectTransactionError {
             // Even though this is an invariant, we consider it a violation to be overly cautious
             ConnectTransactionError::SerializationInvariantError(_) => 100,
             ConnectTransactionError::TimeLockViolation => 100,
+            &ConnectTransactionError::MissingBlockUndo(_) => 0,
+            &ConnectTransactionError::MissingBlockRewardUndo(_) => 0,
+            ConnectTransactionError::MissingTxUndo(_, _) => 0,
+            ConnectTransactionError::UtxoError(err) => err.ban_score(),
```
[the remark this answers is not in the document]
azarovh: these are typos, my vscode highlight is sometimes misleading. thanks

utxo/src/cache.rs
```diff
@@ -145,6 +152,73 @@ impl<'a> UtxosCache<'a> {
...
+    // Marks outputs of a transaction as spent and inputs as unspent
+    pub fn disconnect_transaction(
+        &mut self,
+        tx: &Transaction,
+        tx_undo: TxUndo,
+    ) -> Result<(), Error> {
+        for (i, _) in tx.outputs().iter().enumerate() {
+            let tx_outpoint = OutPoint::new(OutPointSourceId::from(tx.get_id()), i as u32);
+            self.spend_utxo(&tx_outpoint)?;
+        }
```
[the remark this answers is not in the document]
azarovh: It changes utxo entries state to 'spent' in a local cache. When the cache is flushed to db, 'spent' entries are ignored and just won't get stored. I believe this can be described as spend.

iljakuklic (approved):

PR #379, opened 2022-08-24
storage: New interface for storage backends

iljakuklic:
Introduces a new interface for implementing backends. There is a number of inter-related changes motivated by this:

A new trait-based low-level interface for defining DB backens has been introduced in the `storage-core` crate.

* There is also a bunch of tools for implementing backends, like an adaptor that adds transaction capability to a "dumb" backend.
* The backend and schema specification have been decoupled.

A new high-level application-agnostic interface living in the top-level `storage` crate has been introduced.

* This is because the simple backend interface is too low-level and cumbersome to use.
* On top of the low-level backend interface, it adds schema for extra type checking.
* The interface no longer uses any traits. It's simply a collection of types parametrized by backend being used and the schema. This allows us to remove some trait definitions with some of the most intricate lifetime annotations in the project.

The in-memory backend is reworked in terms of two components.

* A thin wrapper over a bunch of `BTreeMap`s defining simple read-write operations.
* A generic adaptor for adding DB transaction capability is applied to create a full-fledged backend.

The implementation of chainstate-storage has been adjusted.

* The drawback of having the storage interface trait-free is that some parts of the internals of `chainstate-storage` are now implemented in terms of macros.
* While the traits defined in `chainstate-storage` have been simplified a bit, the way the outward-facing API is used is unchanged.


chainstate-storage/src/internal/test.rs
```diff
@@ -156,24 +156,20 @@ fn get_set_transactions() {
         let thr1 = {
             let store = Store::clone(&store);
             common::thread::spawn(move || {
-                let _ = store.transaction_rw().run(|tx| {
-                    let v = tx.get_storage_version()?;
-                    tx.set_storage_version(v + 1)?;
-                    storage::commit(())
-                });
+                let mut tx = store.transaction_rw();
```
TheQuantumPhysicist, line 159: No more scoped transactions?

storage/core/src/backend.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn put(&mut self, idx: DbIndex, key: Data, val: Data) -> crate::Result<()>;
+
+    /// Delete the value associated with given key.
+    fn del(&mut self, idx: DbIndex, key: &[u8]) -> crate::Result<()>;
+}
+
+/// Read-only transaction
+///
+/// If a clenup is required when the transaction closes, [Drop] should be implemented too.
+pub trait TxRo: ReadOps {}
```
TheQuantumPhysicist, line 41: Why isn't there a `close()` or whatever method to end a transaction?

Same for TxRw transaction. There's no `revert()`.

storage/core/src/info.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Description of one key-value store in a database
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub struct MapDesc {
+    /// Key-value map name
+    pub name: &'static str,
+    /// Value size hint
+    pub size_hint: core::ops::Range<usize>,
+    // TODO not implemented, is it even needed?
+    // /// Map multiplicity (i.e. is this a key-value or a key-multi-value map)
+    // pub multiplicity: Multiplicity,
```
TheQuantumPhysicist, line 43: I can't really say whether this is needed right now. I needed it in other projects for tokenization, but we may have found a way around it in Mintlayer, still not sure. Though it's reasonable to have this. Does it make things much more complicated or something?

storage/core/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT"
 [dependencies]
 
 # Local dependencies
+common = { path = '../../common' }
```
TheQuantumPhysicist, line 10: I think this is a timebomb waiting to blow up with some circular dependency in the future... I see you're only importing sync primitives from common. Please, either make a separate crate for that, or just use primitives from std directly. I cannot really assess the risk on the latter option, but if you find it risky and can cause big compiler errors and confuse other devs, please go with the former option.

storage/inmemory/src/lib.rs
```diff
@@ -13,316 +13,61 @@
...
-    Modify { add: Vec<T>, del: Vec<T> },
-}
-
-type DeltaMapMulti = BTreeMap<Data, VecDelta<Data>>;
-
-enum DeltaMap {
-    Single(DeltaMapSingle),
-    Multi(DeltaMapMulti),
-}
+use storage_core::{adaptor, backend, Data, DbDesc, DbIndex, Result as BackendResult};
```
TheQuantumPhysicist, line 16: I don't think `Result as BackendResult` aliasing is necessary.

storage/core/src/adaptor/locking.rs
```diff
@@ -0,0 +1,157 @@
...
+impl<T> Clone for TransactionLockImpl<T> {
+    fn clone(&self) -> Self {
+        Self {
+            db: sync::Arc::clone(&self.db),
+            num_maps: self.num_maps,
+        }
+    }
+}
+
+impl<'tx, T: 'tx + ReadOps> traits::TransactionalRo<'tx> for TransactionLockImpl<T> {
```
TheQuantumPhysicist, line 103: Generally speaking, transactions don't need to be thread-safe, and it'd be great if we could have it that way by default. Usually you start a transaction in one thread, and end it in the same thread. Even in async context, we never go across boundaries. In fact, this is a huge performance boost, since only committing transactions will require locking, while the flood of things to do locally (especially writes) to that transaction object is with no locking.

iljakuklic: Sure. Backends are not required to use this for the transaction capability. E.g. for LMDB backend, I plan to just use the LMDB's transaction implementation, which I suppose is way more optimised than this. We can also implement other adaptors that use a more optimistic synchronisation mechanisms. This just splits the in-memory backend implementation into two more reusable components, it does not really change how it is implemented.

storage/inmemory/src/lib.rs
iljakuklic: Will have a look. I'd like to avoid unqualified `Result` since I find it can be easily mistaken for `std::result::Result` by someone not familiar with the pattern of defining `Result<T, E>` specialized for some fixed `E`.

storage/src/interface.rs
```diff
@@ -0,0 +1,190 @@
...
+pub struct Storage<B: Backend, Sch> {
+    backend: B::Impl,
+    _schema: core::marker::PhantomData<Sch>,
+}
+
+impl<B: Backend, Sch> Clone for Storage<B, Sch>
+where
+    B::Impl: Clone,
+{
+    fn clone(&self) -> Self {
```
TheQuantumPhysicist, line 31: I think a derive would work for this. No need to specialize the clone() implementation.

TheQuantumPhysicist (approved):
Besides the few things I mentioned, looks good.

chainstate-storage/src/internal/test.rs
iljakuklic: I decided to remove them since it was not really used in the code at all, only in tests, and there is little point on testing something that is not used in the software itself. Me removing it is just recognition of this reality.

My preference would be to have only the scoped transaction interface, since then you can provide stronger guarantees the storage module is used correctly. For example, it guarantees the transactions are properly finalized, i.e. `close` is called to terminate `ro` transactions and either `commit` or `abort` are called for `rw` transactions. I think it also encourages more principled approach to error handling.

storage/core/src/backend.rs
iljakuklic: Since there is no guarantee that `close` or similar will be called by the user, I decided it's the job of the `Drop::drop` method to do the cleanup, as suggested in the comment. See the discussion on scoped transactions too, it's somewhat related. This applies to backend. The user-facing interface provides `abort` and `close` but these only consume the object and by dosing so `drop()` is triggered which would do the cleanup. This is also approach taken by some of the `lmdb` crates I was looking into.

storage/src/interface.rs
iljakuklic: I just checked and it does not work because `self` contains a `B::Impl` and not `B`. The auto-derived `Clone` impl stupidly puts in the `B: Clone` constraint whereas `B::Impl: Clone` is needed.

storage/core/Cargo.toml
iljakuklic: The `common::concurrency` module switches what is used between primitives from `std` and from `loom`, depending whether we are compiling for loom or not. Using primitives from `std` directly would mean loom would not do any of the fancy thread interleaving stuff.

Maybe `utils` is a more appropriate place to put the `concurrency` stuff in?

TheQuantumPhysicist: That sounds OK, at least for now. Let's move it to utils.

storage/core/src/backend.rs
TheQuantumPhysicist: There's no guarantee, but the choice should probably be there. Now I can't say this must go in here, since this is the backend of the storage. As long as these methods exist for the storage frontend, I'm OK with that.

The way I do it, usually, is that I make drop call the close/revert functions. That way both possibilities exist.

storage/inmemory/src/lib.rs
TheQuantumPhysicist: It's quite a common pattern in rust to see Result being an alias for something else. Don't anyhow and its users do that quite often?

iljakuklic: > It's quite a common pattern in rust to see Result being an alias for something else. Don't anyhow and its users do that quite often?

They do. However, I like to refer to them using a qualified path, like `anyhow::Result` rather than importing it (shadowing `Result` from `std`) and using it as just `Result`. WIll do something similar here, would that be okay?

TheQuantumPhysicist: Yeah, sure. That's fine.

stanislav-tkach: Personally I prefer to shadow `std::result::Result` only if no other result-types are used in the crate/module. It is usually the case with `anyhow`. Additionally I think it is better for the crate (`storage_core` in this case) to export a result with some more specific name. Otherwise every user of a crate will be forced to introduce an alias himself or use the qualified name everywhere.

storage/core/Cargo.toml
iljakuklic: #380

storage/core/src/backend.rs
iljakuklic: > The way I do it, usually, is that I make drop call the close/revert functions. That way both possibilities exist.

Then close/revert would have to take `&mut self` rather than `self` since drop has the signature of `fn drop(&mut self)`. So I think it's cleaner to just use `drop` for cleanup, which has to happen anyway in case the user does not call `abort()` or similar explicitly and avoid duplication of this functionality in the backend code.

The high-level interface includes the `TransactionRo::close` and `TransactionRw::abort` methods but these are merely convenient ways to call `std::mem::drop` early with a more accurate name.

storage/core/src/info.rs
iljakuklic: To be honest I'd like to see if we can get away without multi-valued maps as it would make a number of things simpler. The backend code would be slightly more complicated with multi-valued maps but that I expect not to be a massive issue. In the high-level interface, I'd like to see a type-level distinction between single-valued and multi-valued maps and this would create a fair amount of ceremony around schema definitions.

If the need for this arises, I we could introduce it and think about how to do it nicely, but for the time of being I'd be in favour of getting rid of it. At the moment, there's some amount of dead code in service of feature that's not fully implemented.

storage/src/interface.rs
```diff
@@ -0,0 +1,190 @@
...
+
+//! High-level application-agnostic storage interface
+
+use crate::schema::{self, Schema};
+use storage_core::{backend, Backend, Data, DbIndex};
+
+/// The main storage type
+pub struct Storage<B: Backend, Sch> {
+    backend: B::Impl,
+    _schema: core::marker::PhantomData<Sch>,
```
stanislav-tkach, line 24: I'm just curious: is an underscore prefix really required for a `PhantomData` field?

stanislav-tkach (approved):

chainstate-storage/src/mock.rs
```diff
@@ -313,14 +309,15 @@ mod tests {
         store: &mut BS,
         block: &Block,
     ) -> &'static str {
-        let res: crate::Result<&'static str> = store.transaction_rw().run(|tx| {
+        (|| {
+            let mut tx = store.transaction_rw();
             // Get current best block ID
             let _best_id = match tx.get_best_block_id()? {
-                None => return storage::abort("top not set"),
+                None => return Ok("top not set"),
```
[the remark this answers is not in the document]
iljakuklic: They are used in tests further down. Was too lazy to introduce a dedicated error type just for testing

storage/core/src/backend.rs
TheQuantumPhysicist: That's fair. I forgot about moving self.

storage/src/interface.rs
iljakuklic: I don't think it's strictly required. It's there to suggest the field is not really used conceptually.

storage/core/src/info.rs
TheQuantumPhysicist: OK. Let's get rid of it for now. I can't promise we won't need it in the future though.

muursh (approved):
Realistically I've only skimmed this pretty quickly but it looks fine based on that

PR #381, opened 2022-08-25
common: Mark `Amount` and `SignedAmount` as `must_use`

iljakuklic:
to prevent the results from being ignored accidentally

### DIFF
--- a/chainstate/src/detail/transaction_verifier/mod.rs
+++ b/chainstate/src/detail/transaction_verifier/mod.rs
@@ -533,7 +533,7 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
                 self.precache_inputs(tx.inputs())?;
 
                 // check for attempted money printing
-                self.check_transferred_amounts_and_get_fee(tx)?;
+                let _ = self.check_transferred_amounts_and_get_fee(tx)?;
 
                 // verify input signatures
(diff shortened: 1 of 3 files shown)


muursh (approved):

TheQuantumPhysicist (approved):

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -533,7 +533,7 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
                 self.precache_inputs(tx.inputs())?;
 
                 // check for attempted money printing
-                self.check_transferred_amounts_and_get_fee(tx)?;
+                let _ = self.check_transferred_amounts_and_get_fee(tx)?;
```
azarovh, line 536: Should check return () instead of Amount that is ignored anyway?

azarovh (approved):

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: This is being used in two different contexts (IIRC), one of which requires a return value and the other just does the check.

azarovh: oh it returns fees, then ignore this

stanislav-tkach (approved):

PR #383, opened 2022-08-26
Normal tokens (second attempt)

sinitcin:
This PR is entailing of [the previous one](https://github.com/mintlayer/mintlayer-core/pull/269). 

Opened issues:
* https://github.com/mintlayer/mintlayer-core/issues/384
* https://github.com/mintlayer/mintlayer-core/issues/385
* https://github.com/mintlayer/mintlayer-core/issues/401


chainstate/src/detail/tests/test_framework/framework.rs
```diff
@@ -28,7 +28,7 @@ use crate::{
     detail::{
         tests::{
             test_framework::{BlockBuilder, TestFrameworkBuilder, TransactionBuilder},
-            TestBlockInfo,
+            TestBlockInfo, *,
```
TheQuantumPhysicist, line 31: This doesn't make sense. This is like saying "waiter, please get me the chicken and caeser salad, and get me the whole menu". Well, if you're ordering the whole menu, you don't need to specifically order the chicken... it's part of the menu.

common/src/chain/tokens/errors.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::{
+    chain::{Block, Transaction},
+    primitives::Id,
+};
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
+pub enum TokensError {
+    #[error("Incorrect ticker in issue transaction {0} in block {1}")]
+    IssueErrorIncorrectTicker(Id<Transaction>, Id<Block>),
```
TheQuantumPhysicist, line 26: I would call this Invalid, not Incorrect.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,843 @@
...
+        })];
+        assert!(matches!(
+            process_token(&mut test_framework, ParentBlock::BestBlock, values),
+            Err(BlockError::CheckBlockFailed(
+                CheckBlockError::CheckTransactionFailed(CheckBlockTransactionsError::TokensError(
+                    TokensError::IssueErrorIncorrectMetadataURI(_, _)
+                ))
+            ))
+        ));
+    });
```
TheQuantumPhysicist, line 243: Always when you test something that doesn't work, you end it with something that works so that you have a reference. Here, you tested multiple cases that don't work. You should finally test passing the working case with everything being valid.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,843 @@
...
+                CheckBlockError::CheckTransactionFailed(CheckBlockTransactionsError::TokensError(
+                    TokensError::TransferZeroTokens(_, _)
+                ))
+            ))
+        ));
+    })
+}
+
+#[test]
+fn couple_of_token_issuance_in_one_tx() {
```
TheQuantumPhysicist, line 315: multiple_token_issuance_in_one_tx

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,843 @@
...
+        })];
+        assert!(matches!(
+            process_token(&mut test_framework, ParentBlock::BestBlock, values),
+            Err(BlockError::CheckBlockFailed(
+                CheckBlockError::CheckTransactionFailed(CheckBlockTransactionsError::TokensError(
+                    TokensError::TransferZeroTokens(_, _)
+                ))
+            ))
+        ));
+    })
```
TheQuantumPhysicist, line 311: Same thing here. For all the tests in this file, always end with a working reference case.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,843 @@
...
+            process_token(&mut test_framework, ParentBlock::BestBlock, values),
+            Err(BlockError::StateUpdateFailed(
+                ConnectTransactionError::AttemptToPrintMoney(_, _)
+            ))
+        ));
+    });
+}
+
+#[test]
+fn test_attempt_to_mix_input_tokens() {
```
TheQuantumPhysicist, line 755: What does this test exactly? Can you elaborate, plase?

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    amount_to_issue: &Amount,
+    number_of_decimals: &u8,
+    metadata_uri: &Vec<u8>,
+    tx_id: Id<Transaction>,
+    source_block_id: Id<Block>,
+) -> Result<(), TokensError> {
+    // Check token name
+    if token_ticker.len() > TOKEN_MAX_TICKER_LEN
+        || token_ticker.is_empty()
+        || !String::from_utf8_lossy(token_ticker).is_ascii()
```
TheQuantumPhysicist, line 64: Never, ever, use lossy conversions in consensus. This is a recipe for disaster. Basically people can post things in the token ticker that are not detectable by our software.

The right way to do this check if with `token_ticker.chars().all(char::is_alphanumeric)`.

TheQuantumPhysicist, line 64: Never, ever, use lossy conversions in consensus. This is a recipe for disaster. Basically people can post things in the token ticker that are not detectable by our software.

The right way to do this check if with `from_utf8()`, then do error handling.

Also double-check that the result has only alphanumeric chars with `result.chars().all(char::is_alphanumeric)`.

TheQuantumPhysicist: Also add tests that ensure that token tickers can't have symbols after implementing this.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    // Check decimals
+    if number_of_decimals > &TOKEN_MAX_DEC_COUNT {
+        return Err(TokensError::IssueErrorTooManyDecimals(
+            tx_id,
+            source_block_id,
+        ));
+    }
+
+    // Check URI
+    if metadata_uri.len() > TOKEN_MAX_URI_LEN || !String::from_utf8_lossy(metadata_uri).is_ascii() {
```
TheQuantumPhysicist, line 89: Same here about `from_utf8_lossy`. Never use lossy conversions!

chainstate/src/detail/transaction_verifier/utils.rs
```diff
@@ -0,0 +1,126 @@
...
+    chain::{
+        tokens::{token_id, CoinOrTokenId, OutputValue, TokenData, TokensError},
+        Transaction,
+    },
+    primitives::Amount,
+};
+
+use super::error::ConnectTransactionError;
+
+pub fn insert_or_increase(
```
TheQuantumPhysicist, line 28: I think this can be simplified with `total_amounts.entry(key).unwrap_or(Amount::from_atoms)`. I'm not sure abut the whole picture though.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -598,6 +678,9 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
                         .unspend(outpoint.output_index())
                         .map_err(ConnectTransactionError::from)?;
                 }
+
+                // Remove issued tokens
+                self.remove_tokens(tx)?;
```
TheQuantumPhysicist, line 683: Is this supposed to be `unregister_token_issuance()`?

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -533,7 +598,10 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
                 self.precache_inputs(tx.inputs())?;
 
                 // check for attempted money printing
-                let _ = self.check_transferred_amounts_and_get_fee(tx)?;
+                let _ = self.check_transferred_amounts_and_get_fee(block.get_id(), tx)?;
+
+                // Register tokens if tx has issuance data
+                self.register_tokens(tx)?;
```
TheQuantumPhysicist, line 604: `register_tokens_issuance()`

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +531,29 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
         Ok(())
     }
 
+    fn register_tokens(&mut self, tx: &Transaction) -> Result<(), TokensError> {
+        let token_id = token_id(tx).ok_or(TokensError::TokenIdCantBeCalculated)?;
+
+        let _tokens_op = match self.tokens_cache.entry(token_id) {
+            Entry::Occupied(entry) => entry.into_mut(), // TODO: Tokens already issued? Must we return error?
```
TheQuantumPhysicist, line 538: This should return an error with an invariant being broken as this should never happen.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -571,6 +639,18 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
         Ok(())
     }
 
+    fn remove_tokens(&mut self, tx: &Transaction) -> Result<(), ConnectTransactionError> {
+        let was_tokens_issued =
```
TheQuantumPhysicist, line 643: Again, this is a consistency issue. It's IMPOSSIBLE to disconnect a block with token issuance without finding a token issued. Please don't hide such issues and handle them properly.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -256,66 +282,71 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                        }
+                    }
                 }
             }
         }
-        Ok(total)
+        Ok(result)
+    }
+
+    fn get_paid_fee_unchecked(
```
TheQuantumPhysicist, line 309: What is it exactly that's "unchecked" here?

TheQuantumPhysicist: Also, since you're excluding tokens here, please throw in some "coin" in the name of the function.

TheQuantumPhysicist (commented):
Please add checks that no token issuance is done in block reward branch of connect_transaction.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+) -> Result<(), TokensError> {
+    match token_data {
+        TokenData::TokenTransferV1 {
+            token_id: _,
+            amount,
+        } => {
+            check_tokens_transfer_data(source_block_id, tx, amount)?;
+        }
+        TokenData::TokenIssuanceV1 {
+            token_ticker,
```
TheQuantumPhysicist, line 111: Please also add a public key, basically the controller public key. The controller will have rights to issue more tokens later. It can be an `Option<PublicKey>` because issuers may decide to not allow that.

TheQuantumPhysicist: You know what, don't do that now. Let's postpone this to next PR. Create an issue for it.

In that issue, please reference this comment. I'll add more information in another comment.

TheQuantumPhysicist: Keep in mind that the total supply should be taken from another place if this is enabled. For example, some storage in database. This should also support reorgs and such complex operations with newly minted coins after initial issuance being subtracted when disconnecting a block.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +531,29 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+    }
+
+    fn undo_issuance(&mut self, token_id: TokenId) -> Result<(), TokensError> {
+        match self.tokens_cache.entry(token_id) {
+            Entry::Occupied(entry) => {
+                let tokens_op = entry.into_mut();
+                *tokens_op = CachedTokensOperation::Erase;
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(CachedTokensOperation::Erase);
```
TheQuantumPhysicist, line 551: Same, consistency issue.

chainstate/src/detail/tests/test_framework/framework.rs
sinitcin: It's understandable. This does not remove responsibility from myself, but weird why Clippy don't mark it as an issue. 

Anyway, I itemized imports and removed `*`.

common/src/chain/tokens/errors.rs
sinitcin: Ok. Renamed.

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: I'll move this comment to where appropriate.

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist, line 551: Again, this is a consistency issue. It's IMPOSSIBLE to disconnect a block with token issuance without finding a token issued. Please don't hide such issues and handle them properly.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: I have a valid case at the start of the test. However, I fixed the comment to highlight valid case.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Ok, renamed.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: So, I have here a valid case too at the start of the test, but seems it's a real problem. I will highlight through comments all valid cases in all tests. And I'm going to move them at the end of the tests where it is possible.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: In a few words, we issue tokens A and tokens B, and try to spend tokens A with values A+B. If tokens values calculate correctly in connect block, it will fail with `AttemtToPrintMoney`.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Cool, I've just renamed it.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: And this renamed too

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,867 @@
...
+}
+
+fn process_token_ex(
+    test_framework: &mut TestFramework,
+    additional_inputs: Vec<TxInput>,
+    parent_block: ParentBlock,
+    values: Vec<OutputValue>,
+) -> Result<(Block, Option<BlockIndex>), BlockError> {
+    let receiver = anyonecanspend_address();
+    let parent_block_id = match parent_block {
```
TheQuantumPhysicist, line 80: Why can't you just pass the best block from the test_framework? Isn't it just a function call there? Why the enum? This is just more complexity with no real benefit.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,867 @@
...
+            TokenData::TokenBurnV1 {
+                token_id: _,
+                amount_to_burn: _,
+            } => true,
+        },
+    }
+}
+
+fn process_token_ex(
+    test_framework: &mut TestFramework,
```
TheQuantumPhysicist, line 74: Since you're using the test framework, please use the builder pattern available in it.

chainstate/src/detail/transaction_verifier/utils.rs
```diff
@@ -0,0 +1,126 @@
...
+            }
+            TokenData::TokenBurnV1 {
+                token_id,
+                amount_to_burn,
+            } => Some((CoinOrTokenId::TokenId(*token_id), amount_to_burn)),
+        },
+    }
+}
+
+pub fn filter_for_total_inputs(
```
TheQuantumPhysicist, line 94: This seems like a misnomer. There's no filtering here.

chainstate/src/detail/transaction_verifier/utils.rs
```diff
@@ -0,0 +1,126 @@
...
+            return Err(ConnectTransactionError::AttemptToPrintMoney(
+                *inputs_total,
+                *outputs_total,
+            ));
+        }
+    }
+    Ok(())
+}
+
+pub fn filter_for_total_outputs(output_value: &OutputValue) -> Option<(CoinOrTokenId, &Amount)> {
```
TheQuantumPhysicist, line 67: Same misnomer, there's no filtering, from what I'm seeing here. There's no total, either.

chainstate/src/detail/transaction_verifier/utils.rs
```diff
@@ -0,0 +1,126 @@
...
+            TokenData::TokenTransferV1 { token_id, amount } => {
+                Some((CoinOrTokenId::TokenId(*token_id), amount))
+            }
+            TokenData::TokenIssuanceV1 {
+                token_ticker: _,
+                amount_to_issue: _,
+                number_of_decimals: _,
+                metadata_uri: _,
+            } => {
+                // TODO: Might be it's not necessary at all?
```
TheQuantumPhysicist, line 80: Didn't we agree to have a boolean to enable accounting for issuance?

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -256,66 +282,71 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
     ) -> Result<Amount, ConnectTransactionError> {
-        let inputs = tx.inputs();
-        let outputs = tx.outputs();
+        let inputs_total_map = self.calculate_total_inputs(tx.inputs())?;
+        let outputs_total_map = Self::calculate_total_outputs(tx.outputs())?;
 
-        let inputs_total = self.calculate_coins_total_inputs(inputs)?;
-        let outputs_total = Self::calculate_coins_total_outputs(outputs)?;
+        let total_fee = check_transferred_amount(&inputs_total_map, &outputs_total_map)
+            .and_then(|_| Self::get_paid_fee_unchecked(&inputs_total_map, &outputs_total_map))?;
```
TheQuantumPhysicist, line 332: I don't understand. Why not use a question mark on the check_transferred_amount and then call this independently? Why bind them together like this?

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -571,6 +639,21 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
         Ok(())
     }
 
+    fn unregister_token_issuance(
```
TheQuantumPhysicist, line 642: Please put all register/unregister functions together so that they can be compared (including undo, and whatever else related). Putting them in utils isn't the worst idea, but I don't think that's easily achived because they have to do with the cache. So please just put them above one another somewhere here.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -489,6 +531,29 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
         Ok(())
     }
 
+    fn register_tokens_issuance(&mut self, tx: &Transaction) -> Result<(), TokensError> {
```
TheQuantumPhysicist, line 534: Please add a comment "token registration saves the token id in the database with the transaction that issued it, and possible some additional auxiliary data; this helps in finding the relevant information of the token at any time in the future".

chainstate/src/detail/ban_score.rs
```diff
@@ -99,6 +99,7 @@ impl BanScore for ConnectTransactionError {
             // Even though this is an invariant, we consider it a violation to be overly cautious
             ConnectTransactionError::SerializationInvariantError(_) => 100,
             ConnectTransactionError::TimeLockViolation => 100,
+            ConnectTransactionError::TokensError(tokens_error) => tokens_error.ban_score(),
```
TheQuantumPhysicist, line 102: For brevity, please change `tokens_error` to `err`, just like everything else.

chainstate/src/detail/ban_score.rs
```diff
@@ -120,13 +121,36 @@ impl BanScore for CheckBlockError {
...
+impl BanScore for TokensError {
+    fn ban_score(&self) -> u32 {
+        match self {
+            TokensError::IssueErrorInvalidTicker(_, _) => 100,
+            TokensError::IssueErrorIncorrectAmount(_, _) => 100,
+            TokensError::IssueErrorTooManyDecimals(_, _) => 100,
+            TokensError::IssueErrorIncorrectMetadataURI(_, _) => 100,
+            TokensError::MultipleTokenIssuanceInTransaction(_, _) => 100,
+            TokensError::CoinOrTokenOverflow => 100,
+            TokensError::InsufficientTokenFees(_, _) => 0,
```
TheQuantumPhysicist, line 133: This is a violation. Why is it not punishable? If someone creates a transaction with insufficient fee and include it in a block, they're basically spamming us with invalid blocks.

The only case when this is zero, is when there's an internal error and we're 100% it's not the fault of the peer. But then also some other times it's difficult to make that judgment because we don't know if it can be exploited. 

Meaning: You ban when there's a deliberate violation, e.g., invalid data, transaction or fees etc. You don't ban when it's our problem.

chainstate/src/detail/chainstateref.rs
```diff
@@ -305,6 +305,13 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         self.get_gen_block_index(&self.get_best_block_id()?)
     }
 
+    pub fn get_token_tx(
```
TheQuantumPhysicist, line 308: `get_token_info`, and return that object we talked about in storage.

common/src/chain/config/mod.rs
```diff
@@ -37,6 +36,11 @@ use std::time::Duration;
 
 const DEFAULT_MAX_FUTURE_BLOCK_TIME_OFFSET: Duration = Duration::from_secs(60 * 60);
 pub const DEFAULT_TARGET_BLOCK_SPACING: Duration = Duration::from_secs(120);
+pub const TOKEN_MIN_ISSUANCE_FEE: Amount = Amount::from_atoms(10_000_000_000_000);
+pub const TOKEN_MAX_URI_LEN: usize = 1024;
+pub const TOKEN_MAX_DEC_COUNT: u8 = 18;
+pub const TOKEN_MAX_TICKER_LEN: usize = 5;
+pub const TOKEN_MAX_ISSUANCE_ALLOWED: usize = 1;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: And to be clear, we mean in the ChainConfig struct, not just here as global objects.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,879 @@
...
+    })
+}
+
+#[test]
+fn snapshot_testing_tokens_data() {
+    utils::concurrency::model(|| {
+        // TODO: Add tests, that will prevent change fields order
+    })
+}
+
```
TheQuantumPhysicist, line 878: Add a test too that reorgs over a token issuance and see if all the data of that token will disappear from the mainchain.

chainstate/src/detail/tokens/mod.rs
sinitcin: I've got you. I fixed it. Also, `token_issue_test` has the check that tickers can't have symbols.

sinitcin: I will push all updates later, because my credential expired on github

chainstate/src/detail/tokens/mod.rs
sinitcin: Fixed!

chainstate/src/detail/transaction_verifier/utils.rs
sinitcin: If it can return Err from closure, we might be use something like this: 
```rust
    total_amounts
        .entry(key)
        .and_modify(|value| {
            (*value + amount).ok_or(ConnectTransactionError::TokensError(
                TokensError::CoinOrTokenOverflow,
            ))?
        })
        .or_insert(amount);
```
Unfortunately, I haven't found a way to simplify it. If anyone can advise something, it will be great.

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: I think, we can resolve this because now I use the builder pattern there

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,867 @@
...
+            } => false,
+            TokenData::TokenBurnV1 {
+                token_id: _,
+                amount_to_burn: _,
+            } => true,
+        },
+    }
+}
+
+fn process_token_ex(
```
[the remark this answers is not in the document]
sinitcin: Since I started use the build pattern, I removed this function

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Done

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Fixed. If we can't find tokens here, we return `TokensError::InvariantBrokenUndoIssuanceOnNonexistentToken`

chainstate/src/detail/transaction_verifier/utils.rs
sinitcin: Yes, we agreed that before, but is it really need for us? So far, we don't have any code that can use issuance amount. Moreover, we have to add token_id for this arm. I feel like it mightn't be necessary.

chainstate/src/detail/transaction_verifier/utils.rs
sinitcin: Actually, we take coin or token value from input. I propose `get_input_value`, but not entirely sure.

chainstate/src/detail/transaction_verifier/utils.rs
sinitcin: Same here `get_output_value`

chainstate/src/detail/transaction_verifier/utils.rs
sinitcin: Or might be better to use `get_input_coin_or_tokenid`

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: That was my extra paranoia, I have a look again here and yes we can use just 

```Rust
        check_transferred_amount(&inputs_total_map, &outputs_total_map)?;
        let total_fee = Self::get_paid_fee_unchecked(&inputs_total_map, &outputs_total_map)?;
```

chainstate/src/detail/ban_score.rs
sinitcin: Done

chainstate/src/detail/transaction_verifier/utils.rs
TheQuantumPhysicist: We have to add it because the function name and structure doesn't say it excludes issuance. This function basically returns incorrect information.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: As we discussed before, functions mustn't include unclear words, or words that can confuse. Is it ok to rename it to  `get_paid_fee_coins`

chainstate/src/detail/tokens/mod.rs
sinitcin: Done

chainstate/src/detail/transaction_verifier/utils.rs
TheQuantumPhysicist: `get_token_id_and_amount`

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -256,66 +282,71 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+        let total_fee = check_transferred_amount(&inputs_total_map, &outputs_total_map)
+            .and_then(|_| Self::get_paid_fee_unchecked(&inputs_total_map, &outputs_total_map))?;
 
-        if outputs_total > inputs_total {
-            return Err(ConnectTransactionError::AttemptToPrintMoney(
-                inputs_total,
-                outputs_total,
+        // Check is fee enough for issuance
+        let issuance_count = get_tokens_issuance_count(tx.outputs());
+        if issuance_count == 1 && total_fee < TOKEN_MIN_ISSUANCE_FEE {
```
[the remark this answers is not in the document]
sinitcin: Updated

chainstate/src/detail/transaction_verifier/utils.rs
TheQuantumPhysicist: `get_output_token_id_and_amount`

chainstate/src/detail/transaction_verifier/utils.rs
TheQuantumPhysicist: `get_input_token_id_and_amount`

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: If you agreed with parameter `tokens_cache: &mut BTreeMap<TokenId, CachedTokensOperation>,`, then we can move all these functions into utils

TheQuantumPhysicist: Go for it. Preferably put them all in the tokens submodule, not the generic utils.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Done

chainstate/src/detail/ban_score.rs
sinitcin: That makes sense. Fixed!

chainstate/src/detail/transaction_verifier/utils.rs
sinitcin: Renamed

sinitcin: Thanks

chainstate/src/detail/transaction_verifier/mod.rs
TheQuantumPhysicist: `get_total_fee()`

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Done

chainstate-storage/src/lib.rs
```diff
@@ -104,6 +108,12 @@ pub trait BlockchainStorageWrite: BlockchainStorageRead + UtxosStorageWrite {
 
     /// Remove block id from given mainchain height
     fn del_block_id_at_height(&mut self, height: &BlockHeight) -> crate::Result<()>;
+
+    /// Set token creation tx
+    fn set_token_tx(&mut self, token_id: TokenId, tx_id: Id<Transaction>) -> crate::Result<()>;
```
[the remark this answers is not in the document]
sinitcin: I added the struct `TokenIssuanceTransaction` with one field `tx`. And now token's functions in storage use this type instead of Id.

chainstate/src/detail/chainstateref.rs
sinitcin: Done

common/src/chain/config/mod.rs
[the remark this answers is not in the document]
sinitcin: Done

chainstate/src/detail/transaction_verifier/utils.rs
sinitcin: As we spoke in the chat, we'll use the parameter `include_issuance: Option<&Transaction>,`. Fixed!

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: I added test `test_tokens_reorgs_and_cleanup_data`

chainstate/src/detail/transaction_verifier/error.rs
```diff
@@ -97,6 +98,14 @@ pub enum ConnectTransactionError {
     TimeLockViolation,
     #[error("Utxo error: {0}")]
     UtxoError(#[from] utxo::Error),
+    #[error("Tokens error: {0}")]
+    TokensError(TokensError),
+}
+
+impl From<TokensError> for ConnectTransactionError {
```
TheQuantumPhysicist, line 105: Instead of doing this, just use `#[from]` like the other variants do.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -87,11 +100,12 @@ impl<'a, S> TransactionVerifier<'a, S> {
             tx_index_cache: BTreeMap::new(),
             utxo_cache,
             utxo_block_undo: BTreeMap::new(),
+            tokens_cache: BTreeMap::new(),
         }
     }
 }
 
-impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
+impl<'a, S: BlockchainStorageRead + 'a> TransactionVerifier<'a, S> {
```
TheQuantumPhysicist, line 108: Why?

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -201,60 +215,98 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
                     .ok_or(ConnectTransactionError::MissingOutputOrSpent)?;
                 entry.insert(CachedInputsOperation::Read(tx_index));
             }
-        }
+        };
```
TheQuantumPhysicist, line 218: Why?

sinitcin: Fixed. I had faced semicolon addings a few times, probably some glitch in cargo-fmt.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -201,60 +215,98 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
-                .ok_or(ConnectTransactionError::MissingOutputOrSpent)
-                .map(|utxo| match utxo.output().value() {
-                    OutputValue::Coin(amount) => *amount,
-                })?;
-            (total + amount).ok_or(ConnectTransactionError::OutputAdditionError)
-        })
+                .ok_or(ConnectTransactionError::MissingOutputOrSpent)?;
+
+            match input.outpoint().tx_id() {
+                OutPointSourceId::Transaction(issuance_tx_id) => {
```
TheQuantumPhysicist, line 253: So every OutPointSourceId's parameter is an issuance_tx_id? If I understand correctly, this may or may not be an issuance transaction, we don't know yet. It's just an outpoint tx id.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Due to last merges, and code simplifying, it's not necessary again now. Fixed.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -205,56 +219,94 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                    let token_id = self.db_tx.get_token_id(&issuance_tx_id)?;
```
TheQuantumPhysicist, line 254: So then we assume it's an issuance anyway, and try to get a token id? Why not detect first if this is issuance and then attempt to access the persistent storage?

chainstate/src/detail/transaction_verifier/utils.rs
```diff
@@ -0,0 +1,126 @@
...
+                token_id,
+                amount_to_burn,
+            } => Some((CoinOrTokenId::TokenId(*token_id), amount_to_burn)),
+        },
+    })
+}
+
+pub fn get_input_token_id_and_amount(
+    output_value: &OutputValue,
+    token_id: Option<TokenId>,
```
TheQuantumPhysicist, line 97: What does it mean when this is an Option? If we're asserting in the function name that we're dealing with a token, why is the token_id an Option?

In the caller, the `Option<TokenId>` is None if you don't find a relevant issuance transaction, but here in the implenetation you ignore the parameter when it's transfer. This is way too convoluted. Maybe I'm missing something.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,1558 @@
...
+        ));
+    })
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn test_tokens_reorgs_and_cleanup_data(#[case] seed: Seed) {
+    utils::concurrency::model(move || {
+        const ISSUED_FUNDS: Amount = Amount::from_atoms(1_000_000);
```
TheQuantumPhysicist, line 1396: Why can't this be a random number in a specified range?

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,1558 @@
...
+        // Check tokens available in storage
+        let (issuance_block_id, issuance_tx) = tf.chainstate.get_token_detail(token_id).unwrap();
+        assert!(issuance_block.get_id() == issuance_block_id);
+        assert!(issuance_block.transactions()[0].get_id() == issuance_tx.get_id());
+
+        // Cause reorg
+        tf.create_chain(&tf.genesis().get_id().into(), 5, &mut rng).unwrap();
+
+        // Check that tokens not in storage
+        tf.chainstate.get_token_detail(token_id).unwrap_err();
```
TheQuantumPhysicist, line 1438: Where's the check for the issuance transaction?

TheQuantumPhysicist: Also, please always test for the particular error. Don't just accept a generic error.

TheQuantumPhysicist: Where's the check of the issuance transaction??

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: As we spoke in the chat, I added random variables where it is possible

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Added more comments, and update the code. I hope now is better.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Sam, you fixed this

chainstate/src/detail/tests/normal_tokens.rs
sinitcin: Added check that reorg happens, and check that tx_index no longer exists in mainchain

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -205,56 +221,105 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+
+        check_transferred_amount(&inputs_total_map, &outputs_total_map)?;
+        let total_fee = Self::get_total_fee(&inputs_total_map, &outputs_total_map)?;
+
+        // TODO: the fee has the issue that it doesn't deduct the issuance fee from the total fee,
+        // which means that anyone issuing tokens will have a free-of-charge priority and a possibly
+        // huge transaction compared to what they would get for without the issuance.
+        // This has to be studied
+
+        // Check is fee enough for issuance
```
azarovh, line 315: Check if the fee is enough for issuance

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    // Check amount
+    if amount_to_issue == &Amount::from_atoms(0) {
+        return Err(TokensError::IssueErrorIncorrectAmount(
+            tx_id,
+            source_block_id,
+        ));
+    }
+
+    // Check decimals
+    if number_of_decimals > &chain_config.token_max_dec_count() {
```
azarovh, line 89: I'm confused about the purpose of `number_of_decimals`, could you please explain why do we need it?

common/src/chain/config/mod.rs
```diff
@@ -153,6 +157,26 @@ impl ChainConfig {
...
+    pub fn token_max_uri_len(&self) -> usize {
+        self.token_max_uri_len
+    }
+
+    pub fn token_max_dec_count(&self) -> u8 {
+        self.token_max_dec_count
+    }
+
+    pub fn token_max_ticker_len(&self) -> usize {
+        self.token_nax_ticker_len
```
muursh, line 173: nax...

muursh: I've fixed this and a load of other English issues. I'll push it when I'm done

sinitcin: My bad, another misprint

chainstate/src/detail/tokens/mod.rs
sinitcin: It's quite common practice in blockchain for avoiding use float numbers and all problems that might appear in transactions. We have integer number of tokens. And special number of decimals. Wallets and DApps can show in UI float numbers, for example $123.45. But actually, we use the integer number 12345 in transactions.

chainstate/src/detail/transaction_verifier/mod.rs
sinitcin: Thanks

chainstate-storage/src/lib.rs
```diff
@@ -66,6 +67,12 @@ pub trait BlockchainStorageRead: UtxosStorageRead {
 
     /// Get mainchain block by its height
     fn get_block_id_by_height(&self, height: &BlockHeight) -> crate::Result<Option<Id<GenBlock>>>;
+
+    /// Get token creation tx
+    fn get_token_tx(&self, token_id: &TokenId) -> crate::Result<Option<TokenIssuanceTransaction>>;
+
+    // Get token id by id of the creation tx
+    fn get_token_id(&self, tx_id: &Id<Transaction>) -> crate::Result<Option<TokenId>>;
```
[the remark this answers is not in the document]
sinitcin: We can't due to token id. Token ID essentially a hash of first input of issuance tx, and obviously we have one token_id per tx.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,1615 @@
...
+
+        let block = tf.block(*block_index.block_id());
+        assert_eq!(block.transactions()[0].outputs()[0].value(), &output_value);
+    });
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn token_transfer_tes(#[case] seed: Seed) {
```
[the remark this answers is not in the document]
sinitcin: Fixed

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,1650 @@
...
+                            number_of_decimals: rng.gen_range(1..18),
+                            metadata_uri: random_string(&mut rng, 1..1024).as_bytes().to_vec(),
+                        }),
+                        OutputPurpose::Transfer(Destination::AnyoneCanSpend),
+                    ))
+                    .build(),
+            )
+            .build_and_process();
+
+        assert!(matches!(
```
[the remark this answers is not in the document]
sinitcin: I have doubts. `assert_eq!`  require use real parameters instead of `_`. These parameters might be random or easily changed due to changes in a test.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,1650 @@
...
+    },
+    primitives::{Amount, Idable},
+};
+use crypto::{hash::StreamHasher, random::Rng};
+use expect_test::expect;
+use rand::distributions::uniform::SampleRange;
+use rstest::rstest;
+use test_utils::random::{make_seedable_rng, Seed};
+
+fn random_string<RNG: Rng, R: SampleRange<usize>>(rng: &mut RNG, range_len: R) -> String {
```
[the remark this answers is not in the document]
sinitcin: Yes, that looks better. I've fixed.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+            tx_id,
+            source_block_id,
+        ));
+    }
+
+    // Check URI
+    let is_ascii = String::from_utf8(metadata_uri.to_vec())
+        .map_err(|_| TokensError::IssueErrorIncorrectMetadataURI(tx_id, source_block_id))?
+        .chars()
+        // TODO: this is probably an invalid way to validate URLs. Find the proper way to do this in rust.
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I'm aware of this crate. But this needs more research than just using an arbitrary crate.

chainstate/src/detail/tests/normal_tokens.rs
```diff
@@ -0,0 +1,1650 @@
...
+        return String::new();
+    }
+    let len = rng.gen_range(range_len);
+    Alphanumeric.sample_string(rng, len)
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn token_issue_test(#[case] seed: Seed) {
```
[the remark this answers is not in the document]
sinitcin: @TheQuantumPhysicist what do you think? Some tests required creating a chain, and seems can't be simplified.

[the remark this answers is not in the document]
TheQuantumPhysicist: On one hand, the format of the tests were chaotic when we started the project because the test framework wasn't complete. On the other hand, keep in mind that many tests don't really share the exact same fixture. It's a game of whac-A-mole to hit them all and not make an extended mess with another layer of test-framework that just parametrizes everything in the fixture and just passes it to the test framework.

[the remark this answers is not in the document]
azarovh: Test fixture shouldn't have more code that it's already written for test initialisation.
My point is that it's better to try to write a lot of small unit test than few big.

chainstate/src/detail/mod.rs
```diff
@@ -307,6 +317,95 @@ impl<S: BlockchainStorage> Chainstate<S> {
         Ok(block)
     }
 
+    pub fn get_token_detail(
```
TheQuantumPhysicist, line 320: get_token_detail... what does detail mean? I have no freaking idea. Just another english word that could mean "stuff". A function that says detail, and then returns an issuance transaction and a block id in a tuple/pair... and to make this even better, the issuance transaction is in TokenIssuanceTransaction struct that has nothing but the issuance transaction (even though I requested token information to contain the full thing), yet the block id is returned as a pair with that struct.

Because function names mean nothing. Function interfaces are just tools to make the code compile. And usually when we want to return data, we just stack it in a tuple instead of making a meaningful interfaces.

To top all this up, we have all these getter functions here completely skipping the query interface.

I seriously feel like I'm gonna get a stroke. My head is numb. I'm really tired of reviewing this PR. Every other day I find more bad practices, negligence and arbitrary names that mean nothing to their function or interface. I'm out of options. This PR has been stringing on for over two months now. I'm trying my best now to fix this myself because I'm tired of the ping-pong we're playing to get this PR merged, which is only resulting in more time wasted.

chainstate/src/detail/mod.rs
```diff
@@ -307,6 +317,95 @@ impl<S: BlockchainStorage> Chainstate<S> {
...
+
+        // Find issuance transaction id by token_id
+        let creation_tx =
+            chainstate_ref
+                .get_token_info(&token_id)?
+                .ok_or(PropertyQueryError::TokensError(
+                    TokensError::TokensNotRegistered(token_id),
+                ))?;
+        let tx_index = chainstate_ref
+            .get_mainchain_tx_index(&OutPointSourceId::Transaction(creation_tx.get_id()))?
```
TheQuantumPhysicist, line 334: Even though we talked about this and we mentioned that not every user will necessarily have the transaction index, and that's why we added a database entry for issuance tx id vs token id, yet we have this still sitting here.

Don't worry. I'll fix it for you. That's what I'm here for.

common/src/chain/config/builder.rs
```diff
@@ -195,7 +195,7 @@ impl Builder {
             token_min_issuance_fee,
             token_max_uri_len,
             token_max_dec_count,
-            token_nax_ticker_len,
+            token_nax_ticker_len: token_max_ticker_len,
```
muursh, line 198: Missed one.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,147 @@
...
+    // Check amount
+    ensure!(
+        amount > &Amount::from_atoms(0),
+        TokensError::TransferZeroTokens(tx.get_id(), source_block_id)
+    );
+
+    Ok(())
+}
+
+pub fn check_tokens_burn_data(
```
muursh, line 39: What does `data` have to do with this?

TheQuantumPhysicist: Don't ask for explanation... just make suggestions (and I genuinely mean that). That's the shortest path. Let's try to get this done with.

sinitcin: We literally have the type `TokenData`, that essentially enum with burn, transfer, and issuance data for output.

common/src/chain/config/mod.rs
```diff
@@ -163,6 +187,11 @@ impl ChainConfig {
 const MAX_BLOCK_HEADER_SIZE: usize = 1024;
 const MAX_BLOCK_TXS_SIZE: usize = 524_288;
 const MAX_BLOCK_CONTRACTS_SIZE: usize = 524_288;
+const TOKEN_MIN_ISSUANCE_FEE: Amount = Amount::from_atoms(10_000_000_000_000);
+const TOKEN_MAX_URI_LEN: usize = 1024;
+const TOKEN_MAX_DEC_COUNT: u8 = 18;
+const TOKEN_MAX_TICKER_LEN: usize = 5;
+const TOKEN_MAX_ISSUANCE_ALLOWED: usize = 1;
```
azarovh, line 194: I think this constant is misleading. In current implementation number of issuance in tx cannot be changed, thus it shouldn't be part of chain configuration. If we are planning to configure this I suggest adding a TODO explaining what should be fixed.

azarovh (approved):

common/src/chain/tokens/mod.rs
```diff
@@ -15,14 +15,80 @@
...
+    // TokenReissueV1 {
+    //     token_id: TokenId,
+    //     amount_to_issue: Amount,
+    // },
+    // // A new NFT creation
+    // #[codec(index = 5)]
+    // NftIssuanceV1 {
+    //     data_hash: NftDataHash,
+    //     metadata_uri: Vec<u8>,
+    // },
```
stanislav-tkach, line 93: Should this commented code be removed? Let's add a TODO comment if we still need it.

stanislav-tkach (approved):

common/src/chain/config/mod.rs
sinitcin: I added the comment

common/src/chain/tokens/mod.rs
sinitcin: Added the comment

chainstate/src/detail/transaction_verifier/tokens.rs
```diff
@@ -0,0 +1,140 @@
...
+// TODO: get rid of these mut map references parameters and use self-contained objects for testability
+
+// Token registration saves the token id in the database with the transaction that issued it, and possibly some additional auxiliary data;
+// This helps in finding the relevant information of the token at any time in the future.
+pub fn register_tokens_issuance(
+    tokens_cache: &mut BTreeMap<TokenId, CachedTokensOperation>,
+    block_id: Id<Block>,
+    tx: &Transaction,
+) -> Result<(), TokensError> {
+    let was_tokens_issued = tx.outputs().iter().any(|output| is_tokens_issuance(output.value()));
```
iljakuklic, line 31: `was` -> `were`

chainstate/src/detail/transaction_verifier/tokens.rs
```diff
@@ -0,0 +1,140 @@
...
+    let token_id = token_id(tx).ok_or(TokensError::TokenIdCantBeCalculated)?;
+    match tokens_cache.entry(token_id) {
+        Entry::Occupied(entry) => {
+            let tokens_op = entry.into_mut();
+            *tokens_op = CachedTokensOperation::Erase(tx.get_id());
+        }
+        Entry::Vacant(_) => {
+            return Err(TokensError::InvariantBrokenUndoIssuanceOnNonexistentToken(
+                token_id,
+            ))
```
iljakuklic, line 84: Should this be a panic rather than a runtime error? Seems rolling back state where it was previously should not fail.

iljakuklic (approved):
Adding some comments, can be disregarded or addressed as a follow up. A bunch of things here will have to be reevaluated for the anything-can-be-a-gas-token feature.

TheQuantumPhysicist:
> anything-can-be-a-gas-token feature

It's already part of the plan. I even did one change specifically for that, which allows having tokens in block reward as input source (but it's currently not allowed in outputs, acting as a switch to the feature). 

One problem at a time.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    tx: &Transaction,
+    source_block_id: &Id<Block>,
+    amount_to_burn: &Amount,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: You mean make MLT a token just like others? I'm not sure I understand.

muursh (approved):

chainstate/src/detail/transaction_verifier/amounts_map.rs
```diff
@@ -0,0 +1,124 @@
...
+    amount: Amount,
+) -> Result<(), TokensError> {
+    match total_amounts.get_mut(&key) {
+        Some(value) => {
+            *value = (*value + amount).ok_or(TokensError::CoinOrTokenOverflow)?;
+        }
+        None => {
+            total_amounts.insert(key, amount);
+        }
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I just tried this, and it doesn't work, because `value + amount` returns an Option.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -205,56 +221,103 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
     }
 
-    fn calculate_coins_total_inputs(
+    fn calculate_total_inputs(
         &self,
         inputs: &[TxInput],
-    ) -> Result<Amount, ConnectTransactionError> {
-        inputs.iter().try_fold(Amount::from_atoms(0), |total, input| {
-            let amount = self
+    ) -> Result<BTreeMap<CoinOrTokenId, Amount>, ConnectTransactionError> {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I agree. It's just a quick solution for now as I solved one problem with the amounts map, namely the `&mut BTreeMap` way of doing things... which is a security risk that I had to fix myself. There are still more, but yeah. We can clean this up over time.

chainstate/src/detail/transaction_verifier/tokens.rs
TheQuantumPhysicist: Typically, all InvariantBroken errors are supposed to panic. But I'm trying to be conservative in the interest of protecting the node from crashing because of some mistake we do.

chainstate/src/detail/tokens/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Exactly. We have to be very careful when doing things related to consensus. The last thing we want is to update the node and cause a fork. Cargo can quickly become our enemy.

PR #388, opened 2022-08-29
storage: Marker trait for transaction types

iljakuklic:
Useful for marking types where read/write operations provide transactional guarantees. Operations may require this trait to be implemented to enforce the code to be run inside a DB transaction.

This will only really become useful once operations in `chainstate-storage` start requiring the trait to be implemented. @sinitcin @TheQuantumPhysicist

### DIFF
--- a/chainstate-storage/src/internal/mod.rs
+++ b/chainstate-storage/src/internal/mod.rs
@@ -437,5 +437,8 @@ impl<'st, B: storage::Backend> crate::TransactionRw for StoreTxRw<'st, B> {
     }
 }
 
+impl<'st, B: storage::Backend> crate::IsTransaction for StoreTxRo<'st, B> {}
+impl<'st, B: storage::Backend> crate::IsTransaction for StoreTxRw<'st, B> {}
+
 #[cfg(test)]
 mod test;
--- a/chainstate-storage/src/is_transaction_seal.rs
+++ b/chainstate-storage/src/is_transaction_seal.rs
@@ -1,0 +1,25 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


TheQuantumPhysicist (approved):
Cool!

Maybe we should make this more crazy and mark the trait as unsafe, to prevent anyone from easily using it? Not sure.

iljakuklic:
An alternative approach would be to not provide the access to read/write operations outside of transaction at all, effectively forcing the user to denote which operations form a transaction explicitly. Currently, the direct operations only wrap the StartTransaction-DoOperation-EndTransaction sequence anyway. Thoughts / comments?

iljakuklic:
> Cool!
> 
> Maybe we should make this more crazy and mark the trait as unsafe, to prevent anyone from easily using it? Not sure.

If you mean to prevent it from being implemented on more types, that can be achieved with the [sealed trait pattern](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed)

altonen (approved):

TheQuantumPhysicist:
> An alternative approach would be to not provide the access to read/write operations outside of transaction at all, effectively forcing the user to denote which operations form a transaction explicitly. Currently, the direct operations only wrap the StartTransaction-DoOperation-EndTransaction sequence anyway. Thoughts / comments?

The only problem I can think of with that is that enforcing that every write/read is a transaction isn't like enforcing that a long-lived operation is a transaction. Aka, multiple atomics don't make an atomics. So this isn't really a solution at the same level of abstraction. Thing of this in the context of transaction verifier. Hopefully I'm not missing anything.

TheQuantumPhysicist:
> > Maybe we should make this more crazy and mark the trait as unsafe, to prevent anyone from easily using it? Not sure.
> 
> If you mean to prevent it from being implemented on more types, that can be achieved with the [sealed trait pattern](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed)

Yes and no. Sealed trait works but it just means no one (ever) will be able to make exceptions no matter what. Using unsafe means that people can, but it'll incur an unsafe, making it very obvious at the review with a much higher social approval cost. Both solutions are OK for now. We may, though, need to switch to the unsafe solution if a need every arises, albeit unlikely.

muursh (approved):

PR #389, opened 2022-08-29
p2p: Add tests for connection timeout

altonen:
Add tests to ensure that connection timeout is reported to the `PeerManager` and all the way to RPC if a new outbound connection was initiated by the user.

### DIFF
--- a/p2p/src/peer_manager/tests/connections.rs
+++ b/p2p/src/peer_manager/tests/connections.rs
@@ -54,7 +54,7 @@ where
         swarm.peer_connectivity_handle.poll_next().await,
         Ok(net::types::ConnectivityEvent::ConnectionError {
             address: _,
-            error: P2pError::DialError(DialError::IoError(std::io::ErrorKind::ConnectionRefused))
+            error: P2pError::DialError(DialError::ConnectionRefusedOrTimedOut)
         })
     ));
 }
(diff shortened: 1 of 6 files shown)


p2p/src/peer_manager/tests/connections.rs
```diff
@@ -434,3 +436,122 @@ async fn inbound_connection_too_many_peers_mock() {
...
+        swarm1.peer_connectivity_handle.poll_next(),
+    )
+    .await
+    {
+        Ok(res) => assert!(std::matches!(
+            res,
+            Ok(net::types::ConnectivityEvent::ConnectionError {
+                address: _,
+                error: P2pError::DialError(DialError::IoError(
+                    std::io::ErrorKind::ConnectionRefused
```
TheQuantumPhysicist, line 467: Why is a timed out connection showing as a refused connection?

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

muursh (approved):

iljakuklic (approved):

p2p/src/peer_manager/tests/connections.rs
altonen: Getting the correct error out would require either making all of the errors generic over the network service or having a leaky abstraction where the global error types would contain a `Multiaddr`. Admittedly there is some nuance lost if we treat connection refused and connection timeout both as connection refused but the result is handled the same way in both cases: we retry later again or we abandon this peer completely. Is it ok if I log the dial error using `log:.error!()` and keep this at least for now as `ConnectionRefused`?

TheQuantumPhysicist: Yes. Please log it correctly and change the error label to "ConnectionRefusedOrTimedout"

PR #391, opened 2022-08-30
storage: Backend test suite

iljakuklic:
Introduce a generic test suite that backends can use to check for correct functionality. At the moment, it is instantiated by the in-memory backend only.


storage/inmemory/tests/backend.rs
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+fn main() {
+    storage_backend_test_suite::main(storage_inmemory::InMemory::new())
+}
```
stanislav-tkach, line 18: Do I understand correctly that `libtest_mimic` is used to write backed-agnostic tests? So for several different storage backends it will look the following way:
```rust
fn main() {
    storage_backend_test_suite::main(storage_inmemory::InMemory::new());
    storage_backend_test_suite::main(<another_backend>);
}
```
Right?

Looks neat, though I'm still not sure if the straightforward approach (used in our p2p crate) is worse:
```rust
fn put_and_commit<B: Backend>(backend: B) {
    ...
}

#[test]
fn put_and_commit_inmemory() {
    put_and_commit(storage_inmemory::InMemory::new());
}
```

Sure, it requires a little bit more boilerplate (though you also have to additionally declare tests using the macro), but I'm really curious what output `libtest-mimic` produces.

To be clear: I'm not against using `libtest-mimic` specifically or any other fancy crates. 🙃

stanislav-tkach (approved):

storage/inmemory/tests/backend.rs
iljakuklic: > Do I understand correctly that `libtest_mimic` is used to write backed-agnostic tests? So for several different storage backends it will look the following way:
> 
> ```rust
> fn main() {
>     storage_backend_test_suite::main(storage_inmemory::InMemory::new());
>     storage_backend_test_suite::main(<another_backend>);
> }
> ```
> 
> Right?

Not quite. The idea is to have the test suite instantiated in each backend crate separately. I.e. the `storage-inmemory` crate has a test executable that just calls `storage_backend_test_suite::main::<InMemory>`, the `storage-lmdb` crate would have a test executable calling `storage_backend_test_suite::main::<LmdbBackend>` in its test suite, etc.

Ultimately, this approach stems from the desire for the test suite to be backend agnostic to the extent that it does not have any knowledge about the set of backends using it. In principle, it should be usable by third-party crates that provide storage backend outside of the project.

> Looks neat, though I'm still not sure if the straightforward approach (used in our p2p crate) is worse:
> 
> ```rust
> fn put_and_commit<B: Backend>(backend: B) {
>     ...
> }
> 
> #[test]
> fn put_and_commit_inmemory() {
>     put_and_commit(storage_inmemory::InMemory::new());
> }
> ```

I think the drawback of this is that if I add a test to the suite, a wrapper test needs to be added to each backend, which might be easily forgotten as there would be no warning. Ideally even for third-party backends we have no knowledge of. If there is a way to resolve that easily, I am open to suggestions.

> Sure, it requires a little bit more boilerplate (though you also have to additionally declare tests using the macro), but I'm really curious what output `libtest-mimic` produces.
> 
> To be clear: I'm not against using `libtest-mimic` specifically or any other fancy crates. upside_down_face

I understand that people may be uncomfortable using a fairly heavy dependency. However, it already is used in the project so it does not introduce a new dependency, and also it's a dev-time dependency which does not end up in the final executable. So while it's not risk-free, it's not too bad I think. As I already mentioned, suggestions regarding alternatives are welcome.

storage/backend-test-suite/src/basic.rs
```diff
@@ -0,0 +1,85 @@
...
+    assert_eq!(
+        dbtx.get(IDX.0, b"hello".as_ref()).as_ref().map(|x| x.as_deref()),
+        Ok(Some(b"b".as_ref())),
+    );
+    dbtx.commit().expect("commit to succeed");
+
+    let dbtx = store.transaction_ro();
+    assert_eq!(dbtx.get(IDX.0, b"hello"), Ok(Some(b"b".as_ref())),);
+}
+
```
TheQuantumPhysicist, line 79: If you need more ideas for what to test on such a test suite, take a look here

https://github.com/NeblioTeam/neblio/blob/83ce0da3c307226efa85ef5467ca46eab044ae46/wallet/test/db_tests.cpp

Every test that begins with TEST_P is one that you can implement for us as well, which is executed there for different backends.

TheQuantumPhysicist (approved):
Looking good. Just needs more tests. I understand this may be just a template.

altonen (approved):

storage/inmemory/tests/backend.rs
stanislav-tkach: > I think the drawback of this is that if I add a test to the suite, a wrapper test needs to be added to each backend, which might be easily forgotten as there would be no warning.

Yes, it is a convincing argument.

> However, it already is used in the project 

I missed that, so no objections anymore. 🙃

iljakuklic:
> Looking good. Just needs more tests. I understand this may be just a template.

I added a few more tests. However for many tests, the ability to iterate over all key/value pairs in given DB would be useful, so I plan to extend it a fair bit once that is in place.

PR #392, opened 2022-08-30
Make the Mock networking service generic over transport

stanislav-tkach:
`MockTransport`, `MockListener` and `MockStream` traits were introduced in order to abstract MockService from transport. There are two implementations for now: `ChannelMockTransport` and `TcpMockTransport`.


p2p/src/net/mock/backend.rs
```diff
@@ -133,14 +140,15 @@ impl Backend {
...
         local_peer_id: types::MockPeerId,
         remote_peer_id: types::MockPeerId,
         role: peer::Role,
-        state: ConnectionState,
+        state: ConnectionState<T>,
     ) -> crate::Result<()> {
         let (tx, rx) = mpsc::channel(16);
-        let socket = socket::MockSocket::new(socket);
+        // TODO: FIXME:
+        //let socket = socket::MockSocket::new(socket);
```
altonen, line 151: I think `socket` can be passed on `Peer` without converting it into a `MockSocket` first because `SocketService` provides everything that the `MockSocket` did

p2p/src/net/mock/mod.rs
```diff
@@ -35,47 +34,53 @@ use crate::{
...
     peer_id: types::MockPeerId,
 
     /// TX channel for sending commands to mock backend
-    cmd_tx: mpsc::Sender<types::Command>,
+    cmd_tx: mpsc::Sender<types::Command<U>>,
 
     /// RX channel for receiving connectivity events from mock backend
-    conn_rx: mpsc::Receiver<types::ConnectivityEvent>,
+    conn_rx: mpsc::Receiver<types::ConnectivityEvent<U>>,
     _marker: std::marker::PhantomData<fn() -> T>,
```
altonen, line 61: Since you export `std::marker::PhantomData`, this can be fixed as well

p2p/src/net/mock/mod.rs
```diff
@@ -35,47 +34,53 @@ use crate::{
...
+    U: TransportService,
 {
     /// TX channel for sending commands to mock backend
-    _cmd_tx: mpsc::Sender<types::Command>,
+    _cmd_tx: mpsc::Sender<types::Command<U>>,
 
     /// RX channel for receiving pubsub events from mock backend
-    _pubsub_rx: mpsc::Receiver<types::PubSubEvent>,
+    _pubsub_rx: mpsc::Receiver<types::PubSubEvent<U>>,
     _marker: std::marker::PhantomData<fn() -> T>,
```
altonen, line 74: And this and the one for syncing service

p2p/src/net/mock/transport/channel.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use once_cell::sync::Lazy;
+use tokio::sync::mpsc;
+
+use crate::net::mock::{types::Message, SocketService, TransportService};
+
+static NETWORK_HANDLE: Lazy<mpsc::UnboundedSender<SocketCommand>> = Lazy::new(|| {
+    let (tx, mut rx) = mpsc::unbounded_channel();
+    tokio::spawn(async move {
+        loop {
```
altonen, line 24: I know this is a draft but this can probably but moved into a separate function at some point

altonen: Maybe even an actual `MockNetwork` object

p2p/src/net/mock/transport/channel.rs
```diff
@@ -0,0 +1,96 @@
...
+                }
+            }
+        }
+    });
+    tx
+});
+
+enum SocketCommand {
+    Connect(u64),
+    Bind(u64),
```
altonen, line 45: Just a reminder but when you're implementing this functionality and you want to get the return value for, e.g., the `Connect` command, remember to add `oneshot::sender` to these enum values so the `MockNetwork` can send the return value for the peer that tried to connect.

p2p/src/net/mock/transport/channel.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct ChannelService {}
+
+pub struct ChannelSocket {}
+
+#[async_trait::async_trait]
+impl TransportService for ChannelService {
+    type Socket = ChannelSocket;
+    type Address = u64;
+
+    async fn bind(address: Self::Address) -> crate::Result<Self::Socket> {
```
altonen, line 58: One thing that is very convenient in tests is that we use port 0 to ask for the OS to allocate us a free port. It makes running the tests much easier as there are no port reuse issues. Something similar could be done for this interface as well so when a peer tries to bind to `0u64`, the `MockNetwork` allocates a port for them from its free list. That address (port) can then be queried using `SocketService::local_addr()`.

p2p/src/net/mock/transport/traits.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use crate::net::mock::types::Message;
+
+// TODO: FIXME: Better description.
+/// An abstraction layer for creating "sockets" ("connections").
+#[async_trait]
+pub trait TransportService: Sized {
+    type Socket: Send;
+
+    // TODO: FIXME: Remove Hash and Display?
```
altonen, line 31: If you want to relax these trait bounds, I think you also need to relax them in the `NetworkService` that is defined in `src/net/mod.rs`. I'm indifferent about removing `Hash` if you want to start using B trees but I don't know why you'd want to remove `Display`.

stanislav-tkach: I'm not insisting on these changes, but I thought that we prefer not to use `std::hash::Hash` in order to avoid confusion between cryptographic hashes that we use everywhere. As for `Display`, I think this bound is too restrictive, so `Debug` should be enough.

p2p/tests/sync.rs
```diff
@@ -629,17 +650,46 @@ where
     );
 }
 
+// TODO: FIXME: Move somewhere.
+struct MakeP2pAddress {}
+
+impl MakeAddress for MakeP2pAddress {
```
altonen, line 656: I'm still not convinced if this trait is needed to for the channel support but something like it would very convenient for making the tests less verbose. If the `NetworkService` would implement a `DefaultAddress` trait (or maybe just have a member function `NetworkService::default_address()`, there would be no need to call 'make_mock_addr()`/`make_libp2p_addr()` and the test code could instead initialize the managers like this:

```rust
    let (mut mgr1, mut conn1, _, mut pubsub, _) = make_sync_manager::<T>(T::default_address(), handle1).await;
    let (mut mgr2, mut conn2, _, _, _) = make_sync_manager::<T>(T::default_address(), handle2).await;
```

p2p/src/net/mock/transport/tcp.rs
```diff
@@ -0,0 +1,73 @@
...
+    async fn connect(address: Self::Address) -> Result<Self::Connection, super::Error> {
+        //TcpStream::connect(address)
+        todo!()
+    }
+}
+
+pub struct TcpConnection {}
+
+#[async_trait]
+impl Connection<TcpTransport> for TcpConnection {
```
altonen, line 59: This is probably copied from the example code I provided but I think the design of this is not yet finished.

The `Connection` trait probably has to be split into `Listener` and `Stream` traits because IMO it's a nonsensical operation to call `send()`/`recv()` on an object that outputs TCP connections (because there is nobody to send the data) and on the flip side calling `accept()` on an established TCP connection should not mean anything because the connection has already been accepted so there shouldn't be anything to accept anymore.

p2p/tests/sync.rs
TheQuantumPhysicist: This is actually specifically for the test, because there were methods you created to generate a test address. The trait should be called `MakeTestAddress`. The idea is that we want all addresses to be generated homogeneously so that we can use some generic method to run all the tests. Out of all the options I saw, I think this is the most reasonable one.

stanislav-tkach: I though about that, but in that case the `NetworkService` also needs to be implemented for `Libp2pServiceNetwork` and it doesn't seem right.

Originally I named this trait exactly MakeTestAddress, but then decided to shorten the name.  🙃  I'm going to rename it and move from the mocks to the tests.

p2p/src/net/mock/transport/channel.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        let (sender, receiver) = unbounded_channel();
+        assert!(connections.insert(address, sender).is_none());
+
+        Ok(Self::Listener { address, receiver })
+    }
+
+    async fn connect(address: Self::Address) -> Result<Self::Stream> {
+        // A connection can only be established to a known address.
+        debug_assert_ne!(ZERO_ADDRESS, address);
```
TheQuantumPhysicist, line 71: This is a mock... why debug?

p2p/src/net/mock/transport/traits.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use async_trait::async_trait;
+
+use crate::{net::mock::types::Message, Result};
+
+/// An abstraction layer for creating and opening connections.
+#[async_trait]
+pub trait MockTransport: Send + 'static {
+    /// An address type.
+    type Address: Copy + Clone + Debug + Eq + Hash + Send + Sync + ToString;
```
TheQuantumPhysicist, line 26: This is just for testing so it's not that necessary... but why does Address have to implement the Copy trait? Seems like a huge limitation.

p2p/src/sync/tests/block_response.rs
```diff
@@ -112,23 +139,34 @@ where
...
+    valid_block_invalid_state::<MakeP2pAddress, Libp2pService>(PeerId::random()).await;
 }
 
 // TODO: fix https://github.com/mintlayer/mintlayer-core/issues/375
 #[tokio::test]
 #[cfg(not(target_os = "macos"))]
-async fn valid_block_invalid_state_mock() {
-    valid_block_invalid_state::<MockService>(make_mock_addr(), MockPeerId::random()).await;
+async fn valid_block_invalid_state_mock_tcp() {
+    valid_block_invalid_state::<MakeTcpAddress, MockService<TcpMockTransport>>(MockPeerId::random()).await;
```
TheQuantumPhysicist, line 149: Please make sure that all tests use the same parameters but only differ by generic arguments.

TheQuantumPhysicist (approved):
Besides the few comments, looking good.

p2p/src/net/mock/transport/channel.rs
stanislav-tkach: I thought that having an assertion here isn't strictly necessary because this function returns an error if zero address is used anyway. But passing such address here indicated a developer error, so `debug_assert` feels appropriate. Fixed.

p2p/src/net/mock/transport/traits.rs
stanislav-tkach: I don't like this either and it isn't consistent with the `NetworkingService::Address` requirements, but the previous mocks implementation relies on copy-able addresses, so I decided to preserve this requirement in order to introduce fewer changes (such as `address.clone()`).

I'm going to remove the `Copy` bound.

muursh (approved):
lgtm

PR #393, opened 2022-08-31
Expose chainstate TestFramework to make it usable by other crates

royster57:
In the process of connecting the mempool and the chainstate (and replacing the mock chainstate used in the mempool tests by a real chainstate), the need arose for a way to generate transactions based on a real chainstate rather than just a mock.
To address this issue, we expose the testing framework used by the chainstate crate to be usable by other crates.


TheQuantumPhysicist (commented):
The problem with this is that it'll be compiled with all the code and make the binary much larger. Can you move the framework to another crate and make it a dev-dependency?

iljakuklic:
> The problem with this is that it'll be compiled with all the code and make the binary much larger. Can you move the framework to another crate and make it a dev-dependency?

There may be an issue if the framework uses some private items from the crate it's in. If it does not, moving it into another crate is a good solution.

I was thinking how to expose certain utility functions and modules from crates that are only used for testing that allows for private access if needed. The best I was able to come up with is the following: Define a feature called `testing` and guard the testing stuff with that feature using `#[cfg(feature = "testing")]`. In crate that depends on the crate, the `[dependencies]` section includes the dependency without the `testing` feature, `[dev-dependencies]` declare the dependency with the `testing` feature enabled. It would also be nice to have a standard name for the feature (here I suggest `testing`).

Any thoughts, related ideas, or objections?

chainstate/src/test_framework/framework.rs
```diff
@@ -146,14 +140,104 @@ impl TestFramework {
...
+        let outsrc = OutPointSourceId::BlockReward(id);
+        let txns = vec![(outsrc, genesis.utxos().to_vec())];
+        Self { txns, id }
+    }
+
+    pub(crate) fn from_id(cs: &TestChainstate, id: Id<GenBlock>) -> Self {
+        match id.classify(&cs.chain_config) {
+            GenBlockId::Genesis(_) => Self::from_genesis(cs.chain_config.genesis_block()),
+            GenBlockId::Block(id) => {
+                let block = cs.chainstate_storage.get_block(id).unwrap().unwrap();
```
iljakuklic, line 227: Clippy is configured to allow `unwrap` in the testing mode but not in production code. Since this is now not guarded by `#[cfg(test)]`, it starts complaining. You may want to add `#![allow(clippy::unwrap_used)]` to the top of the `test_framework` module.

TheQuantumPhysicist:
> There may be an issue if the framework uses some private items from the crate it's in. If it does not, moving it into another crate is a good solution.

I'm hoping we can get rid of the private dependencies. There's practically very little need for that. 

The feature stuff I consider a nuclear option... just like macros. It'll complicate things by an order of magnitude and make code readability much worse. I've already discussed the private usage issue with Roy. Let's see if he can find a way. Let's also not forget that when instantiating chainstate, subsystems won't be involved. So, querying things will become much simpler.

stanislav-tkach:
As far as I remember, the private items are used heavily in some of the tests, but I'm all for removing them. Additionally I think that the approach proposed by @iljakuklic is the best way to go.

iljakuklic:
Could the test framework crate be moved into a subfolder of `chainstate` so we don't pollute the top-level directory too much?

stanislav-tkach:
Probably this pull request should fix the https://github.com/mintlayer/mintlayer-core/pull/393 issue.

TheQuantumPhysicist:
> Could the test framework crate be moved into a subfolder of `chainstate` so we don't pollute the top-level directory too much?

That's in my mind. We have to be sure it's accessible though first. One step at a time.

chainstate/src/detail/mod.rs
```diff
@@ -44,28 +52,25 @@ use logging::log;
...
 
 pub type OrphanErrorHandler = dyn Fn(&BlockError) + Send + Sync;
 
 #[must_use]
 pub struct Chainstate<S> {
-    chain_config: Arc<ChainConfig>,
-    chainstate_config: ChainstateConfig,
-    chainstate_storage: S,
-    orphan_blocks: OrphanBlocksPool,
+    pub(crate) chain_config: Arc<ChainConfig>,
```
TheQuantumPhysicist, line 68: Instead of these invasive pub(crate), why not create proper reference getters that are pub?

chainstate/src/lib.rs
```diff
@@ -77,5 +83,23 @@ pub fn make_chainstate<S: chainstate_storage::BlockchainStorage + 'static>(
     Ok(Box::new(cons_interface))
 }
 
+pub fn make_chainstate_no_genesis<S: chainstate_storage::BlockchainStorage + 'static>(
```
TheQuantumPhysicist, line 86: Are you sure this is necessary? I don't think this is used. I followed the call tree and I don't see it used.

chainstate/src/detail/mod.rs
```diff
@@ -138,7 +145,7 @@ impl<S: BlockchainStorage> Chainstate<S> {
         Ok(chainstate)
     }
 
-    fn new_no_genesis(
+    pub(crate) fn new_no_genesis(
```
TheQuantumPhysicist, line 148: Maybe we can get rid of this if there are no callers in lib.rs like I mentioned in the other comment

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -58,4 +64,33 @@ pub trait ChainstateInterface: Send {
...
+    fn get_chainstate_config(&self) -> ChainstateConfig;
+    fn wait_for_all_events(&self);
+    fn get_mainchain_tx_index(
+        &self,
+        tx_id: &OutPointSourceId,
+    ) -> Result<Option<TxMainChainIndex>, ChainstateError>;
+    fn subscribers(&self) -> &Vec<EventHandler<ChainstateEvent>>;
+    fn calculate_median_time_past(&self, starting_block: &Id<GenBlock>) -> BlockTimestamp;
+    fn is_already_an_orphan(&self, block_id: &Id<Block>) -> bool;
+    fn num_orphans(&self) -> usize;
```
TheQuantumPhysicist, line 81: orphans_count

chainstate_test_framework/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+chainstate = { path = '../chainstate' }
+chainstate-storage = { path = '../chainstate-storage' }
+chainstate-types = { path = '../chainstate-types' }
+common = { path = '../common' }
+crypto = { path = '../crypto' }
+rstest = "0.15"
```
TheQuantumPhysicist, line 14: We're using the arbitrary convention to separate external crates with a new line from our internal crates

p2p/src/net/libp2p/behaviour/connection_manager/mod.rs
```diff
@@ -143,9 +143,7 @@ impl ConnectionManager {
                 self.add_event(ConnectionManagerEvent::Behaviour(
                     BehaviourEvent::ConnectionError {
                         address: connection.addr().clone(),
-                        error: P2pError::DialError(DialError::IoError(
-                            std::io::ErrorKind::ConnectionRefused,
-                        )),
+                        error: P2pError::DialError(DialError::ConnectionRefusedOrTimedOut),
```
TheQuantumPhysicist, line 146: This makes me think that a rebase (as opposed to a merge) is necessary to avoid redoing what's being already done in master

p2p/src/peer_manager/tests/connections.rs
```diff
@@ -52,7 +54,7 @@ where
         swarm.peer_connectivity_handle.poll_next().await,
         Ok(net::types::ConnectivityEvent::ConnectionError {
             address: _,
-            error: P2pError::DialError(DialError::IoError(std::io::ErrorKind::ConnectionRefused))
+            error: P2pError::DialError(DialError::ConnectionRefusedOrTimedOut)
```
TheQuantumPhysicist, line 57: This stuff is everywhere... it shouldn't be in your PR.

storage/backend-test-suite/src/basic.rs
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2022 RBB S.r.l
```
TheQuantumPhysicist, line 1: omg... something is really wrong with this branch merge.

TheQuantumPhysicist (commented):
The work is OK. Please mind my comments. 

I think something is wrong with the merge process. This PR somehow includes many irrelevant things from storage that Lukas did and from p2p that Aaro did. Please squash everything and rebase upon master and let's see what will be left.

chainstate/src/detail/mod.rs
```diff
@@ -297,6 +304,13 @@ impl<S: BlockchainStorage> Chainstate<S> {
         self.make_db_tx_ro().get_best_block_id()
     }
 
+    pub fn get_mainchain_tx_index(
```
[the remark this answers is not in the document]
royster57: Done

chainstate/src/interface/chainstate_interface.rs
royster57: Done.

chainstate/src/detail/mod.rs
royster57: Done.

chainstate/src/lib.rs
royster57: It's used in the test framework, but only one test requires it. I'll move this test back to the chainstate crate and have it not use the framework, then we can get rid of this function.

chainstate/src/interface/chainstate_interface.rs
royster57: Done.

chainstate/src/lib.rs
royster57: Done.

Cargo.toml
```diff
@@ -15,6 +15,7 @@ members = [
   "crypto",                       # Cryptographic primitives and their interfaces.
   "consensus",                    # Consensus related logic.
   "chainstate",                   # Code on chainstate of blocks and transactions.
+  "chainstate_test_framework",    # Test framework for chainstate-related tests.
```
TheQuantumPhysicist, line 18: I don't think this should be a part of the workspace... it should just be a dependency. Are you sure this is the right way to do this?
For example, chainstate-types isn't here. Because it's just a dependency.

chainstate_test_framework/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+chainstate = { path = '../chainstate' }
+chainstate-storage = { path = '../chainstate-storage' }
+chainstate-types = { path = '../chainstate-types' }
+common = { path = '../common' }
+crypto = { path = '../crypto' }
+serialization = {path = "../serialization"}
+utils = { path = '../utils' }
+
+#[dev-dependencies]
```
TheQuantumPhysicist, line 19: ?

chainstate/src/test.rs
```diff
@@ -13,9 +13,80 @@
...
+fn empty_chainstate_no_genesis() {
+    utils::concurrency::model(|| {
+        let chain_config = ChainConfigBuilder::new(ChainType::Mainnet)
+            .net_upgrades(NetUpgrades::unit_tests())
+            .genesis_unittest(Destination::AnyoneCanSpend)
+            .build();
+        let chainstate_config = ChainstateConfig::default();
+        let chainstate_storage = Store::new_empty().unwrap();
+        let time_getter = TimeGetter::default();
+        let chainstate = Chainstate::new_no_genesis(
```
TheQuantumPhysicist, line 82: How about we move these tests inside the detail module to avoid that unnecessary pub(crate)?

chainstate/src/detail/mod.rs
```diff
@@ -304,7 +311,20 @@ impl<S: BlockchainStorage> Chainstate<S> {
...
-mod tests;
+    pub fn get_chain_config(&self) -> Arc<ChainConfig> {
+        self.chain_config.clone()
+    }
+
+    pub fn get_chainstate_config(&self) -> &ChainstateConfig {
+        &self.chainstate_config
+    }
+
+    pub fn get_orphan_blocks(&self) -> &OrphanBlocksPool {
```
TheQuantumPhysicist, line 323: orphan_blocks_pool()

TheQuantumPhysicist (approved):
Getting closer! Please address the few issues left and merge.

TheQuantumPhysicist (commented):
I would appreciate it if others can review this branch. Long discussions have happened because of this. Is everyone happy?

iljakuklic (approved):

azarovh (approved):

chainstate/src/detail/block_index_history_iter.rs
```diff
@@ -63,51 +63,128 @@ impl<'a, H: BlockIndexHandle> Iterator for BlockIndexHistoryIterator<'a, H> {
...
-    use super::*;
+    use crate::detail::Chainstate;
+    use crate::BlockSource;
+    use crate::ChainstateConfig;
 
     #[test]
     fn history_iteration() {
         utils::concurrency::model(|| {
-            let mut tf = TestFramework::default();
+            let chain_config = ChainConfigBuilder::new(ChainType::Mainnet)
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Because we can't use the test framework as a dependency in chainstate because it'll cause circular dependency. We either move this to the test framework crate or we make it independent of it. One of these have to happen.

muursh (approved):
I'm happy with this.

chainstate/src/detail/mod.rs
```diff
@@ -44,15 +50,14 @@ use utxo::UtxosDBMut;
...
-    time_getter::TimeGetter,
 };
 use crate::{detail::orphan_blocks::OrphanBlocksPool, ChainstateConfig, ChainstateEvent};
 
 type TxRw<'a, S> = <S as Transactional<'a>>::TransactionRw;
 type TxRo<'a, S> = <S as Transactional<'a>>::TransactionRo;
 type ChainstateEventHandler = EventHandler<ChainstateEvent>;
 
-const HEADER_LIMIT: BlockDistance = BlockDistance::new(2000);
+pub const HEADER_LIMIT: BlockDistance = BlockDistance::new(2000);
```
stanislav-tkach, line 60: It isn't directly related to this pull request, but this constant should probably be moved to the config. Lets add a TODO-node. By the way, the same constant is used in the p2p code...

chainstate_test_framework/src/framework.rs
```diff
@@ -42,6 +39,14 @@ pub struct TestFramework {
 }
 
 impl TestFramework {
+    /// Creates a new test framework
+    pub fn new(chainstate: TestChainstate, block_indexes: Vec<BlockIndex>) -> Self {
```
stanislav-tkach, line 43: As far as I can see, this construct isn't used. Am I missing something?

chainstate_test_framework/src/lib.rs
```diff
@@ -22,9 +22,13 @@ mod transaction_builder;
...
-pub use self::{
-    block_builder::BlockBuilder, framework::TestFramework, framework_builder::TestFrameworkBuilder,
+pub use {
+    block_builder::BlockBuilder, framework::anyonecanspend_address, framework::empty_witness,
+    framework::TestBlockInfo, framework::TestFramework, framework_builder::TestFrameworkBuilder,
     transaction_builder::TransactionBuilder,
 };
+
+#[cfg(test)]
+mod tests;
```
stanislav-tkach, line 34: I feel like these tests belong somewhere else. I would expect for the tests in the `test_framework` crate to test the framework itself and these are chainstate tests. I think it would be better to move them to a separate crate.

stanislav-tkach (approved):

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -59,4 +65,37 @@ pub trait ChainstateInterface: Send {
         &self,
         headers: Vec<BlockHeader>,
     ) -> Result<Vec<BlockHeader>, ChainstateError>;
+
+    fn get_block_index(&self, id: &Id<Block>) -> Result<Option<BlockIndex>, ChainstateError>;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: It's not really that big. That's still nothing. Big in this case is not really measured by number of lines, but whether there's a logical way to split it. I don't see any, so far.

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -59,4 +65,37 @@ pub trait ChainstateInterface: Send {
...
+    ) -> Result<GenBlockIndex, ChainstateError>;
+    fn last_common_ancestor(
+        &self,
+        first_block_index: &GenBlockIndex,
+        second_block_index: &GenBlockIndex,
+    ) -> Result<GenBlockIndex, ChainstateError>;
+    fn get_block_reward(
+        &self,
+        block_index: &BlockIndex,
+    ) -> Result<Option<BlockReward>, ChainstateError>;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Absolutely. This is happening and one of the reasoning why I find this PR useful.

chainstate_test_framework/src/framework.rs
TheQuantumPhysicist: Removed

chainstate_test_framework/src/tests/history_iteration.rs
```diff
@@ -0,0 +1,51 @@
+use chainstate::BlockSource;
+
+use crate::TestFramework;
+
+#[test]
+fn history_iteration() {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Seems like it. I'll remove the other instance.

chainstate_test_framework/src/lib.rs
TheQuantumPhysicist: I'll see what I can do.

PR #396, opened 2022-09-01
Mock interface macOS issue MRE

altonen:


altonen:
@stanislav-tkach @TheQuantumPhysicist 

I think the Mac OS issue is fixed now. It was caused by [`tokio_util::io::read_buf()`](https://docs.rs/tokio-util/latest/tokio_util/io/fn.read_buf.html) returning `Poll::Pending` if the socket didn't have any more data. Unlike UDP, TCP is allowed to coalesce the incoming data into one continuous stream but the code was expecting to read discrete frames from the socket and thus it caused problems because the tests were expecting a certain amount of frames to be received, one for each call to `read_buf()` which was not correct. Sometimes there was some data left over in the stream buffer from previous calls but these were never decoded because the code called `read_buf()` every time and at some point would block if there was no more data to be received from the socket even though everything was received already.

The implementation is updated to first decode the buffer and if it doesn't have enough bytes to decode into a valid frame, it will try to read more bytes from the socket.

TheQuantumPhysicist:
> @stanislav-tkach @TheQuantumPhysicist
> 
> I think the Mac OS issue is fixed now. It was caused by [`tokio_util::io::read_buf()`](https://docs.rs/tokio-util/latest/tokio_util/io/fn.read_buf.html) returning `Poll::Pending` if the socket didn't have any more data. Unlike UDP, TCP is allowed to coalesce the incoming data into one continuous stream but the code was expecting to read discrete frames from the socket and thus it caused problems because the tests were expecting a certain amount of frames to be received, one for each call to `read_buf()` which was not correct. Sometimes there was some data left over in the stream buffer from previous calls but these were never decoded because the code called `read_buf()` every time and at some point would block if there was no more data to be received from the socket even though everything was received already.
> 
> The implementation is updated to first decode the buffer and if it doesn't have enough bytes to decode into a valid frame, it will try to read more bytes from the socket.

Wow. Great stuff. Thanks for having figured it out. I'll test it tomorrow extensively. Cheers!

TheQuantumPhysicist (approved):
Looking good. Thanks, Aaro!

PR #398, opened 2022-09-01
Use utxo set to calculate block total fees

azarovh:
Use utxo set to calculate block total fees.

Pay attention that I had to change the meaning of spender in `DoubleSpendAttempt(Spender)` error. Previously Spender was a tx where the original output was spent. But since utxo set doesn't store spent outputs I redefine Spender as a tx where double spend occurred. Can be clearly seen in the test.

### DIFF
--- a/chainstate/src/detail/chainstateref.rs
+++ b/chainstate/src/detail/chainstateref.rs
@@ -539,23 +539,20 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         debug_assert!(reward_fees.is_none());
 
         // TODO: add a test that checks the order in which txs are connected
-        let total_fees = block
-            .transactions()
-            .iter()
-            .enumerate()
-            .map(|(tx_num, _)| {
-                tx_verifier.connect_transactable(
+        let total_fees = block.transactions().iter().enumerate().try_fold(
+            Amount::from_atoms(0),
+            |total, (tx_num, _)| {
+                let fee = tx_verifier.connect_transactable(
                     BlockTransactableRef::Transaction(block, tx_num),
                     spend_height,
                     &median_time_past,
                     blockreward_maturity,
-                )
-            })
-            .try_fold(Amount::from_atoms(0), |total, fee| {
-                (total + fee?.expect("connect tx should return fees").0).ok_or_else(|| {
+                )?;
+                (total + fee.expect("connect tx should return fees").0).ok_or_else(|| {
                     ConnectTransactionError::FailedToAddAllFeesOfBlock(block.get_id())
                 })
-            })?;
+            },
+        )?;
 
         let block_subsidy = self.chain_config.block_subsidy_at_height(spend_height);
         tx_verifier.check_block_reward(block, Fee(total_fees), Subsidy(block_subsidy))?;
--- a/chainstate/src/detail/chainstateref.rs
+++ b/chainstate/src/detail/chainstateref.rs
@@ -572,7 +572,7 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
             TransactionVerifier::new(&self.db_tx, utxo_view.derive_cache(), self.chain_config);
 
         // TODO: add a test that checks the order in which txs are disconnected
-        block.transactions().iter().rev().enumerate().try_for_each(|(tx_num, _)| {
+        block.transactions().iter().enumerate().rev().try_for_each(|(tx_num, _)| {
             tx_verifier.disconnect_transactable(BlockTransactableRef::Transaction(block, tx_num))
         })?;
         tx_verifier.disconnect_transactable(BlockTransactableRef::BlockReward(block))?;
--- a/chainstate/src/detail/chainstateref.rs
+++ b/chainstate/src/detail/chainstateref.rs
@@ -24,7 +24,7 @@ use common::{
         },
         Block, ChainConfig, GenBlock, GenBlockId, OutPointSourceId,
     },
-    primitives::{BlockDistance, BlockHeight, Id, Idable},
+    primitives::{Amount, BlockDistance, BlockHeight, Id, Idable},
     Uint256,
 };
 use consensus::{BlockIndexHandle, TransactionIndexHandle};
@@ -37,7 +37,9 @@ use crate::{BlockError, BlockSource, ChainstateConfig};
 
 use super::{
     orphan_blocks::{OrphanBlocks, OrphanBlocksMut},
-    transaction_verifier::{BlockTransactableRef, TransactionVerifier},
+    transaction_verifier::{
+        error::ConnectTransactionError, BlockTransactableRef, Fee, Subsidy, TransactionVerifier,
+    },
     BlockSizeError, CheckBlockError, CheckBlockTransactionsError, OrphanCheckError,
 };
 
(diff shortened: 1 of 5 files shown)


chainstate/src/detail/transaction_verifier/error.rs
```diff
@@ -74,7 +74,7 @@ pub enum ConnectTransactionError {
     BlockTimestampArithmeticError,
     #[error("Input addition error")]
     InputAdditionError,
-    #[error("Double-spend attempt")]
+    #[error("Double-spend attempt in `{0}`")]
```
stanislav-tkach, line 77: You can use `{0:?}` in order to not implement `Display` for `Spender`.

chainstate/src/detail/chainstateref.rs
```diff
@@ -528,25 +530,35 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
-                BlockTransactableRef::Transaction(block, tx_num),
-                spend_height,
-                &median_time_past,
-                blockreward_maturity,
-            )?;
-        }
+        let total_fees = block
+            .transactions()
+            .iter()
+            .enumerate()
```
stanislav-tkach, line 545: Nitpick: I find the usage of `enumerate` here slightly misleading. You can use `0..block.transactions().len().map(...)` instead.

chainstate/src/detail/chainstateref.rs
```diff
@@ -558,11 +570,13 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
     ) -> Result<TransactionVerifier<S>, BlockError> {
         let mut tx_verifier =
             TransactionVerifier::new(&self.db_tx, utxo_view.derive_cache(), self.chain_config);
+
         // TODO: add a test that checks the order in which txs are disconnected
-        block.transactions().iter().rev().enumerate().try_for_each(|(tx_num, _tx)| {
-            tx_verifier.disconnect_transaction(BlockTransactableRef::Transaction(block, tx_num))
+        block.transactions().iter().rev().enumerate().try_for_each(|(tx_num, _)| {
```
stanislav-tkach, line 575: Same here.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -40,11 +40,14 @@ use common::{
...
 use utxo::{
     BlockRewardUndo, BlockUndo, ConsumedUtxoCache, FlushableUtxoView, TxUndo, UtxosCache,
-    UtxosDBMut,
+    UtxosDBMut, UtxosView,
 };
 
 use self::error::ConnectTransactionError;
 
+pub struct Fee(pub Amount);
+pub struct Subsidy(pub Amount);
```
stanislav-tkach, line 49: It would be nice to add some doc-comments. The `Fee` type is pretty self explanatory, `Subsidy` is a little less.

stanislav-tkach (approved):

chainstate/src/detail/chainstateref.rs
```diff
@@ -558,11 +570,13 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+            tx_verifier.disconnect_transactable(BlockTransactableRef::Transaction(block, tx_num))
```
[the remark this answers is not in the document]
azarovh: Nice catch! Actually we don't enumerate txs here just use indexes. So rev() has no effect here - which is a bug. I'll rework this as Stas suggested in his comment.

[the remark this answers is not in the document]
TheQuantumPhysicist: Just put rev() after enumerate?

chainstate/src/detail/transaction_verifier/mod.rs
azarovh: These types are pure syntactic sugar to not misuse Amount in arguments. Subsidy is as common term as Fee (meaning amount of new coins minted by new block). So I think comment is redundant here

chainstate/src/detail/chainstateref.rs
stanislav-tkach: @TheQuantumPhysicist I'm not insisting, but for me `enumerate` is a clear sign that we need to have an index in addition to the item we are iterating over, so I have spent some time checking why it is ignored. By the way, I don't see many possibilities for an error in the `0..len` range.

TheQuantumPhysicist: Sure, when you know 0..len works, you do it. But then you get someone else modifying the code for some reason, rewriting it, and doing 0..len-1 reasoning that the length of the vector is len-1, and that happened here in Mintlayer a couple of months ago. My solution: Manual ranges only when we must.

I get your view on enumerate and you're right, but I'm more than happy to abuse it in the interest of avoiding a bug. This whole software is pivoted on security and the best practices we can come up with to reduce the probability of random bugs.

TheQuantumPhysicist:
> But since utxo set doesn't store spent outputs I redefine Spender as a tx where double spend occurred. Can be clearly seen in the test. 

Unfortunately that kind of doesn't make sense for two reasons. First, a spender is an input that spends, and second, this information is useful for debugging to know what tx is the culprit, but now the information is kind of useless because it'll never be queriable. But I understand your motivation since utxo set doesn't include that information. I'll think about it.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -306,12 +246,32 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+        &self,
+        inputs: &[TxInput],
+        spender: Spender,
+    ) -> Result<Amount, ConnectTransactionError> {
+        inputs
+            .iter()
+            .map(|input| {
+                self.utxo_cache
+                    .utxo(input.outpoint())
+                    .ok_or_else(|| ConnectTransactionError::DoubleSpendAttempt(spender.clone()))
```
TheQuantumPhysicist, line 265: The problem here is that calling this double-spend is not really fair or accurate. It may or may not be a double-spend. In bitcoin it's called "InputMissingOrSpent" or something similar. That's more accurate because it could be either.

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -572,13 +520,14 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
-    pub fn connect_transaction(
+    pub fn connect_transactable(
         &mut self,
         spend_ref: BlockTransactableRef,
         spend_height: &BlockHeight,
         median_time_past: &BlockTimestamp,
         blockreward_maturity: &BlockDistance,
-    ) -> Result<(), ConnectTransactionError> {
+    ) -> Result<Option<Fee>, ConnectTransactionError> {
+        let mut fee: Option<Fee> = None;
```
TheQuantumPhysicist, line 530: Nope... no mutable variables.

TheQuantumPhysicist: I pushed a fix.

Please always avoid mutable variables. They're usually the source of undefined behavior. We only use mutable under heavily controlled environments and only when we must.

TheQuantumPhysicist (approved):
LGTM. Please mind the comments. Let's merge this as soon as we can.

chainstate/src/detail/chainstateref.rs
```diff
@@ -528,25 +530,35 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                    spend_height,
+                    &median_time_past,
+                    blockreward_maturity,
+                )
+            })
+            .try_fold(Amount::from_atoms(0), |total, fee| {
+                (total + fee?.expect("connect tx should return fees").0).ok_or_else(|| {
+                    ConnectTransactionError::FailedToAddAllFeesOfBlock(block.get_id())
+                })
+            })?;
```
iljakuklic, line 558: Would it make sense to fuse the `map` and `try_fold` calls?

I.e. replace this pattern:
```rust
iter.map(|a| f(a)).try_fold(b0, |b, fa| g(b, fa))
```
with
```rust
iter.try_fold(b0, |b, a| g(b, f(a)))
```

chainstate/src/detail/transaction_verifier/mod.rs
```diff
@@ -306,12 +246,31 @@ impl<'a, S: BlockchainStorageRead> TransactionVerifier<'a, S> {
...
+                self.utxo_cache
+                    .utxo(input.outpoint())
+                    .ok_or(ConnectTransactionError::MissingOutputOrSpent)
+                    .map(|utxo| match utxo.output().value() {
+                        OutputValue::Coin(amount) => *amount,
+                    })
+            })
+            .try_fold(Amount::from_atoms(0), |total, amount| {
+                (total + amount?).ok_or(ConnectTransactionError::OutputAdditionError)
+            })
```
iljakuklic, line 271: Similar fusion opportunity here.

iljakuklic (approved):
Some minor comments

PR #402, opened 2022-09-06
Isolate query functions

TheQuantumPhysicist:
This PR has no functional benefit, but it makes chainstate main file cleaner and smaller so that it doesn't overwhelm the reader.

### DIFF
--- a/chainstate/src/detail/query.rs
+++ b/chainstate/src/detail/query.rs
@@ -1,0 +1,146 @@
+use chainstate_storage::BlockchainStorageRead;
+use chainstate_types::{BlockIndex, GenBlockIndex, Locator, PropertyQueryError};
+use common::{
+    chain::{
+        block::{BlockHeader, BlockReward},
+        Block, GenBlock,
+    },
+    primitives::{BlockDistance, BlockHeight, Id, Idable},
+};
+
+use super::{chainstateref, orphan_blocks::OrphanBlocks, HEADER_LIMIT};
+
+pub fn locator_tip_distances() -> impl Iterator<Item = BlockDistance> {
+    itertools::iterate(0, |&i| std::cmp::max(1, i * 2)).map(BlockDistance::new)
+}
+
+pub struct ChainstateQuery<'a, S, O> {
+    chainstate_ref: chainstateref::ChainstateRef<'a, S, O>,
+}
+
+impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateQuery<'a, S, O> {
+    pub(crate) fn new(chainstate_ref: chainstateref::ChainstateRef<'a, S, O>) -> Self {
+        Self { chainstate_ref }
+    }
+
+    pub fn get_best_block_id(&self) -> Result<Id<GenBlock>, PropertyQueryError> {
+        self.chainstate_ref.get_best_block_id()
+    }
+
+    #[allow(dead_code)]
+    pub fn get_block_reward(
+        &self,
+        block_index: &BlockIndex,
+    ) -> Result<Option<BlockReward>, PropertyQueryError> {
+        self.chainstate_ref.get_block_reward(block_index)
+    }
+
+    #[allow(dead_code)]
+    pub fn get_header_from_height(
+        &self,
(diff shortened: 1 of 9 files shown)


chainstate/src/detail/query.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use super::{chainstateref, orphan_blocks::OrphanBlocks, HEADER_LIMIT};
+
+pub fn locator_tip_distances() -> impl Iterator<Item = BlockDistance> {
+    itertools::iterate(0, |&i| std::cmp::max(1, i * 2)).map(BlockDistance::new)
+}
+
+pub struct ChainstateQuery<'a, S, O> {
+    chainstate_ref: chainstateref::ChainstateRef<'a, S, O>,
+}
```
iljakuklic, line 19: Why the extra wrapper type?

TheQuantumPhysicist: I couldn't find a better way do isolate query functions in a different module. Either this, or floating references and generics all over the place. Feel free to suggest a different approach.

iljakuklic: I think methods do not need to be defined in the same module as the type. The type just needs to be visible and in the same crate. The following compiles for me:

```rust
struct Foo(i32);

mod query {
    impl super::Foo {
        fn get_x(&self) -> i32 { self.0 }
    }
}
```

TheQuantumPhysicist: Do you find this cleaner? I'm kind of hesitant to scatter impl blocks outside the main file for the same type. I still prefer an extra thin wrapper with the bonus of read-only guarantees.

iljakuklic: More detailed docs here: https://doc.rust-lang.org/reference/items/implementations.html#inherent-implementations

iljakuklic: > Do you find this cleaner? I'm kind of hesitant to scatter impl blocks outside the main file for the same type. I still prefer an extra thin wrapper with the bonus of read-only guarantees.

I think merely separating a big set of methods into multiple files should not introduce a new type just for that reason. People better get used to the fact that methods do not need to be in the same file (IDEs will help here). If there is another motivation or benefit of having the extra type there, I'm fine with that, but just moving methods around does not quite cut it in my book.

TheQuantumPhysicist: I lean more towards the new wrapper type. What do you guys think, @muursh @stanislav-tkach @azarovh ?

stanislav-tkach: I think that multiple impl-blocks in different files can be confusing, but personally I still prefer this approach to introducing wrappers. My motivation is the following: the structure rises more questions, such as “Why is it needed?”. Sure, a comment can help, but separate impl-blocks feel more natural in that case.

azarovh: Imho the optimal solution would be to move chainstateref to a separate folder and then spilt it in multiple files. Thus all the logic will be isolated and split in logical parts in the same time.

muursh: My gut feeling here is that I prefer the wrapper approach.

muursh (approved):
I'm good with this as is

chainstate/src/detail/query.rs
TheQuantumPhysicist: With the split I see here, and after having thought about it and considered all options, I think using the wrapper is worth it for one reason: It enforces that the query interface is only used in read-only context, preventing future queries from becoming mutable on the whims of a mistaken developer misunderstanding how it should be used. It's much harder to make a query mutable than just to change `make_db_tx_ro` to `make_db_tx`. I thought this is just a bonus, but now I realize it prevents possible mistakes where people make the ref mutable.

TheQuantumPhysicist:
Other than the wrapper thing, is there any other problem?

TheQuantumPhysicist:
Btw, chainstateref.rs is also large and I'm wondering how to make it smaller.

iljakuklic (approved):

stanislav-tkach (approved):

PR #403, opened 2022-09-06
Make reward maturity use locked outputs instead of manually checking spending time

TheQuantumPhysicist:


azarovh (approved):

stanislav-tkach (approved):

common/src/chain/pow.rs
```diff
@@ -27,6 +28,8 @@ pub struct PoWChainConfig {
     allow_min_difficulty_blocks: bool,
     /// The lowest possible difficulty
     limit: Uint256,
+    /// The distance required to pass to allow spending the block reward
+    reward_maturity_distance: BlockDistance,
```
iljakuklic, line 32: Why is this specific to `PowChainConfig`? Would it be appropriate for the setting to be agnostic to the consensus engine used?

chainstate/src/detail/chainstateref.rs
```diff
@@ -374,7 +374,62 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
...
+                        ))
+                    }
+                    common::chain::timelock::OutputTimeLock::ForBlockCount(c) => {
+                        let cs: i64 = (*c).try_into().map_err(|_| {
+                            CheckBlockError::InvalidBlockRewardMaturityDistanceValue(
+                                block.get_id(),
+                                *c,
+                            )
+                        })?;
+                        let given = BlockDistance::new(cs);
```
iljakuklic, line 405: Not really related to this PR, but `OutputTimeLock::ForBlockCount` should really take `NonNegativeBlockDistance` or something like that rather than `u64`. The conversions here are a bit awkward.

iljakuklic (approved):
Looks good overall

common/src/chain/pow.rs
TheQuantumPhysicist: Nope. It depends on the consensus algorithm. That's a point we can discuss some time later.

But the technicality of where to put it and where to pull it from is an open question. I used this because it's part of chain config.

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: Making it nice is something we can do later. But having it signed is an issue because we need a compact form for it, and signed doesn't support compact.

PR #404, opened 2022-09-06
Use WithId<Block> instead of Block to avoid recalculating hashes too often.

TheQuantumPhysicist:


stanislav-tkach (approved):

stanislav-tkach:
I'm really curious about how such changes impact our overall performance. Too bad we don't have benchmarks. 🙃

azarovh (approved):

iljakuklic (approved):

TheQuantumPhysicist:
> I'm really curious about how such changes impact our overall performance. Too bad we don't have benchmarks. 🙃

In a previous project, scrypt block hash calculation took 40%-50% of the CPU usage during syncing. It can really be a lot!

PR #405, opened 2022-09-06
Rerun CI when target branch is updated

TheQuantumPhysicist:
With this change, I'm hoping that merging some PR with master will trigger CI in all other PRs. This is necessary to ensure that all merges yield a working copy of master/staging.

The documentation cannot be any more confusing. I'm not sure this does what I think it does:

https://github.com/orgs/community/discussions/24567

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,14 +1,17 @@
 name: build
 
 on:
-    push:
-      branches:
-        - master
-        - staging
-    pull_request:
-      branches:
-        - master
-        - staging
+  push:
+    branches:
+      - master
+      - staging
+  pull_request:
+    branches:
+      - master
+      - staging
+  pull_request_target:
+    types: synchronize
+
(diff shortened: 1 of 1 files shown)


muursh (approved):
Okay so I can only go on the docs too and I think this look right. I guess we can try it

stanislav-tkach (approved):
Interesting, I wasn't aware of this possibility. By the way, there is also the "Require branches to be up to date before merging" option in the repository setting. Doesn't it give the same result?

TheQuantumPhysicist:
> Interesting, I wasn't aware of this possibility. By the way, there is also the "Require branches to be up to date before merging" option in the repository setting. Doesn't it give the same result?

Requiring basically means it won't allow you to merge unless you reverse merge with master or rebase. Quite bad if you ask me.

PR #408, opened 2022-09-07
Move accounting delta to a separate crate

azarovh:
After this I will create separate PR which adds accounting crate to master.


accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -32,7 +45,6 @@ pub struct DeltaDataCollection<K: Ord, T> {
 }
 
 impl<K: Ord + Copy, T> DeltaDataCollection<K, T> {
-    #[cfg(test)]
     pub fn from_data<const N: usize>(data: [(K, DataDelta<T>); N]) -> Self {
```
TheQuantumPhysicist, line 48: If you wanna keep this kind of initialization without it being exclusive for tests, perhaps it makes more sensible to implement the FromIterator trait for `(K, DataDelta<T>)` instead of this. This interface isn't ideal.

TheQuantumPhysicist (approved):
Except for my minor comments, looking good.

accounting/src/delta/delta_amount_collection.rs
```diff
@@ -12,7 +27,6 @@ pub struct DeltaAmountCollection<K: Ord> {
 }
 
 impl<K: Ord> DeltaAmountCollection<K> {
-    #[cfg(test)]
     pub fn from_data<const N: usize>(data: [(K, SignedAmount); N]) -> Self {
```
TheQuantumPhysicist, line 30: Same here

PR #409, opened 2022-09-07
Introduce accounting crate

azarovh:
Introduce crate with basic accounting. It will be used in PoS accounting or any other accounting type we would have.

It contains the logic of working with amount and data deltas as a basic building block for accounting.

### DIFF
--- a/accounting/Cargo.toml
+++ b/accounting/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "accounting"
+license = "MIT"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+common = {path = '../common'}
+serialization = {path = "../serialization"}
+
+thiserror = "1.0"
+parity-scale-codec = "3.1"
--- a/accounting/src/delta/combine.rs
+++ b/accounting/src/delta/combine.rs
@@ -1,0 +1,104 @@
+// Copyright (c) 2021 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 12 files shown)


TheQuantumPhysicist (approved):
I'm assuming you copied our code form the feat/accounting branch.

azarovh:
Correct. This is a part of changes from feat/accounting that is complete and ready to be merged.

accounting/src/delta/combine.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+
+#[cfg(test)]
+pub mod test {
+    use super::*;
+    use common::primitives::{amount::UnsignedIntType, signed_amount::SignedIntType};
+
+    #[test]
+    #[rustfmt::skip]
```
stanislav-tkach, line 73: Just curious: formatting is disabled here because rustfmt will split assertions into several lines? Is that bad? Can only that rule be overridden locally?

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+            })
+            .collect::<Result<BTreeMap<_, _>, _>>()?;
+
+        // TODO: maybe we don't have to run Collect::<_>() twice, but dealing with Result<Option<A>> is tricky in a functional way
+        let data_undo = data_undo
+            .into_iter()
+            .filter_map(|(k, v)| v.map(|v| (k, v)))
+            .collect::<BTreeMap<_, _>>();
+
+        Ok(DeltaDataUndoCollection::new(data_undo))
```
stanislav-tkach, line 66: I suppose the following code can be used to avoid using `collect` twice, but I guess that it will be less clear.
```rust
let data_undo = delta_to_apply
    .data
    .into_iter()
    .filter_map(|(key, other_pool_data)| {
        match self.merge_delta_data_element(key, other_pool_data).map(|v| (key, v)) {
            Err(e) => Some(Err(e)),
            Ok((k, v)) => Ok(v.map(|v| (k, v))).transpose(),
        }
    })
    .collect::<Result<BTreeMap<_, _>, _>>()?;
```

stanislav-tkach (approved):

accounting/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+common = {path = '../common'}
+serialization = {path = "../serialization"}
+
+thiserror = "1.0"
+parity-scale-codec = "3.1"
```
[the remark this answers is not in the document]
iljakuklic: Sadly, it is if you use `#[derive(Encode, Decode)]`. The generated code refers to `parity_scale_codec::*`

iljakuklic (approved):

PR #411, opened 2022-09-09
Iteration over storage items with given key prefix

iljakuklic:
* Backend and frontend interfaces
* In-memory implementation
* Tests
* Some refactoring, mostly just moving stuff into submodules and similar

### DIFF
--- a/storage/backend-test-suite/src/basic.rs
+++ b/storage/backend-test-suite/src/basic.rs
@@ -87,9 +87,145 @@ fn put_twice_then_commit_read_last<B: Backend>(backend: B) {
     assert_eq!(dbtx.get(IDX.0, b"hello"), Ok(Some(b"b".as_ref())),);
 }
 
+fn put_iterator_count_matches<B: Backend>(backend: B) {
+    let store = backend.open(desc(1)).expect("db open to succeed");
+
+    let mut dbtx = store.transaction_rw();
+    dbtx.put(IDX.0, vec![0x00], vec![]).unwrap();
+    dbtx.put(IDX.0, vec![0x01], vec![]).unwrap();
+    dbtx.put(IDX.0, vec![0x02], vec![]).unwrap();
+    dbtx.put(IDX.0, vec![0x03], vec![]).unwrap();
+    dbtx.commit().expect("commit to succeed");
+
+    let dbtx = store.transaction_ro();
+    assert_eq!(dbtx.prefix_iter(IDX.0, vec![]).unwrap().count(), 4);
+}
+
+fn put_and_iterate_over_prefixes<B: Backend>(backend: B) {
+    let store = backend.open(desc(1)).expect("db open to succeed");
+
+    // Populate the database with some values
+    let mut dbtx = store.transaction_rw();
+    dbtx.put(IDX.0, b"ac".to_vec(), b"2".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"bf".to_vec(), b"7".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"ab".to_vec(), b"1".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"aca".to_vec(), b"3".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"bz".to_vec(), b"8".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"x".to_vec(), b"9".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"bb".to_vec(), b"6".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"b".to_vec(), b"5".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"acb".to_vec(), b"4".to_vec()).unwrap();
+    dbtx.put(IDX.0, b"aa".to_vec(), b"0".to_vec()).unwrap();
+    dbtx.commit().expect("commit to succeed");
+
+    // Check for a non-existent prefix
+    assert_eq!(
+        store.transaction_ro().prefix_iter(IDX.0, b"foo".to_vec()).unwrap().next(),
+        None,
+    );
+
+    // Check for items that are supposed to be present
+    let check = |range: std::ops::RangeInclusive<usize>, prefix: Data| {
+        let dbtx = store.transaction_ro();
+        let vals = dbtx.prefix_iter(IDX.0, prefix).unwrap().map(|(_key, val)| val);
+        let expected_vals = range.map(|x| Data::from(x.to_string()));
+        assert!(vals.eq(expected_vals));
+        drop(dbtx);
+    };
+
+    check(0..=9, b"".to_vec());
+    check(0..=4, b"a".to_vec());
+    check(0..=0, b"aa".to_vec());
+    check(2..=4, b"ac".to_vec());
+    check(5..=8, b"b".to_vec());
+    check(9..=9, b"x".to_vec());
+}
+
+// Check for items that are supposed to be present
+fn check_prefix<Tx: ReadOps>(dbtx: &Tx, prefix: Data, expected: &[(&str, &str)]) {
+    let entries = dbtx.prefix_iter(IDX.0, prefix).unwrap();
+    let expected = expected
+        .iter()
+        .map(|(x, y)| (Data::from(x.to_string()), Data::from(y.to_string())));
+    assert!(entries.eq(expected));
+}
+
+fn put_and_iterate_delete_some<B: Backend>(backend: B) {
+    let store = backend.open(desc(1)).expect("db open to succeed");
+
+    let expected_full_0 =
+        [("aa", "0"), ("ab", "1"), ("ac", "2"), ("aca", "3"), ("acb", "4"), ("b", "5")];
+    let expected_aa_0 = [("aa", "0")];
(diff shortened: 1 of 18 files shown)


storage/backend-test-suite/src/basic.rs
```diff
@@ -87,9 +87,65 @@ fn put_twice_then_commit_read_last<B: Backend>(backend: B) {
...
+    check(5..=8, b"b".to_vec());
+    check(9..=9, b"x".to_vec());
+}
+
 tests![
     put_and_abort,
     put_and_commit,
     put_twice_then_commit_read_last,
     put_two_under_different_keys,
+    put_and_iterate_over_prefixes,
```
TheQuantumPhysicist, line 149: Quite frankly, having to manually add tests like this is playing with fire... it's very easy to forget to add them to the list and then have that false sense of security. How hard is it to create a macro that simulates the `#[test]` macro? Obviously this isn't a priority, but we should at least create an issue to investigate this.

iljakuklic: It is quite similar to forgetting `#[test]`. If you do, you get an unused function warning.

TheQuantumPhysicist: Alrighty. That's fair.

storage/core/src/adaptor/locking/mod.rs
```diff
@@ -36,7 +38,15 @@ impl<'tx, T: ReadOps> ReadOps for TxRo<'tx, T> {
...
+
+    fn prefix_iter<'m: 'i>(&'m self, idx: DbIndex, prefix: Data) -> crate::Result<Self::Iterator> {
+        self.0.prefix_iter(idx, prefix)
+    }
+}
+
+impl<'tx, T: ReadOps> backend::TxRo for TxRo<'tx, T> {}
 
 // Tracker for database changes
 type DeltaMap = BTreeMap<Data, Option<Data>>;
```
TheQuantumPhysicist, line 52: I think the correct way to do this is by having an Operation enum, just like in TransactionVerifier. The bonus you get there is that you can support caching read-operations by adding an extra arm, which you can exclude from writing back to db when you commit.

storage/core/src/adaptor/locking/prefix_iter_rw.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// How to merge the items if the keys collide
+fn merger(item: EitherOrBoth<DataPair, DataPairRef<'_>>) -> Option<(Data, Data)> {
+    match item {
+        // Item only in original db, just present it
+        EitherOrBoth::Left(l) => Some(l),
+        // If the entry is present in both database and the delta map, the delta map takes
+        // precedence. If it only is in the delta map, just take that.
+        EitherOrBoth::Right((k, v)) | EitherOrBoth::Both(_, (k, v)) => {
+            v.as_ref().map(|v| (k.to_vec(), v.clone()))
```
TheQuantumPhysicist, line 45: Are delete operations taken into account here?

iljakuklic: If a value for given key has been deleted, `v` will be `None`, so the result of the whole expression is `None`.

TheQuantumPhysicist: Alrighty

storage/backend-test-suite/src/basic.rs
```diff
@@ -87,9 +87,65 @@ fn put_twice_then_commit_read_last<B: Backend>(backend: B) {
...
+    };
+
+    check(0..=9, b"".to_vec());
+    check(0..=4, b"a".to_vec());
+    check(0..=0, b"aa".to_vec());
+    check(2..=4, b"ac".to_vec());
+    check(5..=8, b"b".to_vec());
+    check(9..=9, b"x".to_vec());
+}
+
```
TheQuantumPhysicist, line 143: Can you please add a test of iteration that starts with a set, then deletes some items of the set arbitrarily in the middle, and then iterates again and ensures only the correct items are left?

storage/core/src/adaptor/locking/mod.rs
iljakuklic: Can have a look into it but preferably not as a part of the present PR. There will have to be some thought around how to mutate the data structure.The `get()` method now takes `&self`, not  `&mut self` so some decision whether to introduce mutability into the interface or use some interior mutability mechanism would have to be made.

TheQuantumPhysicist: That's fair. I don't mind doing this in another PR. But I also want to point out that even without reads, mapping Some to Write and None to Erase is confusing. Let's fix this some times later with or without reads.

storage/inmemory/src/lib.rs
```diff
@@ -27,6 +37,22 @@ impl backend::ReadOps for StorageMaps {
     }
 }
 
+impl<'i> backend::PrefixIter<'i> for StorageMaps {
+    // TODO: Define a custom iterator type and do this without boxing
```
TheQuantumPhysicist, line 41: This TODO is probably not relevant anymore?

iljakuklic: Yes, outdated TODO. Thanks for pointing it out

storage/backend-test-suite/src/basic.rs
TheQuantumPhysicist: Btw, I meant within and outside of a transaction. 

This can be seen in multiple cases:
1. Create the data, commit, erase a subset, check, commit, check
2. Create the data, erase a subset, check, commit, check

storage/src/test.rs
```diff
@@ -0,0 +1,100 @@
...
+
+decl_schema! {
+    // Schema with a compound key
+    Compound {
+        Map2: Map<(String, u16), u64>,
+    }
+}
+
+#[test]
+fn prefix_iteration() {
```
TheQuantumPhysicist, line 51: Is there a reason this is not part of the test-suite?

TheQuantumPhysicist (approved):
Looking good. I made some minor comments.

storage/src/test.rs
iljakuklic: The test suite only depends on the backend API. Tests that use the higher-level frontend API are here.

storage/core/src/util.rs
```diff
@@ -0,0 +1,42 @@
...
+use std::collections::BTreeMap;
+
+/// Iterator over entries of a [BTreeMap] with keas starting with given prefix
+pub struct PrefixIter<'m, T> {
+    inner: std::collections::btree_map::Range<'m, Data, T>,
+    prefix: Data,
+}
+
+impl<'m, T> PrefixIter<'m, T> {
+    pub fn new(map: &'m BTreeMap<Data, T>, prefix: Data) -> Self {
```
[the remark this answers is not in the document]
iljakuklic: I did experiment with this. Here are some high-level observations:

Taking prefix by reference `&[u8]` is not quite trivial but achievable:

* The lifetime of the prefix has to be tied to the lifetime of the iterator. That is somewhat less convenient for the user because the burden of making sure the prefix data live for at least as long as the iterator is placed on the user.
* Due to how the `BTreeMap` range iterator works, an extra wrapper type with custom implementation of `std::ops::RangeBounds` is required to match the number of indirections.

Also note this is the low-level storage interface. It is accessed by the high-level interface, which operates by encoding some object and passing it into the `prefix_iter`, and returning a new iterator. For this use case, the owned value is generated anyway, and it has to be passed out of the function. The most straightforward way to do it is to include the prefix in the iterator as an owned value. Reference alone does not work, since the prefix is created inside the function but it has to outlive the function scope.

Clearly, some use cases are better served by `&[u8]` and  some by `Vec<u8>`. In general, it would make sense to parametrize the iterator over type that can give prefix slice, i.e. instead of hard-coding `Vec<u8>` or `&[u8]` make it polymorphic over `T: Borrow<[u8]>`. The `BTreeMap` range iterator does something like this. However, given that all calls ultimately go through the high-level interface which uses the owned variant, I don't see much benefit doing it at present.

storage/backend-test-suite/src/basic.rs
iljakuklic: > Btw, I meant within and outside of a transaction.
> 
> This can be seen in multiple cases:
> 
> 1. Create the data, commit, erase a subset, check, commit, check
> 2. Create the data, erase a subset, check, commit, check

These are covered by randomised property tests but will add some normal tests as well since these tend to be easier to analyse.

storage/core/src/util.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Alrighty. Let's go for it without this for now, but maybe open an issue to see the final outcome later and whether we want to pursue this based on it, and link to this comment.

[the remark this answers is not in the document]
iljakuklic: > Alrighty. Let's go for it without this for now, but maybe open an issue to see the final outcome later and whether we want to pursue this based on it, and link to this comment.

Issue #417

storage/backend-test-suite/src/basic.rs
iljakuklic: Added some more tests for deleting stuff

stanislav-tkach (approved):

PR #415, opened 2022-09-10
Check block small enhancements

azarovh:
Gathered some enhancements that I found while reading the code:

1) Added a check that tx has some inputs or outputs. We previously detected that error while connecting txs, so I moved it to earlier stage.
2) Removed redundant `consensus::validate_consensus` call from `check_block` (it is called in `check_block_header` anyway).
3) Call `check_block_size` before `check_transactions`. Checking size before iterating through all txs is a proper way to handle size constraints.

### DIFF
--- a/chainstate/test-suite/src/tests/double_spend_tests.rs
+++ b/chainstate/test-suite/src/tests/double_spend_tests.rs
@@ -29,6 +29,7 @@ use common::{
     chain::{tokens::OutputValue, OutPointSourceId, Transaction, TxInput, TxOutput},
     primitives::{Amount, Id},
 };
+use crypto::random::SliceRandom;
 
 // Process a block where the second transaction uses the first one as input.
 //
@@ -415,10 +416,12 @@ fn duplicate_tx_in_the_same_block(#[case] seed: Seed) {
         let first_tx = tx_from_genesis(&tf.genesis(), &mut rng, 1);
         let second_tx = tx_from_tx(&first_tx, rng.gen_range(1000..2000));
 
-        let block = tf
-            .make_block_builder()
-            .with_transactions(vec![first_tx, second_tx.clone(), second_tx])
-            .build();
+        let txs = vec![first_tx, second_tx];
+        let tx_duplicate = txs.choose(&mut rng).unwrap().clone();
+        let mut txs = txs.into_iter().chain(vec![tx_duplicate]).collect::<Vec<_>>();
+        txs.shuffle(&mut rng);
+
+        let block = tf.make_block_builder().with_transactions(txs).build();
         let block_id = block.get_id();
         assert_eq!(
             tf.process_block(block, BlockSource::Local).unwrap_err(),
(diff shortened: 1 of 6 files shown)


chainstate/test-suite/src/tests/double_spend_tests.rs
```diff
@@ -426,10 +424,7 @@ fn duplicate_tx_in_the_same_block(#[case] seed: Seed) {
             tf.process_block(block, BlockSource::Local).unwrap_err(),
             ChainstateError::ProcessBlockError(BlockError::CheckBlockFailed(
                 CheckBlockError::CheckTransactionFailed(
-                    CheckBlockTransactionsError::DuplicatedTransactionInBlock(
-                        second_tx_id,
-                        block_id
-                    )
+                    CheckBlockTransactionsError::DuplicateInputInBlock(block_id)
```
azarovh, line 427: At this point I'm not sure if error `DuplicatedTransactionInBlock` is reachable. Because every duplicated tx will have duplicated inputs that will be reported. The case with empty inputs (that gives the same tx id) is handled separately.

I left the check for duplicated txs untouched just as a safe net, but I'd like to know your opinions whether it is still valid.

TheQuantumPhysicist: Usually the right way to solve such problems is to write unit tests that prove whatever you're trying to prove, instead of adding blanket checks that may or may not be reached.

azarovh: That's the problem - I can't figure out the case how to create duplicated transaction that would trigger `DuplicatedTransactionInBlock ` from `check_transactions`

TheQuantumPhysicist: Well, then we can remove the duplicated transactions check and just test for it in unit tests.

chainstate/test-suite/src/tests/double_spend_tests.rs
```diff
@@ -432,6 +432,35 @@ fn duplicate_tx_in_the_same_block(#[case] seed: Seed) {
...
+        let mut tf = TestFramework::default();
+
+        let mut rng = make_seedable_rng(seed);
+        let first_tx = tx_from_genesis(&tf.genesis(), &mut rng, 1);
+        let second_tx = tx_from_tx(&first_tx, rng.gen_range(1000..2000));
+        let third_tx = tx_from_tx(&second_tx, rng.gen_range(1000..2000));
+
+        let block = tf
+            .make_block_builder()
+            .with_transactions(vec![first_tx, second_tx, third_tx.clone(), third_tx])
```
TheQuantumPhysicist, line 449: I would innovate a little more in such a test, picking a random tx out of the three, and inserting it in a random position.

azarovh: I wanted last 2 leaf of merkle tree to be identical to see how we handle this.

TheQuantumPhysicist: Or... try all possibilities over time.

stanislav-tkach (approved):

chainstate/test-suite/src/tests/double_spend_tests.rs
TheQuantumPhysicist: I'll do it and push to save time this morning.

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #416, opened 2022-09-14
Unify undo op in pos_accounting

azarovh:


stanislav-tkach (approved):

TheQuantumPhysicist (approved):

muursh (approved):

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -19,19 +19,19 @@ use serialization::{Decode, Encode};
 
 use crate::error::Error;
 
+use self::undo::*;
+
```
iljakuklic, line 23: Minor: it looks a bit weird to `use` something before it is introduced. I would personally put `use` after the corresponding `mod`

iljakuklic (approved):
Good stuff

TheQuantumPhysicist (approved):

accounting/src/delta/delta_data_collection/mod.rs
azarovh: will add with my next PR

PR #418, opened 2022-09-14
Isolate transaction verification logic from storage

TheQuantumPhysicist:
If the transaction verification logic is to be used in the mempool (since the mempool has to have transactions that are ready to be collected into a block), this means that the mempool should be able to use the chainstate handle as a backend for the transaction verifier for loading history. Currently this is not possible because the transaction verifier has direct access to the database. In this PR, I'm trying to make that isolation happen, where eventually we have a way to replace the database access with a generic implementation that may have a chainstate handle or storage. Let's see if I'm gonna come up with something.


chainstate/src/detail/transaction_verifier/error.rs
```diff
@@ -101,35 +86,60 @@ impl From<chainstate_storage::Error> for ConnectTransactionError {
...
                 tx_id,
                 source_output_index,
             },
         }
     }
 }
 
-impl From<TxMainChainIndexError> for ConnectTransactionError {
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
+pub enum TxIndexError {
```
azarovh, line 106: Moved errors from `ConnectTransactionError` to a separate enum to break circular dependency between `TransactionVerifierError` and `ConnectTransactionError`

chainstate/src/detail/transaction_verifier/flush.rs
```diff
@@ -0,0 +1,101 @@
...
+        CachedInputsOperation::Erase => storage.del_mainchain_tx_index(&tx_id)?,
+    }
+    Ok(())
+}
+
+fn flush_tokens(
+    storage: &mut impl TransactionVerifierStorageMut,
+    token_cache: &ConsumedTokenIssuanceCache,
+) -> Result<(), TransactionVerifierStorageError> {
+    assert_eq!(token_cache.data.len(), token_cache.txid_vs_tokenid.len());
```
azarovh, line 42: Now `token_cache` and `txid_vs_tokenid` are flushed in separate loops.

Also `txid_vs_tokenid` stores operations instead of plain values, because this information is required to properly implement `TransactionVerifierStorage::del_token_id`

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -652,21 +658,26 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         let block_subsidy = self.chain_config.block_subsidy_at_height(spend_height);
         tx_verifier.check_block_reward(block, Fee(total_fees), Subsidy(block_subsidy))?;
 
+        tx_verifier.set_best_block_for_utxos(block.get_id().into());
```
azarovh, line 661: This fixes a bug of storing best block for utxo - it was missed before

azarovh (commented):
Added some comments to help you guys review this PR

chainstate/src/detail/error.rs
```diff
@@ -49,6 +52,8 @@ pub enum BlockError {
     DatabaseCommitError(Id<Block>, usize, chainstate_storage::Error),
     #[error("Block proof calculation error for block: {0}")]
     BlockProofCalculationError(Id<Block>),
+    #[error("Error while flushing state to storage: {0}")]
+    FlushError(#[from] TransactionVerifierStorageError),
```
TheQuantumPhysicist, line 56: I would call this TransactionVerifierError?

TheQuantumPhysicist:
I can't approve since this is my own PR, but this looks OK even though it's incomplete. I don't want to dangle this for longer. So let's merge it and add more tests to it afterwards.

chainstate/src/detail/error.rs
azarovh: why? `FlushError` sounds more precise about what went wrong

TheQuantumPhysicist: Because flushing is an implementation detail.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -116,6 +116,18 @@ impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S> {
         }
     }
 
+    pub fn derive(&'a self) -> TransactionVerifier<'a, Self> {
```
TheQuantumPhysicist, line 119: Now that I think of this, maybe call it `derive_child`.

stanislav-tkach (approved):

chainstate/src/detail/transaction_verifier/flush.rs
TheQuantumPhysicist: Btw, please make this a debug_assert.

iljakuklic (approved):
Looks good

PR #422, opened 2022-09-15
Added a new ChainConfig property and block reward outputs maturity tests

erasmospunk:
A new property `empty_consensus_reward_maturity_distance` has been added to the ChainConfig so that it is possible to specify a non-zero block reward maturity even when the consensus is set to `None`. This is used for testing.

Added tests for 100% coverage of the
`ChainstateRef::check_block_reward_maturity_settings` failure cases.

### DIFF
--- a/chainstate/test-suite/src/tests/processing_tests.rs
+++ b/chainstate/test-suite/src/tests/processing_tests.rs
@@ -25,6 +25,7 @@ use chainstate_test_framework::{
     TransactionBuilder,
 };
 use chainstate_types::{GenBlockIndex, PropertyQueryError};
+use common::primitives::BlockDistance;
 use common::{
     chain::{
         block::{consensus_data::PoWData, timestamp::BlockTimestamp, ConsensusData},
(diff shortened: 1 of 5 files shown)


TheQuantumPhysicist (approved):

iljakuklic (approved):

stanislav-tkach:
Too late for the review, but I'm thinking if it makes sense to somehow enforce that the `empty_consensus_reward_maturity_distance` parameter is ignored if `ConsensusData` isn't equal to `None`. I suppose it would be an overkill to complicate the config, but I feel like at least an assertion can be added to the `Builder::build` method.

TheQuantumPhysicist:
> Too late for the review, but I'm thinking if it makes sense to somehow enforce that the `empty_consensus_reward_maturity_distance` parameter is ignored if `ConsensusData` isn't equal to `None`. 

It's kind of the other way around.

stanislav-tkach:
@TheQuantumPhysicist 
As far as I can see from the `reward_maturity_distance` function implementation, first we perform a match on `ConsensusData`, and the value of the `empty_consensus_reward_maturity_distance` field is only used if it is equal to `None`. Am I missing something?

chainstate/test-suite/src/tests/processing_tests.rs
```diff
@@ -46,6 +47,172 @@ use crypto::{
...
+            )])
+            .add_test_transaction(&mut rng)
+            .build();
+
+        assert!(matches!(
+            tf.process_block(block, BlockSource::Local),
+            Err(ChainstateError::ProcessBlockError(
+                BlockError::CheckBlockFailed(CheckBlockError::InvalidBlockRewardOutputType(_))
+            ))
+        ));
```
TheQuantumPhysicist, line 212: I just noticed this is missing the working reference case

TheQuantumPhysicist:
> @TheQuantumPhysicist As far as I can see from the `reward_maturity_distance` function implementation, first we perform a match on `ConsensusData`, and the value of the `empty_consensus_reward_maturity_distance` field is only used if it is equal to `None`. Am I missing something?

My point is that ConsensusData should pick the number, not the config. The config is the lowest common denominator in the software. There's no way for it to know what ConsensusData will be without resulting to some kind of double-checking and possibly circular logic. ConsensusData is basically chosen because of NetworkUpgrades that come from chain config. So what's the point in doing that twice?

stanislav-tkach:
OK, not I got it.

PR #425, opened 2022-09-15
Support merging values with deltas in pos_accounting

azarovh:


stanislav-tkach (approved):

pos_accounting/src/storage/in_memory.rs
```diff
@@ -8,13 +8,13 @@ use chainstate_types::storage_result::Error;
...
 
-#[derive(Clone, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct InMemoryPoSAccounting {
-    pool_data: BTreeMap<H256, PoolData>,
-    pool_balances: BTreeMap<H256, Amount>,
-    pool_delegation_shares: BTreeMap<(H256, H256), Amount>,
-    delegation_balances: BTreeMap<H256, Amount>,
-    delegation_data: BTreeMap<H256, DelegationData>,
+    pub pool_data: BTreeMap<H256, PoolData>,
```
TheQuantumPhysicist, line 13: Making everything public is a bad idea. Do we really need this? We have all the getters we need to avoid this.

pos_accounting/src/pool/tests/delta_tests.rs
```diff
@@ -162,8 +163,125 @@ fn check_merge_delta() {
...
+    let (_, pub_key1) = PrivateKey::new(KeyKind::RistrettoSchnorr);
+    let (_, pub_key2) = PrivateKey::new(KeyKind::RistrettoSchnorr);
+
+    let mut storage = InMemoryPoSAccounting {
+        pool_data: BTreeMap::from([(
+            H256::from_low_u64_be(1),
+            PoolData::new(pub_key1.clone(), Amount::from_atoms(100)),
+        )]),
+        pool_balances: BTreeMap::from([
+            (H256::from_low_u64_be(3), Amount::from_atoms(100)),
```
TheQuantumPhysicist, line 181: Two things:
1. Maybe some variation in values
2. You're probably making everything public to initialize like this. How about initializing with some operations and ramping up the values, testing their undos incrementally?

pos_accounting/src/pool/delta/mod.rs
```diff
@@ -22,6 +22,9 @@ pub struct PoSAccountingDelta<'a> {
 pub struct DeltaMergeUndo {
     pool_data_undo: DeltaDataUndoCollection<H256, PoolData>,
     delegation_data_undo: DeltaDataUndoCollection<H256, DelegationData>,
+    pool_balances_undo: DeltaAmountCollection<H256>,
```
TheQuantumPhysicist, line 25: I don't think we need collections to undo amounts. The delta itself is enough because you could just reverse the sign.

pos_accounting/src/pool/delta/mod.rs
```diff
@@ -84,13 +84,15 @@ impl<'a> PoSAccountingDelta<'a> {
         &mut self,
         other: PoSAccountingDeltaData,
     ) -> Result<DeltaMergeUndo, Error> {
-        self.data.pool_balances.merge_delta_amounts(other.pool_balances)?;
+        self.data.pool_balances.merge_delta_amounts(other.pool_balances.clone())?;
```
TheQuantumPhysicist, line 87: I bet these clones aren't necessary

pos_accounting/src/pool/storage/helpers.rs
```diff
@@ -0,0 +1,43 @@
+use crate::{error::Error, storage::PoSAccountingStorageWrite};
+use chainstate_types::storage_result;
+
+pub struct StorageTempAccessor<'a, S, Getter, Setter, Deleter> {
```
TheQuantumPhysicist, line 4: Please avoid using "Temp" in names. In theory, everything is temporary.

I would call this StorageAmountController, or StorageAmountWrapper.

pos_accounting/src/storage/mod.rs
```diff
@@ -34,7 +32,7 @@ pub trait PoSAccountingStorageWrite: PoSAccountingStorageRead {
 
     fn del_pool_balance(&mut self, pool_id: H256) -> Result<(), Error>;
 
-    fn set_pool_data(&mut self, pool_id: H256, pool_data: &PoolData) -> Result<(), Error>;
+    fn set_pool_data(&mut self, pool_id: H256, pool_data: PoolData) -> Result<(), Error>;
```
TheQuantumPhysicist, line 35: In theory, setters should take things by value, not by reference. But in practice, for a database, they should be by reference. The reason is that setters for databases don't really consume values. I know you're dealing with maps now for the in-memory storage, but this isn't really how this works in practice for databases. Imagine you want to store PoolData in the database with lmdb or leveldb. Basically, the first thing you'll do is that you'll have to serialize PoolData, which only uses a reference of PoolData, causing an extra unnecessary copy to be destroyed.

If you wish to make this "perfect", you can use the std::borrow::Cow type, but then the problem there is that you can't enforce one way or the other. So users can still pass things by ref when it's in memory.

Or you could support both interfaces, which again has the same issue.

So, the eventually, you'll remember that the only purpose of the that map is testing, so cloning from the ref into the map isn't such a bad idea. It's the right way to do things since serialization is equivalent to cloning, in a way.

pos_accounting/src/pool/storage/helpers.rs
TheQuantumPhysicist: If you're not using this just for amounts, you can call this StorageValueController, or StorageValueWrapper

pos_accounting/src/pool/storage/mod.rs
```diff
@@ -20,7 +28,169 @@ impl<'a, S> PoSAccountingDBMut<'a, S> {
...
+    pool_data_undo: BTreeMap<H256, Option<PoolData>>,
+    delegation_data_undo: BTreeMap<H256, Option<DelegationData>>,
+}
+
 impl<'a, S: PoSAccountingStorageWrite> PoSAccountingDBMut<'a, S> {
+    pub fn merge_with_delta(
+        &mut self,
+        other: PoSAccountingDeltaData,
+    ) -> Result<DataMergeUndo, Error> {
+        let pool_data_undo = self.merge_data_generic(
```
TheQuantumPhysicist, line 41: I'm very happy we managed to get this to work. See how beautiful and clean it's!

pos_accounting/src/pool/storage/mod.rs
```diff
@@ -20,7 +28,169 @@ impl<'a, S> PoSAccountingDBMut<'a, S> {
...
+            delegation_data_undo,
+        })
+    }
+
+    pub fn undo_merge_with_delta(
+        &mut self,
+        other: PoSAccountingDeltaData,
+        undo: DataMergeUndo,
+    ) -> Result<(), Error> {
+        undo.pool_data_undo
```
TheQuantumPhysicist, line 89: You can also create a generic function for the undo, just like merge_data_generic.

pos_accounting/src/pool/storage/mod.rs
```diff
@@ -20,7 +28,169 @@ impl<'a, S> PoSAccountingDBMut<'a, S> {
...
+
+        undo.delegation_data_undo
+            .into_iter()
+            .try_for_each(|(key, undo_data)| match undo_data {
+                Some(data) => self.store.set_delegation_data(key, data),
+                None => self.store.del_delegation_data(key),
+            })?;
+
+        self.merge_balances_generic(
+            other.pool_balances.consume().into_iter().map(|(k, v)| (k, v.neg().unwrap())),
```
TheQuantumPhysicist, line 104: unwrap()?

I bet clippy didn't leave you alone.

Please, always throw in a TODO(PR) when you want to do something temporary that has to be fixed before the PR is merged.

TheQuantumPhysicist: Also, please remember that you can use FallibleIterator to deal with iterators that can fail. See how I did that for tokens in transaction verifier for an example.

pos_accounting/src/pool/storage/mod.rs
```diff
@@ -20,7 +28,169 @@ impl<'a, S> PoSAccountingDBMut<'a, S> {
...
+            |s, id, amount| s.set_pool_balance(id, amount),
+            |s, id| s.del_pool_balance(id),
+        )?;
+
+        self.merge_balances_generic(
+            other
+                .delegation_balances
+                .consume()
+                .into_iter()
+                .map(|(k, v)| (k, v.neg().unwrap())),
```
TheQuantumPhysicist, line 115: unwrap

pos_accounting/src/pool/storage/mod.rs
```diff
@@ -20,7 +28,169 @@ impl<'a, S> PoSAccountingDBMut<'a, S> {
...
+            |s, id, amount| s.set_delegation_balance(id, amount),
+            |s, id| s.del_delegation_balance(id),
+        )?;
+
+        self.merge_balances_generic(
+            other
+                .pool_delegation_shares
+                .consume()
+                .into_iter()
+                .map(|(k, v)| (k, v.neg().unwrap())),
```
TheQuantumPhysicist, line 126: unwrap

pos_accounting/src/pool/storage/mod.rs
```diff
@@ -20,7 +28,169 @@ impl<'a, S> PoSAccountingDBMut<'a, S> {
...
+            |s, (pool_id, delegation_id), amount| {
+                s.set_pool_delegation_share(pool_id, delegation_id, amount)
+            },
+            |s, (pool_id, delegation_id)| s.del_pool_delegation_share(pool_id, delegation_id),
+        )?;
+
+        Ok(())
+    }
+
+    fn merge_balances_generic<Iter, K: Ord + Copy, Getter, Setter, Deleter>(
```
TheQuantumPhysicist, line 137: Please throw in TODOs to test merge_balances_generic and merge_data_generic independenly, since they're now perfectly abstracted.

TheQuantumPhysicist (approved):
This is almost perfect now. Please address the few issues I mentioned then feel free to merge.

Btw, clippy won't be happy with what you've done so far. There's lots of rules broken. Please fix these too.

pos_accounting/src/pool/storage/helpers.rs
azarovh: What if I rename it to `BorrowedStorage`? Imo Accessor, Wrapper or Controller are too abstract. This struct does nothing but borrows and was introduced to please borrow checker.

TheQuantumPhysicist: BorrowedStorageValue, maybe?

muursh (approved):

pos_accounting/src/pool/operations.rs
```diff
@@ -14,6 +14,7 @@ pub(crate) enum PoolDataUndo {
     DataDelta((Amount, DataDeltaUndoOp<PoolData>)),
 }
 
+#[allow(clippy::large_enum_variant)]
```
TheQuantumPhysicist, line 17: Please fix this by using boxes.

azarovh: fixed ini #426

PR #426, opened 2022-09-15
Replace H256 with PoolId/DelegationId

azarovh:


pos_accounting/src/lib.rs
```diff
@@ -1,3 +1,24 @@
+use common::primitives::H256;
+use serialization::{Decode, Encode};
+
 pub mod error;
 pub mod pool;
 pub mod storage;
+
+#[derive(Clone, Copy, Debug, Encode, Decode, Default, Eq, PartialEq, Ord, PartialOrd)]
+pub struct PoolId(H256);
```
azarovh, line 9: not sure whether I should move these new types to `id.rs` or leave here

TheQuantumPhysicist: Perhaps the better way to do this is defining an empty struct, like Pool, and doing `Id<Pool>`. Then we can later decide what we want to attach this to. 

The problem with the current format is that you have to redefine everything we defined for `Id<>`

TheQuantumPhysicist (approved):
Looking good. Just please look into PoolId and DelegationId definitions as I mentioned in the comment

azarovh:
> Just please look into PoolId and DelegationId definitions as I mentioned in the comment

Confused what do you mean? I defined empty structs as you suggested

stanislav-tkach (approved):

TheQuantumPhysicist:
> > Just please look into PoolId and DelegationId definitions as I mentioned in the comment
> 
> Confused what do you mean? I defined empty structs as you suggested

All good. I submitted the comment late.

PR #428, opened 2022-09-16
Use the utxo set to verify signatures instead of the transaction index

TheQuantumPhysicist:
This seems to have been missed before when we moved to use the utxo set. Now it's complete.

Please review this carefully. I hope I did't miss anything. It's Friday night though :-)

### DIFF
--- a/chainstate/src/detail/chainstateref.rs
+++ b/chainstate/src/detail/chainstateref.rs
@@ -168,27 +168,11 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         self.db_tx.get_block_index(block_id).map_err(PropertyQueryError::from)
     }
 
-    pub fn gen_block_index_getter(
-        db_tx: &S,
-        chain_config: &ChainConfig,
-        block_id: Id<GenBlock>,
-    ) -> Result<Option<GenBlockIndex>, PropertyQueryError> {
-        match block_id.classify(chain_config) {
-            GenBlockId::Genesis(_id) => Ok(Some(GenBlockIndex::Genesis(Arc::clone(
-                chain_config.genesis_block(),
-            )))),
-            GenBlockId::Block(id) => db_tx
-                .get_block_index(&id)
-                .map_err(PropertyQueryError::from)
-                .map(|b| b.map(GenBlockIndex::Block)),
-        }
-    }
-
     pub fn get_gen_block_index(
         &self,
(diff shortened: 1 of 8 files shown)


chainstate/src/detail/chainstateref.rs
```diff
@@ -168,18 +168,29 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         self.db_tx.get_block_index(block_id).map_err(PropertyQueryError::from)
     }
 
-    pub fn get_gen_block_index(
-        &self,
-        block_id: &Id<GenBlock>,
+    pub fn gen_block_index_getter(
```
stanislav-tkach, line 171: Should this function be public? It feels like an implementation detail.

stanislav-tkach (approved):

chainstate/src/detail/chainstateref.rs
TheQuantumPhysicist: Not really. There's a TODO related to this. Due to the usage in different contexts this may be used in other places, just like `block_index_ancestor_getter()`.

TheQuantumPhysicist: Boring Saturday. I fixed them all. I'm resolving this.

chainstate/types/src/error.rs
```diff
@@ -32,6 +32,8 @@ pub enum PropertyQueryError {
     BlockNotFound(Id<Block>),
     #[error("Previous block index not found {0}")]
     PrevBlockIndexNotFound(Id<GenBlock>),
+    #[error("Previous block index not found {0}")]
```
azarovh, line 35: wrong error description

azarovh (approved):

chainstate/types/src/error.rs
TheQuantumPhysicist: Fixed

chainstate/types/src/error.rs
```diff
@@ -32,6 +32,8 @@ pub enum PropertyQueryError {
     BlockNotFound(Id<Block>),
     #[error("Previous block index not found {0}")]
     PrevBlockIndexNotFound(Id<GenBlock>),
+    #[error("Block index at height not found {0}")]
```
muursh, line 35:
```suggestion
    #[error("Block index at height {0} not found")]
```

incredibly picky but I feel this would be neater

muursh (approved):

PR #429, opened 2022-09-19
Build for all branches and target branches

TheQuantumPhysicist:


muursh (approved):

.github/workflows/build.yml
```diff
@@ -3,12 +3,7 @@ name: build
 on:
   push:
     branches:
-      - master
```
azarovh, line 6: If I'm not mistaken after this change actions will be triggered on *every* push, no matter if the code is on review or not. Isn't this too much? I'm not sure how concerned we are about CI costs but looks like overkill.

I assume our intent is to run actions on PRs that are created from feature branches? 
https://stackoverflow.com/questions/65093647/in-github-actions-how-can-i-trigger-on-push-but-only-if-a-pr-is-active

Wouldn't the following config work more efficiently?
```
on:
  push:
    branches:
      - master
      - staging
  pull_request:
    branches:
      - '**'
```

PR #431, opened 2022-09-19
Fix spend_inputs_simple test

erasmospunk:
The spend_inputs_simple test was incorrectly checking the spent state of the UTXO set after a new block had been processed.


TheQuantumPhysicist (approved):
Good find!

stanislav-tkach (approved):

muursh (approved):

PR #432, opened 2022-09-20
Nft issuance

sinitcin:
Added simple NFT, without multi-token, approval, payouts features. 

However, I left a few `FIXME` in the code, because it requires discussion. Please write your thoughts.


chainstate/test-suite/src/tests/nft_transfer.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn nft_transfer_test(#[case] seed: Seed) {
+    utils::concurrency::model(move || {
+        let mut tf = TestFramework::default();
+        let mut rng = make_seedable_rng(seed);
+        // To have possibility to send exceed tokens amount than we have, let's limit the max issuance tokens amount
```
muursh, line 57: What does this mean?

sinitcin: It's still developing and some code just a copy of the fungible tokens. The PR is still in draft.

common/src/chain/tokens/mod.rs
```diff
@@ -74,36 +79,46 @@ impl OutputValue {
...
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
+pub struct TokenBurnV1 {
+    pub token_id: TokenId,
+    pub amount_to_burn: Amount,
+}
+
+//FIXME(nft_issuance): The same issue above in OutputValue.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
+#[allow(clippy::large_enum_variant)]
```
TheQuantumPhysicist, line 103: No. This is wrong. 

Use a Box on the large variants.

common/src/chain/tokens/mod.rs
```diff
@@ -52,8 +54,11 @@ impl TokenAuxiliaryData {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
+#[allow(clippy::large_enum_variant)]
```
TheQuantumPhysicist, line 57: Also no. Use a Box.

chainstate/src/detail/tokens/check_utils.rs
```diff
@@ -0,0 +1,65 @@
...
+        .chars()
+        .all(char::is_alphanumeric);
+
+    ensure!(is_alphanumeric, CheckTokensError::InvalidCharacter);
+    Ok(())
+}
+
+pub fn is_rfc1738_valid_symbol(ch: char) -> bool {
+    // RFC 1738 alphabet
+    String::from(":._-~!/?#[]@$&\'()*+,;=")
```
TheQuantumPhysicist, line 43: This is very inefficient. You're constructing a new string everytime you call this function. The right way is to use a static str. Just remove the String::from.

chainstate/src/detail/tokens/check_utils.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    ensure!(
+        is_validated && (uri.len() <= chain_config.token_max_uri_len()) && !uri.is_empty(),
+        CheckTokensError::InvalidURI
+    );
+    Ok(())
+}
+
+pub fn check_media_hash(_hash: &[u8]) -> Result<(), TokensError> {
+    //FIXME(nft_issuance): For now, we can use arbitrary hashes, but should we check them?
```
TheQuantumPhysicist, line 63: There's nothing you can check about the hash except for its length.

TheQuantumPhysicist: Limit it to 32-bytes.

chainstate/src/detail/tokens/check_utils.rs
```diff
@@ -0,0 +1,65 @@
...
+// limitations under the License.
+
+use common::chain::ChainConfig;
+use utils::ensure;
+
+use crate::TokensError;
+
+use super::error::CheckTokensError;
+
+pub fn check_token_text_length(text: &[u8], max_len: usize) -> Result<(), CheckTokensError> {
```
TheQuantumPhysicist, line 23: Instead of passing the max_len, pass in the chain config. This function is kinda useless because you can pass any number you want.

chainstate/src/detail/tokens/check_utils.rs
TheQuantumPhysicist: 32-bytes is good enough. Hashes can be cropped and they're still secure enough.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -48,6 +59,65 @@ pub fn check_tokens_burn_data(
...
 }
 
+pub fn check_nft_issuance_data(
+    chain_config: &ChainConfig,
+    issuance: &NftIssuanceV1,
+    tx_id: Id<Transaction>,
+    source_block_id: Id<Block>,
+) -> Result<(), TokensError> {
+    // Ticker
+    check_token_text_length(
```
TheQuantumPhysicist, line 69: So you call a function that gives you a specific error if the length is more than something... and then you override that by map_err and overwriting the error with something else... why not just do the check here in the first place?! What kind of service is that function providing exactly?

chainstate/src/detail/tokens/check_utils.rs
sinitcin: This function improve readability of `check_nft_issuance_data`. As the parameter, I use `chain_config` constants. Have a look, please. Because before this, `check_nft_issuance_data` looked dirty and unclear due to many code lines that did the same for different metadata items. But if you will confirm that this is unnecessary, I remove it.

chainstate/src/detail/tokens/mod.rs
```diff
@@ -48,6 +59,65 @@ pub fn check_tokens_burn_data(
...
+        &issuance.metadata.ticker,
+        chain_config.token_max_ticker_len(),
+    )
+    .map_err(|_| TokensError::IssueErrorInvalidTickerLength(tx_id, source_block_id))?;
+
+    check_is_text_alphanumeric(&issuance.metadata.ticker).map_err(|_| {
```
TheQuantumPhysicist, line 75: Same here... you return in that function a specific error, which you override and ignore. Isn't it better to return a boolean and then convert that to an error if need be?

chainstate/src/detail/tokens/mod.rs
sinitcin: As I mentioned before, `check_nft_issuance_data` become too big. I tried to improve readability. Probably, better to make a bunch of small functions like `check_nft_something` and remove this layer. What do you think?

chainstate/src/detail/tokens/check_utils.rs
sinitcin: Fixed!

common/src/chain/tokens/mod.rs
sinitcin: Removed attribute

common/src/chain/tokens/mod.rs
sinitcin: Removed attribute

chainstate/src/detail/tokens/mod.rs
TheQuantumPhysicist: I'll resolve this. Move the discussion here: https://github.com/mintlayer/mintlayer-core/pull/432#discussion_r987678827

common/src/chain/tokens/nft.rs
```diff
@@ -0,0 +1,72 @@
...
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct Metadata {
+    // FIXME(nft_issuance): Can it be optional?
+    pub creator: Option<TokenCreator>,
+    pub name: Vec<u8>,
+    pub description: Vec<u8>,
+    pub ticker: Vec<u8>,
+    pub icon_uri: Option<Vec<u8>>,
```
TheQuantumPhysicist, line 68: These 3 should use something like this: https://github.com/mintlayer/mintlayer-core/pull/474

Once it merges, move to that.

chainstate/test-suite/src/tests/nft_issuance.rs
```diff
@@ -0,0 +1,1139 @@
...
+        let mut tf = TestFramework::default();
+        let outpoint_source_id = TestBlockInfo::from_genesis(&tf.genesis()).txns[0].0.clone();
+        let mut rng = make_seedable_rng(seed);
+
+        let max_desc_len = tf.chainstate.get_chain_config().token_max_description_len();
+        let max_name_len = tf.chainstate.get_chain_config().token_max_name_len();
+        let max_ticker_len = tf.chainstate.get_chain_config().token_max_ticker_len();
+
+        // try all possible chars for ticker and ensure everything fails except for alphanumeric chars
+        for c in u8::MIN..u8::MAX {
```
azarovh, line 158: I think `proptest` is ideal for this purpose. You just need to feed a regexp and it fill generate tickers and test the whole range.

The only possible problem would be how to init `make_seedable_rng`, but maybe we don't need it at all if we let proptest generate all the strings

chainstate/src/detail/tokens/check_utils.rs
```diff
@@ -0,0 +1,63 @@
...
+        .chars()
+        .all(char::is_alphanumeric);
+
+    ensure!(is_alphanumeric, CheckTokensError::InvalidCharacter);
+    Ok(())
+}
+
+pub fn is_rfc1738_valid_symbol(ch: char) -> bool {
+    // RFC 1738 alphabet
+    ":._-~!/?#[]@$&\'()*+,;=".chars().any(|rfc1738_ch| ch == rfc1738_ch)
```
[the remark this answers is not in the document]
azarovh: https://www.rfc-editor.org/rfc/rfc3986#section-2.1
but it's kinda obvious because % sign is required by percent encoding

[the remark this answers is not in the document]
TheQuantumPhysicist: Actually I don't mean you, Gosha. I mean that throwing a bunch of characters in the definition isn't good enough. We should have a concrete source on where we got these from.

See for example what I did for Argon2: https://github.com/mintlayer/mintlayer-core/blob/eec27d83212a961ab7e350af161a17f203db0ad1/crypto/src/kdf/argon2.rs#L67

Basically this helps in minimizing disputes on correctness and sources.

chainstate/src/detail/tokens/check_utils.rs
```diff
@@ -0,0 +1,121 @@
...
+fn check_is_text_alphanumeric(str: &[u8]) -> bool {
+    match String::from_utf8(str.to_vec()) {
+        Ok(text) => text.chars().all(char::is_alphanumeric),
+        Err(_) => false,
+    }
+}
+
+pub fn is_rfc1738_valid_symbol(ch: char) -> bool {
+    // RFC 1738 alphabet
+    ":._-~!/?#[]@$&\'()*+,;=".chars().any(|rfc1738_ch| ch == rfc1738_ch)
```
muursh, line 32: RFC 3986 removes \ as a valid character as it's "unsafe"

muursh: And we should use 3986 not 1738

test-utils/src/lib.rs
```diff
@@ -14,3 +14,33 @@
...
+    use crypto::random::distributions::{Alphanumeric, DistString};
+    if range_len.is_empty() {
+        return String::new();
+    }
+    let len = rng.gen_range(range_len);
+    Alphanumeric.sample_string(rng, len)
+}
+
+pub fn gen_text_with_non_ascii(c: u8, rng: &mut impl Rng, max_len: usize) -> Vec<u8> {
+    assert!(!c.is_ascii_alphanumeric());
```
muursh, line 31: Every single time you call this you do something like the below first 

 ```
if c.is_ascii_alphanumeric() {
    continue;
}
```

Why the redundancy?

File: chainstate/src/detail/tokens/check_utils.rs (hunk 12 above)
Comment (muursh) [reply chainstate/src/detail/tokens/check_utils.rs:32]: Then we missed it because it was valid in RFC 1738 :)

File: test-utils/src/lib.rs (hunk 13 above)
Comment (sinitcin) [reply test-utils/src/lib.rs:31]: Actually, this function was written by @TheQuantumPhysicist. And I also started use it for NFT. I guess, he wanted to protect us from the wrong use of the function.

File: chainstate/src/detail/tokens/mod.rs [hunk 14]
```diff
@@ -48,6 +59,65 @@ pub fn check_tokens_burn_data(
...
+        &issuance.metadata.description,
+        chain_config.token_max_description_len(),
+    )
+    .map_err(|_| TokensError::IssueErrorInvalidDescriptionLength(tx_id, source_block_id))?;
+
+    check_is_text_alphanumeric(&issuance.metadata.description).map_err(|_| {
+        TokensError::IssueErrorDescriptionHasNoneAlphaNumericChar(tx_id, source_block_id)
+    })?;
+
+    // Icon URL
```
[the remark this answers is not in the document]
sinitcin: Fixed

chainstate/src/detail/tokens/error.rs
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
+pub enum CheckTokensError {
+    #[error("URL is invalid")]
+    InvalidURI,
```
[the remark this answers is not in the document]
sinitcin: Fixed

common/src/chain/tokens/nft.rs
```diff
@@ -0,0 +1,72 @@
...
+    Eq,
+    PartialOrd,
+    Ord,
+    Encode,
+    Decode,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct TokenCreator {
+    pub public_key_data: Vec<u8>,
```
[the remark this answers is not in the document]
sinitcin: I changed it to `PublicKey`, but now RPC require `serde` for this type. I hope this is not a problem, because even `Id<T>` has `serde` attributes.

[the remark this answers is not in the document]
TheQuantumPhysicist: You probably should separate RPC query types from consensus types.

common/src/chain/tokens/nft.rs
sinitcin: Now, `icon_uri`, `additional_metadata_uri`, `media_uri` are the `DataOrNoVec` type

chainstate/src/detail/tokens/mod.rs
sinitcin: Move the discussion here: https://github.com/mintlayer/mintlayer-core/pull/432#discussion_r987678827

chainstate/src/detail/tokens/check_utils.rs
sinitcin: I hope, it's better now

chainstate/src/detail/tokens/check_utils.rs
sinitcin: I added length checks

sinitcin:
Gentle reminder, we have to decide how long name, description, and URI might be. Now I use:

- name = 10 bytes
- description = 100 bytes 
- URI = 1024 bytes

common/src/chain/tokens/nft.rs
```diff
@@ -0,0 +1,104 @@
...
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Encode,
+    Decode,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct Metadata {
```
TheQuantumPhysicist, line 60: You should separate this from RPC by having a special type just for RPC to avoid having this in RPC. Please create a type that's more friendly with RPC instead of implementing serde for everything.

sinitcin: Ok, I will do that

chainstate/src/detail/tokens/check_utils.rs
```diff
@@ -0,0 +1,121 @@
...
+use utils::ensure;
+
+fn check_is_text_alphanumeric(str: &[u8]) -> bool {
+    match String::from_utf8(str.to_vec()) {
+        Ok(text) => text.chars().all(char::is_alphanumeric),
+        Err(_) => false,
+    }
+}
+
+pub fn is_rfc3986_valid_symbol(ch: char) -> bool {
```
TheQuantumPhysicist, line 30: Why do we need such an invasive function to be public? This is a the kind of function you wanna hide, while you expose something that simply checks a string.

common/src/chain/tokens/nft.rs
sinitcin: I added separated types for RPC. But I have token creator field in `RPCNonFungibleTokenMetadata` that contain a `PublicKey`. I think that encoding it to `Vec<u8>` is better than adding `serde` requirements to many types. Anyway, if you have any suggestions how to do it better, let me know

chainstate/src/detail/tokens/check_utils.rs
sinitcin: As I mentioned in DM, I have a few tests that check URI is valid or not. For example: `nft_issuance_icon_uri_invalid`. In these tests, we try all possible chars for URI, and ensure everything fails except for alphanumeric chars and RFC3986 symbols. The problem is any RFC3986 valid symbol is not ASCII alphanumeric. And I exposed this function to use in tests. Not sure how we can hide the function and use it in tests too.

common/src/chain/tokens/mod.rs
```diff
@@ -54,10 +56,15 @@ impl TokenAuxiliaryData {
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
 pub enum OutputValue {
     Coin(Amount),
-    Token(TokenData),
+    Token(Box<TokenData>),
 }
 
 impl OutputValue {
+    // FIXME(nft_issuance): Maybe we can use a better name
+    pub fn new_boxed_token(token_data: TokenData) -> Self {
```
TheQuantumPhysicist, line 64: No need to call it boxed. It's an implementation detail.

common/src/chain/config/mod.rs
```diff
@@ -188,9 +208,13 @@ const MAX_BLOCK_HEADER_SIZE: usize = 1024;
 const MAX_BLOCK_TXS_SIZE: usize = 524_288;
 const MAX_BLOCK_CONTRACTS_SIZE: usize = 524_288;
 const TOKEN_MIN_ISSUANCE_FEE: Amount = Amount::from_atoms(10_000_000_000_000);
-const TOKEN_MAX_URI_LEN: usize = 1024;
 const TOKEN_MAX_DEC_COUNT: u8 = 18;
 const TOKEN_MAX_TICKER_LEN: usize = 5;
+const TOKEN_MIN_HASH_LEN: usize = 32;
+const TOKEN_MAX_HASH_LEN: usize = 4;
```
TheQuantumPhysicist, line 214: min is 32 and max is 4?

TheQuantumPhysicist: How is this not triggering tests failures? Please investigate how this is not breaking everything.

common/src/chain/tokens/mod.rs
```diff
@@ -74,36 +81,56 @@ impl OutputValue {
...
-    // NftIssuanceV1 {
-    //     data_hash: NftDataHash,
-    //     metadata_uri: Vec<u8>,
-    // },
+}
+
+impl TokenData {
+    // FIXME(nft_issuance): Maybe we can use better names
+
+    pub fn new_boxed_nft_issuance(issuance: NftIssuanceV1) -> Self {
```
TheQuantumPhysicist, line 129: Again, no need for boxed

chainstate/test-suite/src/tests/nft_utils.rs
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common::chain::tokens::TokenCreator;
+use crypto::key::{KeyKind, PrivateKey};
+
+pub fn random_creator() -> Option<TokenCreator> {
```
TheQuantumPhysicist, line 19: Why is the returned value Option?

chainstate/src/detail/tokens/mod.rs
```diff
@@ -48,6 +57,74 @@ pub fn check_tokens_burn_data(
     Ok(())
 }
 
+pub fn check_nft_issuance_data(
+    chain_config: &ChainConfig,
+    issuance: &NftIssuanceV1,
+    tx_id: Id<Transaction>,
+    source_block_id: Id<Block>,
+) -> Result<(), TokensError> {
+    check_token_ticker(
```
TheQuantumPhysicist, line 66: Does an NFT have a ticker too? That doesn't make sense.

chainstate/test-suite/src/tests/nft_issuance.rs
```diff
@@ -0,0 +1,1165 @@
...
+                ))
+            ));
+        }
+    })
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn issue_test_ticker_too_long(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 214: Please make all names homogeneous.

No need to state in the function name it's a test. It's already obvious from the macro above.

common/src/chain/config/mod.rs
sinitcin: As I mentioned in DM, this problem was on my side. After updating rust I didn't call cargo clean and for some reasons tests done fine even if the code changed. I fixed it and added a new test for NFT hash.

common/src/chain/tokens/mod.rs
sinitcin: Cool

common/src/chain/tokens/nft.rs
TheQuantumPhysicist: Vec for RPC is fine.

chainstate/test-suite/src/tests/nft_issuance.rs
```diff
@@ -0,0 +1,1288 @@
...
+            common::primitives::BlockHeight::from(1)
+        );
+
+        let block = tf.block(*block_index.block_id());
+        let outputs = &TestBlockInfo::from_block(&block).txns[0].1;
+        let issuance_output = &outputs[0];
+
+        assert_eq!(issuance_output.value(), &output_value.into());
+    })
+}
```
TheQuantumPhysicist, line 1288: You don't seem to have tests for media hash length. Since you're enforcing that, you should test it.

TheQuantumPhysicist (approved):
Please address the few issues mentioned then feel free to merge.

chainstate/test-suite/src/tests/nft_utils.rs
sinitcin: As we spoke, it was an abstraction problem. 
I fixed it.

chainstate/src/detail/tokens/mod.rs
sinitcin: However, the OpenSea as one of the biggest NFT market have `symbol`, `name`, `description` in metadata. This is the [link to CryptoPank metadata. ](https://api.opensea.io/asset/0xb47e3cd837dDF8e4c57F05d70Ab865de6e193BBB/722). 

I saw that blockchains use words `symbol`, `title`, or `ticker` for the short name of NFT. Probably, we can use symbol instead of ticker. But I don't think that removing it is a good choice.

chainstate/test-suite/src/tests/nft_issuance.rs
sinitcin: Have a look please, is it ok now or not?

chainstate/test-suite/src/tests/nft_issuance.rs
```diff
@@ -0,0 +1,1288 @@
...
+                            TokensError::IssueErrorIncorrectMediaURI(_, _)
+                        )
+                    ))
+                ))
+            ));
+        }
+    })
+}
+
+fn new_block_with_hash(
```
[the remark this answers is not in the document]
sinitcin: Yes, it should be `new_block_with_media_hash`. Fixed!

chainstate/test-suite/src/tests/nft_issuance.rs
sinitcin: So, I have `nft_media_hash_valid`, `nft_media_hash_too_long`, `nft_media_hash_too_short`. Shall I add checks in valid case?

PR #433, opened 2022-09-20
Enable ban tests for mocks

stanislav-tkach:


stanislav-tkach:
`MockPeerId` it is simply generated from an address, so even the same peer would be assigned a different identifier after a reconnect, therefore there is no sense to ban a peer by id. This can be fixed, but it will require to change the way how mock connections are established.

PR #434, opened 2022-09-21
Added transaction_processing_order test

erasmospunk:
The transaction_processing_order test ensures that transactions in a block appear in the order they are being spent.

### DIFF
--- a/chainstate/src/detail/chainstateref.rs
+++ b/chainstate/src/detail/chainstateref.rs
@@ -609,7 +609,6 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         )?;
         debug_assert!(reward_fees.is_none());
 
-        // TODO: add a test that checks the order in which txs are connected
         let total_fees = block.transactions().iter().enumerate().try_fold(
             Amount::from_atoms(0),
             |total, (tx_num, _)| {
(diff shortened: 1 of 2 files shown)


chainstate/src/detail/chainstateref.rs
```diff
@@ -609,7 +609,6 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks> ChainstateRef<'a, S, O> {
         )?;
         debug_assert!(reward_fees.is_none());
 
-        // TODO: add a test that checks the order in which txs are connected
```
azarovh, line 612: I believe this todo is not about the order of txs in the block, but rather the order in which we process them. So essentially some test should fail if the change `enumerate()` here to `enumerate().rev()`

Also there is the same todo in `make_cache_with_disconnected_transactions`

azarovh: hmm, never mind my comment. The test does its job for connectinig txs. But the disconnect is not covered

erasmospunk: Thanks for taking a look.

You are right about the disconnecting part. I am not sure how to approach this yet as there isn't a way to undo a block that has wrongly sorted transactions (otherwise it would not connect in the first place).

TheQuantumPhysicist (approved):
Looks good!

I don't think there's a way to do the disconnect test. We should probably drop it.

muursh (approved):

PR #435, opened 2022-09-21
Fix utxos cache ownership issue in transaction verifier to avoid manually passing utxos cache as a parameter

TheQuantumPhysicist:


utxo/src/cache.rs
```diff
@@ -36,9 +36,53 @@ pub struct ConsumedUtxoCache {
...
+pub enum UtxosViewCow<'a> {
+    Borrowed(&'a dyn UtxosView),
+    Owned(Box<dyn UtxosView + 'a>),
+}
+
+impl<'a> UtxosView for UtxosViewCow<'a> {
+    fn utxo(&self, outpoint: &OutPoint) -> Option<Utxo> {
+        match self {
+            UtxosViewCow::Borrowed(r) => r.utxo(outpoint),
+            UtxosViewCow::Owned(r) => r.utxo(outpoint),
```
stanislav-tkach, line 51: What do you think about implementing something like that in order to avoid matching in every method?
```rust
impl<'a> UtxosViewCow<'a> {
    fn as_ref(&self) -> &dyn UtxosView {
        match self {
            UtxosViewCow::Borrowed(r) => *r,
            UtxosViewCow::Owned(o) => o.as_ref(),
        }
    }
}
```

stanislav-tkach (approved):

azarovh (approved):
cool!

muursh (approved):
Lgtm

PR #438, opened 2022-09-22
Implementation of exporting of blockchain bootstrap

TheQuantumPhysicist:
With this, people can export their blockchain and import it in another node, to avoid having to specifically receive the blockchain from peers.

It's still incomplete, but the relevant TODOs are there.

### DIFF
--- a/chainstate/storage/src/lib.rs
+++ b/chainstate/storage/src/lib.rs
@@ -76,7 +76,7 @@ pub trait BlockchainStorageRead: UtxosStorageRead {
     /// Get token id by id of the creation tx
     fn get_token_id(&self, tx_id: &Id<Transaction>) -> crate::Result<Option<TokenId>>;
 
-    /// Get token id by id of the creation tx
+    /// Get block tree as height vs ids
     fn get_block_tree_by_height(&self) -> crate::Result<BTreeMap<BlockHeight, Vec<Id<Block>>>>;
 }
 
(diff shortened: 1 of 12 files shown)


chainstate/storage/src/lib.rs
```diff
@@ -71,8 +73,11 @@ pub trait BlockchainStorageRead: UtxosStorageRead {
     /// Get token creation tx
     fn get_token_aux_data(&self, token_id: &TokenId) -> crate::Result<Option<TokenAuxiliaryData>>;
 
-    // Get token id by id of the creation tx
+    /// Get token id by id of the creation tx
     fn get_token_id(&self, tx_id: &Id<Transaction>) -> crate::Result<Option<TokenId>>;
+
+    /// Get token id by id of the creation tx
+    fn get_block_tree_by_height(&self) -> crate::Result<BTreeMap<BlockHeight, Vec<Id<Block>>>>;
```
stanislav-tkach, line 80: The comment is from a different function. 🙃

stanislav-tkach (approved):

chainstate/storage/src/lib.rs
TheQuantumPhysicist: Thanks. Fixed.

azarovh (approved):

muursh (approved):
lgtm

PR #442, opened 2022-09-22
Add support for importing blocks from a bootstrap file

TheQuantumPhysicist:


chainstate/src/interface/chainstate_interface_impl.rs
```diff
@@ -340,4 +342,43 @@ impl<S: BlockchainStorage> ChainstateInterface for ChainstateInterfaceImpl<S> {
...
+    fn import_bootstrap_stream<'a>(
+        &mut self,
+        reader: std::io::BufReader<Box<dyn std::io::Read + Send + 'a>>,
+    ) -> Result<(), ChainstateError> {
+        let magic_bytes = self.chainstate.chain_config().magic_bytes().to_vec();
+
+        let mut reader = reader;
+
+        // We clone because borrowing with the closure below borrowing
+        let chainstate_config = self.chainstate.chainstate_config().clone();
```
azarovh, line 355: maybe instead of cloning the whole config just pass a single field that is actually used inside import?

also comment is confusing

chainstate/src/detail/bootstrap.rs
```diff
@@ -0,0 +1,143 @@
...
+        };
+        reader.consume(buf_len);
+    }
+
+    Ok(())
+}
+
+fn read_block_at_pos(buf: &[u8]) -> Result<Block, BootstrapError> {
+    let mut buffer = buf;
+    let block = Block::decode(&mut buffer)?;
```
azarovh, line 117: I understand that this is from 3rd-party code, but I'm curious why `decode` should require mutability?

azarovh (approved):

chainstate/src/interface/chainstate_interface_impl.rs
TheQuantumPhysicist: I'll fix the comment.

But I don't think it's worth it. The interface is better when passing the whole config. The reason I chose this is because the import operation is way more expensive than this copy operation, so it doesn't really matter.

chainstate/src/detail/bootstrap.rs
TheQuantumPhysicist: The slice itself is immutable, but we a copy of the slice to track how the decoding progresses. Imagine the function parameter to be a `const uint8_t*` with a size associated with it (something like `std::span` or `std::string_view`), and the `buffer` in line 116 to be a `uint8_t*` starting at the same starting point of `buf`. While you're decoding, you need to increment some pointer, and the `buffer` pointer is the one you increment. In rust, the second pointer is also a slice that keeps changing and reducing its size while the decoding is happening. So you have the original slice `buf`, and another shrinking slice `buffer` that tracks the position to consume the bytes while decoding.

azarovh: oh, right it's just a pointer that is copied not the data... thanks

chainstate/src/interface/chainstate_interface_impl.rs
```diff
@@ -340,4 +342,44 @@ impl<S: BlockchainStorage> ChainstateInterface for ChainstateInterfaceImpl<S> {
...
             .map_err(ChainstateError::FailedToReadProperty)
     }
+
+    fn import_bootstrap_stream<'a>(
+        &mut self,
+        reader: std::io::BufReader<Box<dyn std::io::Read + Send + 'a>>,
+    ) -> Result<(), ChainstateError> {
+        let magic_bytes = self.chainstate.chain_config().magic_bytes().to_vec();
+
+        let mut reader = reader;
```
stanislav-tkach, line 352: You prefer not to write `mut reader: ...` in the funtion signature, right?..

chainstate/src/detail/bootstrap.rs
```diff
@@ -0,0 +1,143 @@
...
+use chainstate_storage::BlockchainStorageRead;
+use chainstate_types::{BlockIndex, PropertyQueryError};
+use common::{chain::Block, primitives::id::WithId};
+use serialization::{Decode, Encode};
+
+use crate::{BlockError, ChainstateConfig};
+
+use super::{orphan_blocks::OrphanBlocks, query::ChainstateQuery};
+
+const DEFAULT_MIN_IMPORT_BUFFER_SIZE: usize = 1 << 22; // 4 MB
```
stanislav-tkach, line 27: I feel like Im missing something, but in what case do I (as a user) want to change the minimum size? I probably can use the maximum size to reduce a memory consumption or stop the import process if a huge block is encountered unexpectedly. But what benefits of specifying the custom minimal size?

stanislav-tkach (approved):

chainstate/src/config.rs
```diff
@@ -22,20 +22,37 @@ pub struct ChainstateConfig {
...
+        self
+    }
 }
 
 impl Default for ChainstateConfig {
     fn default() -> Self {
         Self {
             max_db_commit_attempts: 10,
             max_orphan_blocks: 512,
+            min_max_bootstrap_import_buffer_sizes: None,
```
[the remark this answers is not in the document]
stanislav-tkach: Oh, I get it. For now we want the config fields to stay public and it probably would be weird to have both public fields and a getter.

chainstate/src/interface/chainstate_interface_impl.rs
TheQuantumPhysicist: Not really... it's weird on so many levels, but I'm not fully decided.

chainstate/src/detail/bootstrap.rs
TheQuantumPhysicist: In no case. This is only usable for tests (both unit and functional tests).

common/src/primitives/id/mod.rs
```diff
@@ -122,6 +122,12 @@ impl<T: ?Sized> Id<T> {
     }
 }
 
+impl<T: Debug> Debug for Id<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Debug::fmt(&self.id, f)
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I think we need a separate PR because the current format of printing is quite useless and there are things to be changed here. I changed them here because this is the format that made sense while debugging. Let's have a discussion about this in the other PR.

p2p/src/pubsub.rs
```diff
@@ -140,6 +140,7 @@ where
                 FailedToInitializeChainstate(_) => 0,
                 ProcessBlockError(err) => err.ban_score(),
                 FailedToReadProperty(_) => 0,
+                chainstate::ChainstateError::BootstrapError(_) => 0,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

test/functional/test_framework/test_node.py
```diff
@@ -217,10 +217,12 @@ def start(self, extra_args=None, *, cwd=None, stdout=None, stderr=None, **kwargs
...
         subp_env = dict(os.environ, LIBC_FATAL_STDERR_="1")
 
-        self.process = subprocess.Popen(self.args + extra_args, env=subp_env, stdout=stdout, stderr=stderr, cwd=cwd, **kwargs)
+        cmd = self.args + extra_args
+        self.log.debug("Starting node with command: {}".format(" ".join(self.args + extra_args)))
+        self.process = subprocess.Popen(cmd, env=subp_env, stdout=stdout, stderr=stderr, cwd=cwd, **kwargs)
 
         self.running = True
-        self.log.debug("bitcoind started, waiting for RPC to come up")
+        self.log.debug("mintlayer-core started, waiting for RPC to come up")
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Mintlayer-core is the name of the node software. At least as far as we've planned so far and I know.

chainstate/src/config.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: It's your design ;-)

chainstate/src/detail/bootstrap.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    loop {
+        if buffer_queue.len() < min_buffer_size {
+            fill_buffer(&mut buffer_queue, file_reader, max_buffer_size)?;
+        }
+
+        // locate magic bytes to recognize the start of a block
+        let current_pos = buffer_queue
+            .windows(expected_magic_bytes.len())
+            .position(|window| window == expected_magic_bytes);
```
iljakuklic, line 78: Why do we search for magic bytes here rather than just assert magic bytes are the first thing in the stream? As far as I can see, each iteration advances the buffer so it starts with the magic bytes introducing the new block

iljakuklic (approved):
Looks OK. Would like to have the reasoning behind the magic bytes behaviour explained though.

chainstate/src/detail/bootstrap.rs
iljakuklic: > Why do we search for magic bytes here rather than just assert magic bytes are the first thing in the stream?

By *assert* here I mean check, not use the `assert` macro

chainstate/src/config.rs
```diff
@@ -22,20 +22,37 @@ pub struct ChainstateConfig {
     pub max_db_commit_attempts: usize,
     /// The maximum capacity of the orphan blocks pool.
     pub max_orphan_blocks: usize,
+    /// When importing bootstrap file, this controls the buffer sizes (min, max)
+    /// (see bootstrap import function for more information)
+    pub min_max_bootstrap_import_buffer_sizes: Option<(usize, usize)>,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I don't see it as a range, really. I find that misleading when I think about it.

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -134,4 +134,16 @@ pub trait ChainstateInterface: Send {
...
     /// Returns a list of all blocks in the block tree, including orphans. The length cannot be predicted before the call
     fn get_block_id_tree_as_list(&self) -> Result<Vec<Id<Block>>, ChainstateError>;
+
+    /// Imports a bootstrap file exported with rpc
+    fn import_bootstrap_stream<'a>(
+        &mut self,
+        reader: std::io::BufReader<Box<dyn std::io::Read + Send + 'a>>,
+    ) -> Result<(), ChainstateError>;
+
+    fn export_bootstrap_stream<'a>(
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

PR #444, opened 2022-09-22
Fix potential feerate overflow issues

### LINKED ISSUE #327: Make FeeRate arithmetic operations safe
See unresolved comments in the original [mempool PR](https://github.com/mintlayer/mintlayer-core/pull/43).

royster57:
This PR is meant to close #327. @iljakuklic and @TheQuantumPhysicist, let me know what you think.

### DIFF
--- a/mempool/src/error.rs
+++ b/mempool/src/error.rs
@@ -81,6 +81,8 @@ pub enum TxValidationError {
     RollingFeeThresholdNotMet { minimum_fee: Amount, tx_fee: Amount },
     #[error("Overflow encountered while updating ancestor fee.")]
     AncestorFeeUpdateOverflow,
+    #[error("Fee overflow")]
+    FeeOverflow,
     #[error("Transaction is a descendant of expired transaction.")]
     DescendantOfExpiredTransaction,
     #[error("Chainstate error")]
--- a/mempool/src/feerate.rs
+++ b/mempool/src/feerate.rs
@@ -29,14 +29,12 @@ lazy_static::lazy_static! {
 // dangerous arithmetic operations
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub(crate) struct FeeRate {
-    atoms_per_kb: u128,
+    atoms_per_kb: Amount,
 }
 
 impl FeeRate {
     pub(crate) fn new(atoms_per_kb: Amount) -> Self {
-        Self {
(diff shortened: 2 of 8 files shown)


mempool/src/error.rs
```diff
@@ -81,6 +81,8 @@ pub enum TxValidationError {
     RollingFeeThresholdNotMet { minimum_fee: Amount, tx_fee: Amount },
     #[error("Overflow encountered while updating ancestor fee.")]
     AncestorFeeUpdateOverflow,
+    #[error("Fee overflow")]
+    FeeOverflow,
```
stanislav-tkach, line 85: I'm just curious if we can have a more specific error (such as `AncestorFeeUpdateOverflow` above) or there are several completely different places where the new error can occur?..

stanislav-tkach (approved):

mempool/src/error.rs
royster57: The new error occurs in just one place. But calling it `FeePrecisionMultiplicationOverflow` or something of the sort seemed over the top....

mempool/src/feerate.rs
```diff
@@ -51,11 +62,12 @@ impl FeeRate {
         self.atoms_per_kb
     }
 
-    // TODO Use NonZeroUsize for divisor
-    fn div_up(dividend: u128, divisor: usize) -> u128 {
-        debug_assert!(divisor != 0);
-        let divisor = u128::try_from(divisor).expect("div_up conversion");
-        (dividend + divisor - 1) / divisor
+    fn div_up(dividend: Amount, divisor: NonZeroUsize) -> Amount {
```
TheQuantumPhysicist, line 65: Please call this `div_and_round_up` or something that indicates that it's dividing then rounding up. `div_up` is not an agreed upon naming standard that I know of.

TheQuantumPhysicist (commented):
Can you please review bitcoin's feerate.cpp and compare with your implementations? For some reason I see you're not rounding as they're doing.

mempool/src/feerate.rs
royster57: Following your other comment, I changed our implementation to follow Bitcoin Core (there was no good reason why I diverged), and the need for this function disappeared.

royster57: I also changed `FeeRate`'s underlying type from `u128` to `Amount`.

TheQuantumPhysicist (approved):
Looks good.

PR #445, opened 2022-09-22
Move the GOSSIPSUB_MAX_TRANSMIT_SIZE constant to the chain config

stanislav-tkach:


TheQuantumPhysicist:
I don't think chain config is the right place for this. This isn't part of consensus.

stanislav-tkach:
@TheQuantumPhysicist I also thought that it wasn't part of consensus, but will the network behave normally if different nodes have different values of this parameter? We currently don't consider this a misbehavior, but it still looks suspicious to me.

Should I move this to the p2p config?

stanislav-tkach:
Sam pointed out that it probably doesn't make sense because we are going to remove gossipsub and replace it with syncing anyway.

PR #448, opened 2022-09-24
Make id debug output readable

TheQuantumPhysicist:
The current debug output is cluttered and unusable for debugging. I noticed this when working on bootstrap. Changing it to make use of the inner type required creating a trait that can spit out the name of the type. It's kind of a mess, but rust, unfortunately, doesn't have a good solution to this that doesn't require manual string processing.

On the way I had to rename the tokens/utils module because it conflicted with the utils crate. Very bad naming choice fixed there.


muursh (approved):

utils/src/typename.rs
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub trait TypeName {
+    fn typename_str() -> &'static str {
+        // This implementation is good enough, though it includes the full qualifiers of a typename, which may not be ideal
+        std::any::type_name::<Self>()
```
stanislav-tkach, line 19: I'm not sure about the default implementation. Yes, it is "good enough", but I guess that we want to provide a custom implementation for any type outside of the tests.

stanislav-tkach (approved):

utils/src/typename.rs
TheQuantumPhysicist: What do you propose?

In its current form, we can choose to use the default or something else.

stanislav-tkach: Initially I thought that it almost never makes sense to use the default implementation, but now I get a slightly different idea: https://github.com/mintlayer/mintlayer-core/pull/449

common/src/primitives/id/mod.rs
```diff
@@ -66,26 +69,32 @@ impl<'de> serde::Deserialize<'de> for H256 {
...
+#[derive(PartialEq, Eq, Encode, Decode, serde::Serialize, serde::Deserialize)]
 #[serde(transparent)]
-pub struct Id<T: ?Sized> {
+pub struct Id<T> {
     id: H256,
     #[serde(skip)]
     _shadow: std::marker::PhantomData<fn() -> T>,
 }
 
+impl<T: Debug + TypeName> Debug for Id<T> {
```
iljakuklic, line 80: I believe the `Debug` bound could be dropped

common/src/primitives/id/mod.rs
```diff
@@ -66,26 +69,32 @@ impl<'de> serde::Deserialize<'de> for H256 {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, Encode, Decode, serde::Serialize, serde::Deserialize)]
+#[derive(PartialEq, Eq, Encode, Decode, serde::Serialize, serde::Deserialize)]
 #[serde(transparent)]
-pub struct Id<T: ?Sized> {
+pub struct Id<T> {
```
iljakuklic, line 74: Why has `?Sized` been removed?

utils/typename-derive/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+    let DeriveInput { ident, .. } = parse_macro_input!(input);
+    let type_name = ident.to_string();
+    let output = quote! {
+        impl TypeName for #ident {
+            fn typename_str() -> &'static str {
+                #type_name
+            }
+        }
+    };
+    output.into()
```
iljakuklic, line 31: Should generics be handled somehow? I.e. should `<Vec<Foo>>::typename_str()` be `Vec<Foo>` or just `Vec`?

utils/typename/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub use typename_derive::TypeName;
+
+/// The interface for getting a name of the type.
+///
+/// The name is used in debug output, so it shouldn't be too verbose (fully qualified) to nut
+/// clutter logs.
+pub trait TypeName {
+    /// Returns a name of the type.
+    fn typename_str() -> &'static str;
```
iljakuklic, line 24: Is there a reason this is a `fn` rather than `const`, given generics are not supported?

iljakuklic (approved):
I am fine with this. Have you researched a possible use of an external crate for the `TypeName` stuff?

common/src/primitives/id/mod.rs
TheQuantumPhysicist: I actually never understood why it's there in the first place. We have no intention to support unsized trait objects and it's actually never used.

common/src/primitives/id/mod.rs
TheQuantumPhysicist: Done

utils/typename/src/lib.rs
TheQuantumPhysicist: No reason other than that I'm basing this on the TypeName crate. I'm not sure whether it's possible to be honest given that there's some macro mess going on there. Not my biggest point of strength.

utils/typename-derive/src/lib.rs
TheQuantumPhysicist: I actually tested it and it seems to not work at all. I would prefer it to be `Vec<Foo>`. I'll have to research how to do this.

common/src/primitives/id/mod.rs
iljakuklic: It's just an unnecessary restriction. Putting the `Sized` constraint there suggests the `T` type is going to be stored or manipulated as a value somewhere within the `Id`, which it is not since it only occurs in `PhantomData`

TheQuantumPhysicist: We discussed, and agreed that there's kind of difference in the way we see this as `?Sized` can be seen as a restriction and also an extension of possible types. We'll keep it as is as it's not a big deal right now. If the need comes to change it back, we'll do it then.

utils/typename/src/lib.rs
TheQuantumPhysicist: We changed this to Cow to support generics.

iljakuklic (approved):

utils/typename-derive/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+                }
+            }
+        }
+    } else {
+        let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+        let type_name = ident.to_string();
+        quote! {
+            impl #impl_generics TypeName for #ident #ty_generics #where_clause {
+                fn typename_str() -> std::borrow::Cow<'static, str> {
+                    std::borrow::Cow::Owned(#type_name.to_owned() + "<" + T::typename_str().as_ref() + ">")
```
iljakuklic, line 44: Does this work? I don't see any `T` in scope for `T::typename_str()` to work. Also I'd expect there to be one cal to `T::typename_str()` per generic parameter, not just one for all params.

TheQuantumPhysicist: You're right. I'll look into it.

stanislav-tkach: Just checked: is only works if a structure has only one generic parameter named `T`.

stanislav-tkach: I guess it is possible to get `TokenStream` from `TypeGenerics` (`ty_generics`) and get a list of type names from this stream, but it won't be easy.

TheQuantumPhysicist: I'm doing that actually as we speak. But it needs some work.

utils/typename-derive/src/lib.rs
```diff
@@ -38,10 +39,21 @@ pub fn derive(input: TokenStream) -> TokenStream {
     } else {
         let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
         let type_name = ident.to_string();
+        #[allow(unstable_name_collisions)]
+        let gen_params = generics
+            .type_params()
+            .into_iter()
```
stanislav-tkach, line 45: I'm not sure if it is worth the effort, but probably we can return `Cow::Borrowed` if the list of generic parameters is empty and fall back to `Cow::Owned` otherwise.

TheQuantumPhysicist: Not really worth it.

utils/typename-derive/src/lib.rs
TheQuantumPhysicist: Issue is resolved.

PR #449, opened 2022-09-25
Add derive macro for the TypeName trait

stanislav-tkach:


TheQuantumPhysicist:
I know this crate and the reason I didn't use it is that it just does exactly what `std::any::typename` does. However, it seems that derive solves that problem because it just prints the token of the impl, so we don't even need my version anymore.

TheQuantumPhysicist:
I'm happy to do this myself, btw. You don't have to do it.

TheQuantumPhysicist:
I'm merging this as soon as it builds. Thanks for helping. Typing the strings manually was disturbing.

stanislav-tkach:
> I'm happy to do this myself, btw. You don't have to do it.

Initially I wanted to check how hard it would be to implement such macro and after seeing that it only takes several lines of code I decided to finish the implementation. 🙃

PR #455, opened 2022-09-27
In Mempool, follow Bitcoin Core and use `fees_with_descendants/size_with_desdendants` as descendant_score

### LINKED ISSUE #93: Implement Transaction "Score" for Txs in Mempool
The transaction _score_ is a number based on the fee (or rather fee/byte), of the transaction, as well as its unconfirmed (i.e. in-mempool) descendants. This is needed for deciding which transactions to evict when the mempool is full.

royster57:
Up until now, we've used the total fees of a transaction and all its in-mempool descendants as the "descendant score" by which we decide which transactions to keep when the mempool is full. This deviated from Bitcoin core, which uses `fees_with_descendants/size_with_descendants`. In this PR we modify the mempool to follow Bitcoin core in this regard.

Closes #93.

### DIFF
--- a/mempool/src/pool/store.rs
+++ b/mempool/src/pool/store.rs
@@ -330,7 +330,6 @@ impl Idable for TxMempoolEntry {
 
 #[derive(Debug, Eq, Clone)]
 pub(super) struct TxMempoolEntry {
-    // TODO(Roy) make members private and add getters
     tx: WithId<Transaction>,
     fee: Amount,
     parents: BTreeSet<Id<Transaction>>,
@@ -338,7 +337,7 @@ pub(super) struct TxMempoolEntry {
     count_with_descendants: usize,
     fees_with_descendants: Amount,
     size_with_descendants: usize,
-    pub(super) creation_time: Time,
+    creation_time: Time,
 }
 
 impl TxMempoolEntry {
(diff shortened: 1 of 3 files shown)


mempool/src/pool/store.rs
```diff
@@ -337,24 +348,53 @@ impl TxMempoolEntry {
...
+        .expect("nonzero tx_size")
+        .into()
+    }
+
     pub(super) fn tx_id(&self) -> Id<Transaction> {
         WithId::id(&self.tx)
     }
 
+    pub(super) fn size(&self) -> usize {
+        // TODO(Roy) this should follow Bitcoin's GetTxSize, which weighs in sigops, etc.
```
stanislav-tkach, line 390: I'm confused. 🙃 Should this comment be addressed before the review or should it be changed to a regular TODO-comment?

stanislav-tkach (approved):

mempool/src/pool/store.rs
royster57: This is a regular TODO, it follows one of the formats suggested by Sam for long-standing todos, `TODO(name)`.

mempool/src/pool/store.rs
```diff
@@ -318,15 +321,23 @@ impl MempoolStore {
     }
 }
 
+impl Idable for TxMempoolEntry {
```
TheQuantumPhysicist, line 324: This is a bad idea. I get the motivation for this, but `TxMempoolEntry` is not "Ideable". Feel free to create a method in it `get_tx_Id()`.

PR #459, opened 2022-09-28
Remove some usages of TestBlockInfo

erasmospunk:
Reduce the dependency on TestBlockInfo where it is not needed

### DIFF
--- a/chainstate/test-framework/src/framework.rs
+++ b/chainstate/test-framework/src/framework.rs
@@ -82,23 +82,18 @@ impl TestFramework {
         blocks: usize,
         rng: &mut impl Rng,
     ) -> Result<Id<GenBlock>, ChainstateError> {
-        // TODO: Instead of creating TestBlockInfo on every iteration, a proper UTXO set
-        // abstraction should be used. See https://github.com/mintlayer/mintlayer-core/issues/312
-        // for the details.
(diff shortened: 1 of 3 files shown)


iljakuklic (approved):

chainstate/test-framework/src/framework.rs
```diff
@@ -82,23 +82,19 @@ impl TestFramework {
...
-        // TODO: Instead of creating TestBlockInfo on every iteration, a proper UTXO set
-        // abstraction should be used. See https://github.com/mintlayer/mintlayer-core/issues/312
-        // for the details.
-        let mut prev_block = TestBlockInfo::from_id(&self.chainstate, *parent_block);
-
+        let mut prev_block_id = *parent_block;
         for _ in 0..blocks {
             let block = self
                 .make_block_builder()
                 // .with_transactions(vec![transaction])
```
stanislav-tkach, line 89: It isn't related to you changes, but I think that we should remove the commented line.

stanislav-tkach (approved):

chainstate/test-framework/src/framework.rs
erasmospunk: Nice catch Stanislav, thank you.

muursh (approved):

PR #461, opened 2022-09-28
Add a method to query the utxo to chainstate

erasmospunk:
Utility method to query the UTXO set.

### DIFF
--- a/chainstate/src/interface/chainstate_interface_impl.rs
+++ b/chainstate/src/interface/chainstate_interface_impl.rs
@@ -24,7 +24,7 @@ use common::chain::block::BlockReward;
 use common::chain::config::ChainConfig;
 use common::chain::tokens::OutputValue;
 use common::chain::tokens::TokenAuxiliaryData;
-use common::chain::TxInput;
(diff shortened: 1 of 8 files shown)


chainstate/src/interface/chainstate_interface_impl.rs
```diff
@@ -382,4 +382,10 @@ impl<S: BlockchainStorage> ChainstateInterface for ChainstateInterfaceImpl<S> {
         )?;
         Ok(())
     }
+
+    fn utxo(&self, outpoint: &OutPoint) -> Option<Utxo> {
```
TheQuantumPhysicist, line 386: This should be fallible

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

chainstate/src/interface/chainstate_interface_impl.rs
erasmospunk: You mean something like `Result<Option<Utxo>, ChainstateError>`? Because none of the methods called is a `Result`

iljakuklic (approved):

PR #463, opened 2022-09-29
Check the tip in the 'p2p_syncing' (functional) test

stanislav-tkach:


chainstate/src/rpc.rs
```diff
@@ -77,6 +81,11 @@ impl ChainstateRpcServer for super::ChainstateHandle {
         handle_error(self.call(move |this| this.get_block_id_from_height(&height)).await)
     }
 
+    async fn block(&self, id: Id<Block>) -> rpc::Result<Option<String>> {
```
TheQuantumPhysicist, line 84: Please call it get_block

TheQuantumPhysicist (approved):

chainstate/src/rpc.rs
stanislav-tkach: I had my doubts about the naming, but for some reason other RPC methods (`best_block_id`, `block_id_at_height`, etc.) don't have the `get_` prefix. Should I rename them too?

TheQuantumPhysicist: The other ones seem OK. "Block" alone seems to be not meaningful enough in rpc.

iljakuklic (approved):

PR #464, opened 2022-10-01
Introduce Argon2 password hashing in crypto

TheQuantumPhysicist:
And also introduce generic hashed-password encoding.

### DIFF
--- a/crypto/src/kdf/mod.rs
+++ b/crypto/src/kdf/mod.rs
@@ -1,0 +1,1 @@
+pub mod argon2;
--- a/crypto/src/kdf/mod.rs
+++ b/crypto/src/kdf/mod.rs
@@ -175,7 +175,7 @@ pub fn verify_password(
                 hashed_password.len().try_into().map_err(|_| KdfError::InvalidHashSize)?,
                 password,
             )?;
-            Ok(equality_checker.is_equal(&new_hashed_password, &hashed_password))
+            Ok(equality_checker.are_equal(&new_hashed_password, &hashed_password))
         }
     }
 }
--- a/crypto/src/kdf/mod.rs
+++ b/crypto/src/kdf/mod.rs
@@ -99,7 +99,7 @@ pub fn hash_password<R: Rng + CryptoRng>(
 pub fn verify_password(
     password: &[u8],
     previously_password_hash: KdfResult,
-    equality_test_type: SliceEqualityCheckMethod,
+    equality_checker: SliceEqualityCheckMethod,
 ) -> Result<bool, KdfError> {
     match previously_password_hash {
         KdfResult::Argon2id {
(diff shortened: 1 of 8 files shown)


crypto/src/util/eq.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use std::ops::BitXor;
+
+pub enum SliceEqualityCheckMethod {
+    Normal,
+    TimingResistant,
+}
+
+impl SliceEqualityCheckMethod {
+    pub fn timing_resistant_equal<T: Eq + BitXor + Copy>(a: &[T], b: &[T]) -> bool
```
muursh, line 24: Is there a specific reason we did this "in house" rather than use something like subtle? https://docs.rs/subtle/2.4.1/subtle/trait.ConstantTimeEq.html

This appears to be fine to me for what it's worth but given we have an off the shelf trustworthy alternative I'm tempted to suggest it.

TheQuantumPhysicist: Because every other implementation has its quirks and this is simple enough I don't want to worry about it. For example, the library you linked does a short-circuit on equality if the lengths are not equal, which basically can be considered an attack vector depending on the context.

muursh (approved):

stanislav-tkach (approved):

erasmospunk:
We would need to save the Argon2 parameters + salt to the database so that we can recreate the `hashed_password ` that would be used as a `chacha20poly1305::Key` encryption key.

crypto/src/kdf/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+    #[codec(index = 0)]
+    Argon2id {
+        #[codec(compact)]
+        m_cost_memory_size: u32,
+        #[codec(compact)]
+        t_cost_iterations: u32,
+        #[codec(compact)]
+        p_cost_parallelism: u32,
+        salt: Vec<u8>,
+        hashed_password: Vec<u8>,
```
erasmospunk, line 58: Probably we would need to remove the `hashed_password` from here.

TheQuantumPhysicist: hmmmm... you're right. I seem to have confused web authentication and wallet authentication. Let me think about it.

erasmospunk:
Is it correct to say that there are 2 use cases for this?
- RPC session authentication
- Wallet private key encryption

TheQuantumPhysicist:
> Is it correct to say that there are 2 use cases for this?
> 
> * RPC session authentication
> * Wallet private key encryption

Yes. At least that's my plan. I'm currently working on restructuring this to make it work for the wallet too.

TheQuantumPhysicist (commented):
So. I added support for both encryption/decryption in addition to the already existing client/server password storage.

TheQuantumPhysicist (commented):
Special thanks to Giannis for having noticed this mistake.

erasmospunk (commented):
Looks good

crypto/src/kdf/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+                p_cost_parallelism,
+                salt,
+                hashed_password,
+            };
+            Ok(result)
+        }
+    }
+}
+
+pub fn verify_password(
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yes.

PR #465, opened 2022-10-01
Add bash file that does all the checks locally with all the relevant commands

TheQuantumPhysicist:


stanislav-tkach (approved):
Can we somehow synchronize Clippy arguments here and in the github action? In that case we don't need to update two files if we want to add more flags.

TheQuantumPhysicist:
> Can we somehow synchronize Clippy arguments here and in the github action? In that case we don't need to update two files if we want to add more flags.

We had that idea in the daily when I proposed it. I'm aiming for something simple where one can copy parts of the file for any reason. Not worth complicating this for now until we define a problem we're trying to solve.

azarovh (approved):

PR #468, opened 2022-10-03
Unify p2p blocks processing (pubsub/sync)

stanislav-tkach:
Previously we had two separate mechanisms:
- Initial syncing was done by the `sync` module.
- After that new blocks were handled by the `pubsub` module.

The pull request unifies this logic by removing the `pubsub` module.


p2p/src/net/libp2p/service/syncing.rs
```diff
@@ -139,6 +137,10 @@ where
                 request_id,
                 error,
             }),
+            types::SyncingEvent::Announcement { .. } => {
+                todo!();
+                todo!()
```
TheQuantumPhysicist, line 142: Wow... there's twice as much to do here :smile:

p2p/src/net/mock/backend.rs
```diff
@@ -319,6 +333,10 @@ where
             .map_err(P2pError::from)
     }
 
+    async fn handle_announcement(&mut self, _announcement: Announcement) -> crate::Result<()> {
+        todo!();
```
TheQuantumPhysicist, line 337: ?

p2p/src/lib.rs
```diff
@@ -203,32 +203,13 @@ where
                     consensus_handle,
```
TheQuantumPhysicist, line 203: We're still calling this consensus for some reason... please rename it to chainstate_handle

p2p/src/net/mock/backend.rs
stanislav-tkach: The block announcement (pubsub) wasn't implemented for the mock backed, so in this pull request I have added the minimal required parts in order to pass the existing tests. I'm going to create an issue to finish this implementation.

stanislav-tkach: https://github.com/mintlayer/mintlayer-core/issues/488

TheQuantumPhysicist:
If I understand correctly, the test that wasn't working now is working. So now this is free to get merged.

TheQuantumPhysicist (approved):

PR #469, opened 2022-10-04
Introduce LogError trait for easy logging and use it in chainstate

TheQuantumPhysicist:
While the Try operator in rust makes it easy to return control to the caller, it introduces the problem that it's not easy to log errors without beefy text or calls. In this PR, I attempt to provide a solution as a compromise that'll simply log errors before returning control to the caller with a simple function call.

### DIFF
--- a/utils/src/tap_error_log.rs
+++ b/utils/src/tap_error_log.rs
@@ -6,6 +6,8 @@ where
 {
     fn log_err(self) -> Self;
     fn log_err_pfx(self, prefix: &str) -> Self;
+    fn log_warn(self) -> Self;
+    fn log_warn_pfx(self, prefix: &str) -> Self;
 }
 
 impl<T, E: Display> LogError for Result<T, E> {
@@ -18,18 +20,26 @@ impl<T, E: Display> LogError for Result<T, E> {
     }
 
     #[inline(always)]
+    fn log_warn(self) -> Self {
+        if let Err(ref err) = self {
+            logging::log::warn!("{}", err);
+        }
+        self
+    }
+
+    #[inline(always)]
     fn log_err_pfx(self, prefix: &str) -> Self {
         if let Err(ref err) = self {
             logging::log::error!("{}{}", prefix, err);
         }
         self
     }
-}
 
-pub fn log_err<T: Display>(err: &T) {
-    logging::log::error!("{}", err);
-}
-
-pub fn log_warn<T: Display>(err: &T) {
-    logging::log::error!("{}", err);
+    #[inline(always)]
+    fn log_warn_pfx(self, prefix: &str) -> Self {
+        if let Err(ref err) = self {
+            logging::log::warn!("{}{}", prefix, err);
+        }
+        self
+    }
 }
--- a/utils/src/tap_error_log.rs
+++ b/utils/src/tap_error_log.rs
@@ -1,0 +1,35 @@
+use std::fmt::Display;
+
+pub trait LogError
+where
+    Self: Sized,
+{
+    fn log_err(self) -> Self;
+    fn log_err_pfx(self, prefix: &str) -> Self;
+}
+
+impl<T, E: Display> LogError for Result<T, E> {
+    #[inline(always)]
+    fn log_err(self) -> Self {
+        if let Err(ref err) = self {
+            logging::log::error!("{}", err);
+        }
+        self
+    }
+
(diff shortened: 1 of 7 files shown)


utils/src/tap_error_log.rs
```diff
@@ -0,0 +1,60 @@
...
+where
+    Self: Sized,
+{
+    fn log_err(self) -> Self;
+    fn log_err_pfx(self, prefix: &str) -> Self;
+    fn log_warn(self) -> Self;
+    fn log_warn_pfx(self, prefix: &str) -> Self;
+}
+
+impl<T, E: Display> LogError for Result<T, E> {
```
stanislav-tkach, line 28: I think that it is better to use `Debug` instead of `Display` for an error. It can be more verbose, but it is what we usually want anyway.

utils/src/tap_error_log.rs
```diff
@@ -0,0 +1,60 @@
...
+        if let Err(ref err) = self {
+            logging::log::warn!("{}", err);
+        }
+        self
+    }
+
+    #[inline(always)]
+    fn log_err_pfx(self, prefix: &str) -> Self {
+        if let Err(ref err) = self {
+            logging::log::error!("{}{}", prefix, err);
```
stanislav-tkach, line 48: Nitpick: perhaps it makes sense to add some separator (such as a white-space or `: `) between the prefix and the error.

utils/src/tap_error_log.rs
```diff
@@ -0,0 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt::Display;
+
+pub trait LogError
+where
+    Self: Sized,
+{
+    fn log_err(self) -> Self;
```
stanislav-tkach, line 22: I'm not sure how useful it is to log errors without any context. I think that usually it would be hard to figure out what's wrong only by the "X error occurred" message.

stanislav-tkach (approved):

stanislav-tkach:
I really like this idea in general, though.

utils/src/tap_error_log.rs
TheQuantumPhysicist: Since we're using thiserror, it'll basically print the error that's defined there above every error. There's no really more information that we need unless our labels are bad.

utils/src/tap_error_log.rs
TheQuantumPhysicist: I thought about this. Let the user decide how to handle it, better than a general solution that may not fit all sizes.

utils/src/tap_error_log.rs
TheQuantumPhysicist: You have to think about this in a hierarchical form. When someone submits a block, everything will return an error, so you can see where the chain of errors started. With our design of errors with thiserror, we should have all the information we need, assuming we use `log_err()` over every returned error.

utils/src/tap_error_log.rs
stanislav-tkach: OK, let's look at the usage across our codebase, but I can hardly imagine other cases except inserting a space at the end of the prefix.

iljakuklic (approved):
+1. Looks useful.

Do you have guidelines on when to use this to avoid duplicate log entries? I can  imagine it can happen with a pattern like this:
```rust
fn foo() -> Result<_, _> { let _ = some_func.log_err()?; do_sttuff() }
// Elsewhere, possibly in a different file:
fn bar() -> Result<_, _> { let _ = foo().log_err()?; do_more_sttuff() }
```
We may end up with two log entries for the same error. (This is an issue even when logging errors manually but having an ergonomic way of doing it makes it more likely)

TheQuantumPhysicist:
> +1. Looks useful.
> 
> Do you have guidelines on when to use this to avoid duplicate log entries? I can imagine it can happen with a pattern like this:
> 
> ```rust
> fn foo() -> Result<_, _> { let _ = some_func.log_err()?; do_sttuff() }
> // Elsewhere, possibly in a different file:
> fn bar() -> Result<_, _> { let _ = foo().log_err()?; do_more_sttuff() }
> ```
> 
> We may end up with two log entries for the same error. (This is an issue even when logging errors manually but having an ergonomic way of doing it makes it more likely)

The rule is, if the error is just being passed or wrapped in the top level, use the top level to log. I guess we're gonna learn more.

PR #470, opened 2022-10-04
Collect Transactions API for Mempool

royster57:
Currently a draft since:
1. The transaction selection algorithm used here is a dummy.
2. I would like to hear what you guys think of this design.


mempool/src/pool/tests.rs
```diff
@@ -1879,3 +1879,88 @@ async fn no_empty_bags_in_indices(#[case] seed: Seed) -> anyhow::Result<()> {
...
+}
+
+impl TransactionAccumulator for TestTxAccumulator {
+    fn add_tx(&mut self, tx: SignedTransaction) {
+        self.total_size += tx.encoded_size();
+        self.txs.push(tx);
+    }
+
+    fn done(&self) -> bool {
+        self.target_size <= self.total_size
```
TheQuantumPhysicist, line 1906: This kind of condition and possible bad use for it, kind of... worries me. This can easily cause an overflow over the limit. We have to be careful how to set the condition. If the limit is 1 MB and you do this check, basically you're guaranteed with a very high probability to be over 1 MB. The probability to hit exactly 1 MB is around 1byte/average_tx_size, which is a very low probability. 

Proof to this is in your test below, where you set the limit to 1 byte, but you get 1 transaction larger than that limit.

This has to be carefully engineered to not cause this problem, where the size-limit is a hard limit that'll never be exceeded.

mempool/src/pool/tests.rs
```diff
@@ -1879,3 +1879,88 @@ async fn no_empty_bags_in_indices(#[case] seed: Seed) -> anyhow::Result<()> {
...
+            txs: Vec::new(),
+            total_size: 0,
+            target_size,
+        }
+    }
+}
+
+impl TransactionAccumulator for TestTxAccumulator {
+    fn add_tx(&mut self, tx: SignedTransaction) {
+        self.total_size += tx.encoded_size();
```
TheQuantumPhysicist, line 1901: The solution to that problem I mentioned in the other comment is to add the new size to a temporary variable, then only push if the variable doesn't exceeds the size limit, otherwise switch "done" to true. 

This, though, introduces the limitation that the mempool should be smart about adding transaction by trying to pack more at the end when the available size becomes smaller and smaller, which is an optimization problem we can worry about later. This problem can be irrelevant if the size of transactions is much smaller than the block size limit, which can be an assumption we accept, but have to acknowledge.

mempool/src/pool/tests.rs
royster57: Fixed.

mempool/src/pool.rs
```diff
@@ -790,6 +801,25 @@ where
...
+        &self,
+        mut tx_accumulator: Box<dyn TransactionAccumulator>,
+    ) -> Vec<SignedTransaction> {
+        if self.store.is_empty() {
+            return Vec::new();
+        }
+        let mut tx_iter = self.store.txs_by_descendant_score.values().flatten();
+        while !tx_accumulator.done() {
+            if let Some(next_tx_id) = tx_iter.next() {
+                let next_tx = self.store.txs_by_id.get(next_tx_id).expect("tx to exist");
```
TheQuantumPhysicist, line 814: This seems like a good place to implement an iterator for the store so that you don't have to expect like this... not a priority though. It'll become something like a multi-index container.

mempool/src/pool.rs
```diff
@@ -790,6 +801,25 @@ where
...
+    fn collect_txs(
+        &self,
+        mut tx_accumulator: Box<dyn TransactionAccumulator>,
+    ) -> Vec<SignedTransaction> {
+        if self.store.is_empty() {
+            return Vec::new();
+        }
+        let mut tx_iter = self.store.txs_by_descendant_score.values().flatten();
+        while !tx_accumulator.done() {
+            if let Some(next_tx_id) = tx_iter.next() {
```
TheQuantumPhysicist, line 813: I don't think the prefix `next_` is necessary anywhere here.

mempool/src/pool.rs
```diff
@@ -790,6 +801,25 @@ where
...
+                let next_tx = self.store.txs_by_id.get(next_tx_id).expect("tx to exist");
+                tx_accumulator.add_tx(next_tx.tx().clone());
+            } else {
```
TheQuantumPhysicist, line 816: Shouldn't we pop transactions from the mempool when they're collected?

mempool/src/pool/tests.rs
```diff
@@ -1879,3 +1879,94 @@ async fn no_empty_bags_in_indices(#[case] seed: Seed) -> anyhow::Result<()> {
...
+
+    let tx_accumulator = TestTxAccumulator::new(0);
+    let collected_txs = mempool.collect_txs(Box::new(tx_accumulator));
+    assert_eq!(collected_txs.len(), 0);
+
+    let tx_accumulator = TestTxAccumulator::new(1);
+    let collected_txs = mempool.collect_txs(Box::new(tx_accumulator));
+    assert_eq!(collected_txs.len(), 0);
+    Ok(())
+}
```
TheQuantumPhysicist, line 1972: You need a more explicit test that checks the total size of transactions and that it's less than 200000.

mempool/src/pool/tests.rs
```diff
@@ -1879,3 +1879,94 @@ async fn no_empty_bags_in_indices(#[case] seed: Seed) -> anyhow::Result<()> {
     mempool.store.assert_valid();
     Ok(())
 }
+
+struct TestTxAccumulator {
```
TheQuantumPhysicist, line 1883: You can call this DefaultTxAccumulator, and make it available for block production. This way you don't have to do it twice, but a test that needs a custom accumulator can write one.

mempool/src/pool.rs
royster57: Dropped it.

mempool/src/pool.rs
royster57: I'm not sure. Bitcoin Core drops transactions from the mempool when a new tip is connected, not when collecting txs for a block.
I can imagine the following scenario: we collect transactions for a block, try to mine it, and before we succeed another, competing block arrives over the network. In this case if we have dropped the txs from the mempool during our attempted block production, those txs will be lost forever.

mempool/src/pool/tests.rs
royster57: Done.

mempool/src/pool/tests.rs
royster57: Changed the limit to 1_000 so that we actually test that the accumulator stops before reaching the limit, and also added an explicit check that the total size of collected txs is less than the limit.

mempool/src/pool.rs
```diff
@@ -790,6 +801,25 @@ where
             .collect()
     }
 
+    fn collect_txs(
+        &self,
+        mut tx_accumulator: Box<dyn TransactionAccumulator>,
+    ) -> Vec<SignedTransaction> {
+        if self.store.is_empty() {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I'm confused. Isn't it an invariant that anything that comes from tx_iter has a corresponding tx_id in the txs_by_id container?

In what situation could we have something in txs_by_descendant_score but not in txs_by_id?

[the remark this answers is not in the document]
royster57: You're absolutely right.
What I overlooked is the fact that if the pool is empty the txs_by_descendant score iterator won't yield any elements. Removed the check since it's redundant.

mempool/src/pool.rs
royster57: Added a `TODO` for this.

PR #473, opened 2022-10-04
storage: Pass backend constructor into test functions

iljakuklic:
Pass a backend constructor function into the tests instead of an instantiated backend. This gives more flexibility in how backend is initialized in tests. Will be useful in LMDB where each test run needs a separate database file.

### DIFF
--- a/storage/backend-test-suite/src/basic.rs
+++ b/storage/backend-test-suite/src/basic.rs
@@ -17,8 +17,8 @@
 
 use crate::prelude::*;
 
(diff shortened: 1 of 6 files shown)


stanislav-tkach (approved):

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
Btw, this is what I also wanted for subsystems... it simplifies initialization by requiring the function to be Send, not the object itself.

PR #474, opened 2022-10-05
Decodable type that can be Some(vec) or None if the vec is empty

TheQuantumPhysicist:
As we focus on encoding/decoding format, it can be useful at times to have a type that doesn't uselessly consume a byte in diskspace just because of the semantic meaning we'd like to achieve. A good example of this is an `Option<Vec<u8>`. In some cases, we'd like to have an interface that enforces testing whether there's a value and treat the situation differently when it's Some or None. But encoding such an object will unnecessarily add a byte because it distinguishes between `Some(vec![])` and `None`, while in the specific application, these cases are not really different.

Note: A cooler implementation of this will have a generic in the option with a generic iterable container instead of just `Option<Vec<T>>` by using some `IntoIterator` implementor. But it's not a priority now. Example here: https://users.rust-lang.org/t/generic-iterator-with-specific-item-as-function-parameter/74731

### DIFF
--- a/serialization/src/extras/non_empty_vec.rs
+++ b/serialization/src/extras/non_empty_vec.rs
@@ -1,3 +1,18 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use serialization_core::{Decode, Encode};
 
 /// A wrapper that saves the extra byte coming from Option by assuming the Vec can never be empty
--- a/serialization/src/extras/non_empty_vec.rs
+++ b/serialization/src/extras/non_empty_vec.rs
@@ -1,0 +1,80 @@
+use serialization_core::{Decode, Encode};
+
+/// A wrapper that saves the extra byte coming from Option by assuming the Vec can never be empty
+///
+/// Description:
+/// - If the Vec has data, it encodes to just the Vec, the Option is omitted
+/// - If the Vec has no data, it encodes to None
+/// - Some(vec![]) and None are equivalent when encoded, but when decoded result in None
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct DataOrNoVec<T>(pub Option<Vec<T>>);
+
+impl<U: Encode> Encode for DataOrNoVec<U> {
+    fn encode_to<T: serialization_core::Output + ?Sized>(&self, dest: &mut T) {
+        match &self.0 {
+            Some(v) => v.encode_to(dest),
+            None => Vec::<U>::new().encode_to(dest),
+        }
+    }
+
+    fn encoded_size(&self) -> usize {
+        match &self.0 {
+            Some(v) => v.encoded_size(),
+            None => Vec::encoded_size(&Vec::<U>::new()),
+        }
+    }
+}
+
+impl<U: Decode> Decode for DataOrNoVec<U> {
+    fn decode<I: serialization_core::Input>(
+        input: &mut I,
+    ) -> Result<Self, serialization_core::Error> {
+        let v = Vec::decode(input)?;
+        if v.is_empty() {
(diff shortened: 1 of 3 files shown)


serialization/src/extras/non_empty_vec.rs
```diff
@@ -0,0 +1,95 @@
...
+use serialization_core::{Decode, Encode};
+
+/// A wrapper that saves the extra byte coming from Option by assuming the Vec can never be empty
+///
+/// Description:
+/// - If the Vec has data, it encodes to just the Vec, the Option is omitted
+/// - If the Vec has no data, it encodes to None
+/// - Some(vec![]) and None are equivalent when encoded, but when decoded result in None
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct DataOrNoVec<T>(pub Option<Vec<T>>);
```
stanislav-tkach, line 25: Is the field supposed to be public?

stanislav-tkach (approved):

serialization/src/extras/non_empty_vec.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: But it's not OptionVec ;-)

TheQuantumPhysicist: Yep. How else would you access the data in it? This is just a wrapper for encoding purposes.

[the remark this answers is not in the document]
azarovh: I would argue that, because `DataOrNo` literally means optional but in a weird way. If you imply that it is a "optimised encoder-decoder for option vector" then yes, but current name imho is confusing. Also I guess it would be used as an `Option` via `.0` in the code.
Or did I miss your point?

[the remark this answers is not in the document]
TheQuantumPhysicist: Unfortunately there's no better name that incorporates the fact that this is a "data in a vector otherwise None". If you call this OptionVec, then it sounds like an `Option<Vec>`, which it's not quite the same. It's something special. I'm all open for better naming, and believe me I renamed it like 3 times before creating the PR... but I can't figure out a better name that I like and makes sense and doesn't take 87 characters.

stanislav-tkach: Personally I prefer a method for that or an implementation of `Into` or `AsRef` traits, but it isn't that important.

TheQuantumPhysicist: That's fair. Maybe I should do that. It's just quick and dirty, to be honest, to solve an urgent problem in the NFTs PR.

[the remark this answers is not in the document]
azarovh: I won't insist on this. `OptionVec` would be my choice, but it should not block this PR

muursh (approved):
I'm good with this

iljakuklic:
The roundtrip property:
```rust
x.encode().decode() == Ok(x)
```
is violated for `vec![]`.

I think a better way would be if the constructors (i.e. `From::from` and `Decode::decode`) enforced canonical representation, i.e. changed `vec![]` into `None`. The decoding method already does that by construction. That way, every time you have `DataOrNoVec<T>`, you know it is already represented as desired (`Some(vec![])` not possible).

iljakuklic (approved):

serialization/src/extras/non_empty_vec.rs
```diff
@@ -0,0 +1,146 @@
...
+impl<U: Decode> Decode for DataOrNoVec<U> {
+    fn decode<I: serialization_core::Input>(
+        input: &mut I,
+    ) -> Result<Self, serialization_core::Error> {
+        let v = Vec::decode(input)?;
+        if v.is_empty() {
+            Ok(DataOrNoVec(None))
+        } else {
+            Ok(DataOrNoVec(Some(v)))
+        }
```
iljakuklic, line 79: I think this can now delegate to the `from` method:
```rust
Vec::decode(input).map(Self::from)
```

serialization/src/extras/non_empty_vec.rs
```diff
@@ -0,0 +1,146 @@
...
+        match val {
+            Some(ref v) => {
+                if v.is_empty() {
+                    DataOrNoVec(None)
+                } else {
+                    DataOrNoVec(val)
+                }
+            }
+            None => DataOrNoVec(None),
+        }
```
iljakuklic, line 44: I can't help but think this could be simplified somewhat but not exactly sure how.

iljakuklic (approved):

serialization/src/extras/non_empty_vec.rs
TheQuantumPhysicist: Not exactly, but I did the closest to that.

TheQuantumPhysicist: Even though it's a little less efficient, btw... because we have to match again in From

PR #475, opened 2022-10-05
Isolate transaction verification strategy

TheQuantumPhysicist:
Provide a mechanism to control how transactions in a block are connected (we call it, transaction verification strategy), especially in a way where the pre-state and post-state can be controlled without affecting the default strategy. 

In other words: Rarely does anyone need to modify the default strategy, but it can now be incorporated in another strategy that can see the state of the transaction verifier before and after the transactions are connected. This can help in studying the state of the transaction verifier and its side effects.


chainstate/src/detail/tx_verification_strategy.rs
```diff
@@ -0,0 +1,127 @@
...
+        chain_config: &'a ChainConfig,
+    ) -> Self {
+        Self {
+            block_index_handle,
+            storage_backend,
+            chain_config,
+        }
+    }
+
+    pub fn connect_block(
```
azarovh, line 50: maybe there should be `trait TransactionVerificationStartegy` so that we can implement different strategies

azarovh (commented):
tbh I already forgot what are the direct benefits of having these strategies? Could you add a description to the PR?

azarovh (approved):

PR #477, opened 2022-10-06
Separate timelock checking from signature verification

TheQuantumPhysicist:
In preparation to make it possible to verify transactions without blocks

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -325,13 +325,7 @@ impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S> {
         Ok(())
     }
 
-    fn verify_signatures<T: Transactable>(
-        &self,
-        block_index: &BlockIndex,
-        tx: &T,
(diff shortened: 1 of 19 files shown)


chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -337,17 +359,13 @@ impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S> {
             None => return Ok(()),
         };
 
-        for (input_idx, input) in inputs.iter().enumerate() {
+        for (_input_idx, input) in inputs.iter().enumerate() {
```
stanislav-tkach, line 362:
```suggestion
        for input in inputs.iter() {
```

stanislav-tkach (approved):

chainstate/tx-verifier/src/transaction_verifier/mod.rs
iljakuklic: I think `inputs.iter()` could be just `&iter` but not sure that would be an improvement

iljakuklic (approved):

PR #478, opened 2022-10-06
storage: Database dump functionality

iljakuklic:
Suggestions for improvements / changes welcome

### DIFF
--- a/chainstate/storage/src/lib.rs
+++ b/chainstate/storage/src/lib.rs
@@ -19,6 +19,7 @@ mod internal;
 mod is_transaction_seal;
 #[cfg(any(test, feature = "mock"))]
 pub mod mock;
+pub mod schema;
 
 use std::collections::BTreeMap;
 
--- a/storage/src/database/mod.rs
+++ b/storage/src/database/mod.rs
@@ -16,6 +16,7 @@
 //! High-level application-agnostic storage interface
 
 mod internal;
+pub mod raw;
 
 use internal::{EntryIterator, TxImpl};
 
@@ -48,6 +49,11 @@ impl<B: Backend, Sch: Schema> Storage<B, Sch> {
             backend: backend.open(Sch::desc_iter().collect())?,
             _schema: Default::default(),
         })
+    }
+
+    /// Dump raw database contents into a data structure
+    pub fn dump_raw(&self) -> crate::Result<raw::RawDb> {
+        raw::RawDb::from_db(self)
     }
 
     /// Start a read-only transaction
--- a/storage/src/database/mod.rs
+++ b/storage/src/database/mod.rs
@@ -52,8 +52,8 @@ impl<B: Backend, Sch: Schema> Storage<B, Sch> {
     }
 
     /// Dump raw database contents into a data structure
-    pub fn dump_raw(&self) -> crate::Result<raw::RawDb<Sch>> {
-        raw::RawDb::from_db(self)
+    pub fn dump_raw(&self) -> crate::Result<raw::StorageContents<Sch>> {
+        raw::dump_storage(self)
     }
 
     /// Start a read-only transaction
--- a/storage/src/database/mod.rs
+++ b/storage/src/database/mod.rs
@@ -52,7 +52,7 @@ impl<B: Backend, Sch: Schema> Storage<B, Sch> {
     }
 
     /// Dump raw database contents into a data structure
-    pub fn dump_raw(&self) -> crate::Result<raw::RawDb> {
+    pub fn dump_raw(&self) -> crate::Result<raw::RawDb<Sch>> {
         raw::RawDb::from_db(self)
     }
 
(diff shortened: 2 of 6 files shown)


storage/src/database/mod.rs
```diff
@@ -50,6 +51,11 @@ impl<B: Backend, Sch: Schema> Storage<B, Sch> {
         })
     }
 
+    /// Dump raw database contents into a data structure
+    pub fn dump_all(&self) -> crate::Result<raw::RawDb> {
```
azarovh, line 55: Could you also add this method to `chainstate-storage`? So that it can be used in test-suite

iljakuklic: I renamed the method to `dump_raw` and exposed it in `chainstate_storage::Store`. Please have a look.

stanislav-tkach (approved):

TheQuantumPhysicist (approved):
I took a quick look, and it looks fine. I'm just not sure about the type of idx. If this is stored in the map as an enum, then great. Otherwise, please let's try to achieve that. It's probably not a lot of work.

iljakuklic:
> I took a quick look, and it looks fine. I'm just not sure about the type of idx. If this is stored in the map as an enum, then great. Otherwise, please let's try to achieve that. It's probably not a lot of work.

Do you mean the index that identifies the key-value map? It's now a string. There is currently no enum and I am not sure it's worth introducing just for dumping the database. The map keys in the high-level interface are just unit-type structs.

muursh (approved):

TheQuantumPhysicist:
> > I took a quick look, and it looks fine. I'm just not sure about the type of idx. If this is stored in the map as an enum, then great. Otherwise, please let's try to achieve that. It's probably not a lot of work.
> 
> Do you mean the index that identifies the key-value map? It's now a string. There is currently no enum and I am not sure it's worth introducing just for dumping the database. The map keys in the high-level interface are just unit-type structs.

Strings are error prone and not limited and can have other issues from my limited experience. I would really appreciate doing something with an enum. A simple solution is just to create a trait, AsEnumIndex, and put a method that returns correct enum arm per struct and implement it for each unit type. Your Schema macro can easily do all that. I don't think it's much work, but it'll improve usage a lot.

iljakuklic:
> Strings are error prone and not limited and can have other issues from my limited experience. I would really appreciate doing something with an enum. A simple solution is just to create a trait, AsEnumIndex, and put a method that returns correct enum arm per struct and implement it for each unit type. Your Schema macro can easily do all that. I don't think it's much work, but it'll improve usage a lot.

Yea maybe it could be done in the macro and strings are error prone. My contention is whether it's worth introducing a specialized enum for each schema just for dumping the database. I am not sure how is this supposed to be used exactly but it feels more like debug output and I'd say it's a bit overkill going for an enum here.

stanislav-tkach (approved):
I like the idea.

TheQuantumPhysicist (approved):
Nice. Thank you. I looked briefly and it looks good. Apologies I can't do full reviews currently.

chainstate/storage/src/lib.rs
```diff
@@ -15,6 +15,7 @@
 
 //! Application-level interface for the persistent blockchain storage.
 
+pub mod db;
```
azarovh, line 18: why not `schema` for module name? looks more specific

azarovh (approved):

chainstate/storage/src/lib.rs
iljakuklic: I called it `schema` initially but then it had the same name `storage_core::schema` and it is often useful to import both, making it awkward to work with. If you have an idea for a better name or module layout that does not collide with the generic `storage_core::schema`, I will be very happy to hear about alternatives.

iljakuklic: Also I thought the names of the database maps could be changed now that the schema is in its own submodule. For exmaple the `DBBlock` map is now referred to as `db::DBBlock` (after importing `chainstate_storage::db`). This could b made more concise by calling it just `db::Block` as it would no longer collide with `chain::Block`. Thoughts/comments?

azarovh: tbh I don't see a problem with both modules being `schema`. We do the same for `error` modules. As a solution for confusing `Schema` names one can be called `ChainstateSchema`.

> This could b made more concise by calling it just db::Block as it would no longer collide with chain::Block.

we can import via `use chainstate_storage::schema::Block as DBBlock`. Though imo this rename can be more confusing than helpful.

But maybe I don't see the full picture, so treat my comment as a suggestion

iljakuklic: > tbh I don't see a problem with both modules being `schema`. We do the same for `error` modules. As a solution for confusing `Schema` names one can be called `ChainstateSchema`.

Well at the moment I import both as modules, not the individual items in the modules, so importing both is a name clash:
```rust
use storage::schema;
use chainstate_storage::schema;  // conflicts with the above
```

I think it can be resolved by renaming it locally only in modules that use both, which may be a small improvement but still not ideal:
```rust
use storage::schema;
use chainstate_storage::schema as db;
```

PR #480, opened 2022-10-10
Tx verifier homomorphic tests

azarovh:
Use `TransactionVerificationStrategy` to test homomorphic property of TransactionVerifiers.

Fix merging of values in `TxIndexCache` and block undo cache.


chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -151,8 +154,7 @@ impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S> {
                 let issuance_token_id_getter =
                     || -> Result<Option<TokenId>, ConnectTransactionError> {
                         // issuance transactions are unique, so we use them to get the token id
-                        self.storage_ref
-                            .get_token_id_from_issuance_tx(tx_id)
+                        self.get_token_id_from_issuance_tx(tx_id)
```
azarovh, line 157: this change makes it possible to issue token in one tx and transfer in another tx in the same block

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
```diff
@@ -81,9 +41,8 @@ impl Default for DefaultTransactionVerificationStrategy {
 }
 
 impl TransactionVerificationStrategy for DefaultTransactionVerificationStrategy {
-    fn connect_block<'a, H, S, M>(
+    fn connect_block<'a, H, S>(
         &self,
-        tx_verifier_maker: M,
```
TheQuantumPhysicist, line 86: Why did you remove this?

chainstate/src/detail/tx_verification_strategy/disposable_strategy.rs
```diff
@@ -0,0 +1,155 @@
...
+                let fee = tx_verifier
+                    .connect_transactable(
+                        block_index,
+                        BlockTransactableRef::Transaction(block, tx_num),
+                        &block_index.block_height(),
+                        &median_time_past,
+                    )
+                    .map_err(BlockError::StateUpdateFailed)
+                    .log_err()?;
+                let consumed_cache = tx_verifier.consume()?;
```
TheQuantumPhysicist, line 91: I see this, and all I can think of is that we can write a test that tests every possible combination and asserts that the result is the same. Then, we make that monstrosity part of all framework tests using rstest.

chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs
```diff
@@ -178,18 +168,18 @@ impl<'a, S: TransactionVerifierStorageMut> TransactionVerifierStorageMut
...
-                    undo: undo.clone(),
-                    is_fresh: true,
-                });
-                Ok(())
-            }
-        }
+        self.utxo_block_undo
+            .entry(id)
+            .and_modify(|cached_undo| {
+                cached_undo.undo.append(new_undo.clone());
```
azarovh, line 176: please pay attention to this line. I have doubts whether it is ok to call `append` to merge `BlockUndo`, because the function is called `set_undo_data` and there is no guarantee that info in `new_undo` actually corresponds to txs with later indexes.  Though this works for our code. What do you think?

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
azarovh: To implement disposable strategy I need to create different verifiers that could be modified as well as not. The function limits this to read only `S`.

azarovh: No, I'm wrong. It was the case but it was fixed by changes in chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs lline 108.
Will revert.

chainstate/test-framework/src/framework_builder.rs
```diff
@@ -35,6 +43,7 @@ pub struct TestFrameworkBuilder {
     chainstate_storage: Store,
     custom_orphan_error_hook: Option<Arc<OrphanErrorHandler>>,
     time_getter: TimeGetter,
+    tx_verification_strategy: TxVerificationStrategy,
```
TheQuantumPhysicist, line 46: I don't understand the need to make this an enum. Why not just put the object in here then use it in Chainstate, instead of using this enum as a proxy to match on when building?

azarovh: What object exactly? It depends on a strategy and I didn't want to make `TestFrameworkBuilder` generic over it.  And `TransactionVerificationStrategy` trait cannot be used for dynamic polymorphism.

azarovh: Strategy customisation is rare, but if I make it generic then every user of `TestFrameworkBuilder` would have to explicitly specify it. With enums it looks odd, but it is hidden inside the implementation.

chainstate/test-suite/src/tests/homomorphism.rs
```diff
@@ -0,0 +1,187 @@
...
+            .unwrap();
+
+        assert_eq!(storage1.dump_raw(), storage2.dump_raw());
+    });
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn tokens_homomorphism(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 118: I wonder whether we need more variations... at least throw in TODOs for that if you don't want to do that right now. I think all output types must go in here in a way, and I'm thinking this needs more thought on whether this should be done manually or in some generic form that guarantees that all new types are covered.

TheQuantumPhysicist: I'm not talking about tokens, specifically, btw.

TheQuantumPhysicist: Perhaps a connection strategy that does the merges randomly solves this problem. The one I mentioned before in my comment in disposable_strategy

chainstate/test-framework/src/framework_builder.rs
TheQuantumPhysicist: I see the convenience there, but that match... I bet there's a way to make it go away or abstract it away. I don't know, ignore it for now.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: Seems correct to me.

stanislav-tkach (approved):

PR #481, opened 2022-10-11
storage: LMDB backend

iljakuklic:
Add LMDB backend for storage. Passes backend test suite, not yet integrated with the node itself.

### DIFF
--- a/storage/lmdb/Cargo.toml
+++ b/storage/lmdb/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "storage-lmdb"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+
+[dependencies]
+storage-core = { path = '../core' }
+utils = { path = '../../utils' }
+
+lmdb-rkv = "0.14"
+
+[dev-dependencies]
+storage-backend-test-suite = { path = "../backend-test-suite" }
+rand = "0.8"
+
+[[test]]
+name = "backend"
+harness = false
--- a/storage/lmdb/Cargo.toml
+++ b/storage/lmdb/Cargo.toml
@@ -12,6 +12,7 @@ lmdb-rkv = "0.14"
 
 [dev-dependencies]
 storage-backend-test-suite = { path = "../backend-test-suite" }
+
 rand = "0.8"
 
 [[test]]
--- a/storage/lmdb/src/error.rs
+++ b/storage/lmdb/src/error.rs
@@ -1,0 +1,91 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use lmdb::Error;
+use std::io::{Error as IoError, ErrorKind};
+
+use storage_core::error::{Fatal, Recoverable};
+
+fn process<T>(err: Error, default: storage_core::Result<T>) -> storage_core::Result<T> {
+    match err {
+        // Not found is signified by succeeding with a value signifying a missing entry
+        Error::NotFound => default,
+
+        // Overwriting a key is not an error
+        Error::KeyExist => default,
+
+        // Initialization issues
+        Error::Invalid | Error::VersionMismatch => Err(Recoverable::DbInit.into()),
+
+        // Transaction failed to commit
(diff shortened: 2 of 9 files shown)


storage/lmdb/src/error.rs
```diff
@@ -0,0 +1,91 @@
...
+                ErrorKind::BrokenPipe
+                | ErrorKind::AlreadyExists
+                | ErrorKind::PermissionDenied
+                | ErrorKind::NotFound => Recoverable::Io(err.kind(), err.to_string()).into(),
+                _ => Fatal::Io(err.kind(), err.to_string()).into(),
+            })
+        }
+
+        // Make this an error for the time of being.
+        // TODO: Would be nice to handle memory map resizes automatically somehow but it looks hard.
```
TheQuantumPhysicist, line 73: Monero does that. You can see how they do it. It's not really that hard: https://github.com/monero-project/monero/blob/fc907a957078cae2dc68348886a33363848dd089/src/blockchain_db/lmdb/db_lmdb.cpp#L494

On the other hand, given that we have recoverable errors, and assuming that this error happens only for transactions, we could just retry.

storage/lmdb/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    type Iterator = PrefixIter<'i>;
+
+    fn prefix_iter<'t: 'i>(
+        &'t self,
+        idx: DbIndex,
+        prefix: Data,
+    ) -> storage_core::Result<Self::Iterator> {
+        let mut cursor = self.tx.open_ro_cursor(self.dbs[idx]).or_else(error::process_with_err)?;
+        let iter = if prefix.is_empty() {
+            cursor.iter_start()
```
TheQuantumPhysicist, line 85: I would not make that distinction... the iter_from should be good enough for all cases.

TheQuantumPhysicist: The homogeneity of this system is very important. If it turns out there's different treatment for different calls, then it's gonna be hard to figure out what's going on at the user level.

iljakuklic: > I would not make that distinction... the iter_from should be good enough for all cases.

That's what I had initially but it did not work. LMDB does not support empty keys, so it has to be done this way as far as I can tell.

TheQuantumPhysicist: Alrighty.

TheQuantumPhysicist (approved):
We're good. I guess we can do the library maintainers a service and push a fix for that cursor lifetime issue upstream.

storage/lmdb/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+storage-core = { path = '../core' }
+utils = { path = '../../utils' }
+
+lmdb-rkv = "0.14"
+
+[dev-dependencies]
+storage-backend-test-suite = { path = "../backend-test-suite" }
+rand = "0.8"
```
stanislav-tkach, line 15: Nitpick: do we want to separate our and external crates in the `dev-dependencies` section too?

stanislav-tkach (approved):

muursh (approved):
lgtm

storage/lmdb/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+use storage_core::{
+    backend::{self, TransactionalRo, TransactionalRw},
+    info::{DbDesc, MapDesc},
+    Data, DbIndex,
+};
+use utils::sync::Arc;
+
+/// Metadata about the list of databases (key-value maps)
+#[derive(Eq, PartialEq, Debug, Clone)]
+struct DbList(Vec<lmdb::Database>);
```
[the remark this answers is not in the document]
iljakuklic: What do you mean by array here? `Box<[lmdb::Database]>` or something else?

The user-facing interface enforces quite a few invariants and also guarantees the indices it passes to the backend are in range. The backend itself does not create or manipulate indices explicitly. I don't think adding strong types here has a good ratio of benefits vs. added complexity.

storage/lmdb/src/error.rs
iljakuklic: Filed an issue for this #483

storage/lmdb/src/lib.rs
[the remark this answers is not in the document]
iljakuklic: The backend does not have access to the schema as a type parameter, so it does not know the size at compilation time. It would be possible to pass it in but not sure it's worth it and I suspect it will be too rigid when implementing e.g. database migrations.

`Box<[lmdb::Database]>` is still dynamically sized, just without some of the `Vec` methods that allow you to modify the list such as `.push()`.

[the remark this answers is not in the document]
TheQuantumPhysicist: The way I solved this problem before was that the Database enum came from the common database interface. The set/get methods would take that enum, which is type-safe, then for every implementation of the database a `const fn` would do the translation to an integer, which makes it even possible to use an array of maps instead of map of maps. 

In our case, this can be a separate module or even crate if we want to maintain perfect core that doesn't know anything about the application layer. You don't have to do this right now though. You can keep this as a future thing to do since it's not specifically related to lmdb.

PR #482, opened 2022-10-12
Implement Ancestor Score Index in Mempool

royster57:
This will be used for selecting the most economically attractive entries for block production.

### DIFF
--- a/mempool/src/pool.rs
+++ b/mempool/src/pool.rs
@@ -848,6 +848,10 @@ where
         while !tx_accumulator.done() {
             if let Some(tx_id) = tx_iter.next() {
                 let next_tx = self.store.txs_by_id.get(tx_id).expect("tx to exist");
+                log::debug!(
+                    "collect_txs: next tx has ancestor score {:?}",
+                    next_tx.ancestor_score()
+                );
                 tx_accumulator.add_tx(next_tx.tx().clone());
             } else {
                 break;
--- a/mempool/src/pool.rs
+++ b/mempool/src/pool.rs
@@ -760,7 +760,7 @@ where
             log::debug!(
                 "Mempool trim: Evicting tx {} which has a descendant score of {:?} and has size {}",
                 removed.tx_id(),
-                removed.fees_with_descendants(),
+                removed.descendant_score(),
                 removed.size()
             );
             removed_fees.push(FeeRate::from_total_tx_fee(
--- a/mempool/src/pool.rs
+++ b/mempool/src/pool.rs
@@ -385,10 +385,18 @@ where
             .filter_map(|input| input.outpoint().tx_id().get_tx_id().cloned())
             .filter_map(|id| self.store.txs_by_id.contains_key(&id).then_some(id))
             .collect::<BTreeSet<_>>();
+        let ancestor_ids =
+            TxMempoolEntry::unconfirmed_ancestors_from_parents(parents.clone(), &self.store);
+        let ancestors = ancestor_ids
+            .0
+            .into_iter()
+            .map(|id| self.store.get_entry(&id).expect("ancestors to exist"))
+            .cloned()
+            .collect();
 
         let fee = self.try_get_fee(&tx).await?;
         let time = self.clock.get_time();
-        Ok(TxMempoolEntry::new(tx, fee, parents, time))
+        TxMempoolEntry::new(tx, fee, parents, ancestors, time)
     }
 
     fn get_update_minimum_mempool_fee(
--- a/mempool/src/pool.rs
+++ b/mempool/src/pool.rs
@@ -304,7 +304,6 @@ where
     pub(crate) fn get_update_min_fee_rate(&self) -> FeeRate {
         log::debug!("get_update_min_fee_rate");
         let rolling_fee_rate = *self.rolling_fee_rate.read();
-        log::debug!("after read");
         if !rolling_fee_rate.block_since_last_rolling_fee_bump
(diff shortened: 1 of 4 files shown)


mempool/src/pool.rs
```diff
@@ -385,10 +377,18 @@ where
             .filter_map(|input| input.outpoint().tx_id().get_tx_id().cloned())
             .filter_map(|id| self.store.txs_by_id.contains_key(&id).then_some(id))
             .collect::<BTreeSet<_>>();
+        let ancestor_ids =
+            TxMempoolEntry::unconfirmed_ancestors_from_parents(parents.clone(), &self.store)?;
+        let ancestors = ancestor_ids
+            .0
```
stanislav-tkach, line 383: It isn't related to your pull request, but I find this "index field access" somewhat confusing: I immediately want to check why it is a tuple and why only one field is utilized. Our `newtype` macro generates the `Deref` implementation, but `*ancestor_ids.into_iter()` breakes the type inference. Maybe we want to add a method such as `inner` to the `newtype` macro or is it only me who is frustrated in such cases?

mempool/src/pool/store.rs
```diff
@@ -77,10 +85,21 @@ pub struct MempoolStore {
     pub(super) spender_txs: BTreeMap<OutPoint, Id<Transaction>>,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub(super) enum MempoolRemovalReason {
+    Block,
+    //Conflict,
+    Expiry,
+    SizeLimit,
+    //Reorg,
```
stanislav-tkach, line 94: Should the commented lines be removed?

stanislav-tkach (approved):

mempool/src/pool/store.rs
```diff
@@ -64,6 +69,9 @@ pub struct MempoolStore {
     // rather a value that takes into account witdess and sigop data (see CTxMemPoolEntry::GetTxSize).
     pub(super) txs_by_descendant_score: BTreeMap<DescendantScore, BTreeSet<Id<Transaction>>>,
 
+    // Mempool entries sorted by ancestor score
```
TheQuantumPhysicist, line 72: Maybe it's a good place to explain what ancestor score means

mempool/src/pool.rs
royster57: Opened an issue for this.

mempool/src/pool/store.rs
```diff
@@ -77,10 +85,21 @@ pub struct MempoolStore {
     pub(super) spender_txs: BTreeMap<OutPoint, Id<Transaction>>,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub(super) enum MempoolRemovalReason {
```
TheQuantumPhysicist, line 89: I'm confused... why does this even exist?

mempool/src/pool/store.rs
royster57: We'll eventually use these as well, but I guess we can delete them for now.

mempool/src/pool/store.rs
TheQuantumPhysicist: To rephrase, who cares why transactions are removed?

royster57: Updated the documentation.

mempool/src/pool.rs
stanislav-tkach: It is just my personal opinion, so if everyone is OK with that, then I can live with that too. 🙃

mempool/src/pool/store.rs
TheQuantumPhysicist: If you wanna keep this, please provide full documentation on it why it exists and what every arm does

mempool/src/pool/store.rs
```diff
@@ -55,15 +60,18 @@ pub struct MempoolStore {
...
-    //
-    // TODO: currently, the descendant score is the sum fee of the transaction to gether with all
-    // of its descendants. If we wish to follow Bitcoin Core, we should use:
-    // max(feerate(tx, tx_with_descendants)),
-    // Where feerate is computed as fee(tx)/size(tx)
-    // Note that if we wish to follow Bitcoin Bore, "size" is not simply the encoded size, but
+    // The descendant score of an entry is defined as:
+    //  max(fee/size of entry's tx, fee/size with all descendants).
+    //  TODO if we wish to follow Bitcoin Bore, "size" is not simply the encoded size, but
     // rather a value that takes into account witdess and sigop data (see CTxMemPoolEntry::GetTxSize).
```
TheQuantumPhysicist, line 66: witness

mempool/src/pool/store.rs
royster57: I copied this enum from Bitcoin Core. They have a mechanism for notifying clients interested in knowing when a transaction was removed for a reason other than inclusion in a block. Also, the Bitcoin Core wallet has some logic for transactions removed from the mempool due to conflicting transactions in a new block.
If this enum is confusing then I think there's no need for it at the moment. Replaced it with a boolean `remove_for_block`, which is all we really need in our logic.

mempool/src/pool/store.rs
royster57: Fixed.

mempool/src/pool/store.rs
TheQuantumPhysicist: Booleans are usually a downgrade. Just some documentation helps.

royster57: Ok, restored it and added some docs.

PR #485, opened 2022-10-13
Migrate keys crypto library

erasmospunk:
This is a two commit change. Initially a new key variant (schnorrkel) is added and the public interfaces are copied and ported for the new library. Same as with the tests. The second commit removes the first key variant and renames the second key variant.

One feature that it is currently hard to port is the signatures homomorphic additions because schorrkel does not allow access to the private data of the signature struct.


crypto/src/key/rschnorr/mod.rs
```diff
@@ -37,170 +35,177 @@ pub enum RistrettoSignatureError {
...
-    ) -> bool {
-        signature.verify_challenge(self.as_native(), challenge)
+        Self::from_native(private_key.key_data.to_public())
     }
 
-    pub(crate) fn verify_message(&self, signature: &RistrettoSchnorrSignature, msg: &[u8]) -> bool {
-        let e = Blake2b32Stream::new().write(msg).finalize();
-        signature.verify_challenge(self.as_native(), &e)
+    pub(crate) fn verify_message(&self, signature: &schnorrkel::Signature, msg: &[u8]) -> bool {
+        // TODO CRYPTOGRAPHY REVIEW
```
TheQuantumPhysicist, line 198: If you're expecting this to be done in this PR, this should be TODO(PR) or FIXME. CI won't allow to merge with these present.

crypto/src/key/rschnorr/mod.rs
```diff
@@ -16,14 +16,12 @@
 mod internal;
 
 use crate::random::{CryptoRng, Rng};
-use generic_array::GenericArray;
 pub use internal::add_sigs;
 pub use internal::RistrettoSchnorrSignature;
```
TheQuantumPhysicist, line 20: Please make sure to clean these up and the tari_crypto dependency

crypto/src/key/rschnorr/mod.rs
```diff
@@ -16,14 +16,12 @@
 mod internal;
```
TheQuantumPhysicist, line 16: There's probably a misunderstanding on what this module is. This is just a portal to the tari crypto library. This has to go if you get rid of tari_crypto

crypto/src/key/rschnorr/mod.rs
```diff
@@ -37,170 +35,177 @@ pub enum RistrettoSignatureError {
...
+        self.pubkey_data.verify(ctx.bytes(msg), signature).is_ok()
     }
 }
 
 impl std::ops::Add for &MLRistrettoPublicKey {
     type Output = MLRistrettoPublicKey;
     fn add(self, rhs: Self) -> MLRistrettoPublicKey {
-        let result = self.as_native() + rhs.as_native();
-        MLRistrettoPublicKey::from_native(result)
+        let result = self.pubkey_data.as_point() + rhs.pubkey_data.as_point();
```
TheQuantumPhysicist, line 207: Is this doing proper elliptic curve addition? From my brief look, seems like it. Now I'm wondering whether verifying the sum of sigs will yield the same result as individual verifications. Can you try this? Create sig1 of msg1 with priv_key1, then sig2 with priv_key2, then verify each, then verify msg1 + msg2 with pub_key1 + pub_key2. I wonder whether that'll work. I wonder whether making hashes internally has ruined this.

crypto/src/key/mod.rs
```diff
@@ -113,11 +114,12 @@ impl PublicKey {
...
-        use crate::key::Signature::RistrettoSchnorr;
-
-        let PublicKeyHolder::RistrettoSchnorr(k) = &self.pub_key;
-        match signature {
-            RistrettoSchnorr(s) => k.verify_message(s, msg),
+        match &self.pub_key {
+            PublicKeyHolder::RistrettoSchnorr(ref k) => match signature {
+                RistrettoSchnorr(s) => k.verify_message(s, msg),
+                #[allow(unreachable_patterns)]
+                _ => panic!("Wrong key/signature combination"),
```
TheQuantumPhysicist, line 121: I think this isn't necessary anymore.

TheQuantumPhysicist (commented):
Please run a build locally before committing and include Cargo.lock changes. When I build locally, I seem to get changes in Cargo.lock.

TheQuantumPhysicist (approved):
Please mind the comments and wait for Ben to review as well. Otherwise, looks good.

crypto/src/key/mod.rs
```diff
@@ -133,7 +135,7 @@ mod test {
     use super::*;
 
     #[test]
-    fn sign_and_verify() {
+    fn sign_and_verify2() {
```
muursh, line 138: Is there a reason this has become 2?

crypto/src/key/rschnorr/mod.rs
```diff
@@ -16,14 +16,12 @@
 mod internal;
 
 use crate::random::{CryptoRng, Rng};
-use generic_array::GenericArray;
 pub use internal::add_sigs;
 pub use internal::RistrettoSchnorrSignature;
-use internal::*;
 use serialization::{Decode, Encode};
-use tari_crypto::{keys::PublicKey, tari_utilities::ByteArray};
+use zeroize::Zeroize;
```
muursh, line 22: If we're doing this manually we should probably consider doing a "proper job" here. Not for this PR though. Might be worth a todo for the future though

crypto/src/key/rschnorr/mod.rs
muursh: The addition seems fine

crypto/src/key/rschnorr/mod.rs
```diff
@@ -241,7 +245,7 @@ mod test {
     }
 
     #[test]
-    fn serialize() {
+    fn serialize2() {
```
muursh, line 248: 2 again?

muursh (commented):
Nothing major from me

iljakuklic (approved):

stanislav-tkach (approved):

muursh (approved):
Just realised I never actually approved. Apologies.

crypto/src/key/rschnorr/mod.rs
erasmospunk: Understood, I have removed the unnecessary tests.

crypto/src/key/mod.rs
erasmospunk: It was a mistake, good catch.

PR #487, opened 2022-10-14
storage: Use the Mintlayer fork of the `lmdb` crate

iljakuklic:


TheQuantumPhysicist (approved):

storage/lmdb/Cargo.toml
```diff
@@ -8,7 +8,9 @@ license = "MIT"
 storage-core = { path = '../core' }
 utils = { path = '../../utils' }
 
-lmdb-rkv = "0.14"
+[dependencies.lmdb-mintlayer]
```
TheQuantumPhysicist, line 11: Why not use braces like others?

muursh (approved):

storage/lmdb/Cargo.toml
iljakuklic: I just found the line too long but can change it back, no strong opinions here

stanislav-tkach: Personally I prefer the "single line" dependency format, but in the p2p we have a lot of dependencies in the "multiline" format.

stanislav-tkach (approved):

storage/lmdb/Cargo.toml
TheQuantumPhysicist: Would be great if you could change it... it makes global searches much easier when everything is homogeneous. I want to change them also in p2p.

iljakuklic: ok

PR #489, opened 2022-10-14
Attempt to make the transaction verifier usable for the mempool without a full block being available

TheQuantumPhysicist:


azarovh (approved):
looks good to me

stanislav-tkach (approved):

PR #491, opened 2022-10-16
Update the merkletree version in order to reduce duplicated dependencies

stanislav-tkach:


common/Cargo.toml
```diff
@@ -18,7 +18,7 @@ anyhow = "1.0"
 bech32 = "0.9"
 generic-array = "0.14"
 lazy_static = "1.4"
-merkletree = "0.22"
+merkletree = "0.22.1"
```
TheQuantumPhysicist, line 21: Instead of specifying the version manually why not run cargo update to do that for you?

stanislav-tkach: I don't wanted to change the versions of other dependencies, but I have run `cargo update` now.

TheQuantumPhysicist (approved):

muursh (approved):

PR #493, opened 2022-10-18
node: Integrate the LMDB backend

iljakuklic:
Notes
* Moved stuff around in a minor way in the node crate
* The storage option / config processing is a bit unwieldy, improvement suggestions welcome
* Functional tests
  * <s>Factored out the mined sample blocks into a separate file so they can be shared between tests</s>
  * Added a test that database state persists across node shutdown/start cycle if LMDB is used


node/src/storage_config.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum StorageBackendParseError {
+    #[error("Unrecognized backend '{name}'")]
+    Unrecognized { name: String },
+}
+
+/// Storage configuration
+#[derive(Serialize, Deserialize, Debug, Default)]
+pub struct StorageConfig {
```
TheQuantumPhysicist, line 50: I don't understand why we need a separate storage config. Storage is part of chainstate. Two things related should be there: The storage type, and the data directory. The relation between lmdb's storage path and the data directory is fixed. It's only a subdirectory of that, say, `chainstate-lmdb`. I still haven't dug deep in the PR to see how it's done here, but seeing a separate storage configuration seems redundant.

node/src/runner.rs
```diff
@@ -37,36 +40,58 @@ use crate::{
...
-            node_config.chainstate,
-            storage.clone(),
-            DefaultTransactionVerificationStrategy::new(),
-            None,
-            Default::default(),
-        )?,
-    );
+    let chainstate = match backend_type {
+        StorageBackend::Lmdb => {
+            let storage_backend = storage_lmdb::Lmdb::new(node_config.datadir);
```
TheQuantumPhysicist, line 86: Yeah, I don't think this is correct. Passing `node_config.datadir` basically means that the lmdb storage will be directly in datadir, which is not really what we need. There are other things that go into the datadir, such as the p2p noise key. We need a subdirectory for lmdb, say `chainstate-lmdb`. So something like `node_config.datadir.join("chainstate-lmdb")` is what we need.

node/src/storage_config.rs
```diff
@@ -0,0 +1,60 @@
...
+            name => {
+                let name = name.into();
+                Err(StorageBackendParseError::Unrecognized { name })
+            }
+        }
+    }
+}
+
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum StorageBackendParseError {
```
TheQuantumPhysicist, line 43: This probably doesn't need to be an enum... kind of an overkill... Maybe using `Box<dyn std::error::Error>` is good enough? I'm not sure.

node/src/runner.rs
```diff
@@ -37,36 +40,58 @@ use crate::{
...
     // INITIALIZE SUBSYSTEMS
 
     let mut manager = subsystem::Manager::new("mintlayer");
     manager.install_signal_handlers();
 
+    // Pick default storage backend based on chain type
+    let backend_type =
+        node_config.storage.backend.unwrap_or_else(|| match chain_config.chain_type() {
+            ChainType::Signet | ChainType::Mainnet | ChainType::Testnet => StorageBackend::Lmdb,
+            ChainType::Regtest => StorageBackend::InMemory,
```
TheQuantumPhysicist, line 80: Actually Regtest should be lmdb by default, as well. The only difference between Regtest and Mainnet is ChainConfig.

In-memory is only a choice to be made when necessary in Regtest.

node/Cargo.toml
```diff
@@ -27,6 +30,7 @@ serde = { version = "1", features = ["derive"] }
 toml = "0.5"
 directories = "4.0"
 paste = "1.0"
+thiserror = "1.0"
```
TheQuantumPhysicist, line 33: Well... we do have anyhow already... even `Box<dyn Error>` isn't necessary, but I'm OK with it. Not sure why we need thiserror.

storage/lmdb/src/lib.rs
```diff
@@ -176,6 +176,9 @@ impl backend::Backend for Lmdb {
     type Impl = LmdbImpl;
 
     fn open(self, desc: DbDesc) -> storage_core::Result<Self::Impl> {
+        // Attempt to create the storage directry
```
TheQuantumPhysicist, line 179: directory

Btw, there are also two typos in lmdb wrapper, also directory and exchange

storage/lmdb/src/lib.rs
```diff
@@ -176,6 +176,9 @@ impl backend::Backend for Lmdb {
...
+        std::fs::create_dir_all(&self.path).map_err(error::process_io_error)?;
```
TheQuantumPhysicist, line 180: The policy for creating directories should be handled with care. The default policy I see reasonable (same one as bitcoin's) is that:
- If the data directory is created by default in the default path, then create everything without question
- If the data directory is chosen manually, it has to exist and nothing should be created.

This is probably not directly related to lmdb. This has to be handled in the node runner. It just occurred to me now.

test/functional/feature_lmdb_backend_test.py
```diff
@@ -0,0 +1,69 @@
...
+from test_framework.util import (
+    assert_equal,
+)
+import test_framework.sample_data
+
+class ExampleTest(BitcoinTestFramework):
+    def set_test_params(self):
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+        self.extra_args = [['--storage-backend=lmdb']]
```
TheQuantumPhysicist, line 17: As mentioned before, persistent storage should be the default.

test/functional/p2p_syncing_test.py
```diff
@@ -40,14 +41,8 @@ def run_test(self):
...
-            "01b02e63a4a8ab958125fcd41e8a6e326c49dcfea387cda8212d79f32537d3036abb14968e23d301cbc5792ea49762e6241287185cfa4cc224700f78e0cc0230d0c7f8e6f882f62c91706d4827b2e8746f99591dfa4492bfd1609d16e793e8af9a031391366301ffff7f20000000000000000000000000000000000004010000000004004ed9e281e0e79d7dbd6be81093fdad09397aff197695669a2b733d76490929170000000004000bd1b77f19c4050003000000000400018d01334a3c0818541f490c413a254f4813243f5259583b4d2d1a31211c26015e27052b4b534261565f1707291d3611600d1b1204234e4c5a452214282e0e5b3540373d390a093850573062631e475c0f2c2f205d2a1955060b1532164310035102463e3444"     
-        ]
-        
+        blocks = test_framework.sample_data.BLOCKS
```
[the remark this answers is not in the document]
iljakuklic: Ok. I don't want to deal with additional RPC calls int this PR so I just copy-pasted the block data so it's more annoying to keep status quo :laughing:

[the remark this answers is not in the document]
TheQuantumPhysicist: Copy/paste is completely acceptable. All will be resolved on time ;-)

node/src/runner.rs
iljakuklic: Ok. I was also wondering if it makes sense to incorporate chain type into the name. For example, when the database contains mainnet blocks, and the user launches the node in testnet mode, different database should be used.

TheQuantumPhysicist: Typically, the testnet has a different data directory because everything in it is different and shouldn't share anything with the mainnet data directory. It could be a subdirectory in the mainnet directory called testnet/regtest/etc.

chainstate/launcher/src/config.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl Default for StorageBackend {
+    fn default() -> Self {
+        Self::Lmdb
+    }
+}
+
+/// Storage configuration
+#[derive(Serialize, Deserialize, Debug, Default)]
+pub struct ChainstateLauncherConfig {
```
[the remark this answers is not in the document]
iljakuklic: `ChainstateConfig` is used if you construct the storage backend manually and pass it to `make_chainstate`. It gives more flexibility in configuring / setting up the backend, which is very useful for testing. In that case, the `storage_backend` option would be unused but you would still have to provide it which is confusing. This is used if the backend specification is a part of the configuration and you want the storage to be initialized for you.

In addition to that, I like the conceptual separation where the `chainstate` crate is 100% backend-agnostic. You can mix-and-match it with any backend without the crate being aware of it. The `chainstate-launcher` crate, on the other hand, just offers a set of pre-defined backends to pick from. That means it also has to pull all of them in as a dependency.

TheQuantumPhysicist (approved):
I'm not particularly happy with the launcher thing despite its apparently benefits, but I'm going to confess that I don't have a better solution for now. The reason I don't like it is that it seems to break the symmetry of subsystem configuration and introduces seemingly unnecessary nesting. Given that I'll give it a shot to fix the configuration file issues, I'll try to figure out something.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ members = [
   "crypto",                       # Cryptographic primitives and their interfaces.
   "consensus",                    # Consensus related logic.
   "chainstate",                   # Code on chainstate of blocks and transactions.
+  "chainstate/launcher",          # Tools to set up chainstate and storage together.
```
TheQuantumPhysicist, line 20: Btw, this is not necessary. Launcher is a dependency of node, so it'll be included automatically.

My understanding is that we should include things here that we want to be autonomously built. There are many things that should be removed from here anyway, like crypto. Maybe I'll create another PR and do that.

PR #494, opened 2022-10-19
Enable mempool to handle chainstate events

royster57:
I think we can/should `expect` instead of having the `SendError` and `RecvError` I added. Also, I think we can `expect` on the results of the `send` calls in the `tokio::select!` event loop, as the receiver is kept alive long enough in `impl MempoolInterface for MempoolInterfaceHandle`. However, I'm wary of adding expect calls so I'd like to hear what you guys think first.

### DIFF
--- a/mempool/src/pool.rs
+++ b/mempool/src/pool.rs
@@ -918,36 +918,44 @@ where
                 Some((block_id, block_height)) = chainstate_event_receiver.recv() =>{
                     self.new_tip_set(block_id, block_height)
                 },
-                Some(method_call) = self.receiver.recv() => match method_call {
-                    MempoolMethodCall::AddTransaction {tx, rtx} => {
-                        if let Err(e) = rtx.send(self.add_transaction(tx).await) {
-                            logging::log::error!("AddTransaction: Error sending response: {:?}", e);
-                        }
+                Some(method_call) = self.receiver.recv() => self.handle_mempool_method_call(method_call).await
+            }
+        }
+    }
 
-                    },
-                    MempoolMethodCall::GetAll {rtx} => {
-                        if let Err(e) = rtx.send(self.get_all()) {
-                            logging::log::error!("GetAll: Error sending response: {:?}", e);
-                        }
-                    },
-                    MempoolMethodCall::CollectTxs {tx_accumulator, rtx} => {
-                        if let Err(e) = rtx.send(self.collect_txs(tx_accumulator)) {
-                            logging::log::error!("CollectTxs: Error sending response: {:?}", e.transactions());
-                        }
-                    },
-                    MempoolMethodCall::ContainsTransaction {tx_id, rtx} => {
-                        if let Err(e) = rtx.send(self.contains_transaction(&tx_id)) {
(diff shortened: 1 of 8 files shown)


mempool/src/pool.rs
```diff
@@ -810,36 +847,137 @@ impl GetMemoryUsage for SystemUsageEstimator {
...
     M: GetMemoryUsage + Send + std::marker::Sync,
 {
-    #[cfg(test)]
-    fn new_tip_set(&mut self) {
-        let mut rolling_fee_rate = self.rolling_fee_rate.write();
-        (*rolling_fee_rate).block_since_last_rolling_fee_bump = true;
+    pub(crate) async fn run(mut self) -> Result<(), Error> {
+        let mut event_receiver = self.subscribe_to_chainstate_events().await?;
+        tokio::spawn(async move {
+            loop {
```
TheQuantumPhysicist, line 927: Why not move this inside a function of its own? Usually the `run()` function contains this, instead of spawning a loop. Just for readability.

mempool/src/pool.rs
```diff
@@ -810,36 +847,137 @@ impl GetMemoryUsage for SystemUsageEstimator {
...
+                tokio::select! {
+                    Some((block_id, block_height)) = event_receiver.recv() =>{
+                        self.new_tip_set(block_id, block_height)
+                    },
+                    Some(method_call) = self.receiver.recv() => match method_call {
+                        MempoolMethodCall::AddTransaction {tx, rtx} => {
+                            if let Err(e) = rtx.send(self.add_transaction(tx).await) {
+                                logging::log::error!("AddTransaction: Error sending response: {:?}", e);
+                            }
+
```
TheQuantumPhysicist, line 937: Format?

mempool/src/pool.rs
```diff
@@ -810,36 +847,137 @@ impl GetMemoryUsage for SystemUsageEstimator {
     }
 }
 
+pub(crate) struct MempoolInterfaceHandle {
```
TheQuantumPhysicist, line 850: Too many `pub(crate)`... which indicate a design issue. I've always said that `pub(crate)` is most of the time not a good sign. Generally speaking, you have a style of putting to many unrelated things in one file. You may wanna think how to structure this into modules that make sense, where `pub` alone can give you the guarantee that only the mempool will see the crates.

For example, in #495, you can see that the `BlockMaker` is only accessible to the crate internally, and I never had to use `pub(crate)`, because the `BlockMaker` is isolated in its own submodule, which is declared only at one level higher, that needs it.

TheQuantumPhysicist (approved):
The functionality seems to be fine. but the mempool needs some restructuring.

stanislav-tkach (approved):
I don't like extensive `pub (crate)` either, but otherwise looking good to me.

mempool/src/pool.rs
royster57: Done.

mempool/src/pool.rs
royster57: Not sure what's wrong with the formatting here?

TheQuantumPhysicist: Select! macro

PR #495, opened 2022-10-19
Block production initial design

TheQuantumPhysicist:
This PR presents the initial design of block production. There's still no consensus in it, but it does the basic things required.

### DIFF
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -98,7 +98,7 @@ impl BlockMaker {
         // TODO: this isn't efficient. We have to create the header first, then see if it obeys consensus rules, then construct the full block
         let current_time = self.time_getter.get_time();
         let block = Block::new(
-            accumulator.txs().clone(),
+            accumulator.transactions().clone(),
             current_tip_id.into(),
             BlockTimestamp::from_duration_since_epoch(current_time),
             common::chain::block::ConsensusData::None,
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -1,0 +1,180 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use chainstate::ChainstateHandle;
+use common::{
+    chain::{
+        block::{timestamp::BlockTimestamp, BlockReward},
+        signed_transaction::SignedTransaction,
+        Block, ChainConfig,
+    },
+    primitives::{BlockHeight, Id, Idable},
+    time_getter::TimeGetter,
+};
+use logging::log;
+use mempool::{pool::DefaultTxAccumulator, MempoolHandle};
+use utils::tap_error_log::LogError;
+
+use crate::BlockProductionError;
+
+pub enum BlockMakerControlCommand {
+    StopBecauseNewTip(Id<Block>, BlockHeight),
+}
+
+/// Slave to the PerpetualBlockBuilder. Every new block tip gets one BlockMaker, and keeps running
+/// until either it's successful in submitting a block, or there's a new tip in chainstate, deeming
+/// the effort pointless
+pub struct BlockMaker {
+    chain_config: Arc<ChainConfig>,
+    chainstate_handle: ChainstateHandle,
+    mempool_handle: MempoolHandle,
+    time_getter: TimeGetter,
+    current_tip_id: Id<Block>,
+    current_tip_height: BlockHeight,
+    block_maker_rx: crossbeam_channel::Receiver<BlockMakerControlCommand>,
+}
+
+enum BlockSubmitResult {
+    Failed,
+    Success,
+}
+
+impl BlockMaker {
+    pub fn new(
+        chain_config: Arc<ChainConfig>,
+        chainstate_handle: ChainstateHandle,
+        mempool_handle: MempoolHandle,
+        time_getter: TimeGetter,
+        current_tip_id: Id<Block>,
+        current_tip_height: BlockHeight,
+        block_maker_rx: crossbeam_channel::Receiver<BlockMakerControlCommand>,
+    ) -> Self {
+        Self {
+            chain_config,
+            chainstate_handle,
+            mempool_handle,
+            time_getter,
+            current_tip_id,
+            current_tip_height,
+            block_maker_rx,
+        }
+    }
+
+    pub async fn collect_transactions(
+        &self,
+    ) -> Result<Vec<SignedTransaction>, BlockProductionError> {
+        let max_block_size = self.chain_config.max_block_size_from_txs();
+        let transactions = self
+            .mempool_handle
+            .call(move |mempool| {
+                mempool.collect_txs(Box::new(DefaultTxAccumulator::new(max_block_size)))
+            })
+            .await?;
+        Ok(transactions)
+    }
+
+    pub fn make_block(
+        &self,
+        current_tip_id: Id<Block>,
+        transactions: Vec<SignedTransaction>,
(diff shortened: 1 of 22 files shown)


mempool/src/tx_accumulator.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum TxAccumulatorError {
+    #[error("Fee overflow: {0:?} + {1:?} failed")]
+    FeeAccumulationError(Amount, Amount),
+}
+
+pub trait TransactionAccumulator: Send {
+    /// Add a transaction to the accumulator and its fee
+    /// This method should not mutate self unless it's successful
+    /// Meaning: If this call returns an error, the callee should guarantee that &self never changed
```
stanislav-tkach, line 28: Should we add a test for this property?

stanislav-tkach (approved):

blockprod/src/detail/builder.rs
```diff
@@ -0,0 +1,220 @@
...
+use futures::FutureExt;
+use logging::log;
+use mempool::{MempoolEvent, MempoolHandle};
+use tokio::sync::mpsc;
+
+use crate::BlockProductionError;
+
+use super::block_maker::{BlockMaker, BlockMakerControlCommand};
+
+/// Master in the master/slave model with BlockMaker. Every time there's a new tip in chainstate or mempool,
```
azarovh, line 33: I think 'master/slave' shouldn't be in modern projects vocabulary, let's replace with something else.

blockprod/src/detail/builder.rs
```diff
@@ -0,0 +1,220 @@
...
+/// the perpetual block builder constructs a new instance of BlockMaker that keeps trying to create a block.
+pub struct PerpetualBlockBuilder {
```
azarovh, line 35: I'd suggest renaming this to something like `BlockProductionController`. Though I like the use of Perpetual here, it is confusing to have both Builder and Maker which are synonyms. Also it requires extensive comments with master/slave to describe the intent.

azarovh (approved):
looks good to me, just some minor comments relating the naming

mempool/src/pool.rs
```diff
@@ -852,17 +816,29 @@ where
                     "collect_txs: next tx has ancestor score {:?}",
                     next_tx.ancestor_score()
                 );
-                tx_accumulator.add_tx(next_tx.tx().clone());
+
+                match tx_accumulator.add_tx(next_tx.tx().clone(), next_tx.fee()) {
+                    Ok(_) => (),
+                    Err(err) => log::error!(
+                        "CRITICAL: Failed to add transaction {} from mempool. Error: {}",
```
royster57, line 823: Shouldn't we break out of this loop if we fail to add a transaction? It seems to me like this will run forever.

TheQuantumPhysicist: It won't run forever, because the iterator will keep proceeding. It'll keep trying again and again though, until the transaction times out in the mempool and gets dropped. But then we should notice this in the logs. This function doesn't have the authority to remove transactions from the mempool.

blockprod/src/detail/builder.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    pub async fn run(&mut self) -> Result<(), BlockProductionError> {
+        let mut mempool_rx = self.subscribe_to_mempool_events().await?;
+        let mut chainstate_rx = self.subscribe_to_chainstate_events().await?;
+
+        loop {
+            tokio::select! {
+                // when we receive information from chainstate that we have a new tip, we stop building the current block,
+                // and expect the mempool to soon send a trigger for a new command
+                block_info = chainstate_rx.recv().fuse() => {
```
royster57, line 132: What problem does the `fuse` call here solve? I read through the docs and I'm still not quite sure. Is it possible that `poll` "may be called too often" on this future like it says in the documentation? Does this have to do with the `tokio::select!` macro?

blockprod/src/detail/block_maker.rs
```diff
@@ -0,0 +1,184 @@
...
+            match new_info {
+                BlockMakerControlCommand::StopBecauseNewTip(block_id, _) => {
+                    // if there is a new tip, no point in continuing to mine this block
+                    if block_id != self.current_tip_id {
+                        break;
+                    }
+                }
+                BlockMakerControlCommand::JustStop => break,
+            }
+        }
```
iljakuklic, line 181: We will have to be careful not to starve other tasks with a busy loop like this. E.g. if `make_block` contains some intensive task like PoW mining or expensive block verification

iljakuklic (approved):
Approving but please mind the comments I added

blockprod/src/detail/builder.rs
TheQuantumPhysicist: Yes. Fuse basically makes it possible to poll a future, get the result, and reset again. It "fuses" the Ready state into another Pending state. At least this is how I understand it.

blockprod/src/detail/block_maker.rs
```diff
@@ -0,0 +1,184 @@
...
+use utils::tap_error_log::LogError;
+
+use crate::BlockProductionError;
+
+pub enum BlockMakerControlCommand {
+    StopBecauseNewTip(Id<Block>, BlockHeight),
+    JustStop,
+}
+
+/// Slave to the PerpetualBlockBuilder. Every new block tip gets one BlockMaker, and keeps running
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Will do. I'm not so familiar with documentation requirements.

blockprod/src/detail/block_maker.rs
TheQuantumPhysicist: Unfortunately, this is how mining works. I already had discussions with Aaro about this and that we may have to introduce multiple executors to solve such problems. But... well, one problem at a time. I'll add a TODO to study this.

blockprod/src/detail/block_maker.rs
```diff
@@ -0,0 +1,184 @@
...
+        current_tip_id: Id<Block>,
+        accumulator: &dyn TransactionAccumulator,
+    ) -> Result<Block, BlockProductionError> {
+        // TODO: this isn't efficient. We have to create the header first, then see if it obeys consensus rules, then construct the full block
+        let current_time = self.time_getter.get_time();
+        let block = Block::new(
+            accumulator.transactions().clone(),
+            current_tip_id.into(),
+            BlockTimestamp::from_duration_since_epoch(current_time),
+            common::chain::block::ConsensusData::None,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yep. Nothing from consensus is here yet. We can't do signing without communication with the wallet, which is why you saw the websocket PR yesterday. That's another problem that's being studied as we speak.

blockprod/src/detail/builder.rs
```diff
@@ -0,0 +1,220 @@
...
+                    self.trigger_new_block_production(block_id, block_height).await?;
+                }
+                event = self.builder_rx.recv().fuse() => match event.ok_or(BlockProductionError::BlockBuilderChannelClosed)? {
+                    BlockBuilderControlCommand::Stop => {
+                        self.enabled = false;
+                        self.stop_all_block_makers()?;
+                    },
+                    BlockBuilderControlCommand::Start => {
+                        self.enabled = true;
+                    },
```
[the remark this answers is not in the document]
TheQuantumPhysicist: It's a convention. 

Stopping must stop everything because there's no point in continuing if we want to stop. While when we start, we may be too late to the game of staking/mining the current block. So we just wait for the next tip to start.

blockprod/src/detail/builder.rs
```diff
@@ -0,0 +1,220 @@
...
+        let subscribe_func = Arc::new(move |mempool_event: MempoolEvent| match mempool_event {
+            MempoolEvent::NewTip(block_id, block_height) => {
+                futures::executor::block_on(async {
+                    if let Err(e) = tx.send((block_id, block_height)) {
+                        log::error!(
+                            "Block production failed to receive event from mempool - channel closed: {:?}",
+                            e
+                        )
+                    }
+                });
```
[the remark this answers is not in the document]
TheQuantumPhysicist: It's a convention we used. Maybe you're right. I'll change it and see how it develops. I'll also change it in p2p.

blockprod/src/detail/mod.rs
```diff
@@ -13,14 +13,60 @@
...
+        };
+        Ok(block_production)
     }
 }
 
-impl BlockProductionInterface for BlockProduction {}
+impl BlockProductionInterface for BlockProduction {
+    fn stop(&self) -> Result<(), BlockProductionError> {
+        self.builder_tx
+            .send(BlockBuilderControlCommand::Stop)
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yes. This is just a signal.

PR #496, opened 2022-10-19
BlockUndo finalise work to make it a map instead of vec

azarovh:
This allows to undo an arbitrary tx if no other tx depends on it

### DIFF
--- a/utxo/src/undo.rs
+++ b/utxo/src/undo.rs
@@ -13,10 +13,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::BTreeMap;
+use std::collections::{btree_map::Entry, BTreeMap, BTreeSet};
 
 use crate::Utxo;
-use common::{chain::Transaction, primitives::Id};
+use common::{
+    chain::{OutPointSourceId, Transaction},
+    primitives::Id,
+};
 use serialization::{Decode, Encode};
 use thiserror::Error;
 
@@ -47,20 +50,8 @@ impl BlockRewardUndo {
 pub struct TxUndo(Vec<Utxo>);
 
 impl TxUndo {
-    pub fn new_empty() -> Self {
-        Self(vec![])
-    }
-
     pub fn new(utxos: Vec<Utxo>) -> Self {
         Self(utxos)
-    }
-
-    pub fn push(&mut self, utxo: Utxo) {
-        self.0.push(utxo)
-    }
-
-    pub fn append(&mut self, other: &mut TxUndo) {
-        self.0.append(&mut other.0)
     }
 
     pub fn inner(&self) -> &[Utxo] {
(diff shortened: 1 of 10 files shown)


utxo/src/undo.rs
```diff
@@ -100,20 +110,47 @@ impl BlockUndo {
...
+            // If not this tx can be taken and returned.
+            // But first, remove itself as a dependency of others.
+            self.dependencies
+                .iter()
+                .filter_map(|v| if v.0 == *tx_id { Some(*v) } else { None })
+                .collect::<Vec<_>>()
+                .iter()
+                .for_each(|p| {
+                    self.dependencies.remove(p);
+                });
```
stanislav-tkach, line 148:
```suggestion
            self.dependencies.retain(|v| &v.0 == tx_id);
```

stanislav-tkach (approved):

utxo/src/undo.rs
```diff
@@ -100,20 +110,47 @@ impl BlockUndo {
...
-        // TODO(PR): this should analyze dependencies before taking an undo
-        self.tx_undos.remove(tx_id)
+        // Check if the tx is a dependency for other txs.
+        let dependencies_count =
+            self.dependencies.iter().filter(|(_child, parent)| parent == tx_id).count();
+        if dependencies_count == 0 {
+            // If not this tx can be taken and returned.
+            // But first, remove itself as a dependency of others.
+            self.dependencies
+                .iter()
```
TheQuantumPhysicist, line 142: Unfortunately, this is a bad design. This being O(N) basically means that undoing blocks is potentially O(N^2). That's a recipe for disaster with DoS attacks. We can't accept that. 

This should be at most O(log(N)). Meaning: You should provide a mechanism to search for transactions that doesn't involve looping over all transactions, but use some kind of tree at worst (like BTreeSet/Map).

utxo/src/undo.rs
```diff
@@ -100,20 +110,47 @@ impl BlockUndo {
...
+
         Ok(())
     }
 
     pub fn take_tx_undo(&mut self, tx_id: &Id<Transaction>) -> Option<TxUndo> {
-        // TODO(PR): this should analyze dependencies before taking an undo
-        self.tx_undos.remove(tx_id)
+        // Check if the tx is a dependency for other txs.
+        let dependencies_count =
+            self.dependencies.iter().filter(|(_child, parent)| parent == tx_id).count();
```
TheQuantumPhysicist, line 137: This is also bad for the same reason. You should never use iter() in this place. You should using something like `self.dependencies.range()`, but it seems that the dependencies are inverted. I expected this to be the other way around.

utxo/src/undo.rs
```diff
@@ -100,20 +118,52 @@ impl BlockUndo {
...
+        let dependencies_count =
+            self.parent_child_dependencies.range(range_start..=range_end).count();
+        if dependencies_count == 0 {
+            // If not this tx can be taken and returned.
+            // But first, remove itself as a dependency of others.
+            let to_remove = self
+                .child_parent_dependencies
+                .range(range_start..=range_end)
+                .copied()
+                .collect::<Vec<_>>();
```
TheQuantumPhysicist, line 157: I don't think you need to collect here... you could just use the iterator in the next step.

azarovh: Actually, I do, because otherwise, I would hold references to `child_parent_dependencies` and won't be able to modify it

TheQuantumPhysicist: I see.

stanislav-tkach: Perhaps the `drain_filter` method can be used here, but it is still unstable.

stanislav-tkach (approved):

Issue #446, opened 2022-09-23
All configuration fields that loads from toml files should be based on Option


TheQuantumPhysicist:
This issue was noticed when we tried to load the [default value of buffer sizes](https://github.com/mintlayer/mintlayer-core/blob/a2b2d6d9ebcd4ba2cdbf61cb6b27abfba8b3ca8a/chainstate/src/config.rs#L27): 

It was seen that any value that isn't an Option, will basically be required in the toml file. This isn't what we need out of configuration files. What we need is to offer the user to *modify* values through the configuration file, instead of having to manually control everything.

So, either everything has to become an Option and the default mechanism should be different, or some other solution must be found that doesn't force the user to input all configuration manually.

PR #497, opened 2022-10-20
Add websocket support to RPC

TheQuantumPhysicist:


node/src/options.rs
```diff
@@ -92,9 +92,21 @@ pub struct RunOptions {
...
+    #[clap(long)]
+    pub http_rpc_enabled: Option<bool>,
+
+    /// Address to bind websocket RPC to.
+    #[clap(long, value_name = "ADDR")]
+    pub ws_rpc_addr: Option<SocketAddr>,
+
+    /// Enable/Disable websocket RPC.
+    #[clap(long)]
+    pub ws_rpc_enabled: Option<bool>,
```
iljakuklic, line 109: Here's not very well thought out suggestion. Maybe the combination of the `SocketAddr` and `bool` could be handled by having a special value for disabling the address, like `--ws-rpc-addr=none` or `--ws-rpc-addr=disable`?

Using a custom type like
```rust
enum SocketAddrConfig { Disabled, Enabled(SorcketAddr) }
impl FormStr for SocketAddrConfig { /* handle "none" here */ }
```
the `ws_rpc_addr` and `http_rpc_addr` could be `Option<SocketAddrConfig>`.

I don't have a problem with the current approach for now, the configuration handling needs a bit more thought anyway so can be done separately.

node/src/runner.rs
```diff
@@ -104,16 +104,19 @@ pub async fn initialize(
...
+            "rpc",
+            rpc::Builder::new(node_config.rpc)
+                .register(chainstate.clone().into_rpc())
+                .register(mempool.into_rpc())
+                .register(NodeRpc::new(manager.make_shutdown_trigger()).into_rpc())
+                .register(p2p.clone().into_rpc())
+                .build()
+                .await?,
+        );
+    }
```
iljakuklic, line 119: Would it make sense for `rpc-http` and `rpc-ws` to be separate subsystems and only start the ones that are enabled? The `rpc` crate now has way too many pattern matches on `Option` to my liking.

After some refactoring in the `rpc` module, this bit could look something like:

```rust
let rpc_builder = rpc::builder::new().register(/*...*/).register(/*...*/);
if node_config.rpc.http_enabled() {
    manager.add_subsystem("rpc-http", rpc_builder.build_http(node_config.rpc.http_addr()).await);
}
if node_config.rpc.ws_enabled() {
    manager.add_subsystem("rpc-ws", rpc_builder.build_ws(node_config.rpc.ws_addr()).await);
}
```

iljakuklic (commented):
Adding some initial comments

node/src/options.rs
TheQuantumPhysicist: I don't want to deal with this until the configuration issue is solved. It's basically beating a dead horse at this point.

node/src/runner.rs
TheQuantumPhysicist: They both use the same function objects, so we'll have to duplicate the handlers. Not really the best choice.

azarovh (approved):

node/src/options.rs
TheQuantumPhysicist: To be clear, I'm talking about #446 . This has wasted enough time in this PR.

TheQuantumPhysicist:
> I'm fine with merging this. However, I'd say having a separate sybsystem for each server (i.e. `rpc-http` and `rpc-ws`) deserves some consideration

I'll have to work on websocket much more later. Let's solve one problem at a time. I'm still not sure we need this at this point.

stanislav-tkach (approved):

PR #500, opened 2022-10-21
chainstate: Genesis check

iljakuklic:
Introduce a check that verifies the genesis block the storage builds on matches the one specified in `ChainConfig`. This is to prevent accidentally using storage with a different chain from the one it was started with (e.g. testnet vs mainnet).

### DIFF
--- a/chainstate/src/detail/mod.rs
+++ b/chainstate/src/detail/mod.rs
@@ -29,7 +29,9 @@ pub use self::error::*;
 pub use self::median_time::calculate_median_time_past;
 pub use self::tokens::is_rfc3986_valid_symbol;
 pub use chainstate_types::Locator;
-pub use error::{BlockError, CheckBlockError, CheckBlockTransactionsError, OrphanCheckError};
+pub use error::{
+    BlockError, CheckBlockError, CheckBlockTransactionsError, InitializationError, OrphanCheckError,
+};
 // TODO: ConnectTransactionError used in unit tests to check block processing results. We have to find more appropriate place for this error.
 pub use transaction_verifier::{
     error::{ConnectTransactionError, TokensError, TxIndexError},
@@ -147,12 +149,7 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V> 
 
         let best_block_id = chainstate_storage
             .get_best_block_id()
-            .map_err(|e| {
-                ChainstateError::FailedToInitializeChainstate(format!(
-                    "Database read error: {:?}",
-                    e
-                ))
-            })
+            .map_err(|e| ChainstateError::FailedToInitializeChainstate(e.into()))
             .log_err()?;
 
         let mut chainstate = Self::new_no_genesis(
@@ -169,7 +166,10 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V> 
                 .process_genesis()
                 .map_err(crate::ChainstateError::ProcessBlockError)
                 .log_err()?;
+        } else {
+            chainstate.check_genesis().map_err(crate::ChainstateError::from)?;
         }
+
         Ok(chainstate)
     }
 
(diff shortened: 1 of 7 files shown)


chainstate/src/detail/mod.rs
```diff
@@ -164,11 +164,24 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
             time_getter,
         );
 
-        if best_block_id.is_none() {
+        if best_block_id.is_some() {
```
TheQuantumPhysicist, line 167: I think this is not correct. It should remain is_none.

chainstate/storage/src/internal/mod.rs
```diff
@@ -151,6 +152,7 @@ impl<B: storage::Backend> BlockchainStorageRead for Store<B> {
     delegate_to_transaction! {
         fn get_storage_version(&self) -> crate::Result<u32>;
         fn get_best_block_id(&self) -> crate::Result<Option<Id<GenBlock>>>;
+        fn get_genesis_block_id(&self) -> crate::Result<Option<Id<Genesis>>>;
```
TheQuantumPhysicist, line 155: I don't think this is necessary. I'll write in the relevant place how to avoid this.

chainstate/src/detail/mod.rs
```diff
@@ -164,11 +164,24 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
             time_getter,
         );
```
TheQuantumPhysicist, line 166: To avoid having to store a separate genesis block id, simply do the following:

instead of that if condition below, match on it.

- If it's None, then the genesis block isn't initialized. No problem there. Just run the normal initialization.
- If it's Some, check the best height:
    - If the height is > 0, retrieve block number 1, check the prev block id. If it's genesis, you've proven what you wanted to prove.
    - If the height is 0, the current best block should be genesis block.

iljakuklic: I started with this but then thought it would be conceptually simpler to just store the genesis ID. Can do what you suggest, I don't have a strong preference either way

TheQuantumPhysicist: Would be great if you could go through with it. Just isolate it in a separate function `check_genesis_matches` or something. I have many fears that relate to people abusing the value of genesis being in storage in the future instead of properly using `ChainConfig`.

chainstate/src/detail/mod.rs
```diff
@@ -194,6 +194,32 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
         }
     }
 
+    fn check_genesis(&self) -> Result<(), InitializationError> {
```
azarovh, line 197: maybe cover it with a simple unit test?

azarovh (approved):

chainstate/src/detail/mod.rs
TheQuantumPhysicist: Yes, please. I was gonna suggest this. A test at height zero, and at another arbitrary height would be great.

TheQuantumPhysicist (approved):
Just the tests missing. Otherwise, all good.

TheQuantumPhysicist (approved):

chainstate/test-suite/src/tests/initialization.rs
```diff
@@ -0,0 +1,134 @@
...
+#[test]
+fn genesis_check_err_height_1() {
+    genesis_check_err(1);
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn genesis_check_err_nonempty_chain(#[case] seed: Seed) {
+    genesis_check_err(make_seedable_rng(seed).gen_range(1..100));
```
stanislav-tkach, line 133: Should the range be `2..100` because the one block case is covered by a separate test? 🙃

stanislav-tkach (approved):

chainstate/test-suite/src/tests/initialization.rs
TheQuantumPhysicist: Doesn't hurt to check twice :-)

azarovh (approved):

PR #502, opened 2022-10-24
Attempt to make configuration more reasonable - Please read the description

TheQuantumPhysicist:
This is more like a proposal. Let's have a discussion about it and find the best way to tackle this problem.

There's a semi-predicate problem happening in the way configuration was done. Everyone using configuration transiently is facing an issue understanding how it's supposed to work. The problem is that the configuration file is used as configuration for subsystems. In many cases, there's no way to distinguish between the user not supplying data and the default value.

To solve this problem, I separated the configuration that comes from files, from the configuration that goes into subsystems. This makes sense because supplying configuration through configuration files is optional, and is not mandatory by any means. That object, though, will tell us *ONLY* whether the user supplied the configuration. Then, we can construct the configuration of a subsystem based on that information. The configuration of a subsystem then can decide how to derive the default values when the user supplied nothing.

Finally, in order to solve the problem of default values, I created a macro that will generate a thin wrapper around configuration options in subsystems. An option will become a type with a default value introduced in a `Default::default()` implementation. Basically, this makes configuration values types, with each type holding its default value in a function.

Let me know what you, guys, think. Hopefully we can come to a good conclusion with this problem.


TheQuantumPhysicist:
closes #446

node/src/config_files/p2p.rs
```diff
@@ -0,0 +1,133 @@
...
+use p2p::config::{MdnsConfig, P2pConfig};
+use serde::{Deserialize, Serialize};
+
+pub const MDNS_DEFAULT_QUERY_INTERVAL: u64 = 0;
+pub const MDNS_DEFAULT_IPV6_STATE: bool = false;
+
+/// Multicast DNS configuration.
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(tag = "state")]
+pub enum MdnsConfigFile {
```
azarovh, line 25: shouldn't this be a struct with optional fields, same way as other *ConfigFile types? Otherwise why not construct MndsConfig at once.

do I understand the idea right? we want to have a chain: toml -> *ConfigFile with optionals -> *Config with concrete values

rpc/src/lib.rs
```diff
@@ -67,15 +67,15 @@ impl Builder {
     /// New builder pre-populated with RPC info methods
     pub fn new(rpc_config: RpcConfig) -> Self {
         // TODO: this is a mess because of the configuration file mapping from files to options. See #446
```
azarovh, line 69: just a reminder to clean up todos if the PR is merged

p2p/src/config.rs
```diff
@@ -13,20 +13,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use serde::{Deserialize, Serialize};
+use utils::make_config_setting;
 
 pub const MDNS_DEFAULT_QUERY_INTERVAL: u64 = 0;
```
azarovh, line 18: these constants are duplicated in node/src/config_files/p2p.rs. I think they should be defined once

utils/src/config_setting.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        impl From<Option<$tp>> for $name {
+            fn from(v: Option<$tp>) -> Self {
+                Self {
+                    value: v.unwrap_or(Self::default().value),
+                }
+            }
+        }
```
iljakuklic, line 42: I think this is just `Option::unwrap_or_default`. Could that be used instead of a custom `From` impl to be more explicit?

iljakuklic (commented):
Looks good. Would like to see the storage backed configuration code duplication addressed before approving. Other comments are just minor suggestions.

node/src/config_files/chainstate/storage_backend.rs
```diff
@@ -0,0 +1,54 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serde::{Deserialize, Serialize};
+
+/// Storage type to use
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub enum StorageBackendFileConfig {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: You're right. Thanks for pointing it out.

p2p/Cargo.toml
```diff
@@ -27,7 +27,6 @@ tap = "1.0"
 once_cell = "1.13"
 jsonrpsee = { version = "0.15", features = ["macros"]}
 libp2p = { version = "0.46", default-features = false, features = ["gossipsub", "identify", "mdns", "mplex", "noise", "ping", "request-response", "tcp-tokio"] }
-serde = { version = "1", default-features = false, features = ["derive"] }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: It was one of the goals of this PR. Why should the configuration of subsystems be serializable? That doesn't make sense in my head. Cheers!

utils/src/config_setting.rs
TheQuantumPhysicist: Unfortunately, that doesn't work, because the default would be for v of type `Option<$tp>`, whose default is not really what we're looking for.

node/src/config_files/p2p.rs
TheQuantumPhysicist: Yes and no. This is an enum, so this can't really be done with Option like structs. But this has a method "from_options", which makes everything optional.

rpc/src/lib.rs
TheQuantumPhysicist: Thank you.

utils/src/config_setting.rs
iljakuklic: Hmm. I see. The impl looks a bit awkward though. Where is it used?

iljakuklic (approved):

utils/src/config_setting.rs
TheQuantumPhysicist: This is basically the de-facto conversion method from "user didn't supply a value" to "default value".

iljakuklic: I guess my question is why not use conversion from `Option<$name>` rather than `Option<$tp>` in that case, and use the more expressive wrapper types in the configuration files / command lines as well. Is there an issue with that? There might be something I'm missing.

TheQuantumPhysicist: Because this type has nothing to do with serialization. The only purpose of this type is to have a well-defined way of interpreting input options. The serialization of `$tp` comes from command line arguments, which have their serialization defined with them. 

If you think about it, it'll become a circular problem. You add serialization to this type and use it in command line arguments, and we're back to the semi-predicate problem, where we can't distinguish between None and default. I don't know, maybe I'm missing something.

PR #506, opened 2022-10-25
Implement block announcements for the p2p mock backend

stanislav-tkach:
Fixes https://github.com/mintlayer/mintlayer-core/issues/488.


p2p/src/net/mock/mod.rs
```diff
@@ -111,9 +110,9 @@ where
     T: MockTransport,
 {
     type Address = T::Address;
-    type PeerId = types::MockPeerId;
-    type SyncingPeerRequestId = types::MockRequestId;
-    type SyncingMessageId = MockMessageId;
+    type PeerId = MockPeerId;
+    type SyncingPeerRequestId = MockRequestId;
+    type SyncingMessageId = ();
```
iljakuklic, line 115: Why the unit type here? Would it make sense to use a unit struct with a proper name instead?

iljakuklic (approved):

p2p/src/net/mock/mod.rs
stanislav-tkach: Good point! I thought that eventually we could remove the associated `SyncingMessageId` type entirely, but it is still a good idea to use a named type for now.

TheQuantumPhysicist (approved):

PR #508, opened 2022-10-26
Use custom time getter by default for chainstate TestFramework

TheQuantumPhysicist:
With this PR, tests should be 100% deterministic unless the `TimeGetter` decides otherwise. 

The design of the solution I provided is up for discussion.

### DIFF
--- a/chainstate/test-framework/src/framework.rs
+++ b/chainstate/test-framework/src/framework.rs
@@ -32,8 +32,10 @@ pub struct TestFramework {
     pub chainstate: super::TestChainstate,
     pub storage: TestStore,
     pub block_indexes: Vec<BlockIndex>,
-    pub time_getter: TimeGetter, // A clone of the TimeGetter supplied to the chainstate
-    pub time_value: Option<Arc<AtomicU64>>, // current time since epoch; if None, it means a custom TimeGetter was supplied and this is useless
+    // A clone of the TimeGetter supplied to the chainstate
+    pub time_getter: TimeGetter,
+    // current time since epoch; if None, it means a custom TimeGetter was supplied and this is useless
+    pub time_value: Option<Arc<AtomicU64>>,
 }
 
 impl TestFramework {
--- a/chainstate/test-framework/src/framework.rs
+++ b/chainstate/test-framework/src/framework.rs
@@ -19,15 +19,21 @@ use chainstate_types::{BlockIndex, GenBlockIndex};
 use common::{
     chain::{Block, GenBlock, Genesis},
     primitives::{id::WithId, BlockHeight, Id, Idable},
+    time_getter::TimeGetter,
 };
 use crypto::random::Rng;
-use std::sync::Arc;
+use std::{
+    sync::{atomic::AtomicU64, Arc},
+    time::Duration,
+};
 
 /// The `Chainstate` wrapper that simplifies operations and checks in the tests.
 pub struct TestFramework {
     pub chainstate: super::TestChainstate,
     pub storage: TestStore,
     pub block_indexes: Vec<BlockIndex>,
+    pub time_getter: TimeGetter, // A clone of the TimeGetter supplied to the chainstate
+    pub time_value: Option<Arc<AtomicU64>>, // current time since epoch; if None, it means a custom TimeGetter was supplied and this is useless
 }
 
 impl TestFramework {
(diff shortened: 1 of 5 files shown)


stanislav-tkach (approved):
I like the idea of a better control over time in the tests, though I'm not 100% happy with the optional `time_value` field. All these "custom time getter" panics don't look pretty, but I think it is OK for the test code.

azarovh (approved):
just curious does it solve the problem seen in bootstrap test?

TheQuantumPhysicist:
> I like the idea of a better control over time in the tests, though I'm not 100% happy with the optional `time_value` field. All these "custom time getter" panics don't look pretty, but I think it is OK for the test code.

Exactly. As long as it's just for the test, it's just for the test. Panics in the tests are acceptable, so we can live with it. I'm happy with changing the design as long as it achieves the same goal.

TheQuantumPhysicist:
> just curious does it solve the problem seen in bootstrap test?

The implementation of bootstrap had a bug and that was fixed in #503. The issue primary issue I complained about in bootstrap, besides the one that was fixed, is that we get different on every run even when using the same seed value. That doesn't happen anymore. At this point, if you have a bug, you'll get the exact same result every time you run the test if you use the same seed.

chainstate/test-framework/src/framework.rs
```diff
@@ -19,15 +19,21 @@ use chainstate_types::{BlockIndex, GenBlockIndex};
...
+    time::Duration,
+};
 
 /// The `Chainstate` wrapper that simplifies operations and checks in the tests.
 pub struct TestFramework {
     pub chainstate: super::TestChainstate,
     pub storage: TestStore,
     pub block_indexes: Vec<BlockIndex>,
+    pub time_getter: TimeGetter, // A clone of the TimeGetter supplied to the chainstate
+    pub time_value: Option<Arc<AtomicU64>>, // current time since epoch; if None, it means a custom TimeGetter was supplied and this is useless
```
iljakuklic, line 36: Minor: I'd prefer the comments to be before the struct member. This makes the lines a bit too long

iljakuklic (approved):
Was thinking what could be done about the time getter option issue without going completely crazy with traits and stuff (only slightly crazy). Here's an approach to consider which might work with some modifications:

1. Parametrize TestFramework by the time getter type
   ```rust
   pub struct TestFramework<TG> {
       /* ... */
       pub time_getter: TG,
   ```
2. Any operation that needs to get time puts a constraint `TG: AsRef<TimeGetter>` and uses that to convert it to the actual time getter
3. Implement `AsRef<TimeGetter> for TimeGetter`
4. Create a `ManualClock` type that contains the time getter and the time value that can be controlled manually
  a. expose the fucntions to control the time manually
  b. Implement `AsRef<TimeGetter>` for it
5. Now `TestFramework<TimeGetter>` can be used for a custom time getter function or when using system clock, `TestFramework<ManualClock>` can be used in testing to control time manually
6. Fun part: since the builder takes ownership of `self`, it can change the type:
   ```rust
    // Inside impl<TG> TestFrameworkBuilder<TG>
    fn with_time_getter<TG1>(
        self /*: TestFrameworkBuilder<TG> */,
        time_getter: TG1,
    ) -> TestFrameworkBuilder<TG1> {...}
    ```

I leave to your best judgement whether this is worth giving a shot

TheQuantumPhysicist:
> Was thinking what could be done about the time getter option issue without going completely crazy with traits and stuff (only slightly crazy). Here's an approach to consider which might work with some modifications:
> 
> 1. Parametrize TestFramework by the time getter type
>    ```rust
>    pub struct TestFramework<TG> {
>        /* ... */
>        pub time_getter: TG,
>    ```
> 2. Any operation that needs to get time puts a constraint `TG: AsRef<TimeGetter>` and uses that to convert it to the actual time getter
> 3. Implement `AsRef<TimeGetter> for TimeGetter`
> 4. Create a `ManualClock` type that contains the time getter and the time value that can be controlled manually
>    a. expose the fucntions to control the time manually
>    b. Implement `AsRef<TimeGetter>` for it
> 5. Now `TestFramework<TimeGetter>` can be used for a custom time getter function or when using system clock, `TestFramework<ManualClock>` can be used in testing to control time manually
> 6. Fun part: since the builder takes ownership of `self`, it can change the type:
>    ```rust
>     // Inside impl<TG> TestFrameworkBuilder<TG>
>     fn with_time_getter<TG1>(
>         self /*: TestFrameworkBuilder<TG> */,
>         time_getter: TG1,
>     ) -> TestFrameworkBuilder<TG1> {...}
>    ```
> 
> I leave to your best judgement whether this is worth giving a shot

One thing I've been very aggressively avoiding is making the TestFrameworkBuilder generic. This is because it'll require generic parameters on construction. Unfortunately, it seems that your plan may require that. We had a similar issue with TransactionVerificationStrategy, and that's why you see that ugly enum.

iljakuklic:
> One thing I've been very aggressively avoiding is making the TestFrameworkBuilder generic. This is because it'll require generic parameters on construction.

The `TestFrameworkBuilder::new` function would just return monomorphic `TestFrameworkBuilder<ManualClock>` (or whatever the default is). Only `TestFrameworkBuilder::with_time_getter` would be generic in my original proposal but even that could just return a builder with fixed parametrization.

TheQuantumPhysicist:
> > One thing I've been very aggressively avoiding is making the TestFrameworkBuilder generic. This is because it'll require generic parameters on construction.
> 
> The `TestFrameworkBuilder::new` function would just return monomorphic `TestFrameworkBuilder<ManualClock>` (or whatever the default is). Only `TestFrameworkBuilder::with_time_getter` would be generic in my original proposal but even that could just return a builder with fixed parametrization.

I will try it tomorrow. I'll merge this now as I need to run a test loop over night.

PR #516, opened 2022-10-27
Tests: More robust way of creating temporary test directories

iljakuklic:
Currently only used in the LMDB backend tests


storage/backend-test-suite/src/prelude.rs
```diff
@@ -46,7 +46,8 @@ pub fn using_proptest<B: Backend, F: BackendFn<B>, S: proptest::prelude::Strateg
 ) {
     let config = {
         let mut config = proptest::prelude::ProptestConfig::with_source_file(source_file);
-        config.cases = 64;
+        // Decrease the number of test cases. By default, this is 256 / 8 = 64.
```
TheQuantumPhysicist, line 49: Why? Can you please provide some context?

test-utils/src/test_dir.rs
```diff
@@ -0,0 +1,134 @@
...
+        atomic::{AtomicU32, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use crypto::random::Rng;
+
+fn make_rng() -> impl Rng {
+    let seed = crate::random::Seed::from_entropy();
```
TheQuantumPhysicist, line 31: Perhaps this should be printed somewhere

TheQuantumPhysicist (commented):
I'll run this branch with all tests all night

storage/backend-test-suite/src/prelude.rs
iljakuklic: By default, this is the same as was before. I just wanted to have the ability to adjust the amount using the `PROPTEST_CASES` env variable.

test-utils/src/test_dir.rs
iljakuklic: It does not seem very useful tbh. This is run when the test machinery is initialized, not inside of the tests so it does not play nice with `--nocapture` (is always printed).

storage/backend-test-suite/src/prelude.rs
```diff
@@ -46,7 +46,8 @@ pub fn using_proptest<B: Backend, F: BackendFn<B>, S: proptest::prelude::Strateg
 ) {
```
TheQuantumPhysicist, line 46: There are two typos on this file, "database" and "description"

test-utils/src/test_dir.rs
```diff
@@ -0,0 +1,134 @@
...
+    sync::{
+        atomic::{AtomicU32, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use crypto::random::Rng;
+
+fn make_rng() -> impl Rng {
```
TheQuantumPhysicist, line 30: The way this is being used, kind of defeats the purpose... we should just throw in a new rng everywhere we need it. The correct strategy is to have a seed generated somewhere, create the rng once (or more with care with deterministic relationship), then use that everywhere.

test-utils/src/test_dir.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        let mut rng = make_rng();
+
+        for delay in backoff_delays() {
+            let path = top_path.as_ref().join(format!("run_{:08x}", rng.gen::<u32>()));
+
+            // Attempt to create the candidate directory
+            match fs::create_dir(&path) {
+                Ok(()) => {
+                    // Sucesfully created
```
TheQuantumPhysicist, line 80: successfully

test-utils/tests/dir.rs
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[test]
+fn existing_dir_error() {
+    pub const TARGET_TMPDIR: &str = env!("CARGO_TARGET_TMPDIR");
+    // Target tmpdir should have been created by cargo, check we get the expected error code
+    match std::fs::create_dir(TARGET_TMPDIR) {
```
TheQuantumPhysicist, line 20: I wonder whether we can wrap this properly so that we're testing our own function instead of testing rust's behavior.

iljakuklic:
> This worked successfully all night.

Good to hear but let's keep hammering it. I can imagine that hitting an issue that happens by chance has a similar probability distribution to bitcoin mining: sometimes there are two blocks coming in a quick succession, sometimes you wait quite a while for one. Therefore it's quite plausible it could have happened by chance.

test-utils/src/test_dir.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct TestRootImpl {
+    path: PathBuf,
+    counter: AtomicU32,
+}
+
+/// Represents a test directory. It gets deleted if the object is dropped, unless there's a panic.
```
stanislav-tkach, line 124: I think this comment isn't quite accurate. We don't delete individual directories, we delete everything that is located inside the root directory. Additionally the deletion isn't performed when this object is dropped. Am I missing something?

stanislav-tkach (approved):

test-utils/src/test_dir.rs
iljakuklic: Yeah, I attempted what's written there but it proved rather difficult to do with how the current storage tests work. Will update the comment. Thanks for pointing this out

test-utils/tests/dir.rs
iljakuklic: The intention is to test Rust's behaviour here, just wanted to make sure what is the error we get when creating an existing dir. I will think a bit how to test our function in addition to it but I'd like to keep this around.

test-utils/src/test_dir.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        Err(io::Error::new(
+            io::ErrorKind::AlreadyExists,
+            "All attempted directory names exist",
+        ))
+    }
+
+    /// Create the new test subdirectory
+    pub fn fresh_test_dir(&self, name: impl AsRef<str>) -> TestDir {
```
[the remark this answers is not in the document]
iljakuklic: I am not sure, it's basically a test name, not necessarily the directory name (although it gets included in the directory name). Would also involve more casting between `&Path` and `&str` to make the `format!` below work.

test-utils/tests/dir.rs
TheQuantumPhysicist: I completely understand what you're trying to achieve. But instead of testing what a generic function does, just wrap it with your own call with proper constraints. For example `create_new_dir_or_panic()`. This way, if someone else sees the logic and thinks of changing it with something else in the future, they realize that it's an invariant in this test that a panic should happen if the dir already exists. Feel free to keep this, but it just seems more logical to restrict behavior to our code.

TheQuantumPhysicist (approved):
Other than the one comment, I think this is good to go.

test-utils/tests/dir.rs
iljakuklic: Created a simple wrapper that is used in both our tests and in the code returning an enum to distinguish between already existing and created dir. Maybe a bit overkill but worked out quite well in the end. It's in the last commit, please take a look.

TheQuantumPhysicist: Looking good. Thank you!

Issue #511, opened 2022-10-26
Shuffle peers when broadcasting messages


stanislav-tkach:
Random peer order should be used every time when we need to broadcast messages to all peers.

PR #519, opened 2022-10-28
Shuffle peers when broadcasting messages in the mock p2p backend

stanislav-tkach:
Closes https://github.com/mintlayer/mintlayer-core/issues/511.

### DIFF
--- a/p2p/src/net/mock/backend.rs
+++ b/p2p/src/net/mock/backend.rs
@@ -28,13 +28,13 @@ use std::{
 };
 
 use futures::{future::join_all, FutureExt, TryFutureExt};
-use rand::seq::SliceRandom;
 use tokio::{
     sync::{mpsc, oneshot},
     time::timeout,
 };
 
 use common::chain::ChainConfig;
+use crypto::random::{make_pseudo_rng, SliceRandom};
 use logging::log;
 use serialization::{Decode, Encode};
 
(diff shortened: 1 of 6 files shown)


p2p/src/net/mock/backend.rs
```diff
@@ -252,13 +253,23 @@ where
...
+            .iter()
+            .map(|(id, p)| {
+                p.tx.send(subscription.clone()).inspect_err(move |e| {
+                    log::error!("Failed to send subscription to {id:?} peer: {e:?}")
+                })
             })
-        }))
-        .await;
+            .collect();
+        futures.shuffle(&mut rand::thread_rng());
```
azarovh, line 270: Shouldn't `crypto::random::make_pseudo_rng()` be used instead?

azarovh (approved):

p2p/src/net/mock/backend.rs
stanislav-tkach: Sure! Fixed.

muursh (approved):
Fine by me

p2p/src/net/mock/types.rs
```diff
@@ -203,7 +203,7 @@ pub enum PeerEvent {
 }
 
 /// Events sent by the mock backend to peers
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
TheQuantumPhysicist, line 206: While this is a mock and doesn't really matter, in production code, cloning the message is not really a good idea. Consider that a message can be a whole block, like 1 MB of data. If the logic requires having multiple copies floating around, then an Rc or Arc is the right way to go.

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #521, opened 2022-10-28
Prepare TransactionVerifier to be able to connect/disconnect txs from mempool

azarovh:
This PR has several part in it:

1) It fixes `TransactionVerifier`'s logic of connecting/disconnecting txs from mempool. Specifically, in terms of managing transaction index and block undo cache.
2) Introduces `can_disconnect_transaction` function
3) Adds tests for checking timelocks for txs from mempool


chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -628,47 +618,113 @@ impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S> {
...
+        tx_source: &TransactionSource,
+        tx_id: &Id<Transaction>,
+    ) -> bool {
+        match tx_source {
+            TransactionSource::Chain(block_id) => {
+                let current_block_height =
+                    match self.storage_ref.get_gen_block_index(&(*block_id).into()) {
+                        Ok(block_index) => match block_index {
+                            Some(block_index) => block_index.block_height(),
+                            None => return false,
```
azarovh, line 632: I don't like all these explicit `return false`, but I don't know how to ignore Err from Result in a cleaner way.
My intention was to treat errors in this function as negative result, so that the user could simply write:
```rust
if verifier.can_disconnect_transaction(tx) {
```

If someone has a suggestion, please comment

chainstate/src/detail/chainstateref/tx_verifier_storage.rs
```diff
@@ -181,15 +182,31 @@ impl<'a, S: BlockchainStorageWrite, O: OrphanBlocks, V: TransactionVerificationS
 
     fn set_undo_data(
         &mut self,
-        id: Id<Block>,
+        tx_source: TransactionSource,
```
TheQuantumPhysicist, line 185: I think this is unnecessary. Instead of making it possible to flush mempool tx then panic, why don't we just prevent it through types?

chainstate/test-suite/src/tests/chainstate_storage_tests.rs
```diff
@@ -34,7 +34,7 @@ use utxo::UtxosStorageRead;
 #[trace]
 #[case(Seed::from_entropy())]
 fn store_coin(#[case] seed: Seed) {
-    utils::concurrency::model(move || {
+    ::utils::concurrency::model(move || {
```
TheQuantumPhysicist, line 37: When this starts happening, it's better to find a better name for the new module.

chainstate/src/detail/chainstateref/tx_verifier_storage.rs
azarovh: I panic only when flushing to ChainstateRef, but at the same time it's ok to flush undo for mempool from one verifier to another

chainstate/test-suite/src/tests/tx_verifier_disconnect.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    let chain_config = ConfigBuilder::test_chain().build();
+    let storage = InMemoryStorageWrapper::new(storage, chain_config.clone());
+
+    (chain_config, storage, tf)
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy(), 5)]
```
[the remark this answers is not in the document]
azarovh: replaced with 2 as a minimal case

chainstate/test-suite/src/tests/mempool_output_timelock.rs
```diff
@@ -0,0 +1,388 @@
...
+        .build();
+
+    let chain_config = ConfigBuilder::test_chain().build();
+    let storage = InMemoryStorageWrapper::new(storage, chain_config.clone());
+
+    (chain_config, storage, tf)
+}
+
+#[test]
+fn output_lock_until_height() {
```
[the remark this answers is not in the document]
azarovh: not sure that it's worth it. Though the frame of test cases is the same they are using different methods to modify chain: `tf.process_block` vs `verifier.connect_trancastion`, which seems to create lot's of boilerplate code to generalize

chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs
```diff
@@ -98,7 +98,7 @@ impl<'a, S: TransactionVerifierStorageRef> UtxosStorageRead for TransactionVerif
         &self,
         id: Id<Block>,
     ) -> Result<Option<utxo::BlockUndo>, storage_result::Error> {
-        match self.utxo_block_undo.get(&id) {
+        match self.utxo_block_undo.get(&TransactionSource::Chain(id)) {
```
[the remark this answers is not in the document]
azarovh: So the problem is the same as with panics: interfaces `TransactionVerifierStorageRef` and `TransactionVerifierStorageMut` connect both `TransactionVerifier` -> `TransactionVerifier` and `TransactionVerifier` -> `ChainstateRef`.  I believe we want to allow flushing/reading undo data from one verifier to another and forbid it in case of ChainstateRef.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -628,47 +637,113 @@ impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S> {
...
+    pub fn disconnect_transaction(
+        &mut self,
+        tx_source: &TransactionSource,
+        tx: &SignedTransaction,
+    ) -> Result<(), ConnectTransactionError> {
+        let tx_undo = match tx_source {
+            TransactionSource::Chain(_) => {
+                // update tx index only for txs from main chain
+                // pre-cache all inputs
+                self.tx_index_cache.precache_inputs(tx.inputs(), |tx_id: &OutPointSourceId| {
```
TheQuantumPhysicist, line 690: Why inline the fetcher function? Make it its own line, then it's more readable.

TheQuantumPhysicist (approved):
Besides the few things we discussed and the last comment, I think we're good to go. Please fix the interface of can_disconnect_transaction, then I'll take a look then we merge.

TheQuantumPhysicist (approved):

PR #522, opened 2022-10-30
Try spend burned output test

azarovh:


erasmospunk:
Do you think it makes sense to add a second case where the spending transaction is in not in the same block but a subsequent one?

TheQuantumPhysicist:
> Do you think it makes sense to add a second case where the spending transaction is in not in the same block but a subsequent one?

I think it makes sense. Particl had once a bug because of something similar, where a cache sync issue load to a double-spend bug.

TheQuantumPhysicist (approved):
Please add a test for two different blocks, then feel free to merge.

iljakuklic (approved):

PR #524, opened 2022-10-30
Added support for basic Hierarchical Deterministic Key Derivation

erasmospunk:
Added support for hard derivations using the schnorrkel's algorithm.


crypto/src/key/hdkd/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+        Ok(())
+    }
+}
+
+impl PrivateKey {
+    /// Derive a child private key given a derivation path
+    pub fn derive_path(&self, path: &DerivationPath) -> Result<PrivateKey, DerivationError> {
+        let mut key: PrivateKey = self.clone();
+        for num in path.0.iter() {
```
TheQuantumPhysicist, line 151: please use try_fold

crypto/src/key/hdkd/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+
+/// BIP32-like child numbers. Currently we only support hardened derivations
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Hash)]
+pub enum ChildNumber {
+    Hardened(u32),
+}
+
+impl ChildNumber {
+    /// Return a hardened child number
+    pub fn hardened(index: u32) -> Result<Self, DerivationError> {
```
TheQuantumPhysicist, line 54: Like discussion in the call, index can mean index and can mean index | HARDENED_BIT, please disambiguate them.

crypto/src/key/hdkd/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+    #[error("Malformed derivation path format")]
+    InvalidDerivationPathFormat,
+    #[error("Unsupported key type")]
+    UnsupportedKeyType,
+    #[error("Unsupported child number type")]
+    UnsupportedChildNumberType,
+}
+
+/// BIP32-like child numbers. Currently we only support hardened derivations
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Hash)]
```
TheQuantumPhysicist, line 47: Please remove Hash. It's not needed.

crypto/src/key/hdkd/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+use schnorrkel::ExpansionMode::Ed25519;
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+const PREFIX: &str = "m";
+const SEPARATOR: &str = "/";
+const HARDENED_BIT: u32 = 0x80000000;
+
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+#[non_exhaustive]
```
TheQuantumPhysicist, line 30: Please avoiding such directives

crypto/src/key/hdkd/mod.rs
erasmospunk: Done

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,248 @@
...
+            DerivationPath::from_str("m/0'/1'/2'/2'/1000000000'"),
+            Ok(vec![
+                ChildNumber::hardened(0).unwrap(),
+                ChildNumber::hardened(1).unwrap(),
+                ChildNumber::hardened(2).unwrap(),
+                ChildNumber::hardened(2).unwrap(),
+                ChildNumber::hardened(1000000000).unwrap(),
+            ]
+            .into())
+        );
```
iljakuklic, line 246: Maybe tests for `"m/2147483647'"` and `m/2147483648'` should be added

iljakuklic (commented):
Adding some comments. Can't say much about security implications.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,248 @@
...
+            ChildNumber::Hardened(_) => true,
+        }
+    }
+}
+
+impl FromStr for ChildNumber {
+    type Err = DerivationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // Check if this child number is hardened
```
TheQuantumPhysicist, line 72: Please treat these strictly as ascii bytes. Like I mentioned before, we should not support any unicode or other weird encoding string that works just by coincidence. There's no reason to support it. This is a recipe for problems. This whole wallet system is based on ascii. Let's try to follow that for simplicity.

The correct way to do this is to start by converting this to bytes, check that all the bytes are ascii, and only then start processing.

I do recognize that you added tests for specific cases, but I don't think we can test all possibilities because encoding is a mess in general.

I don't know if I'm missing something here that makes all this OK. But I have to be cautious.

crypto/src/key/rschnorr/mod.rs
```diff
@@ -129,6 +134,27 @@ impl MLRistrettoPrivateKey {
     }
 }
 
+impl ChildNumber {
```
TheQuantumPhysicist, line 137: This is not really idiomatic rust. We shouldn't have struct impl blocks outside their modules.

crypto/src/key/mod.rs
```diff
@@ -102,13 +104,29 @@ impl PrivateKey {
     }
 }
 
+impl Derivable for PrivateKey {
```
TheQuantumPhysicist, line 107: The problem with this assumption is that we're saying that *every* kind of PrivateKey now, and in the future, must support being derivable. Not sure this is a good assumption, or even a tenable one.

crypto/src/key/hdkd/derivation_path.rs
iljakuklic: As far as I can tell, the parsing algorithm already only accepts ASCII strings. The trick is to specify what is accepted, rather than what is rejected, which already happens. The only possible exception might be third-party code like `u32::parse` which is used here. I expect it only accepts ASCII too but may be worth double checking.

iljakuklic: I just checked with the stdlib implementation of `u32::parse` and it only accepts ASCII chars. However, it seems it still allows the plus sign even in front of an unsigned number. That means we would accept strings like `"m/+42'"` (please check). Do we want this?

TheQuantumPhysicist: Nice catch! Nope, + isn't allowed.

TheQuantumPhysicist: One thing that also occurred to me... what about exponents? would something like 1e6 parse and return 1000000? That would be very ridiculous! 😄

Quite honestly, this is why I'm overly cautious with parsing strings that are meant to be read as bytes... there's always that corner case that you didn't think of... always some exotic case that'll cause trouble.

In C++, I would use `isdigit()` to check every character after the suffix as bytes. I'm sure there's an equivalent thing in Rust.

iljakuklic: > One thing that also occurred to me... what about exponents? would something like 1e6 parse and return 1000000? That would be very ridiculous! smile

Heh. No, from what I've seen, only `+` was there. Also, it may allow leading zeroes.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,248 @@
...
+        let is_hardened = s.ends_with('\'') || s.ends_with('h');
```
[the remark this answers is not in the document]
erasmospunk: Not against that

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+impl ChildNumber {
+    /// Return a hardened child number
+    pub fn hardened(index: u32) -> Result<Self, DerivationError> {
+        if index & HARDENED_BIT == 0 {
+            Ok(ChildNumber::Hardened(index))
+        } else {
+            Err(InvalidChildNumber(index))
+        }
```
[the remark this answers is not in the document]
erasmospunk: Did you mean checking these errors at run time?

Using a `u31` type will push the checking the result in to that type. I took the approach that @TheQuantumPhysicist suggested, wrapping a private enum in a struct:

```rust
pub struct ChildNumber(DerivationType);

enum DerivationType {
    Hardened(u32),
}
```

So this was it would be impossible to create a wrong kind of ChildNumber i.e. with the internal `u32` having the MSB set to 1. It is not exactly what you were referring to but let me know if this approach is better.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,248 @@
...
+        // Check if this child number is hardened
+        let is_hardened = s.ends_with('\'') || s.ends_with('h');
+        if is_hardened {
+            // This will never panic because we know that the string is at least 1 character long
+            // and it is always on the boundary of a UTF-8 code point
+            let (s, _) = s.split_at(s.len() - 1);
+            ChildNumber::hardened(s.parse().map_err(|_| InvalidChildNumberFormat)?)
+        } else {
+            ChildNumber::normal(s.parse().map_err(|_| InvalidChildNumberFormat)?)
+        }
```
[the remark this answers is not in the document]
erasmospunk: Thank you for this. Originally I wanted to write the code like that but I didn't know that it was possible to pass an array to `strip_suffix`.

crypto/src/key/hdkd/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+    InvalidDerivationPathFormat,
+    #[error("Unsupported derivation type")]
+    UnsupportedDerivationType,
+}
+
+pub trait Derivable: Sized + Clone {
+    /// Derive a child private key given a derivation path
+    fn derive_path(&self, path: &DerivationPath) -> Result<Self, DerivationError> {
+        let key = path.into_iter().try_fold(self.clone(), |key, num| key.derive_child(num))?;
+        Ok(key)
```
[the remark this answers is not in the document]
erasmospunk: Thanks for that, maybe it was a remnant from a previous version.

Regarding `self` vs `&self`, what are the pros and cons of each approach?

crypto/src/key/rschnorr/mod.rs
erasmospunk: Would just having a simple function `child_num_to_schnorrkel_chaincode` outside of an `impl` enough?

crypto/src/key/mod.rs
erasmospunk: That is correct. The assumption is that because `derive_child` is fallible it would not create any issues.

Another approach would be to have a new private key type `ExtendedPrivateKey` that behaves like a `PrivateKey`.

crypto/src/key/hdkd/derivation_path.rs
erasmospunk: The idea was that the `parse` to `u32` would fail on non ascii characters. Having said that, I don't have any issues with checking if the `str` is ascii. Better safe than sorry.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl From<Vec<ChildNumber>> for DerivationPath {
+    fn from(path: Vec<ChildNumber>) -> Self {
+        DerivationPath(path)
+    }
+}
+
+impl<'a> IntoIterator for &'a DerivationPath {
+    type Item = &'a ChildNumber;
+    type IntoIter = Iter<'a, ChildNumber>;
```
iljakuklic, line 113: Minor: I'd prefer this to be a fully qualified name `std::slice::Iter` rather than importing `Iter` into the module namespace. Just to be more explicit, `Iter` is quite an ambiguous name.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+}
+
+impl FromStr for ChildNumber {
+    type Err = DerivationError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if !s.is_ascii() {
+            return Err(InvalidDerivationPathFormat);
+        }
```
iljakuklic, line 79: FYI: We have a macro `utils::ensure!` for this pattern.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+}
+
+impl fmt::Display for ChildNumber {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self.0 {
+            DerivationType::Hardened(index) => {
+                fmt::Display::fmt(&index, f)?;
+                let alt = f.alternate();
+                f.write_str(if alt { "h" } else { "'" })
```
iljakuklic, line 95: `HARDENED_H` and `HARDENED_APOS` could be used here

crypto/src/key/hdkd/mod.rs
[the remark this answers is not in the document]
iljakuklic: > Regarding `self` vs `&self`, what are the pros and cons of each approach?

The tradeoffs here may be quite subtle. With `&self` you avoid some cloning/copying but you `.clone()` explicitly here so this advantage is not as pronounced in this case. My motivation for using `self` here is to avoid the initial `.clone()` and also that the previous iteration produces `self` which can be fed into the next iteration so you already have a value there. Also it makes your trait more general. Since you don't require `T: Clone`, it is in principle implementable for more types. I am not sure if that is useful since keys are just byte arrays that tend to be trivially clonable. However, I can imagine a hypothetical scenario where we deliberately avoid implementing `Clone` on some key-like type to let Rust help us to avoid a key reuse when it is undesirable.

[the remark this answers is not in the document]
TheQuantumPhysicist: Usually in rust, you're explicitly cloning, in the sense that you force the user of the library to clone in the implementation, then that option can be given to the user. Using self here is a good example of how that works. If the user wants to clone, they can clone the outer object instead of consuming this one.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -0,0 +1,260 @@
...
+        // Check if this child number is hardened
+        if let Some(s) = s.strip_suffix([HARDENED_APOS, HARDENED_H]) {
+            ChildNumber::hardened(s.parse().map_err(|_| InvalidChildNumberFormat)?)
+        } else {
+            ChildNumber::normal(s.parse().map_err(|_| InvalidChildNumberFormat)?)
+        }
+    }
+}
+
+impl fmt::Display for ChildNumber {
```
TheQuantumPhysicist, line 89: Do we want to instead implement ToString, and make this just call that?

crypto/src/key/hdkd/derivation_path.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Just to clarify, there's also a NonZeroU32, NonZeroU64, etc in rust that solve a similar problem. The benefit here is to ensure that invariants are held across many functions and environments. For example, if function A calls B, and B calls C, ideally you'd have to check the u32 parameter in every call, which is cumbersome. But then if you enforce using U31, then you're guaranteed to do this only once by extracting the logic outside of these functions.

crypto/src/key/rschnorr/mod.rs
TheQuantumPhysicist: There probably should be a trait for this, which you implement for the schnorrkel key.

crypto/src/key/hdkd/derivation_path.rs
iljakuklic: The way it's laid out in the standard library, it's the other way around. This is defined:
```rust
impl<T: fmt::Display + ?Sized> ToString for T
```

Meaning if you implement `fmt::Display`, you get `ToString` for free.

TheQuantumPhysicist: Nice! Didn't know that! Cheers!

crypto/src/key/hdkd/derivation_path.rs
TheQuantumPhysicist: Let's skip that because I wanna avoid adding utils to crypto.

iljakuklic: Ok.

iljakuklic (approved):
This now looks good to me, in terms of code structure and general design. I am not a cryptography expert by any stretch so please wait for an approval by @TheQuantumPhysicist and/or @muursh.

TheQuantumPhysicist (approved):

PR #527, opened 2022-11-01
Fix utxo indexes for txs with burned outputs

azarovh:
The problem was that `filter` should be applied after `enumerate` to take skipped outputs into account.

Also few errors changed to `MissingOutputOrSpent` because we are now skipping burned outputs.

### DIFF
--- a/utxo/src/cache.rs
+++ b/utxo/src/cache.rs
@@ -179,9 +179,9 @@ impl<'a> UtxosCache<'a> {
 
         tx.outputs()
             .iter()
-            // burned outputs should not be included into utxo set
-            .filter(|output| !output.purpose().is_burn())
             .enumerate()
+            // burned outputs should not be included into utxo set
+            .filter(|(_, output)| !output.purpose().is_burn())
             .try_for_each(|(idx, output)| {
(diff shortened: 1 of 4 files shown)


utxo/src/cache.rs
```diff
@@ -179,9 +179,9 @@ impl<'a> UtxosCache<'a> {
 
         tx.outputs()
             .iter()
-            // burned outputs should not be included into utxo set
-            .filter(|output| !output.purpose().is_burn())
             .enumerate()
+            // burned outputs should not be included into utxo set
+            .filter(|(_, output)| !output.purpose().is_burn())
```
TheQuantumPhysicist, line 184: I don't think this makes a difference. You can put it back up there. It's even more readable.

TheQuantumPhysicist (approved):

utxo/src/cache.rs
azarovh: It actually does. I introduced a dedicated test for that. And it fails if I revert this line

TheQuantumPhysicist: Actually you're right. We shouldn't filter before we enumerate. Never mind.

PR #529, opened 2022-11-01
Implement peer banning for the mock backend

stanislav-tkach:


stanislav-tkach:
I'm not entirely satisfied with the result because of the following issues:
- The ugly `GetIp` workaround that is needed because we have different kinds of addresses (`std::net::SocketAddr`, `libp2p::Multiaddr`, `mock::transport::channel::Address`).
- IP ban cannot be implemented for the mock channels transport. I almost want to remove this transport implementation and the only thing that stops me is the time that I spent implementing it. 🙃 
- There is no easy and 100% correct way to get an IP address from `libp2p::Multiaddr`.

p2p/src/peer_manager/peerdb.rs
```diff
@@ -92,8 +121,11 @@ pub struct PeerDb<T: NetworkingService> {
     /// Pending connections
     pending: HashMap<T::Address, T::PeerId>,
 
-    /// Banned peers
-    banned: HashSet<T::PeerId>,
+    /// Banned addresses along with the duration of the ban.
+    ///
+    /// The duration represents the `UNIX_EPOCH + duration` time point, so the ban should end
+    /// when `current_time > ban_duration`.
+    banned: BTreeMap<IpAddr, Duration>,
```
azarovh, line 128: Can `T::Address` be stored here? As far as I get `IpAddr` is only required to unify different types of addresses. But maybe storing polymorpic type is enough?

p2p/src/net/types/mod.rs
```diff
@@ -211,3 +211,7 @@ pub enum ValidationResult {
     /// Message is not invalid but it shouldn't be forwarded to other peers
     Ignore,
 }
+
+pub trait GetIp {
```
TheQuantumPhysicist, line 215: Why don't you define an address type here as an associated type? Wouldn't that solve the problem you're having with channels?

p2p/src/peer_manager/peerdb.rs
TheQuantumPhysicist: Also, there can be another associated type for a "BannableAddress", if this isn't appropriate.

stanislav-tkach: . Currently we have three address types:
- `std::net::SocketAddr`
- `libp2p::MultiAddr`
- A channel address (`u64`)

The last one isn't important for now. A socket address consists of an IP address and a port and I think that we want to ban the "whole" address and not a specific address and port combination. The situation is even more trickier with libp2p because the address can be the following: `/ip4/7.7.7.7/tcp/4242/p2p/QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N` and once again I think that we are only interested in the IP address here.

p2p/src/net/types/mod.rs
stanislav-tkach: I'm not sure. Yes, some kind of an address can be introduced specifically for channels, but it would still have the same issues: it needs to be randomly generated (this can be done for `IpAddr` too) and it needs to be "persistent".

stanislav-tkach:
After the discussions with @azarovh and @TheQuantumPhysicist I have introduced a new `BannableAddress` abstraction. By the way, the `From` trait cannot be used instead of `AsBannableAddress` because both types (for example `MultiAddr` and `IpAddr`) are in "foreign" crates.

p2p/src/net/mod.rs
```diff
@@ -165,3 +182,11 @@ where
     /// Poll syncing-related event from the networking service
     async fn poll_next(&mut self) -> crate::Result<types::SyncingEvent<T>>;
 }
+
+// TODO: We can require infallible conversion after getting rid of `libp2p::MultiAddr`.
+/// Represents a fallible conversion to a bannable address.
+pub trait AsBannableAddress {
+    type BannableAddress;
+
+    fn as_bannable(&self) -> Option<Self::BannableAddress>;
```
TheQuantumPhysicist, line 191: This shouldn't return an Option... what are we saying, there are addresses we can never ban? Either the whole design around banning is wrong or there's something major we don't understand.

p2p/src/net/mod.rs
```diff
@@ -165,3 +182,11 @@ where
     /// Poll syncing-related event from the networking service
     async fn poll_next(&mut self) -> crate::Result<types::SyncingEvent<T>>;
 }
+
+// TODO: We can require infallible conversion after getting rid of `libp2p::MultiAddr`.
+/// Represents a fallible conversion to a bannable address.
+pub trait AsBannableAddress {
```
TheQuantumPhysicist, line 188: Even though I don't want this to return an Option, but the question still stands, why we aren't using TryFrom trait. After removing the Option, maybe we should just go with From/Into.

stanislav-tkach: Either type (in case of `From`/`TryFrom`  any of types) or trait must be in the current crate, otherwise the compiler disallows that:
```rust
only traits defined in the current crate can be implemented for types defined outside of the crate
   --> p2p/src/net/libp2p/mod.rs:150:1
    |
150 | impl From<Multiaddr> for IpAddr {
    | ^^^^^---------------^^^^^------
    | |    |                   |
    | |    |                   `IpAddr` is not defined in the current crate
    | |    `libp2p::Multiaddr` is not defined in the current crate
    | impl doesn't use only types from inside the current crate
```

p2p/src/net/mod.rs
stanislav-tkach: Technically `MultiAddr` can contain no IP address. For example, the following is a type-wise correct libp2p address: `/p2p/QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N` as well as `Multiaddr::empty()`.

TheQuantumPhysicist: But you see, that shouldn't be allowed. We have to think about this. ANY address coming to p2p should be bannable.

stanislav-tkach: In the current implementation such addresses are rejected during connection.

TheQuantumPhysicist: Then we add a test for that, and consider this an invariant, and make the return a value that we panic if not satisfied.

azarovh (approved):

TheQuantumPhysicist (approved):

PR #533, opened 2022-11-03
Refactor `TestBlockInfo`

pavel-kokolemin:
Move functions from `TestBlockInfo` to `TestFramework`, fix missed block reward UTXOs loading.


chainstate/test-framework/src/utils.rs
```diff
@@ -264,3 +265,42 @@ fn new_token_burn_output(
...
+        (
+            OutPointSourceId::Transaction(tx.transaction().get_id()),
+            tx.transaction().outputs().clone(),
+        )
+    }))
+    .collect();
+    utxos
+}
+
+impl TestFramework {
```
TheQuantumPhysicist, line 295: Please put this in framework.rs, with the other impls. It's unidiomatic to separate impls in different modules.

chainstate/test-framework/src/utils.rs
```diff
@@ -264,3 +265,42 @@ fn new_token_burn_output(
...
+    pub fn utxos_from_id(&self, id: Id<GenBlock>) -> TestBlockUtxos {
```
TheQuantumPhysicist, line 296: I would call this `utxos_of_block`

chainstate/test-framework/src/utils.rs
```diff
@@ -264,3 +265,42 @@ fn new_token_burn_output(
         OutputPurpose::Burn,
     )
 }
+
+// TODO: Replace by a proper UTXO set abstraction
+// (https://github.com/mintlayer/mintlayer-core/issues/312).
+pub type TestBlockUtxos = Vec<(OutPointSourceId, Vec<TxOutput>)>;
```
TheQuantumPhysicist, line 271: I would make this a map to protect from accidental duplication.

chainstate/test-framework/src/utils.rs
```diff
@@ -264,3 +265,42 @@ fn new_token_burn_output(
         OutputPurpose::Burn,
     )
 }
+
+// TODO: Replace by a proper UTXO set abstraction
```
TheQuantumPhysicist, line 269: This TODO can go, right?

TheQuantumPhysicist (approved):
The PR looks good. Just please fix the minor issues under the few comments I made.

chainstate/test-framework/src/utils.rs
pavel-kokolemin: You are right, fixed, thanks.

chainstate/test-framework/src/utils.rs
pavel-kokolemin: Sure, fixed.

iljakuklic (approved):

chainstate/test-framework/src/utils.rs
pavel-kokolemin: OK, updated. Though loading outputs looks a bit ugly in the tests now.

chainstate/test-framework/src/utils.rs
pavel-kokolemin: I suppose. Removed.

chainstate/test-suite/src/tests/fungible_tokens.rs
```diff
@@ -753,7 +753,8 @@ fn transfer_split_and_combine_tokens(#[case] seed: Seed) {
         let token_min_issuance_fee = tf.chainstate.get_chain_config().token_min_issuance_fee();
 
         // Issue a new token
-        let genesis_outpoint_id = TestBlockInfo::from_genesis(&tf.genesis()).txns[0].0.clone();
+        let genesis_outpoint_id =
```
azarovh, line 756: Agree that this looks ugly. But I think the intention here was to get `OutPointSourceId` and it can be simply rewritten as:
```rust 
let genesis_outpoint_id = OutPointSourceId::BlockReward(tf.genesis().get_id().into());
```
No need to collect all outputs to create an id.

chainstate/test-suite/src/tests/helpers/mod.rs
```diff
@@ -32,13 +32,13 @@ pub fn add_block_with_locked_output(
 ) -> (InputWitness, TxInput) {
     // Find the last block.
     let current_height = tf.best_block_index().block_height();
-    let prev_block_info = tf.block_info(current_height.into());
+    let prev_block_info = tf.utxos_of_block(tf.block_id(current_height.into()));
```
azarovh, line 35: `prev_block_info` -> `prev_block_outputs`

stanislav-tkach (approved):

chainstate/test-framework/src/utils.rs
```diff
@@ -264,3 +264,27 @@ fn new_token_burn_output(
         OutputPurpose::Burn,
     )
 }
+
+pub(crate) fn utxos_from_genesis(genesis: &Genesis) -> TestBlockUtxos {
```
TheQuantumPhysicist, line 268: Using `pub` is heavily discouraged. Usually we depend on the hierarchical design of the crate to limit the requirements of what's exposed. Make this `pub` and that's good enough.

chainstate/test-framework/src/utils.rs
```diff
@@ -264,3 +264,27 @@ fn new_token_burn_output(
...
+    [(
+        OutPointSourceId::BlockReward(genesis.get_id().into()),
+        genesis.utxos().to_vec(),
+    )]
+    .into_iter()
+    .collect()
+}
+
+pub(crate) fn utxos_from_block(blk: &Block) -> TestBlockUtxos {
```
TheQuantumPhysicist, line 277: Same here

chainstate/test-framework/src/framework.rs
```diff
@@ -38,6 +42,8 @@ pub struct TestFramework {
     pub time_value: Option<Arc<AtomicU64>>,
 }
 
+pub type TestBlockUtxos = BTreeMap<OutPointSourceId, Vec<TxOutput>>;
```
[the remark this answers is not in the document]
pavel-kokolemin: Fixed, thanks.

chainstate/test-framework/src/framework.rs
```diff
@@ -152,15 +152,17 @@ impl TestFramework {
...
+    /// Returns the list of UTXOs from the selected block.
     #[track_caller]
-    pub fn block_info(&self, height: u64) -> TestBlockInfo {
-        let id = self
-            .chainstate
-            .get_block_id_from_height(&BlockHeight::from(height))
-            .unwrap()
-            .unwrap();
-        TestBlockInfo::from_id(&self.chainstate, id)
+    pub fn utxos_of_block(&self, id: Id<GenBlock>) -> TestBlockUtxos {
```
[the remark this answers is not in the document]
pavel-kokolemin: Done, thanks.

chainstate/test-suite/src/tests/fungible_tokens.rs
pavel-kokolemin: Updated. Much better now, thanks!

chainstate/test-suite/src/tests/helpers/mod.rs
pavel-kokolemin: Done, thanks.

chainstate/test-framework/src/utils.rs
pavel-kokolemin: OK, thanks.

chainstate/test-framework/src/utils.rs
pavel-kokolemin: Fixed, thanks.

iljakuklic (approved):
Looks good. Please mind the clippy issue in `block_builder.rs`

chainstate/test-framework/src/utils.rs
azarovh: The following code got me thinking. 
```rust 
let issuance_outpoint_id =
             tf.outputs_from_genblock(block.get_id().into()).keys().next().unwrap().clone();
```
First I thought that it was not a good idea to switch to a map. Many tests rely on the ordering of txs. And I was really surprised that no tests failed because of this. And it just looked ugly anyway. 

But then the more I looked at that TestBlockInfo the more ugly it got. For example:
```rust
        let issuance_outpoint_id = TestBlockInfo::from_block(&block).txns[0].0.clone(); 
```
It's insane: collecting all the outputs from all the txs just to get id for the first tx! Instead of something like `block.transaction().get(0).get_id()`.

I think that in most cases `TestBlockInfo` is heavily misused (as well as `BlockOutputs` because it doesn't solve the root cause). Imo what we actually need is a  way to **get transaction id using block id and index**. This is handy because when constructing blocks with BlockBuilder + TransactionBuilder you won't get ids of txs to be able to use them later in the test.

I suggest adding dedicated method to test framework: `get_transaction(id: Id<Block>, index)`. With it I believe that 90% of the smelly code goes away.

`BlockOutputs` still can be used for other 10% of cases that actually use outputs.

P.S. there is almost exact copy of `TestBlockInfo` in `p2p` [dizzy]

stanislav-tkach: > P.S. there is almost exact copy of TestBlockInfo in p2p [dizzy]

Test client still isn't used in `p2p` because it (or the `p2p` code) requires some adaptation. We have an issue for that: https://github.com/mintlayer/mintlayer-core/issues/182

TheQuantumPhysicist: Answering you, @azarovh , this is what happens when someone created a very bad design and it trickles down everywhere. Hopefully we'll fix this. I think you're mostly right, but we want to give this some good thought. 

We probably do need a way to get all utxos. But if most of the calls just need a single utxo, simplifying this is a good call.

PR #538, opened 2022-11-07
In mempool, cache transaction fees to avoid repeatedly calculating them

royster57:


TheQuantumPhysicist (approved):

mempool/src/interface/mempool_interface_impl/pool/tx_with_fee.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl TxWithFee {
+    pub async fn new<M: GetMemoryUsage + Sync + Send>(
+        mempool: &Mempool<M>,
+        tx: &SignedTransaction,
+    ) -> Result<Self, TxValidationError> {
+        let fee = mempool.try_get_fee(tx).await?;
+        Ok(Self {
+            tx: tx.clone(),
```
iljakuklic, line 35: I think this method should take ownership of `tx` and let the caller handle the potential cloning. If the caller already owns the `tx` and does not need it for anything else, it can be just passed in without cloning.

mempool/src/interface/mempool_interface_impl/pool/tx_with_fee.rs
```diff
@@ -0,0 +1,47 @@
...
+            fee,
+        })
+    }
+
+    pub fn tx(&self) -> &SignedTransaction {
+        &self.tx
+    }
```
iljakuklic, line 42: Maybe also add `take_tx(self) -> SignedTransaction`

iljakuklic, line 42: This could be also exposed as an impl of `AsRef<SignedTransaction>` (or more controversially `Deref<Target = SignedTransaction>`

iljakuklic (approved):
Looks good, just some minor comments.

muursh (approved):

mempool/src/interface/mempool_interface_impl/pool/tx_with_fee.rs
royster57: Done.

mempool/src/interface/mempool_interface_impl/pool/tx_with_fee.rs
royster57: I don't currently see the need for this...

royster57: True, however I think having a designated method is clearer.

stanislav-tkach (approved):

mempool/src/interface/mempool_interface_impl/pool/mod.rs
```diff
@@ -789,6 +813,17 @@ where
...
         self.events_controller.subscribe_to_events(handler)
     }
+
+    pub fn new_tip_set(&mut self, block_id: Id<Block>, block_height: BlockHeight) {
+        log::info!(
+            "new tip with block_id {:?} and block_height {:?}",
+            block_id,
+            block_height
+        );
+        // TODO(Roy) handle the new tip
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Not necessarily. TODO(Name) means that the person in the name is responsible for it.

TODO(PR) means it should be addressed before merge.

PR #540, opened 2022-11-07
PoS accounting integration into TransactionVerifier

azarovh:
This is the first integration of `pos_accounting` into `chainstate`. It has tons of code to make it work, but the idea boils down to this: whenever there is a transaction output with `OutputPurpose::StakePool` - create an accounting pool via `PoSAccountingDelta::create_pool`. More operations will be supported in the future.

Some highlights to help you guys review: 
1. `TransactionVerifier`:
   - Fields for `PoSAccountingDelta` and `AccountsBlockUndoCache` were added.
   - Supported flushing and fetching information for new fields from one verifier to another and to/from storage
   - Moved logic related to `utxo::BlockUndo` cache to a separate module
   - Introduced `AccountsBlockUndoCache` which is almost a copy of `UtxoBlockUndoCache`
2. `chainstate_storage`. Added new indexes to the storage schema with required getters and setters 
3. `pos_accounting`:
   - `PoSAccountingDelta` is now generic in the same way as `UtxoCache`
   - Added `BlockUndo` which is almost a copy of `utxo::BlockUndo` but doesn't contain logic that checks dependencies.
   - Removed `PoSAccountingOperationsCache` struct. Imo it's useless and all the functionality is performed via `PoSAccountingDelta`
   - Removed `PoSAccountingOperatorRead` trait. It's redundant as it was a copy of a `PoSAccountingView`. In addition we already have `PoSAccountingStorageRead` trait,  which in combination with view covers all our needs.


pos_accounting/src/pool/delta/operator_impls.rs
```diff
@@ -280,62 +279,3 @@ impl<'a> PoSAccountingDelta<'a> {
         Ok(())
     }
 }
-
-impl<'a> PoSAccountingOperatorRead for PoSAccountingDelta<'a> {
```
TheQuantumPhysicist, line 284: I think it might be premature to delete this. Are you sure? Maybe we'll need it later when we introduce other operations on accounting.

pos_accounting/src/pool/block_undo.rs
```diff
@@ -0,0 +1,95 @@
...
+        &self.0
+    }
+
+    pub fn into_inner(self) -> Vec<PoSAccountingUndo> {
+        self.0
+    }
+}
+
+#[derive(Clone, Debug, Default, Eq, PartialEq, Encode, Decode)]
+pub struct BlockUndo {
```
TheQuantumPhysicist, line 48: Maybe we should give this a different name? `AccountingBlockUndo`, while the other one `UtxosBlockUndo?`

pos_accounting/src/pool/delta/operator_impls.rs
azarovh: Maybe I don't see the idea you put in it. As for me we have `PoSAccountingView` to get data from deltas, `PoSAccountingStorageRead` for storage. `PoSAccountingOperations` (previously PoSAccountingOperatorsWrite) to perform complex operations. But not sure why dedicated `PoSAccountingOperatorRead` is useful

pos_accounting/src/pool/block_undo.rs
azarovh: I used them as pos_accounting::BlockUndo and utxo::BlockUndo. But I don't mind to rename

utxo/src/storage/view_impls.rs
```diff
@@ -24,21 +24,17 @@ mod utxosdb_utxosview_impls {
     use super::*;
 
     pub fn utxo<S: UtxosStorageRead>(db: &S, outpoint: &OutPoint) -> Option<Utxo> {
-        db.get_utxo(outpoint).unwrap_or_else(|e| {
-            panic!(
-                "Database error while attempting to retrieve utxo from the database: {}",
-                e
-            )
-        })
+        db.get_utxo(outpoint).expect("Database error while attempting to retrieve utxo")
```
TheQuantumPhysicist, line 27: Why remove the extra info?

utxo/src/undo.rs
```diff
@@ -102,7 +102,9 @@ impl BlockUndo {
     ) -> Result<Self, BlockUndoError> {
         let mut block_undo = BlockUndo {
             reward_undo,
-            ..Default::default()
+            tx_undos: Default::default(),
```
TheQuantumPhysicist, line 105: Thank you for doing this!

utxo/src/storage/view_impls.rs
azarovh: It should be the same but simpler. As per documentation `expect` wll print the message from `Err`:

[[Panics](https://doc.rust-lang.org/std/result/enum.Result.html#panics)
Panics if the value is an [Err](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err), with a panic message including the passed message, and the content of the [Err] (https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err).](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect)

TheQuantumPhysicist: I see.

pos_accounting/src/pool/block_undo.rs
TheQuantumPhysicist: In principle it's nice, but the problem is that typically people using IDEs will just import the whole thing when they need it. That's why I'm not really super-excited about being able to distinguish between different types through scopes.

pos_accounting/src/pool/delta/operator_impls.rs
TheQuantumPhysicist: I forgot what I was planning back then, this was like more than 3 months ago. I don't mind removing it, I'm just worried we may need it later. You're free to remove it.

iljakuklic (approved):

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -582,6 +542,92 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
...
+                                )
+                                .map_err(ConnectTransactionError::PoSAccountingError)?;
+
+                            Ok(undo)
+                        }
+                        OutputPurpose::Transfer(_)
+                        | OutputPurpose::LockThenTransfer(_, _)
+                        | OutputPurpose::Burn => unreachable!(),
+                    }
+                },
```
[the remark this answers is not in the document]
azarovh: I need to propagate Err while mapping. Couldn't find better way to return Err and filter-map at the same time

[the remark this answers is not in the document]
iljakuklic: Hmm, that should be possible by returning from the `filter_map` closure:
* `None` if the item should be filtered out
* `Some(Err(e))` if propagating errors down the pipe
* `Some(Ok(blah))` in the success case

But maybe I missed something.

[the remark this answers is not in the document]
iljakuklic: This might work reasonably well:

Replace the `filter` with:
```rust
            .filter_map(|output| match output.purpose() {
                OutputPurpose::StakePool(data) => data,
                OutputPurpose::Transfer(_)
                | OutputPurpose::LockThenTransfer(_, _)
                | OutputPurpose::Burn => None,
            })
```

And in the subsequent `map`, you don't need to deal with the purpose matching and unreachables again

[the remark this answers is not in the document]
azarovh: you are right, thanks!

chainstate/tx-verifier/src/transaction_verifier/utxos_undo_cache.rs
```diff
@@ -0,0 +1,183 @@
...
+        self.data
+    }
+
+    pub fn read_block_undo<F>(
+        &self,
+        tx_source: &TransactionSource,
+        fetcher_func: F,
+    ) -> Result<BlockUndo, ConnectTransactionError>
+    where
+        F: Fn(Id<Block>) -> Result<Option<BlockUndo>, storage_result::Error>,
```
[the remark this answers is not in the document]
azarovh: tried to apply this, but have mixed feelings about it. Though it saves some characters the result is less clear and explicit for me

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
```diff
@@ -101,18 +107,23 @@ impl TransactionVerificationStrategy for DefaultTransactionVerificationStrategy
...
     where
         S: TransactionVerifierStorageRef,
         U: UtxosView,
-        M: Fn(&'a S, &'a ChainConfig, TransactionVerifierConfig) -> TransactionVerifier<'a, S, U>,
+        A: PoSAccountingView,
+        M: Fn(
+            &'a S,
+            &'a ChainConfig,
+            TransactionVerifierConfig,
+        ) -> TransactionVerifier<'a, S, U, A>,
```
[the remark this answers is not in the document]
azarovh: do you mean this? 
https://doc.rust-lang.org/beta/unstable-book/language-features/trait-alias.html

chainstate/tx-verifier/src/transaction_verifier/utxos_undo_cache.rs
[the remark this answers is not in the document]
iljakuklic: Ok. Having `something::Result<X>` as an alias to `Result<X, something::Error>` is a fairly common pattern (even used in standard lib, e.g. `std::io::Result`) so I guess it depends on how familiar are you with the pattern. Ultimately it's your call, do according to your judgement.

TheQuantumPhysicist:
Note: This work constantly takes from #260, and hence #260 will not be merged.

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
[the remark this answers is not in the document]
azarovh: added, looks nicer, thanks

chainstate/storage/src/lib.rs
```diff
@@ -81,10 +82,15 @@ pub trait BlockchainStorageRead: UtxosStorageRead {
...
     fn get_block_tree_by_height(&self) -> crate::Result<BTreeMap<BlockHeight, Vec<Id<Block>>>>;
+
+    /// Get accounting undo for specific block
+    fn get_accounting_undo(&self, id: Id<Block>) -> crate::Result<Option<AccountingBlockUndo>>;
 }
 
 /// Modifying operations on persistent blockchain data
-pub trait BlockchainStorageWrite: BlockchainStorageRead + UtxosStorageWrite {
+pub trait BlockchainStorageWrite:
+    BlockchainStorageRead + UtxosStorageWrite + PoSAccountingStorageWrite
```
TheQuantumPhysicist, line 92: Seems to me this should implement `PoSAccountingStorageRead` as well?

azarovh: `PoSAccountingStorageRead` is a supertrait of `PoSAccountingStorageWrite` so it's already there

chainstate/tx-verifier/src/transaction_verifier/accounts_undo_cache.rs
```diff
@@ -0,0 +1,158 @@
...
+    AccountingBlockUndo, AccountingBlockUndoError, AccountingTxUndo, PoSAccountingUndo,
+};
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct AccountsBlockUndo {
+    tx_undos: BTreeMap<Id<Transaction>, PoSAccountingUndo>,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct AccountsBlockUndoEntry {
```
TheQuantumPhysicist, line 35: Some times accounts... some times accounting... can we stick to Accounting everywhere?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -582,6 +542,87 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
         }
     }
 
+    pub fn connect_pos_accounting_outputs(
```
TheQuantumPhysicist, line 545: I don't think this should be pub

TheQuantumPhysicist: There might be a misunderstanding on what this does. If I understand correctly, this should be called when connecting a regular transaction.

Same applies for the disconnect function.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -582,6 +542,87 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
...
+            .filter_map(|output| match output.purpose() {
+                OutputPurpose::StakePool(data) => Some((data, output.value())),
+                OutputPurpose::Transfer(_)
+                | OutputPurpose::LockThenTransfer(_, _)
+                | OutputPurpose::Burn => None,
+            })
+            .map(
+                |(pool_data, output_value)| -> Result<PoSAccountingUndo, ConnectTransactionError> {
+                    let tx_id = tx.get_id();
+                    let total_inputs = self.calculate_total_inputs(tx.inputs())?;
```
TheQuantumPhysicist, line 562: This is a design flaw. The function `calculate_total_inputs` should be called exactly once, and then the returned amount should be drained through outputs. The way it stands now, there can be money printing, where in accounts, money is spent for delegation, and for utxos, the same money is spent. This has to be redesigned.

TheQuantumPhysicist (changes requested):
There's a major issue with inputs/outputs calculation that allows overspending. I added a comment about that.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -582,6 +542,87 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
...
+                        total_inputs.contains_key(&CoinOrTokenId::Coin),
+                        ConnectTransactionError::MissingCoinOutputToStake
+                    );
+                    ensure!(
+                        total_inputs.len() == 1,
+                        ConnectTransactionError::TokenInputForPoSAccountingOperation(tx_id)
+                    );
+                    // TODO: check StakePoolData fields
+
+                    let input0 = tx.inputs().get(0).expect("must be some");
```
TheQuantumPhysicist, line 573: Please avoid using expect in such short logic situations, because as demonstrated in the call, removing the upper part, which we'll remove, will basically break the invariant.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -582,6 +542,87 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
...
+            self.accounting_delta_undo
+                .get_or_create_block_undo(&tx_source)
+                .insert_tx_undo(tx.get_id(), pos_accounting::AccountingTxUndo::new(tx_undo))
+                .map_err(ConnectTransactionError::AccountingBlockUndoError)
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn disconnect_pos_accounting_outputs(
```
TheQuantumPhysicist, line 602: As discussed, please don't forget to test and verify that reorgs/disconnecting blocks restore the database state 100%.

chainstate/test-suite/src/tests/pos_accounting_reorg.rs
```diff
@@ -0,0 +1,171 @@
...
+            tf.make_block_builder()
+                .with_parent((*block_b.block_id()).into())
+                .add_transaction(tx_c)
+                .build_and_process()
+                .unwrap();
+
+            storage.read_accounting_data()
+        };
+
+        assert_eq!(storage.read_accounting_data(), expected_accounting_data);
```
azarovh, line 169: I haven't replaced this with full storage comparison, because reorg affects db leaving traces in DBBlock and DBBlockIndex. So I cannot compare chain a->b->c that was reorged with iconic chain a->b->c that was not.

I took this approach because I couldn't find other reasonable way to construct full storage without processing blocks.

TheQuantumPhysicist: Can we exclude these, or even better, add them manually to the result?

pos_accounting/src/pool/delta/operator_impls.rs
```diff
@@ -199,15 +198,26 @@ impl<'a> PoSAccountingDelta<'a> {
         self.get_pool_data(undo.pool_id)?
             .ok_or(Error::InvariantErrorPoolCreationReversalFailedDataNotFound)?;
 
-        self.data.pool_data.undo_merge_delta_data_element(undo.pool_id, undo_data)?;
+        // TODO: find better solution
+        // If pool_data is in the current cache then perform undo.
+        // Otherwise it's from parent so add `Delete` to the current cache,
+        // which effectively will perform undo for the parent on flush
```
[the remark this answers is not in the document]
azarovh: Undo has all the data required. The problem is that right now we cannot undo things that are missing in current cache and present in parent.
And I still haven't figure out a way to do it properly.

chainstate/test-suite/src/tests/pos_accounting_reorg.rs
azarovh: done

chainstate/src/detail/ban_score.rs
```diff
@@ -94,9 +95,14 @@ impl BanScore for ConnectTransactionError {
             ConnectTransactionError::InvariantErrorHeaderCouldNotBeLoadedFromHeight(_, _) => 100,
             ConnectTransactionError::BlockIndexCouldNotBeLoaded(_) => 100,
             ConnectTransactionError::TransactionVerifierError(err) => err.ban_score(),
-            ConnectTransactionError::BlockUndoError(_) => 100,
+            ConnectTransactionError::UtxoBlockUndoError(_) => 100,
             ConnectTransactionError::BurnAmountSumError(_) => 100,
             ConnectTransactionError::AttemptToSpendBurnedAmount => 100,
+            ConnectTransactionError::MissingPoSAccountingUndo(_) => 0,
+            ConnectTransactionError::PoSAccountingError(err) => err.ban_score(),
+            ConnectTransactionError::TokenInputForPoSAccountingOperation(_) => 100,
```
TheQuantumPhysicist, line 103: This doesn't seem to be used

chainstate/tx-verifier/src/transaction_verifier/error.rs
```diff
@@ -80,12 +81,22 @@ pub enum ConnectTransactionError {
...
     #[error("Failed to sum amounts of burns in transaction: {0}")]
     BurnAmountSumError(Id<Transaction>),
     #[error("Attempt to spend burned amount in transaction")]
     AttemptToSpendBurnedAmount,
+    #[error("PoS accounting error")]
+    PoSAccountingError(#[from] pos_accounting::Error),
+    #[error("PoS accounting undo is missing for transaction {0}")]
+    MissingPoSAccountingUndo(Id<Transaction>),
+    #[error("No token inputs is allowed for PoS accounting operations {0}")]
+    TokenInputForPoSAccountingOperation(Id<Transaction>),
```
TheQuantumPhysicist, line 97: This doesn't seem to be used

chainstate/tx-verifier/src/transaction_verifier/error.rs
```diff
@@ -80,12 +81,22 @@ pub enum ConnectTransactionError {
...
+    #[error("No token outputs is allowed for PoS accounting operations {0}")]
```
TheQuantumPhysicist, line 98: are allowed

chainstate/tx-verifier/src/transaction_verifier/error.rs
```diff
@@ -80,12 +81,22 @@ pub enum ConnectTransactionError {
...
+    TokenOutputForPoSAccountingOperation(Id<Transaction>),
```
TheQuantumPhysicist, line 99: I would call it "TokenOutputInPoSAccountingOperation", because the presence itself is the issue, not the association.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -611,16 +641,23 @@ where
         // verify input signatures
         self.verify_signatures(tx)?;
 
+        self.connect_pos_accounting_outputs(TransactionSource::from(tx_source), tx.transaction())?;
```
TheQuantumPhysicist, line 644: This could simply be `tx_source.into()`

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -511,77 +543,75 @@ where
...
+            .filter_map(|output| match output.purpose() {
+                OutputPurpose::StakePool(data) => Some((data, output.value())),
+                OutputPurpose::Transfer(_)
+                | OutputPurpose::LockThenTransfer(_, _)
+                | OutputPurpose::Burn => None,
+            })
+            .map(
+                |(pool_data, output_value)| -> Result<PoSAccountingUndo, ConnectTransactionError> {
+                    // TODO: check StakePoolData fields
+                    let input0 =
```
TheQuantumPhysicist, line 563: Why not take this outside the map? Seems to be common.

pos_accounting/src/pool/delta/operator_impls.rs
```diff
@@ -74,14 +74,14 @@ impl<'a> PoSAccountingOperatorWrite for PoSAccountingDelta<'a> {
     }
 
     fn decommission_pool(&mut self, pool_id: PoolId) -> Result<PoSAccountingUndo, Error> {
-        let last_amount = self
-            .get_pool_balance(pool_id)?
-            .ok_or(Error::AttemptedDecommissionNonexistingPoolBalance)?;
-
         let last_data = self
```
TheQuantumPhysicist, line 77: ... why?!

pos_accounting/src/pool/delta/operator_impls.rs
```diff
@@ -186,11 +186,11 @@ impl<'a> PoSAccountingOperatorWrite for PoSAccountingDelta<'a> {
     }
 }
 
-impl<'a> PoSAccountingDelta<'a> {
+impl<P: PoSAccountingView> PoSAccountingDelta<P> {
     fn undo_create_pool(&mut self, undo: CreatePoolUndo) -> Result<(), Error> {
         let (pledge_amount, undo_data) = match undo.data_undo {
             PoolDataUndo::DataDelta(v) => *v,
-            PoolDataUndo::Data(_) => unreachable!("incompatible PoolDataUndo supplied"),
+            PoolDataUndo::Data(_) => panic!("incompatible PoolDataUndo supplied"),
```
TheQuantumPhysicist, line 193: Why changing to panic?

azarovh: The idea was that this code is reachable if someone passes the argument with the wrong type. Maybe this should be an error

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -611,16 +641,23 @@ where
...
+            .insert_tx_undo(tx.transaction().get_id(), tx_undo)?;
+
         match tx_source {
-            TransactionSourceForConnect::Chain {
-                new_block_index: block_index,
-            } => {
+            TransactionSourceForConnect::Chain { new_block_index: _ } => {
+                // update tx index only for txs from main chain
+
                 // update tx index only for txs from main chain
```
TheQuantumPhysicist, line 661: Twice?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -155,23 +164,31 @@ impl<'a> TransactionSourceForConnect<'a> {
...
-    utxo_cache: UtxosCache<U>,
-    utxo_block_undo: BTreeMap<TransactionSource, BlockUndoEntry>,
     token_issuance_cache: TokenIssuanceCache,
-    best_block: Id<GenBlock>,
+
+    utxo_cache: UtxosCache<U>,
+    utxo_block_undo: UtxosBlockUndoCache,
+
+    accounting_delta: PoSAccountingDelta<A>,
+    accounting_delta_undo: AccountingBlockUndoCache,
```
TheQuantumPhysicist, line 186: I would call this `accounting_block_undo`, just for symmetry, since they both serve the same purpose from different angles.

I understand that the "block" terminology might be wrong, in which case I also would seek a better name than `utxo_block_undo` as well.

Not a priority, but maybe give it a thought.

TheQuantumPhysicist (approved):
Looking good. I don't see huge issues here. Please address the little issues I mentioned.

I urge others to review. This is an important PR.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -511,77 +543,74 @@ where
...
+    fn connect_pos_accounting_outputs(
         &mut self,
-        tx_source: &TransactionSource,
-        tx_id: &Id<Transaction>,
-    ) -> Result<TxUndo, ConnectTransactionError> {
-        let block_undo = self.fetch_block_undo(tx_source)?;
+        tx_source: TransactionSource,
+        tx: &Transaction,
+    ) -> Result<(), ConnectTransactionError> {
+        let input0 = tx.inputs().get(0).ok_or(ConnectTransactionError::MissingOutputOrSpent)?;
```
TheQuantumPhysicist, line 551: While taking this out of the map is the right thing to do from an abstraction point of view, keep in mind that it has a different effect that before. Before, it used to do this check only if the output purpose is StakePool. Now it does it all the time.

This may or may not be desired. It may be desired if it's an invariant that inputs can never be empty. If it's not desired, the input0 pulling can be done in a lambda that's called inside. Perhaps we should go with the latter, since we're not restricted by having no inputs in the context of PoS.

I'll push a fix for this.

chainstate/tx-verifier/src/transaction_verifier/tests/mock.rs
```diff
@@ -87,17 +98,58 @@ mockall::mock! {
...
         fn get_best_block_for_utxos(&self) -> Result<Option<Id<GenBlock>>,storage_result::Error>;
-        fn get_undo_data(&self, id: Id<Block>) -> Result<Option<BlockUndo>, storage_result::Error>;
+        fn get_undo_data(&self, id: Id<Block>) -> Result<Option<utxo::UtxosBlockUndo>, storage_result::Error>;
     }
 
     impl FlushableUtxoView for Store {
         fn batch_write(&mut self, utxos: ConsumedUtxoCache) -> Result<(), utxo::Error>;
     }
+
+    impl PoSAccountingView for Store{
```
iljakuklic, line 126: not formatted?

chainstate/tx-verifier/src/transaction_verifier/utxos_undo_cache.rs
```diff
@@ -0,0 +1,186 @@
...
+    primitives::Id,
+};
+use utxo::{UtxosBlockRewardUndo, UtxosBlockUndo, UtxosTxUndo};
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct UtxosBlockUndoEntry {
+    pub undo: UtxosBlockUndo,
+    // indicates whether this BlockUndo was fetched from the db or it's new
+    pub is_fresh: bool,
+}
```
iljakuklic, line 31: This seems to be a recurring pattern. Maybe it could be made generic with something like:
```rust
struct DbEntry<T> { data: T, is_fresh: bool }
```

And reuse that in various places. Not an issue for this PR though.

iljakuklic (approved):
Looks good. A couple of minor comments below.

chainstate/src/detail/chainstateref/tx_verifier_storage.rs
```diff
@@ -209,4 +220,105 @@ impl<'a, S: BlockchainStorageWrite, O: OrphanBlocks, V: TransactionVerificationS
...
+        undo: &AccountingBlockUndo,
+    ) -> Result<(), TransactionVerifierStorageError> {
+        match tx_source {
+            TransactionSource::Chain(id) => self
+                .db_tx
+                .set_accounting_undo_data(id, undo)
+                .map_err(TransactionVerifierStorageError::from),
+            TransactionSource::Mempool => {
+                panic!("Flushing mempool info into the storage is forbidden")
+            }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: That sounds like a nice idea and it deserves a shot. Let's register it somewhere so that we don't forget. Someone please open an issue.

If we can isolate all mempool vs chain operations at compile time, we'll gain a huge security advantage.

chainstate/tx-verifier/src/transaction_verifier/tests/mock.rs
TheQuantumPhysicist: Seems like a lovely macro...

chainstate/src/detail/chainstateref/tx_verifier_storage.rs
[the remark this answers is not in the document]
azarovh: done: #633

alfiedotwtf:
Even though it's already been merged, lgtm

PR #542, opened 2022-11-08
Add tx indexing flag

pavel-kokolemin:
Make tx index optional (for now without ability to change it).

### DIFF
--- a/chainstate/src/config.rs
+++ b/chainstate/src/config.rs
@@ -28,6 +28,7 @@ make_config_setting!(
         DEFAULT_MAX_IMPORT_BUFFER_SIZE,
     )
 );
+make_config_setting!(TxIndexEnabled, bool, true);
 
 /// The chainstate subsystem configuration.
 #[derive(Debug, Clone, Default)]
@@ -39,6 +40,7 @@ pub struct ChainstateConfig {
     /// When importing bootstrap file, this controls the buffer sizes (min, max)
     /// (see bootstrap import function for more information)
     pub min_max_bootstrap_import_buffer_sizes: MinMaxBootstrapImportBufferSizes,
+    pub tx_index_enabled: TxIndexEnabled,
 }
 
 impl ChainstateConfig {
--- a/chainstate/src/config.rs
+++ b/chainstate/src/config.rs
@@ -28,7 +28,7 @@ make_config_setting!(
         DEFAULT_MAX_IMPORT_BUFFER_SIZE,
     )
 );
-make_config_setting!(TxIndexEnabled, bool, true);
+make_config_setting!(TxIndexEnabled, bool, false);
 
 /// The chainstate subsystem configuration.
 #[derive(Debug, Clone, Default)]
--- a/chainstate/src/detail/mod.rs
+++ b/chainstate/src/detail/mod.rs
@@ -43,7 +43,7 @@ use std::sync::Arc;
 
 use itertools::Itertools;
 
-use chainstate_storage::{BlockchainStorage, Transactional};
+use chainstate_storage::{BlockchainStorage, BlockchainStorageRead, Transactional};
 use chainstate_types::{BlockIndex, GenBlockIndex, PropertyQueryError};
 use common::{
     chain::{block::BlockHeader, config::ChainConfig, Block},
@@ -161,6 +161,10 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V> 
             time_getter,
         );
 
+        chainstate
+            .process_tx_index_enabled_flag()
+            .map_err(crate::ChainstateError::from)?;
+
         if best_block_id.is_none() {
             chainstate
                 .process_genesis()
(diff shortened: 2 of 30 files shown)


stanislav-tkach (approved):

node/src/config_files/mod.rs
```diff
@@ -111,6 +111,7 @@ fn chainstate_config(
         max_db_commit_attempts,
         max_orphan_blocks,
         min_max_bootstrap_import_buffer_sizes,
+        tx_index_enabled: is_tx_index_enabled,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yeah... this doesn't seem necessary... please make them both the same name.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -519,6 +519,30 @@ impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S> {
             .undo
     }
 
+    fn get_tx_cache_ref(&self) -> Result<Option<&TxIndexCache>, ConnectTransactionError> {
+        let enabled = self
+            .storage_ref
+            .get_is_mainchain_tx_index_enabled()?
```
TheQuantumPhysicist, line 525: I don't think this is necessary. The way I imagine it, is that the flag is checked for consistency when chainstate is constructed, and then whether tx index is enabled is passed as configuration to TransactionVerifier. Maybe there should be a TransactionVerifier configuration object. In fact, we do need that for the mempool.

chainstate/tx-verifier/src/transaction_verifier/storage.rs
```diff
@@ -65,6 +67,10 @@ pub trait TransactionVerifierStorageRef: UtxosStorageRead {
         block_id: &Id<GenBlock>,
     ) -> Result<Option<GenBlockIndex>, storage_result::Error>;
 
+    fn get_is_mainchain_tx_index_enabled(
```
TheQuantumPhysicist, line 70: I think this has to go.

chainstate/src/detail/mod.rs
```diff
@@ -217,6 +219,38 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
             config_geneis_id == stored_genesis_id,
             InitializationError::GenesisMismatch(config_geneis_id, stored_genesis_id),
         );
+
+        Ok(())
+    }
+
+    /// Check that transaction index state is consistent between DB and config.
+    fn process_tx_indexing(&mut self) -> Result<(), BlockError> {
```
TheQuantumPhysicist, line 227: process_tx_index_enabled_flag, I would call this... does anyone have a better name?

node/src/config_files/mod.rs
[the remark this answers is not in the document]
pavel-kokolemin: It was originally `is_tx_index_enabled` but as a name of option in the `config.toml` file I like `tx_index_enabled` more.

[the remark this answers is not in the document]
TheQuantumPhysicist: That's OK. Just please make them both the same.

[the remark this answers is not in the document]
pavel-kokolemin: Fixed, thanks.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
pavel-kokolemin: I added new `TransactionVerifierConfig` struct which is passed to `TransactionVerifier`. Will that work?

I also added `allow(clippy::too_many_arguments)` in two places to suppress clippy warnings.

chainstate/tx-verifier/src/transaction_verifier/storage.rs
pavel-kokolemin: Replaced with new `TransactionVerifierConfig` config.

chainstate/src/detail/mod.rs
pavel-kokolemin: Sure, renamed.

chainstate/src/config.rs
```diff
@@ -28,6 +28,7 @@ make_config_setting!(
         DEFAULT_MAX_IMPORT_BUFFER_SIZE,
     )
 );
+make_config_setting!(TxIndexEnabled, bool, true);
```
TheQuantumPhysicist, line 31: While thinking about this, I think there's a design mistake here.

First, the default should be false. Because most users will use their wallets to index. However, this configuration is kind of invasive, because once enabled, people will never be able to run the software, ever again, without specifying the flag. I think this should have 3 different types:

- Enabled
- Disabled
- Unspecified

This way, users can set this only once on initialization, then forget about it.

While this sounds nice, it has the ugly side effect that we'll again have the problem of having to put the default of unspeficied somewhere. Maybe we should update `make_config_setting!` to support such cases. I'm not sure.

pavel-kokolemin: I changed default to false.

Maybe we could leave it as-is for now (without switching to tri-state)? Because if we implement re-indexing someday this won't be a problem.

chainstate/test-suite/src/tests/reorgs_tests.rs
```diff
@@ -81,7 +81,12 @@ fn reorg_simple(#[case] seed: Seed) {
 fn test_very_long_reorgs(#[case] seed: Seed) {
     utils::concurrency::model(move || {
         let mut rng = make_seedable_rng(seed);
-        let mut tf = TestFramework::default();
+        let mut tf = TestFramework::builder()
+            .with_chainstate_config(chainstate::ChainstateConfig {
+                tx_index_enabled: true.into(),
+                ..Default::default()
+            })
```
azarovh, line 88: I think for TestFramework we should make `true` the default value. It will increase coverage.

Also I'd suggest adding another case to `tx_verification_simulation.rs`. So that it runs with and without tx index.

pavel-kokolemin: > I think for TestFramework we should make true the default value. It will increase coverage.

Good idea. Fixed, thanks.

> Also I'd suggest adding another case to tx_verification_simulation.rs. So that simulation runs with and without tx index.

Added, thanks.

azarovh (approved):

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -128,10 +128,22 @@ pub struct TransactionVerifierDelta {
     token_issuance_cache: ConsumedTokenIssuanceCache,
 }
 
+#[derive(Clone)]
+pub struct TransactionVerifierConfig {
+    pub tx_index_enabled: bool,
+}
+
+impl TransactionVerifierConfig {
+    pub fn new(tx_index_enabled: bool) -> Self {
```
iljakuklic, line 137: It would be nice to have a custom enum for the parameter (and the struct field), like
```rust
enum TxIndexEnabled { Yes, No }
```

It's difficult to see what the parameter refers to if it's just `TransactionVerifierConfig::new(true)`. You can implement `From<bool>` for it if you like

TheQuantumPhysicist: We have multiple issues with this whole thing... I'll come back to this problem later.

TheQuantumPhysicist: See this for more details: https://github.com/mintlayer/mintlayer-core/pull/542#discussion_r1017612540

PR #543, opened 2022-11-09
Make utxos use generics instead of dyn so that TransactionVerifier can be Send + Sync for the mempool

TheQuantumPhysicist:


utxo/src/cache.rs
```diff
@@ -88,11 +106,11 @@ pub struct UtxosCache<'a> {
     memory_usage: usize,
 }
 
-impl<'a> UtxosCache<'a> {
+impl<'a, P: UtxosView> UtxosCache<'a, P> {
     #[cfg(test)]
-    pub fn new_for_test(best_block: Id<GenBlock>, view: &'a dyn UtxosView) -> Self {
+    pub fn new_for_test(best_block: Id<GenBlock>, parent: UtxosViewCow<'a, P>) -> Self {
```
azarovh, line 111: not directly related, but I think we can remove this function at all. The only difference that it creates cache with specific best block id. Default id::zero from `EmptyUtxosView` is no different for our tests. If needed one can always call `set_best_block`. But without this function `UtxosViewCow` wouldn't be exposed to the rest of the code and will make things cleaner.

azarovh (commented):
looks good. I don't see better solution

utxo/src/cache.rs
TheQuantumPhysicist: Done

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
```diff
@@ -50,11 +51,12 @@ impl TransactionVerificationStrategy for DefaultTransactionVerificationStrategy
         verifier_config: TransactionVerifierConfig,
         block_index: &'a BlockIndex,
         block: &WithId<Block>,
-    ) -> Result<TransactionVerifier<'a, S>, BlockError>
+    ) -> Result<TransactionVerifier<'a, S, P>, BlockError>
     where
         H: BlockIndexHandle,
         S: TransactionVerifierStorageRef,
-        M: Fn(&'a S, &'a ChainConfig, TransactionVerifierConfig) -> TransactionVerifier<'a, S>,
+        P: UtxosView,
```
azarovh, line 58: minor thing but maybe replace P with UV. I guess P stands for parent but in case of verifier it's not that obvious

azarovh (approved):

iljakuklic (approved):

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
TheQuantumPhysicist: Done.

PR #544, opened 2022-11-09
storage: LMDB memory map resizing

iljakuklic:
At the moment this is just a dead simple approach that uses locks to make sure transactions and memory map resize do not happen concurrently. There certainly is a room for optimization that can be discussed as a follow-up.

### DIFF
--- a/storage/lmdb/src/lib.rs
+++ b/storage/lmdb/src/lib.rs
@@ -68,7 +68,7 @@ impl<'tx, C: lmdb::Cursor<'tx>> Iterator for PrefixIter<'tx, C> {
 pub struct DbTx<'m, Tx> {
     tx: Tx,
     dbs: &'m DbList,
-    _map_token: RwLockReadGuard<'m, remap::MemMapToken>,
+    _map_token: RwLockReadGuard<'m, remap::MemMapController>,
 }
 
 type DbTxRo<'a> = DbTx<'a, lmdb::RoTransaction<'a>>;
@@ -132,7 +132,7 @@ pub struct LmdbImpl {
     ///
     /// * Shared (read) lock is required for running transactions, both read-only and read-write.
     /// * Exclusive (write) lock is required to reallocate the memory map
-    map_token: Arc<RwLock<remap::MemMapToken>>,
+    map_token: Arc<RwLock<remap::MemMapController>>,
 
     /// Size required to write a transaction
     tx_size: MemSize,
@@ -240,13 +240,12 @@ impl backend::Backend for Lmdb {
         std::fs::create_dir_all(&self.path).map_err(error::process_io_error)?;
 
         // Set up LMDB environment
-        let environment = {
-            let mut env_builder = lmdb::Environment::new();
-            env_builder.set_max_dbs(desc.len() as u32);
-            env_builder.set_flags(self.flags);
-            env_builder.set_map_size(self.map_size.as_bytes());
-            env_builder.open(&self.path).or_else(error::process_with_err)?
-        };
+        let environment = lmdb::Environment::new()
+            .set_max_dbs(desc.len() as u32)
+            .set_flags(self.flags)
+            .set_map_size(self.map_size.as_bytes())
+            .open(&self.path)
+            .or_else(error::process_with_err)?;
 
         // Set up all the databases
         let dbs = desc
@@ -258,7 +257,7 @@ impl backend::Backend for Lmdb {
         Ok(LmdbImpl {
             env: Arc::new(environment),
             dbs,
-            map_token: Arc::new(RwLock::new(remap::MemMapToken::new())),
+            map_token: Arc::new(RwLock::new(remap::MemMapController::new())),
             tx_size: self.tx_size,
         })
     }
--- a/storage/lmdb/src/lib.rs
+++ b/storage/lmdb/src/lib.rs
@@ -14,6 +14,9 @@
 // limitations under the License.
 
 mod error;
+mod remap;
+
+pub use remap::MemSize;
 
 use std::path::PathBuf;
 
@@ -23,7 +26,7 @@ use storage_core::{
     info::{DbDesc, MapDesc},
     Data, DbIndex,
 };
-use utils::sync::Arc;
+use utils::sync::{Arc, RwLock, RwLockReadGuard};
 
 /// Identifiers of the list of databases (key-value maps)
 #[derive(Eq, PartialEq, Debug, Clone)]
@@ -65,6 +68,7 @@ impl<'tx, C: lmdb::Cursor<'tx>> Iterator for PrefixIter<'tx, C> {
 pub struct DbTx<'m, Tx> {
     tx: Tx,
     dbs: &'m DbList,
+    _map_token: RwLockReadGuard<'m, remap::MemMapToken>,
 }
 
 type DbTxRo<'a> = DbTx<'a, lmdb::RoTransaction<'a>>;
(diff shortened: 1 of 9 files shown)


storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+        Self::from_kilobytes(1024 * megabytes)
+    }
+
+    /// Convert to raw byte count
+    pub const fn as_bytes(self) -> usize {
+        self.0
+    }
+
+    pub fn div_ceil(self, rhs: Self) -> usize {
+        self.0 / rhs.0 + (self.0 % rhs.0 > 0) as usize
```
stanislav-tkach, line 47: I was going to suggest using `self.0.div_ceil(rhs.0)`, but it is still unstable. 🙃

storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+        freelist_pages * page_size,
+        current_size,
+        free_pages * page_size,
+        required_size,
+    );
+
+    // Ensure there is enough space
+    if required_size > current_size {
+        // Remap with double the size but or the required size, whichever is larger
+        let new_size = std::cmp::max(2 * current_size, required_size);
```
stanislav-tkach, line 123: Could you please explain this logic to me? Why in one case (`required_size < 2 * current_size`) we want to have an additional capacity, but in other case (`required_size > 2 * current_size`) we don't?

stanislav-tkach (approved):

storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Support for memory map reallocation
+
+use logging::log;
+use utils::tap_error_log::LogError;
+
+/// Represensts LMDB memory map size
```
TheQuantumPhysicist, line 21: Represents

storage/lmdb/src/remap.rs
iljakuklic: The basic idea is to double the capacity if we run out. However, in some rare cases (when the initial size is small and the expected current tx size is large), that would not be sufficient. The `max` is there to remain correct even in these rare cases.

storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+pub type ExclusiveMemMapToken<'a> = std::sync::RwLockWriteGuard<'a, MemMapToken>;
+
+/// Memory remapping procedure. Ensure at least `headroom` free space is available
+pub fn remap(
+    env: &lmdb::Environment,
+    _map_token: ExclusiveMemMapToken<'_>,
+    headroom: MemSize,
+) -> storage_core::Result<()> {
+    // Get page size
+    let page_size = {
```
TheQuantumPhysicist, line 93: I get that this PR is supposed to be dead-simple, but this call is thread-safe... can't we have a way to call this and do the size check before acquiring the lock of the token, to avoid having to acquire multiple locks on every transaction?

storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+// limitations under the License.
+
+//! Support for memory map reallocation
+
+use logging::log;
+use utils::tap_error_log::LogError;
+
+/// Represents LMDB memory map size
+#[derive(Eq, PartialEq, PartialOrd, Ord, Clone, Copy, Debug)]
+pub struct MemSize(usize);
```
TheQuantumPhysicist, line 23: Using usize here isn't really a good idea... if for any reason it's 32-bit (like if we ever decide to support 32-bit), everything will fall apart. Please use u128 or u64 at least.

I know lmdb-rs uses usize... but this is a mistake. I already got into trouble because of this in C++.

storage/lmdb/src/lib.rs
```diff
@@ -180,21 +240,26 @@ impl backend::Backend for Lmdb {
...
 
         // Set up LMDB environment
-        let mut env = lmdb::Environment::new();
-        env.set_max_dbs(desc.len() as u32);
-        env.set_flags(self.flags);
-        //env.set_map_size(todo!());
-        //env.set_max_readers(todo!());
-        let env = env.open(&self.path).or_else(error::process_with_err)?;
+        let environment = {
+            let mut env_builder = lmdb::Environment::new();
```
TheQuantumPhysicist, line 244: I wouldn't make this mut then use it. Just put all the other commands after new().

storage/lmdb/src/lib.rs
```diff
@@ -180,21 +240,26 @@ impl backend::Backend for Lmdb {
...
+            .map(|desc| Self::open_db(&environment, desc))
             .collect::<storage_core::Result<Vec<_>>>()
             .map(DbList)?;
 
-        let env = Arc::new(env);
-        Ok(LmdbImpl { env, dbs })
+        Ok(LmdbImpl {
+            env: Arc::new(environment),
+            dbs,
+            map_token: Arc::new(RwLock::new(remap::MemMapToken::new())),
```
TheQuantumPhysicist, line 261: I would call this MapSizeController or something... MemMapToken is too opaque.

storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+        let new_size = std::cmp::max(2 * current_size, required_size);
+        log::info!(
+            "Resizing LMDB memory map from {} to {}",
+            current_size,
+            new_size,
+        );
+        env.set_map_size(new_size.as_bytes())
+            .or_else(crate::error::process_with_err)
+            .log_err()?;
+    }
```
TheQuantumPhysicist, line 132: At this point we call the hook.

TheQuantumPhysicist (approved):
Generally this looks alright. There's some room for improvement; it's too simplistic, just like the description. The only alarming problem is using usize for sizes, which is a big mistake. u128 is the best to go for.

storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+
```
TheQuantumPhysicist, line 133: Can you add a debug assert that checks the new size here?

storage/lmdb/src/remap.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub fn div_ceil(self, rhs: Self) -> u64 {
+        self.0 / rhs.0 + (self.0 % rhs.0 > 0) as u64
+    }
+}
+
+impl std::fmt::Display for MemSize {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}B", self.0)
```
TheQuantumPhysicist, line 53: The crate `byte_unit` can do this nicely for you.

storage/lmdb/src/remap.rs
stanislav-tkach: I got that, but would it make sense to double `required_size` too?..

iljakuklic: > I got that, but would it make sense to double `required_size` too?..

Yeah why not, will do

storage/lmdb/src/remap.rs
iljakuklic: Do you think it's worth pulling in given it's just used for logging?

TheQuantumPhysicist: Meh... I don't know... probably not.

muursh (approved):
Looks fine to me

Issue #541, opened 2022-11-08
Add deterministic PrivateKey generation function for tests


azarovh:
In our thrive on making tests 100% deterministic `PrivateKey::new` function that uses `make_true_rng` should be replaced with something that uses `make_seedable_rng`. Otherwise resulting `PublicKey` can affect hashes of transactions and blocks.

pavel-kokolemin:
The PR is merged so I think this could be closed now.

PR #545, opened 2022-11-09
Make private keys deterministic in tests

pavel-kokolemin:
Make private keys generation deterministic in tests (from the @TheQuantumPhysicist request).
I switched to using seeded RNG where it was available.
There are some places (not many) where it's not available.
I did not change keys generations in that parts.

### DIFF
--- a/crypto/src/key/mod.rs
+++ b/crypto/src/key/mod.rs
@@ -23,6 +23,7 @@ use serialization::{Decode, Encode};
 use crate::key::rschnorr::{MLRistrettoPrivateKey, MLRistrettoPublicKey, RistrettoSignatureError};
 use crate::key::Signature::RistrettoSchnorr;
 use crate::random::make_true_rng;
+use crate::random::{CryptoRng, Rng};
 pub use signature::Signature;
 
 use self::hdkd::child_number::ChildNumber;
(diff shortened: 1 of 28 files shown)


TheQuantumPhysicist:
Closes #541

TheQuantumPhysicist:
> There are some places (not many) where it's not available.
> I did not change keys generations in that parts.

If it's not available because the test doesn't have rng, please add it, just like the other tests. I, personally, had to add many when I supported this for block time.

test-utils/src/nft_utils.rs
```diff
@@ -23,14 +23,12 @@ use common::{
...
-    key::{KeyKind, PrivateKey},
-    random::Rng,
-};
+use crypto::key::{KeyKind, PrivateKey};
+use crypto::random::{CryptoRng, Rng};
 use serialization::extras::non_empty_vec::DataOrNoVec;
 
-pub fn random_creator() -> TokenCreator {
-    let (_, public_key) = PrivateKey::new(KeyKind::RistrettoSchnorr);
+pub fn random_creator(rng: &mut (impl Rng + CryptoRng)) -> TokenCreator {
```
TheQuantumPhysicist, line 30: Thank you for changing this!

crypto/src/key/mod.rs
```diff
@@ -64,10 +65,16 @@ impl From<RistrettoSignatureError> for SignatureError {
 
 impl PrivateKey {
     pub fn new(key_kind: KeyKind) -> (PrivateKey, PublicKey) {
```
TheQuantumPhysicist, line 67: Can you rename this to `new_from_entropy`?

TheQuantumPhysicist (approved):
Looking fine. Please try to eliminate all the entropy-based derivations of the private key you can.

stanislav-tkach (approved):

crypto/src/key/mod.rs
pavel-kokolemin: Updated, thanks.

pavel-kokolemin:
> Looking fine. Please try to eliminate all the entropy-based derivations of the private key you can.

Update in all remaining tests.

PR #546, opened 2022-11-10
Propagate announcements

### LINKED ISSUE #513: Propagate announcements to other peers
When a node receives an announcement from a peer it should check and propagate it to other peers.

stanislav-tkach:
Closes #513.

### DIFF
--- a/p2p/src/message.rs
+++ b/p2p/src/message.rs
@@ -113,7 +113,39 @@ pub enum Response {
 }
 
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
-pub enum Announcement {
+pub struct Announcement {
+    announcement: AnnouncementType,
+    /// This flag distinguishes between an original announcement and announcement propagated from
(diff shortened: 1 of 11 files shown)


p2p/src/message.rs
```diff
@@ -113,7 +113,39 @@ pub enum Response {
 }
 
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
-pub enum Announcement {
+pub struct Announcement {
+    announcement: AnnouncementType,
+    /// This flag distinguishes between an original announcement and announcement propagated from
+    /// another peer.
+    is_propagated: bool,
```
TheQuantumPhysicist, line 120: This distinction shouldn't exist... something is wrong, most probably.

TheQuantumPhysicist (commented):
Unit tests are not working in this PR. The tests: 
```
announcements_propagation_mock_tcp
announcements_propagation_libp2p
```

are taking forever

PR #547, opened 2022-11-10
Fix minor TODO in `outputs_from_block` from tests

pavel-kokolemin:
Fix remaining TODO after recent TestBlockInfo refactor in `outputs_from_block`.
The fix is a bit ugly but I unable to find a better way.

There also was a fair [comment](https://github.com/mintlayer/mintlayer-core/pull/533#discussion_r1013330998) from @azarovh about collecting all outputs just to get an tx id. I replaced that with getting tx id directly in some places.

### DIFF
--- a/chainstate/test-framework/src/block_builder.rs
+++ b/chainstate/test-framework/src/block_builder.rs
@@ -113,7 +113,7 @@ impl<'f> BlockBuilder<'f> {
     }
 
     /// Returns regular transaction output(s) if any, otherwise returns block reward outputs
-    fn get_regular_block_outputs(outputs: BlockOutputs) -> BlockOutputs {
+    fn filter_outputs(outputs: BlockOutputs) -> BlockOutputs {
         let has_tx_outputs = outputs
             .iter()
             .any(|(output, _)| matches!(output, OutPointSourceId::Transaction(_)));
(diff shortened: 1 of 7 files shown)


chainstate/test-framework/src/block_builder.rs
```diff
@@ -112,6 +112,19 @@ impl<'f> BlockBuilder<'f> {
         self
     }
 
+    /// Returns regular transaction output(s) if any, otherwise returns block reward outputs
+    fn get_regular_block_outputs(outputs: BlockOutputs) -> BlockOutputs {
```
azarovh, line 116: Comment describes logic that is more complex then a function name suggests. Maybe if we need to distinguish outputs from reward and tx we should make `BlockOutputs` type a struct with 2 dedicated fields instead of mixing them in one collection? That would be more explicit. Does it make sense?

azarovh: looking again at this function I still have a filling that something should be done about it. Because it's kind of misleading: not only regular outputs can be returned but also reward. What if we rename it to `filter_outputs`?

azarovh (approved):
looks good to me. That thing from my previous comment bothers me but not blocking from merging it

stanislav-tkach (approved):

PR #550, opened 2022-11-10
Add randomness to test framework builder

TheQuantumPhysicist:


chainstate/test-framework/src/framework_builder.rs
```diff
@@ -57,7 +58,7 @@ pub struct TestFrameworkBuilder {
 
 impl TestFrameworkBuilder {
     /// Constructs a builder instance with values appropriate for most of the tests.
-    pub fn new() -> Self {
+    pub fn new(_rng: &mut (impl Rng + CryptoRng)) -> Self {
```
azarovh, line 61: should we use it to generate flag for tx index?

azarovh (approved):

chainstate/test-framework/src/framework_builder.rs
TheQuantumPhysicist: This is the first step.

stanislav-tkach (approved):

muursh (approved):
Looks fine to me

PR #551, opened 2022-11-10
Randomize tx_index_enabled in tests

pavel-kokolemin:
And make required fixes

### DIFF
--- a/chainstate/test-framework/src/framework_builder.rs
+++ b/chainstate/test-framework/src/framework_builder.rs
@@ -58,7 +58,7 @@ pub struct TestFrameworkBuilder {
 
 impl TestFrameworkBuilder {
     /// Constructs a builder instance with values appropriate for most of the tests.
-    pub fn new(_rng: &mut (impl Rng + CryptoRng)) -> Self {
+    pub fn new(rng: &mut (impl Rng + CryptoRng)) -> Self {
         let chain_config = ChainConfigBuilder::new(ChainType::Mainnet)
             .net_upgrades(NetUpgrades::unit_tests())
             .genesis_unittest(Destination::AnyoneCanSpend)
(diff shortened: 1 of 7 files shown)


azarovh (approved):

chainstate/test-framework/src/framework_builder.rs
```diff
@@ -67,7 +67,7 @@ impl TestFrameworkBuilder {
             max_db_commit_attempts: Default::default(),
             max_orphan_blocks: Default::default(),
             min_max_bootstrap_import_buffer_sizes: Default::default(),
-            tx_index_enabled: true.into(),
+            tx_index_enabled: rng.gen_bool(0.5).into(),
```
iljakuklic, line 70: I am somewhat uncomfortable that something called `new` is not deterministic. Not quite obvious from call sites but not sure what a better approach could be

chainstate/test-framework/src/framework_builder.rs
```diff
@@ -58,7 +58,7 @@ pub struct TestFrameworkBuilder {
 
 impl TestFrameworkBuilder {
     /// Constructs a builder instance with values appropriate for most of the tests.
-    pub fn new(_rng: &mut (impl Rng + CryptoRng)) -> Self {
+    pub fn new(rng: &mut (impl Rng + CryptoRng)) -> Self {
```
iljakuklic, line 61: I'd say the doc comment here should mention that `tx_index_enable` is determined randomly by default

iljakuklic (approved):

chainstate/test-framework/src/framework_builder.rs
TheQuantumPhysicist: I don't think this is helpful because this will change in the future as more things will become random. Docs tend to not be updated.

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

PR #555, opened 2022-11-11
Speedup calculate_tx_index_from_block

pavel-kokolemin:
Calculate each transaction offset just once.

### DIFF
--- a/chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
+++ b/chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
@@ -21,7 +21,7 @@ use common::{
 };
 use tx_verifier::transaction_verifier::{config::TransactionVerifierConfig, error::TxIndexError};
 
-pub fn take_tx_index(
+pub fn take_front_tx_index(
     tx_indices: &mut Option<VecDeque<TxMainChainIndex>>,
 ) -> Option<TxMainChainIndex> {
     match tx_indices {
--- a/chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
+++ b/chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
@@ -1,0 +1,76 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::VecDeque;
+
+use common::{
+    chain::{calculate_tx_offsets_in_block, Block, TxMainChainIndex, TxMainChainIndexError},
+    primitives::Idable,
+};
+use tx_verifier::transaction_verifier::{error::TxIndexError, TransactionVerifierConfig};
+
+pub fn take_tx_index(
+    tx_indices: &mut Option<VecDeque<TxMainChainIndex>>,
+) -> Option<TxMainChainIndex> {
+    match tx_indices {
+        Some(v) => v.pop_front(),
+        None => None,
+    }
+}
+
+fn assert_tx_indices_sanity(
+    verifier_config: &TransactionVerifierConfig,
+    tx_indices: &Option<VecDeque<TxMainChainIndex>>,
+    block: &Block,
+) {
+    // Either tx_index is disabled, or enabled and tx count is equal to tx count in block
+    assert!(
+        !verifier_config.tx_index_enabled
+            || tx_indices.as_ref().expect("Guaranteed by verifier config").len()
+                == block.transactions().len()
+    );
+}
+
+pub fn construct_tx_indices(
+    verifier_config: &TransactionVerifierConfig,
+    block: &Block,
+) -> Result<Option<VecDeque<TxMainChainIndex>>, TxIndexError> {
+    let tx_indices = verifier_config
+        .if_tx_index_enabled(|| calculate_tx_offsets_in_block(block).map_err(TxIndexError::from));
+    let tx_indices = tx_indices?;
+
+    assert_tx_indices_sanity(verifier_config, &tx_indices, block);
+
+    Ok(tx_indices)
+}
+
+pub fn construct_reward_tx_indices(
+    verifier_config: &TransactionVerifierConfig,
+    block: &Block,
+) -> Result<Option<TxMainChainIndex>, TxIndexError> {
+    let result = verifier_config.if_tx_index_enabled(|| {
+        TxMainChainIndex::new(
+            block.get_id().into(),
+            block
+                .block_reward()
+                .outputs()
+                .len()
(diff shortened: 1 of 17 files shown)


common/src/chain/transaction/transaction_index/mod.rs
```diff
@@ -144,40 +144,23 @@ pub struct TxMainChainIndex {
...
-
-    TxMainChainIndex::new(
-        SpendablePosition::from(tx_position),
-        tx.outputs()
-            .len()
-            .try_into()
-            .expect("Number conversion from usize to u32 should not fail here (3)"),
-    )
+pub fn calculate_tx_offsets_in_block(block: &Block) -> Vec<u32> {
+    // Please add new test if new block format version is added
```
TheQuantumPhysicist, line 148: You don't need to add this comment. The match below will yell at whoever tries to ignore this ;-)

chainstate/tx-verifier/src/transaction_verifier/tx_index_cache.rs
```diff
@@ -48,10 +48,24 @@ impl TxIndexCache {
         self.data
     }
 
-    pub fn add_tx_index(&mut self, spend_ref: BlockTransactableRef) -> Result<(), TxIndexError> {
+    pub fn add_tx_index(
+        &mut self,
+        spend_ref: BlockTransactableRef,
+        tx_offsets: &[u32],
+    ) -> Result<(), TxIndexError> {
         let tx_index = match spend_ref {
```
TheQuantumPhysicist, line 56: Please debug_assert here that the lengths are equal.

TheQuantumPhysicist (approved):
Besides the minor things I mentioned, all good :-)

chainstate/tx-verifier/src/transaction_verifier/tx_index_cache.rs
pavel-kokolemin: Added, thanks.

common/src/chain/transaction/transaction_index/mod.rs
TheQuantumPhysicist: The better way is to create a superficial match there that triggers an error :wink:

stanislav-tkach (approved):

common/src/chain/transaction/transaction_index/mod.rs
```diff
@@ -144,40 +144,21 @@ pub struct TxMainChainIndex {
...
-    let tx_position = TxMainChainPosition::new(block.get_id(), offset_tx);
-
-    TxMainChainIndex::new(
-        SpendablePosition::from(tx_position),
-        tx.outputs()
-            .len()
-            .try_into()
-            .expect("Number conversion from usize to u32 should not fail here (3)"),
-    )
+pub fn calculate_tx_offsets_in_block(block: &Block) -> Vec<u32> {
```
azarovh, line 147: why not returning `Vec<TxMainChainIndex>`? Having raw offsets in u32 form all over the code is not the best imo

TheQuantumPhysicist: This sounds like a better solution. Pavel, do you mind trying it out?

TheQuantumPhysicist:
There's actually an issue I noticed now. It's that we have to calculate and pass the offset even if we don't need it because the transaction index is disabled... if we want to calculate it only if we need it, then we have to uphold that invariant, which is ugly...

I don't know what the best solution could be.

common/src/chain/transaction/transaction_index/mod.rs
TheQuantumPhysicist: Now I also noticed that if we pass the index by reference, we'll have to copy it. It seems that the reasonable solution is to pass the tx index of a transaction while connecting it, not the whole vector as a slice.

pavel-kokolemin: > This sounds like a better solution. Pavel, do you mind trying it out?

Sure, updated.

> Now I also noticed that if we pass the index by reference, we'll have to copy it. It seems that the reasonable solution is to pass the tx index of a transaction while connecting it, not the whole vector as a slice.

Updated. Looks better now, thanks.

pavel-kokolemin:
> There's actually an issue I noticed now. It's that we have to calculate and pass the offset even if we don't need it because the transaction index is disabled... if we want to calculate it only if we need it, then we have to uphold that invariant, which is ugly...
> I don't know what the best solution could be.

Do I need to make `calculate_tx_offsets_in_block` call optional? It's indeed a bit ugly.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -700,11 +720,11 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
     pub fn connect_transactable(
         &mut self,
         block_index: &BlockIndex,
-        spend_ref: BlockTransactableRef,
+        spend_ref: BlockTransactableWithIndexRef,
```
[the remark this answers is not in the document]
azarovh: sorry, won't work, never mind

TheQuantumPhysicist:
I revamped the whole solution around TransactionVerifierStrategy. Besides the unnecessary cloning of the transaction index, the primary fear is to have inconsistent state of the transaction index taken from the configuration. If transaction index is disabled, there's no need to calculate the transaction indices. Holding this invariant consistently is the goal here.

pavel-kokolemin:
> I revamped the whole solution around TransactionVerifierStrategy. Besides the unnecessary cloning of the transaction index, the primary fear is to have inconsistent state of the transaction index taken from the configuration. If transaction index is disabled, there's no need to calculate the transaction indices. Holding this invariant consistently is the goal here.

Looks good, thanks.

chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
```diff
@@ -0,0 +1,76 @@
...
+use common::{
+    chain::{calculate_tx_offsets_in_block, Block, TxMainChainIndex, TxMainChainIndexError},
+    primitives::Idable,
+};
+use tx_verifier::transaction_verifier::{config::TransactionVerifierConfig, error::TxIndexError};
+
+pub fn take_tx_index(
+    tx_indices: &mut Option<VecDeque<TxMainChainIndex>>,
+) -> Option<TxMainChainIndex> {
+    match tx_indices {
```
azarovh, line 27:
```rust
tx_indices.map(|v| v.pop_from())
```

chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use std::collections::VecDeque;
+
+use common::{
+    chain::{calculate_tx_offsets_in_block, Block, TxMainChainIndex, TxMainChainIndexError},
+    primitives::Idable,
+};
+use tx_verifier::transaction_verifier::{config::TransactionVerifierConfig, error::TxIndexError};
+
+pub fn take_tx_index(
```
azarovh, line 24: `take_front_tx_index`?

azarovh (approved):

chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
TheQuantumPhysicist: map can't take a mutable reference.

chainstate/src/detail/tx_verification_strategy/tx_verification_strategy_utils.rs
TheQuantumPhysicist: Fair

stanislav-tkach (approved):

PR #556, opened 2022-11-11
mempool: Mempool constructor does not need to be async

iljakuklic:
The mempool object is initialized in a separate task, meaning the constructor itself does not need to be `async`.

### DIFF
--- a/mempool/src/interface/mempool_interface_impl.rs
+++ b/mempool/src/interface/mempool_interface_impl.rs
@@ -40,7 +40,7 @@ pub struct MempoolInterfaceImpl {
 }
 
 impl MempoolInterfaceImpl {
-    pub async fn new<M: GetMemoryUsage + Sync + Send + 'static>(
+    pub fn new<M: GetMemoryUsage + Sync + Send + 'static>(
         chain_config: Arc<ChainConfig>,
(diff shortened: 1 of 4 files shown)


TheQuantumPhysicist (approved):

muursh (approved):

stanislav-tkach (approved):
It is interesting that the compiler doesn’t produce the “a function doesn’t need to be async” warning.

iljakuklic:
> 

Just checked and Clippy has a lint for this [1] but it's disabled by default.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#unused_async

iljakuklic:
> > It is interesting that the compiler doesn’t produce the “a function doesn’t need to be async” warning.
> 
> Just checked and Clippy has a lint for this [1] but it's disabled by default.
> 
> [1] https://rust-lang.github.io/rust-clippy/master/index.html#unused_async

Interesting, it does not seem to flag up anything for me, probably broken?

TheQuantumPhysicist:
> > It is interesting that the compiler doesn’t produce the “a function doesn’t need to be async” warning.
> 
> Just checked and Clippy has a lint for this [1] but it's disabled by default.
> 
> [1] https://rust-lang.github.io/rust-clippy/master/index.html#unused_async

It did work for us before. It made me remove a few asyncs from p2p

PR #558, opened 2022-11-14
Refrain from taking exclusive lock on every transaction for map-resize check in lmdb

TheQuantumPhysicist:


TheQuantumPhysicist:
Converting this to draft because there seems to be a thread-safety issue within `lmdb-rs`. The current form of these fixes must be done above the thread-safety fixes.

The issue is visible with `two_concurrent_writers` test when executed multiple times.

TheQuantumPhysicist:
Replaced with #684

PR #560, opened 2022-11-14
mempool: Remove some occurrences `pub(..)`

iljakuklic:
Changing them to `pub` does not increase visibility since the module itself is private and all the `use` items in the parent are private too.

Had to move some struct definitions around a bit to make it work but nothing too big.

### DIFF
--- a/mempool/src/config.rs
+++ b/mempool/src/config.rs
@@ -15,19 +15,19 @@
 
 use std::time::Duration;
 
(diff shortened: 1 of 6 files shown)


muursh (approved):
All fine by me

TheQuantumPhysicist (approved):

azarovh (approved):

PR #562, opened 2022-11-15
Fix rust analyzer errors in p2p

pavel-kokolemin:
Rust analyzer shows some false positive errors in p2p integration tests:

From the rust analyzer logs:
```
[ERROR project_model::workspace] cyclic deps: p2p_test_utils(CrateId(263)) -> p2p(CrateId(257)), alternative path: p2p(CrateId(257)) -> p2p_test_utils(CrateId(263))
```
I moved some code from `p2p/p2p-test-utils` crate to `p2p/tests/p2p_tests_common/mod.rs` file. This fixes the problem for me.

### DIFF
--- a/p2p/backend-test-suite/src/ban.rs
+++ b/p2p/backend-test-suite/src/ban.rs
@@ -24,9 +24,10 @@ use p2p::{
     net::{
         self, types::SyncingEvent, ConnectivityService, NetworkingService, SyncingMessagingService,
     },
+    peer_manager::helpers::connect_services,
     sync::BlockSyncManager,
 };
-use p2p_test_utils::{connect_services, MakeTestAddress, TestBlockInfo};
+use p2p_test_utils::{MakeTestAddress, TestBlockInfo};
 
 tests![invalid_pubsub_block, invalid_sync_block,];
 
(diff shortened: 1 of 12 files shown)


stanislav-tkach:
I think that this whole crate can be moved like this.

pavel-kokolemin:
> I think that this whole crate can be moved like this.

Sorry, I have a problem with this. Some helper functions are used in both unit and integration tests. For example `create_n_blocks` is used in `p2p/tests/ban.rs` (integration test) and also in `p2p/src/sync/tests/block_response.rs` (unit test). Perhaps `create_n_blocks` is belong to integration tests and the tests should be reorganized accordingly (`p2p/src/sync/tests/block_response.rs` to be moved to integration tests). After that we should be able move `p2p-test-utils` crate.

pavel-kokolemin:
`connect_services` is used in both integration and unit tests so I moved it as a public function to a new mod (`p2p::peer_manager::helpers`). This fixes the problem with rust analyzer.  I also removed one more `connect_services` copy.

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

muursh (approved):

PR #566, opened 2022-11-17
Speedup `block_announcement_too_big_message` test

pavel-kokolemin:
Most of the time spent in the test was constructing block with 200k empty transactions. I replaced that with constructing 1 big transaction instead. This results in 95% speedup when running only `block_announcement_too_big_message` tests (0.661s vs 11.683s) and 20% time speedup (16.937s vs 21.343s) when running all tests on my machine (i7-12700H).
I suspect there could be other offenders.

### DIFF
--- a/p2p/backend-test-suite/src/block_announcement.rs
+++ b/p2p/backend-test-suite/src/block_announcement.rs
@@ -20,8 +20,13 @@ use tokio::{pin, select, time::Duration};
 use common::{
     chain::{
         block::{consensus_data::ConsensusData, timestamp::BlockTimestamp, Block, BlockReward},
+        signature::{
+            inputsig::{InputWitness, StandardInputSignature},
+            sighashtype,
+        },
         transaction::signed_transaction::SignedTransaction,
         transaction::Transaction,
+        TxInput,
     },
     primitives::{Id, H256},
 };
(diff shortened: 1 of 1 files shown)


p2p/backend-test-suite/src/block_announcement.rs
```diff
@@ -183,12 +188,19 @@ where
...
 
-    let txs = (0..200_000)
-        .map(|_| {
-            SignedTransaction::new(Transaction::new(0, vec![], vec![], 0).unwrap(), vec![])
-                .expect("invalid witness count")
-        })
-        .collect::<Vec<_>>();
+    let input = TxInput::new(config.genesis_block_id().into(), 0);
+    let mut signature = Vec::new();
+    signature.resize(3 * 1024 * 1024, 0);
```
TheQuantumPhysicist, line 193: Please don't define mutable vectors then do operations on them when you can make them in one liners. Here's how you do this:

```
let signature = (0..3 * 1024 * 1024).into_iter().map(|_| 0).collect::<Vec<u8>>();
```

Rule of thumb: If there's `mut`, then maybe this can be fixed. Only use `mut` if there's absolutely no other way.

First best solution, no mut.
Second best solution, move mutations to a separate function and return something immutable.

TheQuantumPhysicist (approved):
Please fix the little thing then feel free to merge.

p2p/backend-test-suite/src/block_announcement.rs
TheQuantumPhysicist: I agree with the magic number thing, let's do that.

But the range I provided is correct. If you want size `N`, then it's `(0..N)`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1fa00a4836d43b58e8f84b27b1161b12

pavel-kokolemin: Updated, thanks for the hint.

PR #571, opened 2022-11-21
Use `Snowstorm` (noise protocol impl) in `TcpMockTransport`

pavel-kokolemin:
Add very basic `Encryption` trait with two implementations (`NoEncryption` and `NoiseEncryption`), use `NoiseEncryption` in `TcpMockTransport` by default.
`Encryption::handshake` will probably need to be updated later (`&self` added) to be able configure encryption parameters (for example remember local private key).


p2p/src/net/mock/transport/tcp.rs
```diff
@@ -73,16 +81,44 @@ impl MockListener<TcpMockStream, SocketAddr> for TcpListener {
 }
 
 pub struct TcpMockStream {
-    stream: TcpStream,
+    stream: snowstorm::NoiseStream<TcpStream>,
     buffer: BytesMut,
 }
 
+enum Side {
+    Initiator,
```
TheQuantumPhysicist, line 89: Inbound, Outbound

pavel-kokolemin: Sure, updated.

pavel-kokolemin:
Rekey is not implemented in `NoiseEncryption`. It should be done manually somehow (not sure if we really need it).
Here are some details in noise protocol specification: https://noiseprotocol.org/noise.html#rekey

pavel-kokolemin:
Also it's not clear what should we do with the public keys from peers.

p2p/src/net/mock/transport/tcp/encryption.rs
```diff
@@ -0,0 +1,84 @@
...
+use tokio::{
+    io::{AsyncRead, AsyncWrite},
+    net::TcpStream,
+};
+
+use super::Side;
+use crate::{error::P2pError, Result};
+
+#[async_trait]
+pub trait Encryption: Send {
```
TheQuantumPhysicist, line 27: StreamAdapter, I would call this.

p2p/src/net/mock/transport/tcp/encryption.rs
```diff
@@ -0,0 +1,84 @@
...
+    type Stream: AsyncRead + AsyncWrite + Send + Unpin;
+
+    /// Wraps base async TcpStream into AsyncRead/AsyncWrite stream that implements encryption.
+    async fn handshake(base: TcpStream, side: Side) -> Result<Self::Stream>;
+}
+
+#[derive(Debug)]
+pub struct NoEncryption {}
```
TheQuantumPhysicist, line 35: IdentityStreamAdapter, I'd call this.

p2p/src/net/mock/transport/tcp/encryption.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+static NOISE_HANDSHAKE_PATTERN: &str = "Noise_XX_25519_ChaChaPoly_SHA256";
+
+static NOISE_HANDSHAKE_PARAMS: once_cell::sync::Lazy<snowstorm::NoiseParams> =
+    once_cell::sync::Lazy::new(|| NOISE_HANDSHAKE_PATTERN.parse().expect("valid pattern"));
+
+#[derive(Debug)]
+pub struct NoiseEncryption {}
```
TheQuantumPhysicist, line 52: NoiseEncryptionAdapter

TheQuantumPhysicist: Also please move this to its own file (adapter/noise.rs)

p2p/src/net/mock/transport/tcp/encryption.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[async_trait]
+impl Encryption for NoiseEncryption {
+    type Stream = snowstorm::NoiseStream<TcpStream>;
+
+    async fn handshake(base: TcpStream, side: Side) -> Result<Self::Stream> {
+        // TODO: Check the data directory first, and use keys from there if available
+        let local_key = snowstorm::Builder::new(NOISE_HANDSHAKE_PARAMS.clone())
+            .generate_keypair()
```
TheQuantumPhysicist, line 61: Key-pair generation should be part of the struct, not part of the implementation. So the construction of the struct should contain this data, which can be loaded from a file or otherwise.

Please create a more elaborate design around this.

p2p/src/error.rs
```diff
@@ -164,6 +164,8 @@ pub enum P2pError {
     DatabaseFailure,
     #[error("Failed to convert data `{0}`")]
     ConversionError(ConversionError),
+    #[error("Noise protocol handshake error")]
+    NoiseHandshakeError,
```
stanislav-tkach, line 168: I think it would be better to store the original error here. At least in the form of a string for debug purposes.

p2p/src/net/mock/transport/tcp.rs
```diff
@@ -40,58 +42,75 @@ use crate::{
...
     async fn connect(address: Self::Address) -> Result<Self::Stream> {
-        let stream = TcpStream::connect(address).await?;
-        Ok(TcpMockStream::new(stream))
+        let tcp_stream = TcpStream::connect(address).await?;
+        let noise_stream = TcpMockStream::new(tcp_stream, Side::Outbound).await?;
+        Ok(noise_stream)
     }
 }
 
+pub struct TcpMockListener<E: Encryption>(TcpListener, std::marker::PhantomData<E>);
```
TheQuantumPhysicist, line 72: You don't need this trait-bound here

p2p/src/net/mock/transport/tcp.rs
```diff
@@ -40,58 +42,75 @@ use crate::{
     P2pError, Result,
 };
 
+use self::encryption::{Encryption, NoiseEncryption};
+
 #[derive(Debug)]
-pub struct TcpMockTransport {}
+pub struct TcpMockTransportBase<E: Encryption>(std::marker::PhantomData<E>);
```
TheQuantumPhysicist, line 48: Also here, no need for trait bound

p2p/src/net/mock/transport/tcp.rs
```diff
@@ -40,58 +42,75 @@ use crate::{
...
+
+// By default the transport uses Noise protocol encryption
```
TheQuantumPhysicist, line 50: No, it shouldn't. There's no default. The instantation of TcpMockStream should establish whether we want encryption or not.

pavel-kokolemin:
I noticed two major problems in `TcpMockListener::accept` after my changes:
1. `accept` become not cancel safe (new TCP connections might drop in the main loop if some other event received).
2. `accept` become blocking (event loop will not progress until noise handshake complete).

I will need to fix those two problems first.

p2p/src/net/mock/transport/tcp/encryption.rs
pavel-kokolemin: Sure, updated.

p2p/src/net/mock/transport/tcp/encryption.rs
pavel-kokolemin: Done, thanks.

p2p/src/net/mock/transport/tcp/encryption.rs
pavel-kokolemin: Generic `StreamConfig` added. It can be loaded when `NetworkingService` is started (default implementation is used in `NetworkingService::start` for now).

p2p/src/error.rs
pavel-kokolemin: Good idea, thanks. I did not add original error here because `snowstorm::Error` does not implement `PartialEq` and `Eq`, which is needed for `P2pError`. Replaced it with `String` here.

p2p/src/net/mock/transport/tcp.rs
pavel-kokolemin: Sure, removed. I just wanted to make it a little more readable.

p2p/src/net/mock/transport/tcp.rs
pavel-kokolemin: Default removed. Will that work?

pavel-kokolemin:
> I noticed two major problems in `TcpMockListener::accept` after my changes:
>     1. `accept` become not cancel safe (new TCP connections might drop in the main loop if some other event received).
>     2. `accept` become blocking (event loop will not progress until noise handshake complete).
> I will need to fix those two problems first.

I added generic high-level `Stream` which is upgraded from generic low-level `Socket` in the peer handshake process. `TcpMockListener::accept` is reverted like it was before. Both issues should be fixed now.

p2p/src/net/mock/transport/tcp/adapter/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use crate::{net::mock::peer::Role, Result};
+
+#[async_trait]
+pub trait StreamAdapter: Send {
+    type StreamConfig: Send + Sync + Default;
+
+    type Stream: AsyncRead + AsyncWrite + Send + Unpin;
+
+    /// Wraps base async TcpStream into AsyncRead/AsyncWrite stream that implements encryption.
```
TheQuantumPhysicist, line 32: "that may implement encryption"

p2p/src/net/mock/transport/tcp/adapter/noise.rs
```diff
@@ -0,0 +1,76 @@
...
+    once_cell::sync::Lazy::new(|| NOISE_HANDSHAKE_PATTERN.parse().expect("valid pattern"));
+
+#[derive(Debug)]
+pub struct NoiseEncryptionAdapter {}
+
+pub struct StreamConfig {
+    local_key: snowstorm::Keypair,
+}
+
+impl Default for StreamConfig {
```
TheQuantumPhysicist, line 36: There seems to be a misunderstanding regarding the workflow with keys here.

First, there's no default. Well, there's default for configuration typically, but you can't call "keys" configuration, then generate new keys on every instantiation by calling it a default.

Keys can come from two sources:
- From a file, as in from the data directory for someone who already initialized their node
- New key, which you're calling default.

In my opinion, instead of calling this configuration, you should call it "keys". So something like StreamKey, then you don't use "Default" for that at all. Instead, you either generate a new key, or you load it from a file.

TheQuantumPhysicist (commented):
We're working on a solution for the CI problem.

p2p/src/net/mock/transport/tcp/adapter/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    async fn handshake(
+        stream_config: &Self::StreamConfig,
+        base: TcpStream,
+        role: Role,
+    ) -> Result<Self::Stream>;
+}
+
+#[derive(Debug)]
+pub struct IdentityStreamAdapter {}
```
TheQuantumPhysicist, line 41: Please move this to identity.rs

p2p/src/net/mock/transport/tcp/mod.rs
```diff
@@ -40,58 +43,75 @@ use crate::{
     P2pError, Result,
 };
 
+use self::adapter::StreamAdapter;
+
 #[derive(Debug)]
-pub struct TcpMockTransport {}
+pub struct TcpMockTransport<E>(std::marker::PhantomData<E>);
```
TheQuantumPhysicist, line 49: Why are we owning `PhantomData<E>` instead of E?

p2p/src/net/mock/transport/tcp/adapter/noise.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use super::StreamAdapter;
+
+static NOISE_HANDSHAKE_PATTERN: &str = "Noise_XX_25519_ChaChaPoly_SHA256";
+
+static NOISE_HANDSHAKE_PARAMS: once_cell::sync::Lazy<snowstorm::NoiseParams> =
+    once_cell::sync::Lazy::new(|| NOISE_HANDSHAKE_PATTERN.parse().expect("valid pattern"));
+
+#[derive(Debug)]
+pub struct NoiseEncryptionAdapter {}
```
TheQuantumPhysicist, line 30: Why not put the key in here, and ditch the whole new `StreamConfig` object?

p2p/src/net/mock/transport/tcp/adapter/noise.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+#[async_trait]
+impl StreamAdapter for NoiseEncryptionAdapter {
+    type StreamConfig = StreamConfig;
+
+    type Stream = snowstorm::NoiseStream<TcpStream>;
+
+    async fn handshake(
+        stream_config: &Self::StreamConfig,
```
TheQuantumPhysicist, line 52: Why not pass `&self` instead of this config, and load the key from it?

p2p/src/net/mock/transport/tcp/adapter/noise.rs
pavel-kokolemin: Because `StreamAdapter` is only relevant to `TcpMockTransport`.
It is irrelevant to `ChannelMockTransport` because `ChannelMockTransport` does not work with `TcpStream`s.
But we still need somehow pass own keys from `MockTransport` to `TcpMockTransport`.
That's why I added generic `StreamConfig` type to `MockTransport`.

p2p/src/net/mock/transport/tcp/adapter/noise.rs
pavel-kokolemin: I think I answered that above.

p2p/src/net/mock/transport/tcp/mod.rs
```diff
@@ -40,58 +43,75 @@ use crate::{
...
-impl MockTransport for TcpMockTransport {
+impl<E: StreamAdapter + 'static> MockTransport for TcpMockTransport<E> {
     type Address = SocketAddr;
     type BannableAddress = IpAddr;
-    type Listener = TcpListener;
-    type Stream = TcpMockStream;
+    type Listener = TcpMockListener;
+    type Socket = TcpStream;
+    type Stream = TcpMockStream<E>;
+    type StreamConfig = E::StreamConfig;
```
TheQuantumPhysicist, line 58: If E contains StreamConfig, why redefine it here'?

p2p/src/net/mock/transport/tcp/adapter/mod.rs
pavel-kokolemin: Fixed, thanks.

p2p/src/net/mock/transport/tcp/adapter/noise.rs
pavel-kokolemin: Sure. For now only `gen_new` is added, but we can add later loading from file as well. Default is removed.

p2p/src/net/mock/transport/tcp/adapter/mod.rs
pavel-kokolemin: Done, thanks.

p2p/src/net/mock/transport/tcp/mod.rs
pavel-kokolemin: Not sure why we would want to own `StreamAdapter`, `StreamAdapter` trait has only static methods right now.

p2p/src/net/mock/transport/tcp/adapter/noise.rs
pavel-kokolemin: I tried again but don't see why it would be any better.
We would need to pass `StreamAdapter` around (instead of `StreamKey`) and it simply adds more code.

p2p/src/net/mock/transport/tcp/mod.rs
```diff
@@ -40,58 +48,130 @@ use crate::{
...
+
+pub struct TcpMockListener<E: StreamAdapter> {
+    receiver: UnboundedReceiver<(TcpMockStream<E>, SocketAddr)>,
+    local_address: SocketAddr,
+    join_handle: tokio::task::JoinHandle<()>,
+    _phantom: PhantomData<E>,
+}
+
+impl<E: StreamAdapter + 'static> TcpMockListener<E> {
+    async fn start(stream_key: &E::StreamKey, address: SocketAddr) -> Result<Self> {
```
TheQuantumPhysicist, line 86: Why pass the key by reference then clone it?

Rule of thumb: If you pass something by reference then clone it, then it makes sense to just pass it by value.

p2p/src/net/mock/transport/traits.rs
```diff
@@ -46,11 +47,14 @@ pub trait MockTransport: Send + 'static {
     /// A messages stream.
     type Stream: MockStream;
```
TheQuantumPhysicist, line 48: I'm having trouble understanding why StreamKey can't be just part of Stream... something doesn't add up.

p2p/src/net/mock/transport/tcp/mod.rs
pavel-kokolemin: Updated, thanks.

p2p/src/net/mock/transport/traits.rs
pavel-kokolemin: Moved to `MockStream`. Though referencing `StreamKey` become a bit convoluted ("ambiguous associated type" error):

```
-        stream_key: T::StreamKey,
+        stream_key: <<T as MockTransport>::Stream as MockStream>::StreamKey,
```

p2p/src/net/mock/transport/adapted.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::{error::P2pError, net::mock::peer::Role, Result};
+
+use super::{MockListener, MockTransport, StreamAdapter};
+
+// How much time is allowed to spend setting up (optionally) encrypted stream.
+const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Debug)]
+pub struct AdaptedMockTransport<S, T> {
+    stream_adapter: Arc<S>,
```
TheQuantumPhysicist, line 30: It's probably a bad idea to have this as an Arc. Why not have the stream object owned by the transport layer?

p2p/src/net/mock/transport/adapted.rs
```diff
@@ -0,0 +1,138 @@
...
+            receiver,
+            local_address,
+            join_handle,
+        })
+    }
+}
+
+impl<S: StreamAdapter<T::Stream>, T: MockTransport> Drop for AdaptedListener<S, T> {
+    fn drop(&mut self) {
+        self.join_handle.abort();
```
TheQuantumPhysicist, line 123: Is aborting join handles the correct way to do this? Usually you want to wait for it to finish or something. I'm not sure.

p2p/src/net/mock/transport/adapted.rs
```diff
@@ -0,0 +1,138 @@
...
+
+use crate::{error::P2pError, net::mock::peer::Role, Result};
+
+use super::{MockListener, MockTransport, StreamAdapter};
+
+// How much time is allowed to spend setting up (optionally) encrypted stream.
+const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Debug)]
+pub struct AdaptedMockTransport<S, T> {
```
TheQuantumPhysicist, line 29: Please add some documentation to what this is

p2p/src/net/mock/transport/adapted.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    async fn connect(&self, address: Self::Address) -> Result<Self::Stream> {
+        let base = self.base_transport.connect(address).await?;
+        let stream = self.stream_adapter.handshake(base, Role::Outbound).await?;
+        Ok(stream)
+    }
+}
+
+pub struct AdaptedListener<S: StreamAdapter<T::Stream>, T: MockTransport> {
```
TheQuantumPhysicist, line 66: Documentation... I don't know what I'm reading.

p2p/src/net/mock/transport/traits.rs
TheQuantumPhysicist: This is now irrelevant after the changes. Resolving.

p2p/src/net/mock/transport/adapted.rs
```diff
@@ -0,0 +1,300 @@
...
+const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Transport layer that wraps a lower-level transport layer (can be seen like an onion with multiple layer)
+/// Simplest version of this can be seen as a tcp transport layer, with an Identity stream_adapter. That would
+/// be equivalent to the tcp transport layer with nothing done to it.
+/// More layers can be added on top of this, with this struct, where we add encryption on top.
+#[derive(Debug)]
+pub struct WrappedTransportSocket<S, T> {
+    stream_adapter: Arc<S>,
+    base_transport: T,
```
TheQuantumPhysicist, line 41: One conclusion from our call is that having this owned by WrappedTransportSocket will result in the initialization of the base_transport being monolithic, which will make it impossible for keys to be imported from disk. This is a problem that already extends to the current implementation in libp2p, where the keys are also automatically generated. The keys being automatically (newly) generated on every connection is the only reason this is possible in its current form. Generally speaking, every stream layer should hand its prepared state to its parent in an owning fashion, and then the highest layer should go into the transport layer. This makes it possible to initialize keys separately for every layer in different ways, instead of in the current monolithic fashion.

This doesn't have to be fixed in this PR since this will touch the libp2p implementation, but this has to be fixed next.

p2p/src/net/mock/transport/adapted.rs
pavel-kokolemin: `Arc` was removed after recent changes (replacing task spawning).

p2p/src/net/mock/transport/adapted.rs
pavel-kokolemin: Some documentation was added (by @TheQuantumPhysicist himself).

p2p/src/net/mock/transport/adapted.rs
pavel-kokolemin: Same here.

p2p/src/net/mock/transport/adapted.rs
pavel-kokolemin: OK, thank you for the summary. Let's do that in later PR.

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/wrapped_listener.rs
```diff
@@ -0,0 +1,70 @@
...
+            tokio::select! {
+                handshake = HandshakeFut::<S, T>(&mut self.handshakes) => {
+                    return Ok(handshake);
+                }
+                accept_res = self.listener.accept() => {
+                    match accept_res {
+                        Ok((base, addr)) => {
+                            // Store active handshakes because accept must be cancel safe
+                            let handshake = self.stream_adapter.handshake(base, Role::Inbound);
+                            self.handshakes.push((handshake, addr));
```
TheQuantumPhysicist, line 55: Maybe I'm missing something... but seems like this vector will be growing forever? Shouldn't this be some kind of double-ended queue, where handshakes that are done are popped out?

pavel-kokolemin: No, it won't be growing forever because we remove complete handshakes here: https://github.com/mintlayer/mintlayer-core/blob/70222be5398046ebeb80258be13a2602cfe0dc98/p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/utils.rs#L39
Double-ended queue won't help because handshakes might complete in random order.
Though it might be a good idea to bound total count of active handshakes.

TheQuantumPhysicist: Alrighty. I'll comment there for follow-up.

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/utils.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl<'a, S: StreamAdapter<T::Stream>, T: TransportSocket> Future for HandshakeFut<'a, S, T> {
+    type Output = (S::Stream, T::Address);
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
+        'outer: loop {
+            for i in 0..self.0.len() {
+                match Future::poll(self.0[i].0.as_mut(), cx) {
+                    Poll::Ready(res) => {
+                        let (_, addr) = self.0.remove(i);
```
TheQuantumPhysicist, line 39: This seems dangerous for a few reasons:
1. It seems that the Vec growth is unbounded.
2. Couple the unbounded growth with that we're looping over the whole vector in every poll, making this unbounded in execution time, opening nodes to remote DDoS.
3. Finally, to make all this worse, we remove arbitrary elements from the middle of a Vec, causing huge clones all the time...

The algorithm is OK, but the data structuring used is very bad. I can suggest a simple solution to removal, which is swapping element `i` and the last element of the Vec, but then I can imagine pinning might be an issue, besides the complexity in doing this.

We have to think of a better to do all this. Vec is wrong here.

- Worst case scenario, we should make the size of this Vec bounded.
- Best case sceanrio, we should be able to pop handshakes in O(1) time.

pavel-kokolemin: The number of concurrent handshakes is limited to 100 and `Vec::swap_remove` is used instead (which runs in O(1)).
So all points should be addresses now.

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/wrapped_listener.rs
```diff
@@ -29,6 +29,8 @@ use crate::{
     Result,
 };
 
+const MAX_CONCURRENT_HANDSHAKES: usize = 100;
```
TheQuantumPhysicist, line 32: This setting should be moved somewhere else. Please throw in a TODO above it as I don't care right now.

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/wrapped_listener.rs
```diff
@@ -43,11 +45,12 @@ impl<S: StreamAdapter<T::Stream>, T: TransportSocket> TransportListener<S::Strea
 {
     async fn accept(&mut self) -> Result<(S::Stream, T::Address)> {
         loop {
+            let accept_new = self.handshakes.len() < MAX_CONCURRENT_HANDSHAKES;
             tokio::select! {
                 handshake = HandshakeFut::<S, T>(&mut self.handshakes) => {
                     return Ok(handshake);
                 }
-                accept_res = self.listener.accept() => {
+                accept_res = self.listener.accept(), if accept_new => {
```
TheQuantumPhysicist, line 53: What happens now if a client connects and there are already 100 handshakes in progress?

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/wrapped_listener.rs
pavel-kokolemin: Sure, added.

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/wrapped_listener.rs
pavel-kokolemin: I think the TCP connection will be in pending state and once some handshake is complete we will process it normally (I have not tested this).

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/utils.rs
```diff
@@ -36,7 +36,7 @@ impl<'a, S: StreamAdapter<T::Stream>, T: TransportSocket> Future for HandshakeFu
             for i in 0..self.0.len() {
                 match Future::poll(self.0[i].0.as_mut(), cx) {
                     Poll::Ready(res) => {
-                        let (_, addr) = self.0.remove(i);
+                        let (_, addr) = self.0.swap_remove(i);
```
TheQuantumPhysicist, line 39: First, really impressive that rust provides this. I'm very happy.

Second, your usage of the implementation is now incorrect. This is the complexity I was talking about. It would've been hell to implement this whole thing ourselves, but while rust removed part of the burden, this still has corner cases that should be treated with care. The problem with your current implementation is that we should subtract `1` from `i` to re-account for the last element that replaced element number `i`. A corner case there would be when `i == 0`, which will cause an integer underflow. Funnily enough, if you blindly subtract `1` in C++, it would work, because the next `+1` in the loop will cause another overflow and fix itself. But now in rust, with `usize`, this should be treated with care because it'll get asserts to fire in debug mode. Please isolate this treatment in a function that can be tested and don't be hasty in fixing this.

pavel-kokolemin: Yes, you are absolutely right about such corner cases in general. But here when a future is complete and removed we either return result with `Poll::Ready` (if handshake succeeds) or start looping over (if handshake fails). So we don't need to bother adjusting the index.

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/wrapped_listener.rs
TheQuantumPhysicist: Please add a test for this, and also add a test to check what happens locally when the other peer times out.

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/utils.rs
TheQuantumPhysicist: Alrighty. Please add a comment above the remove saying "Index `i` will be skipped after the swap, but this is OK because the next poll will go over it again".

TheQuantumPhysicist: Feel free to resolve this comment after adding the comment in the code.

TheQuantumPhysicist (approved):
Besides the tests for the future to ensure the behavior is how we expect it, all looks good. Good job!

p2p/src/net/mock/transport/impls/stream_adapter/wrapped_transport/utils.rs
pavel-kokolemin: Outer loop was removed and we now always return.

pavel-kokolemin:
> Besides the tests for the future to ensure the behavior is how we expect it, all looks good. Good job!

Two tests added.
There was a logical error in the handshakes polling. `HandshakeFut` did not return `Poll::Ready` if some handshake failed for any reason. And this resulted in no `accept` call when there are too many active handshakes.
That was a right call to add test, thanks!

p2p/src/net/mock/transport/impls/stream_adapter/noise.rs
```diff
@@ -0,0 +1,96 @@
...
+            let state = match role {
+                Role::Outbound => snowstorm::Builder::new(NOISE_HANDSHAKE_PARAMS.clone())
+                    .local_private_key(&local_key.private)
+                    .build_initiator()
+                    .expect("snowstorm builder must succeed"),
+                Role::Inbound => snowstorm::Builder::new(NOISE_HANDSHAKE_PARAMS.clone())
+                    .local_private_key(&local_key.private)
+                    .build_responder()
+                    .expect("snowstorm builder must succeed"),
+            };
```
stanislav-tkach, line 81: It is up to you, but I think that it would be better to move the same code from the match arms:
```suggestion
            let builder = snowstorm::Builder::new(NOISE_HANDSHAKE_PARAMS.clone())
                .local_private_key(&local_key.private);
            let state = match role {
                Role::Outbound => builder.build_initiator(),
                Role::Inbound => builder.build_responder(),
            }
            .expect("snowstorm builder must succeed");
```

p2p/src/net/mock/transport/impls/stream_adapter/noise.rs
```diff
@@ -0,0 +1,96 @@
...
+
+            let stream = timeout(HANDSHAKE_TIMEOUT, NoiseStream::handshake(base, state))
+                .await
+                .map_err(|err| P2pError::NoiseHandshakeError(err.to_string()))?
```
stanislav-tkach, line 85: I'm not quite sure, but perhaps it makes sense to introduce the specific `NoiseHandshakeTimeout` error for that.

Even if it isn't worth the effort I propose to use a custom error string here, because otherwise it would be `"NoiseHandshakeError(Elapsed)"` which isn't very descriptive.

stanislav-tkach (approved):

p2p/src/net/mock/transport/impls/stream_adapter/noise.rs
pavel-kokolemin: Good idea, thanks.

p2p/src/net/mock/transport/impls/stream_adapter/noise.rs
pavel-kokolemin: I don't think we need specific error here. The text is updated to "Handshake timeout".

iljakuklic (approved):
Looks good overall. Added some comments, most of them minor. Please pay attention to the one mentioning `decode_all`.

p2p/src/net/mock/transport/traits/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod listener;
+pub mod socket;
+pub mod stream;
```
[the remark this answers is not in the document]
pavel-kokolemin: Sure, updated.

p2p/src/net/mock/transport/traits/listener.rs
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use async_trait::async_trait;
+
+use crate::Result;
+
+/// An abstraction layer over a potential inbound network connection (acceptor in boost terminology).
+#[async_trait]
+pub trait TransportListener<Stream, Address>: Send {
```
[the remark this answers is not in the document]
pavel-kokolemin: I think you are right. But this code was already here so let's please do something about it later. I added TODO.

p2p/src/net/mock/transport/impls/stream_adapter/identity.rs
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+use futures::future::{ready, BoxFuture};
+
+use crate::net::mock::{peer::Role, transport::PeerStream};
+
+use super::StreamAdapter;
+
+#[derive(Debug)]
+pub struct IdentityStreamAdapter;
```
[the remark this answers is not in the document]
pavel-kokolemin: I will take a note, thanks.
But `new` here is only because of the trait requirement (and this part will probably change soon).
Let's please keep it for now.

p2p/src/net/mock/transport/impls/stream_adapter/noise.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use super::StreamAdapter;
+
+// How much time is allowed to spend setting up (optionally) encrypted stream.
+const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
+
+static NOISE_HANDSHAKE_PATTERN: &str = "Noise_XX_25519_ChaChaPoly_SHA256";
+
+static NOISE_HANDSHAKE_PARAMS: once_cell::sync::Lazy<snowstorm::NoiseParams> =
+    once_cell::sync::Lazy::new(|| NOISE_HANDSHAKE_PATTERN.parse().expect("valid pattern"));
```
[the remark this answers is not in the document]
pavel-kokolemin: Noise protocol names are well known and documented in the spec: https://noiseprotocol.org/noise.html#protocol-names-and-modifiers
So I think it's OK to parse it. lib-p2p also parses it from the string.

p2p/src/net/mock/transport/message_codec.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        if src.len() < 4 + length {
+            src.reserve(4 + length - src.len());
+            return Ok(None);
+        }
+
+        let data = src[4..4 + length].to_vec();
+        src.advance(4 + length);
+
+        match Message::decode(&mut &data[..]) {
```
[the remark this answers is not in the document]
pavel-kokolemin: Nice catch, thanks. Fixed.

p2p/src/net/mock/transport/message_codec.rs
```diff
@@ -0,0 +1,129 @@
...
+            )
+            .into());
+        }
+
+        if src.len() < 4 + length {
+            src.reserve(4 + length - src.len());
+            return Ok(None);
+        }
+
+        let data = src[4..4 + length].to_vec();
```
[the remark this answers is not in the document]
pavel-kokolemin: Fixed, thanks. This code was already here so I have not looked carefully.

p2p/src/net/mock/transport/message_codec.rs
```diff
@@ -0,0 +1,129 @@
...
+    type Item = Message;
+    type Error = P2pError;
+
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>> {
+        if src.len() < 4 {
+            return Ok(None);
+        }
+
+        let mut length_bytes = [0u8; 4];
+        length_bytes.copy_from_slice(&src[..4]);
```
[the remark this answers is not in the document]
pavel-kokolemin: Fixed, thanks.

p2p/src/net/mock/transport/traits/listener.rs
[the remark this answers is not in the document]
iljakuklic: Ok, cheers

p2p/src/net/mock/transport/impls/stream_adapter/noise.rs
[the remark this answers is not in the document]
iljakuklic: My motivation for pointing this out was that you could avoid `once_cell` and such and make it just a `const`. Your call in the end, I don't think it's a big deal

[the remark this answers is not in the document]
pavel-kokolemin: Right, static initialization would be better here.
But `NoiseParams` has some annoying information duplication. It requires both text noise protocol name and protocol details. So we will need to make sure that the protocol name and actual values are in sync:

```
pub struct NoiseParams {
    pub name:      String,
    pub base:      BaseChoice,
    pub handshake: HandshakeChoice,
    pub dh:        DHChoice,
    #[cfg(feature = "hfs")]
    pub kem:       Option<KemChoice>,
    pub cipher:    CipherChoice,
    pub hash:      HashChoice,
}
```

Also we won't be able const struct because of `String` .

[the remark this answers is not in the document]
iljakuklic: Ok, that's quite annoying. Let's leave it as is with the string parsing then.

PR #573, opened 2022-11-22
Remove retries from the sync manager and implement timeouts for the mock backend

stanislav-tkach:
The retries logic was removed because it doesn't have much sense. TCP guaranties delivery, so there is no reason to send a peer a new request if the previous one was ignored. `SyncingEvent::Error` is removed because it was only used for timeouts. I also have removed some of the `BlockSyncManager` tests and while I feel bad removing tests, they actually check the libp2p timeout related logic and not the sync manager behavior.


pavel-kokolemin (approved):

p2p/src/sync/tests/request_response.rs
```diff
@@ -163,166 +170,17 @@ where
...
+    multiple_requests_and_responses::<MakeTcpAddress, MockService<TcpMockTransport>>().await;
 }
 
 #[tokio::test]
-async fn test_multiple_requests_and_responses_mock_channels() {
-    test_multiple_requests_and_responses::<MakeChannelAddress, MockService<ChannelMockTransport>>()
+async fn multiple_requests_and_responses_mock_channels() {
+    multiple_requests_and_responses::<MakeChannelAddress, MockService<ChannelMockTransport>>()
         .await;
 }
```
TheQuantumPhysicist, line 186: Can you add tests that show that this still works as expected?

TheQuantumPhysicist (approved):
Please add tests that show that ignoring requests isn't a disaster.

Also, are we sure that messages without response won't cause state issues? Like if 3 messages have to go in a row, A then B then C, then ignoring B won't put us in an invalid state?

p2p/src/net/types/mod.rs
```diff
@@ -162,8 +155,7 @@ pub enum SyncingEvent<T: NetworkingService> {
         /// Received request
         request: message::Request,
     },
-
-    /// Incoming response to a sent request
+    /// An incoming response to a sent request.
```
muursh, line 158: Incredibly picky but the second half of the comment is redundant. It wouldn't be a response if it wasn't as a result of an outgoing request.

muursh (approved):
No qualms from my side

iljakuklic (approved):

stanislav-tkach:
I have made the following changes:
- The block sync manager now keeps track of requests timeouts and unresponsive peers are disconnected. We previously had this functionality for the `libp2p` backend only.
- The timeout value has been moved to the config (`P2pConfig::sync_manager_response_timeout`).
- More tests.

iljakuklic (approved):

stanislav-tkach:
The third and hopefully final iteration. 🙃 The changes from the previous version:
- The timeouts have been implemented for the mock backend.
- Some of the tests have been restored.

p2p/src/net/mock/backend.rs
```diff
@@ -208,21 +238,30 @@ where
...
 
+        let is_inserted = self.pending_responses.insert(request_id);
+        debug_assert!(is_inserted);
+
+        let timeout = self.p2p_config.request_timeout.clone().into();
+        let sender = self.timeouts_sender.clone();
+        tokio::spawn(async move {
+            time::sleep(Duration::from_secs(timeout)).await;
+            let _ = sender.send((request_id, peer_id));
+        });
```
iljakuklic, line 263: Would it be worth it to somehow kill the timeout task once the response is received so it does not hang around for the full timeout period?

p2p/src/net/mock/backend.rs
```diff
@@ -407,6 +452,12 @@ where
                 command = self.cmd_rx.recv().fuse() => {
                     self.handle_command(command.ok_or(P2pError::ChannelClosed)?).await?;
                 }
+                ids = self.timeouts_receiver.recv() => {
+                    match ids {
+                        Some((request_id, peer_id)) => self.handle_request_timeout(peer_id, request_id).await?,
+                        None => return Err(P2pError::Other("Timeouts channels closed")),
```
iljakuklic, line 458: Since `timeouts_sender` is a part of `self`, I think this could be a panic.

iljakuklic (approved):
Looks good. Just a few minor nitpicks and comments

p2p/src/sync/tests/request_response.rs
```diff
@@ -201,128 +203,42 @@ where
...
-                mgr1.process_error(peer_id, request_id, error).await.unwrap();
+                assert_eq!(peer_id, peer2_id);
+                mgr1.process_timeout(peer_id, request_id).await.unwrap();
             }
             _ => panic!("invalid event received"),
         }
     });
 
-    match timeout(Duration::from_secs(15), mgr2.peer_sync_handle.poll_next()).await {
+    match timeout(Duration::from_secs(5), mgr2.peer_sync_handle.poll_next()).await {
```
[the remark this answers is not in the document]
stanislav-tkach: I haven't, thanks for pointing this out function out! Though the timeout is currently set in the config and I believe that the value of 1 second used for the tests is enough to deal with this issue.

p2p/src/net/mock/backend.rs
stanislav-tkach: You are right, but I'm very reluctant to use panics in the code. 🙃 I can change it if you think it is OK.

p2p/src/net/mock/backend.rs
stanislav-tkach: I haven’t been able to come up with a solution how to do that properly. I'm open to suggestions. 🙃

iljakuklic: Hmm not sure. Maybe add the task handle to `pending_responses` and call `Task::abort` when the request is resolved. Would that work?

stanislav-tkach: Good idea, I will try that!

p2p/src/config.rs
```diff
@@ -25,6 +25,12 @@ make_config_setting!(OutboundConnectionTimeout, u64, 10);
 make_config_setting!(MdnsConfigSetting, MdnsConfig, MdnsConfig::Disabled);
 make_config_setting!(MdnsQueryInterval, u64, MDNS_DEFAULT_QUERY_INTERVAL);
 make_config_setting!(MdnsEnableIpV6Discovery, bool, MDNS_DEFAULT_IPV6_STATE);
+make_config_setting!(
+    /// The request timeout value in seconds.
+    RequestTimeout,
+    u64,
```
TheQuantumPhysicist, line 31: How about we make this a Duration?

p2p/src/net/mock/backend.rs
```diff
@@ -208,21 +238,30 @@ where
...
-
+        let (request_id, request) = self.request_mgr.make_request(request)?;
         peer.tx.send(MockEvent::SendMessage(request)).await.map_err(P2pError::from)?;
 
+        let is_inserted = self.pending_responses.insert(request_id);
+        debug_assert!(is_inserted);
+
+        let timeout = self.p2p_config.request_timeout.clone().into();
+        let sender = self.timeouts_sender.clone();
+        tokio::spawn(async move {
```
TheQuantumPhysicist, line 260: A new task for every request sent? This seems very excessive and can easily blow up in memory in the future as a result to bad maintenance...

Why not just share an atomic of the latest time a request was sent? The way I would do this, is have an async timer in a task that's launched when the peer is connected. In that task, we own an atomic variable that contains the time of the latest request. Once a response is received, that atomic variable is reset (to zero, unfortunately using something nice like an Option would remove atomicity). 

The timer ticks every time that timeout duration passes (or twice of that, that's an OK error to accept). Once a tick happens, we check current time versus stored atomic time. If the difference is larger than timeout, we send the timeout signal.

p2p/src/sync/tests/mod.rs
```diff
@@ -61,11 +63,17 @@ where
     );
     tokio::spawn(async move { man.main().await });
 
-    let config = Arc::new(common::chain::config::create_unit_test_config());
-    let (conn, sync) = T::start(addr, Arc::clone(&config), Default::default()).await.unwrap();
+    let chain_config = Arc::new(common::chain::config::create_unit_test_config());
+    let p2p_config = Arc::new(P2pConfig {
+        request_timeout: 1.into(),
+        ..Default::default()
```
TheQuantumPhysicist, line 69: Please don't use the pattern `..Default..default()`... it's even preferable that you get rid of any instances of that you see around.

TheQuantumPhysicist (approved):
Looking good. Please check the few comments I made. Especially the timer and task parts. Please fix them and let me know.

p2p/src/sync/tests/mod.rs
stanislav-tkach: I understand your concerns and I don't mind removing `..Default..default()` from all p2p code (all of them are in the tests code by the way), but what if we decide to change the default values in the config? In that case all tests will use some hardcoded values that are no longer default. Is that OK?

TheQuantumPhysicist: Theoretically, the values should be chosen per object and per test. So, if the test is meant to use the default configuration, then that's OK, but if not, the full settings that get the test to function should be selected. There's no way to win all possible scenarios. I'd rather manually choose new values when adding a new member that an unrealistic blanket untraceable changes on tests just because the default was changed.

p2p/src/net/mock/backend.rs
```diff
@@ -208,21 +238,30 @@ where
...
+            time::sleep(Duration::from_secs(timeout)).await;
```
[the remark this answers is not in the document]
stanislav-tkach: Just to be clear: this is `tokio::time::sleep`, not `std::thread::sleep`. According to the documentation, "no work is performed while awaiting on the sleep future to complete".

p2p/src/sync/tests/mod.rs
stanislav-tkach: Makes sense. I have removed all `..Default::default` usages from the p2p code.

p2p/src/net/mock/backend.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I know. But it's the wrong tool for the job. 

Btw, also no work is performed on sleeps because a yield happens. But it's still the wrong way. Timers can be optimized by OS schedulers in ways that are blocked by sleeping.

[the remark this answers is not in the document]
stanislav-tkach: > Timers can be optimized by OS schedulers in ways that are blocked by sleeping.

Tokio uses a timer under the hood of the `Sleep` feature.

p2p/src/net/mock/backend.rs
stanislav-tkach: While I agree that spawning a new task each time is little too much, the tasks are fairly lightweight. In the documentation is stated that a task itself requires only a single allocation and 64 bytes of memory. I understand that there is an additional overhead, but spawning thousands of tasks is OK for the tokio runtime.

Anyway, I'm going to implement your approach.

TheQuantumPhysicist: If you're looking for a good argument why this is a bad idea, just imagine someone trying to debug while all these tasks are running ;-)

In multithreaded applications, in general, we try to minimize concurrency/tasks to make reasoning about the program easier.

Issue #514, opened 2022-10-26
Announcements subscription during handshake


stanislav-tkach:
A peer should only be able to subscribe to announcements during the handshake process, so the list of announcement topics cannot be changed after the connection is established.

PR #575, opened 2022-11-24
Obtain a peer subscription during the handshake process

stanislav-tkach:
- Subscriptions are now exchanged as part of peer information.
- The `subscribe` method has been removed from the `SyncingMessagingService` trait.
- The subscriptions are stored in the config now.

Closes https://github.com/mintlayer/mintlayer-core/issues/514.


node/src/config_files/p2p.rs
```diff
@@ -112,6 +113,8 @@ impl From<P2pConfigFile> for P2pConfig {
             outbound_connection_timeout: c.outbound_connection_timeout.into(),
             mdns_config: mdns_config.into(),
             request_timeout: c.request_timeout.map(Duration::from_secs).into(),
+            // TODO: This should probably be set indirectly, for example by the node type.
+            announcement_subscriptions: AnnouncementSubscriptions::default(),
```
TheQuantumPhysicist, line 117: I don't think having announcement subscriptions are the correct thing to do. This is too low-level. The right thing to do is to define multiple sets of possible subscriptions with node types. Currently, we can have one node type, the full node, which subscribes to whatever is available now. There's another type that we can support, which is the "BlocksOnly" mode for the sake of variety to be able to test it, at least. Please see how bitcoin does this. The entry point is here:

https://github.com/bitcoin/bitcoin/blob/9e229a542ff2107be43eff2e4b992841367f0366/src/init.cpp#L419

p2p/src/sync/mod.rs
```diff
@@ -355,30 +355,29 @@ where
     ///
     /// The node is considered fully synced (its initial block download is done) if all its peers
```
TheQuantumPhysicist, line 356: As we've discussed this is totally wrong and has to be changed in some subsequent PR.

p2p/src/net/mock/peer.rs
```diff
@@ -90,25 +96,23 @@ where
...
         match self.role {
             Role::Inbound => {
-                let (peer_id, network, version, protocols) =
-                    if let Ok(Some(types::Message::Handshake(types::HandshakeMessage::Hello {
+                let Ok(Some(types::Message::Handshake(types::HandshakeMessage::Hello {
                         peer_id,
                         version,
                         network,
                         protocols,
+                        subscriptions,
```
TheQuantumPhysicist, line 104: Please make sure there are protections against having huge values in these Vec and BTreeSet

TheQuantumPhysicist (approved):
Looking good. Please fix the little things I mentioned then feel free to merge. 

To fix the config issue, please create a NodeType enum, and put in it FullNode and BlocksOnlyNode, and then map these to announcement values.

p2p/src/net/mock/peer.rs
stanislav-tkach: I have rechecked and the subscriptions are only stored in `BTreeSet`, therefore it is impossible to send duplicated topics. Personally I would be more concerned with the `PeerInfo::agent` field because an arbitrary long string can be stored there.

Additionally, we already have the `MAX_MESSAGE_SIZE` constant, though it is equal to 10 megabytes. Do we need additional limitation specifically for subscriptions or peer info?

TheQuantumPhysicist: Theoretically speaking, it should be enough. Though every layer should have its own protection. If you wish, you could make this an issue with testnet milestone. No hurry on this.

stanislav-tkach: https://github.com/mintlayer/mintlayer-core/issues/594

iljakuklic (approved):

PR #576, opened 2022-11-29
subsystem: Command submission + unbounded queues

iljakuklic:
Please see the commit messages for description and rationale

### DIFF
--- a/subsystem/src/subsystem.rs
+++ b/subsystem/src/subsystem.rs
@@ -96,7 +96,7 @@ pub type ActionSender<T> = mpsc::UnboundedSender<Action<T, ()>>;
 /// There are two sets of methods for communication with and control of subsystem:
 /// * Methods starting with `submit_` will submit the closure given as the argument for processing
 ///   by the subsystem. The result is not immediately ready and the current task is free to
-///   continue its operation and `.await` the return value at a latet point (or decide not to).
+///   continue its operation and `.await` the return value at a later point (or decide not to).
 /// * Methods starting with `call_` will also submit the closure and suspend the current task until
 ///   the result is ready, returning it directly.
 pub struct Handle<T> {
@@ -153,7 +153,7 @@ impl<T: Send + 'static> Handle<T> {
         self.submit_async_mut(|this| func(this))
     }
 
-    /// Subimt a procedure to the subsystem.
+    /// Submit a procedure to the subsystem.
     pub fn submit_mut<R: Send + 'static>(
         &self,
         func: impl for<'a> FnOnce(&'a mut T) -> R + Send + 'static,
--- a/subsystem/src/subsystem.rs
+++ b/subsystem/src/subsystem.rs
@@ -64,7 +64,7 @@ impl<T> std::future::Future for CallResponse<T> {
     type Output = Result<T, CallError>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
-        Pin::new(&mut self.0).poll(cx).map(|r| r.map_err(|_| CallError::SubsystemDead))
+        Pin::new(&mut self.0).poll(cx).map_err(|_| CallError::ResultFetchFailed)
     }
 }
 
@@ -114,8 +114,30 @@ impl<T> Clone for Handle<T> {
 
 #[derive(Debug, Ord, PartialOrd, PartialEq, Eq, Clone, Copy, thiserror::Error)]
 pub enum CallError {
-    #[error("Callee subsystem has terminated")]
-    SubsystemDead,
+    #[error("Call submission failed")]
+    SubmissionFailed,
+    #[error("Result retrieval failed")]
+    ResultFetchFailed,
+}
+
+pub struct CallResult<T>(Result<CallResponse<T>, CallError>);
+
+impl<T> CallResult<T> {
+    /// Get the corresponding [`CallResponse`], with the opportunity to handle errors at send time.
+    pub fn response(self) -> Result<CallResponse<T>, CallError> {
+        self.0
+    }
+}
+
+impl<T> std::future::Future for CallResult<T> {
+    type Output = Result<T, CallError>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
+        self.0.as_mut().map_or_else(
+            |err| task::Poll::Ready(Err(*err)),
+            |res| Pin::new(res).poll(cx),
+        )
+    }
 }
 
 impl<T: Send + 'static> Handle<T> {
@@ -124,15 +146,15 @@ impl<T: Send + 'static> Handle<T> {
         Self { action_tx }
     }
 
-    /// Submit an async procedure to the subsystem. Result has to be await-ed explicitly
-    pub fn submit_async_mut<R: Send + 'static>(
+    /// Call an async procedure to the subsystem. Result has to be await-ed explicitly
+    pub fn call_async_mut<R: Send + 'static>(
         &self,
         func: impl for<'a> FnOnce(&'a mut T) -> BoxFuture<'a, R> + Send + 'static,
-    ) -> Result<CallResponse<R>, CallError> {
+    ) -> CallResult<R> {
         let (rtx, rrx) = oneshot::channel::<R>();
-        let rrx = CallResponse(rrx);
 
-        self.action_tx
+        let res = self
+            .action_tx
             .send(Box::new(move |subsys| {
                 Box::pin(async move {
                     if rtx.send(func(subsys).await).is_err() {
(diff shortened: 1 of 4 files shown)


stanislav-tkach (approved):
Seems reasonable.

subsystem/src/subsystem.rs
```diff
@@ -27,37 +30,29 @@ pub trait Subsystem: 'static + Send + Sized {
...
     }
 }
 
 impl Default for SubsystemConfig {
     fn default() -> Self {
-        Self {
-            subsystem_name: Self::DEFAULT_SUBSYSTEM_NAME,
-            call_queue_capacity: Self::DEFAULT_CALL_QUEUE_CAPACITY,
-        }
+        let subsystem_name = Self::DEFAULT_SUBSYSTEM_NAME;
```
stanislav-tkach, line 46: It isn't related to your pull request, but I'm curious: why do we even have the `DEFAULT_SUBSYSTEM_NAME` constant? As far as I can see, we always use explicitly named subsystems even in the tests.

iljakuklic: Good point. Will have a look and possibly remove it if it turns out it's indeed not used.

iljakuklic: Removed

subsystem/src/subsystem.rs
```diff
@@ -89,13 +86,22 @@ impl ShutdownRequest {
...
+
 /// Subsystem handle.
 ///
 /// This allows the user to interact with the subsystem from the outside. Currently, it only
 /// supports calling functions on the subsystem.
+///
+/// There are two sets of methods for communication with and control of subsystem:
+/// * Methods starting with `submit_` will submit the closure given as the argument for processing
+///   by the subsystem. The result is not immediately ready and the current task is free to
+///   continue its operation and `.await` the return value at a latet point (or decide not to).
```
TheQuantumPhysicist, line 99: "later" typo

subsystem/src/subsystem.rs
```diff
@@ -114,52 +120,85 @@ pub enum CallError {
...
+
+    /// Submit an async procedure to the subsystem (immutable).
+    pub fn submit_async<R: Send + 'static>(
+        &self,
+        func: impl for<'a> FnOnce(&'a T) -> BoxFuture<'a, R> + Send + 'static,
+    ) -> Result<CallResponse<R>, CallError> {
+        self.submit_async_mut(|this| func(this))
+    }
+
+    /// Subimt a procedure to the subsystem.
```
TheQuantumPhysicist, line 156: Submit

TheQuantumPhysicist (commented):
From looking around, if I understand correctly, the difference between submit and call is that call will return a future, `await`ing which will cause the evaluation to happen in the subsystem. Submit, on the other hand, will get the subsystem to start the evaluation of the function on call, and `await`ing the returned future will specifically await the result on the receiving channel (instead of awaiting the function call itself in the subsystem).

If what I understand is correct, then I have to propose two things:
1. Tests that prove this behavior
2. Documentation that describe the behavior on submit/call functions in the same way I did.

Having said all this, I have to say that submit is the behavior I would expect and I don't see a need to make this distinction. I feel like we have this distinction now because of an implementation misunderstanding we had earlier. In my earliest static dispatch example in C++ (and your translation of it to rust), awaiting the availability of a result (instead of the availablity of the function call in the subsystem) is the intended behavior. In fact, I can even content that `call` is completely useless and no one wants that. Why would the caller care when the function is submitted to the executor? Usually you want the subsystem (or the manager) to decide when the function call happens, not the caller.

iljakuklic:
> Having said all this, I have to say that submit is the behavior I would expect and I don't see a need to make this distinction. I feel like we have this distinction now because of an implementation misunderstanding we had earlier. In my earliest static dispatch example in C++ (and your translation of it to rust), awaiting the availability of a result (instead of the availablity of the function call in the subsystem) is the intended behavior. In fact, I can even content that `call` is completely useless and no one wants that. Why would the caller care when the function is submitted to the executor? Usually you want the subsystem (or the manager) to decide when the function call happens, not the caller.

As you can see, `call` trivially delegates to `submit` using `await`, so having only one seems sensible to me. There is one subtle difference in that submitting a call as well as receiving can fail. That means that if we change `call` to behave like `submit`, all its uses will have to be slightly updated `subsys.call(todo!()).await?` to `subsys.call(todo!())?.await?`

iljakuklic:
> > Having said all this, I have to say that submit is the behavior I would expect and I don't see a need to make this distinction. I feel like we have this distinction now because of an implementation misunderstanding we had earlier. In my earliest static dispatch example in C++ (and your translation of it to rust), awaiting the availability of a result (instead of the availablity of the function call in the subsystem) is the intended behavior. In fact, I can even content that `call` is completely useless and no one wants that. Why would the caller care when the function is submitted to the executor? Usually you want the subsystem (or the manager) to decide when the function call happens, not the caller.
> 
> As you can see, `call` trivially delegates to `submit` using `await`, so having only one seems sensible to me. There is one subtle difference in that submitting a call as well as receiving can fail. That means that if we change `call` to behave like `submit`, all its uses will have to be slightly updated `subsys.call(todo!()).await?` to `subsys.call(todo!())?.await?`

This is now included in the PR. There are many more changes but almost all are concerned with the increased amount of error handling required. That is because now both method submission and result fetching can fail while before the two were rolled into one.

TheQuantumPhysicist:
So, as per our discussion, it seems that handling two different kinds of errors is an overkill, especially that the meaning of an error doesn't matter (in both cases the program should be shutdown afterwards, so this error is just a marker for stopping gracefully). The extra error handling makes everything overly complex. The other side of the story is that we may want to support "fire-and-forget", where we don't care about the result, but want to know that the call went through, but I don't see why we would want to support that. Subsystems should not be dependent on each other on shutdown. 

My opinion is that we should fuse both errors into one sent through the oneshot channel, since we don't care. However, Lukas still wants to give it a shot to separate the errors (separate between submission error vs channel error). As long as we don't introduce extra bloat on every call where the end of the line has 13 `map_err`s and 55 `????????` :smile:, I'm OK with any similar solution.

iljakuklic:
I pushed an update to make the error handling somewhat less annoying.

Like before, the result can be `await`ed and we get an error if either submission or result retrieval fails. That means no changes to existing code are required, the following works as before (but is implemented slightly differently using a custom type that implements `Future`):

```rust
let result = subsys.call(foo).await.expect("Call failed");
```

The caller now also has an option to check if the submission failed separately from the result retrieval handling. To do that, the `response` method is used instead of `await`ing right away:

```rust
let response = subsys.call(foo).response().expect("Call submission failed");
// do stuff here, then await the result:
let result = response.await.expect("Result retrieval failed");
```

TheQuantumPhysicist (approved):
Very nice! You have my blessing :smile:

subsystem/src/subsystem.rs
```diff
@@ -108,58 +116,82 @@ impl<T> Clone for Handle<T> {
...
         &self,
         func: impl for<'a> FnOnce(&'a mut T) -> BoxFuture<'a, R> + Send + 'static,
-    ) -> Result<R, CallError> {
+    ) -> CallResult<R> {
         let (rtx, rrx) = oneshot::channel::<R>();
 
-        self.action_tx
+        let res = self
+            .action_tx
             .send(Box::new(move |subsys| {
```
TheQuantumPhysicist, line 160: Please see if you can get rid of this Box, since everything is a generic. Doesn't have to be in this PR.

subsystem/src/subsystem.rs
```diff
@@ -108,58 +116,82 @@ impl<T> Clone for Handle<T> {
...
+    SubmissionFailed,
+    #[error("Result retrieval failed")]
+    ResultFetchFailed,
+}
+
+pub struct CallResult<T>(Result<CallResponse<T>, CallError>);
+
+impl<T> CallResult<T> {
+    /// Get the corresponding [`CallResponse`], with the opportunity to handle errors at send time.
+    pub fn response(self) -> Result<CallResponse<T>, CallError> {
```
TheQuantumPhysicist, line 129: Maybe adding a test with this usage as example will help. The only example I see is in github.

subsystem/src/subsystem.rs
```diff
@@ -71,6 +57,19 @@ impl<T: 'static> CallRequest<T> {
...
+/// Call response that can be polled for result
+pub struct CallResponse<T>(oneshot::Receiver<T>);
+
+impl<T> std::future::Future for CallResponse<T> {
+    type Output = Result<T, CallError>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
+        Pin::new(&mut self.0)
+            .poll(cx)
+            .map(|r| r.map_err(|_| CallError::ResultFetchFailed))
```
stanislav-tkach, line 69: Nitpick: `Poll` provides the `map_err` function, so it can be used directly:
```suggestion
        Pin::new(&mut self.0).poll(cx).map_err(|_| CallError::ResultFetchFailed)
```

stanislav-tkach (approved):

subsystem/src/subsystem.rs
iljakuklic: Very nice, cheers!

subsystem/src/subsystem.rs
iljakuklic: I don't think it's easily achievable. The channel endpoint has type `Sender<Box<dyn Fn(_)>>`. It has to be a box since the `Fn` closure may a different type for each invocation of the `call` procedure.

subsystem/src/subsystem.rs
iljakuklic: Added

iljakuklic:
I'd like to merge this soon. Any more comments / objections?

TheQuantumPhysicist:
> I'd like to merge this soon. Any more comments / objections?

Looking good. I just would like to request adding explanation to the test, as in, explaining that the test shows how to separate call errors from result retrieval errors. Otherwise feel free to merge. Cheers!

muursh (approved):
Looks fine to me

PR #577, opened 2022-11-30
Rework undo for DeltaDataCollection

azarovh:
This PR aims to address the issue with the accounting that I encountered while working on the reorg test for stake pool operation.

The main problem is that undo for `DeltaDataCollection` was implemented with the assumption that while undoing the merge of 2 deltas the collection must not be empty. This results in the behaviour that if we flush deltas to the storage we could not undo these operations via applying undo on the newly created deltas.

To be more specifically I removed `DeltaMapOp` which could erase data from the collection on undo. And replaced this mechanism with creation of reverse deltas for undo, which could be propagated over different collections. 

I recommend to start by reading 2 high-level tests that better illustrate the problem: `create_pool_decomission_pool_merge_undo_merge`  and `create_pool_flush_undo` in pos_accounting.

### DIFF
--- a/accounting/src/delta/delta_data_collection/mod.rs
+++ b/accounting/src/delta/delta_data_collection/mod.rs
@@ -25,11 +25,11 @@ use self::undo::*;
 
 #[derive(PartialEq, Eq, Clone, Encode, Decode, Debug)]
 pub enum DataDelta<T: Clone> {
-    // stores new value
+    // Stores new value
     Create(Box<T>),
-    // stores prev and new values
+    // Stores prev and new values
     Modify((Box<T>, Box<T>)),
-    // stores prev value before deletion
+    // Stores prev value before deletion
     Delete(Box<T>),
 }
 
@@ -100,8 +100,8 @@ impl<K: Ord + Copy, T: Clone> DeltaDataCollection<K, T> {
         key: K,
         other: DeltaMapElement<T>,
     ) -> Result<Option<DataDeltaUndo<T>>, Error> {
-        let current_el = self.data.get(&key);
-        let current_delta = match current_el {
+        let current_element = self.data.get(&key);
+        let current_delta = match current_element {
             Some(d) => match d {
                 DeltaMapElement::Delta(d) => Some(d),
                 DeltaMapElement::Undo(_) => None,
@@ -117,7 +117,7 @@ impl<K: Ord + Copy, T: Clone> DeltaDataCollection<K, T> {
             DeltaMapElement::Undo(_) => None,
         };
 
-        let new_element = match current_el {
+        let new_element = match current_element {
             Some(current) => combine_delta_elements(current, other)?,
             None => other,
         };
@@ -165,7 +165,7 @@ impl<K: Ord + Copy, T: Clone> FromIterator<(K, DataDelta<T>)> for DeltaDataColle
     }
 }
 
-/// This function returns a delta that if applied to the result of merge(lhs,rhs) gives original lhs
+/// This function returns a delta that if applied to the result of merge(delta1,delta2) gives original delta1
 fn create_undo_delta<T: Clone>(
     lhs: Option<&DataDelta<T>>,
     rhs: DataDelta<T>,
--- a/accounting/src/delta/delta_data_collection/mod.rs
+++ b/accounting/src/delta/delta_data_collection/mod.rs
@@ -52,17 +52,14 @@ impl<K: Ord + Copy, T: Clone> DeltaDataCollection<K, T> {
         }
     }
 
+    pub fn data(&self) -> &BTreeMap<K, DeltaMapElement<T>> {
+        &self.data
+    }
+
     pub fn get_data_delta(&self, key: &K) -> Option<&DataDelta<T>> {
         self.data.get(key).map(|el| match el {
             DeltaMapElement::Delta(d) => d,
             DeltaMapElement::Undo(d) => d,
-        })
-    }
-
-    pub fn delta_iter(&self) -> impl Iterator<Item = (&K, &DataDelta<T>)> {
-        self.data.iter().map(|(k, el)| match el {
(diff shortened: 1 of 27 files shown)


pos_accounting/src/storage/in_memory.rs
```diff
@@ -59,6 +59,16 @@ impl InMemoryPoSAccounting {
             delegation_data,
         }
     }
+
+    pub fn dump(&self) -> PoSAccountingData {
```
TheQuantumPhysicist, line 63: Why not just use the clone method from the Clone trait?

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -110,170 +140,111 @@ impl<K: Ord + Copy, T> DeltaDataCollection<K, T> {
...
     rhs: DataDelta<T>,
-) -> Result<DeltaMapOp<DataDelta<T>>, Error> {
+) -> Result<DataDelta<T>, Error> {
     match (lhs, rhs) {
-        (DataDelta::Create(_), DataDelta::Create(_)) => Err(Error::DeltaDataCreatedMultipleTimes),
-        (DataDelta::Create(_), DataDelta::Modify(d)) => Ok(DeltaMapOp::Write(DataDelta::Create(d))),
-        (DataDelta::Create(_), DataDelta::Delete) => {
-            // if lhs had a creation, and we delete, this means nothing is left and there's a net zero to return
-            Ok(DeltaMapOp::Delete)
+        (DataDelta::Create(_), DataDelta::Create(d)) => Ok(DataDelta::Create(d)),
```
TheQuantumPhysicist, line 213: Why are we now accepting two consecutive Create ops? This should be an error.

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -110,170 +140,111 @@ impl<K: Ord + Copy, T> DeltaDataCollection<K, T> {
...
+        (DataDelta::Create(_), DataDelta::Modify((_, d))) => Ok(DataDelta::Create(d)),
+        (DataDelta::Create(_), DataDelta::Delete(d)) => Ok(DataDelta::Delete(d)),
+
+        (DataDelta::Modify((d1, _)), DataDelta::Create(d2)) => {
```
TheQuantumPhysicist, line 217: This should be error too

accounting/src/delta/delta_data_collection/mod.rs
TheQuantumPhysicist: To be clear, it was an error... why was this function changed this way?

azarovh: This is valid case if we undo, because undo for Delete is Create.
Take a look at `create_delete_undo_merge` test.

azarovh: Though I agree that it's confusing. Maybe we need some kind of `UndoDelta` that would identify undo operation and not user operation

accounting/src/delta/delta_data_collection/mod.rs
azarovh: Same here, to be able to undo properly. See `create_modify_delete_undo_merge` test

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -23,43 +23,64 @@ use crate::error::Error;
...
 use self::undo::*;
 
-/// The outcome of combining two deltas for a given key upon the map that contains it
-#[derive(PartialEq, Eq, Debug)]
-enum DeltaMapOp<T> {
-    /// Write a specific value (for example, to write a Create or Modify operation)
-    Write(T),
-    /// Erase the value at the relevant key spot (for example, a modify followed by Erase yields nothing)
-    Delete,
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug)]
```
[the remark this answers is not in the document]
azarovh: Seems irrelevant to me, because it's about `DeltaMapOp` which was removed

accounting/src/delta/delta_data_collection/mod.rs
TheQuantumPhysicist: We definitely need that... this is absolutely unreadable and even incorrect. In its current form, if a user creates twice, it'll succeed at the lowest level, and high levels should detect the error. This isn't acceptable.

pos_accounting/src/pool/tests/operations_tests.rs
```diff
@@ -0,0 +1,111 @@
...
+            BTreeMap::from([(pool_id, pledge_amount)]),
+            BTreeMap::new(),
+            BTreeMap::new(),
+            BTreeMap::new(),
+        );
+        assert_eq!(storage, expected_storage);
+    }
+}
+
+// TODO: more tests with operations here
```
TheQuantumPhysicist, line 111: Please clarify what else has to be added/what's missing.

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -23,43 +23,58 @@ use crate::error::Error;
...
 }
 
 #[derive(PartialEq, Eq, Clone, Encode, Decode, Debug)]
-pub enum DataDelta<T> {
-    Create(Box<T>),
-    Modify(Box<T>),
-    Delete,
+pub enum DeltaMapElement<T: Clone> {
+    Delta(DataDelta<T>),
+    Undo(DataDelta<T>),
```
TheQuantumPhysicist, line 39: I would call this DeltaUndo, not just Undo.

pos_accounting/src/pool/tests/operations_tests.rs
azarovh: not sure that it's worth it - these tests are exactly what I'm working on right now

azarovh:
Closing in favour of #586

PR #578, opened 2022-12-01
Add secp256k1 keys and BIP340 Schnorr signatures

erasmospunk:


crypto/src/key/mod.rs
```diff
@@ -133,19 +153,34 @@ impl PublicKey {
     }
 
     pub fn verify_message(&self, signature: &Signature, msg: &[u8]) -> bool {
-        let PublicKeyHolder::RistrettoSchnorr(k) = &self.pub_key;
-        match signature {
-            RistrettoSchnorr(s) => k.verify_message(s, msg),
+        match &self.pub_key {
+            PublicKeyHolder::Secp256k1Schnorr(ref k) => match signature {
+                Secp256k1Schnorr(s) => k.verify_message(s, msg),
+                _ => panic!("Wrong key/signature combination"),
```
stanislav-tkach, line 159: I'm just curious if it is OK to panic here.

stanislav-tkach (approved):

crypto/src/key/rschnorr/mod.rs
```diff
@@ -155,6 +162,105 @@ impl Derivable for MLRistrettoPrivateKey {
...
 
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct MLPrivateKey {
+    key_data: secp256k1::SecretKey,
+}
+
+impl zeroize::Zeroize for MLPrivateKey {
+    fn zeroize(&mut self) {
+        // TODO find a safe way to retrieve a mutable slice of the underlying SecretKey data array
+        let key_data = unsafe {
```
TheQuantumPhysicist, line 173: So, like I mentioned the other day, doing this manually with unsafe is basically playing with fire and quite frankly unnecessary (at this point). Off the top of my head, you can't guarantee the allocation length, which may or may not be a problem.

The issue here is that the slightest mistake will have a tremendous cost in the form memory corruption.

What I suggest doing is disabling this implementation (not necessarily completely removing it), and opening a github issue stating that this has to be studied, justified then done with address sanitizer checks combined with C code review, so that we guarantee we don't do it wrong.

crypto/src/key/rschnorr/mod.rs
```diff
@@ -155,6 +162,105 @@ impl Derivable for MLRistrettoPrivateKey {
...
+            std::slice::from_raw_parts_mut(
+                self.key_data.as_mut_ptr(),
+                secp256k1::constants::SECRET_KEY_SIZE,
+            )
+        };
+        key_data.zeroize();
+    }
+}
+
+impl Drop for MLPrivateKey {
```
TheQuantumPhysicist, line 183: I think you don't need to implement this... just inherit Zeroize and that'll do it.

crypto/src/key/rschnorr/mod.rs
```diff
@@ -155,6 +162,105 @@ impl Derivable for MLRistrettoPrivateKey {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct MLPrivateKey {
```
TheQuantumPhysicist, line 166: I would call this Secp256k1PrivateKey. The reason we used ML in the other one was because I didn't find a better wrapper name. Embarrassing... I know.

crypto/src/key/rschnorr/mod.rs
```diff
@@ -15,7 +15,9 @@
 
 mod internal;
 
+use crate::hash::{Blake2b32Stream, StreamHasher};
 use crate::random::{CryptoRng, Rng};
+use bitcoin::secp256k1;
```
TheQuantumPhysicist, line 20: Why are we doing this implementation in the same file... rschnoor? why not start another separate submodule where we put the new secp256k1 stuff?

Please create a directory, call it secp256k1schnorr, and put the new stuff in it... no need to crowd this file... it's becoming confusing.

crypto/src/key/key_holder.rs
```diff
@@ -13,16 +13,23 @@
...
 // limitations under the License.
 
+use crate::key::rschnorr::{MLPrivateKey, MLPublicKey};
 use serialization::{Decode, Encode};
 
 use super::rschnorr::{MLRistrettoPrivateKey, MLRistrettoPublicKey};
 
 #[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
 pub enum PrivateKeyHolder {
+    #[codec(index = 0)]
```
TheQuantumPhysicist, line 23: Thank you for adding indexes! Please add it to any enum that needs to be encoded if we forget.

crypto/src/key/mod.rs
TheQuantumPhysicist: No. I think it's not the right thing to do.

However, whether we return a `Result<bool>` or just `false` is an open question. I'm tempted to use `Result<bool>` and then allow the dev to use `unwrap_or(false)`, but the issue here is that we're gonna later remove the RistrettoSchnorr in a subsequent PR, so this won't be useful and the panic will disappear anyway because we'll go back to having one key type.

TheQuantumPhysicist (approved):
Besides the few issues I mentioned, looking good. Please fix them then feel free to merge. 

I'm mostly upset with the decision of unnecessarily intermixing both implementations in one file... I'm fighting so hard in all our crates to make files smaller as much as possible because we want our code to be readable. The other day I had a similar adventure in p2p, and before that in "derivables". Let's try to uphold that standard across all our code :smile:

crypto/src/key/signature/mod.rs
```diff
@@ -53,13 +60,18 @@ impl Decode for Signature {
...
+                Ok(Signature::Secp256k1Schnorr(sig))
+            }
             SignatureKind::RistrettoSchnorr => {
-                let sig = schnorrkel::Signature::from_bytes(&data).map_err(|_| {
-                    serialization::Error::from("Private Key deserialization failed")
-                })?;
+                let data = Vec::decode(input)?;
+                let sig = schnorrkel::Signature::from_bytes(&data)
+                    .map_err(|_| serialization::Error::from("Signature deserialization failed"))?;
                 Ok(Signature::RistrettoSchnorr(sig))
```
iljakuklic, line 75: There seems to be a bit of an inconsistency. When decoding `Secp256k1`, the payload is supposed to be a fixed number of bytes while `Ristrertto` uses a variable length encoding (based on `Vec`). Is that intentional? If new key types are added while the network is in operation, how is that supposed to be handled?

crypto/src/key/rschnorr/mod.rs
iljakuklic: I did some digging around how this could be done and here is an idea. Not sure it is viable but may be worth trying out.

The main insight is that the `Zeroize` trait is automatically implemented tor any type that implements `DefaultIsZeroes`. That could be used as follows:

1. Create an internal wrapper over `secp256k1::SecretKey`, with the following impls:
  a) `Copy`, `Clone`, `From<secp256k1::SecretKey>`
  b) `Default` that creates an all-zero private key. Could be done using `secp256k1::SecretKey::from_slice` but there may be a better way
  c) `DefaultIsZeroes`
  d) `Deref(Mut)<Target = secp256k1::SecretKey>` if the amount of unwrapping becomes tedious
2. If the wrapper type is used as `MLPrivateKey::key_data`, it should be enough to just `#[derive(Zeroize)]`

I know the likes of `Default` and `Deref` are somewhat controversial, that's why I suggest the wrapper type to be private or even separated into its own module. Care should be taken it is not exposed outside of this module.

iljakuklic (commented):
Adding a couple of comments to the outdated version of the code. May still be useful.

crypto/src/key/secp256k1/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+        self.as_bytes().size_hint()
+    }
+
+    fn encode_to<T: serialization::Output + ?Sized>(&self, dest: &mut T) {
+        self.as_bytes().encode_to(dest)
+    }
+
+    fn encode(&self) -> Vec<u8> {
+        self.as_bytes().encode()
+    }
```
iljakuklic, line 124: Same as above, default impls are good enough imho

crypto/src/key/secp256k1/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+        self.key_data.as_ref().encoded_size()
+    }
+}
+
+impl Decode for Secp256k1PrivateKey {
+    fn decode<I: serialization::Input>(input: &mut I) -> Result<Self, serialization::Error> {
+        let mut v = <[u8; secp256k1::constants::SECRET_KEY_SIZE]>::decode(input)?;
+        let result = secp256k1::SecretKey::from_slice(&v)
+            .map(|r| Secp256k1PrivateKey { key_data: r })
+            .map_err(|_| serialization::Error::from("Private Key deserialization failed"));
```
iljakuklic, line 60: Could we make the error more informative, something like the following:

```suggestion
            .map_err(|e| serialization::Error::from(format!("Private Key deserialization failed: {e}")));
```

(Not tested, please don't just blindly apply the change)

iljakuklic (commented):
Some more minor comments

crypto/src/key/secp256k1/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+        self.key_data.as_ref().size_hint()
+    }
+
+    fn encode_to<T: serialization::Output + ?Sized>(&self, dest: &mut T) {
+        self.key_data.as_ref().encode_to(dest);
+    }
+
+    fn encode(&self) -> Vec<u8> {
+        self.key_data.as_ref().encode()
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I thought about it too... but it doesn't really matter because they all reference the same type and same process. It's just a thin wrapper.

crypto/src/key/rschnorr/mod.rs
TheQuantumPhysicist: Your solution is a good one, Lukas, in the sense that no more unsafe is needed. Though I think we'd need another struct layer on top of that to prevent `Default::default()` from being used by developers... because a default for a private key being 0 is basically playing with fire... and that's an understatement.

iljakuklic: > Though I think we'd need another struct layer on top of that to prevent `Default::default()` from being used by developers...

That's exactly what I meant when I suggested the wrapper struct should be private/internal and used within `MLPrivateKey` so the users don't have direct access to it.

crypto/src/key/secp256k1/mod.rs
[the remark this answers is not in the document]
erasmospunk: Thanks for the tip, that code was copied from rschnorr.

crypto/src/key/rschnorr/mod.rs
erasmospunk: `secp256k1::SecretKey::from_slice` returns an error when the key is zero.

Furthermore if we managed the create a zeroed key, it would be still problematic because when signing with that key a panic would happen.

This is actually why the rust-bitcoin developers decided not to implement zeroize on the secret key.

What we could do, is to implement zeroing on drop. This can be studied on another issue.

iljakuklic: > `secp256k1::SecretKey::from_slice` returns an error when the key is zero.

Ok, this is an issue which will make my approach with `Default` not viable sadly.

crypto/src/key/signature/mod.rs
```diff
@@ -114,9 +128,45 @@ mod test {
...
+        let sig = sk.sign_message(msg).unwrap();
+        assert!(pk.verify_message(&sig, msg));
+
+        let encoded_sig = sig.encode();
+        let decoded_sig = Signature::decode(&mut encoded_sig.as_slice()).unwrap();
+        assert_eq!(decoded_sig, sig);
+    }
+
+    #[test]
+    fn serialize_chosen_data2() {
```
TheQuantumPhysicist, line 163: I guess here instead of 2 (and no number in the other instance) you can put the names of the algos because we don't have to read the first byte every time :smile:

TheQuantumPhysicist (approved):
Please fix the few minor issues I mentioned then feel free to merge

iljakuklic (approved):

PR #584, opened 2022-12-06
mempool: Simplify tasks that mempool is composed of 

iljakuklic:
Use only one task for mempool handling. Two tasks were used before to handle inter-subsystem calls as well as chainstate event handling. One task contained the logic and the seconf one was just dispatching calls and chainstate events. The call and event dispatch task had substantial amount of duplication with the subsystem mechanism. Letting the subsystem mechanism handle calls and events natively simplifies the task machinery to be on par with other subsystems. Clean mempool shutdown is now correctly handled as a side effect as well.


mempool/src/lib.rs
```diff
@@ -54,14 +54,14 @@ pub fn make_mempool<M>(
     chainstate_handle: subsystem::Handle<Box<dyn ChainstateInterface>>,
     time_getter: TimeGetter,
     memory_usage_estimator: M,
-) -> crate::Result<Box<dyn MempoolInterface>>
+) -> MempoolInterfaceImpl<M>
```
TheQuantumPhysicist, line 57: While this is better from a pure programmatic point of view, where we have less Box usage (even though compilers can optimize them nowadays), I actually prefer returning `Box<dyn MempoolInterface>` because that enforces adhering to the interface and not using some functionality the interface doesn't provide. As you probably have noticed, I've done this in all subsystems, so that we never have this problem in the future. A middle-ground solution is that you return an `impl MempoolInterface` on the stack instead of a Box, this way both problems are solved. No Box + no bypassing the interface methods.

TheQuantumPhysicist (approved):
Looking good. The as mentioned in the comment, maybe change that removed Box to impl and we're good. 

The removal of the impl_delegation I found to be unnecessary, it still can be useful for mocks to simulate errors when necessary. This is another reason why Box is preferred.

mempool/src/pool/mod.rs
```diff
@@ -115,7 +93,7 @@ where
         chainstate_handle: subsystem::Handle<Box<dyn ChainstateInterface>>,
         clock: TimeGetter,
         memory_usage_estimator: M,
-        receiver: tokio::sync::mpsc::UnboundedReceiver<MempoolMethodCall>,
+        //receiver: tokio::sync::mpsc::UnboundedReceiver<MempoolMethodCall>,
```
azarovh, line 96: remove commented code?

mempool/src/pool/mod.rs
```diff
@@ -128,97 +106,12 @@ where
...
-                    logging::log::error!("ContainsTransaction: Error sending response: {:?}", e);
-                }
-            }
-            MempoolMethodCall::SubscribeToEvents { handler, rtx } => {
-                self.subscribe_to_events(handler);
-                if let Err(e) = rtx.send(()) {
-                    logging::log::error!("SubscribeToEvents: Error sending response: {:?}", e);
-                }
-            }
+            //receiver,
```
azarovh, line 109: same

azarovh (approved):

mempool/src/interface/mempool_interface_impl.rs
```diff
@@ -13,100 +13,118 @@
...
+    pub async fn run(
+        mut self,
+        mut call_rq: CallRequest<dyn MempoolInterface>,
+        mut shut_rq: ShutdownRequest,
+    ) {
+        let mut chainstate_events_rx = self
+            .subscribe_to_chainstate_events()
+            .await
+            .log_err()
+            .expect("chainstate event subscribrtion");
```
TheQuantumPhysicist, line 61: subscription

mempool/src/lib.rs
iljakuklic: Understood. However in this case, however, `impl MempoolInterface` is not sufficient, since the method to start the subsystem task needs to be exposed too to start the subsystem.

In the node setup code (modified to emphasise the relevant parts):
```rust
let mempool_obj: MempoolInterfaceImpl<_> = make_mempool(/* args */);
let mempool: subsystem::Handle<dyn MempoolInterface> =
    manager.add_raw_subsystem("mempool", move |c, s| mempool_obj.run(c, s));
```

A couple of points here:
* The `run` method needs to be temporarily exposed to register the subsystem, and I don't think it's appropriate to add it to `MempoolInterface`.
* I added type annotations to make it clear that the mempool subsystem handle type is `subsystem::Handle<dyn MempoolInterface>`, so the scope where the concrete type is exposed is very limited. Once you get the subsystem handle, only `dyn MempoolInterface` is exposed.

I can try to hide `MempoolInterfaceImpl` a bit more, maybe by adding a trait that exposes the `run` method and has `MempoolInterface` as a supertrait. All that while only exposing `MempoolInterface` from the handle. I am not sure if it would work but can give this a shot. If it does not, I still think it's not a huge deal since the amount of exposure `MempoolInterfaceImpl` gets is so limited.

iljakuklic: I now added an extra trait so that the `MempoolInterfaceImpl` is not exposed

mempool/src/interface/mempool_interface_impl.rs
```diff
@@ -13,100 +13,141 @@
...
+            .expect("chainstate event subscription");
+        loop {
+            tokio::select! {
+                () = shut_rq.recv() => break,
+                call = call_rq.recv() => call(&mut self).await,
+                Some((block_id, block_height)) = chainstate_events_rx.recv() => {
+                    self.pool.new_tip_set(block_id, block_height);
+                }
+            }
+        }
```
iljakuklic, line 103: As per call: This code mixes mempool-specific code with code that handles calls and shutdown. The call and shutdown handling is normally provided by the subsystem framework but is unavailable if custom event loop is used. It would be preferable for the event loop + call handling + shutdown handling to be still provided by the subsystem framework while exposing extension points to give each subsystem the ability to react to inputs other than calls and shutdown requests. This is a matter for subsequent research.

PR #586, opened 2022-12-07
Rework undo for DeltaDataCollection v2

azarovh:
This is a continuation of work started in #577, but as a separate PR to avoid painful rebase. This time based on master.

I'll try to formulate the problem once more. I need deltas and undos to be associative. Meaning sequence like `(DB <- Delta) <- Undo` should be equal to `DB <- (Delta <- Undo)` where `<-` is merge operation. Previous approach where undo didn't contain information and only Erase operation didn't satisfy this property. 
This PR implements undo operation through separate delta type `DataDeltaUndo`, which can be applied to a delta from another collection or to a DB. Thus the information is not lost for undo, allowing us to flush delta to db at any moment and still be able to undo it later.

Moreover in a arbitrary chain of deltas and undos they can be merged in different order and still produce the same result, e.g:
Delta + Delta + Undo + Undo 
(Delta + Delta) + (Undo + Undo) 
Delta + ((Delta + Undo) + Undo) 
Delta + (Delta + (Undo + Undo))


accounting/src/delta/delta_data_collection/tests.rs
```diff
@@ -0,0 +1,786 @@
...
+fn test_combine_delta_with_undo() {
+    let delta_create = |c| DataDelta::Create(Box::new(c));
+    let delta_modify = |c1, c2| DataDelta::Modify((Box::new(c1), Box::new(c2)));
+    let delta_delete= |c| DataDelta::Delete(Box::new(c));
+
+    let undo_create = |c| DataDeltaUndo::Create(Box::new(c));
+    let undo_modify = |c1,c2| DataDeltaUndo::Modify((Box::new(c1), Box::new(c2)));
+    let undo_delete= |c| DataDeltaUndo::Delete(Box::new(c));
+
+    assert_eq!(combine_delta_with_undo(&delta_create('a'),      undo_create('b')),      Ok(undo_create('b')));
```
TheQuantumPhysicist, line 56: I have a feeling this might be incorrect. If I create then undo, the result is a no-op, or "termination" of the operation. In other words, if block at height N has a create operation, undoing N shouldn't result in an undo operation, but in a "no-operation", because there's nothing left to do... the undo has been done.

Maybe there should be tests with multiple steps that show that this preserves its properties. Something like:

A -> (B -> C)
(A -> B) -> C

where:

- A is a no-op
- B is a create 
- C is an undo of B

what we wanna ensure is that no matter how we do it, the final result is a no-op once we reach A and anything before A.

azarovh: Maybe you are right, indeed some cases might be converted to no-op instead of propagating further. I'll try out

accounting/src/delta/combine.rs
```diff
@@ -15,26 +15,46 @@
...
-            DataDelta::Delete => Err(Error::RemoveNonexistingData),
+            DeltaMapElement::Delta(d) => match d {
+                DataDelta::Create(d) => Ok(Some(*d.clone())),
+                DataDelta::Modify(_) => Err(Error::ModifyNonexistingData),
+                DataDelta::Delete(_) => Err(Error::RemoveNonexistingData),
+            },
+            DeltaMapElement::DeltaUndo(u) => match u {
+                DataDeltaUndo::Create(d) => Ok(Some(*d.clone())),
+                DataDeltaUndo::Modify(_) => Err(Error::ModifyNonexistingData),
+                // This is Ok because the Create + Undo = Delete
```
azarovh, line 41: This is not Ok. I'll rework and add real tests for data + delta, considering new changes with No-op. In the mean time please see `delta_data_collection` module

iljakuklic (commented):
Looks alright to me. Adding some minor comments and a point of discussion below:

Wondering if we can get away with something a bit simpler. Instead of the distinction between `DataDelta`, `DataDeltaUndo` and `DeltaDataCollection`, maybe it would be sufficient to have just one delta with the representation introduced in this PR. This delta can now be made invertible: `Create` becomes `Delete` and vice versa, `Modify` gets the elements flipped around.

This way, the `Option<Delta>` type would behave somewhat like a group (the mathematical structure):

* We have a type that implements `merge(x, y)` and `x.invert()`
* Identity is `None`, `merge(x, None) == merge(None, x) == x`
* Associative: `merge(x, merge(y, z)) == merge(merge(x, y), z)`
* Inverse: `merge(x, x.invert()) == None`
* Distributing inverses: `merge(x, y).inverse() == merge(y.inverse(), x.inverse())`

Or something close to a group if error handling and other details get in the way. Although I think the representation can be modified to be a group in presence of errors too. The `merge` operation composes deltas together, the `invert` operation implements undo, `None` is a trivial no-op delta. There could be a wrapper over `Option<Delta>` as to not expose the fact it's represented internally as an `Option`.

Now there is the question of having a type-level distinction between 'forward' and 'undo' deltas. It may be useful to know if the delta is intended to be used to apply "new" changes or undo existing changes. If that is required or desirable given the increased complexity, it could be done using a phantom type or wrappers instead without implementing the operations again. There are some complications though. What should `merge(x, y)` be where `x` is a forward delta and `y` is an undo delta? Should it be a forward or undo delta? It seems this would be fairly difficult to track at compilation time. This issue is not specific to having the, the current version in the PR has the same problem.

Any comments? What was the original motivation for the separation into three types? I may be missing something.

azarovh:
> What was the original motivation for the separation into three types? I may be missing something.

The idea to distinguish `DeltaDataUndo` from `DeltaData` was introduced to make the code more explicit. Indeed undo can be represented by a simple delta, but that would make user intention unclear. For example, we want to be able to apply additional rules for undo (right now applying delta over undo is forbidden for simplicity).
I agree that it looks duplicated, maybe an undo wrapper over DeltaData is better. Will play with it.

Regarding `DeltaDataCollection`. It's just a container for deltas. A map of arbitrary keys to either `DeltaData` or `DeltaDataUndo`. It is also useful because it encapsulates the rules of undo creation: undo is only created on delta+delta merge (not undo merge or delta+undo).

> Wondering if we can get away with something a bit simpler. 

I believe that your main concern is the reimplementation of operations for undo. Otherwise, the current approach is more restrictive in terms of how undo can be created and used.

accounting/src/delta/delta_data_collection/combine.rs
```diff
@@ -0,0 +1,205 @@
...
+                // Delete + Undo(Delete) produces Modify if the data has changed.
+                // This is introduced specifically for the case: (Modify + Delete) + Undo(Delete) = Modify
+                Ok(Some(DataDelta::Modify((lhs.clone(), rhs))))
+            }
+        }
+        (DataDelta::Delete(_), DataDeltaUndo::Modify(_)) => Err(Error::DeltaDataModifyAfterDelete),
+        (DataDelta::Delete(_), DataDeltaUndo::Delete(_)) => {
+            Err(Error::DeltaDataDeletedMultipleTimes)
+        }
+    }
```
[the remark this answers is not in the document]
azarovh: agree, this duplication is caused by how DeltaData and DeltaDataUndo are implemented. A wrapper over undo can eliminate this whole match.

accounting/src/delta/delta_data_collection/combine.rs
```diff
@@ -0,0 +1,205 @@
...
+        (DataDelta::Modify((d, _)), DataDelta::Delete(_)) => Ok(Some(DataDelta::Delete(d.clone()))),
+        (DataDelta::Delete(lhs), DataDelta::Create(rhs)) => {
+            if lhs == &rhs {
+                Ok(None)
+            } else {
+                Ok(Some(DataDelta::Modify((lhs.clone(), rhs))))
+            }
+        }
+        (DataDelta::Delete(_), DataDelta::Modify(_)) => Err(Error::DeltaDataModifyAfterDelete),
+        (DataDelta::Delete(_), DataDelta::Delete(_)) => Err(Error::DeltaDataDeletedMultipleTimes),
```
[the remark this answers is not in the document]
azarovh: That's a good question, but I'm not sure how important that is to solve. Curious what @TheQuantumPhysicist think.

One way I can think of is to accumulate deltas on merge instead of folding them. Then on some final step (like combining with data), we can apply accumulated deltas in a predefined order. This is pretty straightforward and avoids the complex logic of recovering information from error types. But again I'm not sure if we actually need this.

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -101,134 +143,119 @@ impl<K: Ord + Copy, T> DeltaDataCollection<K, T> {
...
+            Ok(None)
+        }
+        (DataDelta::Modify(_, _), DataDelta::Create(_)) => {
+            Err(Error::DeltaDataCreatedMultipleTimes)
+        }
+        (DataDelta::Modify(lhs1, lhs2), DataDelta::Modify(rhs1, rhs2)) => {
+            if lhs1 == &rhs2 && lhs2 == &rhs1 {
+                Ok(None)
+            } else {
+                Ok(Some(DataDelta::Modify(lhs1.clone(), rhs2)))
```
iljakuklic, line 241: Why the clone here? Seems `lhs1` is consumed by the condition, changing it to borrow `&lhs1 == &rhs2` should make it possible to avoid the clone here.

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -101,134 +143,119 @@ impl<K: Ord + Copy, T> DeltaDataCollection<K, T> {
...
-        (DataDelta::Delete, DataDelta::Create(d)) => Ok(DeltaMapOp::Write(DataDelta::Create(d))),
-        (DataDelta::Delete, DataDelta::Modify(_)) => Err(Error::DeltaDataModifyAfterDelete),
-        (DataDelta::Delete, DataDelta::Delete) => Err(Error::DeltaDataDeletedMultipleTimes),
+        (DataDelta::Delete(lhs), DataDelta::Create(rhs)) => {
+            if lhs == &rhs {
+                Ok(None)
+            } else {
+                Ok(Some(DataDelta::Modify(lhs.clone(), rhs)))
+            }
+        }
```
iljakuklic, line 254: I think there's a subtle issue here.  `None` is not the same as `Some(DeltaData::Modify(x, x))`. The former succeeds unconditionally. The latter errors out if the original value is not `x` functionally being equivalent to and assertion that the value is equal to `x`.

Imho this should be `Modify(lhs, rhs)` even if `lhs` and `rhs` are equal. That way we get the same result applying the deltas one by one as we do combining the deltas into one and applying it.

Similar issue is present in the `Create`/`Delete` pair. The `None` result is returned there which succeeds unconditionally. However, the individual operations only succeed if the value is initially not present (`None`).

In addition to that, the comment about cloning above also applies here.

iljakuklic (commented):
Just adding comments regarding some potential issues I noticed with the combine operation.

iljakuklic:
> The idea to distinguish `DeltaDataUndo` from `DeltaData` was introduced to make the code more explicit. Indeed undo can be represented by a simple delta, but that would make user intention unclear. For example, we want to be able to apply additional rules for undo (right now applying delta over undo is forbidden for simplicity). I agree that it looks duplicated, maybe an undo wrapper over DeltaData is better. Will play with it.
> 
> Regarding `DeltaDataCollection`. It's just a container for deltas. A map of arbitrary keys to either `DeltaData` or `DeltaDataUndo`. It is also useful because it encapsulates the rules of undo creation: undo is only created on delta+delta merge (not undo merge or delta+undo).

Sorry I meant whether `DeltaData`, `DeltaDataUndo` and `DeltaMapElement` could be somehow streamlined. I don't have a problem with `DeltaDataCollection`, that one just slipped in by accident.

accounting/src/delta/delta_data_collection/mod.rs
iljakuklic: In case of create/delete and create/modify pairs, there is an `ensure!` verifying the intermediate values match but not here. That is an inconsistency.

accounting/src/delta/delta_data_collection/mod.rs
azarovh: Interesting. I agree that producing a delta in all the cases can give us additional protection when merging the result with data. Or at least for the case with Modify.

But this raises a question of what the result of Create + Delete might be. Delete? 
Consider the cases:
```
Create + Delete = Delete
None + Create + Delete = Some + Delete = None
None + (Create + Delete) = None + Delete = Error
Some + Create + Delete = Error + Delete = Error
Some + (Create + Delete) = Some + Delete = None
```
or if it's a No-op:
```
Create + Delete = No-op
None + Create + Delete = Some + Delete = None
None + (Create + Delete) = None + No-op = None
Some + Create + Delete = Error + Delete = Error
Some + (Create + Delete) = Some + No-op = Some
```
The second variant is closer to the truth because it gives None for the 2nd case. But both of them don't produce an error for the 4th case, which unfortunately breaks our rule of associativity.

iljakuklic:
I gave the issue of composition in presence of errors some thought and this is what I ended up with. Please let me know your opinion.

### Specification

Before diving into the implementation, let me first lay down what we want to achieve accompanied with some high-level observations. Hopefully, we will be able to derive the correct implementation from that later on. Let us ignore for the moment the issue of representing missing values. For now, assume we are working with a value of type `T` that is always present and we only have the `Modify` operation, no `Create` or `Delete`. This apparent limitation will be reviewed later.

Fundamentally, a delta `Delta<T>` represents a function of type `Fn(T) -> Result<T>`. That is, a function that takes some data point of type `T` and either returns a new data point of type `T` or fails. We can find a concrete data type to use to represent deltas efficiently later. For now, to lay out semantics of deltas, how should deltas behave when applied to a value and how should they compose together, it is instructive to think of them as if they were implemented as the function type above.

The `Delta` type has a method `Delta::apply` taking `Self` and a `T` and producing a `Result<T>` which actually turns the delta into its function representation `|x| my_delta.apply(x) : impl Fn(T) -> Result<T>`. This is basically your `combine_delta_with_data`. The apply function for the "primitive" create/modify/delete operations is obvious enough, it just performs given operation.

Let us now look into how should deltas combine. Since deltas represent functions, combining deltas should behave like function composition (i.e. applying one function after another). However, the usual notion of function composition `compose(f, g) { |x| f(g(x)) }` does not quite cut it, because the result of `f` is a `Result<T>` and `g` expects a `T` as its argument. We need a modified version of function composition for fallible functions which looks something like `compose_fallible(f, g) { |x| Ok(f(g(x)?)?) }`. Here we take `f` and `g` satisfying `Fn(T) -> Result<T>` and we get one function that also satisfies `Fn(T) -> Result<T>`.

So for delta composition to be well behaved, the following three should all be equivalent functions:

```
|x| Ok(d1.apply(d0.apply(x)?)?)
|x| combine(d0, d1).apply(x)
compose_fallible(|x1| d1.apply(x1), |x0| d0.apply(x0))
```

In the above, the convention that `d0`, `d1`, etc are deltas `Delta<T>` whereas `x`, `x0`, `x1`, etc are values of type `T`.

Since function composition is associative (including the fallible version, assuming no side effects), making delta composition in effect implement function composition will make delta composition associative too once it is applied to data.

### Implementation

Here is one representation of the `Delta` type that seems to satisfy the above requirements:
```rust
enum Delta<T> {
    Modify(T, T), // Require the old value to be equal to the first component, update it to the second
    Mismatch,     // There was a old/new mismatch somewhere along the way
}
```

The semantics are defined in terms of apply as follows:

```rust
fn apply(self, x: T) -> Option<T> {
    match self {
        Delta::Modify(old, new) if x == old => Ok(new),
        _ => None,  // we have Delta::Mismatch or the old value does not match
    }
}
```

Using `Option` in place of `Result` since there is only one error state. Supporting multiple error types is a bit tricky as discussed below.

The delta composition goes like this:

```rust
fn combine<T>(d0: Delta<T>, d1: Delta<T>) -> Delta<T> {
    match (d0, d1) {
        (Delta::Modify(x0, x1), Delta::Modify(x2, x3)) if x1 == x2 => Delta::Modify(x0, x3),
        _ => Delta::Mismatch, // either one of deltas is a mismatch or the equality check above fails
    }
}
```

I think this satisfies the specification outlined in the previous section. Proof left as an exercise for the reader :laughing: 

#### Handling missing values

The above does not handle creating and deleting values. That functionality is easily recovered by using `Delta<Option<T>>`.

* `Create(x)` becomes `Delta::Modify(None, Some(x))`
* `Modify(x0, x1)` becomes `Delta::Modify(Some(x0), Some(x1))`
* `Delete(x)` becomes `Delta::Modify(Some(x), None)`
* The remaining combination `Delta::Modify(None, None)` amounts to failing unless the value is absent. This is different from a no-op delta (which is not representable with the current iteration of the `Delta` type, see below)

Note that since `Delta<U>` has various desirable properties (associativity of the combine operation) for all `U`,  `Delta<Option<T>>` has them too by setting `U = Option<T>`.

### Undo

The notion of inverting something is captured by the mathematical structure *group*. To make `Delta` into a group, we need two extra ingredients: the identity element and the inverse operation.

#### Identity

 We add the `Noop` arm as the identity element:
```rust
enum Delta<T> {
    Noop,
    Modify(T, T),
    Mismatch,
}
```

The `Delta::Noop` behaves like the identity function when applied to data element `x`. Composing `Noop` with any other delta `d` gives `d` (no matter whether noop is lhs or rhs).

It may be sensible to keep separate types for deltas with the noop arm and for deltas without it. When put in a map, the noop case can be represented by key not being present. The new `Delta` is isomorphic to `Option<OldDelta>`.

#### Inverse

Inverse has to satisfy `combine(d.inverse(), d) == combine(d, d.inverse()) == Delta::Noop`. The following does that:

```rust
Noop => Noop,
Modify(old, new) => Modify(new, old),
Mismatch => Mismatch,
```

#### Distinguishing "forward" and "backward" deltas at type level

Also touched upon in one of the other comments was that it may be useful to distinguish using types whether a delta applies new changes or undoes previously applied changes.

My preferred solution would be to add a phantom type that captures whether the intention is for the delta to be applied forward or backward. Inverse turns the forward annotation to backward and vice versa. The combine operation requires the directions to be the same but does not care otherwise. It is not clear to me what the resulting annotation should be when mixing directions, that's why I suggest for combine to work only on deltas going in the same direction. This could be revisited later.

### Tradeoffs

There are two things that I identified that we lose by employing this approach.

#### Less granular errors

Due to a more general representation, the various error cases (like double creation, modifying non-existing value, deleted value mismatch, etc) have been collapsed into one. Is this a serious issue?

Seems we could change `Mismatch` to `Mismatch(expected_value, actual_value)` for a richer error reporting, and the original error cases (like deleting a non-existing value, etc.) could be extracted from that. However, it gets more complicated once inverses enter the scene. Since we want the first error encountered to be propagated and inverting changes which error we encounter first, we would have to track two errors, one for the case when we apply the operations in the specified order and one for backwards order. At least all errors in the middle can still be dropped. Not sure it's worth investing the effort and extra complexity into this.

#### Early exit less convenient

Using `Result` or `Option` allows us to use the `?` operator when something goes wrong to exit early. When working with deltas, we often want to exit early as soon as we encounter a `Delta::Mismatch`. A nice principled way to do it is to implement `std::ops::Try` for `Delta` but it is currently unstable. An ugly option is to use a `Result<(T, T), MismatchError>` to represent `Delta` or have `Delta` easily convertible to it, so `?` can be used on it directly.

The combine operation requires both arguments to be evaluated no matter what which may result in inefficiency. We can provide a version of `combine` that evaluates rhs lazily:

```rust
fn combine_lazy<T>(d0: Delta<T>, d1: impl FnOnce() -> Delta<T>) -> Delta<T> {
    if matches!(d0, Delta::Mismatch) { return Delta::Mismatch }
    combine(d0, d1())
}
// use combine_lazy(a, || expensive_operation(b, c)) instead of combine(a, expensive_operation(b, c)) 
```

I don't particularly like any of these options apart from the one that requires unstable features.

### Conclusion

I hope I have not missed something obvious. This seems to work in my head :rofl:.

There is a number of advantages to the use of well-established mathematical structures. They are well researched so for example any result that someone has proven about groups also applies to deltas if we make deltas into a group. For some reason I suspect the crypto people here are familiar with groups in particular. Also the well-established structures tend to be very well behaved, predictable and composable.

accounting/src/delta/delta_data_collection/mod.rs
iljakuklic: > But this raises a question of what the result of Create + Delete might be.

I think to make `Create` + `Delete` equivalent to applying `Create` and  then applying `Delete` one by one, we need a delta that fails when an element is already there and does nothing otherwise. I.e. another case like `LeaveDeleted` is needed to handle it (in addition to `Create`, `Delete`, `Modify`).

I briefly mention the issue in the long comment I just posted. There a different representation is used where only `Modify` operation is there and missing values are represented by using `Option<_>` as `T`. Then it becomes `Modify(None, None)`

iljakuklic:
> Inverse has to satisfy `combine(d.inverse(), d) == combine(d, d.inverse()) == Delta::Noop`. The following does that:
> 
> ```rust
> Noop => Noop,
> Modify(old, new) => Modify(new, old),
> Mismatch => Mismatch,
> ```

Turns out this is not correct. The `Modify` case does not satisfy the inverse law as it gives a `Modify(old, old)` instead of a `Noop`. The `Modify(old, old)` is correct according to the original spec where we require applying deltas one by one to be equivalent to combining deltas and applying the combined delta.

There is something called [Inverse semigroup](https://en.wikipedia.org/wiki/Inverse_semigroup) which seems to fit the use case here much better. It has a weaker notion of inverse where `combine(d, d.inverse(), d) == d`. The following inverse operation, formulated on the original `Delta` without the extra identity element, satisfies that:

```rust
Modify(old, new) => Modify(new, old),
Mismatch => Mismatch,
```

Also we don't have to artificially introduce the identity element to make this formulation work, so the whole thing becomes somewhat simpler.

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+#[test]
+#[rustfmt::skip]
+fn test_combine_deltas() {
+    use DataDelta::{Create, Delete, Modify};
+
+    assert_eq!(combine_delta_data(&Create('a'), Create('b')),      Err(Error::DeltaDataCreatedMultipleTimes));
+    assert_eq!(combine_delta_data(&Create('a'), Modify('a', 'b')), Ok(Some(Create('b'))));
+    assert_eq!(combine_delta_data(&Create('a'), Delete('a')),      Ok(None));
+
+    assert_eq!(combine_delta_data(&Modify('a', 'b'), Create('c')),      Err(Error::DeltaDataCreatedMultipleTimes));
```
TheQuantumPhysicist, line 32: Can we have a separate error for this, `DeltaDataCreateAfterModify`?

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+    assert_eq!(combine_delta_data(&Modify('a', 'b'), Modify('c', 'd')), Ok(Some(Modify('a', 'd'))));
```
TheQuantumPhysicist, line 33: Are we sure this is valid, to modify a -> b then c -> d? Shouldn't it be a -> b then b -> c? Is allowing this OK for some reason I'm not noticing?

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+    assert_eq!(combine_delta_data(&Modify('a', 'b'), Modify('b', 'a')), Ok(None));
+    assert_eq!(combine_delta_data(&Modify('a', 'b'), Delete('b')),      Ok(Some(Delete('a'))));
+
+    assert_eq!(combine_delta_data(&Delete('a'), Create('a')),      Ok(None));
```
TheQuantumPhysicist, line 37: I think this is incorrect. Create then Delete will result in None. But Delete then Create should result in Create.

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -101,134 +143,119 @@ impl<K: Ord + Copy, T> DeltaDataCollection<K, T> {
...
     }
 }
 
-impl<K: Ord, T> Default for DeltaDataCollection<K, T> {
-    fn default() -> Self {
-        Self {
-            data: Default::default(),
-        }
+/// Returns an undo delta that has the opposite result of the provided delta
+fn create_undo_delta<T: Clone>(delta: DataDelta<T>) -> DataDeltaUndo<T> {
```
TheQuantumPhysicist, line 181: I would call this `undo_from_delta`

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+    //         Create - Delete
+    //         Modify - Modify
+    //         Modify - Delete
+    //         Delete - Create
+    //         Create - None
+    //         Modify - None
+    //         Delete - None
+    //         None   - Create
+    //         None   - Modify
+    //         None   - Delete
```
TheQuantumPhysicist, line 89: Please also list under those the invalid ones, so that we can verify that nothing is missed. I would even go as far as stating that the total possible combinations is 4^2=16, 11 of which are valid, and these 5 are not valid, and specify the reason if possible. These are the parts of the code that can face rigorous (human) audits.

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+            (10, DataDelta::Modify('a', 'n')),
+            (11, DataDelta::Delete('o')),
+        ]
+        .into_iter(),
+    );
+
+    let expected_data_after_merge = BTreeMap::from_iter(
+        [
+            (1, DeltaMapElement::Delta(DataDelta::Create('i'))),
+            // 2 is No-op
```
TheQuantumPhysicist, line 122: I like these comments. Please add them also in collection1 and collection2.

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        assert_eq!(delta1, undo.0);
+    };
+
+    check_undo(DataDelta::Modify('a', 'b'), DataDelta::Modify('b', 'a'));
+    check_undo(DataDelta::Create('a'), DataDelta::Delete('a'));
+    check_undo(DataDelta::Delete('a'), DataDelta::Create('a'));
+}
+
+#[test]
+fn merge_collections_and_undo() {
```
TheQuantumPhysicist, line 75: I'm OK with having this test, but do we have a similar test at the combine level? Or is this just at the collection merge level?

TheQuantumPhysicist: I'm talking specifically about the undos, btw. It seems that this test is special because it tests the undo cases.

accounting/src/delta/delta_data_collection/tests/delta_delta_undo_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        let _ = collection2.merge_delta_data(collection3).unwrap();
+        let _ = collection1.merge_delta_data(collection2).unwrap();
+
+        assert_eq!(collection1, expected_collection);
+    }
+}
+
+#[test]
+fn create_delete_undo_associativity() {
```
TheQuantumPhysicist, line 58: Why is this a separate test from the one above?

accounting/src/delta/delta_data_collection/tests/delta_delta_undo_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+);
+
+#[rstest]
+#[case(make_collections_with_undo(Create('a'), Modify('a', 'b')))]
+#[case(make_collections_with_undo(Modify('a', 'b'), Modify('b', 'c')))]
+#[case(make_collections_with_undo(Modify('a', 'b'), Delete('b')))]
+#[case(make_collections_with_undo(Delete('a'), Create('b')))]
+fn delta_delta_undo_associativity(#[case] collections: ThreeCollections) {
+    {
+        // Delta1 + Delta2 + Undo = Delta1
```
TheQuantumPhysicist, line 35: When you say Delta1 + Delta2 + Undo, it's not clear which addition is done first. I know this is clear here as it's opposed to the case below, but still, please state it as (Delta1 + Delta2) + Undo

accounting/src/delta/delta_data_collection/tests/delta_delta_undo_undo_tests.rs
```diff
@@ -0,0 +1,91 @@
...
+#[case(make_collections_with_undo(DataDelta::Create('a'), DataDelta::Delete('a')))]
+#[case(make_collections_with_undo(DataDelta::Create('a'), DataDelta::Modify('a', 'b')))]
+#[case(make_collections_with_undo(DataDelta::Modify('a', 'b'), DataDelta::Modify('b', 'c')))]
+#[case(make_collections_with_undo(DataDelta::Modify('a', 'b'), DataDelta::Delete('b')))]
+#[case(make_collections_with_undo(DataDelta::Delete('a'), DataDelta::Create('b')))]
+fn delta_delta_undo_undo_associativity(#[case] collections: FourCollections) {
+    let expected_collection = DeltaDataCollection::new();
+
+    {
+        // Delta1 + Delta2 + Undo1 + Undo2 = No-op
```
TheQuantumPhysicist, line 37: Same here, please write it as ((Delta1 + Delta2) + Undo1) + Undo2

accounting/src/delta/delta_data_collection/tests/delta_delta_undo_undo_tests.rs
```diff
@@ -0,0 +1,91 @@
...
+        let (mut collection1, collection2, mut collection3, collection4) = collections.clone();
+        let _ = collection1.merge_delta_data(collection2).unwrap();
+        let _ = collection3.merge_delta_data(collection4).unwrap();
+        let _ = collection1.merge_delta_data(collection3).unwrap();
+
+        assert_eq!(collection1, expected_collection);
+    }
+
+    {
+        // Delta1 + (Delta2 + Undo1 + Undo2) = No-op
```
TheQuantumPhysicist, line 57: Same here, please state which comes first

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+#[case(
+    make_expected_collection(DataDelta::Delete('a')),
+    make_three_collections(Modify('a', 'b'), Modify('b', 'c'), Delete('c'))
+)]
+fn delta_delta_delta_associativity(
+    #[case] expected_collection: DeltaDataCollection<i32, char>,
+    #[case] collections: ThreeCollections,
+) {
+    {
+        // Deleta + Delta + Delta = Delta
```
TheQuantumPhysicist, line 66: Typo, also please make the order clear like mentioned before.

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+        let (mut collection1, collection2, collection3) = collections.clone();
+        let _ = collection1.merge_delta_data(collection2).unwrap();
+        let _ = collection1.merge_delta_data(collection3).unwrap();
+
+        assert_eq!(collection1, expected_collection);
+    }
+
+    {
+        // Deleta + (Delta + Delta) = Delta
```
TheQuantumPhysicist, line 75: Typo

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+)]
+#[case(
+    DeltaDataCollection::new(),
+    make_three_collections(Create('a'), Modify('a', 'b'), Delete('b'))
+)]
+#[case(
+    make_expected_collection(DataDelta::Create('b')),
+    make_three_collections(Create('a'), Delete('a'), Create('b'))
+)]
+#[case(
```
TheQuantumPhysicist, line 57: I think there are way more possible cases than just these 4

accounting/src/delta/delta_data_collection/tests/delta_delta_undo_tests.rs
```diff
@@ -0,0 +1,103 @@
...
+        let (mut collection1, collection2, collection3) = collections.clone();
+        let expected_collection = collection1.clone();
```
TheQuantumPhysicist, line 37: Shouldn't the expected result be calculated in common outside the scope?

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_delta_tests.rs
```diff
@@ -0,0 +1,139 @@
...
+#[case(
+    make_expected_collection(DataDelta::Delete('a')),
+    make_four_collections(Delete('a'), Create('b'), Modify('b', 'c'), Delete('c'))
+)]
+fn delta_delta_delta_delta_associativity(
+    #[case] expected_collection: DeltaDataCollection<i32, char>,
+    #[case] collections: FourCollections,
+) {
+    {
+        // Delta1 + Delta2 + Delta3 + Delta4 = Delta
```
TheQuantumPhysicist, line 101: Please specify the order here too, and all the following ones

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_delta_tests.rs
```diff
@@ -0,0 +1,139 @@
...
+    (
+        DeltaDataCollection::from_iter([(1, delta1)]),
+        DeltaDataCollection::from_iter([(1, delta2)]),
+        DeltaDataCollection::from_iter([(1, delta3)]),
+        DeltaDataCollection::from_iter([(1, delta4)]),
+    )
+}
+
+#[rstest]
+#[case(
```
TheQuantumPhysicist, line 48: I don't think these cases cover all the possibilities. I'm even OK with writing a little script to generate everything possible.

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -14,52 +14,94 @@
 // limitations under the License.
 
 pub mod undo;
+use self::undo::*;
 
 use std::collections::BTreeMap;
 
 use serialization::{Decode, Encode};
+use utils::ensure;
```
TheQuantumPhysicist, line 22: I really don't think this is necessary... import the whole utils module just for this thing... I don't really care for it. I'd rather have one less dependency unless we must.

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -14,52 +14,94 @@
...
+    pub fn get_data_delta(&self) -> &DataDelta<T> {
+        match self {
+            DeltaMapElement::Delta(d) => d,
+            DeltaMapElement::DeltaUndo(d) => &d.0,
+        }
+    }
 }
 
-impl<K: Ord + Copy, T> DeltaDataCollection<K, T> {
+pub enum GetDataResult<T> {
```
TheQuantumPhysicist, line 59: I'm wondering whether this is necessary... can't we learn this from the map itself without having this intermediary type?

accounting/src/delta/combine.rs
```diff
@@ -66,23 +69,43 @@ pub fn combine_amount_delta(
...
+        let delta_delete= Some(DeltaMapElement::Delta(DataDelta::Delete('b')));
+
+        assert_eq!(combine_data_with_delta::<i32>(None, None),                  Ok(None));
+        assert_eq!(combine_data_with_delta(None,        delta_create.as_ref()), Ok(Some('b')));
+        assert_eq!(combine_data_with_delta(None,        delta_modify.as_ref()), Err(Error::ModifyNonexistingData));
+        assert_eq!(combine_data_with_delta(None,        delta_delete.as_ref()), Err(Error::RemoveNonexistingData));
+        assert_eq!(combine_data_with_delta(Some(&'a'),  None),                  Ok(Some('a')));
+        assert_eq!(combine_data_with_delta(Some(&'a'),  delta_create.as_ref()), Err(Error::DataCreatedMultipleTimes));
+        assert_eq!(combine_data_with_delta(Some(&'a'),  delta_modify.as_ref()), Ok(Some('b')));
+        assert_eq!(combine_data_with_delta(Some(&'a'),  delta_delete.as_ref()), Ok(None));
```
TheQuantumPhysicist, line 91: So `Some('a')` then `Delete('b')` is `None`?

TheQuantumPhysicist: Please include the content in the variable name so that such mistakes don't happen.

TheQuantumPhysicist:
> 

I like the formulation and I think it's worth giving a try. I'd like to expand on this and say that the elements you mentioned and the semi-group seem to form a symmetric/hermitian transformation in Hilbert space. One can see that a data (even empty data) is a column vector (A, B) that represents the data of an account from state A to state B. The transformation function is a hermitian/symmetric matrix multiplication that moves the data among these states. If we represent the data in these states, they're guaranteed to be reversible. I think if we want to go nuts we can have a full mathematical formulation of this.

TheQuantumPhysicist:
> I hope I have not missed something obvious. This seems to work in my head

I like the formulation and I think it's worth giving a try. I'd like to expand on this and say that the elements you mentioned and the semi-group seem to form a symmetric/hermitian transformation in Hilbert space. One can see that a data (even empty data) is a column vector (A, B) that represents the data of an account from state A to state B. The transformation function is a hermitian/symmetric matrix multiplication that moves the data among these states. If we represent the data in these states, they're guaranteed to be reversible. I think if we want to go nuts we can have a full mathematical formulation of this.

azarovh:
Added more tests and some fixes. New tests fail which shows that the current approach has problems.

azarovh:
Closing this PR in favor of #613. All comments are addressed there.

PR #592, opened 2022-12-10
Try to find what causes the functional test to fail

stanislav-tkach:


stanislav-tkach:
OK, so bumping semver compatible versions of all dependencies doesn’t cause any issues.

stanislav-tkach:
The issue is in the `is-terminal` crate that is now used by `env_logger` instead of `atty`.

PR #600, opened 2022-12-14
Minor bans fixes

pavel-kokolemin:
Some disabled unit tests were failing because peer manager was expecting that the network backend will block incoming connections from banned addresses. At first I tried to re-implement same logic in mock backend. That worked but I think the better way is to keep bans logic in one place (peer manager). Overall I think what we already have is enough for bans to work.

For reference how bans work in bitcoin:
Bitcoin has two mechanisms: Banning and Discouraging.
1. Banning is done via RPC only. IP addresses or subnets are banned (for 24 hours by default). DB is persistent.
2. Discouraging is done automatically for misbehaved or incompatible peers. DB is not persistent but IP addresses are stored in rolling bloom filter (revoking in not possible but older entries replaced as new entries are added).
Incoming connections from discoraged peers are allowed but are first candidates for eviction if incoming slots are full.
Outgoing connections and gossips are not allowed. 

Our implementation is different because we don't separate "banning" and "discouraging".

I also noticed that `bitcoind` never bans local and manual connections:

```
    if (pnode.HasPermission(NetPermissionFlags::NoBan)) {
        // We never disconnect or discourage peers for bad behavior if they have NetPermissionFlags::NoBan permission
        LogPrintf("Warning: not punishing noban peer %d!\n", peer.m_id);
        return false;
    }

    if (pnode.IsManualConn()) {
        // We never disconnect or discourage manual peers for bad behavior
        LogPrintf("Warning: not punishing manually connected peer %d!\n", peer.m_id);
        return false;
    }

    if (pnode.addr.IsLocal()) {
        // We disconnect local peers for bad behavior but don't discourage (since that would discourage
        // all peers on the same local address)
        LogPrint(BCLog::NET, "Warning: disconnecting but not discouraging %s peer %d!\n",
                 pnode.m_inbound_onion ? "inbound onion" : "local", peer.m_id);
        pnode.fDisconnect = true;
        return true;
    }
```

### DIFF
--- a/node/src/config_files/mod.rs
+++ b/node/src/config_files/mod.rs
@@ -135,6 +135,7 @@ fn p2p_config(config: P2pConfigFile, options: &RunOptions) -> P2pConfigFile {
     let P2pConfigFile {
         bind_address,
         ban_threshold,
+        ban_duration,
         outbound_connection_timeout,
         mdns_config: _,
         request_timeout,
@@ -157,6 +158,7 @@ fn p2p_config(config: P2pConfigFile, options: &RunOptions) -> P2pConfigFile {
     P2pConfigFile {
         bind_address,
         ban_threshold,
+        ban_duration,
         outbound_connection_timeout,
         mdns_config,
         request_timeout,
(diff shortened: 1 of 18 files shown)


TheQuantumPhysicist (approved):
Currently we care only about discouraging, but we certainly need both in the future as nodes should be able to ban peers they don't want to connect to.

iljakuklic (approved):

muursh (approved):

PR #601, opened 2022-12-15
Add support for derivable extended keys for Secp256k1

erasmospunk:
This commit adds an implementation of the BIP32 derivation algorithm that supports "hard" derivations only. In addition to that the bip39 dependency is added for converting mnemonic phrases to extended master key secrets.

Note that NFKD normalisation is disabled in bip39 because of pinned dependencies and it's version should be bumped once it becomes available.

### DIFF
--- a/crypto/src/key/hdkd/child_number.rs
+++ b/crypto/src/key/hdkd/child_number.rs
@@ -34,42 +34,29 @@ enum DerivationType {
 
 impl ChildNumber {
     /// Return a hardened child number
-    pub fn from_hardened(index: U31) -> Result<Self, DerivationError> {
-        Ok(ChildNumber(DerivationType::Hardened(index)))
-    }
-
-    /// Return a normal child number
-    fn from_normal(_index: U31) -> Result<Self, DerivationError> {
-        // For the time being we don't support non-hardened derivations
-        Err(DerivationError::UnsupportedDerivationType)
+    pub fn from_hardened(index: U31) -> Self {
+        ChildNumber(DerivationType::Hardened(index))
     }
 
     /// Return a child based on the hardened bit (MSB bit)
     pub fn from_index_with_hardened_bit(index: u32) -> Result<Self, DerivationError> {
         let (index_u31, msb) = U31::from_u32_with_msb(index);
         if msb {
-            Self::from_hardened(index_u31)
+            Ok(Self::from_hardened(index_u31))
         } else {
-            Self::from_normal(index_u31)
+            Err(DerivationError::UnsupportedDerivationType)
         }
     }
 
     /// Return a BIP32-like child number index that has the hardened bit set if needed
-    pub fn to_encoded_index(self) -> u32 {
+    pub fn into_encoded_index(self) -> u32 {
         match self.0 {
             DerivationType::Hardened(i) => i.into_encoded_with_msb(true),
         }
     }
 
-    pub fn to_encoded_be_bytes(self) -> [u8; 4] {
-        self.to_encoded_index().to_be_bytes()
-    }
-
-    /// Returns true if this child number is hardened
-    pub fn is_hardened(&self) -> bool {
-        match self.0 {
-            DerivationType::Hardened(_) => true,
-        }
+    pub fn into_encoded_be_bytes(self) -> [u8; 4] {
+        self.into_encoded_index().to_be_bytes()
     }
 }
 
@@ -96,9 +83,9 @@ impl FromStr for ChildNumber {
         let index = U31::from_str(index_str)?;
 
         if is_hardened {
-            ChildNumber::from_hardened(index)
+            Ok(ChildNumber::from_hardened(index))
         } else {
-            ChildNumber::from_normal(index)
+            Err(DerivationError::UnsupportedDerivationType)
         }
     }
 }
--- a/crypto/src/key/hdkd/child_number.rs
+++ b/crypto/src/key/hdkd/child_number.rs
@@ -61,6 +61,10 @@ impl ChildNumber {
         }
     }
 
+    pub fn to_encoded_be_bytes(self) -> [u8; 4] {
+        self.to_encoded_index().to_be_bytes()
+    }
+
     /// Returns true if this child number is hardened
     pub fn is_hardened(&self) -> bool {
         match self.0 {
--- a/crypto/src/key/hdkd/derivation_path.rs
+++ b/crypto/src/key/hdkd/derivation_path.rs
@@ -27,10 +27,12 @@ const SEPARATOR: &str = "/";
 pub struct DerivationPath(Vec<ChildNumber>);
 
 impl DerivationPath {
-    pub fn extend(&self, num: ChildNumber) -> Self {
-        let mut new_path = self.0.clone();
-        new_path.push(num);
-        new_path.into()
+    pub fn push(&mut self, num: ChildNumber) {
+        self.0.push(num);
+    }
+
+    pub fn pop(&mut self) -> Option<ChildNumber> {
+        self.0.pop()
     }
 }
 
@@ -82,7 +84,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_parse_derivation_path() {
+    fn parse_derivation_path() {
         assert_eq!(
             DerivationPath::from_str(""),
             Err(DerivationError::InvalidDerivationPathFormat)
@@ -172,45 +174,59 @@ mod tests {
         assert_eq!(DerivationPath::from_str("m"), Ok(vec![].into()));
         assert_eq!(
             DerivationPath::from_str("m/0'"),
-            Ok(vec![ChildNumber::from_hardened(0.try_into().unwrap()).unwrap()].into())
+            Ok(vec![ChildNumber::from_hardened(0.try_into().unwrap())].into())
         );
         assert_eq!(
             DerivationPath::from_str("m/0h/1'/2'"),
             Ok(vec![
-                ChildNumber::from_hardened(0.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(1.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(2.try_into().unwrap()).unwrap(),
+                ChildNumber::from_hardened(0.try_into().unwrap()),
+                ChildNumber::from_hardened(1.try_into().unwrap()),
+                ChildNumber::from_hardened(2.try_into().unwrap()),
             ]
             .into())
         );
         assert_eq!(
             DerivationPath::from_str("m/0'/1'/2h/2h"),
             Ok(vec![
-                ChildNumber::from_hardened(0.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(1.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(2.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(2.try_into().unwrap()).unwrap(),
+                ChildNumber::from_hardened(0.try_into().unwrap()),
+                ChildNumber::from_hardened(1.try_into().unwrap()),
+                ChildNumber::from_hardened(2.try_into().unwrap()),
+                ChildNumber::from_hardened(2.try_into().unwrap()),
             ]
             .into())
         );
         assert_eq!(
             DerivationPath::from_str("m/0'/1'/2'/2'/1000000000'"),
             Ok(vec![
-                ChildNumber::from_hardened(0.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(1.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(2.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(2.try_into().unwrap()).unwrap(),
-                ChildNumber::from_hardened(1000000000.try_into().unwrap()).unwrap(),
+                ChildNumber::from_hardened(0.try_into().unwrap()),
+                ChildNumber::from_hardened(1.try_into().unwrap()),
+                ChildNumber::from_hardened(2.try_into().unwrap()),
+                ChildNumber::from_hardened(2.try_into().unwrap()),
+                ChildNumber::from_hardened(1000000000.try_into().unwrap()),
             ]
             .into())
         );
         assert_eq!(
             DerivationPath::from_str("m/2147483647'"),
-            Ok(vec![ChildNumber::from_hardened(2147483647.try_into().unwrap()).unwrap()].into())
+            Ok(vec![ChildNumber::from_hardened(2147483647.try_into().unwrap())].into())
         );
         assert_eq!(
             DerivationPath::from_str("m/2147483648'"),
             Err(DerivationError::InvalidChildNumber(2147483648))
         );
+    }
+
+    #[test]
+    fn push_pop_path() {
+        let path_1_2 = DerivationPath::from_str("m/1'/2'").unwrap();
+        let mut path = path_1_2.clone();
+        let index_3 = ChildNumber::from_hardened(3.try_into().unwrap());
+        path.push(index_3);
+        assert_eq!(path, DerivationPath::from_str("m/1'/2'/3'").unwrap());
+        assert_eq!(path.pop(), Some(index_3));
+        assert_eq!(path, path_1_2);
+        path.pop();
+        path.pop();
+        assert_eq!(path.pop(), None);
     }
 }
--- a/crypto/src/key/hdkd/derivation_path.rs
+++ b/crypto/src/key/hdkd/derivation_path.rs
@@ -26,6 +26,14 @@ const SEPARATOR: &str = "/";
 #[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
 pub struct DerivationPath(Vec<ChildNumber>);
 
+impl DerivationPath {
+    pub fn extend(&self, num: ChildNumber) -> Self {
+        let mut new_path = self.0.clone();
+        new_path.push(num);
+        new_path.into()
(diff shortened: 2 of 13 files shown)


crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+        mac.update(seed);
+        let mut result = mac.finalize().into_bytes();
+
+        let private_key = secp256k1::SecretKey::from_slice(&result[..32])
+            .map_err(|_| DerivationError::KeyDerivationError)?
+            .into();
+
+        let chain_code: [u8; CHAINCODE_LENGTH] =
+            result[32..].try_into().expect("Chaincode size is 32 bytes");
+        let chain_code = ChainCode::from(chain_code);
```
iljakuklic, line 55: You can use the `.split()` method on `GenericArray` to extract the key part and the chain code part with compile-time bounds checking.

Something like this (may need adjustment):
```rust
use generic_array::{GenericArray, typenum::U32, sequence::Split};
type ByteArray32 = GenericArray<u8, U32>;
let (private_key, chian_code): (ByteArray32, ByteArray32) = result.split();
```

The `ByteArray32` type is going to be convertible to `[u8; 32]` and `&[u8]`.

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// Chain code
+    pub chain_code: ChainCode,
+    /// Private key
+    pub private_key: Secp256k1PrivateKey,
+}
+
+impl Secp256k1ExtendedPrivateKey {
+    pub fn new_master(seed: &[u8]) -> Result<Secp256k1ExtendedPrivateKey, DerivationError> {
+        let mut mac =
+            HmacSha512::new_from_slice(b"Bitcoin seed").expect("HMAC can take key of any size");
```
iljakuklic, line 45: What's the purpose of the `"Bitcoin seed"` string here. Is that for compatibility so that we generate identical key tree to Bitcoin?

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+        let private_key = secp256k1::SecretKey::from_slice(&result[..32])
+            .map_err(|_| DerivationError::KeyDerivationError)?
+            // TODO [SECURITY] a Scalar is created here but not erased
+            .add_tweak(&secp_key.into())
+            .map_err(|_| DerivationError::KeyDerivationError)?
+            .into();
+
+        let chain_code: [u8; CHAINCODE_LENGTH] =
+            result[32..].try_into().expect("Chaincode size is 32 bytes");
+        let chain_code = ChainCode::from(chain_code);
```
iljakuklic, line 117: This part looks very similar to code in `Self::new_master`, could that be factored out?

iljakuklic (approved):
Looks fine to me, just a couple of comments / suggestions. Please wait for an approval by @muursh or @TheQuantumPhysicist before merging.

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+    use super::*;
+    use crate::key::hdkd::derivation_path::DerivationPath;
+    use bip39::Mnemonic;
+    use hex::ToHex;
+    use serialization::DecodeAll;
+    use serialization::Encode;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_serialization() {
```
stanislav-tkach, line 160: Nitpick: there isn't much sense to use the `test_` prefix.

stanislav-tkach (approved):

crypto/src/key/hdkd/child_number.rs
```diff
@@ -61,6 +61,10 @@ impl ChildNumber {
         }
     }
 
+    pub fn to_encoded_be_bytes(self) -> [u8; 4] {
```
TheQuantumPhysicist, line 64: into, instead of to.

You can also implement the From trait as an alternative solution, which will implement Into for you.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -26,6 +26,14 @@ const SEPARATOR: &str = "/";
 #[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
 pub struct DerivationPath(Vec<ChildNumber>);
 
+impl DerivationPath {
+    pub fn extend(&self, num: ChildNumber) -> Self {
+        let mut new_path = self.0.clone();
+        new_path.push(num);
+        new_path.into()
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I would rather call this join than extend. Both Rust and Python call it join. C++ calls it append.

Also one nice thing boost does is using the operator `/` for these kinds of extensions... it's kind of weird, but shows nicely in code. So if `a`, `b` and `c` are paths, we write in code:

```
let d = a / b / c;
```

In fact, the C++ standard adopted it since C++17: https://en.cppreference.com/w/cpp/filesystem/path/append

I'm not suggesting we go that far, but it's an idea worth thinking about if these joins become annoying or something. If I see it, I won't fight it.

crypto/src/key/secp256k1/extended_keys.rs
TheQuantumPhysicist: Yes, it's: https://en.bitcoin.it/wiki/BIP_0032#Master_key_generation

But I'm wondering now whether we want to parametrize this for our own use. Is this a standard that everyone uses in every crypto? Or is it just in bitcoin and everyone adapts it to whatever they want?

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+            return Err(DerivationError::UnsupportedDerivationType);
+        }
+
+        // Add the child number
+        mac.update(&num.to_encoded_be_bytes());
+
+        let mut result = mac.finalize().into_bytes();
+        let private_key = secp256k1::SecretKey::from_slice(&result[..32])
+            .map_err(|_| DerivationError::KeyDerivationError)?
+            // TODO [SECURITY] a Scalar is created here but not erased
```
TheQuantumPhysicist, line 110: We usually follow the format TODO(TAG). It makes it easier to search for all tagged TODOs.

crypto/src/key/hdkd/child_number.rs
TheQuantumPhysicist: Btw, same applies to the function above

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+            .expect("HMAC can take key of any size");
+
+        let secp_key = self.private_key.data;
+
+        // We only support hard derivations
+        if num.is_hardened() {
+            mac.update(&[0u8]);
+            mac.update(&secp_key[..]);
+        } else {
+            return Err(DerivationError::UnsupportedDerivationType);
```
TheQuantumPhysicist, line 101: Isn't this unreachable from ChildNumber itself, guaranteed from the type? Why are we checking for it here again?

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        // Add the child number
+        mac.update(&num.to_encoded_be_bytes());
```
TheQuantumPhysicist, line 105: Please move all these stateful mac updates into a separate function. The rule of thumb is: Whenever we play with `mut`, we try to isolate it. It's not always possible, but we do it whenever we can, and it seems to be possible in your case.

So, do the check if `num` is hardened (which I'm not convinced is necessary, since like I said, it's not supported anyway at the `ChildNumber` level). I don't know why we're checking then appending `0u8` followed by `secp_key`. Does being hardened have any bearing on that? If yes, it would be nice to move any specifical part of "being hardened" into the `ChildNumber` struct, and just retrieve it with a method. I think we can at least take the `secp_key` appending out as it's neutral and doesn't matter.

Then what do we have left? We create a function that takes a ref to the private key data + a ref to num, and returns the result array. That would be beautifully more readable and separately testable.

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -0,0 +1,256 @@
...
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode)]
+pub struct Secp256k1ExtendedPublicKey {
+    /// Chain code
+    pub chain_code: ChainCode,
+    /// Public key
+    pub public_key: Secp256k1PublicKey,
+}
+
+impl Secp256k1ExtendedPublicKey {
+    pub fn as_public_key(&self) -> &Secp256k1PublicKey {
```
TheQuantumPhysicist, line 137: Please call this function `public_key()`. It's a simple getter.

crypto/src/key/mod.rs
```diff
@@ -43,10 +48,18 @@ pub enum SignatureError {
 pub enum KeyKind {
     #[codec(index = 0)]
     Secp256k1Schnorr,
+    // Index 1 is used for extended Secp256k1Schnorr
```
TheQuantumPhysicist, line 51: I don't know why we're doing this. These don't have to be mutually exclusive. In fact, it's not really possible to encode/decode one into the other. Usually, when you decode bytes into whole objects, you use `decode_all`, which will guarantee that the whole object is filled.

Please don't use this unnecessary and confusing limitation. The only relationship between KeyKind and ExtendedKeyKind is that ExtendedKey can be converted into Key. There's no other hidden relationships here. We don't want the reader of the code to be solving puzzles.

crypto/src/key/mod.rs
```diff
@@ -43,10 +48,18 @@ pub enum SignatureError {
...
     #[codec(index = 2)]
     RistrettoSchnorr,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
+pub enum ExtendedKeyKind {
+    // Index 0 is used for Secp256k1Schnorr
+    #[codec(index = 1)]
```
TheQuantumPhysicist, line 59: Based on the discussion above, this has to be `indes = 0`

crypto/src/key/mod.rs
```diff
@@ -140,6 +163,81 @@ impl From<MLRistrettoPrivateKey> for PrivateKey {
     }
 }
 
+impl ExtendedPrivateKey {
```
TheQuantumPhysicist, line 166: Shouldn't there be a method that extracts the private key from the extended private key? How else are we gonna sign stuff?

crypto/src/key/mod.rs
```diff
@@ -140,6 +163,81 @@ impl From<MLRistrettoPrivateKey> for PrivateKey {
...
+            ExtendedPrivateKeyHolder::Secp256k1Schnorr(_) => ExtendedKeyKind::Secp256k1Schnorr,
+        }
+    }
+
+    pub(crate) fn get_internal_key(&self) -> &ExtendedPrivateKeyHolder {
+        &self.key
+    }
+}
+
+impl ExtendedPublicKey {
```
TheQuantumPhysicist, line 205: Same... where's the conversion function?

crypto/src/key/mod.rs
```diff
@@ -57,6 +70,16 @@ pub struct PublicKey {
     pub_key: PublicKeyHolder,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
```
TheQuantumPhysicist, line 73: Please move all these new things into an `extended.rs` module. You're also free to move everything else and make this file just for `mod` and `use` statements. We try everything we can to avoid clutter in code.

crypto/src/key/mod.rs
```diff
@@ -140,6 +163,81 @@ impl From<MLRistrettoPrivateKey> for PrivateKey {
...
+    pub fn from_private_key(private_key: &ExtendedPrivateKey) -> ExtendedPublicKey {
+        match private_key.get_internal_key() {
+            ExtendedPrivateKeyHolder::Secp256k1Schnorr(ref k) => {
+                Secp256k1ExtendedPublicKey::from_private_key(k).into()
+            }
+        }
+    }
+}
+
+impl From<Secp256k1ExtendedPrivateKey> for ExtendedPrivateKey {
```
TheQuantumPhysicist, line 225: This trait impl shouldn't exist. The struct `Secp256k1ExtendedPrivateKey` is not publicly accessible, and this is only used in new. So to save 20 more characters in the `new` function(s), we're adding a whole trait impl. Not a good trade-off. 

Same applies to `Secp256k1ExtendedPublicKey`.

crypto/src/key/mod.rs
```diff
@@ -229,6 +337,26 @@ mod test {
     #[trace]
     #[case(Seed::from_entropy())]
     fn derive(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 339: derive_secp256k1schnorr

TheQuantumPhysicist (approved):
Looks good and fulfills its purpose. There are many little things that have to be fixed. Please update them and request another review from me. Since this is the first time we're adding extended private keys, I'm gonna be a little nit-picky in this :-)

crypto/src/key/secp256k1/extended_keys.rs
iljakuklic: The second hardened component is usually a coin type in the usual layout. That way you can have a wallet that communicates with multiple networks and is able to generate keys for all from the same seed. Since the Mintlayer wallet is supposed to interface both Mintlayer and Bitcoin, I think we should just use the `"Bitcoin seed"` string and allocate a new number for the second derivation path component. Unless there are other complications I have not considered, that is.

https://learnmeabitcoin.com/technical/derivation-paths#wallet-structure

crypto/src/key/hdkd/derivation_path.rs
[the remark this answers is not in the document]
iljakuklic: Join and append differ in that the RHS can have multiple components whereas the method presented here only adds one component at the end. In that respect, it's more similar to `Vec::push`. If it was mutating the argument in place, it would make sense to just call it `push`.

Since it is internally and conceptually a `Vec`, it makes sense to me for the API to mimic `Vec`. Maybe it is sensible to change this method to use `&mut self` and call it `push`. If you look how it's implemented, there is hidden clone followed by a push anyway. For completenes can also add `FromIterator` to add multiple child numbers in one go and `join` to concatenate two paths together. I think this would make the API largely unsurprising.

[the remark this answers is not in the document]
TheQuantumPhysicist: Structurally, it's a path. But it's not implemented like a path in the API for some reason. If I understand correctly, you're not happy calling this function a join because the API we created doesn't act like a path. Well, I'd call that a problem in the API, not in the idea, since this structure is basically a path in every way. I hope I'm not missing something in this story.

And just to not give the vibe that I'm insistent, I wanna emphasize that I don't care too much about this detail since this is some esoteric thing that'll barely be used in the code. But it'd be nice to have something consistent coming out of it.

crypto/src/key/mod.rs
erasmospunk: I have reverted this have the same index for both extended and normal keys. The rationale is that by having a different "version" number it is clear which key is which.

crypto/src/key/hdkd/derivation_path.rs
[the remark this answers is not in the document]
erasmospunk: I have changed the API by adding a push & pop operation.

crypto/src/key/secp256k1/extended_keys.rs
erasmospunk: The "Bitcoin seed" constant is defined in BIP39 and it is used for most of the coins. Some coins do change it but it is when they use different elliptic curve. I our case we don't need to change it.

crypto/src/key/secp256k1/extended_keys.rs
erasmospunk: Good point, I changed it. I have searched and there are many tests that are `fn test_`

crypto/src/key/secp256k1/extended_keys.rs
erasmospunk: Removed this

crypto/src/key/secp256k1/extended_keys.rs
erasmospunk: I moved the some of the functions to the `to_key_and_chain_code` function. I tried moving out the mut parts to a separate function but it makes the code harder to follow because it breaks in half the BIP32 algorithm.

I simplified the ChildNumber API so we don't have to deal with non-hardened paths. The `0u8` is needed for the private key padding and it is unrelated to the child number.

crypto/src/key/mod.rs
erasmospunk: This was a mistake. I have added a `private_key()` method. Unfortunately I couldn't find a way to make it return a reference.

iljakuklic: > This was a mistake. I have added a `private_key()` method. Unfortunately I couldn't find a way to make it return a reference.

I don't think it's possible to return the final private key as a reference unless you pre-compute the key and store it with in the struct. Otherwise, you get a reference to a local variable which is not allowed. In that case, I'd say just return it by value.

iljakuklic (approved):
The updated version looks good to me

crypto/src/key/mod.rs
TheQuantumPhysicist: While that sounds nice, the problem becomes consistency in the future if we ever add something else. The size should be enough, even though it's not perfect in a long stream.

TheQuantumPhysicist (approved):
All good! Cheers!

PR #602, opened 2022-12-20
Update the env_logger version

stanislav-tkach:
The [issue](https://github.com/mintlayer/mintlayer-core/pull/592) with the `is-terminal` crate is fixed, so the `env_logger` version can be updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,9 +85,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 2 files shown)


TheQuantumPhysicist (approved):
Nice. I'm relieved it's not an issue in lmdb.

iljakuklic (approved):

muursh (approved):

PR #603, opened 2022-12-23
Refactor `MockPeerId` 

pavel-kokolemin:
I noticed some issues with peer ids handling in the mock p2p backend:
- `MockPeerId` is derived from remote peer address and port (using a non-cryptographic hash functions). It's not clear how safe this approach.
- Own peer id is sent in handshake message.

I think it has been done this way because peer ids was supposed to be globally unique. After recent discussion it's decided that we don't really need peers ids to be globally unique and so we could simply things a bit. And so I made some changes in that area:
- `MockPeerId` is removed from the handshake, remote peers won't be able to set arbitrary data there.
- New `MockPeerId` generated from a counter.

This changes allowed remove `local_peer_id` which is not really needed.

We will also need to remove `peer_id` from `AddrInfo`:

```
/// Discovered peer address information
#[derive(Debug, PartialEq, Eq)]
pub struct AddrInfo<T: NetworkingService> {
    /// Unique ID of the peer
    pub peer_id: T::PeerId,

    /// List of discovered IPv4 addresses
    pub ip4: Vec<T::Address>,

    /// List of discovered IPv6 addresses
    pub ip6: Vec<T::Address>,
}
```

I will do that later.

### DIFF
--- a/p2p/src/interface/p2p_interface.rs
+++ b/p2p/src/interface/p2p_interface.rs
@@ -17,7 +17,7 @@
 pub trait P2pInterface: Send + Sync {
     async fn connect(&mut self, addr: String) -> crate::Result<()>;
 
-    async fn disconnect(&self, peer_id: String) -> crate::Result<()>;
+    async fn disconnect(&self, addr: String) -> crate::Result<()>;
 
     async fn get_peer_count(&self) -> crate::Result<usize>;
 
--- a/p2p/src/interface/p2p_interface.rs
+++ b/p2p/src/interface/p2p_interface.rs
@@ -23,7 +23,7 @@ pub struct ConnectedPeer {
 pub trait P2pInterface: Send + Sync {
     async fn connect(&mut self, addr: String) -> crate::Result<()>;
 
-    async fn disconnect(&self, peer_id: String) -> crate::Result<()>;
+    async fn disconnect(&mut self, peer_id: String) -> crate::Result<()>;
 
     async fn get_peer_count(&self) -> crate::Result<usize>;
 
--- a/p2p/src/interface/p2p_interface.rs
+++ b/p2p/src/interface/p2p_interface.rs
@@ -17,7 +17,7 @@
 pub trait P2pInterface: Send + Sync {
     async fn connect(&mut self, addr: String) -> crate::Result<()>;
 
-    async fn disconnect(&self, addr: String) -> crate::Result<()>;
+    async fn disconnect(&self, peer_id: String) -> crate::Result<()>;
 
     async fn get_peer_count(&self) -> crate::Result<usize>;
 
--- a/p2p/src/interface/p2p_interface.rs
+++ b/p2p/src/interface/p2p_interface.rs
@@ -23,7 +23,5 @@ pub trait P2pInterface: Send + Sync {
 
     async fn get_bind_address(&self) -> crate::Result<String>;
 
-    async fn get_peer_id(&self) -> crate::Result<String>;
-
     async fn get_connected_peers(&self) -> crate::Result<Vec<String>>;
 }
--- a/p2p/src/interface/p2p_interface.rs
+++ b/p2p/src/interface/p2p_interface.rs
@@ -13,6 +13,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[derive(Debug, serde::Serialize)]
+pub struct ConnectedPeer {
+    pub peer_id: String,
+    pub addr: String,
+}
+
 #[async_trait::async_trait]
 pub trait P2pInterface: Send + Sync {
     async fn connect(&mut self, addr: String) -> crate::Result<()>;
@@ -23,5 +29,5 @@ pub trait P2pInterface: Send + Sync {
 
     async fn get_bind_address(&self) -> crate::Result<String>;
 
-    async fn get_connected_peers(&self) -> crate::Result<Vec<String>>;
+    async fn get_connected_peers(&self) -> crate::Result<Vec<ConnectedPeer>>;
 }
--- a/p2p/src/interface/types.rs
+++ b/p2p/src/interface/types.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2022 RBB S.r.l
+// Copyright (c) 2023 RBB S.r.l
 // opensource@mintlayer.org
 // SPDX-License-Identifier: MIT
 // Licensed under the MIT License;
@@ -13,17 +13,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::types::ConnectedPeer;
-
-#[async_trait::async_trait]
-pub trait P2pInterface: Send + Sync {
-    async fn connect(&mut self, addr: String) -> crate::Result<()>;
-
-    async fn disconnect(&mut self, peer_id: String) -> crate::Result<()>;
-
-    async fn get_peer_count(&self) -> crate::Result<usize>;
-
-    async fn get_bind_address(&self) -> crate::Result<String>;
-
-    async fn get_connected_peers(&self) -> crate::Result<Vec<ConnectedPeer>>;
+// TODO: Implement From PeerContext for ConnectedPeer
+#[derive(Debug, serde::Serialize)]
+pub struct ConnectedPeer {
+    // TODO: Replace String with actual type, once libp2p removed
+    pub peer_id: String,
+    // TODO: Replace String with actual type, once libp2p removed
+    pub addr: String,
 }
--- a/p2p/src/net/mod.rs
+++ b/p2p/src/net/mod.rs
@@ -114,7 +114,7 @@ where
     ///
     /// # Arguments
     /// `id` - socket address of the peer or peer id
-    async fn disconnect(&mut self, id: DisconnectId<T::Address, T::PeerId>) -> crate::Result<()>;
+    async fn disconnect(&mut self, peer_id: T::PeerId) -> crate::Result<()>;
 
     /// Return the socket address of the network service provider
     ///
@@ -192,10 +192,4 @@ pub trait AsBannableAddress {
 /// Checks if an address can be converted to bannable.
 pub trait IsBannableAddress {
     fn is_bannable(&self) -> bool;
-}
-
-#[derive(Debug, Eq, PartialEq)]
-pub enum DisconnectId<A, P> {
-    Address(A),
-    PeerId(P),
 }
--- a/p2p/src/net/mod.rs
+++ b/p2p/src/net/mod.rs
@@ -113,7 +113,7 @@ where
     /// Disconnect active connection
     ///
     /// # Arguments
-    /// `id` - socket address of the peer or peer id
+    /// `peer_id` - Peer ID of the remote node
     async fn disconnect(&mut self, peer_id: T::PeerId) -> crate::Result<()>;
 
     /// Return the socket address of the network service provider
--- a/p2p/src/net/mod.rs
+++ b/p2p/src/net/mod.rs
@@ -121,9 +121,6 @@ where
     /// If the address isn't available yet, `None` is returned
     async fn local_addr(&self) -> crate::Result<Option<T::Address>>;
 
-    /// Return peer id of the local node
-    fn peer_id(&self) -> &T::PeerId;
-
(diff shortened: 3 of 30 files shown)


pavel-kokolemin:
RPC updated to use and return peer addresses (`disconnect` and `get_connected_peers` calls).

p2p/src/peer_manager/helpers.rs
```diff
@@ -40,21 +40,21 @@ where
...
             event => panic!("expected `InboundAccepted`, got {event:?}"),
         },
         Err(_err) => panic!("did not receive `InboundAccepted` in time"),
     };
 
-    match timeout(Duration::from_secs(5), conn1.poll_next()).await {
+    let (_address2, peer_info2) = match timeout(Duration::from_secs(5), conn1.poll_next()).await {
         Ok(event) => match event.unwrap() {
-            ConnectivityEvent::OutboundAccepted { .. } => {}
+            ConnectivityEvent::OutboundAccepted { address, peer_info } => (address, peer_info),
```
stanislav-tkach, line 53: Nitpick:
```suggestion
            ConnectivityEvent::OutboundAccepted { address: _, peer_info } => peer_info,
```

stanislav-tkach (approved):

test/functional/test_framework/test_framework.py
```diff
@@ -559,24 +557,26 @@ def wait_for_node_exit(self, i, timeout):
         self.nodes[i].process.wait(timeout)
 
     def connect_nodes(self, a, b):
-        id_a = self.nodes[a].p2p_get_peer_id()
-        id_b = self.nodes[b].p2p_get_peer_id()
+        count_a = self.nodes[a].p2p_get_peer_count()
+        count_b = self.nodes[b].p2p_get_peer_count()
```
iljakuklic, line 561: Have we lost the ability to request peer ID via RPC? If so, it looks like a functionality regression. Do you intend to re-add it or provide an alternative way of identifying peers of given node?

pavel-kokolemin: There are two RPC calls: `p2p_get_peer_id` and `get_connected_peers`.
`p2p_get_peer_id` (used to return globally unique local node id) is gone forever because peer ID is now a simple counter. This RPC call become meaningless (at least in the mock backend).
`get_connected_peers` can be updated later to also return this new id, for example to allow disconnect specific peers. Currently it's possible to disconnect peers by their remote socket address.

p2p/src/peer_manager/helpers.rs
pavel-kokolemin: Sure, updated.

p2p/src/rpc.rs
```diff
@@ -65,11 +61,6 @@ impl P2pRpcServer for super::P2pHandle {
         handle_error(res)
     }
 
-    async fn get_peer_id(&self) -> rpc::Result<String> {
```
TheQuantumPhysicist, line 68: We still need the get_peer_id function. The peer id is a unique identifier for the node for operational purposes (like in functional tests), not for consensus purposes.

test/functional/test_framework/test_framework.py
TheQuantumPhysicist: count is not like id. `count` is supposed to be "the number of peers available right now".

TheQuantumPhysicist: This test shouldn't be changed at all. The unique id per connection should be used to identify nodes at the connection level. For some reason we seem to be doing an "all or nothing" thing...

p2p/src/rpc.rs
pavel-kokolemin: I'm not sure it's a good idea to add such RPC call just for operation purposes. Bitcoin core manages without it.
Why RPC socket address or PID is not enough to identify the node you are talking to?

p2p/src/net/mod.rs
```diff
@@ -196,3 +193,9 @@ pub trait AsBannableAddress {
 pub trait IsBannableAddress {
     fn is_bannable(&self) -> bool;
 }
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum DisconnectId<A, P> {
```
TheQuantumPhysicist, line 198: I think this should be done with only peer id... using an address is confusing and is error prone.

test/functional/test_framework/test_framework.py
pavel-kokolemin: > count is not like id. `count` is supposed to be "the number of peers available right now".

Yes, count is the number of connected peers.
We check how many nodes was connected before and wait until we get expected count.
Sorry, I don't understand what is supposed to be wrong here.

> This test shouldn't be changed at all. The unique id per connection should be used to identify nodes at the connection level. For some reason we seem to be doing an "all or nothing" thing...

It's not possible to proceed without updating the test.
The test worked before because it's assumed that the peer ids are globally unique.
It was OK for libp2p where public key from noise was used to identify the nodes, but it was a hack in the mock backed, where peer id was derived from the socket address.
As it was decided we are not going to use globally unique ids and so the tests should not assume that we can somehow guess our node id when we connect to some other peer. `id_a` and `id_b` here was supposed to be such guessed id and I was forced to change it.

TheQuantumPhysicist (commented):
In general I'm OK with the changes, but some things have to be fix. Most notably, we still need the peer id at the RPC level, and another one is that disconnecting peers should be done through their ids, not their String addresses or whatever type address they have. The whole point of a peer id is to be able to identify a connection at any point in time. This is useful for a whole spectrum of purposes. The good thing in this PR is that it removes this information from the remote peer, which is the right thing to do, but we still need the id everywhere else in any operation done on a peer. Everything else is prone to being ambiguous.

pavel-kokolemin:
> In general I'm OK with the changes, but some things have to be fix. Most notably, we still need the peer id at the RPC level, and another one is that disconnecting peers should be done through their ids, not their String addresses or whatever type address they have. The whole point of a peer id is to be able to identify a connection at any point in time. This is useful for a whole spectrum of purposes. The good thing in this PR is that it removes this information from the remote peer, which is the right thing to do, but we still need the id everywhere else in any operation done on a peer. Everything else is prone to being ambiguous.

OK, thanks, let me change that.

p2p/src/rpc.rs
TheQuantumPhysicist: That's actually not true. The peer id is used in bitcoin as a standard way to identify connections. See here as an example: 

https://github.com/bitcoin/bitcoin/blob/d8bdee0fc889def7c5f5076da13db4fce0a3728a/src/rpc/net.cpp#L406

Because we need a deterministic way to identify connections by peer info in scripting and similar.

test/functional/test_framework/test_framework.py
TheQuantumPhysicist: I understand what you mean and I see where this misunderstanding is coming from. The test is using ids, which is not a good handle. There should not be usage of ids in the test, but the object of the node should be used. Then, to communicate with the node specifically, the id should be used in the context of one connection.

So let's say NodeA, for testing purposes, needs to send a specific message to NodeB. This is how it should work in functional tests:

- You call `node_b_id = NodeB.get_peer_id()`
- You call `NodeA.send_my_special_message(node_b_id)`

This way, you shouldn't care about the address. Dealing with addresses is messy at the test level and is its own abstraction layer. The only practical purpose of addresses, besides the obvious communication, is banning addresses and associating behavior. But we shouldn't be throwing strings all over the place in our interfaces. That's playing with fire.

p2p/src/rpc.rs
pavel-kokolemin: Right, I'm adding it now.
But I was referring to `get_peer_id` RPC call specifically, which is supposed to return some kind of "my node id", which is become meaningless now.

test/functional/test_framework/test_framework.py
pavel-kokolemin: Sorry, I still don't see how it's relevant to my changes. I removed `get_peer_id` from RPC because after recent discussion it  was decided that we are not going to have global ids. Without global ids `get_peer_id` becomes meaningless (what can we return there?). So I removed it and I updated the functional tests that depend on it (`connect_nodes` and `disconnect_nodes` helper functions). That functions also assumed that the addresses are from libp2p which I also fixed.

p2p/src/net/mod.rs
pavel-kokolemin: OK, removed. I still think that disconnecting by socket address is nice to have. Bitcoin Core has it.

TheQuantumPhysicist: It's OK to have that in RPC. Then it'll do a reverse-search in peers and disconnect by id. You can even disconnect a subnet. But it shouldn't be the internal implementation by string.

p2p/src/interface/p2p_interface.rs
```diff
@@ -13,6 +13,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[derive(Debug, serde::Serialize)]
+pub struct ConnectedPeer {
+    pub peer_id: String,
+    pub addr: String,
+}
```
iljakuklic, line 20: Is it possible to have more descriptive types that serialize to / deserialize from a `String`?

iljakuklic (commented):
Looks fine to me but have not looked too thoroughly. Added some very minor comments

p2p/src/interface/p2p_interface.rs
```diff
@@ -13,6 +13,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[derive(Debug, serde::Serialize)]
+pub struct ConnectedPeer {
```
TheQuantumPhysicist, line 17: Why not just return PeerInfo, or some mapping of it?

p2p/src/rpc.rs
```diff
@@ -34,13 +34,9 @@ trait P2pRpc {
     #[method(name = "get_bind_address")]
     async fn get_bind_address(&self) -> rpc::Result<String>;
 
-    /// Get peer ID of the local node
-    #[method(name = "get_peer_id")]
-    async fn get_peer_id(&self) -> rpc::Result<String>;
-
-    /// Get peer IDs of connected peers
+    /// Get deatils of connected peers
```
TheQuantumPhysicist, line 37: typo

TheQuantumPhysicist (approved):

p2p/src/rpc.rs
pavel-kokolemin: Fixed, thanks.

p2p/src/interface/p2p_interface.rs
pavel-kokolemin: Nice idea, thanks. But let's please wait until we replace libp2p. Peer IDs and addresses have different types in libp2p and mock. Once removed we should be able to use `MockPeerId` and some type for peer address directly.
I added TODO.

p2p/src/sync/tests/mod.rs
```diff
@@ -127,6 +127,6 @@ impl MakeTestPeerId for MockPeerId {
     type PeerId = Self;
 
     fn random() -> Self::PeerId {
-        MockPeerId::random()
+        MockPeerId::new()
```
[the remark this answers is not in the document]
pavel-kokolemin: Renamed to `new`.

p2p/src/interface/types.rs
```diff
@@ -0,0 +1,23 @@
...
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO: Implement From PeerContext for ConnectedPeer
```
TheQuantumPhysicist, line 16: I think it's PeerInfo we're interested in, not PeerContext

TheQuantumPhysicist (approved):

p2p/src/interface/types.rs
pavel-kokolemin: `p2p::peer_manager::peerdb::PeerContext` contains some interesting stuff (address and ban score for example). We probably can move some of it to `PeerInfo`.

iljakuklic (approved):

PR #604, opened 2022-12-27
Basic peer discovery

pavel-kokolemin:
Changes:

**Allow listening on multiple addresses.**
The node was listening on only one IPv6 socket by default and so it was not clear how incoming connections from IPv4 peers should be handled.
There are two ways to fix this:

1. Unset `IPV6_V6ONLY` socket flag and bind to default IPv6 address ([::]). Incoming IPv4 connections will be reported as IPv6 (like `::ffff:192.168.1.2`).
2. Set `IPV6_V6ONLY` and bind to two sockets (unless user specified something different).

I decided to go with the second option because it appears to be more robust and future-proof (bitcoin core and libp2p-rust also work this way). Users will be able to specify any number of bind addresses.
We need to set `IPV6_V6ONLY` explicitly in any case because its default value differs between platforms. For this I used `socket2` crate which is de-factor standard (and we already have it in the dependency tree).

**Add `PeerAddress` type**
I needed some concrete type which we can send over network and to work with (take IP address and apply port change).
`TransportAddress` trait is added to be able covert from abstract transport address type to `PeerAddress` and back.
`PeerAddress` might be extended later to also include addresses from other transports (for example TOR onion address).

**Add `--p2p-add-node` command line parameter**
It's used to manually specify node addresses to which node will try to connect, which is useful in tests. Works like `-addnode=` in `bitcoind`.

**Send listening port**
Mock backend might send its listining port in outbound connection handshake.
If listining port is set, inbound nodes will take IP address of the remote socket and apply this listening socket.
This new socket address will be handled as newly discovered peer address.
This approach will solve the problem with discovering IP address for nodes with public IP addresses and for nodes behind NAT with port forwarding.

**Allow only one connection per remote socket address**
I noticed in tests that we do not prevent duplicated connections to same socket address.
That was not problem with libp2p, because only one connection is allowed to same peer id.
But this is a problem with mock backend where we do not have globally unique peer ids.
To fix this I updated PeerManager to allow only one connection to same socket address.
Still peers on mock backend might have two connections to each others:

- A connects to B.
- B discovers listening address of A.
- B connects to newly discovered address of A.

We can fix this later allowing only 1 connection per IP address.

**PeerDb updated**
`PeerDb` was updated to not rely on globally unique peer ids.
Also `PeerDb` unit tests moved to submodule of peerdb mod. This allowed make some internal functions private.


common/src/primitives/net.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2022 RBB S.r.l
```
TheQuantumPhysicist, line 1: I would call this file ip.rs

TheQuantumPhysicist: Or ip_address.rs

common/src/primitives/net.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn decode<I: serialization::Input>(input: &mut I) -> Result<Self, serialization::Error> {
+        let mut octets = [0u8; IPV6_SIZE];
+        input.read(&mut octets)?;
+        Ok(Self {
+            inner: octets.into(),
+        })
+    }
+}
+
+impl From<std::net::Ipv4Addr> for Ip4 {
```
TheQuantumPhysicist, line 89: You probably also need the conversion the other way. It's 1:1.

node/src/config_files/mod.rs
```diff
@@ -137,6 +137,7 @@ fn chainstate_config(
 fn p2p_config(config: P2pConfigFile, options: &RunOptions) -> P2pConfigFile {
     let P2pConfigFile {
         bind_address,
+        add_node,
```
TheQuantumPhysicist, line 140: There's an issue that has been discussed earlier but has been omitted for a while and we have to come back to it sooner or later. The issue is that not all peers are equal. There are peers that we want to persistently have, and others that we want to just give a "shot" at connecting at. I think in Substrate they call the former "reserved" peers or something. We should also consider doing that. This is important because some times you have peers with whom you want to maintain a connection no matter what. I don't know if you're gonna do that in this PR, but please put that into consideration, whether in an issue or this PR. Please research this topic when you can.

p2p/src/interface/types.rs
```diff
@@ -13,11 +13,15 @@
...
     // TODO: Replace String with actual type, once libp2p removed
     pub peer_id: String,
+
     // TODO: Replace String with actual type, once libp2p removed
-    pub addr: String,
+    pub address: String,
+
+    pub inbound: bool,
+
+    pub score: u32,
```
TheQuantumPhysicist, line 26: ban_score

common/src/primitives/net.rs
pavel-kokolemin: Sure, updated.

common/src/primitives/net.rs
pavel-kokolemin: Done.

node/src/config_files/mod.rs
pavel-kokolemin: Right, we will need this too. But let's please do it later.

p2p/src/interface/types.rs
pavel-kokolemin: Done, thanks.

common/src/primitives/ip_address.rs
```diff
@@ -0,0 +1,111 @@
...
+/// IPv4 address.
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
+pub struct Ip4 {
+    pub inner: std::net::Ipv4Addr,
+}
+
+/// IPv6 address.
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
+pub struct Ip6 {
+    pub inner: std::net::Ipv6Addr,
```
TheQuantumPhysicist, line 30: This doesn't need to be public anymore. Now you can use From/Into. In addition, you can implement AsRef to access the internal values.

common/src/primitives/ip_address.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use serialization::{Decode, Encode};
+
+const IPV4_SIZE: usize = 4;
+const IPV6_SIZE: usize = 16;
+
+/// IPv4 address.
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq, Hash)]
+pub struct Ip4 {
+    pub inner: std::net::Ipv4Addr,
```
TheQuantumPhysicist, line 24: Same applies to this

node/src/config_files/mod.rs
TheQuantumPhysicist: Please open an issue for it then resolve this comment.

p2p/src/net/mock/transport/traits/address.rs
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::types::peer_address::PeerAddress;
+
+pub trait TransportAddress: Sized {
```
TheQuantumPhysicist, line 18: Please specify above this what problem it solves.

I wonder whether this is any different from BannableAddress.

p2p/src/net/mock/types.rs
```diff
@@ -199,13 +205,20 @@ pub enum MockEvent {
...
     Hello {
         version: common::primitives::semver::SemVer,
         network: [u8; 4],
         protocols: Vec<Protocol>,
         subscriptions: BTreeSet<PubSubTopic>,
+
+        // Specify on which port this node can be reached for incoming connections.
+        // If not set, then this node does not want incoming connections.
+        // Actual IP is taken from the remote socket address.
+        listening_port: Option<u16>,
```
TheQuantumPhysicist, line 221: This is a weird thing... why? When someone connects, they advertise their socket address through their connection... I kind of don't understand this.

pavel-kokolemin: When you connect with TCP OS chooses some random port for you. That is because listening socket and connecting sockets are independent from the OS viewpoint.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    pub info: types::PeerInfo<T>,
+
+    /// Peer's address
+    pub address: T::Address,
+
+    /// Peer's role (inbound or outbound)
+    pub role: Role,
+
+    /// Peer score
+    pub score: u32,
```
TheQuantumPhysicist, line 58: ban_score...

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+impl<T: NetworkingService> From<&PeerContext<T>> for ConnectedPeer {
```
TheQuantumPhysicist, line 61: Why convert from ref?

TheQuantumPhysicist: And then clone strings?

It's always preferable to delegate the clone decision to the developer using the feature.

common/src/primitives/ip_address.rs
pavel-kokolemin: I replaced with fixed array and removed manual encode/decode implementation. Public removed. I think AsRef is not very useful now so not added.

common/src/primitives/ip_address.rs
pavel-kokolemin: Same done

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Right, ban_score is better, thanks.

p2p/src/peer_manager/helpers.rs
```diff
@@ -61,3 +60,27 @@ where
 
     (address, peer_info1, peer_info2)
 }
+
+/// Return first event that is accepted by predicate.
+///
+/// Used to skip events that are not of interest or that are different beween backends
```
TheQuantumPhysicist, line 66: between

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: > Why convert from ref?

Because we convert from ref here:

```
        self.peers.values().map(Into::into).collect()
```

Converting from value will require implement `Clone` for `PeerContext`.
And I don't think it's a good idea:

- `PeerContext` probably should not be cloneable for safety.
- Cloning `PeerContext` will make a clones of some non-trivial types (two `BTreeSet` for example) that will be dropped once we convert to `ConnectedPeer`.
- We might want to add some other types to `PeerContext` someday that do not implement `Clone`.

So I think converting from ref is better.

p2p/src/peer_manager/helpers.rs
pavel-kokolemin: Fixed, thanks.

p2p/src/net/mock/types.rs
TheQuantumPhysicist: Still. That isn't really the way this should work. The way I see it (and correct me if I'm wrong here), a node connects to another node, and broadcasts the addresses it finds working, including its own. Ports should never be expected to be different.

This even ignores NAT issues. Behind NAT subnets, no one can know this value because the router decides it and even the value the node thinks it has is firewalled, which basically means our software should be able to detect whether a node is in a private network... why bother?

How is this even gonna be verified? Are we gonna create a second hand-shake process to ensure that the sender isn't lying or something?

I think this should go, and we should just expect ports to be part of the broadcast addresses at worst.

p2p/src/peer_manager/peerdb/mod.rs
TheQuantumPhysicist: I'm not really convinced, but I'm gonna let this go for now because it doesn't really matter and maybe I just don't have the full picture.

p2p/src/net/mock/types.rs
pavel-kokolemin: Sure, different ways are possible.
The main question is how the node will find own IP address.
1. The node might try detect own IP addresses using OS specific API (I think that is what you propose).
2. Remote peers can read listening port for all incoming connections and get peer address directly (what is implemented here).

Problems with the first approach:
- Need to use OS specific API, IP addresses might change with time so that should be done periodically.
- Does not work with NAT and port forwarding (when forwarded port same).

The second approach does not have this cons and is easier and so I decided to use it.
Grin works this way. In bitcoin core public IP also loaded from remote peers, but handled locally instead.

> This even ignores NAT issues.

I don't see a problem here. And this is done specifically to handle NAT with port forwarding, because it's difficult to find out public IP address.

> which basically means our software should be able to detect whether a node is in a private network... why bother?

We don't need to detect if we are in a private network. Remote peer will only see public IP address.

> How is this even gonna be verified? Are we gonna create a second hand-shake process to ensure that the sender isn't lying or something?

This new address will be verified same way like any other discovered address. The node might not be reachable but that is not a problem.

TheQuantumPhysicist: Maybe there's a misunderstanding, but I have mentioned many times that we're not gonna do anything to circumvent NAT. No hole-punching. Nothing. If someone is interested in using DMZ to open their network to the mintlayer network, they should do it. We shouldn't attempt to do anything for NAT.

You're saying "We don't need to detect if we are in a private network. Remote peer will only see public IP address"... it might be a misunderstanding also on how NAT works. Even if we see a public IP address, this doesn't mean that this IP address will work, because again, it's behind a firewall and a NAT mapping should be created beforehand to make that work. So the question is, if someone is going to configure their router to do that mapping through DMZ or some other routing/mapping, what use does this port value serve exactly? We aren't expecting people to use arbitrary port numbers. Bitcoin doesn't do that either.

And just to give you an example from real life, apart from bitcoin which does exactly that (please correct me if I'm wrong here), the SMTP protocol for MTA (Mail Transfer Agents) works on port 25. No one expects other agents to use other ports. No one broadcasts different ports. No one tries to circumvent NAT to reach these ports. It's the administrator's job to maintain that promise. Same for http, https, etc.

You're saying we need OS-specific APIs. How do other blockchains deal with this? How does substrate deal with this? Grin? There are many other blockchains in rust. Check them out, please.

TheQuantumPhysicist: One point I wanna add here is that the best we can do here is make peers broadcast socket addresses instead of just addresses. This way people can announce hosts with different ports, but it's not the default behavior. I someone wants to break that promise, they have to go through the work of broadcasting that address. Now the automation of this process is a whole other question. Again, let's see how other blockchains do this.

pavel-kokolemin: I looked at Bitcoin, Monero and Grin (I skipped Bitcoin forks and blockchains that use libp2p).
In all three, public node addresses might be constructed using socket addresses, as seen by remote peers and local listening ports.

Bitcoin:

```
if (!pfrom.IsInboundConn() && SetupAddressRelay(pfrom, *peer)) {
    if (fListen && !m_chainman.ActiveChainstate().IsInitialBlockDownload())
    {
        CAddress addr{GetLocalAddress(pfrom.addr), peer->m_our_services, Now<NodeSeconds>()};
        if (addr.IsRoutable())
        {
            LogPrint(BCLog::NET, "ProcessMessages: advertising address %s\n", addr.ToString());
            PushAddress(*peer, addr, insecure_rand);
        } else if (IsPeerAddrLocalGood(&pfrom)) {
            addr.SetIP(addrMe);
            LogPrint(BCLog::NET, "ProcessMessages: advertising address %s\n", addr.ToString());
            PushAddress(*peer, addr, insecure_rand);
        }
    }
}
```

So if a node behind NAT and made an outgoing connection, `GetLocalAddress` will return local listining address and the node will change IP address to public socket address and use that to advertise as its own address.

Grin:

```
/// Resolve the correct peer_addr based on the connection and the advertised port.
fn resolve_peer_addr(advertised: PeerAddr, conn: &TcpStream) -> PeerAddr {
    let port = advertised.0.port();
    if let Ok(addr) = conn.peer_addr() {
        PeerAddr(SocketAddr::new(addr.ip(), port))
    } else {
        advertised
    }
}

pub fn accept(
    &self,
    capab: Capabilities,
    total_difficulty: Difficulty,
    conn: &mut TcpStream,
) -> Result<PeerInfo, Error> {
...
    // all good, keep peer info
    let peer_info = PeerInfo {
        ...
        addr: resolve_peer_addr(hand.sender_addr, &conn),
    };
```

When a new peer is connected, IP address of the socket and advertised port is used to construct a public address of this peer.

Monero:

```
  template<class t_payload_net_handler>
  int node_server<t_payload_net_handler>::handle_handshake(int command, typename COMMAND_HANDSHAKE::request& arg, typename COMMAND_HANDSHAKE::response& rsp, p2p_connection_context& context)
  {
    if(!context.m_is_income)
    {
      return 1;
    }

    if(arg.node_data.my_port && zone.m_can_pingback)
    {
      uint32_t port_l = arg.node_data.my_port;
      //try ping to be sure that we can add this peer to peer_list
      try_ping(arg.node_data, context, [peer_id_l, port_l, context, this]()
      {
        //called only(!) if success pinged, update local peerlist
...
  }

  bool node_server<t_payload_net_handler>::get_local_node_data(basic_node_data& node_data, const network_zone& zone)
  {
    if(!m_hide_my_port && zone.m_can_pingback)
      node_data.my_port = m_external_port ? m_external_port : m_listening_port;
    else
      node_data.my_port = 0;
  }

```

For incoming connections `try_ping` is called, to check if a remote node is also accepting incoming connections. `my_port` is set by the remote peer itself and IP is taken from the remote socket address.

So in all three blockchains the same mechanism is used, in one way or another:
When a node is accepting an incoming connection, it takes IP address from remote socket and advertised listening port. This new socket address is then used as a new potential address.
This approach is universal enough to work with different setups, there is no need to load interface addresses manually (although bitcoin loads them).
And that is what I have implemented here.

BTW, I found two popular crates that can be used to find local network interface addresses (`if-addrs` and `if-watch`). I missed them before.

pavel-kokolemin:
There was a problem with functional test in `p2p_syncing_test.py`.
After `disconnect_nodes` is called, remote `PeerManager` will try to establish connection to newly discovered peer address:
1. A connects to B.
2. B discovers listening address of A.
3. A disconnects from B.
4. B connects to A (to discovered address).

And the test fails because the nodes are expected to be disconnected.

I "fixed" the problem by limiting rate of heartbeats in `PeerManager` (which was already marked as TODO).
So when A disconnects from B it now takes some time before B tries to connect to A, which is enough for test to pass.
That is probably not a proper fix but I don't see how it should be fixed.

pavel-kokolemin:
After discussion with @TheQuantumPhysicist I replaced `listening_port` with `receiver_address`, like it's done in Bitcoin Core. This way peers will be able find out their own IP addresses and advertise those to other peers.
But let's please do it in other PRs, this branch is already too big and in development for too long.

TheQuantumPhysicist (approved):

node/src/config_files/p2p.rs
```diff
@@ -122,7 +122,9 @@ impl FromStr for NodeTypeConfigFile {
 #[derive(Serialize, Deserialize, Debug, Default)]
 pub struct P2pConfigFile {
     /// Address to bind P2P to.
-    pub bind_address: Option<String>,
+    pub bind_address: Option<Vec<String>>,
+    /// Optional list of initial node addresses to connect.
+    pub add_node: Option<Vec<String>>,
```
stanislav-tkach, line 127: Nitpick: Should the plural form (`add_nodes` instead of `add_node`) be use?

p2p/src/peer_manager/helpers.rs
```diff
@@ -61,3 +60,27 @@ where
 
     (address, peer_info1, peer_info2)
 }
+
+/// Return first event that is accepted by predicate.
+///
+/// Used to skip events that are not of interest or that are different between backends
+/// (for example ConnectivityEvent::Discovered).
+/// Can be used in tests only, will panic in case of errors.
+pub async fn filter_connectivity_event<T, F>(
```
stanislav-tkach, line 69: Can we move this function or even the whole `peer_manager/helpers.rs` module to `testing_utils.rs`? Can we at least rename the helpers module to "test helpers" or something like that.

p2p/src/interface/p2p_interface.rs
```diff
@@ -23,7 +23,7 @@ pub trait P2pInterface: Send + Sync {
 
     async fn get_peer_count(&self) -> crate::Result<usize>;
 
-    async fn get_bind_address(&self) -> crate::Result<String>;
+    async fn get_bind_address(&self) -> crate::Result<Vec<String>>;
```
stanislav-tkach, line 26: Nitpick:
```suggestion
    async fn get_bind_addresses(&self) -> crate::Result<Vec<String>>;
```
?

common/src/primitives/ip_address.rs
```diff
@@ -0,0 +1,52 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+/// IPv4 address
+#[derive(Debug, Copy, Clone, PartialOrd, Ord, PartialEq, Eq, Hash, Decode, Encode)]
```
stanislav-tkach, line 19: Am I right that we need these types to implement `Encode` and `Decode` for them? Would it make sense to add a documentation note explaining why we need almost exact copies of the std-types?..

p2p/src/net/mock/transport/impls/channel.rs
```diff
@@ -34,13 +34,31 @@ use tokio::{
...
 
-type Address = u64;
+type Address = u32;
+
+impl TransportAddress for Address {
+    fn as_peer_address(&self) -> PeerAddress {
+        PeerAddress::Ip4(PeerAddressIp4 {
+            // Address of the first "host" will be 0.0.0.1
+            ip: std::net::Ipv4Addr::from(*self).into(),
+            // There is only "port" in MockChannelTransport per "host", use arbitrary value
```
stanislav-tkach, line 53:
```suggestion
            // There is only one "port" in MockChannelTransport per "host", use arbitrary value
```

stanislav-tkach (approved):

p2p/src/peer_manager/helpers.rs
pavel-kokolemin: Good idea. Moved out all to `testing_utils.rs`.

p2p/src/interface/p2p_interface.rs
pavel-kokolemin: Sure, updated, thanks.

common/src/primitives/ip_address.rs
pavel-kokolemin: Yes, correct, to be able implement `Encode` and `Decode`. Documentation added, thanks.

p2p/src/net/mock/transport/impls/channel.rs
pavel-kokolemin: Updated, thanks.

Issue #596, opened 2022-12-13
Fix the block sync manager logic


stanislav-tkach:
Related to https://github.com/mintlayer/mintlayer-core/issues/583.

PR #605, opened 2022-12-29
Fix the sync manager "initial block download" logic

### LINKED ISSUE #583: Rework or remove timeouts in p2p
Currently we have per request timeouts solely because of the `BlockSyncManager` (incorrect) logic. Instead we should use timeouts per peer, not per request. Perhaps timeouts can be completely removed because we have heartbeats, so unresponsive peers should be disconnected anyway.

stanislav-tkach:
- Requests timeouts have been removed.
- The `max_tip_age` parameter has been added.
- The initial block download is considered to be done when the best block timestamp isn't older than the current time minus `max_tip_age`.

Closes #583 and #596.


chainstate/src/detail/query.rs
```diff
@@ -98,6 +98,22 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks, V: TransactionVerificationSt
...
 
+    pub fn get_best_block_header(&self) -> Result<Option<BlockHeader>, PropertyQueryError> {
+        let best_block_index = self
+            .chainstate_ref
+            .get_best_block_index()?
+            .ok_or(PropertyQueryError::BestBlockIndexNotFound)?;
+        self.chainstate_ref.get_header_from_height(&best_block_index.block_height())
+    }
+
+    pub fn get_best_block_timestamp(&self) -> Result<BlockTimestamp, PropertyQueryError> {
```
TheQuantumPhysicist, line 109: Why not just use the header for this?

chainstate/src/detail/query.rs
```diff
@@ -98,6 +98,22 @@ impl<'a, S: BlockchainStorageRead, O: OrphanBlocks, V: TransactionVerificationSt
         self.chainstate_ref.get_best_block_index()
     }
 
+    pub fn get_best_block_header(&self) -> Result<Option<BlockHeader>, PropertyQueryError> {
+        let best_block_index = self
+            .chainstate_ref
+            .get_best_block_index()?
+            .ok_or(PropertyQueryError::BestBlockIndexNotFound)?;
+        self.chainstate_ref.get_header_from_height(&best_block_index.block_height())
```
TheQuantumPhysicist, line 106: The block index does contain the block header. No need for another query.

node/src/config_files/p2p.rs
```diff
@@ -131,10 +131,13 @@ pub struct P2pConfigFile {
...
     pub mdns_config: Option<MdnsConfigFile>,
-    /// The request timeout value in seconds.
-    pub request_timeout: Option<u64>,
     /// A node type.
     pub node_type: Option<NodeTypeConfigFile>,
+    /// A maximum tip age in seconds.
+    ///
+    /// The initial block download is finished if the difference between the current time and the
+    /// tip time is less than this value.
+    pub max_tip_age: Option<u64>,
```
TheQuantumPhysicist, line 140: I don't think this should be parametrizable. It should be a constant.

chainstate/src/detail/mod.rs
```diff
@@ -83,6 +86,7 @@ pub struct Chainstate<S, V> {
     custom_orphan_error_hook: Option<Arc<OrphanErrorHandler>>,
     events_controller: EventsController<ChainstateEvent>,
     time_getter: TimeGetter,
+    is_initial_block_download_finished: AtomicBool,
```
TheQuantumPhysicist, line 89: I don't think this needs to be atomic.

chainstate/test-suite/src/tests/syncing_tests.rs
```diff
@@ -333,3 +333,24 @@ fn filter_already_existing_blocks_detached_headers(#[case] seed: Seed) {
...
+#[case(Seed::from_entropy())]
+fn initial_block_download(#[case] seed: Seed) {
+    utils::concurrency::model(move || {
+        let mut rng = make_seedable_rng(seed);
+        let mut tf = TestFramework::builder(&mut rng).build();
+
+        // Only genesis block, so is_initial_block_download should return true.
+        assert!(tf.chainstate.is_initial_block_download().unwrap());
+
+        // The added block will have a very fresh timestamp.
```
TheQuantumPhysicist, line 348: I think that by default the TestFramework uses simulated time. Please use simulated and controlled time with controlled current-time to simulate the cases true/false for initial block download.

test/functional/p2p_syncing_test.py
```diff
@@ -13,7 +13,7 @@ class ExampleTest(BitcoinTestFramework):
     def set_test_params(self):
         self.setup_clean_chain = True
         self.num_nodes = 2
-        self.extra_args = [[], []]
+        self.extra_args = [["--max-tip-age=63070000"], ["--max-tip-age=63070000"]]
```
TheQuantumPhysicist, line 16: Why does this have to be manually put here instead of using the default value? This doesn't look right.

TheQuantumPhysicist (approved):
Please fix the issues I noted then feel free to merge.

iljakuklic (approved):

chainstate/src/detail/mod.rs
stanislav-tkach: It is needed for "interior mutability", so this flag can be changed in the non-mut function. Do you think it would be better to change the function signature to `is_initial_block_download(&mut self)`?

test/functional/p2p_syncing_test.py
stanislav-tkach: This is needed because w use the pre-made blocks (by the `generate_blocks_for_functional_tests` test), so all timestamps are old and the initial block download is never finished with the default max tip age value.

chainstate/src/detail/mod.rs
iljakuklic: I am not sure whether we should go for `&mut self` or not but if we go down the interior mutability route, I'd prefer it to be `Cell<bool>`. It communicates the intent better. An atomic type kind of suggests you access the field from multiple threads.

TheQuantumPhysicist: No, we shouldn't. We don't have to copy bitcoin. We can just update this value after every new block.

test/functional/p2p_syncing_test.py
TheQuantumPhysicist: Please put a TODO on this to remove it later.

PR #608, opened 2023-01-01
Remove reference flagged by Clippy

alfiedotwtf:


alfiedotwtf:
Again, wrong commits...

PR #609, opened 2023-01-01
Minor documentation fixes

alfiedotwtf:


alfiedotwtf:
I didn't see the `staging` branch in the `git branch -a` list, hence the change, then while I was there decided to add `cmake`.

muursh:
Ah it's the old way the branches used to work, we plan to revert back to the old staging branch system at some point soon so I'm tempted to leave the docs as outdated as they represent a future goal

alfiedotwtf:
Ok cool, sounds good.

stanislav-tkach:
I have made the same mistake before: https://github.com/mintlayer/mintlayer-core/pull/221. 🙃

PR #611, opened 2023-01-01
Add pip requirements.txt for people to run locally

alfiedotwtf:


.github/scripts/requirements.txt
```diff
@@ -0,0 +1 @@
+toml==0.10.2
```
iljakuklic, line 1: There is a `pip install toml` command in `.github/workflows/code_checks.yml`. Should it be pointed to this file or will it be picked up automatically somehow?

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
Whatever you guys arrive at I'm OK with. Just to be clear what my approval means.

.github/scripts/requirements.txt
alfiedotwtf: To get the best of both worlds, I've changed it so that Github Actions will read this new `requirements.txt` file

alfiedotwtf:
This PR looks stuck, even after the push as a direct fork. I'll open a new PR.

PR #613, opened 2023-01-03
Inverse semigroup approach for deltas

azarovh:
This is an attempt to implement what Lukas suggested in the comment to my previous PR:
https://github.com/mintlayer/mintlayer-core/pull/586#issuecomment-1367360036

Imho it looks and works better than the previous version: all tests are passing and the result of combining deltas is cleaner and less debatable.


accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -19,29 +19,28 @@ use self::undo::*;
...
-    Delete(T),
+    // TODO: better error reporting?
+    Mismatch,
+    Modify(Option<T>, Option<T>),
+}
+
+impl<T: Clone> DataDelta<T> {
+    /// Returns an invert delta that has the opposite effect of the provided delta
+    /// and serves as an undo object
+    fn invert(&self) -> DataDeltaUndo<T> {
```
TheQuantumPhysicist, line 36: invert should take self as argument to avoid unnecessary cloning. The caller should decide whether cloning is desirable.

accounting/src/delta/tests.rs
```diff
@@ -101,22 +107,22 @@ fn data_delta_delta(
...
+#[case(None,      DataDelta::new(None, Some('a')),      DataDelta::new(Some('a'), Some('b')),  DataDelta::new(Some('b'), None),      None)]
+#[case(None,      DataDelta::new(None, Some('a')),      DataDelta::new(Some('a'), Some('b')),  DataDelta::new(Some('b'), Some('c')), Some('c'))]
+#[case(None,      DataDelta::new(None, Some('a')),      DataDelta::new(Some('a'), None),       DataDelta::new(None, Some('a')),      Some('a'))]
+#[case(Some('a'), DataDelta::new(Some('a'), None),      DataDelta::new(None, Some('a')),       DataDelta::new(Some('a'), None),      None)]
+#[case(Some('a'), DataDelta::new(Some('a'), None),      DataDelta::new(None, Some('a')),       DataDelta::new(Some('a'), Some('b')), Some('b'))]
+#[case(Some('a'), DataDelta::new(Some('a'), Some('b')), DataDelta::new(Some('b'), None),       DataDelta::new(None, Some('a')),      Some('a'))]
+#[case(Some('a'), DataDelta::new(Some('a'), Some('b')), DataDelta::new(Some('b'), Some('c')),  DataDelta::new(Some('c'), None),      None)]
+#[case(Some('a'), DataDelta::new(Some('a'), Some('b')), DataDelta::new(Some('b'), Some('c')),  DataDelta::new(Some('c'), Some('d')), Some('d'))]
```
iljakuklic, line 117: You can add some helper constructors to streamline this a bit:

* `DataDelta::modify('a', 'b')` => `DataDelta::new(Some('a'), Some('b'))`
* `DataDelta::create('c')` => `DataDelta::new(None, Some('b'))`
* etc...

Maybe it should be called `new_modify` instead of just `modify` (similar with the rest), not a strong opinion on that either way

iljakuklic (commented):
Added quite a few comments. Not sure what parts were up for review and which are still subject to change so commented everything :slightly_smiling_face:. Nice change overall and nice to see the complexity come down a bit.

azarovh:
> Not sure what parts were up for review and which are still subject to change so commented everything

I started this branch as an experiment based on the previous branch. My initial idea was that we want to quickly decide whether we want this approach or not. But now it seems that it's better to change the base to master and drop previous PR (applying all comments from there)

accounting/src/delta/delta_data_collection/mod.rs
```diff
@@ -14,52 +14,133 @@
...
+///   (Delta(Some, None) + Delta(Some, None)) + Delta(None, Some) = Error + Delta(None, Some)
+///   Delta(Some, None) + (Delta(Some, None) + Delta(None, Some)) = Delta(Some, None) + Delta(Some, Some) = Error
+///
+/// - Inversion in the sense that:
+///   (d + d.invert()) + d = d
+///
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug)]
+pub struct DataDelta<T> {
+    prev: Option<T>,
+    next: Option<T>,
```
TheQuantumPhysicist, line 42: Naming these `prev` and `next` is kind of confusing... I suggest something `cur` and `new`. Or `from` and `to`. The labels prev/next usually indicate iteration.

accounting/src/delta/delta_data_collection/undo.rs
```diff
@@ -15,30 +15,29 @@
...
-/// The operations we have to do in order to undo a delta
-pub(super) enum DataDeltaUndoOpInternal<T> {
-    Write(DataDelta<T>),
-    Erase,
-}
+use super::DataDelta;
 
-pub struct DataDeltaUndoOp<T>(pub(super) DataDeltaUndoOpInternal<T>);
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug)]
+pub struct DataDeltaUndo<T>(pub DataDelta<T>);
```
TheQuantumPhysicist, line 23: Instead of having `pub` inside, create a function `as_delta` (which takes a reference), and also implement `From<DataDeltaUndo> for DataDelta`. I hate having to write `.0` or `.1`... nothing could look uglier and more confusing.

pos_accounting/src/pool/storage/mod.rs
```diff
@@ -184,7 +184,7 @@ impl<'a, S: PoSAccountingStorageWrite> PoSAccountingDBMut<'a, S> {
         })
     }
 
-    fn merge_data_generic<K: Ord + Copy, T: Clone, Getter, Setter, Deleter>(
+    fn merge_data_generic<K: Ord + Copy, T: Clone + PartialEq, Getter, Setter, Deleter>(
```
iljakuklic, line 187: Not sure `PartialEq` is completely well-behaved here as opposed to `Eq`. Have you considered the implications of it?

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -15,308 +15,123 @@
...
-        ]
-        .into_iter(),
-    );
-    collection1.undo_merge_delta_data(undo_data).unwrap();
-    assert_eq!(collection1.data, expected_data_after_undo);
+#[rstest]
+#[case(DataDelta::new(None, None))]
+#[case(DataDelta::new(None, Some('a')))]
+#[case(DataDelta::new(Some('a'), None))]
+#[case(DataDelta::new(Some('a'), Some('b')))]
```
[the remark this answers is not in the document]
iljakuklic: > I'm thinking whether we should just convert `(Some('x'), Some('x'))` to some kind of Noop. The reason is specifically to save memory and kind of an optimization. Does that make sense?

`(Some(x), Some(x))` is not equivalent to a noop. It also checks the value is `'x'` and gives an error otherwise. The proposed optimisation would violate the property that applying deltas one after another gives the same result as combining deltas and applying the result.

accounting/src/delta/delta_data_collection/tests/delta_delta_undo_tests.rs
```diff
@@ -0,0 +1,89 @@
...
+fn delta_delta_undo_associativity(
+    #[case] delta1: DataDelta<char>,
+    #[case] delta2: DataDelta<char>,
+) {
+    {
+        // (Delta1 + Delta2) + Undo = Delta1
+        // every delta goes into separate collection
+        let (mut collection1, collection2, collection3) =
+            make_collections_with_undo(delta1.clone(), delta2.clone());
+        let expected_collection = collection1.clone();
```
TheQuantumPhysicist, line 42: If both are expected to have the same result, please take this out of the scope.

accounting/src/delta/combine.rs
```diff
@@ -15,30 +15,41 @@
...
-use crate::{
-    delta::delta_data_collection::{DataDelta, DeltaMapElement},
-    error::Error,
-};
+use crate::{delta::delta_data_collection::DeltaMapElement, error::Error};
 
 /// Combine data with an element of `DeltaDataCollection`.
 /// An element can be either a Delta or a result of delta undo.
-pub fn combine_data_with_delta<T: Clone>(
+pub fn combine_data_with_delta<T: Clone + PartialEq>(
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Strictly speaking, we only use PartialEq without Eq when it's justified. Otherwise, it's mandatory to use Eq. PartialEq alone can easily become a hidden bug.

TheQuantumPhysicist (approved):
Nice. Please address the few issues mentioned and will look again, but this looks good so far and is a big improvement. Also would be great to rebase the TransactionVerifier updates on this branch (PoS Accounting, is it?) and make sure that everything adds up. As far as I understand, all these issues are fixed now.

accounting/src/delta/combine.rs
```diff
@@ -15,30 +15,41 @@
...
-        (Some(_), Some(d)) => match d.get_data_delta() {
-            DataDelta::Create(_) => Err(Error::DataCreatedMultipleTimes),
-            DataDelta::Modify(_, d) => Ok(Some(d.clone())),
-            DataDelta::Delete(_) => Ok(None),
-        },
+        (Some(data), Some(delta)) => {
+            let delta = delta.get_data_delta();
+            match (&delta.prev(), &delta.next()) {
+                (None, None) => Err(Error::RemoveNonexistingData),
+                (None, Some(_)) => Err(Error::DataCreatedMultipleTimes),
```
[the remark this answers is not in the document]
azarovh: My intention was to treat these cases more user-friendly. For example, if Data is `Some` and Delta is `(None, Some)` that means that the data was already created and we try to create it once more.

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
```diff
@@ -17,66 +17,55 @@ use super::*;
...
-    make_three_collections(Modify('a', 'b'), Modify('b', 'c'), Delete('c'))
-)]
+#[rustfmt::skip]
+#[case(DataDelta::new(None,      Some('a')), DataDelta::new(Some('a'), Some('b')), DataDelta::new(Some('b'), Some('c')), DataDelta::new(None, Some('c')))]
+#[case(DataDelta::new(None,      Some('a')), DataDelta::new(Some('a'), Some('b')), DataDelta::new(Some('b'), None),      DataDelta::new(None, None))]
+#[case(DataDelta::new(None,      Some('a')), DataDelta::new(Some('a'), None),      DataDelta::new(None,      Some('b')), DataDelta::new(None, Some('b')))]
+#[case(DataDelta::new(Some('a'), None),      DataDelta::new(None,      Some('b')), DataDelta::new(Some('b'), None),      DataDelta::new(Some('a'), None))]
+#[case(DataDelta::new(Some('a'), None),      DataDelta::new(None,      Some('b')), DataDelta::new(Some('b'), Some('c')), DataDelta::new(Some('a'), Some('c')))]
+#[case(DataDelta::new(Some('a'), Some('b')), DataDelta::new(Some('b'), Some('c')), DataDelta::new(Some('c'), None),      DataDelta::new(Some('a'), None))]
+#[case(DataDelta::new(Some('a'), Some('b')), DataDelta::new(Some('b'), None),      DataDelta::new(None, Some('c')),      DataDelta::new(Some('a'), Some('c')))]
```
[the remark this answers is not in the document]
azarovh: I added proptests to check associativity as you suggested. 
But I'm not sure that checking paths with errors explicitly is worth it. Minimal error cases are covered in `test_combine_deltas` and proptest check errors associativity.

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
```diff
@@ -0,0 +1,107 @@
...
+    // Invalid combinations can be generated, but the associativity property
+    // doesn't apply for such cases as different Error can be produced depending on the
+    // order of operations. So for the sake of this test errors are treated as None
+    // and it is only expected that both sequences must produce either the same valid result
+    // or both fail with some error.
+    #[test]
+    fn random_delta_associativity(
+        delta1 in any::<DataDelta<char>>(),
+        delta2 in any::<DataDelta<char>>(),
+        delta3 in any::<DataDelta<char>>(),
```
iljakuklic, line 82: If I recall correctly, if you use the `any` generator, there is a shortcut where you can say just `delta1: DataDelta<char>` etc. It follows more closely the usual function syntax. It's up to you whether you deem it nicer or not.

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
```diff
@@ -0,0 +1,107 @@
...
+        // every delta is applied to the same collection
+        let mut collection = DeltaDataCollection::new();
+        let _ = collection.merge_delta_data_element(1, delta1).unwrap();
+        let _ = collection.merge_delta_data_element(1, delta2).unwrap();
+        let _ = collection.merge_delta_data_element(1, delta3).unwrap();
+
+        assert_eq!(collection, expected_collection);
+    }
+}
+
```
iljakuklic, line 70: I think we can test the errors in a limited way too. If there is only one error, applying delta one by one should give the same error as combining deltas. Since errors only care about whether the value is `None` or `Some(_)`, we can use `Delta<()>` for testing. We have two deltas, first one applies cleanly, then the second one fails. Applying the second delta to the value updated by the first should give the same error as combining the two deltas together.

Here is the pseudocode:
```rust
fn error_consistency(x0, x1, x3: Delta<()>) {
    // Ensure x2 does not match
    let x2 = match x1 { None => Some(()), Some(()) => None };
    let d1 = Delta::new(x0, x1);
    let d2 = Delta::new(x2, x3);
    let apply_err = d2.apply(x1).expect_err("apply should fail");
    let combine_err = combine(d1, d2).expect_err("combine should fail");
    assert_eq!(apply_err, combine_err);
}
```

Then instantiate the test with all 8 combinations of `x0`, `x1`, `x3` = {`None`, `Some(())`}.

This does not give full associativity in presence of multiple errors but at least we can verify there is some consistency between updating a value using a delta and combining deltas together.

accounting/src/delta/delta_data_collection/tests/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+        });
+        let some_none = any::<T>().prop_map(|v| DataDelta {
+            old: Some(v),
+            new: None,
+        });
+        let some_some = (any::<T>(), any::<T>()).prop_map(|(old, new)| DataDelta {
+            old: Some(old),
+            new: Some(new),
+        });
+        prop_oneof![none, none_some, some_none, some_some].boxed()
```
iljakuklic, line 46: Why not something like this?
```rust
any::<(Option<T>, Option<T>)>().prop_map(DataDelta::new)
```

iljakuklic (approved):
Looks good. Just a couple of final minor suggestions.

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
iljakuklic: This comment is probably attached to the wrong file but the idea still stands I think.

accounting/src/delta/combine.rs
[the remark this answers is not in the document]
iljakuklic: In the first case, for example, the data do exist (look at the `match` arm one level up) but the error says `RemoveNonexistingData`. This might be fixed in the more recent version. If so, please ignore this comment.

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
azarovh: indeed, thanks

accounting/src/delta/delta_data_collection/tests/mod.rs
azarovh: nice, didn't figure that out, thanks

accounting/src/delta/delta_data_collection/tests/delta_delta_delta_tests.rs
azarovh: Currently, errors for applying delta over data are more descriptive than for combining deltas. I agree that it makes sense to add such test, but it will compare the fact that an error is there (i.e. 'is_err()')

iljakuklic: > Currently, errors for applying delta over data are more descriptive than for combining deltas. I agree that it makes sense to add such test, but it will compare the fact that an error is there (i.e. 'is_err()')

Ok. I think it's fine to ignore this.

TheQuantumPhysicist (approved):

azarovh:
Finally merging this to master... Thank you guys for the support!

TheQuantumPhysicist:
> Finally merging this to master... Thank you guys for the support!

The most important thing to remember is that we don't give up until things are provably correct. Cheers!

PR #620, opened 2023-01-05
utxos: Generalize how lifetimes are handled

iljakuklic:
Generalize the generic parametres of `UtxosDB`, `UtxosView`, `UtxosCache` and such. Before, the values were hard-coded to be a reference (with a lifetime associated with it). Instead use the value directly and let the user define whether the objects (such as `UtxosDB`) take the underlying storage by value, by reference, by mut reference, using `Arc` or whatever. Managed to get rid of `UtxosViewCow` in the process.

Also some mild refactoring to unify `UtxosDB` and `UtxosDBMut` which makes things a bit simpler and removes some code duplication.

This is just a first step, I'd like to do the same with the remaining objects in `TransactionVerifier` but want to gather some early feedback regarding the general approach. Hopefully it will be sufficient to make the lifetimes work in the case where the underlying storage is actually a handle to another subsystem.

### DIFF
--- a/chainstate/src/detail/chainstateref/tx_verifier_storage.rs
+++ b/chainstate/src/detail/chainstateref/tx_verifier_storage.rs
@@ -34,7 +34,7 @@ use common::{
     primitives::Id,
 };
 use tx_verifier::transaction_verifier::TransactionSource;
-use utxo::{ConsumedUtxoCache, FlushableUtxoView, UtxosDBMut, UtxosStorageRead};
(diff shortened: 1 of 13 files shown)


TheQuantumPhysicist (approved):
Looks good. I just hope removing the lifetime from UtxosView doesn't clip some corner case.

azarovh (approved):
looks good to me

muursh (approved):

PR #622, opened 2023-01-06
Generalize more lifetimes in `TransactionVerifier`

iljakuklic:
Similar changes as in #620 but applied to the remaining fields in `TransactionVerifier`. Also small refactoring to make it work.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -181,11 +181,11 @@ impl GuardedTxIndexCache {
     }
 
     fn as_ref(&self) -> Option<&TxIndexCache> {
-        self.enabled.then(|| &self.inner)
+        self.enabled.then_some(&self.inner)
     }
 
     fn as_mut(&mut self) -> Option<&mut TxIndexCache> {
-        self.enabled.then(|| &mut self.inner)
+        self.enabled.then_some(&mut self.inner)
     }
 
     /// Take the inner cache, even if disabled
@@ -257,9 +257,7 @@ where
     S: TransactionVerifierStorageRef,
     U: UtxosView,
 {
-    pub fn derive_child<'a>(
-        &'a self,
-    ) -> TransactionVerifier<&'a ChainConfig, &'a Self, &'a UtxosCache<U>> {
+    pub fn derive_child(&self) -> TransactionVerifier<&ChainConfig, &Self, &UtxosCache<U>> {
         TransactionVerifier {
             storage: self,
             chain_config: self.chain_config.as_ref(),
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -198,7 +198,6 @@ impl GuardedTxIndexCache {
 pub struct TransactionVerifier<'a, S, U> {
     chain_config: &'a ChainConfig,
     storage_ref: S,
-    verifier_config: TransactionVerifierConfig,
     tx_index_cache: GuardedTxIndexCache,
     utxo_cache: UtxosCache<U>,
     utxo_block_undo: BTreeMap<TransactionSource, BlockUndoEntry>,
@@ -221,7 +220,6 @@ impl<'a, S: TransactionVerifierStorageRef + Clone> TransactionVerifier<'a, S, Ut
         Self {
             storage_ref,
             chain_config,
-            verifier_config,
             tx_index_cache,
             utxo_cache,
             utxo_block_undo: BTreeMap::new(),
@@ -253,7 +251,6 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView + Send + Sync>
             utxo_block_undo: BTreeMap::new(),
             token_issuance_cache: TokenIssuanceCache::new(),
             best_block,
-            verifier_config,
         }
     }
 }
@@ -263,7 +260,6 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
         TransactionVerifier {
             storage_ref: self,
             chain_config: self.chain_config,
-            verifier_config: self.verifier_config.clone(),
             tx_index_cache: GuardedTxIndexCache::new(self.tx_index_cache.enabled),
             utxo_cache: UtxosCache::new(&self.utxo_cache),
             utxo_block_undo: BTreeMap::new(),
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -195,8 +195,8 @@ impl GuardedTxIndexCache {
 }
 
 /// The tool used to verify transaction and cache their updated states in memory
-pub struct TransactionVerifier<'a, S, U> {
-    chain_config: &'a ChainConfig,
+pub struct TransactionVerifier<CC, S, U> {
+    chain_config: CC,
     storage_ref: S,
     tx_index_cache: GuardedTxIndexCache,
     utxo_cache: UtxosCache<U>,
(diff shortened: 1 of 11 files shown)


chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -157,81 +157,118 @@ pub struct TransactionVerifierDelta {
...
     best_block: Id<GenBlock>,
 }
 
-impl<'a, S: TransactionVerifierStorageRef> TransactionVerifier<'a, S, UtxosDB<&'a S>> {
-    pub fn new(
-        storage_ref: &'a S,
-        chain_config: &'a ChainConfig,
-        verifier_config: TransactionVerifierConfig,
-    ) -> Self {
+impl<CC, S: TransactionVerifierStorageRef + Clone> TransactionVerifier<CC, S, UtxosDB<S>> {
```
iljakuklic, line 208: I am a bit uncomfortable with the `Clone` here. The idea is that storage is some form of reference, smart pointer or handle and cloning it does not clone the contents. Like for example when cloning a `&T` or using `Arc::clone`.

Maybe we can introduce a marker `trait ShallowClone: Clone {}` that is only implemented for types that do shallow clone. Then use it here instead of plain `Clone` to constrain it a bit more precisely.

Comments / counterproposals?

azarovh: I think the idea with `ShallowClone` is worth implementing

azarovh (approved):

chainstate/tx-verifier/src/transaction_verifier/mod.rs
alfiedotwtf: > I am a bit uncomfortable with the Clone here. The idea is that storage is some form of reference, smart pointer or handle and cloning it does not clone the contents.

Should it be renamed something like StorageHandle? That way, you know it's a handle to the underlying, just like a database handle.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -675,10 +693,10 @@ impl<'a, S: TransactionVerifierStorageRef, U: UtxosView> TransactionVerifier<'a,
         // TODO: test spending block rewards from chains outside the mainchain
         if let Some(inputs) = reward_transactable.inputs() {
             let tx_index_fetcher =
-                |tx_id: &OutPointSourceId| self.storage_ref.get_mainchain_tx_index(tx_id);
+                |tx_id: &OutPointSourceId| self.storage.get_mainchain_tx_index(tx_id);
```
alfiedotwtf, line 696: Just a quick coding style question - why do you create a fetcher here, but on line 664 you inline instead?

alfiedotwtf (approved):

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: I don't mind creating a ShallowClone trait... at least now I don't think of a reason why we shouldn't do that.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -157,81 +157,118 @@ pub struct TransactionVerifierDelta {
...
+    }
+}
+
 /// The tool used to verify transaction and cache their updated states in memory
-pub struct TransactionVerifier<'a, S, U> {
-    chain_config: &'a ChainConfig,
-    storage_ref: &'a S,
-    verifier_config: TransactionVerifierConfig,
-    tx_index_cache: TxIndexCache,
+pub struct TransactionVerifier<CC, S, U> {
```
TheQuantumPhysicist, line 198: CC breaks the formatting rules. It should be just `C`.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -157,81 +157,118 @@ pub struct TransactionVerifierDelta {
     token_issuance_cache: ConsumedTokenIssuanceCache,
 }
 
+/// [`TxIndexCache`] that can be enabled or disabled (using a config).
+pub struct GuardedTxIndexCache {
+    enabled: bool,
```
TheQuantumPhysicist, line 162: Maybe `Option<TxIndexCache>`?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -157,81 +157,118 @@ pub struct TransactionVerifierDelta {
     token_issuance_cache: ConsumedTokenIssuanceCache,
 }
 
+/// [`TxIndexCache`] that can be enabled or disabled (using a config).
+pub struct GuardedTxIndexCache {
```
TheQuantumPhysicist, line 161: The file is becoming large again. Can you please move this to some submodule?

TheQuantumPhysicist: I have to say also... I don't like the word "guarded"... because "guarding" is reserved for safety operations. I don't have another word in mind though... how about `MaybeTxIndexCache`? Cheesy but expressive xD.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -157,81 +157,118 @@ pub struct TransactionVerifierDelta {
...
+    fn as_ref(&self) -> Option<&TxIndexCache> {
+        self.enabled.then_some(&self.inner)
+    }
+
+    fn as_mut(&mut self) -> Option<&mut TxIndexCache> {
+        self.enabled.then_some(&mut self.inner)
+    }
+
+    /// Take the inner cache, even if disabled
+    fn take_always(self) -> TxIndexCache {
```
TheQuantumPhysicist, line 192: Instead of take_always(), we can just use the Option as returned value and then do `or_default()` and the map will be empty.

TheQuantumPhysicist (approved):

chainstate/tx-verifier/src/transaction_verifier/mod.rs
iljakuklic: > > I am a bit uncomfortable with the Clone here. The idea is that storage is some form of reference, smart pointer or handle and cloning it does not clone the contents.
> 
> Should it be renamed something like StorageHandle? That way, you know it's a handle to the underlying, just like a database handle.

Hmm maybe the field, which I have renamed from `storage` to `storage_ref` should have stayed called `storage_ref`.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
iljakuklic: That's just a result of me trying out different approaches to refactoring this. Will make it more consistent, thanks.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
iljakuklic: Hmm `OptionalTxIndexCache`?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: It's supposed to be fixed once the node is initialized and never change unless a reindexing happens.

TheQuantumPhysicist: If I wasn't clear, yes, it should be set on start up but never change later.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: That too is OK with me.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
iljakuklic: > If I wasn't clear, yes, it should be set on start up but never change later.

It's still not quite clear to me whether it is a coding time decision, with some instances of tx verifier having the index and others lacking it, or it is something user can set when starting the node e.g. via config or command line.

TheQuantumPhysicist: Yep. It's configurable. A transaction index is only required if you want to index transactions so that you can access all transaction history using their id. This is usually only required for block explorers, not normal users. Normal users usually have their transactions indexed by their wallets.

So, as a user, you set whether you want the transaction index to be supported when you start the node the first time, and we expect that to never change. Changing that requires to dump the current index completely, and reindex from genesis. This cannot be changed in the middle of operation or even when you start the node after shutting it down. A full reindex is required.

stanislav-tkach (approved):

PR #623, opened 2023-01-07
Remove libp2p backend

stanislav-tkach:


stanislav-tkach:
The functional tests passed after switching to mock backend, but the peer manager peerdb tests need to be made generic and some additional tweaks are required.

p2p/src/net/types/mod.rs
```diff
@@ -46,7 +46,7 @@ pub struct AddrInfo<T: NetworkingService> {
 /// (both are Mintlayer nodes and that both support mandatory protocols). If those checks pass,
 /// the information is passed on to [crate::peer_manager::PeerManager] which decides whether it
 /// wants to keep the connection open or close it and possibly ban the peer from.
-#[derive(Debug)]
+#[derive(Debug, Eq)]
```
TheQuantumPhysicist, line 49: I think this is playing with fire. We shouldn't inherit Eq here. What does it mean that PeerInfo A == PeerInfo B? What if peer_id of A == peer_id of B, but the agent is different. Does that mean they're not the same peer?

Please understand that I know that this is "impossible" based on the invariants of our software. But you gotta ask yourself what it means that PeerInfo can have equality, but not the peer_id. I think this is wrong and should be removed, and wherever equality is needed should be examined and we should understand "what we're trying to prove" with that equality, and then see what equality we should do there. Maybe just equating the peer_id is more than enough. In the case of trait-bounds, maybe we should create another trait that satisfies this property.

I'm a huge opponent of throwing Eq and other inherited traits without carefully considering what that means. It can easily become a bug.

pavel-kokolemin: `Eq` is only needed in one test (`connect_to_remote`). I think it will be better rewire the test instead.

p2p/src/testing_utils.rs
```diff
@@ -88,17 +79,79 @@ impl TestTransportMaker for TestTransportChannel {
...
+    /// An identifier type.
+    type PeerId;
+
+    /// Creates a new peer identifier.
+    fn new() -> Self::PeerId;
+}
+
+pub struct TestMockPeerIdMaker {}
+
+impl RandomPeerIdMaker for TestMockPeerIdMaker {
```
pavel-kokolemin, line 110: Just an nitpick.
Do we really need `RandomPeerIdMaker`? There is just one implementation of that trait and it's only used to call `MockPeerId::new()`.
And I think later backend can reuse same type for peer ids (`MockPeerId` currently).

pavel-kokolemin (commented):
From `test/functional/test_framework/.mintlayer/config.toml`:
```
[p2p]
bind_address = "/ip6/::1/tcp/3031"
```
Please replace libp2p addresses there and in some other places (can be found with `/ip6/`).

test/functional/test_framework/test_framework.py
```diff
@@ -557,20 +557,11 @@ def restart_node(self, i, extra_args=None):
...
     def connect_nodes(self, a, b):
         count_a = self.nodes[a].p2p_get_peer_count()
         count_b = self.nodes[b].p2p_get_peer_count()
 
-        # TODO: Replace with get_node_addr_mock once libp2p removed
-        addr_a = self.get_node_addr_libp2p(a)
+        addr_a = p2p_url(a)
         ret = self.nodes[b].p2p_connect(addr_a)
 
         wait_until_helper(lambda:
```
TheQuantumPhysicist, line 567: I want to say that while this works now, this can easily lead to a dead-lock. You can image a situation where a connection is lost from one peer and they never reach the +1 value. It's rare, but possible. There should be a better way to do this... though not in this PR. Please create an issue for this.

The better solution would be checking that we're connecting to a peer by some kind of id, as node label. I think bitcoin does something like that. When a node is created and run, it's given a label that's attached to its version that can be used to identify the node.

TheQuantumPhysicist (approved):
Looking good. Please fix the minor issue I mentioned.

p2p/src/lib.rs
```diff
@@ -151,10 +157,12 @@ pub async fn make_p2p(
...
     mempool_handle: mempool::MempoolHandle,
-) -> crate::Result<Box<dyn P2pInterface>> {
-    let transport = net::libp2p::make_transport(&p2p_config);
+) -> Result<Box<dyn P2pInterface>> {
+    let stream_adapter = NoiseEncryptionAdapter::gen_new();
+    let base_transport = net::mock::transport::TcpTransportSocket::new();
+    let transport = NoiseTcpTransport::new(stream_adapter, base_transport);
 
-    let p2p = P2p::<net::libp2p::Libp2pService>::new(
+    let p2p = P2p::<MockService<NoiseTcpTransport>>::new(
```
TheQuantumPhysicist, line 165: Maybe in a next PR we can fix the "mock" naming.

TheQuantumPhysicist: Maybe we should call it DefaultNetworkingService?

test/functional/test_framework/test_framework.py
stanislav-tkach: https://github.com/mintlayer/mintlayer-core/issues/624

pavel-kokolemin (approved):

PR #625, opened 2023-01-09
Shallow clone

iljakuklic:
As proposed in discussion for #622

### DIFF
--- a/utils/src/shallow_clone.rs
+++ b/utils/src/shallow_clone.rs
@@ -1,0 +1,45 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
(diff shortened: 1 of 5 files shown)


utils/src/shallow_clone.rs
```diff
@@ -0,0 +1,45 @@
...
+pub trait ShallowClone: Clone {}
+
+// Some impls for types from the standard library
+impl<T> ShallowClone for &T {}
+impl<T> ShallowClone for &[T] {}
+impl<T> ShallowClone for *const T {}
+impl<T> ShallowClone for *mut T {}
+impl<T> ShallowClone for std::rc::Rc<T> {}
+impl<T> ShallowClone for std::sync::Arc<T> {}
+
```
TheQuantumPhysicist, line 32: Maybe you wanna implement this for Subsystem handles too 😄

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

utils/src/shallow_clone.rs
iljakuklic: Yea there's probably more stuff like this. For example the `Storage` type itself can also be cloned without duplicating the database. Will add that as the need arises.

PR #628, opened 2023-01-11
Prevent self-connects

pavel-kokolemin:
Prevent when node is trying to connect to itself (which is pointless). We already checked dialed socket address against listening address but that's not robust enough. For example this check won't work if node is listening on 0.0.0.0 and trying to connect to public IP of the host. I replaced this check with random nonces sent in handshake. Similar checks are done in Bitcoin Core and Grin, for example.

### DIFF
--- a/p2p/src/net/mock/backend.rs
+++ b/p2p/src/net/mock/backend.rs
@@ -461,7 +461,7 @@ where
 
             if let Some(outbound_peer_id) = outbound_peer_id {
                 let outbound_pending =
-                    self.pending.remove(&outbound_peer_id).expect("peer must exists");
+                    self.pending.remove(&outbound_peer_id).expect("peer must exist");
 
                 log::info!(
                     "self-connection detect on address {:?}",
--- a/p2p/src/net/mock/backend.rs
+++ b/p2p/src/net/mock/backend.rs
@@ -441,6 +441,50 @@ where
         Ok(())
     }
 
+    async fn is_connection_from_self(
+        &mut self,
+        peer_role: PeerRole,
+        incoming_nonce: u64,
+    ) -> crate::Result<bool> {
+        if peer_role == PeerRole::Inbound {
+            // Look for own outbound connection with same nonce
+            let outbound_peer_id = self
+                .pending
+                .iter()
+                .find(|(_peer_id, pending)| {
+                    pending.peer_role
+                        == PeerRole::Outbound {
+                            handshake_nonce: incoming_nonce,
+                        }
+                })
+                .map(|(peer_id, _pending)| *peer_id);
+
+            if let Some(outbound_peer_id) = outbound_peer_id {
+                let outbound_pending =
+                    self.pending.remove(&outbound_peer_id).expect("peer must exists");
+
+                log::info!(
+                    "self-connection detect on address {:?}",
+                    outbound_pending.address
+                );
+
+                // Report outbound connection failure
+                self.conn_tx
+                    .send(ConnectivityEvent::ConnectionError {
+                        address: outbound_pending.address,
+                        error: P2pError::DialError(DialError::AttemptToDialSelf),
+                    })
+                    .await
+                    .map_err(P2pError::from)?;
+
+                // Nothing else to do, just drop inbound connection
+                return Ok(true);
+            }
+        }
+
+        Ok(false)
+    }
+
     async fn handle_peer_event(
         &mut self,
         peer_id: MockPeerId,
@@ -464,37 +508,8 @@ where
                     None => return Ok(()),
                 };
 
-                if peer_role == PeerRole::Inbound {
-                    // Look for own outbound connection with same nonce
-                    let outbound_peer_id = self
-                        .pending
-                        .iter()
-                        .find(|(_peer_id, pending)| {
-                            pending.peer_role == PeerRole::Outbound { handshake_nonce }
-                        })
-                        .map(|(peer_id, _pending)| *peer_id);
-
-                    if let Some(outbound_peer_id) = outbound_peer_id {
-                        let outbound_pending =
-                            self.pending.remove(&outbound_peer_id).expect("peer must exists");
-
-                        log::info!(
-                            "self-connection detect on address {:?}",
-                            outbound_pending.address
-                        );
-
-                        // Report outbound connection failure
-                        self.conn_tx
-                            .send(ConnectivityEvent::ConnectionError {
-                                address: outbound_pending.address,
-                                error: P2pError::DialError(DialError::AttemptToDialSelf),
-                            })
-                            .await
-                            .map_err(P2pError::from)?;
-
-                        // Nothing else to do, just drop inbound connection
-                        return Ok(());
-                    }
+                if self.is_connection_from_self(peer_role, handshake_nonce).await? {
+                    return Ok(());
                 }
 
                 match peer_role {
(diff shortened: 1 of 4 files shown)


p2p/src/net/mock/peer.rs
```diff
@@ -74,10 +82,25 @@ where
...
+            Role::Inbound => None,
+            Role::Outbound => Some(make_true_rng().gen()),
+        };
+
+        if let Some(handshake_nonce) = handshake_nonce {
+            handshake_nonce_store
+                .lock()
+                .expect("handshake_nonce_store mutex must be valid")
+                .insert(handshake_nonce);
+        }
```
iljakuklic, line 102: Minor: I'd probably prefer to fold the `if let` into the `Role::Outbound` arm of the preceding `match`. That way you don't need to analyse data flow to see what control flow paths are possible.

iljakuklic (approved):
Looks good

p2p/src/net/mock/peer.rs
pavel-kokolemin: Sure, updated.

p2p/src/net/mock/types.rs
```diff
@@ -193,6 +193,8 @@ pub enum MockEvent {
     SendMessage(Box<Message>),
 }
 
+pub type HandshakeNonce = u128;
```
TheQuantumPhysicist, line 196: I think a u64 is more than enough for this purpose. This doesn't have to be cryptographically secure.

p2p/src/net/mock/peer.rs
```diff
@@ -60,6 +61,13 @@ pub struct Peer<T: TransportSocket> {
 
     /// RX channel for receiving commands from backend
     rx: mpsc::Receiver<MockEvent>,
+
+    /// List of nonces of all active connections
+    handshake_nonce_store: Arc<HandshakeNonceStore>,
```
TheQuantumPhysicist, line 66: I'm very confused to why this even exists. Why do we have to store a list of all nonces from all peers?

The only purpose of this nonce is to detect if we're connecting to ourselves somehow, which can never be verified with networking means because proxies exist. So, we have a very simple tool to make that check.

Now I see we're storing the nonce of every single peer. What could we possibly use that for? We cannot use it to id the peer because a malicious peer can just create an infinite number of them arbitrarily... these numbers aren't authenticated by any means. We're back at the "peer id" thing, I feel.

Maybe I'm missing something... I don't know.

p2p/src/net/mock/peer.rs
```diff
@@ -74,10 +82,25 @@ where
...
         tx: mpsc::Sender<(MockPeerId, PeerEvent)>,
         rx: mpsc::Receiver<MockEvent>,
     ) -> Self {
         let socket = BufferedTranscoder::new(socket);
+
+        // Generate random nonce
+        let handshake_nonce = match role {
+            Role::Inbound => None,
+            Role::Outbound => {
+                let nonce = make_true_rng().gen();
```
TheQuantumPhysicist, line 95: Two things:

1. This doesn't have to be a true random number. Pseudo-random is more than enough.
2. This number should be generated exactly once when the p2p system is initialized, and never generated again.

I have to say I'm not fully sure how this works together.

pavel-kokolemin: > This number should be generated exactly once when the p2p system is initialized, and never generated again.

That will be much easier. But are you sure about this? I used random number generator to make fingerprinting more difficult. For example if someday we add TOR addresses then it will be very easy to associate IP address and TOR address of some node. For example Bitcoin Core and Grin use random nonces to detect self-connections.

p2p/src/net/mock/types.rs
pavel-kokolemin: Sure, updated.

p2p/src/net/mock/peer.rs
TheQuantumPhysicist: Does bitcoin generate a new number on every connection? If I remember correctly, it's a static variable that they initialize once. Did they change that?

Let's have a call about it.

p2p/src/net/mock/peer.rs
pavel-kokolemin: Sorry, the comment was wrong. I updated it:

```
    /// List of nonces of all outbound connections in handshake state.
    /// Used to detect and drop self connections.
    handshake_nonce_store: Arc<HandshakeNonceStore>,
```

I also now remove own nonce from the store once handshake complete, because we don't need it after that.

p2p/src/net/mock/peer.rs
pavel-kokolemin: They derive new value for every peer:

```
    NodeId id = GetNewNodeId();
    uint64_t nonce = GetDeterministicRandomizer(RANDOMIZER_ID_LOCALHOSTNONCE).Write(id).Finalize();
```

p2p/src/net/mock/backend.rs
```diff
@@ -59,7 +58,7 @@ use crate::{
     types::peer_address::PeerAddress,
 };
 
-use super::transport::TransportAddress;
+use super::{transport::TransportAddress, types::HandshakeNonce};
```
TheQuantumPhysicist, line 62: Please fix the word "recevied" below

pavel-kokolemin: Fixed, thanks.

pavel-kokolemin:
@TheQuantumPhysicist asked to not introduce additional state and check nonces differently.
So I updated the logic to check nonces inside `Backend` after receiving `PeerInfoReceived` event.

But after adding unit test I noticed some unexpected timeout.
It was caused because `Backend` event loop was blocked when new connection is requested (and that become apparent when I was testing self-connections).
So I spent some time fixing that too (I was planing to fix this for a while).
New outbound connections are added to `FuturesUnordered` and handled in the even loop.

p2p/src/net/mock/peer.rs
```diff
@@ -91,29 +107,22 @@ where
...
-                            network: *self.chain_config.magic_bytes(),
-                            subscriptions: (*self.p2p_config.node_type.as_ref()).into(),
-                            receiver_address: self.receiver_address.clone(),
-                        },
-                    ))
-                    .await?;
-
+                // Send PeerInfoReceived before sending handshake to remote peer!
+                // Backend is expected to receive PeerInfoReceived before outgoing connection has chance to complete handshake,
+                // It's required to relaeble detect self-connects.
```
TheQuantumPhysicist, line 125: reliable

pavel-kokolemin: Fixed, thanks.

pavel-kokolemin:
I reverted my fix for even loop blocking, it will be done in different PR. Once fixed, `self_connect_channels` and `self_connect_noise` tests can be enabled.

p2p/src/net/mock/backend.rs
```diff
@@ -444,12 +452,53 @@ where
...
+
+                        // Nothing else to do, just drop inbound connection
+                        return Ok(());
+                    }
+                }
 
-                match role {
-                    Role::Outbound => {
+                match peer_role {
+                    PeerRole::Outbound { .. } => {
```
TheQuantumPhysicist, line 501: Please don't use `..`. If you don't need the variable, put explicitly a `_`. The problem with `..` is that it prevents the compiler from reacting when you add more variables in the future.

pavel-kokolemin: Sure, fixed.

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

PR #632, opened 2023-01-13
Fix backend event loop blocking

pavel-kokolemin:
Make outbound connections in background.

### DIFF
--- a/p2p/src/net/mock/transport/traits/socket.rs
+++ b/p2p/src/net/mock/transport/traits/socket.rs
@@ -16,6 +16,7 @@
 use std::{fmt::Debug, hash::Hash, str::FromStr};
 
 use async_trait::async_trait;
+use futures::future::BoxFuture;
 
 use crate::{net::AsBannableAddress, Result};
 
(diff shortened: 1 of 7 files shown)


TheQuantumPhysicist (approved):

p2p/src/net/mock/transport/traits/socket.rs
```diff
@@ -52,5 +53,5 @@ pub trait TransportSocket: Send + Sync + 'static {
     async fn bind(&self, address: Vec<Self::Address>) -> Result<Self::Listener>;
 
     /// Open a connection to the given address.
-    async fn connect(&self, address: Self::Address) -> Result<Self::Stream>;
+    fn connect(&self, address: Self::Address) -> BoxFuture<'static, crate::Result<Self::Stream>>;
```
stanislav-tkach, line 56: Nitpick: perhaps the comment should be updated to reflect the changes. Something like "...returns a future that opens a connection...".

stanislav-tkach (approved):
👍

p2p/src/net/mock/transport/traits/socket.rs
pavel-kokolemin: Sure, will do, thanks. Will do in next PR (don't want to wait for CI checks).

PR #638, opened 2023-01-18
Remove the 'mock' part from the p2p naming

stanislav-tkach:
This pull request doesn't change any part of p2p logic, only removes the "mock" part from the naming, because the entity that was previously known as a "mock backend" is now a default and the only existent backend at the moment.

Notable changes:
- The `mock` module was renamed to `default_backend`.
- `MockChannelTransport` was renamed to `TestChannelTransport` because it is only used in tests.
- `MockPeerInfo` was completely removed because `PeerInfo` is exactly the same, so it is used everywhere instead.
- `MockService` was renamed to `Service` and this change I'm mostly unsure of.
- In other places the "mock" part was simply removed, for example `MockPeerId` becomes `PeerId`.

I will gladly will do more renaming if anyone has good suggestions.


pavel-kokolemin (approved):

p2p/src/net/default_backend/transport/impls/channel.rs
```diff
@@ -74,28 +74,30 @@ static CONNECTIONS: Lazy<Mutex<BTreeMap<Address, UnboundedSender<IncomingConnect
...
     pub fn new() -> Self {
         let local_address = NEXT_ADDRESS.fetch_add(1, Ordering::Relaxed);
-        MockChannelTransport { local_address }
+        TestChannelTransport { local_address }
     }
 }
 
 #[async_trait]
-impl TransportSocket for MockChannelTransport {
+impl TransportSocket for TestChannelTransport {
```
TheQuantumPhysicist, line 96: I would call this MpscChannelTransport? Feel free to suggest a better name. The fact that it's for testing doesn't need to be in the name.

p2p/src/net/default_backend/mod.rs
```diff
@@ -33,87 +33,70 @@ use crate::{
...
         },
-        types::{ConnectivityEvent, PeerInfo, PubSubTopic, SyncingEvent},
+        types::{ConnectivityEvent, PubSubTopic, SyncingEvent},
         ConnectivityService, NetworkingService, SyncingMessagingService,
     },
 };
 
 #[derive(Debug)]
-pub struct MockService<T: TransportSocket>(PhantomData<T>);
+pub struct Service<T: TransportSocket>(PhantomData<T>);
```
TheQuantumPhysicist, line 47: DefaultNetworkingService?

TheQuantumPhysicist (approved):
Looking good... but some names can be improved. Please make sure that you don't name anything like TestSomething.

stanislav-tkach:
I have renamed `TestChannelTransport`, but we still have entities with the word "test" in them, such as `TestTransport`. I suppose it is OK, because these names were there before my renaming.

PR #639, opened 2023-01-18
Enable p2p test that was disabled before

pavel-kokolemin:


TheQuantumPhysicist (approved):

p2p/backend-test-suite/src/connect.rs
```diff
@@ -48,9 +43,6 @@ where
     .unwrap();
 }
 
-// Check that connecting twice to the same address isn't possible.
```
stanislav-tkach, line 51: Am I missing something or the comment is still relevant?

iljakuklic (approved):

p2p/backend-test-suite/src/connect.rs
pavel-kokolemin: Sorry, my bad, reverted.

PR #640, opened 2023-01-20
Remove duplicated copyright

alfiedotwtf:
Minor fix to remove a duplicated copyright notice

### DIFF
--- a/common/src/primitives/amount.rs
+++ b/common/src/primitives/amount.rs
@@ -1,26 +1,4 @@
-// Copyright (c) 2022 RBB S.r.l
-// opensource@mintlayer.org
-// SPDX-License-Identifier: MIT
-// Licensed under the MIT License;
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


common/src/primitives/amount.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 RBB S.r.l
```
iljakuklic, line 1: Btw: should these years be just updated (as proposed here) or should it be a range like `2022 - 2023`?
Also all headers should be updated (not for this PR).

iljakuklic (approved):

common/src/primitives/amount.rs
alfiedotwtf: Ranges vs static value - good question.

As for all headers, I didn't want to touch all files as this was a small change, but I guess I could do it for all once the "range vs static value" is sorted.

iljakuklic: Just checked and we already have ranges in some of the older files. E.g. `crypto/src/hash/mod.rs` has `2021-2022`, so let's stick with that. No need to bother updating all headers in this PR.

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

common/src/primitives/amount.rs
```diff
@@ -12,6 +12,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//
```
stanislav-tkach, line 15: A very tiny nitpick: I would remove "//" from this line because otherwise the following comment looks like a part of the license header.

alfiedotwtf: Sounds good 👍

PR #641, opened 2023-01-20
Refactor p2p types

pavel-kokolemin:
Let's replace `PeerId` and `PeerRequestId` associated types with concrete types instead. I think that will simplify code a bit. I don't see why we would ever need this after removing libp2p.
There are some other minor updates as well (which don't worth own PR).


stanislav-tkach (approved):

TheQuantumPhysicist:
If PeerId and RequestId are specifically never sent through wire then I'm OK with this change, otherwise this should be reverted. The reason I'm saying this is that I know now PeerId is only an atomic counter that's never shared, but I'm not sure about RequestId. Can you please confirm that both are never sent over wire? Things that are sent over wire should be part of the NetworkingService abstraction.

Also, in the future, when moving modules like `ip_address`, please make them in another PR. If there's conflicts, you can always chain PRs by making of them on top of the other. This seems like mixing unrelated changes.

pavel-kokolemin:
Yes, `RequestId` is sent through wire. I don't think that we ever going to implement yet another `NetworkingService` and even if we do, then I don't see why not re-use same `RequestId` type. So I think that keeping `RequestId` in `NetworkingService` is excessive. But that is a minor issue so let's keep it. I will close this PR and send updated changes in different PRs.

PR #646, opened 2023-01-23
Basic peer discovery

pavel-kokolemin:
This adds a basic peer discovery mechanism similar to Bitcoin's. When a node connects somewhere, it tries to load addresses from remote peers and also advertises its own discovered address. Such advertised addresses are relayed to random peers (two peers) and spread across the network this way.

### DIFF
--- a/p2p/src/config.rs
+++ b/p2p/src/config.rs
@@ -31,6 +31,8 @@ make_config_setting!(
 );
 make_config_setting!(NodeTypeSetting, NodeType, NodeType::Full);
 make_config_setting!(DiscoverPrivateIps, bool, false);
+make_config_setting!(HeartbeatIntervalMin, Duration, Duration::from_secs(5));
+make_config_setting!(HeartbeatIntervalMax, Duration, Duration::from_secs(30));
 
 /// A node type.
 #[derive(Debug, Copy, Clone)]
@@ -74,4 +76,8 @@ pub struct P2pConfig {
     pub node_type: NodeTypeSetting,
     /// Allow announcing and discovering local and private IPs. Should be used for testing only.
     pub discover_private_ips: DiscoverPrivateIps,
+    /// Lower bound for how often `PeerManager::heartbeat` is called. Should be used for testing only.
+    pub heartbeat_interval_min: HeartbeatIntervalMin,
+    /// Upper bound for how often `PeerManager::heartbeat` is called. Should be used for testing only.
+    pub heartbeat_interval_max: HeartbeatIntervalMin,
 }
--- a/p2p/src/config.rs
+++ b/p2p/src/config.rs
@@ -30,7 +30,7 @@ make_config_setting!(
     [PubSubTopic::Blocks, PubSubTopic::Transactions].into_iter().collect()
 );
 make_config_setting!(NodeTypeSetting, NodeType, NodeType::Full);
-make_config_setting!(DiscoverPrivateIps, bool, false);
+make_config_setting!(AllowDiscoverPrivateIps, bool, false);
 make_config_setting!(HeartbeatIntervalMin, Duration, Duration::from_secs(5));
 make_config_setting!(HeartbeatIntervalMax, Duration, Duration::from_secs(30));
 
@@ -75,7 +75,7 @@ pub struct P2pConfig {
     /// A node type.
     pub node_type: NodeTypeSetting,
     /// Allow announcing and discovering local and private IPs. Should be used for testing only.
-    pub discover_private_ips: DiscoverPrivateIps,
+    pub allow_discover_private_ips: AllowDiscoverPrivateIps,
     /// Lower bound for how often `PeerManager::heartbeat` is called. Should be used for testing only.
     pub heartbeat_interval_min: HeartbeatIntervalMin,
     /// Upper bound for how often `PeerManager::heartbeat` is called. Should be used for testing only.
--- a/p2p/src/config.rs
+++ b/p2p/src/config.rs
@@ -30,6 +30,7 @@ make_config_setting!(
     [PubSubTopic::Blocks, PubSubTopic::Transactions].into_iter().collect()
 );
 make_config_setting!(NodeTypeSetting, NodeType, NodeType::Full);
+make_config_setting!(DiscoverPrivateIps, bool, false);
 
 /// A node type.
 #[derive(Debug, Copy, Clone)]
@@ -71,4 +72,6 @@ pub struct P2pConfig {
     pub outbound_connection_timeout: OutboundConnectionTimeout,
     /// A node type.
     pub node_type: NodeTypeSetting,
+    /// Allow announcing and discovering local and private IPs. Should be used for testing only.
+    pub discover_private_ips: DiscoverPrivateIps,
 }
--- a/p2p/src/peer_manager/global_ip.rs
+++ b/p2p/src/peer_manager/global_ip.rs
@@ -1,0 +1,146 @@
+// Copyright 2021 Protocol Labs.
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Based on libp2p sources:
+// https://github.com/libp2p/rust-libp2p/blob/73cbbe29679f5d56d81d5477d3796e82712f9ac2/protocols/autonat/src/behaviour.rs
+
+pub trait GlobalIp {
+    fn is_global_unicast_ip(&self) -> bool;
+}
+
+impl GlobalIp for std::net::Ipv4Addr {
+    // NOTE: The below logic is based on `std::net::Ipv4Addr::is_global`,
+    // which is at the time of writing behind the unstable `ip` feature.
+    // See https://github.com/rust-lang/rust/issues/27709 for more info.
+    fn is_global_unicast_ip(&self) -> bool {
+        // Copied from the unstable method `std::net::Ipv4Addr::is_shared`.
+        fn is_shared(addr: &std::net::Ipv4Addr) -> bool {
+            addr.octets()[0] == 100 && (addr.octets()[1] & 0b1100_0000 == 0b0100_0000)
+        }
+
+        // Copied from the unstable method `std::net::Ipv4Addr::is_reserved`.
+        fn is_reserved(addr: &std::net::Ipv4Addr) -> bool {
+            addr.octets()[0] & 240 == 240 && !addr.is_broadcast()
+        }
+
+        // Copied from the unstable method `std::net::Ipv4Addr::is_benchmarking`.
+        fn is_benchmarking(addr: &std::net::Ipv4Addr) -> bool {
+            addr.octets()[0] == 198 && (addr.octets()[1] & 0xfe) == 18
+        }
+
+        // Addresses reserved for future protocols (`192.0.0.0/24`)
+        fn is_future_protocol(addr: &std::net::Ipv4Addr) -> bool {
+            addr.octets()[0] == 192 && addr.octets()[1] == 0 && addr.octets()[2] == 0
+        }
+
+        // The 0.0.0.0/8 block is reserved
+        fn is_reserved2(addr: &std::net::Ipv4Addr) -> bool {
+            addr.octets()[0] == 0
+        }
+
+        !self.is_multicast()
+            && !self.is_private()
+            && !self.is_loopback()
+            && !self.is_link_local()
+            && !self.is_broadcast()
+            && !self.is_documentation()
+            && !is_shared(self)
+            && !is_future_protocol(self)
+            && !is_reserved(self)
+            && !is_benchmarking(self)
+            && !is_reserved2(self)
+    }
+}
+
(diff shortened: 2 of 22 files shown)


pavel-kokolemin:
File `p2p/src/peer_manager/global_ip.rs` is based on a source file from `libp2p`, that's why the copyright line is different there. BTW, I said earlier that Rust's stdlib has `is_global` method for IP addresses but actually it's behind unstable feature flag (and probably won't be stabilized any time soon). That's why we have `global_ip.rs` file.

p2p/src/peer_manager/global_ip.rs
```diff
@@ -0,0 +1,146 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Based on libp2p sources:
+// https://github.com/libp2p/rust-libp2p/blob/73cbbe29679f5d56d81d5477d3796e82712f9ac2/protocols/autonat/src/behaviour.rs
+
+pub trait GlobalIp {
```
TheQuantumPhysicist, line 20: I think this trait should be called `IsGlobalIp`. This is because in its current form, it sounds like it's assuming that a type that implements it is a global ip, but it just adds the ability to check that property.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -131,11 +131,23 @@ impl<T: NetworkingService> PeerDb<T> {
 
     /// Checks if the given address is already connected.
     pub fn is_address_connected(&self, address: &T::Address) -> bool {
-        self.addresses.contains(address)
+        self.connected_addresses.contains(address)
     }
 
-    pub fn known_addresses(&self) -> Vec<PeerAddress> {
-        self.addresses.iter().map(TransportAddress::as_peer_address).collect()
+    pub fn random_known_addresses(&self, count: usize) -> Vec<T::Address> {
```
TheQuantumPhysicist, line 137: Please explain what these methods do and why they exist in a little snippet of documentation

p2p/src/peer_manager/tests/ban.rs
```diff
@@ -60,20 +60,12 @@ where
     )
     .await;
     let peer_id = peer_info.peer_id;
-    pm2.accept_inbound_connection(address, peer_info, None).unwrap();
+    pm2.accept_inbound_connection(address, peer_info, None).await.unwrap();
 
     assert_eq!(pm2.adjust_peer_score(peer_id, 1000).await, Ok(()));
```
TheQuantumPhysicist, line 65: I just noticed this number... I know it's not from you, Pavel. But why is this 1000? Seems arbitrary. Please throw in a TODO around it to check how to do this properly.

If I have to guess, someone did this like that because they want to just ban the peer no matter what. This doesn't seem to be the correct approach.

p2p/src/config.rs
```diff
@@ -71,4 +74,10 @@ pub struct P2pConfig {
     pub outbound_connection_timeout: OutboundConnectionTimeout,
     /// A node type.
     pub node_type: NodeTypeSetting,
+    /// Allow announcing and discovering local and private IPs. Should be used for testing only.
+    pub discover_private_ips: DiscoverPrivateIps,
```
TheQuantumPhysicist, line 78: allow_discover_private_ips

p2p/src/peer_manager/mod.rs
```diff
@@ -83,7 +91,13 @@ where
     /// Peer database
     peerdb: peerdb::PeerDb<T>,
 
+    /// Last time when heartbeat was called
     last_heartbeat: Instant,
+
+    /// All addresses that were announced to or from some peer.
+    /// Used to prevent infinity loops while broadcasting addresses.
+    /// Bitcoin Core uses bloom filter for that.
```
TheQuantumPhysicist, line 99: Why don't we do that?

p2p/src/peer_manager/mod.rs
```diff
@@ -123,20 +133,99 @@ where
...
+                    || *self.p2p_config.discover_private_ips
+            }
+            PeerAddress::Ip6(socket) => {
+                std::net::Ipv6Addr::from(socket.ip).is_global_unicast_ip()
+                    || *self.p2p_config.discover_private_ips
+            }
+        }
+    }
+
+    /// Discover public addresses for this node after new outbound connection is made
```
TheQuantumPhysicist, line 150: connections are made

p2p/src/types/peer_address.rs
```diff
@@ -35,7 +35,7 @@ pub struct PeerAddressIp6 {
 ///
 /// Same as std::net::SocketAddr for now but can be later extended with other address types.
 /// Use custom type to be able implement Encode and Decode.
-#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, Hash)]
```
TheQuantumPhysicist, line 38: As I always say, please do not use `Hash` trait as it's easily confused with other Hash things in our software. For maps, just use BTreeMap and implement `Ord` instead.

p2p/src/peer_manager/mod.rs
```diff
@@ -123,20 +133,99 @@ where
...
     /// connection from a remote peer or a response to an outbound connection that was initiated
     /// by the node as result of the peer manager maintenance.
-    fn accept_connection(
+    async fn accept_connection(
         &mut self,
         address: T::Address,
         role: Role,
         info: PeerInfo<T::PeerId>,
-        _receiver_address: Option<PeerAddress>,
+        receiver_address: Option<PeerAddress>,
```
TheQuantumPhysicist, line 227: Why is this an Option? Is it because of Inbound/Outbound? In that case, maybe we should put that in the Role enum. I'm not sure. But keeping variables that represent unexplainable states as Option is not really good. Another non-invasive option is, there could be a `ReceiverAddress` enum, which can have two possible values, Inbound(PeerAddress), and Outbound with no associated value. This way, the reader will immediately understand that this doesn't have a value because it's outgoing.

p2p/src/peer_manager/mod.rs
```diff
@@ -162,6 +251,19 @@ where
             P2pError::PeerError(PeerError::PeerAlreadyExists),
         );
 
+        if let (Some(receiver_address), Role::Outbound) = (receiver_address, role) {
+            self.handle_outbound_receiver_address(peer_id, receiver_address).await?;
+        }
+
+        if role == Role::Outbound {
```
TheQuantumPhysicist, line 258: This place doesn't seem right. I'm trying to understand what's going on here. We should not make any communication before having completed the handshake. Is this done after the completion of the handshake and confirming that we can communicate with the peer? I see `validate_version` above, but I'm not sure about the order of things in both cases of inbound and outbound.

TheQuantumPhysicist (approved):
In general this looks OK. Please address the few issues I mentioned. Feel free to ping me with a DM to discuss details.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Sure, done.

p2p/src/config.rs
pavel-kokolemin: Updated.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: Let's please do this later. I added TODO.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: Sorry, but are you sure? It's called every time when new connection is established. So I think "connections are made" will be wrong here. I used "addresses" because node may be listening on multiple ports so multiple addresses will be advertised.
Actually it just occurred to me that we should not try to change the port if node is already listening on a global IP (e.g. if a user has specified it from the command line). I will update this next time.

p2p/src/types/peer_address.rs
pavel-kokolemin: This is the first time I have heard about it. But I think banning `Hash` trait from stdlib just because it might be confusing is too rough. Anyway, I removed it because `Hash` is not actually needed currently.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: No, it's in `Option` only because remote peers might decide not send it in handshake:

```
pub enum HandshakeMessage {
    Hello {
    ...
        /// Socket address of the remote peer as seen by this node (addr_you in bitcoin)
        receiver_address: Option<PeerAddress>,
    },
    HelloAck {
    ...
        /// Socket address of the remote peer as seen by this node (addr_you in bitcoin)
        receiver_address: Option<PeerAddress>,
    },
}
```

So there is no difference in inbound and outbound peers on a protocol level.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: Handshake exchange happens earlier in a different place (when peer event loop is started, for both inbound and outbound peers).

p2p/src/peer_manager/mod.rs
TheQuantumPhysicist: So it's optional? Peers can choose not to send it? Weird.

p2p/src/peer_manager/mod.rs
TheQuantumPhysicist: Alrighty.

p2p/src/peer_manager/peerdb/mod.rs
TheQuantumPhysicist: Also the other method, please...

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -131,11 +131,25 @@ impl<T: NetworkingService> PeerDb<T> {
...
-        self.addresses.contains(address)
+        self.connected_addresses.contains(address)
     }
 
-    pub fn known_addresses(&self) -> Vec<PeerAddress> {
-        self.addresses.iter().map(TransportAddress::as_peer_address).collect()
+    /// Selects requested count of peer addresses from the DB randomly.
+    /// Result could be shared with remote peers over network.
+    pub fn random_known_addresses(&self, count: usize) -> Vec<T::Address> {
+        // TODO: Use something more efficient
```
TheQuantumPhysicist, line 140: How about this one? https://stackoverflow.com/a/42272866/1317944
Please make sure it works when the number required is less than the one available.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: Fixed, thanks :smile:

p2p/src/types/peer_address.rs
TheQuantumPhysicist: This is a discussion I'm more than happy to have with the team. Anyway, I think hash maps are overused anyway and people treat them as silver bullets for performance. It doesn't matter that much and people cannot prove the difference except with benchmarks. Who cares.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Updated, thanks.
We still copy all addresses into a `Vec`, so it's still not very good. But I expect that this part will change, so I think we should not spend too much time on it now.
Requesting more than available should be fine, `choose_multiple` will update the number internally as needed:
```
let amount = ::core::cmp::min(amount, self.len());
```

p2p/src/peer_manager/mod.rs
pavel-kokolemin: If someone is behind a proxy, sending the TCP address of the remote peer would be wrong (although using proxies is not currently supported).

TheQuantumPhysicist: This actually reminds me that we should add support for SOCKS proxy. I hope our design doesn't make it so hard.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Added, thanks.

p2p/src/peer_manager/peerdb/mod.rs
TheQuantumPhysicist: You don't need to copy all addresses. That's the whole point. You should only copy the addresses that are picked randomly by `choose_multiple`. You're currently cloning them twice.

pavel-kokolemin: Yes, but `choose_multiple` only works with slices and `known_addresses` is a `HashSet`.
I think we can take into account that entries already randomized in `HashSet` and simply do something like this:

```
self.known_addresses.iter().skip(some_random_number).take(count).cloned().collect::<Vec<_>>()
```

p2p/src/peer_manager/mod.rs
pavel-kokolemin: OK, I will take a note.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -131,11 +131,28 @@ impl<T: NetworkingService> PeerDb<T> {
...
+    pub fn random_known_addresses(&self, count: usize) -> Vec<T::Address> {
+        // TODO: Use something more efficient (without iterating over the all addresses first)
+        let all_addresses = self.known_addresses.iter().cloned().collect::<Vec<_>>();
+        all_addresses
+            .choose_multiple(&mut make_pseudo_rng(), count)
+            .cloned()
+            .collect::<Vec<_>>()
+    }
+
+    /// Selects requested count of connected peer ids randomly.
```
TheQuantumPhysicist, line 148: This isn't really what I meant. Please add what purpose this serves.

TheQuantumPhysicist (approved):

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Updated one more time.

PR #648, opened 2023-01-24
Support lmdb database map resize with minimal overhead

TheQuantumPhysicist:
Replaces #558


iljakuklic (approved):

iljakuklic (approved):
Looks fine. Some suggestions below

storage/lmdb/src/lib.rs
```diff
@@ -104,19 +104,46 @@ impl backend::WriteOps for DbTx<'_, lmdb::RwTransaction<'_>> {
...
+    }
+    err
+}
+
+fn schedule_map_resize_if_map_full(backend: &LmdbImpl, err: lmdb::Error) -> lmdb::Error {
+    if err == lmdb::Error::MapFull {
+        backend.schedule_map_resize();
+    }
+    err
+}
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Residue from the dev process. I moved them in there.

storage/lmdb/src/lib.rs
```diff
@@ -67,8 +67,8 @@ impl<'tx, C: lmdb::Cursor<'tx>> Iterator for PrefixIter<'tx, C> {
 
 pub struct DbTx<'m, Tx> {
     tx: Tx,
+    backend: &'m LmdbImpl,
     dbs: &'m DbList,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

storage/lmdb/src/lib.rs
```diff
@@ -165,58 +204,43 @@ impl<'tx> TransactionalRo<'tx> for LmdbImpl {
...
-    /// Amount of space needed for a transaction by default
-    pub const DEFAULT_TX_SIZE: MemSize = MemSize::from_megabytes(8);
-
     /// New LMDB database backend
-    pub fn new(path: PathBuf) -> Self {
+    pub fn new(
+        path: PathBuf,
+        inital_map_size: Option<usize>,
+        resize_settings: DatabaseResizeSettings,
+        resize_callback: Option<Box<dyn Fn(DatabaseResizeInfo)>>,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I thought it'd be an overkill... but I guess it's not. Here we go, all done except for transaction_rw.

storage/lmdb/src/initial_map_size.rs
```diff
@@ -0,0 +1,46 @@
...
+impl InitialMapSize {
+    pub fn into_memsize(&self) -> MemSize {
+        self.initial_map_size
+    }
+}
+
+impl Default for InitialMapSize {
+    fn default() -> Self {
+        Self {
+            initial_map_size: MemSize::from_bytes(0),
```
iljakuklic, line 31: Is this correct to pass 0 as the initial size? I think to get the original behaviour, `InitialMapSize` would contain and `Option<MemSize>` and default would be `None`. That way we are left with whatever LMDB does by default when `InitialMapSize::default` is used. I don't know whether 0 behaves the same or not.

storage/lmdb/src/resize_callback.rs
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+use lmdb::DatabaseResizeInfo;
+
+pub type MapResizeCallbackFn = dyn Fn(DatabaseResizeInfo);
+
+/// A function wrapper that represents the callback when the map within lmdb resizes
+#[derive(Default)]
+pub struct MapResizeCallback {
+    f: Option<Box<MapResizeCallbackFn>>,
```
iljakuklic, line 23: Opposite suggestion to the previous one :laughing:... How about getting rid of the `Option` and just pass in a function that does nothing in the `default` constructor?

storage/lmdb/src/memsize.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+impl std::ops::Mul<MemSize> for u64 {
+    type Output = MemSize;
+
+    fn mul(self, rhs: MemSize) -> Self::Output {
+        MemSize(self * rhs.0)
+    }
+}
```
iljakuklic, line 75: Are these operations used? I introduced them to calculate the new memmap size after reallocation but that is now done by the upstream lmdb crate. Maybe we can get rid of them.

storage/lmdb/src/memsize.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// Get raw byte count in native representation
+    pub fn as_bytes(self) -> u64 {
+        self.0
+    }
+
+    /// Division, rounding up
+    pub fn div_ceil(self, rhs: Self) -> u64 {
+        // TODO: Use u64::div_ceil once stable
+        self.0 / rhs.0 + (self.0 % rhs.0 > 0) as u64
+    }
```
iljakuklic, line 52: Same here, probably not needed?

storage/lmdb/src/memsize.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// Get raw byte count as u64
+    pub fn as_bytes_u64(self) -> u64 {
+        self.0
+    }
+
+    /// Get raw byte count in native representation
+    pub fn as_bytes(self) -> u64 {
+        self.0
+    }
```
iljakuklic, line 46: After changing the representation to `u64`, these are now duplicate.

iljakuklic (commented):
Looking good. Just a few minor nitpicks below:

storage/lmdb/src/resize_callback.rs
TheQuantumPhysicist: This all depends on your assumptions... I agree with you on the other one, but this one maps better with the Option because it calls nothing (even an empty function) when it's empty... there's an argument to be made that calling nothing is nothing... but again, all about the assumptions. In this one, I'm mapping 1:1 with what we have in lmdb.

storage/lmdb/src/initial_map_size.rs
TheQuantumPhysicist: I'll go with None. I think that's more reasonable.

Zero in lmdb means "nothing will happen"... it'll basically be ignored. But I guess you're right it's up to lmdb.

TheQuantumPhysicist: Heads up on the ugliness because of this... you'll see why I chose this.

storage/lmdb/src/memsize.rs
TheQuantumPhysicist: I thought they could become useful. I removed all the extras.

storage/lmdb/src/resize_callback.rs
iljakuklic: Ok, I don't feel very strongly about this.

iljakuklic (approved):

PR #649, opened 2023-01-25
Sqlite integration

erasmospunk:


storage/sqlite/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+
+[dependencies]
+logging = { path = '../../logging' }
+storage-core = { path = '../core' }
+utils = { path = '../../utils' }
+
+ouroboros = "0.15"
```
iljakuklic, line 12: This is no longer needed as far a I can tell

PR #650, opened 2023-01-25
Make PeerDb persistent

pavel-kokolemin:


p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -90,11 +93,16 @@ pub struct PeerDb<T: NetworkingService> {
...
+    banned_addresses: BTreeMap<T::BannableAddress, Duration>,
+
+    store: Box<dyn PeerDbStore>,
 }
 
 impl<T: NetworkingService> PeerDb<T> {
-    pub fn new(p2p_config: Arc<config::P2pConfig>) -> crate::Result<Self> {
+    pub fn new(
+        p2p_config: Arc<config::P2pConfig>,
+        store: Box<dyn PeerDbStore>,
```
TheQuantumPhysicist, line 104: Why use dyn when you can use generics?

p2p/src/peer_manager/peerdb/store.rs
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::time::Duration;
+
+pub trait PeerDbStore: Send {
```
TheQuantumPhysicist, line 18: This interface is overly simplistic. It doesn't support transactions too. I see that in the implementation you have transactions all over. There should be a way to start transaction through this interface, then use that to do these operations.

p2p/src/lib.rs
```diff
@@ -159,13 +162,15 @@ pub async fn make_p2p(
     let stream_adapter = NoiseEncryptionAdapter::gen_new();
     let base_transport = net::default_backend::transport::TcpTransportSocket::new();
     let transport = NoiseTcpTransport::new(stream_adapter, base_transport);
+    let peerdb_store = Box::new(PeerDbStoreImpl::new(storage::inmemory::InMemory::new())?);
```
TheQuantumPhysicist, line 165: This should be initialized in a way similar to how chainstate storage is initialized, where it can depend on the data dir.

TheQuantumPhysicist (changes requested):
Great effort. Some tweaks are required to make this right. Please see the comments.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Is it OK now? I still think it was better with `dyn`.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -71,9 +76,7 @@ impl<T: NetworkingService> From<&PeerContext<T>> for ConnectedPeer {
 }
 
 // TODO: Store available addresses in a binary heap (sorting by their availability).
```
TheQuantumPhysicist, line 78: I'm not sure why you're using the word "heap" here... a heap is a specific data structure. It does support some kind of order, but usually it solves niche problems. A BTreeMap is more than enough to provide some order.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -71,9 +76,7 @@ impl<T: NetworkingService> From<&PeerContext<T>> for ConnectedPeer {
...
-// TODO: Find a way to persist this data in some database for when the node is restarted
-// (banned, available, and at-least-once used should be restored)
-pub struct PeerDb<T: NetworkingService> {
+pub struct PeerDb<T: NetworkingService, P> {
```
TheQuantumPhysicist, line 79: I'm very sorry for this dumb comment... but can you please use the letter S for store, like we do everywhere?

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -162,14 +184,21 @@ impl<T: NetworkingService> PeerDb<T> {
 
     /// Checks if the given address is banned.
     pub fn is_address_banned(&mut self, address: &T::BannableAddress) -> bool {
-        if let Some(banned_till) = self.banned.get(address) {
+        if let Some(banned_till) = self.banned_addresses.get(address) {
             // Check if the ban has expired.
             let now = SystemTime::now()
```
TheQuantumPhysicist, line 189: I just noticed this now. We never do that. We never pass time directly from the system. Otherwise we can have huge conflicts in tests.

Whether in this PR or another one, please use a `TimeGetter` object passed from the P2P object to know what time it's.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: These words were already there. I am also not sure why use "heap", but I expect this part to change so I have not spent much time looking into it.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -259,7 +293,14 @@ impl<T: NetworkingService> PeerDb<T> {
                 // This can fail only if `SystemTime::now()` returns the time before `UNIX_EPOCH`.
```
TheQuantumPhysicist, line 293: OK I can see there's an infection of SystemTime all over the place...

p2p/src/peer_manager/peerdb/store.rs
pavel-kokolemin: Transactions added. I used chainstate storage as a base, but simplified it (to avoid using macros).
`PeerManager` will now stop working if a storage error is detected (for example if a commit fails).

p2p/src/lib.rs
pavel-kokolemin: Should be done. Default is the LMDB backend (until the Sqlite backend is ready) and unit tests must specify InMemory backend explicitly. Otherwise tests might interfere with each other.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Sure, updated.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: OK, will do. But let's please do this in different PR.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: I will change that.

pavel-kokolemin:
There were too many changes, so I squashed and rebased old commits.

iljakuklic (approved):

p2p/src/lib.rs
TheQuantumPhysicist: There's a misunderstanding here. There's no default other than what you provide. Obviously we shouldn't use lmdb for tests, but in runner.rs, this should use lmdb. In unit tests, we pass inmemory. In functional tests, lmdb will be used. Am I missing something?

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

p2p/src/lib.rs
pavel-kokolemin: I updated that part. Should be OK now, thanks.

node/src/runner.rs
```diff
@@ -71,13 +71,18 @@ pub async fn initialize(
     });
 
     // P2P subsystem
+    // TODO: Replace Lmdb with Sqlite backend when it's read
```
TheQuantumPhysicist, line 74: ready

TheQuantumPhysicist (approved):

node/src/runner.rs
pavel-kokolemin: Thanks, I will fix it next time (don't want for CI).

PR #652, opened 2023-01-25
Return Cow<[u8]> instead of &[u8] in storage abstraction layer

TheQuantumPhysicist:
Generally speaking, it's not possible to get a slice of bytes from all storage backends with arbitrary lifetime. This is because this assumes that all the data is readily available for reading directly from the database. This is more of a wish.

In this PR we return `Cow<[u8]>` instead of `&[u8]` from the storage interface's get() function. 

Using an owned value would be the last resort if a storage layer does not provide solutions for getting values by reference.

### DIFF
--- a/storage/inmemory/src/lib.rs
+++ b/storage/inmemory/src/lib.rs
@@ -15,7 +15,7 @@
 
 use storage_core::{adaptor, backend, util, Data, DbDesc, DbIndex};
 
-use std::collections::BTreeMap;
+use std::{borrow::Cow, collections::BTreeMap};
 
 type Map = BTreeMap<Data, Data>;
 
@@ -32,8 +32,8 @@ impl<'i> Iterator for PrefixIter<'i> {
 pub struct StorageMaps(Vec<Map>);
 
 impl backend::ReadOps for StorageMaps {
-    fn get(&self, idx: DbIndex, key: &[u8]) -> storage_core::Result<Option<&[u8]>> {
-        Ok(self.0[idx.get()].get(key).map(AsRef::as_ref))
+    fn get(&self, idx: DbIndex, key: &[u8]) -> storage_core::Result<Option<Cow<[u8]>>> {
+        Ok(self.0[idx.get()].get(key).map(|p| p[..].into()))
     }
 }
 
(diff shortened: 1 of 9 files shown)


storage/inmemory/src/lib.rs
```diff
@@ -32,8 +32,8 @@ impl<'i> Iterator for PrefixIter<'i> {
 pub struct StorageMaps(Vec<Map>);
 
 impl backend::ReadOps for StorageMaps {
-    fn get(&self, idx: DbIndex, key: &[u8]) -> storage_core::Result<Option<&[u8]>> {
-        Ok(self.0[idx.get()].get(key).map(AsRef::as_ref))
+    fn get(&self, idx: DbIndex, key: &[u8]) -> storage_core::Result<Option<Cow<[u8]>>> {
+        Ok(self.0[idx.get()].get(key).map(|p| p[..].into()))
```
stanislav-tkach, line 36:
```suggestion
        Ok(self.0.get(idx.get()).and_then(|m| m.get(key).map(|p| p.into())))
```
Am I missing the point of `[..]` here? And wouldn't it be better to to use the `get` method instead of potentially panicking `[]`?

stanislav-tkach (approved):

storage/inmemory/src/lib.rs
iljakuklic: The key-value maps should be initialised at the beginning, and the storage layer statically enforces the existence. Panicking is the right thing to do here, otherwise we mask potential issues with the backend not maintaining invariants properly.

stanislav-tkach: @iljakuklic Got it!

stanislav-tkach, line 36:
```suggestion
        Ok(self.0[idx.get()].get(key).map(|p| p.into()))
```

TheQuantumPhysicist: `[..]` doesn't panic unless you dereference the slice coming out of it incorrectly. The primitive `[..]` is an idiomatic way to convert something to a slice. Even if the size is zero, it won't panic.

The reason I chose `[..]` over as_ref() is because there's type inference with as_ref(), which makes into() fail. So we have to convert to a slice that infers the type.

TheQuantumPhysicist: Done

iljakuklic (commented):
Looks good.

PR #653, opened 2023-01-27
Fix new clippy lints (rust 1.67)

stanislav-tkach:
By the way, there is a bunch of `eprintln` in `mempool`. Should these be replaced by proper logging?


TheQuantumPhysicist (approved):

iljakuklic (approved):

iljakuklic:
> By the way, there is a bunch of `eprintln` in `mempool`. Should these be replaced by proper logging?

These can probably be just deleted

PR #657, opened 2023-01-27
Workspace dependencies

stanislav-tkach:
This is mostly for discussion. I think that it is convenient to specify the versions of widely used dependencies in the root toml file. It is even more important for things such as `parity-scale-codec`, even though we have a CI check.


pavel-kokolemin (approved):

stanislav-tkach:
Yes, it was added in the 1.64 [version](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds).

It is convenient to specify the version only once, but I'm not sure when exactly a dependency should be moved to the workspace file. For now I have moved the dependencies that are used in 4+ crates.

TheQuantumPhysicist (approved):

PR #659, opened 2023-01-30
Add ping messages

pavel-kokolemin:
This PR will add a check that peers are online and responding in time. This should make p2p more robust and prevent some problems caused by malicious/slow clients (such as unlimited send buffer growth).
Ping requests are sent to all connected peers every 60 seconds, and if no response is received after 150 seconds, the peer is disconnected. A duplicate request is sent until a response is received.
I have moved some functionality from PeerDb to PeerManager. I think PeerDb should be only deal with peer addresses (so it would be more like AddrMan in Bitcoin Core).


p2p/backend-test-suite/src/peer_manager_peerdb.rs
```diff
@@ -20,124 +20,42 @@ use std::{sync::Arc, time::Duration};
...
     peer_manager::peerdb::{
-        storage::{PeerDbStorage, PeerDbStorageRead, PeerDbTransactional},
+        storage::{PeerDbStorageRead, PeerDbTransactional},
         PeerDb,
     },
-    testing_utils::{peerdb_inmemory_store, RandomAddressMaker},
+    testing_utils::{peerdb_inmemory_store, P2pTestTimeGetter, RandomAddressMaker},
 };
 
-tests![
```
TheQuantumPhysicist, line 35: Where did these tests go?

pavel-kokolemin: These two specific tests have been removed. Not much is done there, and `adjust_peer_score` calls (below and above threshold) are tested in `p2p/src/peer_manager/tests/ban.rs`.

p2p/src/net/default_backend/mod.rs
```diff
@@ -60,6 +60,21 @@ pub struct ConnectivityHandle<S: NetworkingService, T: TransportSocket> {
     _marker: PhantomData<fn() -> S>,
```
TheQuantumPhysicist, line 60: I just noticed this... why not `PhantomData<S>`? It has the same variance.

pavel-kokolemin: I thought about changing it, but decided that `fn()` is there for a reason :smile: 
I also removed it from a few other places too.

TheQuantumPhysicist: Maybe there's a reason I'm not remembering... I vaguely am now recalling there's a reason other than variance... but can't exactly remember.

p2p/src/peer_manager/mod.rs
```diff
@@ -313,26 +350,37 @@ where
...
             return Err(P2pError::PeerError(PeerError::TooManyPeers));
         }
 
         self.accept_connection(address, Role::Inbound, info, receiver_address).await
     }
 
-    /// Close connection to a remote peer
+    /// Connection to a remote peer closed
     ///
     /// The decision to close the connection is made either by the user via RPC
```
TheQuantumPhysicist, line 362: This is kind of a heavy promise to make in a comment... I'm confused. This is a state-machine hell.

pavel-kokolemin: Sorry, my bad, I forgot to update the comment here.
I think the connect/disconnect flow was broken before.
`PeerManager` will now update itself only from events from the backend.
Connect and disconnect flows are consistent and simple now:
got connect/disconnect request => sent request to backend => received event from backend => state updated.
For example, this should work correctly if `PeerManager` sends a disconnect request but the peer drops own connection before backend have a chance to process the request.

p2p/src/peer_manager/mod.rs
```diff
@@ -313,26 +350,36 @@ where
...
-    /// or by the [`PeerManager::heartbeat()`] function which has decided to cull
-    /// this connection in favor of another potential connection.
-    fn close_connection(&mut self, peer_id: T::PeerId) -> crate::Result<()> {
+    /// This can happen when the remote peer has dropped its connection
+    /// or if a disconnect request has been sent by PeerManager to the backend.
+    fn connection_closed(&mut self, peer_id: T::PeerId) -> crate::Result<()> {
         // The backend is always sending ConnectionClosed event when somebody disconnects, ensure that the peer is active
-        if self.peerdb.is_active_peer(&peer_id) {
-            log::debug!("connection closed for peer {peer_id}");
+        if self.is_active_peer(&peer_id) {
```
TheQuantumPhysicist, line 366: I'm not sure I understand the specific reason for this logic. You check if a peer exists, then remove it, then expect if it's not there... why not do something like `if let Some(peer) = self.peers.remove(peer_id)`? This looks ugly due to the statefulness of it, but then this is what's practically being done. This should be cleaned up a little, I think.

p2p/src/peer_manager/mod.rs
```diff
@@ -690,13 +812,80 @@ where
...
+        self.peers.values().map(Into::into).collect()
+    }
+
+    /// Selects requested count of connected peer ids randomly.
+    ///
+    /// It can be used to distribute data in the gossip protocol
+    /// (for example, to relay announced addresses to a small group of peers).
+    pub fn random_peer_ids(&self, count: usize) -> Vec<T::PeerId> {
+        // There are normally not many connected peers, so iterating over the whole list should be OK
+        let all_peer_ids = self.peers.keys().cloned().collect::<Vec<_>>();
```
TheQuantumPhysicist, line 831: Please throw in a TODO here to optimize this. Always do this when we have to unnecessarily clone instead of pick random items. And btw, I really think implementing this is not that hard, but I'm not gonna pressure for it now.

p2p/src/peer_manager/mod.rs
```diff
@@ -68,6 +73,11 @@ const PEER_MGR_HEARTBEAT_INTERVAL_MIN: Duration = Duration::from_secs(5);
 /// Upper bound for how often [`PeerManager::heartbeat()`] is called
 const PEER_MGR_HEARTBEAT_INTERVAL_MAX: Duration = Duration::from_secs(30);
 
+/// How often send ping requests to peers
+const PEER_MGR_PING_CHECK_PERIOD: Duration = Duration::from_secs(60);
+/// When a peer is detected as dead and disconnected
+const PEER_MGR_PING_TIMEOUT: Duration = Duration::from_secs(150);
```
TheQuantumPhysicist, line 79: This should be controllable through command line arguments in the future. Maybe not necessarily now. But this is definitely going into functional tests in the future.

TheQuantumPhysicist (approved):

p2p/src/net/default_backend/mod.rs
pavel-kokolemin: @iljakuklic Do you remember why `PhantomData<fn() -> S>` was used instead `PhantomData<S>`?
The code seems to build normally with the later.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: Sure, updated, thanks.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: Sure, done.

p2p/src/peer_manager/mod.rs
```diff
@@ -690,13 +812,80 @@ where
...
+        let all_peer_ids = self.peers.keys().cloned().collect::<Vec<_>>();
+        all_peer_ids
+            .choose_multiple(&mut make_pseudo_rng(), count)
+            .cloned()
+            .collect::<Vec<_>>()
+    }
+
+    /// Checks if the peer is active
+    pub fn is_active_peer(&self, peer_id: &T::PeerId) -> bool {
+        self.peers.get(peer_id).is_some()
```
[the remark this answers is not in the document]
pavel-kokolemin: Good question. Probably the name `is_active_peer` is no longer appropriate. I have renamed the function to `is_peer_connected` (returns if the peer with the specific `peer_id` is currently connected).

p2p/src/peer_manager/mod.rs
pavel-kokolemin: OK. I added command line arguments.

pavel-kokolemin:
I noticed a recent regression in the command line handling. The node won't print anything to stderr if start fails. I have fixed the problem here (reverted like it was before):

```
 #[tokio::main]
 async fn main() {
     run().await.unwrap_or_else(|err| {
-        logging::log::error!("Mintlayer node launch failed: {err:?}");
+        eprintln!("Mintlayer node launch failed: {err:?}");
         std::process::exit(1)
     })
 }
```

p2p/src/net/default_backend/mod.rs
pavel-kokolemin: I reverted my changes.

p2p/src/peer_manager/mod.rs
TheQuantumPhysicist: Optimize. We use American English, not British :-)

p2p/src/peer_manager/tests/ping.rs
```diff
@@ -0,0 +1,135 @@
...
+                    })
+                    .await
+                    .unwrap();
+            }
+            _ => panic!("unexpected event: {event:?}"),
+        }
+    }
+
+    // Receive one more ping request but do not send a ping response
+    time_getter.advance_time(ping_check_period).await;
```
TheQuantumPhysicist, line 113: Btw, I forgot to say: Thank you for using TimeGetter and not "sleeping" to ensure that the test works as expected. The same should be done for all the other flakey tests that depend on waiting periods.

PR #662, opened 2023-01-31
Use unbounded channels in default p2p backend

pavel-kokolemin:
Using unbounded channels should fix the problem with potential deadlocks in p2p (when two channels are full).
I have removed a lot of async annotations that are no longer needed.

I think to apply back-pressure on the connected peers we could do something like this: count in `p2p::net::default_backend::peer::Peer::run` how many unprocessed requests  there are and if more than some limit then stop reading from the network socket, until some responses have been processed and sent back. I think this should solve the node overload problem nicely. It may look like this:

```
response = self.cmd_rx.recv() {
    // Request was processed, send response to the peer...
    request_count -= 1;
},
// Read from the socket if only there is some free capacity
request = self.socket.recv(), if request_count < 10 => {
    // New request received, send it to backend...
    request_count += 1;
}
```

We will need to ensure that for each request received, the backend would send a response. But I suppose it should be easy enough to do.

### DIFF
--- a/p2p/src/net/default_backend/backend.rs
+++ b/p2p/src/net/default_backend/backend.rs
@@ -22,11 +22,7 @@ use std::{
     sync::Arc,
 };
 
-use futures::{
-    future::{join_all, BoxFuture},
-    stream::FuturesUnordered,
-    FutureExt, StreamExt, TryFutureExt,
-};
+use futures::{future::BoxFuture, stream::FuturesUnordered, FutureExt, StreamExt};
 use tokio::{sync::mpsc, time::timeout};
 
 use common::chain::ChainConfig;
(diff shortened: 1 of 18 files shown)


p2p/src/net/default_backend/backend.rs
```diff
@@ -166,26 +169,27 @@ where
...
-        }
+
+        let peer = self
+            .peers
+            .remove(peer_id)
+            .ok_or(P2pError::PeerError(PeerError::PeerDoesntExist))?;
+
+        peer.tx.send(Event::Disconnect).map_err(P2pError::from)?;
+
+        Ok(())
```
iljakuklic, line 188: Could it be just this?

```suggestion
        peer.tx.send(Event::Disconnect).map_err(P2pError::from)
```

There are quite a few occurrences of this pattern all over the place but not sure my proposal is much better. Not a big difference.

iljakuklic (approved):
Nice to see a bunch of `async`/`await` go away

stanislav-tkach (approved):

TheQuantumPhysicist:
I'm not fully convinced this is needed. We should talk about this. Let's have a discussion about it tomorrow.

p2p/src/net/default_backend/backend.rs
pavel-kokolemin: Sure, updated, thanks.

TheQuantumPhysicist (approved):

PR #665, opened 2023-02-01
Add a simple monitoring channel

pavel-kokolemin:
We recently switched to unbounded channels in p2p.
@TheQuantumPhysicist asked to print to the log when channels get too big to see the problems. I have added a simple tokio channel wrapper. The check is done in a background tokio task.

### DIFF
--- a/p2p/src/utils/monitored_channel.rs
+++ b/p2p/src/utils/monitored_channel.rs
@@ -1,0 +1,154 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


p2p/src/utils/monitored_channel.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+pub struct UnboundedReceiver<T> {
+    receiver: mpsc::UnboundedReceiver<T>,
+    recv_count: Arc<AtomicU64>,
+}
+
+impl<T> UnboundedSender<T> {
+    pub fn send(&self, message: T) -> Result<(), SendError<T>> {
+        // Incrementing the counter if sending fails is OK
```
TheQuantumPhysicist, line 55: Why?

Isn't it better to match on send()?

pavel-kokolemin:
I will close the PR for now. Later I will update the PR to work without spawning new tasks.

PR #666, opened 2023-02-01
Unify DeltaMergeUndo and DataMergeUndo

### LINKED ISSUE #654: Unify DeltaMergeUndo and DataMergeUndo
It seems that types `pos_accouting::DeltaMergeUndo` and `pos_accounting::DataMergeUndo` can be unified and represented as a delta. This will simplify code and can be used to return a single undo type from `batch_write_delta`.

Also, the same unification can be done for `PoSAccountingUndo`. I'd say that this enum is completely unnecessary and `DeltaMergeUndo` or a similar type can be used everywhere.  

In the end, all operations are merges of deltas, so no need to manage 3 different types.

azarovh:
1. The main purpose of this PR is to return undo object from `FlushablePoSAccountingView::batch_write_delta`, which I want to use in #647 . To do that it needs a single undo object to return because both `PoSAccountingDB` and `PoSAccountingDelta` can flush. I reimplemented `PoSAccountingDB::merge_with_delta` to utilize Delta as an undo object, where previously undo was done by setting previous state.
2. While doing step 1 I also simplified the code in `accounting` by removing `DeltaMapElement` enum. It played intermediary role but I don't see how it helps now after the last big refactoring. 

Closes #654

### DIFF
--- a/accounting/src/delta/combine.rs
+++ b/accounting/src/delta/combine.rs
@@ -15,18 +15,18 @@
 
 use common::primitives::{signed_amount::SignedAmount, Amount};
 
-use crate::{delta::delta_data_collection::DeltaMapElement, error::Error};
+use crate::{error::Error, DataDelta};
 
 /// Combine data with an element of `DeltaDataCollection`.
 /// An element can be either a Delta or a result of delta undo.
 pub fn combine_data_with_delta<T: Clone + Eq>(
     lhs: Option<T>,
-    rhs: Option<DeltaMapElement<T>>,
+    rhs: Option<DataDelta<T>>,
 ) -> Result<Option<T>, Error> {
     match (lhs, rhs) {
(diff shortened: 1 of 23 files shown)


TheQuantumPhysicist (approved):
Looks fine... I just hope we're not losing information with this "simplification"... I can't pinpoint to a specific issue, but I'd like to ask reviewers to double-check.

azarovh:
> Looks fine... I just hope we're not losing information with this "simplification"... I can't pinpoint to a specific issue, but I'd like to ask reviewers to double-check.

The only piece of information that is lost is how that undo was created: from merging to DB or to a Delta. But if we have a notion of unified  merging (via `FlushablePoSAccountingView`) I think it is logical to have unified undo as well

stanislav-tkach (approved):

PR #667, opened 2023-02-01
Integrate Sqlite backend

erasmospunk:
First unoptimised Sqlite integration. Some of the limitations are related to query construction and caching, the PrefixIter filters results with code rather an SQL query, the results are read all at once rather row by row.


storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    info::DbDesc,
+    Data, DbIndex,
+};
+use utils::shallow_clone::ShallowClone;
+use utils::sync::Arc;
+
+/// Return the table name that corresponds to a specific index
+/// TODO this is unoptimised and the returned values should be cached
+fn db_table_name(idx: usize) -> String {
+    format!("db_{}", idx)
```
TheQuantumPhysicist, line 39: Is it not possible to create a database name by just doing `idx.to_string()`?

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+
+impl Drop for DbTx<'_> {
+    fn drop(&mut self) {
+        if self.connection.is_autocommit() {
+            return;
+        }
+
+        let res = self.connection.execute("ROLLBACK TRANSACTION", ());
+        if let Err(err) = res {
+            println!("Error: transaction rollback failed: {}", err);
```
TheQuantumPhysicist, line 107: Please use logging.

TheQuantumPhysicist:
Looking good. Please try to resolve as many TODOs as you can and let's get this merged.

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+            .map_err(process_sqlite_error)?;
+        Ok(())
+    }
+}
+
+impl Drop for DbTx<'_> {
+    fn drop(&mut self) {
+        if self.connection.is_autocommit() {
+            return;
+        }
```
iljakuklic, line 103: Would the backend transaction mechanism behave correctly if autocommit is set to `true`? Maybe this should be an assertion rather than a check that silently handles that case.

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+    backend::{self, TransactionalRo, TransactionalRw},
+    info::DbDesc,
+    Data, DbIndex,
+};
+use utils::shallow_clone::ShallowClone;
+use utils::sync::Arc;
+
+/// Return the table name that corresponds to a specific index
+/// TODO this is unoptimised and the returned values should be cached
+fn db_table_name(idx: usize) -> String {
```
iljakuklic, line 38: Since this is only used with database indices, it would be more appropriate to take `DbIndex` rather than plain `usize` to make the types more precise. (Also spares you a bunch of `.get()` elsewhere)

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+            .prepare_cached(
+                format!(
+                    "SELECT key, value FROM {} ORDER BY key",
+                    db_table_name(idx.get())
+                )
+                .as_str(),
+            )
+            .map_err(process_sqlite_error)?;
+
+        let mut rows = stmt.query(()).map_err(process_sqlite_error)?;
```
iljakuklic, line 133: It may be more convenient to use [`Statement::query_map`](https://docs.rs/rusqlite/latest/rusqlite/struct.Statement.html#method.query_map) here, it supposedly gives a standard Rust iterator. Have you explored that option?

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+
+// Since this is a key-value store, we have a specific structure for the underlying databases,
+// where it conforms to map<index, map<key, value>>. The outer map size almost never needs to
+// change, and hence we fix it and we use it as a constant
+const MAX_DB_COUNT: usize = 64;
+const DB_NAMES: Lazy<Vec<String>> = Lazy::new(|| make_db_names(MAX_DB_COUNT));
+
+fn make_db_names(max_size: usize) -> Vec<String> {
+    (0..max_size).map(|i| format!("db_{}", i)).collect()
+}
```
iljakuklic, line 45: I am not convinced pre-calculating DB names into a global variable with an arbitrary upper bound like this is a good approach.

An alternative is to pre-calculate the names when the database is open. The information about the number of key-value maps is already available at that point. The result can be stored in `DbList` and accessed by indexing a `DbIndex` into it. It can stored in a struct together with connection so the transactions can access it.

The structure of `SqliteImpl` would then be something like (excuse the made up syntax):
```rust
Arc<{ connection: Mutex<sqlite::Connection>, db_names: Vec<String> }>
```

storage/sqlite/tests/backend.rs
```diff
@@ -0,0 +1,45 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use storage_sqlite::Sqlite;
+
+fn main() {
+    let test_root = test_utils::test_root!("backend-tests").unwrap();
```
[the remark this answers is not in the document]
iljakuklic: Correction: the `test_root!` macro already incorporates the crate name into the path, so we already have: `target/tmp/storage-{sqlite,lmdb}/backend-tests`. Resolving.

storage/sqlite/src/lib.rs
TheQuantumPhysicist: That sounds better. Let's go for it! 

@erasmospunk could you give this a shot? I think all you have to do is eliminate the global variable, the once_cell dependency, and put the implementation inside open_db(), and store the values in the struct. Please give it a shot.

storage/sqlite/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+
+[dependencies]
+logging = { path = '../../logging' }
+storage-core = { path = '../core' }
+utils = { path = '../../utils' }
+
+rusqlite = { version = "0.28.0", features = ["bundled"] }
```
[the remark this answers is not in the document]
erasmospunk: Changed it

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+#[derive(Eq, PartialEq, Debug, Clone)]
+struct DbList(Vec<()>);
+
+impl std::ops::Index<DbIndex> for DbList {
+    type Output = ();
+
+    fn index(&self, index: DbIndex) -> &Self::Output {
+        &self.0[index.get()]
+    }
+}
```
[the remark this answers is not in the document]
erasmospunk: Renamed and implemented as SqliteQueries

storage/sqlite/src/lib.rs
erasmospunk: This is needed otherwise ROLLBACK TRANSACTION might fail

storage/sqlite/src/lib.rs
erasmospunk: Removed this

storage/sqlite/src/lib.rs
erasmospunk: PrefixIter needs to be optimized so this will change

storage/sqlite/src/lib.rs
erasmospunk: Implemented your idea as `SqliteQueries`

storage/lmdb/tests/backend.rs
```diff
@@ -16,7 +16,7 @@
 use storage_lmdb::Lmdb;
 
 fn main() {
-    let test_root = test_utils::test_root!("backend-tests").unwrap();
+    let test_root = test_utils::test_root!("backend-tests/lmdb").unwrap();
```
iljakuklic, line 19: I later realised that the `test_root!` macro already incorporates the crate name into the path, so this change is not needed and can be reverted. Same in the sqlite backend. My mistake, apologies.

storage/sqlite/src/lib.rs
iljakuklic: I was thinking whether we should make sure at the beginning that autocommit is false, and have only an assertion here. Not preforming transaction rollback may leave the database in an unexpected state, so we should make effort to prevent it. But then again I am not sure what exactly the semantics of autocommit are, so maybe it's fine like this.

iljakuklic (approved):
Looks good. Please revert the test path changes (this is my mistake, reason pointed out below) and check what `autocommit` does so we know what we do with it is correct. Tests seem to pass, and rollbacks are tested so it may be fine but it still seems weird to omit a statement that closes the transaction.

storage/sqlite/src/queries.rs
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use storage_core::{DbDesc, DbIndex};
+
+/// Return the prefix of the database key/value tables
+#[inline]
+pub(crate) fn db_table_name(idx: usize) -> String {
```
iljakuklic, line 20: This is a private submodule, all the `pub(crate)` in this module can be just `pub` without exposing them externally as far as I can tell.

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+
+pub struct DbTx<'m> {
+    connection: MutexGuard<'m, Connection>,
+    queries: &'m SqliteQueries,
+}
+
+impl<'m> DbTx<'m> {
+    pub fn start_transaction(sqlite: &'m SqliteImpl) -> storage_core::Result<Self> {
```
iljakuklic, line 68: Why is this `pub`? As far as I can see it is only used internally by this module.

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+            .map_err(|_| storage_core::error::Recoverable::TemporarilyUnavailable)?;
+        let tx = DbTx {
+            connection,
+            queries: &sqlite.queries,
+        };
+        tx.connection.execute("BEGIN TRANSACTION", ()).map_err(process_sqlite_error)?;
+        Ok(tx)
+    }
+
+    pub fn commit_transaction(&self) -> storage_core::Result<()> {
```
iljakuklic, line 81: Same here, why `pub`?

iljakuklic (commented):
Just some more minor comments regarding visibility:

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+    connection: MutexGuard<'m, Connection>,
+    queries: &'m SqliteQueries,
+}
+
+impl<'m> DbTx<'m> {
+    fn start_transaction(sqlite: &'m SqliteImpl) -> storage_core::Result<Self> {
+        let connection: MutexGuard<Connection> = sqlite
+            .connection
+            .lock()
+            .map_err(|_| storage_core::error::Recoverable::TemporarilyUnavailable)?;
```
TheQuantumPhysicist, line 72: I think this is incorrect. Failing to lock a mutex is not a recoverable error. It should either block or there's a mutex poisoning which is practically a dead-lock.

If you want this to be recoverable, then you should use `try_lock()` instead of `lock()`, and then see what the error is.

erasmospunk: You are right, I changed the error to Fatal.

storage/sqlite/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub struct SqliteImpl {
+    /// Handle to an Sqlite database connection
+    connection: Arc<Mutex<Connection>>,
+
+    /// List of sql queries
+    queries: SqliteQueries,
```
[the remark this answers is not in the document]
iljakuklic: Minor nitpick: Both `queries` and `connection` can just share one `Arc`, since they are always cloned together. One reference counter is enough. It does however mean introducing an intermediate struct to keep it readable so I don't mind either way.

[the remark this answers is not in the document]
erasmospunk: We could do that. Something like [this](https://github.com/mintlayer/mintlayer-core/commit/dce92f9bb71f79be6161a4d34302efd707df878b)?

iljakuklic (approved):

PR #668, opened 2023-02-02
Update channel type used in `PeerManagerEvent`

pavel-kokolemin:
I think using the `oneshot` channel in `PeerManagerEvent` is not very safe.
It's easy to imagine a situation where the receiver is dropped before the event is processed.
For example here:

```
#[async_trait::async_trait]
impl<T> P2pInterface for P2p<T>
where
    T: NetworkingService,
{
    async fn connect(&mut self, addr: String) -> crate::Result<()> {
        let (tx, rx) = oneshot_nofail::channel();
        let addr = addr
            .parse::<T::Address>()
            .map_err(|_| P2pError::ConversionError(ConversionError::InvalidAddress(addr)))?;
        self.tx_peer_manager
            .send(PeerManagerEvent::Connect(addr, tx))
            .map_err(|_| P2pError::ChannelClosed)?;
        rx.await.map_err(P2pError::from)?
    }
...    
```

Here `connect` is async and so can be cancelled at any `await` point (I think we are not doing this now, but this could easily change later).
And yet, sending to a closed channel in `PeerManager` is considered a fatal error and `PeerManager` would stop working.
One fix would be to review all `oneshot` usages and fix them accordingly. However, I decided to write a small wrapper for the `oneshot` channel that ignores send errors, as this should be safer.
It would be nice if you could suggest a better name (than `oneshot_nofail`) for the new wrapper.

### DIFF
--- a/p2p/src/utils/oneshot_nofail.rs
+++ b/p2p/src/utils/oneshot_nofail.rs
@@ -15,6 +15,9 @@
 
 //! Simple wrapper for the `tokio::sync::oneshot` channel
 //! that does not return an error if the receiver is disconnected.
+//!
+//! The wrapper could be used when sending to a closed channel is not considered an error
+//! (for example, when the async receiver was canceled for some reason).
 
 use std::{
     future::Future,
--- a/p2p/src/utils/oneshot_nofail.rs
+++ b/p2p/src/utils/oneshot_nofail.rs
@@ -1,0 +1,50 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Simple wrapper for the `tokio::sync::oneshot` channel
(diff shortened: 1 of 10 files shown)


p2p/src/utils/oneshot_nofail.rs
```diff
@@ -0,0 +1,50 @@
...
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Simple wrapper for the `tokio::sync::oneshot` channel
+//! that does not return an error if the receiver is disconnected.
```
TheQuantumPhysicist, line 17: Please add here why this exists and what problem it solves (pretty much what you wrote in the PR)

TheQuantumPhysicist (approved):
Just more documentation is needed. Otherwise, I'm fine with this.

(Even though we're not supposed to allow canceling unless on shutdown... but sure, I don't mind)

p2p/src/utils/oneshot_nofail.rs
pavel-kokolemin: Sure, updated, thanks.

stanislav-tkach (approved):
I like this idea. While changing the sync manager I also had to carefully examine channels errors.

iljakuklic:
Wondering if we could use this for subsystem call return values. Subsystems are meant to be long-lived, and sending the return value only really fails if if the target subsystem shuts down before awaiting the return value future, which is a corner case in which not much can be done anyway. Did I miss something?

pavel-kokolemin:
I agree that reporting send errors to long-lived tasks is usually pointless. Personally, I would remove most of the `P2pError::ChannelClosed` reports from the p2p, and the code should be much cleaner after that. I even started to do this, but decided that it was too radical.

PR #670, opened 2023-02-02
Remove unnecessary map description

TheQuantumPhysicist:
The storage interface is a key-value store that only needs to know the number of databases required. MapDesc seems to be unnecessary, causing complications and confusion when implementing other storage backends, like we faced with sqlite. This removes that layer and simplifies the interface.


iljakuklic:
I don't agree with this change. It deprives the storage backends of potentially useful information for dubious benefit. Having more information about the database structure in the backend potentially improves debugability, enables some optimizations, and may have use in future implementation of database migrations. With this, everything has to be inferred from just one number, the number of key-value maps. Juggling numeric indexes around may prove pretty brittle and error prone.

TheQuantumPhysicist:
> I don't agree with this change. It deprives the storage backends of potentially useful information for dubious benefit. Having more information about the database structure in the backend potentially improves debugability, enables some optimizations, and may have use in future implementation of database migrations. With this, everything has to be inferred from just one number, the number of key-value maps. Juggling numeric indexes around may prove pretty brittle and error prone.

I thought we agreed on this change in the last call, after having seen how SQLite wrapper design became difficult to understand. But apparently this is a misunderstanding.

This is only causing more confusion with little benefit. In reality, we'll never see the name of the database anywhere, because that's within the database implementation, which is a black-box for all we know. The only case where we can see words instead of numbers is in the in-memory implementation, which is extremely simple. I don't see much benefit in the debug process because of this. It's superficial. On the contrary, I can even imagine cases where people use wrong names or bad names and cause even more confusion, compared to just a number that can be learned once during the debug process, and then the debug continues. 

There is even a case to be made that entering names manually and allowing duplicates can cause undefined behavior that's difficult to detect due to using the same map twice with different indexes. You solved this problem with compile-time beautiful things in chainstate, but the backend is still prone to this problem because of this interface.

From my perspective, this just causes more confusion on what links to what, and what indexes relate to one, compared to one number (index) going all the way from the application to storage.

If you feel strongly about this, let's have a call about it to see if I'm missing something.

iljakuklic:
> I thought we agreed on this change in the last call, after having seen how SQLite wrapper design became difficult to understand. But apparently this is a misunderstanding.

I don't think it was difficult to understand. If there are places where it could be more streamlined, we can think about how to refactor it preferably without sacrificing the features described further below. I agreed it's not strictly necessary to have the metadata about databases there but it was not my preferred option.

> This is only causing more confusion with little benefit. In reality, we'll never see the name of the database anywhere, because that's within the database implementation, which is a black-box for all we know. The only case where we can see words instead of numbers is in the in-memory implementation, which is extremely simple. I don't see much benefit in the debug process because of this. It's superficial. On the contrary, I can even imagine cases where people use wrong names or bad names and cause even more confusion, compared to just a number that can be learned once during the debug process, and then the debug continues.

You can inspect both sqlite file and lmdb file using command line tools. It is then more appropriate to run queries like `select * from Blocks` rather than `select * from db_19`. Or perhaps if you enable logging of SQL queries while the node is running, the logs will be more readable. It is immediately obvious to see which table corresponds to which schema entry. With indices, it depends on the order in which the map descriptions occur in the schema. It is difficult to associate numbers with entries in the schema.

The original vision was to use the names as a semi-stable identifier and the indices as purely an internal identifier. Using a string to identify databases has a couple of advantages:

* You can reorder schema entries without breaking the database. I.e. when grouping entries together for readability
* When adding a new key-value map into the schema, it does not need to be added at the end. Quite often, the most natural place to add an entry is close to other related entries which does not need to be at the end.
* In general, I don't like the idea of making the order of schema entries significant. Will be confusing for new contributors and a rather artificial restriction.
* The exact nature of schema changes can be more easily observed. For example, removing a map `A` and adding maps `B` and `C`, you will be able to see that one database was removed, two added, and which ones. Having numeric index only signifies just there is the net increase of the number of databases by one. While this sort of migration is mostly an application-level thing, the backend can still do a more precise check that everything is fine.

Now it would be possible to use strings alone to identify the key-value databases. That however introduces an unnecessary overhead of looking up stuff in `BTreeMap` (or similar) by string. Using numeric indexes internally gets around that issue since then you can just do a fast `Vec` access. Converting an index to a name is also trivial by a `Vec` access into a `DbDesc`. Converting the other way should never be needed, since all the methods are given the key-value map index; there is no case where you start with a string.

> There is even a case to be made that entering names manually and allowing duplicates can cause undefined behavior that's difficult to detect due to using the same map twice with different indexes. You solved this problem with compile-time beautiful things in chainstate, but the backend is still prone to this problem because of this interface.

The duplicate issue is handled by a separate mechanism, `DbDesc` is never created manually outside of tests. Backend cannot manipulate it because mutating methods on `DbDesc` are not exposed.

> From my perspective, this just causes more confusion on what links to what, and what indexes relate to one, compared to one number (index) going all the way from the application to storage.
> 
> If you feel strongly about this, let's have a call about it to see if I'm missing something.

It's not a hill I am willing to die on but I don't think this is an improvement.

TheQuantumPhysicist:
Too many things to respond to. I guess a call is better than write them here.

But just as heads up for what's upsetting me and will be the center of the problem: We either have a map or a vector. We can't have both. In a couple of months we both will forget why we have both; it doesn't make sense. The overhead of using a map is negligible compared to any I/O operations we're doing. So to keep the code clean I'm OK with moving to a map completely. But this twisted situation where half is this and half is that is crazy...

PR #671, opened 2023-02-03
Database maps documentation and refactoring

TheQuantumPhysicist:


iljakuklic:
I pushed the number of changes:

Renamed a bunch of things (improvement suggestions welcome):
* Most types have gotten `Db` prefix to avoid confusion e.g. between `DbMap`s and other kind of maps
* DbIndex -> MapIndex -> DbMapId (to de-emphasise it's a numeric value under the hood)
* `storage_core::info` -> `storage_core::types`

Introduced a new polymorphic wrapper `DbMapsData<T>` that can be used to associate arbitrary data of type `T` with each DB map. The interface is limited to make it hard to use in an inappropriate way. Since it is indexed into with `DbMapId`, there is not much chance of mixup and you no longet neex to extract the numeric index out of it to index into a `Vec`.
* Storage layer uses it to pass DB map info into the backend (names, size hints, etc)  `DbMapsData<DbMapDesc>`
* LMDB uses it to store database handles `DbMapsData<lmdb::Database>`
* SQLite uses it to store pre-calculated queries on per-table basis `DbMapsData<SqlQuery>`
* In-memory backend uses it to store the data
* etc...

Changed the sqlite backend to use string DB map names for table names.

Added a high-level description and backend implementation overview to the top-level `storage_core` crate documentation.

Please let me know what you think. Especially please pay attention if the terminology is still not consistent in places, this stuff is easily missed.

storage/core/src/types.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Like [Self::transform] but fallible
+    pub fn try_transform<U, E>(
+        &self,
+        func: impl FnMut(&T) -> Result<U, E>,
+    ) -> Result<DbMapsData<U>, E> {
+        self.0.iter().map(func).collect::<Result<Vec<_>, _>>().map(DbMapsData)
+    }
+
+    /// Convert into iterator over map indices together with associated data
+    pub fn into_idx_iter(self) -> impl Iterator<Item = (DbMapId, T)> + ExactSizeIterator {
```
alfiedotwtf, line 95: Should this be called `into_mapid_iter()`?

alfiedotwtf:
LGTM

alfiedotwtf (approved):

storage/core/src/types.rs
```diff
@@ -0,0 +1,185 @@
...
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Database backend support buildng blocks
```
TheQuantumPhysicist, line 16: building

storage/core/src/lib.rs
```diff
@@ -14,17 +14,91 @@
...
+//! [types::DbMapsData::transform] method can be used to transform backend-agnostic information
+//! about DB maps into backend-specific handle to the DB maps.
+//!
+//! The result of opening a backend is something of type [Backend::Impl] that implements the
+//! [backend::BackendImpl] trait. It represents an initialized and running database.
+//!
+//! ## Transactions
+//!
+//! Currently, all data accesses in backend happen through transactions. This restriction may be
+//! lifted in the future, allowing primitive opertations to be performed without a transaction for
```
TheQuantumPhysicist, line 62: opertations

storage/core/src/types.rs
```diff
@@ -0,0 +1,185 @@
...
+        Self(map_count.indices().map(constructor).collect())
+    }
+
+    /// Get the number of maps
+    pub fn db_map_count(&self) -> DbMapCount {
+        DbMapCount(self.0.len())
+    }
+
+    /// Apply given function to each map and collect the results
+    pub fn transform<U>(&self, func: impl FnMut(&T) -> U) -> DbMapsData<U> {
```
TheQuantumPhysicist, line 82: Isn't it easier to just implement IntoIterator for this type and then get all these functions for free?

storage/core/src/types.rs
```diff
@@ -0,0 +1,185 @@
...
+/// It can be constructed in two ways:
+/// 1. The [Self::new] method takes in the number of DB maps. The map count is in a type wrapper to
+///    make it harder to use arbitrary number there.
+/// 2. The [Self::transform] method is used to derive one piece of data from another. It preserves
+///    the data count and ensures the DB map IDs remain consistent between the original and the
+///    derived [DbMapsData] instance.
+/// 3. The data can only be accessed by a [DbMapId] so it's only indexed into using a valid DB
+///    map identifier.
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub struct DbMapsData<T>(Vec<T>);
```
TheQuantumPhysicist, line 68: I maintain that this should be an array, not a Vec... the size of this array is determined at compile-time... but again, you don't have to do this now.

What hurts my eyes most in this is the excessive usage of the index operator for every operation, which is constantly done with a boundary checks. The reason this isn't so bad is that it's accompanied with I/O operations, so in comparison it's negligible. But from a type system point of view, this type's size should be fixed at compile-time.

TheQuantumPhysicist:
Looking good.

storage/core/src/types.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+/// Number of key-value maps in the database
+#[derive(Eq, PartialEq, PartialOrd, Ord, Clone, Copy, Debug)]
+pub struct DbMapCount(usize);
+
+impl DbMapCount {
+    /// Convert to `usize`
+    pub const fn as_usize(self) -> usize {
```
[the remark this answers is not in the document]
iljakuklic: It's deliberately to make it more explicit when these conversions are used. The backend is not supposed to construct `DbMapCount` but it still needs to be available for tests and the backend-agnostic storage layer.

storage/core/src/types.rs
iljakuklic: I considered that but in the end decided for a more limited interface. Iterator is allowed to reorder, drop, duplicate elements, all sorts of stuff like that. This ensures the map IDs match between the original `DbMapData` and the resulting one.

storage/core/src/types.rs
iljakuklic: Well I see the point but the tradeoff is that the compiler will generate the backend for each map count with almost identical code which seems a bit excessive. The backends are deliberately independent of the schema, so you can build them once and apply to many databases. It also provides conceptual separation at design level where the schema does not leak into type of the backend.

iljakuklic: By the way changing this to an array does not eliminate the boundary checks

TheQuantumPhysicist: Out of curiosity, isn't the separation you're referring to achieved by having `T` as a generic type?

Also I think the checks will be eliminated because all accesses are done through enums, which are defined at compile-time. The compiler is smart enough to do that.

iljakuklic: > Out of curiosity, isn't the separation you're referring to achieved by having `T` as a generic type?

I don't see how that's related. The `T` refers to data stored for each db map and can be picked by the backend at will. Nothing to do with schema or number of maps as such. Maybe I misunderstand your comment.
 
> Also I think the checks will be eliminated because all accesses are done through enums, which are defined at compile-time. The compiler is smart enough to do that.

Currently, there are no enums as far as I can tell. Maybe you have a different design in mind. At the front end, the db map is identified by a unit struct, not an enum. If we were to change the db map index to be represented by an enum, then in order for the compiler to take the advantage of it to eliminate bounds checking, the backend would have to be parametrized by the enum.

storage/core/src/types.rs
```diff
@@ -0,0 +1,185 @@
...
+use core::ops::Range;
+
+/// Used to identify particular key-value map in the database
+#[derive(Eq, PartialEq, PartialOrd, Ord, Clone, Copy, Debug)]
+pub struct DbMapId(usize);
+
+impl DbMapId {
+    /// New index
+    pub const fn new(id: usize) -> Self {
+        DbMapId(id)
```
stanislav-tkach, line 27: Super tiny nitpick: Self also can be used here. 🙃 
```suggestion
        Self(id)
```

stanislav-tkach (approved):

storage/core/src/types.rs
TheQuantumPhysicist: I meant the enum at the application layer.

Not important. Maybe one day I'll try it and then I'll figure out what you're saying myself 😄

PR #674, opened 2023-02-06
Update the sync manager logic

stanislav-tkach:
- Announce headers instead of blocks.
- Request multiple blocks at once, but send them one by one.
- Slightly different error handling.

Closes https://github.com/mintlayer/mintlayer-core/issues/596.


p2p/src/config.rs
```diff
@@ -80,4 +83,10 @@ pub struct P2pConfig {
     pub node_type: NodeTypeSetting,
     /// Allow announcing and discovering local and private IPs. Should be used for testing only.
     pub allow_discover_private_ips: AllowDiscoverPrivateIps,
+    /// A maximum allowed number of headers in one message.
+    pub header_limit: HeaderLimit,
```
TheQuantumPhysicist, line 87: header_count_limit?

p2p/src/config.rs
```diff
@@ -80,4 +83,10 @@ pub struct P2pConfig {
...
+    /// A maximum number of the elements in the locator.
+    pub max_locator_size: MaxLocatorSize,
```
TheQuantumPhysicist, line 89: max_locator_count?

p2p/src/config.rs
```diff
@@ -80,4 +83,10 @@ pub struct P2pConfig {
...
+    /// A maximum number of blocks that can be requested from a single peer.
+    pub requested_blocks_limit: RequestedBlocksLimit,
```
TheQuantumPhysicist, line 91: max_request_blocks_count?

p2p/src/peer_manager/tests/connections.rs
```diff
@@ -714,6 +717,9 @@ where
         ping_timeout: Default::default(),
         node_type: Default::default(),
         allow_discover_private_ips: Default::default(),
+        header_limit: Default::default(),
+        max_locator_size: Default::default(),
+        requested_blocks_limit: Default::default(),
     });
```
TheQuantumPhysicist, line 723: There's a typo "mades" down there... please fix it.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,492 @@ where
...
     }
 
-    /// Validate incoming block response
-    async fn validate_block_response(
-        &mut self,
-        peer_id: &T::PeerId,
-        blocks: Vec<Block>,
-    ) -> crate::Result<Option<BlockHeader>> {
-        let peer = self
+    // TODO: This shouldn't be public.
```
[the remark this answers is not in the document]
stanislav-tkach: Unfortunately, it is an artifact from the previous design where almost everything was made public for tests. I wasn't sure that I would be able to fix everything in this pull request, but currently I'm trying to rewrite the tests and hopefully this can be removed too.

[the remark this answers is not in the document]
stanislav-tkach: Most of the methods become private now, but unfortunately not all of them.

p2p/src/sync/mod.rs
```diff
@@ -61,25 +62,32 @@ const HEADER_LIMIT: usize = 2000;
...
-    /// Subsystem handle to Chainstate
+    /// A handle to the chainstate subsystem.
     chainstate_handle: subsystem::Handle<Box<dyn chainstate_interface::ChainstateInterface>>,
+
+    /// A queue of the blocks requested by peers.
+    ///
+    /// The block identifiers are added to the queue as a result of BlockListRequest processing
+    /// and removed either after sending a response or when the peer is disconnected. A number of
+    /// blocks is limited by `P2pConfig::requested_blocks_limit` per peer.
+    blocks_queue: VecDeque<(T::PeerId, Id<Block>)>,
```
pavel-kokolemin, line 90: What is the point of `blocks_queue`?
I think it's pretty useless because there is no backpressure.
This line will most likely be selected first if `blocks_queue` is not empty, and the code will work much the same as without using `blocks_queue` at all.

```
tokio::select! {
...
    (peer, block) = async { self.blocks_queue.pop_front().expect("The block queue is empty") }, if !self.blocks_queue.is_empty() => {
        let res = self.send_block(peer, block).await;
        self.handle_result(peer, res).await?;
    }
}
```

pavel-kokolemin:
Overall, I don't like how the sync manager works because there are a lot of synchronization points with the chainstate subsystem (when we request something from the chainstate). I think this will result in high latency and low throughput.
IMO a better way would be to do the processing directly in the peer's event loop. This would solve the backpressure and congestion problems (because we won't start processing new messages until current message is processed and response is sent, and peers won't block each other). But this requires big changes.

p2p/src/sync/mod.rs
pavel-kokolemin: @stanislav-tkach explained that it's needed to allow sync manager to receive messages without blocking the event loop for too long.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,484 @@ where
...
+                },
+                event = self.peer_event_receiver.recv() => match event.ok_or(P2pError::ChannelClosed)? {
+                    SyncControlEvent::Connected(peer_id) => self.register_peer(peer_id).await?,
+                    SyncControlEvent::Disconnected(peer_id) => self.unregister_peer(peer_id),
+                },
+                block_id = new_tip_receiver.recv(), if !self.chainstate_handle.call(|c| c.is_initial_block_download()).await?? => {
+                    // This error can only occur when chainstate drops an events subscriber.
+                    let block_id = block_id.ok_or(P2pError::ChannelClosed)?;
+                    self.handle_new_tip(block_id).await?;
+                },
```
pavel-kokolemin, line 146: I think this line will slow down the event loop a lot, because it will block processing until the `chainstate` call returns. I think it would be much better to somehow listen for `is_initial_block_download` updates instead.

dns_server/src/main.rs
```diff
@@ -47,6 +47,9 @@ async fn run(config: Arc<DnsServerConfig>) -> Result<void::Void, error::DnsServe
         ping_timeout: Default::default(),
         node_type: NodeType::DnsServer.into(),
         allow_discover_private_ips: Default::default(),
+        header_count_limit: Default::default(),
```
TheQuantumPhysicist, line 50: Sorry to comment on this again. But I suggest `msg_header_count_limit`

dns_server/src/main.rs
```diff
@@ -47,6 +47,9 @@ async fn run(config: Arc<DnsServerConfig>) -> Result<void::Void, error::DnsServe
...
+        max_locator_count: Default::default(),
```
TheQuantumPhysicist, line 51: Also here `msg_max_locator_count`

p2p/src/sync/mod.rs
TheQuantumPhysicist: OK I now see the context of that message on slack. 

We don't need to do that. The answer is much simpler: You can always check against the best block known from the last announcement, then check only once if that changed. You can even cache the result on every announcement and just load it here.

p2p/src/sync/tests/A_REMOVE_block_response.rs
```diff
@@ -0,0 +1,305 @@
...
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
```
TheQuantumPhysicist, line 15: This file has an interesting filename... what's the story on this?

p2p/src/sync/tests/block_announcement.rs
```diff
@@ -0,0 +1,136 @@
...
+    let blocks = create_n_blocks(
+        Arc::clone(&chain_config),
+        TestBlockInfo::from_genesis(chain_config.genesis_block()),
+        2,
+    );
+    handle.make_announcement(peer, Announcement::Block(blocks[1].header().clone()));
+
+    let (sent_to, message) = handle.message().await;
+    assert_eq!(sent_to, peer);
+    assert!(matches!(message, SyncMessage::HeaderListRequest(_)));
```
TheQuantumPhysicist, line 65: Maybe this should be punished. Please create a TODO or an issue about it to study it further.

The problem is that being graceful about such a move basically means malicious peers can make us do an infinite amount of database accesses, which is an easy DoS attack.

p2p/src/sync/tests/block_announcement.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    let (adjusted_peer, score) = handle.adjust_peer_score_event().await;
+    assert_eq!(peer, adjusted_peer);
+    assert_eq!(score, 20);
+}
+
+// The peer ban score is increased if it sends an invalid header.
+#[tokio::test]
+async fn invalid_consensus_data() {
+    let chain_config = Arc::new(create_mainnet());
```
TheQuantumPhysicist, line 95: Please use the ChainConfig builder... we should have a starting point from a given configuration and then modify it a little for tests like this.

p2p/src/sync/tests/block_list_request.rs
```diff
@@ -0,0 +1,136 @@
...
+        .map(|b| b.get_id())
+        .collect();
+    handle.send_message(
+        peer,
+        SyncMessage::BlockListRequest(BlockListRequest::new(unknown_blocks)),
+    );
+
+    let (adjusted_peer, score) = handle.adjust_peer_score_event().await;
+    assert_eq!(peer, adjusted_peer);
+    assert_eq!(score, 20);
```
TheQuantumPhysicist, line 96: Why are we testing against literal numbers, instead of using error values?

dns_server/src/main.rs
stanislav-tkach: Renaming is very easy thanks to the IDE support, so I don't mind renaming things any number of times if we can have better names as a result.

p2p/src/sync/tests/A_REMOVE_block_response.rs
stanislav-tkach: It is a story about an oversight. 🙃  While writing new tests I preserved the old ones to double check that I wasn't missing some of the test cases, but I forgot to remove this file after that.

p2p/src/sync/tests/block_list_request.rs
```diff
@@ -0,0 +1,145 @@
...
+    );
+    import_blocks(&chainstate, vec![block.clone()]).await;
+
+    let mut handle =
+        SyncManagerHandle::with_chainstate_and_config(Arc::clone(&chain_config), chainstate).await;
+
+    let peer = PeerId::new();
+    handle.connect_peer(peer).await;
+
+    let blocks = iter::repeat(block.get_id()).take(501).collect();
```
TheQuantumPhysicist, line 60: Please don't use literals...

p2p/src/sync/tests/block_list_request.rs
```diff
@@ -0,0 +1,145 @@
...
+        SyncMessage::BlockResponse(BlockResponse::new(blocks[0].clone()))
+    );
+
+    let (sent_to, message) = handle.message().await;
+    assert_eq!(peer, sent_to);
+    assert_eq!(
+        message,
+        SyncMessage::BlockResponse(BlockResponse::new(blocks[1].clone()))
+    );
+}
```
TheQuantumPhysicist, line 145: In these "valid" cases, please assert that the score never changed.

p2p/src/sync/tests/block_response.rs
```diff
@@ -15,294 +15,104 @@
...
 
-    assert_eq!(
-        mgr.validate_header_response(&peer_id, vec![]).await,
-        Err(P2pError::PeerError(PeerError::PeerDoesntExist)),
+    let block = create_block(
+        Arc::clone(&chain_config),
+        TestBlockInfo::from_genesis(chain_config.genesis_block()),
     );
+    handle.send_message(peer, SyncMessage::BlockResponse(BlockResponse::new(block)));
 }
```
TheQuantumPhysicist, line 45: Does this test ever fail?

p2p/src/sync/tests/block_response.rs
```diff
@@ -15,294 +15,104 @@
...
-    register_peer(&mut mgr, peer_id).await;
-
-    let blocks = p2p_test_utils::create_n_blocks(
-        Arc::clone(&config),
-        TestBlockInfo::from_genesis(config.genesis_block()),
-        1,
+    let blocks = create_n_blocks(
+        Arc::clone(&chain_config),
+        TestBlockInfo::from_genesis(chain_config.genesis_block()),
+        2,
```
TheQuantumPhysicist, line 80: Why not a random number?

p2p/src/sync/tests/header_list_request.rs
```diff
@@ -0,0 +1,92 @@
...
+#[tokio::test]
+async fn nonexistent_peer() {
+    let mut handle = SyncManagerHandle::start().await;
+
+    let peer = PeerId::new();
+    handle.send_message(
+        peer,
+        SyncMessage::HeaderListRequest(HeaderListRequest::new(Locator::new(Vec::new()))),
+    );
+}
```
TheQuantumPhysicist, line 39: Again, can this ever fail?

p2p/src/sync/tests/header_list_request.rs
```diff
@@ -0,0 +1,92 @@
...
+
+#[tokio::test]
+async fn locator_size_exceeded() {
```
TheQuantumPhysicist, line 42: max_locator_size_exceeded

p2p/src/sync/tests/peer_events.rs
```diff
@@ -0,0 +1,61 @@
...
+};
+
+// Check that the header list request is sent to a newly connected peer.
+#[tokio::test]
+async fn connect_peer() {
+    let mut handle = SyncManagerHandle::start().await;
+
+    let peer = PeerId::new();
+    handle.connect_peer(peer).await;
+}
```
TheQuantumPhysicist, line 28: Maybe it's best to assert the state after doing these operations, like assert that the peer exists

p2p/backend-test-suite/src/ban.rs
```diff
@@ -99,13 +110,13 @@ where
...
 
-        sync2.make_announcement(Announcement::Block(blocks[2].clone())).unwrap();
+        sync2.make_announcement(Announcement::Block(block.header().clone())).unwrap();
     });
 
     match rx_peer_manager.recv().await {
         Some(PeerManagerEvent::AdjustPeerScore(peer_id, score, _)) => {
             assert_eq!(peer_id, peer_info2.peer_id);
-            assert_eq!(score, 100);
+            assert_eq!(score, 20);
```
TheQuantumPhysicist, line 119: literals

p2p/src/sync/tests/block_response.rs
stanislav-tkach: I'm not sure what randomness adds here. The test checks that all requested blocks are sent and in my understanding any number more than 1 proves the point.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+        let mut new_tip_receiver = self.subscribe_to_new_tip().await?;
+        loop {
+            tokio::select! {
+                event = self.messaging_handle.poll_next() => match event? {
+                    SyncingEvent::Message { peer, message } => {
+                        let res = self.handle_message(peer, message).await;
+                        self.handle_result(peer, res).await?;
+                    },
+                    SyncingEvent::Announcement{ peer, announcement } => {
+                        let res = self.handle_announcement(peer, announcement).await;
```
TheQuantumPhysicist, line 134: Why do we even need announcements? Why not just use the headers message?

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+            .await
+            .map_err(|_| P2pError::SubsystemFailure)?;
+
+        Ok(receiver)
     }
 
-    /// Unregister peer from the `SyncManager`
-    pub fn unregister_peer(&mut self, peer_id: T::PeerId) {
-        self.peers.remove(&peer_id);
+    async fn handle_message(&mut self, peer: T::PeerId, message: SyncMessage) -> Result<()> {
```
TheQuantumPhysicist, line 176: I suggest moving all these handler functions to a `protocol.rs` file

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+        log::trace!("locator: {locator:#?}");
+
+        if self.chainstate_handle.call(|c| c.is_initial_block_download()).await?? {
+            // TODO: Check if a peer has permissions to ask for headers during the initial block download.
+            log::debug!("Ignoring headers request because the node is in initial block download");
+            return Ok(());
+        }
+
+        let headers = self.chainstate_handle.call(|c| c.get_headers(locator)).await??;
+        debug_assert!(headers.len() <= self.p2p_config.msg_header_count_limit.clone().into());
```
TheQuantumPhysicist, line 213: The header-limit in the function c.get_headers() is not taken from p2p. Perhaps you should modify that function to take the limit as a function parameter instead of using the HEADER_LIMIT constant.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+        peer_state.num_blocks_to_send += block_ids.len();
+        debug_assert!(peer_state.num_blocks_to_send <= requested_blocks_limit);
+        self.blocks_queue.extend(block_ids.into_iter().map(|id| (peer, id)));
+
+        Ok(())
     }
 
-    /// Validate incoming header response
-    async fn validate_header_response(
+    async fn handle_header_response(
```
TheQuantumPhysicist, line 263: Well, if we remove announcements, this becomes `handle_headers_message`

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+        let requested_blocks_limit = self.p2p_config.max_request_blocks_count.clone().into();
+        if block_ids.len() > requested_blocks_limit {
+            return Err(P2pError::ProtocolError(
+                ProtocolError::BlocksRequestLimitExceeded(block_ids.len(), requested_blocks_limit),
+            ));
         }
+        log::trace!("Requested block ids: {block_ids:#?}");
+
+        // Check that all blocks are known.
+        for id in block_ids.clone() {
```
TheQuantumPhysicist, line 249: This is really bad. Every call will lock a mutex through the channels... why? Why not run the loop in the closure?

TheQuantumPhysicist: Also... out of curiosity, why clone if this isn't passed into the lambda? It's irrelevant now since it should go inside the closure.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+            self.chainstate_handle.call(move |c| c.get_block_index(&id)).await??.ok_or(
+                P2pError::ProtocolError(ProtocolError::UnknownBlockRequested),
+            )?;
+        }
+
+        block_ids.truncate(requested_blocks_limit - peer_state.num_blocks_to_send);
```
TheQuantumPhysicist, line 255: Instead of checking then separately subtracting again, creating an unnecessary invariant, you can use match or if-else with an expression to extract this value or return an error.

p2p/src/sync/tests/block_response.rs
TheQuantumPhysicist: You think it proves the point, until it finds a bug you didn't expect :wink:

The idea of randomizing tests is that we don't test an arbitrary case, but we test a model that makes us understand the code better and find bugs we never imagined.

In other words: You think that your point is proven with anything more than `1`? Great. Make a test that tests this assumption. There's no loss here. Only gain.

This principle should be applied everywhere in our code, because mistakes and oversights happen.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
-        match self.validate_header_response(&peer_id, headers).await {
-            Ok(Some(header)) => self.send_block_request(peer_id, header.get_id()),
-            Ok(None) => {
-                self.peers
-                    .get_mut(&peer_id)
-                    .ok_or(P2pError::PeerError(PeerError::PeerDoesntExist))?
-                    .set_state(peer::PeerSyncState::Idle);
-                Ok(())
+        let is_max_headers =
+            headers.len() == Into::<usize>::into(self.p2p_config.msg_header_count_limit.clone());
```
TheQuantumPhysicist, line 320: Isn't it easier to use Deref instead of this huge Into? I think this works:

```
headers.len() == *self.p2p_config.msg_header_count_limit;
```

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
-                ChainstateError::FailedToInitializeChainstate(_) => 0,
-                ChainstateError::ProcessBlockError(err) => err.ban_score(),
-                ChainstateError::FailedToReadProperty(_) => 0,
-                ChainstateError::BootstrapError(_) => 0,
-            },
-        };
+    /// Sends a header list request to the given peer.
+    async fn request_headers(&mut self, peer: T::PeerId) -> Result<()> {
+        let locator = self.chainstate_handle.call(|this| this.get_locator()).await??;
+        debug_assert!(locator.len() <= self.p2p_config.msg_max_locator_count.clone().into());
```
TheQuantumPhysicist, line 546: Again, just deref instead of into

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+    /// The number of headers sent equals to `P2pConfig::requested_blocks_limit`, the remaining
+    /// headers are stored in the peer context.
+    fn request_blocks(&mut self, peer: T::PeerId, mut headers: Vec<BlockHeader>) -> Result<()> {
+        let peer_state = self
+            .peers
+            .get_mut(&peer)
+            .ok_or(P2pError::PeerError(PeerError::PeerDoesntExist))?;
+
+        debug_assert!(peer_state.known_headers.is_empty());
+        if headers.len() > self.p2p_config.max_request_blocks_count.clone().into() {
```
TheQuantumPhysicist, line 567: OK... this is becoming a pattern. In many places you do clone then into unnecessarily. Please look into them all and simply use deref

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
 
-        for (a, b) in headers.iter().zip(&headers[1..]) {
-            ensure!(
-                b.prev_block_id() == &a.get_id(),
-                P2pError::ProtocolError(ProtocolError::InvalidMessage),
-            );
+        // Each header must be connected to the previous one.
+        if !headers
+            .iter()
+            .zip(&headers[1..])
```
TheQuantumPhysicist, line 298: Isn't easier to use the `windows(2)` method in `Vec`?

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+        }
 
-        let block = blocks.into_iter().next().expect("block to exist");
-        let header = block.header().clone();
-
-        let result = match self
+        match self
             .chainstate_handle
-            .call(move |this| this.preliminary_block_check(block))
+            .call_mut(|c| {
```
TheQuantumPhysicist, line 362: It's an open question whether we want to first run `call` and do a check on the block before calling the with `call_mut` and `process_block`. The reasoning is that `call` can run in parallel, hence an attacker can't block our subsystems.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
-                        }
-                    }
-                },
-            );
+    /// Sends a block to the peer.
+    async fn send_block(&mut self, peer: T::PeerId, block: Id<Block>) -> Result<()> {
+        self.peers
+            .get_mut(&peer)
+            .ok_or(P2pError::PeerError(PeerError::PeerDoesntExist))?
+            .num_blocks_to_send -= 1;
```
TheQuantumPhysicist, line 476: This is playing with fire. We must ensure somehow that this is checked, and this looks too naked for me. 

Again, I wanna reiterate that there's lots of arithmetic here that needs to be revised for security.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
 
-        Ok(rx)
-    }
+        match error {
+            // A protocol error - increase the ban score of a peer.
+            e @ (P2pError::ProtocolError(_)
+            | P2pError::ChainstateError(ChainstateError::ProcessBlockError(
+                BlockError::CheckBlockFailed(_),
+            ))) => {
+                log::debug!(
```
TheQuantumPhysicist, line 506: This should be `info!`

TheQuantumPhysicist: Also maybe print the current peer score.

p2p/src/sync/mod.rs
```diff
@@ -89,445 +97,485 @@ where
...
+    /// - Ignored errors aren't propagated and don't affect the peer score.
+    pub async fn handle_result(&mut self, peer: T::PeerId, result: Result<()>) -> Result<()> {
+        let error = match result {
+            Ok(()) => return Ok(()),
+            Err(e) => e,
+        };
 
-        Ok(rx)
-    }
+        match error {
```
TheQuantumPhysicist, line 500: I don't like this complex result processing. This isn't only hard to read but also qualifies to "playing with fire"...

The whole point of the `BanScore` trait is to be able to delegate all this to a single function call that gives you a number that can be audited...

p2p/src/sync/peer_context.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use std::collections::BTreeSet;
+
+use common::{
+    chain::block::{Block, BlockHeader},
+    primitives::Id,
+};
+
+// TODO: Investigate if we need some kind of "timeouts" (waiting for blocks or headers).
+pub struct PeerContext {
```
TheQuantumPhysicist, line 24: It's probably useful to track the current best block for the peer and the current unknown best block.

TheQuantumPhysicist:
Great first iteration! This is close to what we need to have proper block syncing. I'll give it another review later. Please handle the issues I mentioned in the comments.

p2p/src/sync/tests/peer_events.rs
stanislav-tkach: There is an assertion inside the `connect_peer` function:
```rust
pub async fn connect_peer(&mut self, peer: PeerId) {
    self.try_connect_peer(peer);

    let (sent_to, message) = self.message().await;
    assert_eq!(peer, sent_to);
    assert!(matches!(message, SyncMessage::HeaderListRequest(_)));
}
```
Also I have added more assertions everywhere.

p2p/src/sync/mod.rs
stanislav-tkach: I wanted to avoid copying the whole vector in case one of the first identifiers is invalid, but I missed the call overhead. Fixed.

p2p/src/sync/mod.rs
stanislav-tkach: I don't like `std::slice::windows` because this method returns a slice, so I have rewritten this code using `tuple_windows` from `itertools`.

p2p/src/sync/mod.rs
stanislav-tkach: There is a special logic for the announcements in the bitcoin code even though there is no special message, so I decided to keep the separation that we already had.

p2p/src/sync/mod.rs
stanislav-tkach: I don't like the complex logic here either, but I still think that three distinct cases should be handled differently. There are a few errors (such as `SubsystemFailure`) that indicate a completely broken state, so there is no reason to continue processing. Also some errors should never occur in normal situations (such as `NoiseHandshakeError` or `DialError`). I suppose that ideally we can use more specific errors where it is appropriate instead of general `P2pError` everywhere. I can try to do that if you like this idea, but I prefer to do that in a separate pull request.

p2p/src/sync/mod.rs
stanislav-tkach: Initially I tried to make it harder to break the invariants by making the `PeerContext` fields private, but it was clumsy, so I removed that. I will think more about that.

p2p/src/sync/mod.rs
stanislav-tkach: I think that I have fixed this issue.

stanislav-tkach:
I think that I have fixed most of the comments and this is ready for the next review round. I want to highlight the following:

- The checked arithmetic is used where appropriate.
- I'm not 100% sure about the new block announcement and the "headers response" separation, but as far as I can see, the logic is slightly different for the announcements in the Bitcoin code (it can be found by the `MAX_BLOCKS_TO_ANNOUNCE` constant usage). I suppose it becomes even more important when a more intricate peer logic (such as disconnecting stalling peers in favor of more active ones) is implemented. That's why I prefer to keep the announcements for now.
- While the error processing logic is somewhat complicated, I still believe that it makes the code more robust, because unexpected errors aren't ignored.
- Several additional TODO-notes were added.

p2p/src/sync/mod.rs
```diff
@@ -488,46 +173,110 @@ where
...
+    }
+
+    /// Sends a block to the peer.
+    async fn send_block(&mut self, peer: T::PeerId, block: Id<Block>) -> Result<()> {
+        self.peers
+            .get_mut(&peer)
+            .ok_or(P2pError::PeerError(PeerError::PeerDoesntExist))?
+            .num_blocks_to_send
+            .checked_sub(1)
+            // This is safe because the function is called after the `blocks_queue.is_empty` check.
```
TheQuantumPhysicist, line 235: Two things:

1. Why not add that assert at the beginning of the function?
2. Why do we even need to track the "number" if we have a queue, whose `len()` can be called?

p2p/src/sync/mod.rs
```diff
@@ -488,46 +173,110 @@ where
...
+        if self.peers.remove(&peer).is_some() {
+            log::warn!("Unregistering unknown peer: {peer}");
+        }
+    }
+
+    /// Announces the header of a new block to peers.
+    async fn handle_new_tip(&mut self, block_id: Id<Block>) -> Result<()> {
+        if self.is_initial_block_download {
+            self.is_initial_block_download =
+                self.chainstate_handle.call(|c| c.is_initial_block_download()).await??;
```
TheQuantumPhysicist, line 210: Why call twice, once for is_initial_block_download() and once for get_block()? Why not both in one call?

p2p/src/sync/tests/peer_events.rs
TheQuantumPhysicist: Keep in mind that we can't rely on in-code asserts for tests. This is an anti-pattern. The idea is that code changes, but tests should be the judge on whether the code is valid.

If you've depended on this in other pleces, please track that and fix it.

TheQuantumPhysicist (approved):
I think this is in a good state now. Please fix the minor issues I mentioned and feel free to merge.

pavel-kokolemin (approved):

p2p/src/sync/mod.rs
stanislav-tkach: 1. The code looks the following:
```rust
tokio::select! {
    { ... },
    (peer, block) = async { self.blocks_queue.pop_front().expect("The block queue is empty") }, if !self.blocks_queue.is_empty() => {
        let res = self.send_block(peer, block).await;
        self.handle_result(peer, res).await?;
    }
}

async fn send_block(&mut self, peer: T::PeerId, block: Id<Block>) -> Result<()> {
    self.peers
        .get_mut(&peer)
        .ok_or(P2pError::PeerError(PeerError::PeerDoesntExist))?
        .num_blocks_to_send
        .checked_sub(1)
        // This is safe because the function is called after the `blocks_queue.is_empty` check.
        .expect("Trying to send a block when num_blocks_to_send is zero");

    ...
}
```
So it is too late to assert in the `send_block` function because an item is extracted from the queue before calling it.

2. `blocks_queue` is used to store all the pending requests for every connected peer, so if there are two peers and both of them requested 400 blocks each, than the queue will contain 800 block identifiers.

I suppose this code will be change after introducing the queue for futures anyway.

p2p/src/sync/tests/peer_events.rs
stanislav-tkach: > Keep in mind that we can't rely on in-code asserts for tests.

Sure, but it is actually a part of the tests. The actual function is called `register_peer` and `connect_peer` is just a helper for the tests.

Issue #498, opened 2022-10-21
Node directory handling


iljakuklic:
First, the datadir should be locked to prevent multiple node processes from accessing it simultaneously.

Second, it would be nice to honor `XDG_DATA_DIR`, `XDG_CONFIG_DIR` and such unless overridden.

Third, we there should be a cleaner approach to directory creation.

Relevant comment by @TheQuantumPhysicist from #493:

> The policy for creating directories should be handled with care. The default policy I see reasonable (same one as bitcoin's) is that:
> 
> * If the data directory is created by default in the default path, then create everything without question
> * If the data directory is chosen manually, it has to exist and nothing should be created.
> 
> This is probably not directly related to lmdb. This has to be handled in the node runner. It just occurred to me now.

PR #683, opened 2023-02-07
Basic DNS seed server implementation

pavel-kokolemin:
Add a basic scanner/DNS server that nodes can use for initial seeding.

To keep things simple, the server will try to keep connections open to all reachable nodes.
When a new outbound connection is made, a new DNS record is added (but only for nodes on default ports).
When the connection is closed, the DNS record is removed.
When a connection fails, the server increases the backoff time between connection attempts.
If the number of failed connection attempts exceeds the limit, the address is removed from the list.
Once-reachable and newer-reachable addresses have different connection failure limits
(equivalent to about 1 month and about 1 hour, respectively).

I'm attaching sample DNS responses from this server and from Bitcoin's ([https://github.com/sipa/bitcoin-seeder](https://github.com/sipa/bitcoin-seeder)):
[mintlayer.txt](https://github.com/mintlayer/mintlayer-core/files/10674942/mintlayer.txt)
[bitcoin.txt](https://github.com/mintlayer/mintlayer-core/files/10674943/bitcoin.txt)


p2p/src/config.rs
```diff
@@ -19,6 +19,7 @@ use utils::make_config_setting;
 
 use crate::net::types::PubSubTopic;
 
+// TODO: Use different ports on testnet and regtest
 pub const DEFAULT_BIND_PORT: u16 = 3031;
```
TheQuantumPhysicist, line 23: The ports should be taken from ChainConfig

dns_server/src/server/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+
+pub struct Server {
+    auth: Arc<AuthorityImpl>,
+
+    server: ServerFuture<Catalog>,
+
+    command_tx: mpsc::UnboundedReceiver<ServerCommands>,
+}
+
+// Same values as in `https://github.com/sipa/bitcoin-seeder/blob/master/dns.cpp`
```
TheQuantumPhysicist, line 58: Always use permalinks in such situations

dns_server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// Hostname of the nameserver
+    #[clap(long)]
+    pub ns: String,
+
+    /// E-Mail address reported in SOA records (RNAME)
+    #[clap(long)]
+    pub mbox: String,
+}
+
+const DEFAULT_DATA_DIR_NAME: &str = ".mintlayer";
```
TheQuantumPhysicist, line 48: Is it normal to use the same data directory for the node and the dns server?

dns_server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// Hostname of the DNS seed
+    #[clap(long)]
+    pub host: String,
+
+    /// Hostname of the nameserver
+    #[clap(long)]
+    pub ns: String,
+
+    /// E-Mail address reported in SOA records (RNAME)
```
TheQuantumPhysicist, line 43: Well, I'd be too hesitant to enforce using a valid email address... I'm always worried about spam.

dns_server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+// TODO: Allow using with testnet
+
+#[derive(Parser, Debug)]
+pub struct DnsServerConfig {
+    /// The path to the data directory
+    #[clap(long, default_value_t = default_data_dir())]
+    pub datadir: String,
+
+    /// IP address and UDP port to listen on
+    #[clap(long, default_values_t = vec!["[::]:53".to_string()])]
```
TheQuantumPhysicist, line 28: Why not 0.0.0.0:53?

dns_server/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+mod server;
+
+async fn run(config: Arc<DnsServerConfig>) -> Result<void::Void, error::DnsServerError> {
+    let (command_tx, command_rx) = mpsc::unbounded_channel();
+
+    let stream_adapter = NoiseEncryptionAdapter::gen_new();
+    let base_transport = TcpTransportSocket::new();
+    let transport = NoiseTcpTransport::new(stream_adapter, base_transport);
+    let chain_config = Arc::new(common::chain::config::create_mainnet());
+    let p2p_config = Default::default();
```
TheQuantumPhysicist, line 39: Do we really have to instantiate a full Mintlayer node server? Can't we get away with simple pinging that port? Or just opening a connection and closing it?

dns_server/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+mod config;
+mod crawler;
+mod error;
+mod server;
+
+async fn run(config: Arc<DnsServerConfig>) -> Result<void::Void, error::DnsServerError> {
+    let (command_tx, command_rx) = mpsc::unbounded_channel();
+
+    let stream_adapter = NoiseEncryptionAdapter::gen_new();
```
TheQuantumPhysicist, line 35: Please do all the instantiations of p2p in a separate function and take the result and use it. There seems to be many steps that shouldn't be done in here.

dns_server/src/server/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+use crate::{config::DnsServerConfig, error::DnsServerError};
+
+#[derive(Debug)]
+pub enum ServerCommands {
+    AddAddress(IpAddr),
+
+    DelAddress(IpAddr),
+}
+
+pub struct Server {
```
TheQuantumPhysicist, line 50: Why do we call this Server, and not DnsServer? I would prefer the latter unless there's a good reason not to do it.

dns_server/src/server/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+        lookup_options: LookupOptions,
+    ) -> Result<Self::Lookup, LookupError> {
+        self.inner.get_nsec_records(name, lookup_options).await
+    }
+}
+
+fn handle_command(auth: &AuthorityImpl, command: ServerCommands) {
+    match command {
+        ServerCommands::AddAddress(IpAddr::V4(ip)) => {
+            auth.ip4.lock().expect("mutex must be valid").push(ip);
```
TheQuantumPhysicist, line 274: These expects are too aggressive... deeming the whole program dead just because a wrong format was submitted. Perhaps we should just return an error with one shot.

dns_server/src/server/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+            .as_secs() as u32;
+        let old_serial = self.serial.swap(new_serial, Ordering::Relaxed);
+        if old_serial == new_serial {
+            return;
+        }
+
+        let ipv4 = self
+            .ip4
+            .lock()
+            .expect("mutex must be valid")
```
TheQuantumPhysicist, line 164: Perhaps a dumb comment... but please change "mutex must be valid" to "mutex must be valid (ipv4)". Using the same message in multiple places can lead to confusion when an error happens.

dns_server/src/server/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/// Wrapper for InMemoryAuthority that selects random addresses every second
+struct AuthorityImpl {
+    serial: AtomicU32,
+    host: Name,
+    ns: Name,
+    mbox: Name,
+    inner: InMemoryAuthority,
+    ip4: Mutex<Vec<Ipv4Addr>>,
```
TheQuantumPhysicist, line 146: Please do not use std::Mutex in async code. Using tokio mutex and away on it.

dns_server/src/crawler/storage_impl.rs
```diff
@@ -0,0 +1,94 @@
...
+use super::storage::{
+    DnsServerStorage, DnsServerStorageRead, DnsServerStorageWrite, DnsServerTransactionRo,
+    DnsServerTransactionRw, DnsServerTransactional,
+};
+
+storage::decl_schema! {
+    /// Database schema for peer db storage
+    pub Schema {
+        /// Table for all reachable addresses
+        pub DBAddresses: Map<String, ()>,
```
TheQuantumPhysicist, line 27: Please add a schema for versioning too. You need to store the version of your schema.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,516 @@
...
+        {
+            log::debug!("purge old (unreachable) address {}", address.to_string());
+
+            return false;
+        }
+
+        true
+    }
+
+    /// Address list maintaince
```
TheQuantumPhysicist, line 462: maintaince?

TheQuantumPhysicist:
Very nice work.

This is great, but we're still missing many things. Besides the comments I made, I think there are tons of tests that are still needed before getting this to be merged.

p2p/src/config.rs
pavel-kokolemin: OK, thanks. I added `p2p_port` field to `ChainConfig`.

dns_server/src/server/mod.rs
pavel-kokolemin: Sure, updated.

dns_server/src/config.rs
pavel-kokolemin: I switched to using `BaseDirs::data_dir` instead (on Linux this is `$HOME/.local/share/mintlayer_dns_server`).

dns_server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub ns: String,
+
+    /// E-Mail address reported in SOA records (RNAME)
+    #[clap(long)]
+    pub mbox: String,
+}
+
+const DEFAULT_DATA_DIR_NAME: &str = ".mintlayer";
+
+fn default_data_dir() -> String {
```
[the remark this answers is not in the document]
pavel-kokolemin: I could not use `PathBuf` with `clap::Parser` because `PathBuf` does not implement `std::fmt::Display`.

dns_server/src/config.rs
pavel-kokolemin: Looks like the SOA record must always be present: https://serverfault.com/a/1024024
And `rname` is the admin's email address in the SOA.
It's possible to run DNS without the SOA record, but all of Bitcoin's public seed DNSs have it, so I decided to make it mandatory.

dns_server/src/config.rs
pavel-kokolemin: On Linux, `[::]` works with both IPv6 and IPv4 clients (by default).

dns_server/src/main.rs
pavel-kokolemin: I'm using the generic `NetworkingService` from p2p, which works like a low-level networking implementation.
It's definitely not a full node server.
Simply pinging the ports is not enough, because we also need to know when new nodes join the network.

dns_server/src/main.rs
pavel-kokolemin: OK, updated.

dns_server/src/server/mod.rs
pavel-kokolemin: Sure, updated.

dns_server/src/server/mod.rs
pavel-kokolemin: Sure, updated.

dns_server/src/server/mod.rs
pavel-kokolemin: Std `Mutex` is OK here because the lock times are short.
From https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html:
> Contrary to popular belief, it is ok and often preferred to use the ordinary [Mutex](https://doc.rust-lang.org/nightly/std/sync/mutex/struct.Mutex.html) from the standard library in asynchronous code.

dns_server/src/crawler/storage_impl.rs
pavel-kokolemin: Sure, added.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Fixed, thanks.

dns_server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Parser, Debug)]
+pub struct DnsServerConfig {
+    /// The path to the data directory
+    #[clap(long, default_value_t = default_data_dir())]
+    pub datadir: String,
+
+    /// IP address and UDP port to listen on
+    #[clap(long, default_values_t = vec!["[::]:53".to_string()])]
+    pub bind_addr: Vec<String>,
```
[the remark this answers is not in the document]
pavel-kokolemin: Right, updated, thanks.

dns_server/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[clap(long)]
+    pub add_node: Vec<String>,
+
+    /// Hostname of the DNS seed
+    #[clap(long)]
+    pub host: String,
+
+    /// Hostname of the nameserver
+    #[clap(long)]
+    pub ns: String,
```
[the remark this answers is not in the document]
pavel-kokolemin: Updated, thanks.

pavel-kokolemin:
> Very nice work.
> 
> This is great, but we're still missing many things. Besides the comments I made, I think there are tons of tests that are still needed before getting this to be merged.

Thanks! I'm working on tests.

dns_server/src/config.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: That's unfortunate.

dns_server/src/config.rs
TheQuantumPhysicist: I would rather keep it in $HOME... I hate it when things go into .local/share. That's a personal preference though... is there a reason for going in there?

dns_server/src/config.rs
TheQuantumPhysicist: I would want to keep it optional and use a bogus email address if none is provided. This isn't the 90s anymore.

Can we technically make it optional?

dns_server/src/config.rs
```diff
@@ -0,0 +1,61 @@
...
+
+#[derive(Parser, Debug)]
+pub struct DnsServerConfig {
+    /// The path to the data directory
+    #[clap(long, default_value_t = default_data_dir())]
+    pub datadir: String,
+
+    /// Use testnet
+    #[clap(long)]
+    pub testnet: bool,
```
TheQuantumPhysicist, line 27: Please don't use boolean for this. We may have to instantiate multiple testnets. The right thing here is to use network ids that parse from strings. So "testnet" is testnet, and "mainnet" is mainnet. The default is mainnet.

dns_server/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+    let base_transport = TcpTransportSocket::new();
+    NoiseTcpTransport::new(stream_adapter, base_transport)
+}
+
+async fn run(config: Arc<DnsServerConfig>) -> Result<void::Void, error::DnsServerError> {
+    let (command_tx, command_rx) = mpsc::unbounded_channel();
+
+    let chain_config = if config.testnet {
+        Arc::new(common::chain::config::create_testnet())
+    } else {
```
TheQuantumPhysicist, line 43: And after using network ids, we match each possible string. If none matches, we panic.

dns_server/src/server/mod.rs
TheQuantumPhysicist: It's a bad habit and can lead to deadlocks. It's hard here because it's just guarding data, but as per our rules of thumb we don't this because people misunderstand it. But fine. Let's go with it here.

p2p/src/config.rs
TheQuantumPhysicist: I really remember it was there before... I wonder where it went. Anyway, thank you.

dns_server/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+    DefaultNetworkingService,
+};
+use tokio::sync::mpsc;
+
+mod config;
+mod crawler;
+mod dns_server;
+mod error;
+
+fn make_p2p_transport() -> NoiseTcpTransport {
```
TheQuantumPhysicist, line 32: I think this can carry risk. The decision of transport should be made in p2p and exported here. This is code duplication.

The bad thing with this is that in the future, if we change the wrapping protocol, this will break the DNS server.

dns_server/src/config.rs
pavel-kokolemin: There is a specification for this: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
Some random article from the Internet: https://0x46.net/thoughts/2019/02/01/dotfile-madness/
And it's mentioned in our issues: https://github.com/mintlayer/mintlayer-core/issues/498

Maybe we should just remove default from the `datadir` argument (make it mandatory)?

dns_server/src/config.rs
pavel-kokolemin: Sure, I made both the `nameserver` and `mbox` arguments optional (the same way it's done in bitcoin-seeder).

dns_server/src/config.rs
pavel-kokolemin: OK, updated. `clap::ValueEnum` worked nicely.

dns_server/src/main.rs
pavel-kokolemin: Should be OK now.

dns_server/src/main.rs
pavel-kokolemin: Updated, thanks.

pavel-kokolemin:
I added unit tests.

There is a risk that at some point the DNS server will not conform to the network protocol.
For example, by not responding to some requests from the sync manager (currently all requests are ignored).
After that, it would be banned by all nodes and stop working properly. I don't know what to do about it.

pavel-kokolemin:
I forgot to mention that with these changes the node will not bind to the listening port (unless explicitly specified with `--p2p-addr`). Curiously enough, all tests still pass.
I guess I should revert this first. I will mark the PR as draft for now.

dns_server/src/config.rs
TheQuantumPhysicist: I never realized people hated this... I actually like it. But sure, let's follow that standard.

But no, we can't just force people to use that directory without the option to put it somewhere else. There can be all kinds of issues, like diskspace or permission issues.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+        conn: N::ConnectivityHandle,
+        sync: N::SyncingMessagingHandle,
+        storage: S,
+        command_tx: mpsc::UnboundedSender<ServerCommands>,
+    ) -> Result<Self, DnsServerError> {
+        // Load all persistent addresses
+        let tx = storage.transaction_ro()?;
+
+        let storage_version = tx.get_version()?;
+        if storage_version.is_some() && storage_version != Some(STORAGE_VERSION) {
```
TheQuantumPhysicist, line 151: Why not use if let or match? This is more like the C++ way of doing things...

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+            return Err(DnsServerError::Other("Unexpected storage version"));
+        }
+
+        let mut addresses = BTreeMap::new();
+        for address in tx.get_addresses()?.iter().filter_map(|address| address.parse().ok()) {
+            Self::new_address(&mut addresses, address, true);
+        }
+        tx.close();
+
```
TheQuantumPhysicist, line 160: Please move all the selected code into a private function and return `addresses` from that.

TheQuantumPhysicist: This code can be reordered to separate concerns much better than it looks now. I see storage_version is used later... so my exact suggestion from above isn't directly possible, but this can still be improved. Just know that whenever I see `mut` and then usage the alarms go off in my head.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+    fn handle_conn_request(
+        &mut self,
+        _peer_id: N::PeerId,
+        request_id: N::PeerRequestId,
+        request: PeerManagerRequest,
+    ) {
+        match request {
+            PeerManagerRequest::AddrListRequest(_) => {
+                // Ignored
```
TheQuantumPhysicist, line 192: Why not provide an address list? I don't know if it's the right thing to do, but I'm interested in looking into this as a possibility.

pavel-kokolemin: For security the nodes only load addresses (in a list) from outbound connections and the DNS server does not accept inbound connections at all. So normally nobody would do that. And the DNS server has other means to sharing addresses (via the DNS protocol).
Later we might implement loading addresses from the nodes to the server.

TheQuantumPhysicist: What are the security concerns?

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+            }
+            PeerManagerRequest::AnnounceAddrRequest(AnnounceAddrRequest { address }) => {
+                // TODO: Rate limit AnnounceAddrRequest requests
```
TheQuantumPhysicist, line 195: Please add more information, as to why rate limit (what are the concerns) and against what (ip addresses?)

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+        &mut self,
+        address: N::Address,
+        peer_info: PeerInfo<N::PeerId>,
+        _receiver_address: Option<PeerAddress>,
+    ) {
+        let address_data = self
+            .addresses
+            .get_mut(&address)
+            .expect("address in the Connecting state must be known");
+        let valid = peer_info.network == self.config.network;
```
TheQuantumPhysicist, line 228: This doesn't seem to be a good strategy to do this check. The PeerInfo object should have something like `peer_info.is_compatible(&other_peer_info)` (or whatever else you find appropriate). There should not be any processing that's independent of p2p. It doesn't make sense that we're using the mintlayer p2p protocol, yet we invent our own rules here. Consider the effect of updates. 

I see here we're dealing with the crawler config. But being invasive like this is not the way to go. Please study this in a broader scope and let's see if we can come with a better compatibility check than just checking an arbitrary member variable in a struct.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+
+        if !valid {
+            log::debug!("invalid peer detected at {}", address.to_string());
```
TheQuantumPhysicist, line 231: this is `info!`, not `debug!`.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+            self.conn.disconnect(peer_info.peer_id).expect("disconnect must succeed");
+            Self::change_address_state(
+                &self.config,
+                &address,
+                address_data,
+                AddressState::Disconnected,
+                &mut self.storage,
+                &self.command_tx,
+            );
+            return;
```
TheQuantumPhysicist, line 241: Please put both these calls in a separate function. We don't update states within code logic.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+        Self::change_address_state(
+            &self.config,
+            &address,
+            address_data,
+            AddressState::Connected,
+            &mut self.storage,
+            &self.command_tx,
+        );
+        self.peers.insert(peer_info.peer_id, address);
+    }
```
TheQuantumPhysicist, line 253: Same here.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+                address_data,
+                AddressState::Disconnected,
+                &mut self.storage,
+                &self.command_tx,
+            );
+        }
+    }
+
+    fn handle_misbehaved(&mut self, _peer_id: N::PeerId, _error: P2pError) {
+        // Ignore all misbehave reports
```
TheQuantumPhysicist, line 299: Are we sure? Why not disconnect them or remove them from lists?

dns_server/src/crawler/mod.rs
pavel-kokolemin: This is because inbound peers should be treated as more dangerous than outbound peers. An attacker could easily target a specific node if it accepts inbound connections, connecting to it from different IPs, for example.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,560 @@
...
+        {
+            log::debug!("purge old (unreachable) address {}", address.to_string());
+
+            return false;
+        }
+
+        true
+    }
+
+    /// Address list maintenance
```
TheQuantumPhysicist, line 501: Please type more description to what's supposed to happen in this function.

dns_server/src/dns_server/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+fn handle_command(auth: &AuthorityImpl, command: ServerCommands) {
+    match command {
+        ServerCommands::AddAddress(IpAddr::V4(ip)) => {
+            auth.ip4.lock().expect("mutex must be valid (ipv4)").push(ip);
+        }
+        ServerCommands::AddAddress(IpAddr::V6(ip)) => {
+            auth.ip6.lock().expect("mutex must be valid (ipv6)").push(ip);
+        }
+        ServerCommands::DelAddress(IpAddr::V4(ip)) => {
+            auth.ip4.lock().expect("mutex must be valid (ipv4)").retain(|val| *val != ip);
```
TheQuantumPhysicist, line 278: Again, I see "mutex must be valid (ipv4)" repeated twice. Please make the repeated expect messages distinguishable.

p2p/src/lib.rs
```diff
@@ -164,12 +167,29 @@ pub async fn make_p2p<S: PeerDbStorage + 'static>(
     time_getter: TimeGetter,
     peerdb_storage: S,
 ) -> Result<Box<dyn P2pInterface>> {
-    let stream_adapter = NoiseEncryptionAdapter::gen_new();
-    let base_transport = net::default_backend::transport::TcpTransportSocket::new();
-    let transport = NoiseTcpTransport::new(stream_adapter, base_transport);
+    let transport = make_p2p_transport();
 
-    let p2p = P2p::<DefaultNetworkingService<NoiseTcpTransport>>::new(
+    let bind_addresses = if !p2p_config.bind_addresses.is_empty() {
```
TheQuantumPhysicist, line 172: Perhaps this logic should be moved to a separate function.

Also I keep thinking whether we want BindAddresses to be a trait with a Default impl coming from the application... I don't know. But seems like it's "its own thing".

TheQuantumPhysicist (commented):
Looking much better. I have to say that there are many things that are questionable in the crawler. This doesn't mean I have a better answer. But generally speaking, I don't like seeing these big state machines that just keep getting updated all the time. The crawler impl is large (500+ lines) and is not separated in anyway. Perhaps considering separating it into multiple modules that are tested on their own is a good idea. 

Also the tests have many helpers that are thrown within the tests. Maybe using a module for these is a cleaner way.

Please address the comments and I'll take another look.

dns_server/src/crawler/mod.rs
TheQuantumPhysicist: Gotcha.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Sure, updated.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Right. Updated. I hope it's OK now.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Sure, done.

dns_server/src/crawler/mod.rs
pavel-kokolemin: I added `is_compatible` to `PeerInfo`. For now it only checks `network`, but I added TODO to check version as well.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Sure, updated.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Sorry, but I don't like this. It seems like extra boilerplate to me. Let's keep it, please.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Also same.

dns_server/src/crawler/mod.rs
pavel-kokolemin: `NetworkingService` normally does not send `Misbehaved` messages (except for `ANNOUNCEMENT_MAX_SIZE` check), so I decided to keep things simpler.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Sure, updated.

dns_server/src/crawler/tests.rs
```diff
@@ -0,0 +1,419 @@
...
+
+    let time_getter = P2pTestTimeGetter::new();
+
+    (crawler, state, command_rx, time_getter)
+}
+
+async fn advance_time(
+    crawler: &mut Crawler<MockNetworkingService, DnsServerStorageImpl<storage::inmemory::InMemory>>,
+    time_getter: &P2pTestTimeGetter,
+    hours: u32,
```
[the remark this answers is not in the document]
pavel-kokolemin: Sure, updated.

dns_server/src/dns_server/mod.rs
pavel-kokolemin: Updated. I did not pay much attention because `expect` would also print the filename and line number, which is usually enough.

p2p/src/lib.rs
pavel-kokolemin: Sure, updated.

dns_server/src/dns_server/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+};
+
+use crate::{config::DnsServerConfig, error::DnsServerError};
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ServerCommands {
+    AddAddress(IpAddr),
+
+    DelAddress(IpAddr),
+}
```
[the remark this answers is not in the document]
pavel-kokolemin: Sorry, but this seems like a micro-optimization to me. Nodes normally become online/offline one by one, and we are probably talking about maximum of 1000 reachable nodes (I think the bitcoin network has less than 20000 reachable nodes). Let's not spend too much time on this yet.

pavel-kokolemin:
> The crawler impl is large (500+ lines) and is not separated in anyway. Perhaps considering separating it into multiple modules that are tested on their own is a good idea.

OK, I will make a note. But let's please keep it for now.

> Also the tests have many helpers that are thrown within the tests. Maybe using a module for these is a cleaner way.

Sure, updated.

dns_server/src/crawler/tests/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+    let crawler =
+        Crawler::<MockNetworkingService, _>::new(crawler_config, conn, sync, store, command_tx)
+            .unwrap();
+
+    let time_getter = P2pTestTimeGetter::new();
+
+    (crawler, state, command_rx, time_getter)
+}
+
+async fn advance_time(
```
TheQuantumPhysicist, line 81: Can you please provide description for this function that explains how it works?

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+        }
+
+        let mut addresses = BTreeMap::new();
+        // Load all persistent addresses
+        for address in tx.get_addresses()?.iter().filter_map(|address| address.parse().ok()) {
+            Self::new_address(&mut addresses, address, true);
+        }
+        tx.close();
+
+        if storage_version.is_none() {
```
TheQuantumPhysicist, line 179: Move into match

dns_server/src/crawler/mod.rs
pavel-kokolemin: Updated one more time.

dns_server/src/crawler/mod.rs
pavel-kokolemin: I looked one more time and still think that we should keep the code as it is.
That's because connected peers could only ever be created in the `handle_outbound_accepted` function.
So this new proposed function could only be called from one place - `handle_outbound_accepted`.

dns_server/src/crawler/tests/mod.rs
pavel-kokolemin: Sure, done.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Done.

dns_server/src/crawler/mod.rs
TheQuantumPhysicist: What if someone in the future needed another way to create/delete peers? Instead of calling this function that does two things, they have to copy two things, hopefully.

p2p/src/net/types/mod.rs
```diff
@@ -60,6 +60,13 @@ pub struct PeerInfo<P> {
     pub subscriptions: BTreeSet<PubSubTopic>,
 }
 
+impl<P> PeerInfo<P> {
+    pub fn is_compatible(&self, network: [u8; 4]) -> bool {
```
TheQuantumPhysicist, line 64: The `is_compatible` function should take a config, not a network, where multiple things can be tested.

p2p/src/lib.rs
```diff
@@ -156,6 +151,33 @@ impl subsystem::Subsystem for Box<dyn P2pInterface> {}
...
 
+pub type P2pNetworkingService = DefaultNetworkingService<NoiseTcpTransport>;
+
+pub fn make_p2p_transport() -> NoiseTcpTransport {
+    let stream_adapter = NoiseEncryptionAdapter::gen_new();
+    let base_transport = net::default_backend::transport::TcpTransportSocket::new();
+    NoiseTcpTransport::new(stream_adapter, base_transport)
+}
+
+fn get_p2p_bind_addresses(bind_addresses: &[String], p2p_port: u16) -> Result<Vec<SocketAddr>> {
```
TheQuantumPhysicist, line 162: A better signature here would be

```
fn get_p2p_bind_addresses<S: AsRef<str>>(bind_addresses: &[S], ...)
```

Then add `as_ref()` to addresses elements, and instead of `clone()`, use `to_owned()`.

dns_server/src/crawler/mod.rs
TheQuantumPhysicist: Don't you think this could open the door to flooding attacks or crash the node with huge messages?

dns_server/src/dns_server/mod.rs
TheQuantumPhysicist: That assumes you know what version you're testing against. It often happens that we don't know what version caused an issue with a user.

dns_server/src/crawler/mod.rs
```diff
@@ -0,0 +1,567 @@
...
+                Some(Ipv6Addr::from(addr.ip).into())
+            }
+            _ => None,
+        }
+    }
+
+    /// Update address state.
+    ///
+    /// The only place where the address state can be updated.
+    fn change_address_state(
```
TheQuantumPhysicist, line 389: I think this function should be implemented as a finite state machine. Let's talk about it.

TheQuantumPhysicist (commented):
This is getting much better. Thank you! Please do the little fixes that are left.

About the state thing, let's talk about it and see how we'll go about it.

Again, apologies for holding this long on the queue, but this is a new program, so it kind of deserves extra scruitiny :smile:

dns_server/src/crawler/mod.rs
TheQuantumPhysicist: So as agreed, this is gonna be the next PR.

dns_server/src/crawler/mod.rs
pavel-kokolemin: I added `create_outbound_peer` and `remove_outbound_peer` functions. Should be OK now.
Also renamed `peers` to `outbound_peers` to make it clearer that only outbound peers are stored there.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Huge messages are not a problem because the backend limits the maximum message size to 10MB.
Flooding attacks shouldn't be a problem either, because we don't do any processing.
Anyway, I added a TODO and created an issue (https://github.com/mintlayer/mintlayer-core/issues/699) for this.

p2p/src/net/types/mod.rs
pavel-kokolemin: Sure, updated.

p2p/src/lib.rs
pavel-kokolemin: Updated, thanks.

pavel-kokolemin:
I added `update_db` helper function to retry DB transactions in case of recoverable errors.

I also noticed some duplicated code when processing untrusted data from the network (`PeerAddress` values).
And some of the checks in the DNS server were missing because of this duplication.
So I refactored the `TransportAddress` trait to also do the validation. It should work, but I'm not sure if this is a good approach.

TheQuantumPhysicist (approved):

PR #688, opened 2023-02-08
Remove custom config file path and enforce it being under data dir + allow empty config files

### LINKED ISSUE #446: All configuration fields that loads from toml files should be based on Option

### LINKED ISSUE #498: Node directory handling

TheQuantumPhysicist:
- Configuration must stricly be in data directory
- Data directory cannot be specified in configuration
- All configuration options are optional. Options that aren't specified revert to default values
- Data directory is only created if the default path is used, to avoid mistakes where one uses different data directory due to typos

Closes #446 and closes #498

### DIFF
--- a/node/src/config_files/mod.rs
+++ b/node/src/config_files/mod.rs
@@ -45,7 +45,6 @@ pub struct NodeConfigFile {
 }
 
 impl NodeConfigFile {
-    /// Creates a new `Config` instance with the given data directory path.
     pub fn new() -> Result<Self> {
         Ok(Self {
             chainstate: None,
--- a/node/src/config_files/mod.rs
+++ b/node/src/config_files/mod.rs
@@ -22,12 +22,7 @@ mod chainstate_launcher;
 mod p2p;
 mod rpc;
 
-use std::{
-    fs,
-    net::SocketAddr,
-    path::{Path, PathBuf},
-    str::FromStr,
-};
+use std::{fs, net::SocketAddr, path::Path, str::FromStr};
 
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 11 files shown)


node/src/config_files/mod.rs
```diff
@@ -40,58 +35,62 @@ use self::{
...
-    pub chainstate: ChainstateLauncherConfigFile,
-    pub p2p: P2pConfigFile,
-    pub rpc: RpcConfigFile,
+    pub chainstate: Option<ChainstateLauncherConfigFile>,
+    pub p2p: Option<P2pConfigFile>,
+    pub rpc: Option<RpcConfigFile>,
 }
 
 impl NodeConfigFile {
     /// Creates a new `Config` instance with the given data directory path.
```
stanislav-tkach, line 48: I think this comment should be removed or updated because the data directory is no longer stored in the config file.

node/src/config_files/mod.rs
```diff
@@ -40,58 +35,62 @@ use self::{
...
-            rpc,
+            chainstate: None,
+            p2p: None,
+            rpc: None,
         })
     }
 
+    fn read_to_string_with_policy<P: AsRef<Path>>(config_path: P) -> Result<String> {
+        let config_as_str = if config_path.as_ref().exists() {
+            if config_path.as_ref().is_file() {
```
stanislav-tkach, line 59: Why is this check is important? If it is only needed for a better error reporting than I think that the standard errors are pretty descriptive (something like `kind: IsADirectory, message: "Is a directory"`).

node/src/options.rs
```diff
@@ -37,18 +37,12 @@ pub struct Options {
     #[clap(short, long = "datadir")]
     pub data_dir: Option<PathBuf>,
 
-    /// The path to the config file.
-    #[clap(short, long = "conf")]
-    config_path: Option<PathBuf>,
```
stanislav-tkach, line 42: Bitcoin accepts such command line option, but I also think that removing it makes things easier.

node/tests/cli.rs
```diff
@@ -116,107 +127,78 @@ fn read_config_override_values() {
...
+            .chainstate
+            .clone()
+            .unwrap_or_default()
+            .chainstate_config
+            .max_orphan_blocks,
         Some(max_orphan_blocks)
     );
     assert_eq!(
-        config.chainstate.chainstate_config.tx_index_enabled,
+        config.chainstate.clone().unwrap_or_default().chainstate_config.tx_index_enabled,
```
stanislav-tkach, line 151: Why `unwrap_or_default` instead of `unwrap` is used here?

stanislav-tkach (approved):

node/src/config_files/mod.rs
iljakuklic: I think there's a merit to mentioning the error is related to the config file. Maybe it is possible to attach extra info to the default error message.

node/src/config_files/mod.rs
TheQuantumPhysicist: Done

node/src/config_files/mod.rs
TheQuantumPhysicist: Correct. There's merit in that, it's just difficult to do both (write OS message + detect) and it'll be considered code bloat in a way. This error will almost never happen, so we catch it just in case and move on.

node/src/options.rs
TheQuantumPhysicist: Yeah, I don't think this is useful... I'm not convinced. It just complicates things for no good reason. Unless a usecase appears, I'd like to keep this off.

node/tests/cli.rs
TheQuantumPhysicist: Because this was the behavior before. But maybe I should change it back.

iljakuklic (approved):

node/src/config_files/mod.rs
stanislav-tkach: I still don't get it. 🙃  I think that the message produced by the following code is enough to understand the root of the issue:
```rust
    fn read_to_string_with_policy<P: AsRef<Path>>(config_path: P) -> Result<String> {
        let config_as_str = if config_path.as_ref().exists() {
            fs::read_to_string(config_path.as_ref()).context(format!(
                "Unable to read config file in {}",
                config_path.as_ref().display()
            ))?
        } else {
            "".into()
        };
        Ok(config_as_str)
    }
```

By the way, the `is_file` can return false because of a permission error and in that case the current message will be misleading.

TheQuantumPhysicist: Alright. Went with that.

PR #689, opened 2023-02-08
Initial implementation of wallet storage

erasmospunk:
I based this on the chainstate code so there is duplicate boilerplate code.


wallet/storage/src/internal/test.rs
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+type TestStore = super::Store<storage_sqlite::Sqlite>;
+
+#[test]
+fn test_storage_get_default_version_in_tx() {
+    utils::concurrency::model(|| {
```
TheQuantumPhysicist, line 22: I think you need to import the utils crate in Cargo.toml

storage/sqlite/src/lib.rs
```diff
@@ -283,19 +286,28 @@ impl Sqlite {
     }
 }
 
+/// Implements in memory database, useful for testing
+impl Default for Sqlite {
```
TheQuantumPhysicist, line 290: This is wrong.

We have a new() function that doesn't support in-memory storage, and a default trait that does that. This seems too abrupt.

Please remove the default, and modify the new() function to support both.

storage/sqlite/src/lib.rs
```diff
@@ -227,13 +227,13 @@ impl backend::BackendImpl for SqliteImpl {}
 
 #[derive(Eq, PartialEq, Clone, Debug)]
 pub struct Sqlite {
-    path: PathBuf,
+    path: Option<PathBuf>,
```
TheQuantumPhysicist, line 230: It's much better to use a custom enum for this. Something like:

```
enum SqliteBackend {
    InMemory,
    File(PathBuf),
}
```

wallet/storage/src/internal/test.rs
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+type TestStore = super::Store<storage_sqlite::Sqlite>;
+
+#[test]
+fn test_storage_get_default_version_in_tx() {
```
TheQuantumPhysicist, line 21: I think we'll need a little more tests :smile:

wallet/storage/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+[dependencies]
+common = { path = '../../common' }
+serialization = { path = "../../serialization" }
+storage = { path = '../../storage', features = ['inmemory'] }
+storage-sqlite = { path = "../../storage/sqlite" }
+
+utxo = { path = '../../utxo' }
+
+[dev-dependencies]
+test-utils = {path = '../../test-utils'}
```
TheQuantumPhysicist, line 18: You should add utils here to get that test to work

TheQuantumPhysicist (commented):
Looking good! There's a compilation error though + some fixes here and there + some more tests and we're good to go with this one! Cheers!

storage/sqlite/src/lib.rs
```diff
@@ -283,19 +286,28 @@ impl Sqlite {
...
-                std::io::ErrorKind::NotFound,
-                "Cannot find the parent directory".to_string(),
-            )
-            .into());
+        // Attempt to create the parent storage directory if using a file
+        if let Some(ref path) = self.path {
+            if let Some(parent) = path.parent() {
+                std::fs::create_dir_all(parent).map_err(error::process_io_error)?;
+            } else {
+                return Err(storage_core::error::Recoverable::Io(
```
stanislav-tkach, line 305: I had an impression that recoverable errors are errors that can be fixed by retrying the operation. In that case this error doesn't look like a recoverable one. On the other hand we have the `DbInit` recoverable error. Can someone explain this to me? 🙃

iljakuklic: In my conception, recoverable errors are ones that can be handled gracefully, irrecoverable are ones where very little can be done about (e.g. out of memory, etc.). It's fine to re-categorize existing errors if the current classification does not make sense. It's probably a good idea to add another category to separate "retry-able errors" as you suggest.

storage/sqlite/src/lib.rs
iljakuklic: Or leave `new` to open a database from file and add a separate `new_inmemory` constructor.

erasmospunk: I added a `new_in_memory` in `Sqlite` that does that.

However the `wallet/storage/src/internal/mod.rs` contains the following:

```rust
impl<B: Default + storage::Backend> Store<B> {
    /// Create a default storage (mostly for testing, may want to remove this later)
    pub fn new_empty() -> crate::Result<Self> {
        Self::new(B::default())
    }
}
```

I am not sure how I can create a new "in memory" database without parametrising over `Sqlite`. Alternatively I can move this impl to the test.rs file.

storage/sqlite/src/lib.rs
erasmospunk: Much better indeed

storage/sqlite/src/lib.rs
erasmospunk: This makes sense, I am changing it to a Fatal error.

wallet/storage/src/internal/test.rs
erasmospunk: I somehow commit but didn't push the changes

storage/sqlite/src/lib.rs
TheQuantumPhysicist: Default has to go... it doesn't make any sense.

storage/sqlite/src/lib.rs
TheQuantumPhysicist: The difference between recoverable and irrecoverable errors is simply whether we have a plan to recover. That's all. Do we know how to recover if we get error X? That's what decides it. I/O is usually an irrecoverable error, unless we know what to do and how to fix it, which quite frankly is rare. If a file doesn't work or you can't write on it, what do you really hope to do exactly? And I keep saying this: It could be a hardware failure. That's a game we shouldn't play.

I don't think we need a "retry-able errors"... that's what recoverable is for. The application should decide whether it needs to retry.

An example of a recoverable error is a transaction failing in lmdb due to size limitations. We can "try again" and expect a resize to have happened. We try a few times, then we give up and crash.

Another example is a transaction temporal serializability failure. Hence if tx A starts, then tx B, then both modify the same value. Only one of them can commit. The other should "retry" and hope no other conflicts happen.

storage/sqlite/src/lib.rs
TheQuantumPhysicist: I don't think we need the new_empty() function for the wallet. You can remove that.

storage/sqlite/src/lib.rs
```diff
@@ -25,6 +25,7 @@ use std::path::PathBuf;
 use std::sync::{Mutex, MutexGuard};
 
 use crate::queries::SqliteQueries;
+use crate::SqliteBackend::{File, InMemory};
```
TheQuantumPhysicist, line 28: Please don't do this. Use the fully qualified enum arm name.

wallet/storage/src/internal/test.rs
```diff
@@ -0,0 +1,69 @@
...
+        assert_eq!(vtx, 1, "Default storage version wrong");
+        assert_eq!(vtx, vst, "Transaction and non-transaction inconsistency");
+    })
+}
+
+#[cfg(not(loom))]
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn utxo_test(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 49: Better than prefixing or suffixing test names with `test`, it's better to write something about the test. Like `read_write_utxo`, or `read_write_utxo_in_db_transaction`. This applies to any test.

TheQuantumPhysicist (approved):

wallet/storage/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+license = "MIT"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+common = { path = '../../common' }
+serialization = { path = "../../serialization" }
+storage = { path = '../../storage', features = ['inmemory'] }
+storage-sqlite = { path = "../../storage/sqlite" }
+
```
stanislav-tkach, line 14: Nitpick: needless new line. 🙃

stanislav-tkach (approved):

wallet/storage/src/internal/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// Dump raw database contents
+    pub fn dump_raw(&self) -> crate::Result<storage::raw::StorageContents<Schema>> {
+        self.0.dump_raw().map_err(crate::Error::from)
+    }
+
+    /// Collect and return all utxos from the storage
+    #[allow(clippy::let_and_return)]
+    pub fn read_utxo_set(&self) -> crate::Result<BTreeMap<OutPoint, Utxo>> {
+        let db = self.transaction_ro()?;
+        let map = db.0.get::<db::DBUtxo, _>();
```
iljakuklic, line 49: Nitpick: rename the `db` variable to `dbtx`, since it's a transaction object not a database object. Also it clashes with the `db` module

iljakuklic (approved):
Looks good. There is a lot of code that is similar between the wallet storage crate and chainstate storage crates. We should have a look into the possibility of sharing/abstracting it.

storage/sqlite/src/lib.rs
```diff
@@ -225,15 +225,29 @@ impl ShallowClone for SqliteImpl {}
...
+
     /// New Sqlite database backend
     pub fn new(path: PathBuf) -> Self {
-        Self { path }
+        Self {
+            backend: SqliteBackend::File(path),
+        }
     }
 
     // fn open_db(self, desc: &MapDesc) -> storage_core::Result<Connection> {
```
azarovh, line 253: commented code

azarovh (approved):

storage/sqlite/src/lib.rs
```diff
@@ -225,26 +225,42 @@ impl ShallowClone for SqliteImpl {}
 
 impl backend::BackendImpl for SqliteImpl {}
 
+#[derive(Eq, PartialEq, Clone, Debug)]
+enum SqliteBackend {
```
iljakuklic, line 229: Hmm `SqliteBackend` may be a bit confusing name. Not that I have a better alternative though. `SqliteStorageMode`? Not sure, and maybe not too important for now, it's not a public type.

wallet/storage/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+common = { path = '../../common' }
+serialization = { path = "../../serialization" }
+storage = { path = '../../storage', features = ['inmemory'] }
+storage-sqlite = { path = "../../storage/sqlite" }
+utxo = { path = '../../utxo' }
+
+rstest = "0.16"
```
iljakuklic, line 16: Why is `rstest` in dependencies and not `dev-dependencies`?

wallet/storage/src/internal/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /// Collect and return all utxos from the storage
+    #[allow(clippy::let_and_return)]
+    pub fn read_utxo_set(&self) -> crate::Result<BTreeMap<OutPoint, Utxo>> {
+        let db_tx = self.transaction_ro()?;
+        let map = db_tx.0.get::<db::DBUtxo, _>();
+        let res = map.prefix_iter_decoded(&())?.collect::<BTreeMap<_, _>>();
+
+        Ok(res)
```
iljakuklic, line 52: Why did you decide to silence clippy rather than apply the suggestion?

iljakuklic (approved):
Looks good. Some minor issues and questions pointed out below.

wallet/storage/src/internal/mod.rs
azarovh: This code was taken from `chainstate/storage` and if the suggestion is applied the code won't compile, saying that `map` won't live long enough.

iljakuklic: This seems to work and may be worth considering:
```suggestion
    pub fn read_utxo_set(&self) -> crate::Result<BTreeMap<OutPoint, Utxo>> {
        let db_tx = self.transaction_ro()?;
        let dbmap = db_tx.0.get::<db::DBUtxo, _>();
        dbmap.prefix_iter_decoded(&()).map(Iterator::collect::<BTreeMap<_, _>>)
```

iljakuklic: Or even:
```suggestion
    pub fn read_utxo_set(&self) -> crate::Result<BTreeMap<OutPoint, Utxo>> {
        let db_tx = self.transaction_ro()?;
        db_tx.0.get::<db::DBUtxo, _>().prefix_iter_decoded(&()).map(Iterator::collect)
```

azarovh: nice

Issue #164, opened 2022-05-05
consensus: Remove Arc from OrphanBlocksPool


iljakuklic:
The `OrphanBlocksPool` has to have the ability to be moved between threads due to being used in `async` context. Currently it uses `Arc` to be usable from multiple threads. However, all aliasing instances of `Arc` are moved between threads together, so multi-threaded access is not actually needed. Using plain `Rc` also provides a compilation-time guarantee the objects are not accessed concurrently.

iljakuklic:
Introduced in #157

PR #690, opened 2023-02-08
Create a proxy that keeps the orphan blocks pool in one thread

TheQuantumPhysicist:
Solution to #164

Next, we'll replace OrphanBlocksPool in Chainstate with this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -434,6 +434,7 @@ dependencies = [
  "jsonrpsee",
  "logging",
  "mockall",
+ "oneshot",
  "pos_accounting",
  "rpc",
  "rstest",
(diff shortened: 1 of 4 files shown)


stanislav-tkach (approved):
Perhaps I'm missing something, but the `stop_control` atomic variable is never set and it looks like that the thread can be stopped by sending `None` anyway. Otherwise looking good.

TheQuantumPhysicist:
> Perhaps I'm missing something, but the `stop_control` atomic variable is never set and it looks like that the thread can be stopped by sending `None` anyway. Otherwise looking good.

A relic from dev work. I'll remove it. Thanks for noticing.

iljakuklic (approved):

PR #691, opened 2023-02-09
Use orphans proxy instead of orphans pool directly in chainstate to avoid having to move it among threads

### LINKED ISSUE #164: consensus: Remove Arc from OrphanBlocksPool

TheQuantumPhysicist:
Fixes #164


stanislav-tkach (approved):

chainstate/src/detail/orphan_blocks/orphans_proxy_impl.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.deref_mut()
+            .call_mut(move |o| o.add_block(block))
+            .recv()
+            .expect(RECV_ERR_MSG)
+    }
+
+    fn take_all_children_of(
+        &mut self,
+        block_id: &common::primitives::Id<common::chain::GenBlock>,
+    ) -> Vec<common::primitives::id::WithId<common::chain::Block>> {
```
azarovh, line 74: I'd import these types to make it less verbose

azarovh (approved):

chainstate/src/detail/orphan_blocks/orphans_proxy_impl.rs
TheQuantumPhysicist: Done

PR #695, opened 2023-02-13
Initial blockprod unit tests

alfiedotwtf:
Initial unit tests to blockprod before we move mining to outside the main executor

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -150,8 +150,6 @@ mod tests {
         )
         .expect("Error initializing Builder");
 
-        if !block_production.is_connected() {
-            panic!("Builder is not connected");
-        }
+        assert!(block_production.is_connected(), "Builder is not connected");
     }
 }
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -74,3 +74,84 @@ impl BlockProductionInterface for BlockProduction {
         !self.builder_tx.is_closed()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::super::*;
+    use super::*;
+    use tokio::time::{timeout, Duration};
+
+    #[tokio::test]
+    async fn test_stop() {
+        let (_manager, chain_config, chainstate, mempool) = setup_blockprod_test().await;
+
+        let (builder_tx, mut builder_rx) = mpsc::unbounded_channel();
+
+        let block_production = BlockProduction::new(
+            chain_config,
+            chainstate,
+            mempool,
+            Default::default(),
+            builder_tx,
+        )
+        .expect("Error initializing Builder");
+
+        block_production.stop().expect("Error stopping Builder");
+
+        let recv = timeout(Duration::from_millis(1000), builder_rx.recv());
+
+        tokio::select! {
+            msg = recv => match msg.expect("Builder timed out").expect("Error reading from Builder") {
+                BlockBuilderControlCommand::Stop => {},
+                _ => panic!("Invalid message recevied from Builder"),
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_start() {
+        let (_manager, chain_config, chainstate, mempool) = setup_blockprod_test().await;
+
+        let (builder_tx, mut builder_rx) = mpsc::unbounded_channel();
+
+        let block_production = BlockProduction::new(
+            chain_config,
+            chainstate,
+            mempool,
+            Default::default(),
+            builder_tx,
+        )
+        .expect("Error initializing Builder");
+
+        block_production.start().expect("Error starting Builder");
+
+        let recv = timeout(Duration::from_millis(1000), builder_rx.recv());
+
+        tokio::select! {
+            msg = recv => match msg.expect("Builder timed out").expect("Error reading from Builder") {
+                BlockBuilderControlCommand::Start => {},
+                _ => panic!("Invalid message received from Builder"),
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_is_connected() {
+        let (_manager, chain_config, chainstate, mempool) = setup_blockprod_test().await;
+
+        let (builder_tx, _builder_rx) = mpsc::unbounded_channel();
+
+        let block_production = BlockProduction::new(
+            chain_config,
+            chainstate,
+            mempool,
+            Default::default(),
+            builder_tx,
+        )
+        .expect("Error initializing Builder");
+
+        if !block_production.is_connected() {
+            panic!("Builder is not connected");
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


alfiedotwtf:
Ah. That last diff in `do_checks.sh` was a `cargo fmt` error not just a diff!

blockprod/src/lib.rs
```diff
@@ -89,3 +91,71 @@ pub async fn make_blockproduction(
 
     Ok(Box::new(result))
 }
+
+#[allow(clippy::unused_async)]
+pub async fn setup_blockprod_test() -> (Manager, Arc<ChainConfig>, ChainstateHandle, MempoolHandle)
```
azarovh, line 96: Does it really have to be exposed to non-test code? Maybe if many tests across the package require a complex setup all these tests can be moved to a separate module.

Also it should fix `pub` access level.

chainstate/src/detail/chainstateref/tx_verifier_storage.rs
```diff
@@ -269,9 +269,9 @@ impl<'a, S: BlockchainStorageWrite, O: OrphanBlocksRef, V: TransactionVerificati
                     .db_tx
                     .get_block_index(&id)
                     .map_err(TransactionVerifierStorageError::from)?
-                    .ok_or(
-                        TransactionVerifierStorageError::GenBlockIndexRetrievalFailed(id.into()),
-                    )?;
+                    .ok_or_else(|| {
```
azarovh, line 272: not related to your PR but just to point out the problem. I remember that I definitely run `do_checks` on this code. The difference between local runs and CI for clippy becomes annoying

azarovh (approved):

blockprod/src/lib.rs
alfiedotwtf: Yeah I wasn't too sure about this one, but I did find examples elsewhere in our code so kept it exposed. But I've now fixed it.

chainstate/src/detail/chainstateref/tx_verifier_storage.rs
alfiedotwtf: Yeah I had wild differences and that's why I made that little `python3` change so I can now run it locally :)

blockprod/src/lib.rs
azarovh: I mean this function should be under `#[cfg(test)]`, otherwise it can be misused in production and also require all these workarounds with dead and unused.

alfiedotwtf: hmm.. I was under the assumption code in `#[cfg(test)]` wasn't shared to lower modules. I'll give it a try and see if it works

azarovh: it might not, I'm not sure. But that's why I suggested moving everything to a separate module, something like `tests`

blockprod/src/lib.rs
```diff
@@ -89,3 +91,70 @@ pub async fn make_blockproduction(
 
     Ok(Box::new(result))
 }
+
+#[allow(clippy::unused_async, dead_code)]
+async fn setup_blockprod_test() -> (Manager, Arc<ChainConfig>, ChainstateHandle, MempoolHandle) {
```
iljakuklic, line 96: If this is only used for testing, can we guard it by `#[cfg(test)]` and move the in-memory storage dependency to `dev-dependencies`?

iljakuklic (approved):

blockprod/src/lib.rs
alfiedotwtf: Yep, I've managed to find a way to move it all into `#[cfg(test)]`. Just tidying up my PR now

blockprod/src/detail/mod.rs
```diff
@@ -69,4 +69,88 @@ impl BlockProductionInterface for BlockProduction {
...
+        .expect("Error initializing Builder");
+
+        block_production.stop().expect("Error stopping Builder");
+
+        let recv = timeout(Duration::from_millis(1000), builder_rx.recv());
+
+        tokio::select! {
+            msg = recv => match msg.expect("Builder timed out").expect("Error reading from Builder") {
+                BlockBuilderControlCommand::Stop => {},
+                _ => panic!("Invalid message recevied from Builder"),
```
TheQuantumPhysicist, line 107: recevied

blockprod/src/lib.rs
```diff
@@ -89,3 +89,78 @@ pub async fn make_blockproduction(
...
+    use chainstate_storage::inmemory::Store;
+    use common::chain::{config::create_unit_test_config, ChainConfig};
+    use mempool::{MempoolHandle, MempoolSubsystemInterface};
+    use subsystem::Manager;
+
+    use super::*;
+
+    #[allow(clippy::unused_async)]
+    pub async fn setup_blockprod_test(
+    ) -> (Manager, Arc<ChainConfig>, ChainstateHandle, MempoolHandle) {
```
iljakuklic, line 105: What's the business with `unused_async` here? Would it be possible to just remove `async`?

alfiedotwtf: It's complaining because we don't actually do any async work in here, but from where it's called, it definitely needs it. I tried removing it and funnily enough Cargo tells me to add the `async` keyword in

TheQuantumPhysicist: This is probably because `.await` is used. If you remove async from here and `.await` from all the callers, this won't be needed.

blockprod/src/lib.rs
```diff
@@ -89,3 +89,78 @@ pub async fn make_blockproduction(
...
+        let mempool = manager
+            .add_subsystem_with_custom_eventloop("mempool", move |call, shutdn| {
+                mempool.run(call, shutdn)
+            });
+
+        (manager, chain_config, chainstate, mempool)
+    }
+
+    #[tokio::test]
+    async fn test_makeblockproduction() {
```
TheQuantumPhysicist, line 139: Nitpick: `test_make_blockproduction`

blockprod/src/lib.rs
iljakuklic: Can you remove `.await` from the call sites? I don't see why it's needed there

do_checks.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 cargo fmt --check &&
-python .github/scripts/codecheck.py &&
+PYTHON=$(which python || which python3)
```
TheQuantumPhysicist, line 3: Python never stops to piss me off... it seems I'll rewrite this script in rust and move on with my life.

TheQuantumPhysicist (approved):
Looks good. Please fix the little issues here and there and then feel free to merge.

blockprod/src/lib.rs
alfiedotwtf: Gah! I know I sprinkled them all here for some reason, but yes it works without those await tails. Removed. Nice!

do_checks.sh
alfiedotwtf: Xcode wanted to install python each time I tried to running non-existent `python`, and then failed installing it each time lol... so big hammer meets nail!

PR #696, opened 2023-02-13
Fix ShallowClone soundness and introduce ConstValue

### LINKED ISSUE #645: ShallowClone trait is broken
The `ShallowClone` trait on a struct is supposed to signify that cloning the struct will never yield new copies of any members that can be mutated at a later stage. This contract is currently broken in the way `ShallowClone` is used. 

For example, `LmdbImpl` implements `ShallowClone`, but it's allowed to put variables inside of it that can be mutated and Cloned. If we add a `u32` or a `bool`, nothing will break (a successful compile will happen). This basically means that cloning that struct will cause undefined behavior, because the "shallow" clone is not a shallow clone anymore, but it holds an independent value that can be mutated independently.

`ShallowClone` should be some kind of auto trait that enforces `ShallowClone` on all members of a struct. It should also cause an error if any type underneath supports deep Cloning.

TheQuantumPhysicist:
Fixes #645

### DIFF
--- a/utils/src/const_value.rs
+++ b/utils/src/const_value.rs
@@ -14,12 +14,13 @@
 // limitations under the License.
 
 use std::{
+    borrow::Borrow,
     fmt::{Debug, Display},
     hash::{Hash, Hasher},
     ops::Deref,
 };
 
-/// A wrapper for a type that is not supposed to be modified, no matter one
+/// A wrapper for a type that is not supposed to be modified, no matter what
 /// This particularly solves the problem of not being able to mark member variables as const in Rust
 pub struct ConstValue<T> {
     value: T,
@@ -28,6 +29,10 @@ pub struct ConstValue<T> {
 impl<T> ConstValue<T> {
     pub fn new(value: T) -> Self {
         Self { value }
+    }
+
+    pub fn take(self) -> T {
+        self.value
     }
 }
 
@@ -95,6 +100,12 @@ impl<T: Hash> Hash for ConstValue<T> {
 
 impl<T> AsRef<T> for ConstValue<T> {
     fn as_ref(&self) -> &T {
+        &self.value
+    }
+}
+
+impl<T> Borrow<T> for ConstValue<T> {
+    fn borrow(&self) -> &T {
         &self.value
     }
 }
--- a/utils/src/const_value.rs
+++ b/utils/src/const_value.rs
@@ -1,0 +1,108 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


utils/src/shallow_clone.rs
```diff
@@ -17,6 +17,12 @@
 
 use std::marker::PhantomData;
 
+#[cfg(not(loom))]
+use std::sync::Arc;
+
+#[cfg(loom)]
+use loom::sync::Arc;
```
iljakuklic, line 24:
```suggestion
use crate::sync::Arc;
```

utils/src/const_value.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.value.hash(state)
+    }
+}
+
+impl<T> AsRef<T> for ConstValue<T> {
+    fn as_ref(&self) -> &T {
+        &self.value
+    }
+}
```
iljakuklic, line 100: Also implement `Borrow`?

utils/src/const_value.rs
```diff
@@ -0,0 +1,108 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{
+    fmt::{Debug, Display},
+    hash::{Hash, Hasher},
+    ops::Deref,
+};
+
+/// A wrapper for a type that is not supposed to be modified, no matter one
```
iljakuklic, line 22: Unfinished sentence?

utils/src/const_value.rs
```diff
@@ -0,0 +1,108 @@
...
+/// This particularly solves the problem of not being able to mark member variables as const in Rust
+pub struct ConstValue<T> {
+    value: T,
+}
+
+impl<T> ConstValue<T> {
+    pub fn new(value: T) -> Self {
+        Self { value }
+    }
+}
```
iljakuklic, line 32: Should there also be a method to take the inner value out of the `ConstValue` wrapper?

utils/src/const_value.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use std::{
+    fmt::{Debug, Display},
+    hash::{Hash, Hasher},
+    ops::Deref,
+};
+
+/// A wrapper for a type that is not supposed to be modified, no matter one
+/// This particularly solves the problem of not being able to mark member variables as const in Rust
+pub struct ConstValue<T> {
```
iljakuklic, line 24: I understand this wrapper works on best effort basis. It blocks calling `&mut self` methods on the inner value.

As far as I can see, one can still do something like:
```rust
fn modify_const_value(cv: &mut ConstValue<u32>) {
    *cv = ConstValue::new(42);
}
```

Same applies if you're given a `&mut` of a struct containing a `ConstValue`. Also anything with interior mutability can still be modified.

utils/src/shallow_clone.rs
```diff
@@ -15,20 +15,67 @@
...
+    }
+}
+
+impl<T> ShallowClone for PhantomData<T> {
+    fn shallow_clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T: Clone> ShallowClone for ConstValue<T> {
```
iljakuklic, line 74: A comment explaining why it's okay to `shallow_clone` a `ConstValue` would be nice

iljakuklic (approved):
Looks good. Some minor suggestions and comments below

utils/src/const_value.rs
TheQuantumPhysicist: Done.

utils/src/const_value.rs
TheQuantumPhysicist: Done

iljakuklic (approved):

PR #698, opened 2023-02-13
Fix lmdb test dir cleanup

TheQuantumPhysicist:
This fixes the issue that the TempDir object in the test is destroyed early, which leads to lack of cleanup.


muursh (approved):

iljakuklic (approved):

iljakuklic:
Should check whether `sqlite` backend has the same issue

TheQuantumPhysicist:
> Should check whether `sqlite` backend has the same issue

`into_path()` only exists in this file.

PR #701, opened 2023-02-15
Move PoolId and DelegationId to common

azarovh:
I'll need to use `PoolId` in `consensus` package in upcoming PR with PoS. To avoid `consensus` dependency on `pos_accounting` decided to move these types to `common`.

Publishing as a separate PR to avoid spoiling PoS PR with boilerplate.

### DIFF
--- a/common/src/chain/pos.rs
+++ b/common/src/chain/pos.rs
@@ -13,25 +13,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub mod block;
-pub mod config;
-pub mod gen_block;
-pub mod genesis;
-mod mlt;
-mod pos;
-mod pow;
-pub mod tokens;
-pub mod transaction;
-mod upgrades;
+use typename::TypeName;
 
-pub use signed_transaction::SignedTransaction;
-pub use transaction::*;
+use crate::primitives::Id;
 
-pub use block::Block;
-pub use config::ChainConfig;
-pub use gen_block::{GenBlock, GenBlockId};
-pub use genesis::Genesis;
-pub use mlt::Mlt;
-pub use pos::{DelegationId, PoolId};
-pub use pow::PoWChainConfig;
-pub use upgrades::*;
+#[derive(Eq, PartialEq, TypeName)]
+pub enum Pool {}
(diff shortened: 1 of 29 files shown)


azarovh:
On the other hand, `consensus` dependency on `pos_aacounting` might not be that big of a deal. Moving to draft for now to think about it

azarovh:
`PoolId` will also be used on `common::consensus_data::PoSData`, so yeah seems like the right place

stanislav-tkach (approved):

common/src/chain/pos.rs
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use typename::TypeName;
+
+use crate::primitives::Id;
+
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, TypeName)]
+pub struct Pool;
```
iljakuklic, line 21: Not suggesting a change, just thinking aloud. Maybe these types that are only used as phantom types should be void types rather than unit types? There's no point constructing them anyway. Also why all the `derive` impls? If they are required, the trait bounds on `Id` are probably too strict.

iljakuklic (approved):
+1

common/src/chain/pos.rs
azarovh: I wanted PoolId and DelegationId to be different types so that they cannot be accidentally interchanged.  By void types what exactly do you mean?

stanislav-tkach: `enum Pool {}` would be a "void" type because it cannot be instantiated.

azarovh: > should be void types rather than unit types

done

> Also why all the derive impls?

removed ordering

TheQuantumPhysicist (approved):
I'm not happy with moving more things inside common, but I currently don't have a better solution because Block depends on PoSData and PoSData depends on PoolId, so moving PoolId to consensus isn't possible because it depends PoSData needs it and that will create a circular dependency.

Until we find a solution with a better taxonomy for this software, I'm OK with this.

PR #702, opened 2023-02-15
Use strong type for fees in the mempool

### LINKED ISSUE #476: Use `Fee` type instead of `Amount` in mempool for tx fees

TheQuantumPhysicist:
Closes #476

### DIFF
--- a/mempool/src/pool/fee.rs
+++ b/mempool/src/pool/fee.rs
@@ -1,0 +1,37 @@
+use std::{
+    iter::Sum,
+    ops::{Add, Sub},
+};
+
+use common::primitives::Amount;
+use utils::newtype;
+
+newtype! {
+    #[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Copy, Clone)]
+    pub struct Fee(Amount);
+}
+
+impl Add for Fee {
+    type Output = Option<Self>;
+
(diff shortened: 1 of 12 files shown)


mempool/src/pool/fee.rs
```diff
@@ -0,0 +1,52 @@
...
+    iter::Sum,
+    ops::{Add, Sub},
+};
+
+use common::primitives::Amount;
+use utils::newtype;
+
+newtype! {
+    #[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Copy, Clone)]
+    pub struct Fee(Amount);
```
azarovh, line 26: imo it looks generic enough to be moved to `common` and reused in `tx_verifier`, not insisting though

azarovh (approved):

mempool/src/pool/fee.rs
TheQuantumPhysicist: Fee doesn't really mean anything other than a wrapper for an Amount that's type-safe. I would agree with you and do this if there's a meaning other than the lexical one of the word.

muursh (approved):
lgtm

stanislav-tkach (approved):

iljakuklic (approved):

Issue #697, opened 2023-02-13
Proof of stake basics


azarovh:
This issue is created to track progress and to document issues related to the proof of stake.
Ordered by highest priority first.

- [x] Finalize #260 with initial PoS implementation
- [x] Use pool balance for PoS target check
- [x] Store EpochData
- [x] Unit tests
- [x] Add Uint512 type
- [x] Spending stake output on block production #734
- [x] Support target calculation and retargeting
- [x] Decommission pool
- [x] Get rid of `OutputValue` in `TxOutput`, and make `TxOutput` itself an enum instead of using `OutputPurpose`, because we expect some output purpose types not to have value associated with them; e.g., accounting utxos.
- [x] (low priority due to release deadline) TransactionVerifier needs another look to be made smaller, more manageable and more testable
- [x] Stake delegations subscription
- [x] Stake delegations cancelling
- [x] #834
- [x] Reconsider the longest chain rule #827 
- [x] #752
- [x] #787

PR #705, opened 2023-02-15
Add proof of stake primitives

azarovh:
This is a continuation of work started in #260, where PoS was first added. The work is not complete and many aspects are still to be added. I will try to track high-level subtasks in #697 

This PR covers calculating randomness via VRF and comparing it with a target to prove a block winner, taking pool balance into account, integrating PoS consensus into chainstate and unit tests. 

I intentionally left out some kernel checks from #260, because I think they better be covered when implementing proper spending of a kernel. For example, checks if the kernel is spent, is ancestor valid etc.

Closes #260


chainstate/src/detail/tx_verification_strategy/default_strategy.rs
```diff
@@ -115,6 +106,15 @@ impl TransactionVerificationStrategy for DefaultTransactionVerificationStrategy
             .check_block_reward(block, Fee(total_fees), Subsidy(block_subsidy))
             .log_err()?;
 
+        let reward_fees = tx_verifier
```
TheQuantumPhysicist, line 109: Why move this here?

azarovh: `connect_transactable` spends the kernel, as a result, `check_block_reward` won't pass. So I switched the order.  It was not a problem before because we didn't have inputs in block reward

common/src/chain/block/consensus_data.rs
```diff
@@ -52,22 +51,28 @@ impl ConsensusData {
...
         kernel_inputs: Vec<TxInput>,
         kernel_witness: Vec<InputWitness>,
+        stake_pool_id: PoolId,
+        vrf_data: VRFReturn,
         bits: Compact,
     ) -> Self {
         Self {
             kernel_inputs,
             kernel_witness,
+            stake_pool_id,
```
TheQuantumPhysicist, line 74: There's no need to manually put the pool id in the PoSData. The pool id can be derived from the block/transaction by reading its first input.

We can input the pool id when we want to stake, for example, because then we have to say "what output we're staking".

chainstate/test-suite/src/tests/pos_processing_tests.rs
```diff
@@ -0,0 +1,401 @@
...
+    let difficulty =
+        Uint256([0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF]);
+    let prev_randomness = tf.chainstate.get_chain_config().initial_randomness();
+    let block_timestamp = BlockTimestamp::from_duration_since_epoch(tf.current_time());
+    let vrf_transcript = construct_transcript(1, &prev_randomness, block_timestamp);
+    let vrf_data = vrf_sk.produce_vrf_data(vrf_transcript.into());
+    let pos_data = PoSData::new(
+        vec![TxInput::from_outpoint(stake_pool_outpoint)],
+        vec![InputWitness::NoSignature(None)],
+        pool_id,
```
TheQuantumPhysicist, line 130: You can in fact process the transaction and then check if the pool id that's used to store the accounting data matches the one you expect.

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
TheQuantumPhysicist: I see. The issue here is that we should verify the reward before getting into the block because the reward is the key part of a block's validity. Maybe throw in a TODO here and explain this problem.

chainstate/test-suite/src/tests/pos_processing_tests.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        let res = tf
+            .make_block_builder()
+            .with_consensus_data(ConsensusData::PoS(pos_data))
+            .build_and_process()
+            .unwrap_err();
+
+        assert_eq!(res, expected_error);
+    }
+}
```
TheQuantumPhysicist, line 340: Always when you test multiple invalid cases, end it with a valid case to ensure that your reference case works.

common/src/chain/block/consensus_data.rs
TheQuantumPhysicist: Maybe I'm wrong here. Let's talk about it in chat.

TheQuantumPhysicist (approved):
Looking good. I'd like to invite others to review this PR.

common/src/chain/config/builder.rs
```diff
@@ -129,6 +129,7 @@ impl Builder {
             max_future_block_time_offset: super::DEFAULT_MAX_FUTURE_BLOCK_TIME_OFFSET,
             epoch_length: super::DEFAULT_EPOCH_LENGTH,
             sealed_epoch_distance_from_tip: super::DEFAULT_SEALED_EPOCH_DISTANCE_FROM_TIP,
+            initial_randomness: H256::zero(),
```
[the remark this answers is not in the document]
muursh: "a little"

consensus/src/pos/error.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2021 RBB S.r.l
```
muursh, line 1: 2021?

muursh: Quite a few 2021s

muursh (approved):
All good to me

stanislav-tkach (approved):

chainstate/src/detail/chainstateref/epoch_seal.rs
```diff
@@ -110,3 +111,44 @@ fn rollback_epoch_seal<S: BlockchainStorageWrite>(
...
+            }
+        }
+        BlockStateEventWithIndex::Disconnect(tip_height) => {
+            let disconnected_tip = tip_height.next_height();
+            if chain_config.is_last_block_in_epoch(&disconnected_tip) {
+                db_tx
+                    .del_epoch_data(chain_config.epoch_index_from_height(&disconnected_tip))
+                    .log_err()?;
+            }
+        }
```
iljakuklic, line 150: This makes my head spin somewhat. The if block is entered if the next block after `top_height` is the last block? So the epoch switches at block heights (epoch_length - 1 + K * epoch_length) where K is a non-negative integer, and tip_height refers to the tip after disconnection? That's how I see it making some sense, is it correct? At the very least a comment explaining why this is not an off-by-one (or two?) error would be welcome.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -926,4 +913,30 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
...
+        epoch_seal::update_epoch_seal(
+            &mut self.db_tx,
+            self.chain_config,
+            epoch_seal::BlockStateEvent::Disconnect(tip_height),
+        )?;
+        epoch_seal::update_epoch_data(
+            &mut self.db_tx,
+            self.chain_config,
+            epoch_seal::BlockStateEventWithIndex::Disconnect(tip_height),
+        )
```
iljakuklic, line 940: Do these two function calls commute? My instinct would be fort hem to be applied in the opposite order compared to `connect_tip`.

chainstate/types/src/pos_randomness.rs
```diff
@@ -0,0 +1,104 @@
...
+        block_height: &BlockHeight,
+        header: &BlockHeader,
+        previous_randomness: Option<&PoSRandomness>,
+        kernel_output: &TxOutput,
+        pos_data: &PoSData,
+    ) -> Result<Self, PoSRandomnessError> {
+        use crypto::hash::StreamHasher;
+
+        let prev_randomness =
+            previous_randomness.cloned().unwrap_or_else(|| Self::at_genesis(chain_config));
```
iljakuklic, line 60: Minor nitpick: It would be more flexible to take `Option<PosRandomness>` by value so the cloning performed here does not need to happen if caller already owns the value.

consensus/src/pos/kernel.rs
```diff
@@ -0,0 +1,41 @@
...
+        ConsensusPoSError::NoKernel,
+    );
+    // in general this should not be an issue, but we have to first study this security model with one kernel
+    ensure!(
+        pos_data.kernel_inputs().len() == 1,
+        ConsensusPoSError::MultipleKernels,
+    );
+
+    let kernel_outpoint =
+        pos_data.kernel_inputs().get(0).ok_or(ConsensusPoSError::NoKernel)?.outpoint();
```
iljakuklic, line 37: I am wondering if it would be cleaner to use pattern matching here, especially considering the `ok_or` case is already covered by the first `ensure!`.

Very roughly something like:
```rust
match pos_data.kernel_inputs() {
    [] => Err(NoKernel),
    [kernel_input] => todo!("handle happy case"),
    _ => Err(MultipleKernels),
}
```

It may be good idea to first change `kernel_inputs()` to return a slice rather than a vec reference.

iljakuklic (commented):
Looks fine to me overall. However I think some parts deserve some clarification.

alfiedotwtf (approved):

common/src/chain/config/mod.rs
```diff
@@ -192,19 +193,26 @@ impl ChainConfig {
...
     #[must_use]
     pub fn epoch_index_from_height(&self, height: &BlockHeight) -> EpochIndex {
         let height: u64 = (*height).into();
         height / self.epoch_length
     }
 
+    pub fn is_last_block_in_epoch(&self, height: &BlockHeight) -> bool {
+        let next_height: u64 = height.next_height().into();
+        next_height % self.epoch_length() == 0
+    }
```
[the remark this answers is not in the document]
azarovh: I believe it is correct. For example, if epoch takes 3 blocks, given the sequence: 0 1 2 3 4 5, blocks at heights 2 and 5 would be the last. 
Your approach of comparing whether the current and the next blocks are in different epochs would work too, but modulus seems like a good option to me.
Also you can take a look at unit tests in the file below.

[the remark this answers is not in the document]
azarovh: btw, maybe the name is confusing. The last means end - 1, not the only one left.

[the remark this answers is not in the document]
alfiedotwtf: Ah, in that case yeah I think I misunderstood the purpose. All good them :)

consensus/src/pos/kernel.rs
azarovh: nice! didn't know you can match a slice

chainstate/src/detail/chainstateref/mod.rs
azarovh: they don't, so the order doesn't matter

chainstate/src/detail/chainstateref/epoch_seal.rs
azarovh: thanks for noticing, there is a bug indeed! fixed and covered with tests

chainstate/test-framework/src/tx_verification_strategy/disposable_strategy.rs
```diff
@@ -126,6 +117,15 @@ impl TransactionVerificationStrategy for DisposableTransactionVerificationStrate
...
             .log_err()?;
 
+        let reward_fees = base_tx_verifier
+            .connect_transactable(
+                block_index,
+                BlockTransactableWithIndexRef::BlockReward(block, block_reward_tx_index),
+                &median_time_past,
+            )
+            .log_err()?;
+        debug_assert!(reward_fees.is_none());
```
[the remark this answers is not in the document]
azarovh: tbh I don't know. I'd say it's more like a temp check because will reconsider this rule soon

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
azarovh: I think we can separate txs checks and fee collection. The only downside is that we will have to enumerate and perform some calculations twice. I'll place a TODO to look into this later

chainstate/types/src/vrf_tools.rs
```diff
@@ -0,0 +1,83 @@
...
+        .attach(
+            b"Randomness",
+            TranscriptComponent::RawData(random_seed.as_bytes().to_vec()),
+        )
+        .attach(
+            b"Slot",
+            TranscriptComponent::U64(block_timestamp.as_int_seconds()),
+        )
+        .attach(b"EpochIndex", TranscriptComponent::U64(epoch_index))
+        .finalize()
```
iljakuklic, line 50: Minor: the hard-coded strings here could be named constants

iljakuklic (approved):
+1

chainstate/src/detail/tx_verification_strategy/default_strategy.rs
```diff
@@ -111,10 +102,21 @@ impl TransactionVerificationStrategy for DefaultTransactionVerificationStrategy
             })
             .log_err()?;
 
+        // TODO: reconsider the order of connect txs and check block reward.
```
TheQuantumPhysicist, line 105: Always when writing a TODO, also add why this TODO came to existence. Please add that this was X before, and it was changed to Y because of Z, and so on.

PR #708, opened 2023-02-16
Make the DNS server core deterministic

pavel-kokolemin:
So I tried a different approach in the DNS server, I separated core and IO layers. The core is now fully deterministic and much easier to test. New randomized tests should be able to catch corner cases. Failed tests could be easily reproduced from the generated seed.
I suppose the same could be done in some parts of p2p to improve code quality there.


dns_server/src/crawler/core.rs
```diff
@@ -0,0 +1,546 @@
...
+                    3 => age > Duration::from_secs(3600),
+                    4 => age > Duration::from_secs(3 * 3600),
+                    5 => age > Duration::from_secs(6 * 3600),
+                    6 => age > Duration::from_secs(12 * 3600),
+                    _ => age > Duration::from_secs(24 * 3600),
+                }
+            }
+            AddressState::Disconnected => {
+                // The address was never reachable, try to connect just once
+                address_data.fail_count == 0
```
TheQuantumPhysicist, line 403: Why does this variable even exist for the disconnected state?

dns_server/src/crawler/core.rs
```diff
@@ -0,0 +1,546 @@
...
+    /// Connection state
+    state: AddressState,
+
+    /// Last time when the connection state was updated
+    state_updated_at: Duration,
+
+    /// The number of failed connection attempts.
+    /// New connection attempts are made after a progressive backoff time.
+    /// Resets to 0 when an outbound connection to the address is successful.
+    fail_count: u32,
```
TheQuantumPhysicist, line 68: How about we move this inside `AddressState::Disconnected` and `AddressState::Connecting`?

pavel-kokolemin: Sure, updated, thanks.

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// Addresses that were once reachable are stored in the DB.
+    was_reachable: bool,
+
+    /// Whether the address was added from the command line
+    user_added: bool,
+}
+
+#[derive(Clone)]
+pub struct CrawlerCoreConfig {
+    pub p2p_port: u16,
```
TheQuantumPhysicist, line 103: Aren't we not supposed to open a port for DNS in p2p?

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,582 @@
...
+
+        let addresses = loaded_addresses
+            .union(&added_addresses)
+            .map(|addr| {
+                (
+                    addr.clone(),
+                    AddressData {
+                        state: AddressState::Disconnected { fail_count: 0 },
+                        state_updated_at: now,
+                        was_reachable: loaded_addresses.contains(addr),
```
TheQuantumPhysicist, line 165: This is such a weird way for initialization... why not initialize each alone, then chain them together instead of doing a union and finding out where they come from?

pavel-kokolemin: It was done such way before, but user added addresses and addresses from the DB could be in both lists at the same time. And when you construct `BTreeMap` from such a chained iterator, some of the flags get lost (`was_reachable` or `user_added` depending on the chaining order). This caused problems in tests.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: The server does not open any listening ports. This is the default p2p port on which nodes normally accept inbound connections. The name was a bit confusing. I have renamed it and added a comment.

dns_server/src/crawler/core/mod.rs
TheQuantumPhysicist: I see.

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+    DbStoreAddress(N::Address),
+    DbEraseAddress(N::Address),
+
+    Connect(N::Address),
+    Disconnect(N::PeerId),
+    SendResponse(N::PeerId, PeerManagerMessage),
+}
+
+pub trait CoreCallback<N: NetworkingService> {
+    fn exec(&mut self, command: CoreCommand<N>);
```
TheQuantumPhysicist, line 146: Why not use something like `FnMut()` and alias it, instead of creating a trait?

TheQuantumPhysicist: To be clear, I don't mind. Was just wondering.

dns_server/src/crawler/mod.rs
```diff
@@ -14,127 +14,71 @@
...
+    last_core_timer: Instant,
+
+    /// CrawlerCore
+    core: CrawlerCore<N, CrawlerCallback<N, S>>,
+
+    /// CrawlerCallback implementation
+    callback: CrawlerCallback<N, S>,
+}
+
 impl<N: NetworkingService + 'static, S: DnsServerStorage> Crawler<N, S>
```
TheQuantumPhysicist, line 82: I'm not sure this `'static` is necessary.

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    /// Last time when the connection state was updated
+    state_updated_at: Duration,
+
+    /// Whether the address was reachable at least once.
+    /// Addresses that were once reachable are stored in the DB.
+    was_reachable: bool,
+
+    /// Whether the address was added from the command line
+    user_added: bool,
```
TheQuantumPhysicist, line 98: Since this is a constant value, please use `ConstValue` for this.

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+    /// Time of some monotonic timer, started from 0
+    now: Duration,
+
+    /// Crawler config
+    config: CrawlerCoreConfig,
+
+    /// Chain config
+    chain_config: Arc<ChainConfig>,
+
+    /// Map of all known addresses (including currently unreachable)
```
TheQuantumPhysicist, line 117: Map of all known addresses (including currently unreachable); these addresses will be periodically tested, and reachable addresses will be handed to the DNS server to be returned to the user on DNS queries, and unreachable addresses will be removed from the DNS server

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+    addresses: BTreeMap<N::Address, AddressData>,
+
+    /// Map of all currently connected outbound peers
```
TheQuantumPhysicist, line 120: Map of all currently connected outbound peers that we successfully reached and are still connected to (generally speaking, we don't have to stay connected to those peers, but this is an implementation detail)

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+    /// Addresses that were once reachable are stored in the DB.
+    was_reachable: bool,
+
+    /// Whether the address was added from the command line
+    user_added: bool,
+}
+
+#[derive(Clone)]
+pub struct CrawlerCoreConfig {
+    /// Default p2p port on which nodes normally accept inbound connections
```
TheQuantumPhysicist, line 103: Default p2p port on which nodes normally accept inbound connections; the DNS server has no way to communicate port numbers, and hence we only add addresses that are reachable through this default port

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+    },
+    Connecting {
+        /// Same as in the Disconnected state
+        fail_count: u32,
+    },
+    Connected {},
+}
+
+/// Additional state of a potential node address
+struct AddressData {
```
TheQuantumPhysicist, line 86: We discussed this in the call. Please limit the possible states further. This is just a reminder.

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AddressTransition {
+    Disconnected,
+    Connecting,
+    Connected,
+}
+
+/// Connection state of a potential node address (outbound only)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AddressState {
```
TheQuantumPhysicist, line 72: As discussed in the call, please consider making this a parameter in an AddressData method, `AddressData::transition_to()`

dns_server/src/crawler/mod.rs
```diff
@@ -14,127 +14,71 @@
...
-    addresses: BTreeMap<N::Address, AddressData>,
-
-    /// Map of all currently connected outbound peers
-    outbound_peers: BTreeMap<N::PeerId, N::Address>,
-
     /// Storage implementation
     storage: S,
 
     /// Channel used to manage the DNS server
     command_tx: mpsc::UnboundedSender<ServerCommands>,
```
TheQuantumPhysicist, line 68: Two things:

1. Please rename `ServerCommands` to `DNSServerCommand`
2. Please rename `command_tx` to `dns_server_cmd_tx`

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,589 @@
...
+pub struct CrawlerCoreConfig {
+    /// Default p2p port on which nodes normally accept inbound connections
+    pub default_p2p_port: u16,
+}
+
+/// The `CrawlerCore` is the component that communicates with Mintlayer peers using p2p,
+/// and based on the results, commands the DNS server to add/remove ip addresses.
+/// The `CrawlerCore` emits events that communicate whether addresses were reached or,
+/// are unreachable anymore.
+pub struct CrawlerCore<N: NetworkingService, C> {
```
TheQuantumPhysicist, line 111: As discussed, please rename to `CrawlerP2p`

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,589 @@
...
+    _phantom: PhantomData<C>,
+}
+
+pub enum CoreEvent<N: NetworkingService> {
+    Timer { period: Duration },
+    Connectivity { event: ConnectivityEvent<N> },
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum CoreCommand<N: NetworkingService> {
```
TheQuantumPhysicist, line 137: As discussed, please rename to `CrawlerP2pEvent`, which should emit events about addresses, without concrete decisions on what to do with addresses that are reachable.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: I started with `FnMut`, but then decided that a separate trait would be easier to use.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Removed, thanks.

dns_server/src/crawler/core/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+            "change address {} state to {:?}",
+            address.to_string(),
+            transition
+        );
+
+        address_data.state_updated_at = now;
+
+        let dns_ip = Self::get_dns_ip(address, config.default_p2p_port);
+
+        let fail_count_old = match address_data.state {
```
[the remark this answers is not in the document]
pavel-kokolemin: Done.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: Updated, thanks.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: Updated, thanks.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: Updated, thanks.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: Updated, thanks.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: Sure, updated, thanks. Actually all state is now moved to `AddressState`.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: Added to `AddressState`.

dns_server/src/crawler/mod.rs
pavel-kokolemin: Sure, updated.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: I decided that the names `Crawler` and `CrawlerManager` would be a bit better. I hope that is OK.

dns_server/src/crawler/core/mod.rs
pavel-kokolemin: > which should emit events about addresses, without concrete decisions on what to do with addresses that are reachable.

I tried that, but it did not work. Some addresses should be kept in the DB even if they are temporarily unreachable. This is needed if the server is restarted while having temporary network problems. This could be solved by having a third state (temporarily unreachable), but this would complicate the logic in `CrawlerManager`, which I prefer not to do. Also later we may want to store some additional details in the DB (for example `fail_count`).
Considering all this, I think we should keep what we have now.

TheQuantumPhysicist: In its current form it's simply wrong and unnecessarily complex. Having a mutable state in a random callback doing everything is very difficult to understand. It's not imperative not functional programming. We should strive to do things right. I suggest you try to do the right thing (whether it requires the new state or not), and then we try to improve it. This is much better than giving up this early. 

Again, what used to be called "core" doesn't have to know that there's a database or DNS server.

pavel-kokolemin: OK, I updated again. Now `Crawler` only notifies about address state changes:
```
pub trait Callback<N: NetworkingService> {
...
    fn address_updated(
        &mut self,
        address: &N::Address,
        old_state: &AddressState,
        new_state: &AddressState,
    );
}
```
And `CrawlerManager` now reacts as needed (updates DNS or DB).
Old address state is passed to keep `CrawlerManager` stateless:
```
        match (old_state.is_persistent(), new_state.is_persistent()) {
            (false, true) => {
                storage::update_db(&self.storage, |tx| tx.add_address(&address.to_string()))
                    .expect("update_db must succeed (add_address)");
            }
            (true, false) => {
                storage::update_db(&self.storage, |tx| tx.del_address(&address.to_string()))
                    .expect("update_db must succeed (del_address)");
            }
            _ => {}
        }
```

dns_server/src/crawler_p2p/crawler/address_data.rs
```diff
@@ -0,0 +1,326 @@
...
+        fail_count: u32,
+
+        /// Same as above
+        was_reachable: bool,
+
+        /// The time when the address went into the disconnected state
+        disconnected_at: Duration,
+    },
+
+    PermanentlyUnreachable {
```
TheQuantumPhysicist, line 70: If I understand correctly, perhaps you mean something like "never reached so far". `PermanentlyUnreachable` would mean that it's unreachable forever by design. I think we can use something `NeverReachedYet`.

dns_server/src/crawler_p2p/crawler/address_data.rs
```diff
@@ -0,0 +1,326 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::time::Duration;
+
+use utils::const_value::ConstValue;
+
+/// Address state transition
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AddressTransition {
```
TheQuantumPhysicist, line 22: AddressStateTransitionTo

dns_server/src/crawler_p2p/crawler/address_data.rs
```diff
@@ -0,0 +1,326 @@
...
+            AddressState::PermanentlyUnreachable {
+                erase_after: _,
+                was_reachable,
+                fail_count: _,
+            } => *was_reachable,
+        }
+    }
+
+    /// Whether the address is currently recognized as reachable (available from DNS)
+    pub fn is_reachable(&self) -> bool {
```
TheQuantumPhysicist, line 141: is_connected?

dns_server/src/crawler_p2p/crawler/address_data.rs
pavel-kokolemin: This is a final state where an address is marked as unreachable and there will be no more attempts to connect to it. After `erase_after` time it would be removed from memory (it can be added as new after that).
I will rename the state to `Unreachable` and add a comment.

dns_server/src/crawler_p2p/crawler/address_data.rs
```diff
@@ -0,0 +1,326 @@
...
+            AddressState::PermanentlyUnreachable {
+                fail_count: _,
+                was_reachable: _,
+                erase_after: _,
+            } => false,
+        }
+    }
+
+    /// Whether to retain the address between node restarts (store in DB)
+    pub fn is_persistent(&self) -> bool {
```
TheQuantumPhysicist, line 166: I don't think this is valid. The crawler has no idea about persistence.

pavel-kokolemin: It's only used in `CrawlerManager`.

dns_server/src/crawler_p2p/crawler/address_data.rs
TheQuantumPhysicist: Cool.

dns_server/src/crawler_p2p/crawler/address_data.rs
pavel-kokolemin: We can change the logic later to mark the node as reachable even if we are not currently connected. So I think `is_persistent` is a better name. This is only used in `CrawlerManager`.

dns_server/src/crawler_p2p/crawler/address_data.rs
```diff
@@ -0,0 +1,326 @@
...
+                    0 => true,
+                    1 => age > Duration::from_secs(60),
+                    2 => age > Duration::from_secs(360),
+                    _ => age > Duration::from_secs(3600),
+                }
+            }
+
+            AddressState::Disconnected {
+                fail_count,
+                was_reachable: true,
```
TheQuantumPhysicist, line 222: Please create an if condition inside that the expression below. These splits can lead to issues in the future if we add more variables as tracking them is harder than just full arms.

pavel-kokolemin: OK, updated.

dns_server/src/crawler_p2p/crawler/address_data.rs
pavel-kokolemin: Sure, updated.

pavel-kokolemin:
I replaced the callback trait with `FnMut` in the crawler. There is no more mutable callback objects and everything looks good IMO.
The crawler does not depend on `NetworkingService` and only handles relevant messages.
To simplify things a bit, I removed `PeerId` associated type from the `NetworkingService` trait. I remember that we decided it's acceptable because we don't send `PeerId` over the network.
The networking backend will now send `ConnectionClosed` event just once and only if it has sent `InboundAccepted` or `OutboundAccepted` before.

pavel-kokolemin:
I decided not to use channels for `Crawler`/`CrawlerManager` communication, because it's not really needed and would only complicate the code.

TheQuantumPhysicist (approved):
I think it's much better than before now. At least the state is trackable. I invite others to review and scruitinize the design. If anyone thinks we can do better, please don't hold back.

dns_server/Cargo.toml
```diff
@@ -27,3 +28,7 @@ parity-scale-codec.workspace = true
 void = "1.0"
 clap = { version = "4", features = ["derive"] }
 directories = "4.0"
+
+[dev-dependencies]
+rstest = "0.16"
+test-utils = { path = "../test-utils" }
```
stanislav-tkach, line 34: Nitpick: a new line between "foreign" and internal dependencies?.. 🙃

stanislav-tkach (approved):
👍

I think that "simple but excessive" changes such as making `PeerId` non-generic is better to do in a separate pull request, but I'm happy that we are finally able to do that.

dns_server/Cargo.toml
pavel-kokolemin: I will fix, thanks.

pavel-kokolemin:
> I think that "simple but massive" changes such as making `PeerId` non-generic is better to do in a separate pull request, but I'm happy that we are finally able to do that.

You are absolutely right, I will try to do it that way next time. But I needed this change in this PR and decided that separate PR would take too much time and create a lot of conflicts with my current work.

PR #711, opened 2023-02-16
Blocking subsystem calls

iljakuklic:
Introduces a wrapper over `subsystem::Handle` that blocks, so can be used without `async`. Not intended for merge yet, just filing a draft PR to gather early feedback.

### DIFF
--- a/subsystem/src/blocking.rs
+++ b/subsystem/src/blocking.rs
@@ -1,0 +1,110 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Blocking interface to subsystem.
+
+use crate::{subsystem::CallError, CallResult, Handle};
+use futures::future::BoxFuture;
+
+/// Blocking version of [subsystem::Handle].
+///
+/// This should be used sparingly as blocking calls induce non-trivial overhead. The call takes up
+/// a thread in the runtime thread pool. If there is not enough threads for all simultaneous
+/// blocking calls, a new one is spawned.
+pub struct BlockingHandle<T: ?Sized>(Handle<T>);
+
+impl<T: 'static + Send + ?Sized> BlockingHandle<T> {
+    /// A new blocking handle with a dedicated worker
+    pub fn new(handle: Handle<T>) -> Self {
+        Self(handle)
+    }
+
+    /// Get the inner handle
+    pub fn handle(&self) -> &Handle<T> {
+        &self.0
+    }
+
+    /// Perform given closure in a worker, passing the handle to it, get the result
+    fn with_handle<R: 'static + Send>(
+        &self,
+        func: impl 'static + Send + FnOnce(&Handle<T>) -> CallResult<R>,
+    ) -> Result<R, CallError> {
+        // Get the future associated with the function call result
+        let result = func(self.handle());
+
+        // Deal with it according to the current runtime context
+        match tokio::runtime::Handle::try_current() {
+            Ok(rt) => {
+                assert_eq!(
+                    rt.runtime_flavor(),
+                    tokio::runtime::RuntimeFlavor::MultiThread,
+                    "Only multi-threaded Tokio runtime supported by blocking subsystem handle"
+                );
+                tokio::task::block_in_place(|| result.blocking_get())
+            }
+            Err(err) => {
+                if err.is_missing_context() {
+                    result.blocking_get()
+                } else {
+                    panic!("Unexpected error while getting tokio runtime")
+                }
+            }
+        }
+    }
+
+    /// Blocking variant of [Handle::call_async_mut]
+    pub fn call_async_mut<R: Send + 'static>(
+        &self,
+        func: impl FnOnce(&mut T) -> BoxFuture<R> + Send + 'static,
+    ) -> Result<R, CallError> {
+        self.with_handle(|h| h.call_async_mut(func))
+    }
+
+    /// Blocking variant of [Handle::call_async]
+    pub fn call_async<R: Send + 'static>(
+        &self,
+        func: impl FnOnce(&T) -> BoxFuture<R> + Send + 'static,
+    ) -> Result<R, CallError> {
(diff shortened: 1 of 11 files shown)


subsystem/src/blocking.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Blocking variant of [Handle::call_async]
+    pub fn call_async<R: Send + 'static>(
+        &self,
+        func: impl FnOnce(&T) -> BoxFuture<R> + Send + 'static,
+    ) -> Result<R, CallError> {
+        self.with_handle(|h| h.call_async(func))
+    }
+
+    /// Blocking variant of [Handle::call_mut]
+    pub fn call_mut<R: Send + 'static>(
```
TheQuantumPhysicist, line 84: I would call these too `call_sync_mut` and `call_sync_mut` as well. I saw other libraries (like json_rpc) that use call for async and call_sync for sync... no need for making assumptions.

subsystem/src/subsystem.rs
```diff
@@ -126,6 +132,13 @@ impl<T> CallResult<T> {
     pub fn response(self) -> Result<CallResponse<T>, CallError> {
         self.0
     }
+
```
TheQuantumPhysicist, line 135: See "suceeded" typo above

TheQuantumPhysicist:
I know this isn't over and it's in draft form still, but I wanna say it's looking good. I like the idea of an isolated struct that converts calls to sync. Though I wanna emphasize that async can look good and still lead to dead-locks when called in blocking mode. Let's try to scrutinize this as much as possible to make it impossible to get that.

One hint here too is that functional tests can help as an external software for testing. It's totally possible to have everything work fine here in unit tests, but then see dead-locks in functional tests. This, actually, happened with me once in p2p.

subsystem/src/blocking.rs
```diff
@@ -0,0 +1,110 @@
...
+                if err.is_missing_context() {
+                    result.blocking_get()
+                } else {
+                    panic!("Unexpected error while getting tokio runtime")
+                }
+            }
+        }
+    }
+
+    /// Blocking variant of [Handle::call_async_mut]
```
TheQuantumPhysicist, line 67: Comment wrong

subsystem/src/blocking.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub fn call_async_mut<R: Send + 'static>(
```
TheQuantumPhysicist, line 68: Is there value to having both async and sync variants in BlockingHandle?

subsystem/src/blocking.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use crate::{subsystem::CallError, CallResult, Handle};
+use futures::future::BoxFuture;
+
+/// Blocking version of [subsystem::Handle].
+///
+/// This should be used sparingly as blocking calls induce non-trivial overhead. The call takes up
+/// a thread in the runtime thread pool. If there is not enough threads for all simultaneous
+/// blocking calls, a new one is spawned.
+pub struct BlockingHandle<T: ?Sized>(Handle<T>);
```
TheQuantumPhysicist, line 26: Maybe... `SyncHandle`?

subsystem/src/blocking.rs
iljakuklic: It's already available in a sense. You can get a reference to the inner handle and do the call through it. The advantage of only having one flavour on a type is that certain degree of consistency is enforced.

subsystem/src/blocking.rs
iljakuklic: I don't see what's wrong with it?

subsystem/src/blocking.rs
iljakuklic: I am rather torn on this one. The type already communicates it's not an async method, and it will also be obvious on call sites from the fact that there is no `.await` after it. But maybe even more explicit is better? What do others think?

subsystem/src/blocking.rs
TheQuantumPhysicist: It says blocking, and it's "async"... maybe I'm missing something.

subsystem/src/blocking.rs
alfiedotwtf: I personally always prefer explicit rather than having to guess when not totally obvious. But making the choice now doesn't necessarily lock it into stone - we could always change it back at a later stage if we all get weary of the explicit names.

subsystem/src/blocking.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl<T: ?Sized> Clone for BlockingHandle<T> {
+    fn clone(&self) -> Self {
+        Self::new(self.0.clone())
+    }
+}
+
+impl<T> ShallowClone for BlockingHandle<T> {
```
TheQuantumPhysicist, line 35: No `?Sized` here?

TheQuantumPhysicist (approved):
Looking good. I'm hoping we'll have some aggressive testing for this soon.

subsystem/src/blocking.rs
iljakuklic: The closure that's passed in the call yields a future but the call itself is still blocking (it waits for the whole thing to resolve). I.e. the `async` part refers to the inner closure, the blocking part refers to the whole call.

subsystem/src/blocking.rs
iljakuklic: Does not hurt to add

subsystem/src/blocking.rs
iljakuklic: > I personally always prefer explicit rather than having to guess when not totally obvious. But making the choice now doesn't necessarily lock it into stone - we could always change it back at a later stage if we all get weary of the explicit names.

Yea maybe. I'm just wondering whether presence/absence of `.await` makes it explicit enough already.

subsystem/src/blocking.rs
```diff
@@ -0,0 +1,125 @@
...
+                assert_eq!(
+                    rt.runtime_flavor(),
+                    tokio::runtime::RuntimeFlavor::MultiThread,
+                    "Only multi-threaded Tokio runtime supported by blocking subsystem handle"
+                );
+                tokio::task::block_in_place(|| result.blocking_get())
+            }
+            Err(err) => {
+                if err.is_missing_context() {
+                    result.blocking_get()
```
stanislav-tkach, line 74: Could you please explain this to me? In my understanding this error can only occur if the tokio runtime hasn't been started. Are there cases when we want to have a subsystem without starting the runtime?

stanislav-tkach (approved):

subsystem/src/blocking.rs
iljakuklic: My understanding is that it also happens when called from a thread that is not currently being managed by a runtime. Which is a use case worth supporting imo. Is that incorrect?

stanislav-tkach: Yes, you are right. Perhaps it is worth to add a test for such case?..

iljakuklic: Added a test for this

PR #712, opened 2023-02-16
Unify versions in Cargo.toml

### LINKED ISSUE #451: Decide how the semantic versioning of the code will be done
Rust 1.64 [release](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html) allows inheriting the version number. By mainnet, we should've decided how to do this versioning. Specifically, what crates we want to follow the version of the whole workspace, and whether we want a single version number for everything.

Video on this: https://www.youtube.com/watch?v=CO5Xo76CVCk

TheQuantumPhysicist:
Closes #451

I guess we can all agree at this point that having individual versions for crates is just an unnecessary burden. If we decide to isolate crates in the future, we can still do that.

And while at it, did the same for licenses.

The only issue is that I couldn't have a sinlge number for both packages and the workspace. I added a check in codechecks script.

### DIFF
--- a/.github/scripts/codecheck.py
+++ b/.github/scripts/codecheck.py
@@ -75,9 +75,13 @@ def check_workspace_and_package_versions_equal():
     workspace_version = root['package']['version']
     package_version = root['workspace']['package']['version']
 
-    if workspace_version != package_version:
+    result = workspace_version == package_version
+
+    if not result:
         print("Workspace vs package versions mismatch in Cargo.toml: '{}' != '{}'".format(workspace_version, package_version))
-    print()
+        print()
+
+    return result
 
 
 # Check crate versions
--- a/.github/scripts/codecheck.py
+++ b/.github/scripts/codecheck.py
@@ -67,6 +67,19 @@ def check_crate_version_unique(crate_name):
 
     return len(versions) == 1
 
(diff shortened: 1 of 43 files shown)


stanislav-tkach (approved):

muursh (approved):

.github/scripts/codecheck.py
```diff
@@ -67,6 +67,23 @@ def check_crate_version_unique(crate_name):
 
     return len(versions) == 1
 
+# Ensure that the versions in the workspace's Cargo.toml are consistent
+def check_workspace_and_package_versions_equal():
+    print("==== Ensuring workspace and package versions are equal in the workspace's Cargo.toml")
+    root = toml.load('Cargo.toml')
+
+    workspace_version = root['package']['version']
+    package_version = root['workspace']['package']['version']
```
iljakuklic, line 76: Will this throw an exception of these keys are not defined? Maybe it could be handled explicitly so it does not prevent subsequent checks being run.

iljakuklic (approved):

.github/scripts/codecheck.py
TheQuantumPhysicist: Yes, it will. I tested it yesterday. But I don't think it's worth it. This should never happen + debugging it will tell you exactly what's going on (key isn't in dict).

PR #713, opened 2023-02-17
Added a WalletTx struct and a basic integration with the db store

erasmospunk:


wallet/src/wallet.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(thiserror::Error, Debug, Eq, PartialEq)]
+pub enum WalletError {
+    #[error("Wallet database error: {0}")]
+    DatabaseError(wallet_storage::Error),
+    #[error("Transaction already present: {0}")]
+    DuplicateTransaction(Id<Transaction>),
+}
+#[allow(dead_code)] // TODO remove
+pub struct Wallet {
+    db: WalletStorageImpl,
```
TheQuantumPhysicist, line 37: A better way to do this is to use a generic, like we do in chainstate storage (see chainstate launcher). The wallet shouldn't only work with sqlite storage. There should be a generic parameter that determintes the storage backend at compile-time.

wallet/src/wallet.rs
```diff
@@ -0,0 +1,165 @@
...
+        let wallet = Wallet::open_wallet_in_memory();
+        assert!(wallet.is_ok())
+    }
+
+    #[test]
+    fn wallet_transactions() {
+        let mut wallet_path = std::env::temp_dir();
+        let current_time =
+            SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_millis();
+        wallet_path.push(format!("wallet_transactions_{current_time}.sqlite"));
```
TheQuantumPhysicist, line 137: While this might work, it is an invitation to future problems where someone copies this test and forgets to change the string, and then we have unexpected errors because of a 1/1000 chance both tests run in parallel and click at the same time. 

There are already solutions for this. Please see how `TempDir` is used for lmdb and config files. Grep the code. It's much simpler to use than this overly complex solution, in comparison.

wallet/storage/src/internal/mod.rs
```diff
@@ -111,6 +123,7 @@ impl<B: storage::Backend> WalletStorageWrite for Store<B> {
         fn set_storage_version(&mut self, version: u32) -> crate::Result<()>;
         fn set_utxo(&mut self, outpoint: &OutPoint, entry: Utxo) -> crate::Result<()>;
         fn del_utxo(&mut self, outpoint: &OutPoint) -> crate::Result<()>;
+        fn set_transaction(&mut self, id: &Id<Transaction>, tx: &WalletTx) -> crate::Result<()>;
     }
```
TheQuantumPhysicist, line 127: del_transaction?

wallet/storage/src/lib.rs
```diff
@@ -77,3 +83,17 @@ pub trait Transactional<'t> {
...
 
 pub trait WalletStorage: WalletStorageWrite + for<'tx> Transactional<'tx> + Send {}
+
+impl Store<storage_sqlite::Sqlite> {
+    /// Create a default storage (mostly for testing, may want to remove this later)
+    pub fn new_in_memory() -> crate::Result<Self> {
+        Self::new(storage_sqlite::Sqlite::new_in_memory())
+    }
+
+    pub fn new_from_path(path: PathBuf) -> crate::Result<Self> {
```
TheQuantumPhysicist, line 93: You can use the signature

```
new_from_path<P: AsRef<Path>>(path: P)
```

This basically says "take a path from anything that can become path when calling `as_ref()`"

TheQuantumPhysicist: I will fix this problem myself later. I'm see an issue in Backend that it doesn't take Impl as parameter. This will lead to a cascade of fixes.

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,120 @@
...
+        let ordering = self.order.cmp(&other.order);
+        match ordering {
+            // If the order is the same compare the tx ids
+            Ordering::Equal => self.tx.get_id().cmp(&other.tx.get_id()),
+            Ordering::Greater | Ordering::Less => ordering,
+        }
+    }
+}
+
+impl PartialOrd<Self> for WalletTx {
```
TheQuantumPhysicist, line 69: No need to manually add `<Self>`. It's the default.

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
```
TheQuantumPhysicist, line 70: Please also implement `Ord`. Ordering here has the transitive property.

We always implement Eq with PartialEq, and Ord with PartialOrd unless a reason exists.

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,120 @@
...
+        Some(self.cmp(other))
+    }
+}
+
+impl WalletTx {
+    pub fn new(tx: Transaction, state: TxState) -> Self {
+        WalletTx {
+            tx,
+            state,
+            order: -1,
```
TheQuantumPhysicist, line 80: Please don't use special values. Use `Option<u64>` to indicate that a value is uninitialized.

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,120 @@
...
+pub struct WalletTx {
+    /// The actual transaction
+    tx: Transaction,
+    /// The state of this transaction
+    state: TxState,
+    /// The order of the transaction
+    order: i64,
+}
+
+impl PartialEq for WalletTx {
```
TheQuantumPhysicist, line 52: The more I think about this, the more I realize that this is not right. Why bother and implement Ord and PartialEq for this? It's at best ambiguous. 

Why not just compare `wallet_tx.get_tx_id()` instead of specializing Eq this way?

By the same token, why not just extra the order when it's needed instead of using this as a generic ordering?

TheQuantumPhysicist (commented):
Great work. In general this is OK, but I think we're borrowing too much from bitcoin core here to the point that we're doing it wrong in rust. Let's discuss these problems and find the best solution that uses the power of rust.

wallet/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod wallet;
```
stanislav-tkach, line 16: Perhaps it makes sense to reexport the `Wallet` struct here, because otherwise using it in other crate would be somewhat awkward: `use wallet::wallet::Wallet`. Maybe `WalletError` can be exported too.

TheQuantumPhysicist: while `use wallet::wallet::Wallet` is ugly, I hate reexporting... it's heavily error prone because people tend to forget to reexport when adding new symbols. I can't count how many times I did this mistake. I like the idea of once I export, everything inside is public, and we're done.

stanislav-tkach: I'm not sure I understand the reasoning. Sure, it is possible to forget to reexport a new symbol, but it is also possible to forget to make it public. This can easily be fixed by the next person who needs to use this new symbol. I believe that reexports lead to a better and more clear interface: it is easier to see what is available outside of the crate and it prevents exposing everything.

By the way, the current situation with imports is somewhat inconsistent. For example, we have `common::chain::Block` and `common:chain::block::BlockHeader`.

Saying all of that, I still think that it isn't super important, so I'm OK with leaving things as they are.

TheQuantumPhysicist: The reasoning is simple: One export is easier than two.

I faced this issue with crates that are in different repos and it was a pain. So I just hate that habit.

wallet/storage/src/internal/mod.rs
```diff
@@ -48,6 +52,13 @@ impl<B: storage::Backend> Store<B> {
         let db_tx = self.transaction_ro()?;
         db_tx.0.get::<db::DBUtxo, _>().prefix_iter_decoded(&()).map(Iterator::collect)
     }
+
+    /// Collect and return all transactions from the storage
+    #[allow(clippy::let_and_return)]
```
azarovh, line 57: I believe this `allow` is unnecessary

wallet/src/wallet.rs
erasmospunk: Yeah, it's better with the `TempDir`

wallet/storage/src/lib.rs
erasmospunk: I implemented it as you proposed. There is still a `PathBuf` inside the `SqliteStorageMode` struct as a simple `Path` is unsized.

wallet/types/src/wallet_tx.rs
erasmospunk: I originally implemented it with `Optional<u64>` but I didn't like how the code looked like. I gave it another shot and it looks better than before.

wallet/types/src/wallet_tx.rs
erasmospunk: I needed the `PartialEq` for `WalletTx` to be equal even if the order changed. For now I reverted the `PartialEq` as it was only needed for tests.

The `Ord` is needed so the transactions can be properly sorted.

common/src/primitives/id/with_id.rs
```diff
@@ -78,12 +79,19 @@ impl<T: Idable> From<T> for WithId<T> {
     }
 }
 
+impl<T: Idable> WrapperTypeDecode for WithId<T> {
```
TheQuantumPhysicist, line 82: Why would we need to serialize and deserialize WithId? We shouldn't be storing serialized objects with their hashes.

This is actually a recipe for bugs. A scenario for this is as follows: We serialize something, break the invariant and modify the hash, then deserialize it. 

Why not deserialize objects then put them in WithId?

wallet/src/wallet.rs
```diff
@@ -0,0 +1,194 @@
...
+    #[test]
+    fn in_memory_wallet() {
+        let wallet = open_wallet_in_memory();
+        assert!(wallet.is_ok())
+    }
+
+    #[test]
+    fn wallet_transactions() {
+        let wallet_path =
+            tempfile::TempDir::new().unwrap().path().join("wallet_transactions.sqlite");
```
TheQuantumPhysicist, line 155: This is wrong. This is the mistake I warned from.

The object of TempDir must be held until the end of the test, because the Drop trait, or the destructor, of TempDir deletes the file. Which means, you're creating the directory and destroying it immediately. If you check your /tmp/ directory, you'll probably find all these remnants sitting there undeleted.

wallet/types/src/wallet_tx.rs
TheQuantumPhysicist: I need a little more insight from you to understand the problem. When you say "properly sorted". What do you mean? Sorting them where?

If you're sorting them in a Vec, you can use `sort_by` and provide a closure (functor, in C++ lingo) to tell it how to sort.

If you're ordering them for a BTreeMap, then use the key for the ordering.

If you're ordering them for a BTreeSet, then create a wrapper type (equivalent to functors in C++, but in a different way) that implements that sorting with deref and as_ref impls, depending on how it should be used.

Even in C++, we don't implement `WalletTx::operator<(int order1, int order2)` just for convenience. We don't do it, because it's confusing, because order means something. Types should have a meaning as well, and hence we emphasize making operations on types meaningful.

common/src/primitives/id/with_id.rs
erasmospunk: The `WrapperTypeEncode` does not encode the hash. For example in the `serialization()` test in `with_id.rs` encodes this struct

```
TestStruct {
    num: 1337,
    blurb: "Hello!".into(),
}
```

as `[39, 05, 00, 00, 00, 00, 00, 00, 18, 48, 65, 6C, 6C, 6F, 21]`.

Here is the relevant source of that trait:

```
/// A marker trait for types that wrap other encodable type.
///
/// Such types should not carry any additional information
/// that would require to be encoded, because the encoding
/// is assumed to be the same as the wrapped type.
///
/// The wrapped type that is referred to is the [`Deref::Target`].
pub trait WrapperTypeEncode: Deref {}
```

wallet/types/src/wallet_tx.rs
erasmospunk: I need to keep the transactions sorted in memory in a `BTreeMap` and the order must follow `order: Option<u64>` using a custom logic. This is to respond quickly to the RPC requests to the transaction history.

common/src/primitives/id/with_id.rs
TheQuantumPhysicist: I see. That's fair. Then please add it to the `with_id.rs` file instead of here. Then we can scrutinize this as a generic solution instead of just for the wallet.

wallet/types/src/wallet_tx.rs
TheQuantumPhysicist: I don't understand. Why not make the key of the map the order number?

erasmospunk: Because I need to query by transaction hash.

common/src/primitives/id/with_id.rs
erasmospunk: Added some comments. As it is now, anywhere `WithId<T>` is used it can be serialised as if `T` was serialised.

TheQuantumPhysicist: Can you please move this to `with_id.rs` file and import it from there?

wallet/types/src/wallet_tx.rs
TheQuantumPhysicist: I understand you're trying to solve such a problem.

Please see how the mempool solves this problem.

https://github.com/mintlayer/mintlayer-core/blob/master/mempool/src/pool/store.rs

There's no reason to implement an arbitrary order just for convenience. There's always another correct way. Let's chat about it.

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl Ord for WalletTx {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let ordering = self.order.cmp(&other.order);
+        match ordering {
+            // If the order is the same compare the tx ids
+            Ordering::Equal => self.tx.get_id().cmp(&other.tx.get_id()),
+            Ordering::Greater | Ordering::Less => ordering,
+        }
```
iljakuklic, line 60: I understand it is being discussed whether the `Ord` instance is needed at all.

However, from the coding perspective, if you want to compare by one thing and if it's equal check another thing, there is a method for it in the `Ordering` struct:
```rust
    self.order.cmp(&other.order).then_with(|| self.tx.get_id().cmp(other.tx.get_id()))
```

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
+pub struct WalletTx {
+    /// The actual transaction
+    tx: WithId<Transaction>,
+    /// The state of this transaction
+    state: TxState,
+    /// The order of the transaction
+    order: Option<u64>,
```
iljakuklic, line 50: When is this `None`?

iljakuklic (commented):
Adding some initial comments. I understand this is still in flux which may make some of them a bit premature. Feel free to ignore these for the time of being and check back later.

common/src/primitives/id/with_id.rs
erasmospunk: I am not sure what do you mean by moving this to `with_id.rs`. The impl is already in `with_id.rs` but there is nothing to import.

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,153 @@
...
+    InMempool,
+    /// Conflicted transaction with a confirmed block
+    #[codec(index = 2)]
+    Conflicted(Id<GenBlock>),
+    /// Transaction that is not confirmed or conflicted and is not in the mempool.
+    #[codec(index = 3)]
+    Inactive,
+    /// Unrecognized state
+    #[codec(index = 4)]
+    Unrecognized,
```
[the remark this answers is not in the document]
erasmospunk: I removed it for now and re-add it if necessary.

wallet/types/src/wallet_tx.rs
erasmospunk: Nice trick, thank you.

wallet/types/src/wallet_tx.rs
erasmospunk: When the WalletTx does not have a certain order. I removed this code for the time being.

wallet/types/src/wallet_tx.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Transaction that is not confirmed or conflicted and is not in the mempool.
+    #[codec(index = 3)]
+    Inactive,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
+pub struct WalletTx {
+    /// The actual transaction
+    tx: WithId<Transaction>,
+    /// The state of this transaction
```
TheQuantumPhysicist, line 42: Please add more comments on what state really means, like why we need this.

stanislav-tkach (approved):

iljakuklic (approved):

PR #715, opened 2023-02-17
Add Block Builder tests that touch the mempool

alfiedotwtf:


blockprod/src/detail/builder.rs
```diff
@@ -421,4 +455,61 @@ mod tests {
...
+
+            let _block_id = block.get_id();
+
+            chainstate.call_mut(move |this| {
+                this.process_block(block, BlockSource::Local).expect("Error processing block");
+            });
+
+            let _recv = timeout(Duration::from_millis(1000), mempool_rx.recv());
+
+            // TODO: the following assertion will fail until we turn on mempool broadcasts
```
TheQuantumPhysicist, line 499: While I'm OK with this PR, one thing to mention here is that we use `Box<dyn Trait>` for subsystems because we want to make it possible to mock the interfaces of these elements. So a solution could be that you implement a mock interface for the mempool (and chainstate too, if you wish), and then make it do whatever you want :wink:

TheQuantumPhysicist (approved):

blockprod/src/detail/builder.rs
alfiedotwtf: Gotcha, that's useful. Ok cool, I'll update with a mock.

azarovh (approved):

stanislav-tkach (approved):

muursh (approved):

blockprod/src/detail/builder.rs
alfiedotwtf: I've had a look at replacing what's currently there with mocked objects, but I don't think they'll help here because what we need to trigger the right messages is not part of the public interface - so in order to get them to fire, we'll have to eventually call out to the real `process_block()` anyway (which is what's already there)...

So I'll merge this as is, but in future I'll mock either by hand or use `mock!` when feasible.

PR #717, opened 2023-02-21
Replace Uint256 with crypto_bigint::U256

azarovh:


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -889,8 +889,10 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
             GenBlockIndex::Genesis(_) => ConsensusExtraData::None,
         };
 
-        let chain_trust =
-            *prev_block_index.chain_trust() + self.get_block_proof(block).log_err()?;
+        let chain_trust = prev_block_index
+            .chain_trust()
+            .checked_add(&self.get_block_proof(block).log_err()?)
+            .unwrap();
```
TheQuantumPhysicist, line 895: I don't think clippy will let you do that

chainstate/test-suite/src/tests/processing_tests.rs
```diff
@@ -693,8 +695,12 @@ fn consensus_type(#[case] seed: Seed) {
     let ignore_again = BlockHeight::new(10);
     let pow_again = BlockHeight::new(15);
 
-    let min_difficulty =
-        Uint256([0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF]);
+    let min_difficulty = U256::from_words([
```
TheQuantumPhysicist, line 698: MAX?

common/src/uint/mod.rs
```diff
@@ -41,3 +40,101 @@ pub trait BitArray {
     /// Create value representing one
     fn one() -> Self;
 }
+
+#[derive(Debug, Clone)]
+pub struct U256Encodable(pub U256);
```
TheQuantumPhysicist, line 45: Instead of this, maybe we should just use H256 and convert back and forth.

And btw, it's always preferable to use From+Into+Deref+AsRef instead of making the inner a `pub`.

common/src/uint/mod.rs
```diff
@@ -41,3 +40,101 @@ pub trait BitArray {
...
+}
+
+impl Decode for U256Encodable {
+    fn decode<I: serialization::Input>(input: &mut I) -> Result<Self, serialization::Error> {
+        let v = <[u8; 32]>::decode(input)?;
+        Ok(U256Encodable(U256::from_le_bytes(v)))
+    }
+}
+
+pub fn into_u512(u: U256) -> U512 {
```
TheQuantumPhysicist, line 68: I really thought the library provided this conversion... are you sure you have to do this like that? I have a feeling there should be a conversion for this already with `into()` or `try_into()`

TheQuantumPhysicist: Also, if we're gonna do this, there's not even close to enough tests for it.

chainstate/types/src/block_index.rs
```diff
@@ -48,7 +50,7 @@ impl BlockIndex {
             block_header: block.header().clone(),
             block_id: block.get_id(),
             some_ancestor,
-            chain_trust,
+            chain_trust: U256Encodable(chain_trust),
```
TheQuantumPhysicist, line 53: Like I said elsewhere... H256 is maybe better than making U256 encodable.

common/src/uint/mod.rs
azarovh: surprisingly the is no, expect from tuples

common/src/chain/pow.rs
```diff
@@ -91,21 +91,21 @@ const fn allow_min_difficulty_blocks(chain_type: ChainType) -> bool {
     }
 }
 
-pub(crate) const fn limit(chain_type: ChainType) -> Uint256 {
+pub(crate) const fn limit(chain_type: ChainType) -> U256 {
     match chain_type {
-        ChainType::Mainnet | ChainType::Testnet => Uint256([
+        ChainType::Mainnet | ChainType::Testnet => U256::from_words([
```
TheQuantumPhysicist, line 96: Using "from words" is wrong. This basically means mintlayer-core will never compile on 32-bit systems, because the size of a word in 32-bit systems is 32-bit.

Please use the hex versions so that it doesn't depend on word size.

common/src/primitives/compact.rs
```diff
@@ -51,15 +52,15 @@ impl TryFrom<Compact> for Uint256 {
...
-impl From<Uint256> for Compact {
-    fn from(value: Uint256) -> Self {
-        let mut size = (value.bits() + 7) / 8;
+impl From<U256> for Compact {
+    fn from(value: U256) -> Self {
+        let mut size = (value.bits_vartime() + 7) / 8;
 
         let mut compact = if size <= 3 {
-            value.low_u64() << (8 * (3 - size))
+            value.to_words()[0] << (8 * (3 - size)) // FIXME: a problem for 32bit OS???
```
TheQuantumPhysicist, line 60: At this point I'm getting more and more convinced that this is becoming a mess very quickly. Perhaps we should just establish our own `U256` type as a wrapper for `crypto_bigint::U256`, where inside it, we limit all these operations to what's necessary and to what works everywhere, so these mistakes can never be made.

PR #719, opened 2023-02-21
Add Uint512 and use in PoS

azarovh:
I didn't duplicate all the tests from `Uint256` and checked only corner cases specific to `Uint512`. I think it should be enough because the same code is already tested for `Uint128` and `Uint256`

### DIFF
--- a/common/src/uint/impls.rs
+++ b/common/src/uint/impls.rs
@@ -622,6 +622,7 @@ mod tests {
 
     use super::*;
     use crate::uint::BitArray;
+    use proptest::prelude::*;
     use test_utils::random::Seed;
 
     #[rstest]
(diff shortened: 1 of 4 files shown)


stanislav-tkach (approved):

common/src/uint/impls.rs
```diff
@@ -1027,4 +1076,13 @@ mod tests {
         test_op_equality(10000000000000000000000000);
         test_op_equality(100000000000000000000000000);
     }
+
+    #[test]
+    pub fn uint512_from_uint256() {
```
TheQuantumPhysicist, line 1081: Please add more aggressive testing. See how I tested 128 vs 256.

TheQuantumPhysicist (approved):

iljakuklic (approved):

common/src/uint/impls.rs
azarovh: I rewrote this test and an old one using proptest. Please take a look

muursh (approved):

TheQuantumPhysicist (approved):

PR #722, opened 2023-02-23
Minor p2p updates

pavel-kokolemin:
Just some minor improvements in p2p types (removing some duplicate and unused types).

### DIFF
--- a/dns_server/Cargo.toml
+++ b/dns_server/Cargo.toml
@@ -8,7 +8,6 @@ rust-version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-# Local dependencies
 crypto = { path = "../crypto" }
 p2p = { path = "../p2p" }
 common = { path = "../common" }
@@ -18,7 +17,6 @@ serialization = { path = "../serialization" }
 logging = { path = "../logging" }
 utils = { path = '../utils' }
 
-# External dependencies
 tokio = { workspace = true, default-features = false }
 trust-dns-client = "0.22.0"
 trust-dns-server = "0.22.0"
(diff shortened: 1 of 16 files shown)


dns_server/Cargo.toml
```diff
@@ -30,5 +30,8 @@ clap = { version = "4", features = ["derive"] }
 directories = "4.0"
 
 [dev-dependencies]
-rstest = "0.16"
+# Local dependencies
```
TheQuantumPhysicist, line 33: You really don't need to write these comments... stating the obvious isn't usually a good practice in comments.

p2p/src/interface/types.rs
```diff
@@ -13,12 +13,14 @@
...
+
+use crate::types::peer_id::PeerId;
+
+#[derive(Debug, Serialize)]
 pub struct ConnectedPeer {
-    // TODO: Replace String with actual type, once libp2p removed
-    pub peer_id: String,
+    pub peer_id: PeerId,
 
-    // TODO: Replace String with actual type, once libp2p removed
```
stanislav-tkach, line 21: Are there chances that we can replace this string with the actual type too?

stanislav-tkach (approved):

dns_server/Cargo.toml
pavel-kokolemin: OK, sure, removed.

p2p/src/interface/types.rs
pavel-kokolemin: The problem is that we don't know the actual type.
For example here:
```
impl<T: ToString> From<&PeerContext<T>> for ConnectedPeer {
    fn from(context: &PeerContext<T>) -> Self {
        ConnectedPeer {
            peer_id: context.info.peer_id,
            address: context.address.to_string(),
            inbound: context.role == Role::Inbound,
            ban_score: context.score,
        }
    }
}
```

I suppose we could try to replace it with the `PeerAddress` type, which can be constructed from a generic network address (`PeerContext` would need to implement `serde::Serialize` then).

TheQuantumPhysicist (approved):

PR #724, opened 2023-02-24
Add Block Maker unit tests

alfiedotwtf:


mempool/src/interface/mock.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::MempoolEvent;
+use common::chain::{SignedTransaction, Transaction};
+use common::primitives::Id;
+use std::sync::{Arc, Mutex};
+use subsystem::{CallRequest, ShutdownRequest};
+
+use super::mempool_interface::{MempoolInterface, MempoolSubsystemInterface};
+
+#[derive(Clone)]
+pub struct MempoolInterfaceMock {
```
azarovh, line 29: could it be done using `mockall` instead?

alfiedotwtf: I tried. Although `mockall` says that it supports `async`, no matter what I tried, I couldn't get it to work with our Subsystem

mempool/src/interface/mock.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub get_all_called: Arc<Mutex<bool>>,
+    pub get_all_should_error: Arc<Mutex<bool>>,
+    pub contains_transaction_called: Arc<Mutex<bool>>,
+    pub contains_transaction_should_error: Arc<Mutex<bool>>,
+    pub collect_txs_called: Arc<Mutex<bool>>,
+    pub collect_txs_should_error: Arc<Mutex<bool>>,
+    pub subscribe_to_events_called: Arc<Mutex<bool>>,
+    pub subscribe_to_events_should_error: Arc<Mutex<bool>>,
+    pub run_called: Arc<Mutex<bool>>,
+    pub run_should_error: Arc<Mutex<bool>>,
```
iljakuklic, line 41: Not that it is a big deal but maybe this could be improved:

* Use one `Arc` to guard all the fields (unless you want to keep references to the individual fields independently of life of the struct. It involves adding an intermediate struct though
* Use `AtomicBool` in place of a `Mutex`. I think it also removes the need for some `unwrap`s down below.

mempool/src/interface/mock.rs
```diff
@@ -0,0 +1,143 @@
...
+use common::chain::{SignedTransaction, Transaction};
+use common::primitives::Id;
+use std::sync::{Arc, Mutex};
+use subsystem::{CallRequest, ShutdownRequest};
+
+use super::mempool_interface::{MempoolInterface, MempoolSubsystemInterface};
+
+#[derive(Clone)]
+pub struct MempoolInterfaceMock {
+    pub add_transaction_called: Arc<Mutex<bool>>,
```
stanislav-tkach, line 30: I'm not sure if it is really needed, but if in the future we want to have more complex tests, then having counters instead of boolean flags will be more flexible.

stanislav-tkach (approved):

mempool/src/interface/mock.rs
```diff
@@ -0,0 +1,143 @@
...
+        mut self,
+        mut call_rq: CallRequest<dyn MempoolInterface>,
+        _shut_rq: ShutdownRequest,
+    ) {
+        *self.run_called.lock().unwrap() = true;
+
+        if !*self.run_should_error.lock().unwrap() {
+            tokio::select! {
+            call = call_rq.recv() => call(&mut self).await
+                }
```
iljakuklic, line 140: Strange formatting

mempool/src/interface/mock.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl MempoolSubsystemInterface for MempoolInterfaceMock {
+    async fn run(
+        mut self,
+        mut call_rq: CallRequest<dyn MempoolInterface>,
+        _shut_rq: ShutdownRequest,
```
iljakuklic, line 133: Why is shutdown request not handled. Is that intentional?

blockprod/src/detail/block_maker.rs
```diff
@@ -185,3 +185,460 @@ impl BlockMaker {
...
+        let (mut manager, chain_config, chainstate, _mempool) = setup_blockprod_test();
+
+        let mock_mempool = MempoolInterfaceMock::new();
+
+        let mock_mempool_subsystem = manager.add_subsystem_with_custom_eventloop("mock-mempool", {
+            let mock_mempool = mock_mempool.clone();
+            move |call, shutdn| async move {
+                mock_mempool.run(call, shutdn).await;
+            }
+        });
```
iljakuklic, line 225: I don't think you need a custom eventloop unless you're testing some special corner cases, for example mempool subsystem dying prematurely due to a panic.

blockprod/src/detail/block_maker.rs
```diff
@@ -185,3 +185,460 @@ impl BlockMaker {
...
+            let shutdown = manager.make_shutdown_trigger();
+            move |_| shutdown.initiate()
+        });
+
+        // shutdown straight after startup, *then* call collect_transactions()
+        manager.main().await;
+
+        let (_tx_builder, rx_builder) = crossbeam_channel::unbounded();
+
+        tokio::spawn(async move {
```
iljakuklic, line 237: You can make the tester task into a subsystem with a custom event loop (which does not actually contain a loop). When it exits, the subsystem manager will be automatically notified and will tear down the remaining subsystems in a clean way, no custom shutdown handling required.

iljakuklic (commented):
Added a couple of initial comments

mempool/src/interface/mock.rs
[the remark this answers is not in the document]
alfiedotwtf: Doh! I've noticed `rustfmt` has been playing up for me lately with Emacs 🤦‍♂️

mempool/src/interface/mock.rs
alfiedotwtf: It wasn't needed for this change so I just went with `bool`, but yes if we need it sometime in the future, it's an easy change.

mempool/src/interface/mock.rs
alfiedotwtf: Oh, `AtomicBool`... of course!

There wasn't any immediate need for the separation, it way just more convenient to copy-paste. I'll change to `AtomicBool` and see how it looks with a single containing `Arc`.

mempool/src/interface/mock.rs
alfiedotwtf: Yeah, `rustic-mode` and `rust-analyzer` seem to struggle with macros, but `rustfmt` from the command line works no issue.

mempool/src/interface/mock.rs
alfiedotwtf: Not intentional, I just didn't need it at the time and missed it during cleanup. I'll fix this, thanks.

blockprod/src/detail/block_maker.rs
alfiedotwtf: Of course! Ok cool, I'll change this which should make things simpler.

blockprod/src/detail/block_maker.rs
alfiedotwtf: Ah ok. I was just following the way `node` adds the `mempool` subsystem. Ok cool, I'll fix these.

chainstate/src/interface/mod.rs
```diff
@@ -17,5 +17,4 @@ pub mod chainstate_interface;
 pub mod chainstate_interface_impl;
 pub mod chainstate_interface_impl_delegation;
 
-#[cfg(any(test, feature = "mock"))]
 pub mod mock;
```
TheQuantumPhysicist, line 20: This is an issue that I honestly don't know how to solve. We can't use test-guarded things outside of its crate. Part of me thinks we should do the mock implementations in a separate crate then only import it where its needed as part of `dev-dependencies`. What do you guys think about this?

We can have like a separate crate, call it `mock-impls` (?), where all these impls go. This way it doesn't build with the software every time, except for tests.

blockprod/src/detail/block_maker.rs
```diff
@@ -185,3 +185,460 @@ impl BlockMaker {
...
+                accumulator.is_ok(),
+                "Expected collect_transactions() to succeed"
+            );
+        });
+
+        manager.main().await;
+    }
+
+    //
+    // Skipping unit tests as we're about rework make_block()
```
TheQuantumPhysicist, line 353: Is this comment relevant to this test? I don't see `ignore` flag anywhere

TheQuantumPhysicist (approved):
Looking good. No major points from my end. The mock crate thing should be solved sooner or later.

chainstate/src/interface/mod.rs
alfiedotwtf: Nice!

Yeah I was thinking `common` was a good place for them because it's the place the share things, but a separate crate only for `dev-dependencies` is better. Cool, I'll move all of them out as part of this change.

blockprod/src/detail/block_maker.rs
alfiedotwtf: Sorry, what I meant by skipping was as in me skipping implementing them.

chainstate/src/interface/mod.rs
TheQuantumPhysicist: Oh, please not common. We try to take things out of it. Only put it in common if there's no other way (if circular dependencies occur)

alfiedotwtf: Hehe yeah I figured there was no difference between `common` and exposing it where it was. But I do like the `mock/mock-impls` crate idea

mempool/src/interface/mock.rs
alfiedotwtf: I think for a mock object, and the way how I'm using it, a single `Arc` per field is fine for now.

But yes, `AtomicBool` was much nicer.

mempool/src/interface/mock.rs
alfiedotwtf: This was fixed manually, and hopefully wasn't wiped out with my last `cargo fmt`.

blockprod/src/detail/block_maker.rs
alfiedotwtf: For the tests where `MempoolInterfaceMock` were used, the call to the mocked `run()` was needed. Otherwise, the real `run()` was called instead.

Also, `mempool` seems to be a special case with `add_subsystem()` as it doesn't implement the `Subsystem` interface like the others i.e `shutdown()` doesn't exist, and so `add_subsystem_with_custom_eventloop()` was the only way I could see it added. If I'm wrong about this, let me know how and I'll be happy to fix.

blockprod/src/detail/block_maker.rs
alfiedotwtf: Apart from the tests that have to run after `master.main()` is called, all of the `spawn()` calls have now been converted.

chainstate/src/interface/mod.rs
alfiedotwtf: So I've only moved `mempool` and `chainstate`, and they should both only be built as dev dependencies . Let me know if there is any issues with this change.

TheQuantumPhysicist: Yep. Looking good!

blockprod/src/detail/block_maker.rs
iljakuklic: Is there anything preventing implementing `Subsystem` for `MempoolInterfaceMock`?

blockprod/src/detail/block_maker.rs
```diff
@@ -185,3 +185,466 @@ impl BlockMaker {
...
+            let mock_mempool = mock_mempool.clone();
+            move |call: CallRequest<dyn MempoolInterface>, shutdn| async move {
+                mock_mempool.run(call, shutdn).await;
+            }
+        });
+
+        mock_mempool_subsystem.call_mut({
+            let shutdown = manager.make_shutdown_trigger();
+            move |_| shutdown.initiate()
+        });
```
iljakuklic, line 233: Do you need `call_mut` here, as opposed to plain `call`? It does not touch the mempool object

blockprod/src/detail/block_maker.rs
```diff
@@ -185,3 +185,466 @@ impl BlockMaker {
...
+    //
+
+    #[tokio::test]
+    async fn attempt_submit_new_block_subsystem_error() {
+        let (manager, chain_config, chainstate, mempool) = setup_blockprod_test();
+
+        chainstate.call_mut({
+            let shutdown = manager.make_shutdown_trigger();
+            move |_| shutdown.initiate()
+        });
```
iljakuklic, line 376: `call_mut` here as well

iljakuklic (commented):
More feedback below:

blockprod/src/detail/block_maker.rs
alfiedotwtf: I had a lot of issues when I tried this, but the main issue was that `manager.add_subsystem("mock", mock_mempool)` expects its second parameter to be:

    Box<dyn MempoolInterface>

This then returns:

    subsystem::Handle<Box<dyn MempoolInterface>>

However, now there is a type difference with the current `MempoolHandle`, as it's defined as:

    subsystem::Handle<dyn MempoolInterface>

Because of that type difference, I can't use the handle that `add_subsystem()` returns. Maybe I'm missing something simple, but I couldn't find an easy way to convert between types, and so creating a custom event look just like how it's done in in the Node runner was the simpler solution.

mocks/src/mempool.rs
```diff
@@ -0,0 +1,145 @@
...
+        mut shut_rq: ShutdownRequest,
+    ) {
+        self.run_called.store(true, Relaxed);
+
+        if !self.run_should_error.load(Relaxed) {
+            tokio::select! {
+                call = call_rq.recv() => call(&mut self).await,
+                () = shut_rq.recv() => return,
+            }
+        }
```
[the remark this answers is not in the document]
alfiedotwtf: This time, yes that was the intended behaviour - these "_should_error" fields are just `bool` for now as that's what I needed for my blockprod tests.

If we need more complex behaviour in the future, then we can move to mimicking `mockall` a bit more, but I think it's sufficient for now.

blockprod/src/detail/block_maker.rs
alfiedotwtf: Ah cool, I've updated everywhere it's a `call_mut()` when it didn't need to be

blockprod/src/detail/block_maker.rs
iljakuklic: Hmm okay it seems the best way of fixing this would be at subsystem or mempool level. Let's leave it like it is, at least for now.

iljakuklic (approved):

PR #725, opened 2023-02-24
`PeerManager` updates

pavel-kokolemin:
I made some updates to `PeerManager` and `PeerDb` to implement some missing basic parts.
`PeerManager` will now load IPs from DNS seeds on first start and gracefully reconnect to peers.
Added two separate inbound and outbound peer limits (inbound limit is configurable).
I also removed error reporting from `PeerManager` functions where possible to simplify things.
This PR depends on https://github.com/mintlayer/mintlayer-core/pull/727.


.github/scripts/codecheck.py
```diff
@@ -140,9 +140,9 @@ def check_todos():
 
         with open(path) as file:
             file_data = file.read()
-            if 'TODO(PR)' in file_data or 'FIXME' in file_data:
+            if 'TODO(PR)' in file_data or 'FIXME' in file_data or 'todo!()' in file_data:
```
TheQuantumPhysicist, line 143: Thank you for doing this... you beat me to it.

p2p/src/peer_manager/peerdb/address_data/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+use utils::const_value::ConstValue;
+
+/// When the node drops the unreachable node address. Used for negative caching.
+pub const PURGE_UNREACHABLE_TIME: Duration = Duration::from_secs(3600);
+
+/// When the server drops the unreachable node address that was once reachable. This should take about a month.
+/// Such a long time is useful if the node itself has prolonged connectivity problems.
+pub const PURGE_REACHABLE_FAIL_COUNT: u32 = 35;
+
+pub enum AddressState {
```
TheQuantumPhysicist, line 27: Are we copy-pasting this from the DNS server?

TheQuantumPhysicist: I'm talking about the whole file, btw.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -16,197 +16,237 @@
...
+    banned_addresses: BTreeMap<B, Duration>,
 
     time_getter: TimeGetter,
 
     storage: S,
 }
 
-impl<T: NetworkingService, S: PeerDbStorage> PeerDb<T, S> {
+impl<
+        A: Ord + FromStr + ToString + Clone + AsBannableAddress<BannableAddress = B>,
```
TheQuantumPhysicist, line 72: When trait-bounds get so long, using `where` is recommended.

p2p/src/peer_manager/peerdb/address_data/mod.rs
pavel-kokolemin: The approach is similar, but not the same. Most importantly, addresses in `PeerDb` do not have `Connecting` and `Disconnecting` states. This is because `PeerManager` already has information about pending connection states, and so duplicating it in `PeerDb` is not very good.

p2p/src/peer_manager/peer_context.rs
```diff
@@ -45,6 +45,9 @@ pub struct PeerContext<A> {
     /// Sent ping details
     pub sent_ping: Option<SentPing>,
 
+    /// Whether an address list request was sent and no response was received.
```
TheQuantumPhysicist, line 48: Please explain with more details why this is necessary and how it helps in the lifetime of the application.

p2p/src/peer_manager/mod.rs
```diff
@@ -65,8 +60,9 @@ use self::{
     peerdb::storage::PeerDbStorage,
 };
 
-/// Maximum number of connections the [`PeerManager`] is allowed to have open
-const MAX_ACTIVE_CONNECTIONS: usize = 128;
+/// Maximum number of outbound connections the [`PeerManager`] is allowed to have open
+/// This value is constant because users should not change this.
+const MAX_OUTBOUND_CONNECTIONS: usize = 8;
```
TheQuantumPhysicist, line 65: Two things:

1. I think bitcoin does 16.
2. I hope this doesn't prevent users from choosing to connect to someone if the max is hit. A user should be able to choose to connect to peers through RPC.

p2p/src/peer_manager/mod.rs
```diff
@@ -495,25 +332,273 @@ where
...
             Err(e) => {
+                log::debug!("disconnecting new peer {peer_id} failed: {e}");
                 if let Some(response) = response {
                     response.send(Err(e));
                 }
             }
         }
+    }
+
+    /// Check if the (inbound or outbound) peer connection can be accepted
```
TheQuantumPhysicist, line 353: Please give context to this, like an example on when it's not accepted.

TheQuantumPhysicist:
If this PR depends on #727, then please base it on #727... otherwise we see both changes in one PR.

TheQuantumPhysicist:
The PR looks fine to me, except that it depends on the disputed #727. 

I also would like to ask @stanislav-tkach also to review it.

TheQuantumPhysicist (approved):
The PR looks fine to me, except that it depends on the disputed https://github.com/mintlayer/mintlayer-core/pull/727.

I also would like to ask @stanislav-tkach also to review it.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: > I think bitcoin does 16. Right?

No, bitcoin does 8. From https://github.com/bitcoin/bitcoin/blob/519ec2650e7a4dd2850ff74cba403cf446a44a5a/src/net.h#L68:

```
/** Maximum number of automatic outgoing nodes over which we'll relay everything (blocks, tx, addrs, etc) */
static const int MAX_OUTBOUND_FULL_RELAY_CONNECTIONS = 8;
```

> I hope this doesn't prevent users from choosing to connect to someone if the max is hit. A user should be able to choose to connect to peers through RPC.

`MAX_OUTBOUND_CONNECTIONS` is not checked from RPC connect command, so users should be able always connect with RPC.
BTW, should RPC connect work the same way like the `addnode` flag? If yes, then we would need to add/update the address in `PeerDb` with the `user_added` flag set. Right now the node will try to connect just once after the RPC request. I guess we could check how bitcoin works.
And I just noticed that the node won't enforce connections to added addresses (I copied the code from the DNS server where it's not a problem because there is no limit on outbound connections). I will add TODO for this issue.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: Updated, thanks for the hint.

p2p/src/peer_manager/peerdb/storage.rs
```diff
@@ -16,19 +16,27 @@
 use std::time::Duration;
 
 pub trait PeerDbStorageRead {
-    fn get_known_addresses(&self) -> crate::Result<Vec<String>>;
+    fn get_version(&self) -> Result<Option<u32>, storage::Error>;
 
-    fn get_banned_addresses(&self) -> crate::Result<Vec<(String, Duration)>>;
+    fn get_known_addresses(&self) -> Result<Vec<String>, storage::Error>;
```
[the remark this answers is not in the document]
pavel-kokolemin: Probably not, because we only store reachable nodes there. For example, the bitcoin network has less than 20k reachable nodes.
I added TODO.

p2p/src/peer_manager/peer_context.rs
pavel-kokolemin: Sure, updated.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: Sure, done.

pavel-kokolemin:
> The PR looks fine to me, except that it depends on the disputed #727.

I switched to using `tokio::time::Instant` (like we do in other p2p parts) and rebased the PR to the master branch for now.
We can decide what to do with `tokio::time::Instant` later.

p2p/src/peer_manager/mod.rs
TheQuantumPhysicist: Connecting from RPC will need a little more work, because we want to make it possible to maintain connections with a specific peer. So it's a superset of what `addnode` does in bitcoin. Substrate/Parity has 3 types of peers for this.

There's an issue about this.

stanislav-tkach (approved):

p2p/src/peer_manager/mod.rs
pavel-kokolemin: OK, sure, I will add this. But let's please do this in the next PR.

TheQuantumPhysicist: Sure.

PR #726, opened 2023-02-24
Initial implementation of the KeyChain

erasmospunk:


crypto/Cargo.toml
```diff
@@ -12,7 +12,7 @@ authors = ["Samer Afach <samer.afach@mintlayer.org>", "Ben Marsh <benjamin.marsh
 serialization = { path = "../serialization" }
 secp256k1 = { version = "0.24.2", default-features = false, features = ["global-context", "rand-std"] }
 hmac = "0.12.1"
-bip39 = { version = "1.0.1", default-features = false }
+bip39 = { version = "1.0", default-features = false }
```
TheQuantumPhysicist, line 15: Maybe do the same for `hmac` :smile:

Btw, I see there's a new version for secp256k1 crate.

wallet/types/Cargo.toml
```diff
@@ -8,8 +8,13 @@ rust-version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-serialization = { path = "../../serialization" }
 common = { path = "../../common/" }
+serialization = { path = "../../serialization" }
 
+bip39 = { version = "1.0", default-features = false} #TODO use `features = ["std"]`
+zeroize = "1.5.7"
```
TheQuantumPhysicist, line 15: "1.5"

wallet/Cargo.toml
```diff
@@ -9,14 +9,19 @@ rust-version.workspace = true
...
+crypto = { path = "../crypto/" }
 logging = { path = "../logging" }
 serialization = { path = "../serialization" }
 storage = { path = '../storage', features = ['inmemory'] }
 utxo = { path = '../utxo' }
 wallet-storage = {path = './storage'}
 wallet-types = {path = './types'}
 
+bip39 = { version = "1.0", default-features = false} #TODO use `features = ["std"]`
+zeroize = "1.5.7"
```
TheQuantumPhysicist, line 21: "1.5"

wallet/src/key_chain.rs
```diff
@@ -0,0 +1,186 @@
...
+            db,
+            lookahead_size: LOOKAHEAD_SIZE,
+        })
+    }
+
+    pub fn load_key_chain(db: Arc<Store<B>>) -> KeyChainResult<Self> {
+        // TODO remove this
+        let _ = db.get_storage_version().expect("This should work?");
+        // TODO implement loading from database
+        Ok(KeyChain {
```
TheQuantumPhysicist, line 110: To encrypt things you need to store, please use `chacha20poly1305`. It's already done in the crypto trait.

erasmospunk: Thanks, I will in a new PR.

wallet/src/key_chain.rs
```diff
@@ -0,0 +1,186 @@
...
+        Ok(address)
+    }
+
+    /// Get a new derived key that hasn't been used before
+    pub fn get_new_key(&self, purpose: KeyPurpose) -> KeyChainResult<ExtendedPrivateKey> {
+        // TODO implement with correct paths
+        let hd_path = match purpose {
+            KeyPurpose::ReceiveFunds => DerivationPath::from_str("m/0'/0'/0'/0'")?,
+            KeyPurpose::Change => DerivationPath::from_str("m/0'/0'/1'/0'")?,
+        };
```
iljakuklic, line 138: Could the derivation path be passed in as a slice of indices rather than string? Avoids potential parsing issues.

wallet/src/key_chain.rs
```diff
@@ -0,0 +1,186 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! # BIP39 key chain
+//! The KeyChain struct holds and constantly derives keys for the wallet addresses
+//! It uses the following derivation scheme:
+//!
+//! m/0'/<account_number>'/<key_purpose>'/<key_index>'
```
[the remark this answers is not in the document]
erasmospunk: This is because we use hardened derivation for key_purpose and key_index so this does not comply with the standards. We might re-define this path to be compatible with the BIP44 standard.

wallet/src/key_chain.rs
erasmospunk: Those will be created from ChildNumber directly.

wallet/src/key_chain.rs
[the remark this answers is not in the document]
iljakuklic: Some questions:
* Is there a plan to adopt the standard once soft derivations are supported?
* Would it make sense to adjust the path now to be more in sync with the standard? With the following modifications:
  * Not sure what to do with the initial `44'`. It's probably a good idea to change it to something else until the standard is followed correctly
  * Use something else for coin type, `0'` is bitcoin.
  * receiving/change and index components are hardened

TheQuantumPhysicist (approved):

PR #727, opened 2023-02-25
Add Instant to TimeGetter

pavel-kokolemin:
I need a monotonic clock in `PeerDb` where we can only use `TimeGetter`. Using the system clock is not a good option because it can go backwards. To fix this, I added the instant clock to `TimeGetter`.
I removed global `TIME_SOURCE` variable because it is not used in tests and is generally not very useful (doesn't work properly if tests run in parallel).

### DIFF
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -33,7 +33,7 @@ use common::{
         Block, ChainConfig, GenBlock, GenBlockId, OutPointSourceId, Transaction,
     },
     primitives::{id::WithId, BlockDistance, BlockHeight, Id, Idable},
-    time_getter::TimeGetterFn,
+    time_getter::TimeGetter,
     Uint256,
 };
 use consensus::compute_extra_consensus_data;
@@ -63,7 +63,7 @@ pub struct ChainstateRef<'a, S, V> {
     chainstate_config: &'a ChainstateConfig,
     tx_verification_strategy: &'a V,
     db_tx: S,
-    time_getter: &'a TimeGetterFn,
+    time_getter: &'a TimeGetter,
 }
 
 impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> BlockIndexHandle
@@ -116,7 +116,7 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         chainstate_config: &'a ChainstateConfig,
         tx_verification_strategy: &'a V,
         db_tx: S,
-        time_getter: &'a TimeGetterFn,
+        time_getter: &'a TimeGetter,
     ) -> ChainstateRef<'a, S, V> {
         ChainstateRef {
             chain_config,
@@ -132,7 +132,7 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         chainstate_config: &'a ChainstateConfig,
         tx_verification_strategy: &'a V,
         db_tx: S,
-        time_getter: &'a TimeGetterFn,
+        time_getter: &'a TimeGetter,
     ) -> ChainstateRef<'a, S, V> {
         ChainstateRef {
             chain_config,
@@ -153,7 +153,7 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
     }
 
     pub fn current_time(&self) -> std::time::Duration {
-        (self.time_getter)()
+        self.time_getter.get_time()
     }
 
     pub fn get_best_block_id(&self) -> Result<Id<GenBlock>, PropertyQueryError> {
--- a/chainstate/src/detail/median_time.rs
+++ b/chainstate/src/detail/median_time.rs
@@ -46,7 +46,8 @@ mod test {
         detail::tx_verification_strategy::DefaultTransactionVerificationStrategy, BlockSource,
         Chainstate, ChainstateConfig,
     };
-    use common::time_getter::TimeGetter;
+    use common::primitives::time;
+    use test_utils::mock_time_getter::mocked_time_getter_seconds;
 
     use super::*;
     use chainstate_storage::inmemory::Store;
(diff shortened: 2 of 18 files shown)


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -63,7 +63,7 @@ pub struct ChainstateRef<'a, S, V> {
     chainstate_config: &'a ChainstateConfig,
     tx_verification_strategy: &'a V,
     db_tx: S,
-    time_getter: &'a TimeGetterFn,
+    time_getter: &'a dyn TimeGetterFn,
```
TheQuantumPhysicist, line 66: I think the right thing to put here is `TimeGetter`

TheQuantumPhysicist:
> I need a monotonic clock in `PeerDb` where we can only use `TimeGetter`. Using the system clock is not a good option because it can go backwards. To fix this, I added the instant clock to `TimeGetter`. I removed global `TIME_SOURCE` variable because it is not used in tests and is generally not very useful (doesn't work properly if tests run in parallel).

Sorry, Pavel. We do need the global mocked time. You seem to have misunderstood its use. The global mocked time will be used in functional tests to test things like staking. The issue with PoS is that we need to have full control over time from the outside, otherwise stakes cannot happen. 

To clarify: The global mocked time isn't supposed to be used in unit tests. Only in functional tests. For unit tests, we use the `TimeGetter`.

chainstate/src/detail/chainstateref/mod.rs
pavel-kokolemin: OK, updated.

pavel-kokolemin:
> To clarify: The global mocked time isn't supposed to be used in unit tests. Only in functional tests. For unit tests, we use the `TimeGetter`.

I see. I reverted `TIME_SOURCE`.

chainstate/src/detail/median_time.rs
```diff
@@ -46,7 +46,10 @@ mod test {
         detail::tx_verification_strategy::DefaultTransactionVerificationStrategy, BlockSource,
         Chainstate, ChainstateConfig,
     };
-    use common::time_getter::TimeGetter;
+    use common::{
```
TheQuantumPhysicist, line 49: I think you can change `primitives::time::{self}` to `primitives::time`

pavel-kokolemin: Right, thanks.

TheQuantumPhysicist (commented):
I'm not happy with this PR. This is becoming confusing.

Time should come from exactly one source, through exactly one function. There should be no way for any component in the software to know the real time except from that component. Anything else will risk inconsistencies in functional tests and will lead to vulnerabilities due to the inability of fully simulating real time progression.

pavel-kokolemin:
> Time should come from exactly one source, through exactly one function. 

Both now come from one source (`get_mocked_system_time`), when mocked:

```
fn get_mocked_system_time() -> Option<Duration> {
    ...
}

pub fn get_system_time() -> Duration {
    match get_mocked_system_time() {
        Some(mocked_time) => mocked_time,
        None => SystemTime::now()
            .duration_since(SystemTime::UNIX_EPOCH)
            .expect("Time went backwards"),
    }
}

pub fn get_instant_time() -> Instant {
    match get_mocked_system_time() {
        Some(mocked_time) => *BASE_MOCK_INSTANT + mocked_time,
        None => Instant::now(),
    }
}
```

Is it better now? `BASE_MOCK_INSTANT` is a constant.

TheQuantumPhysicist:
> > Time should come from exactly one source, through exactly one function.
> 
> Both now come from one source (`get_mocked_system_time`), when mocked:
> 
> ```
> fn get_mocked_system_time() -> Option<Duration> {
>     ...
> }
> 
> pub fn get_system_time() -> Duration {
>     match get_mocked_system_time() {
>         Some(mocked_time) => mocked_time,
>         None => SystemTime::now()
>             .duration_since(SystemTime::UNIX_EPOCH)
>             .expect("Time went backwards"),
>     }
> }
> 
> pub fn get_instant_time() -> Instant {
>     match get_mocked_system_time() {
>         Some(mocked_time) => *BASE_MOCK_INSTANT + mocked_time,
>         None => Instant::now(),
>     }
> }
> ```
> 
> Is it better now? `BASE_MOCK_INSTANT` is a constant.

It looks fine from the outside. I'll do a thorough review soon.

common/src/time_getter.rs
```diff
@@ -13,33 +13,137 @@
...
+    fn system_time(&self) -> Duration {
+        time::get_system_time()
+    }
+
+    fn instant(&self) -> Instant {
+        time::get_instant_time()
+    }
+}
+
+pub struct MockedMsecTimeGetterFn {
```
TheQuantumPhysicist, line 90: I think this can go.

TheQuantumPhysicist (commented):
Please remove that extra MockedMsecTimeGetterFn struct, then feel free to merge.

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

common/src/time_getter.rs
pavel-kokolemin: I think you missed that `MockedMsecTimeGetterFn` is used in unit tests (from the `TimeGetter::mocked_seconds` constructor). I don't see how we can remove it.

TheQuantumPhysicist: I don't think this should exist. We should just use the absolute time getters and instant getters.

pavel-kokolemin: Can you please elaborate on this? I'm still not sure what should be done here.
BTW, I made `MockedMsecTimeGetterFn` and `DefaultTimeGetterFn` private so they are not even visible from the outside.

TheQuantumPhysicist: Like I mentioned in the meeting, there has to be exactly one source of time in TimeGetter. There shouldn't be the possibility to query different kinds of time.

muursh (approved):

common/src/time_getter.rs
pavel-kokolemin: Maybe it would be better to close this PR and keep using `tokio::time::Instant` in p2p (where absolute time is not needed)?

TheQuantumPhysicist: The current usage of instant is a problem because it's decoupled from mocked time. We need this PR.

TheQuantumPhysicist (approved):
Looking good.

Please consider rebasing on master. I see duplicated changes.

pavel-kokolemin:
> Please consider rebasing on master. I see duplicated changes.

It's already rebased on latest master (0a5fd8086c68a75d31f8740b19a9f077f4b5c94e).

PR #729, opened 2023-02-27
Use task queue in the sync manager

stanislav-tkach:


stanislav-tkach:
The current implementation doesn't solve anything and only introduces new issues.

PR #730, opened 2023-02-27
Use biased select loop in the sync manager

stanislav-tkach:
I think that this is a much simpler alternative to https://github.com/mintlayer/mintlayer-core/pull/729. I'm not sure that the same approach that is used in the backend will work for the sync manager, so I have tried something different.

### DIFF
--- a/p2p/src/sync/mod.rs
+++ b/p2p/src/sync/mod.rs
@@ -43,7 +43,7 @@ use crate::{
     error::{P2pError, PeerError, ProtocolError},
     event::{PeerManagerEvent, SyncControlEvent},
     message::{Announcement, BlockListRequest, BlockResponse, HeaderListRequest, SyncMessage},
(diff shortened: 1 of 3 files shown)


pavel-kokolemin (approved):

TheQuantumPhysicist (commented):
I don't think this solves the problem. Yes, it's *unlikely* that a series of blocks block connections, but we still cannot do both concurrently. This can be seen as a temporary solution... unless I'm missing something.

stanislav-tkach:
I don't think there is a real issue with blocking because the sync manager doesn't establish connections itself and all its operations aren't blocking (in terms of async). Anyway, I have a different idea that I would like to explore, so this can be seen as a temporary improvement.

stanislav-tkach:
Superseded by https://github.com/mintlayer/mintlayer-core/pull/742.

Issue #621, opened 2023-01-06
Differentiate types of peer


pavel-kokolemin:
From @TheQuantumPhysicist 

> There's an issue that has been discussed earlier but has been omitted for a while and we have to come back to it sooner or later. The issue is that not all peers are equal. There are peers that we want to persistently have, and others that we want to just give a "shot" at connecting at. I think in Substrate they call the former "reserved" peers or something. We should also consider doing that. This is important because some times you have peers with whom you want to maintain a connection no matter what. I don't know if you're gonna do that in this PR, but please put that into consideration, whether in an issue or this PR. Please research this topic when you can.

PR #733, opened 2023-03-02
`PeerManager` updates

pavel-kokolemin:
From @TheQuantumPhysicist 

> Connecting from RPC will need a little more work, because we want to make it possible to maintain connections with a specific peer. So it's a superset of what addnode does in bitcoin. Substrate/Parity has 3 types of peers for this.
There's an issue about this.

I looked at Substrate but did not find 3 types there. I only saw 2 types there: normal and reserved.
RPC now has two commands to add and remove "reserved" node addresses. Adding a "reserved" node address with RPC won't initiate an outbound connection until `heartbeat` is called, and removing it won't disconnect currently connected peers. Reserved nodes are not banned.
I have renamed `addnode` to `reserved_node` everywhere because it sounds a bit better.
I also added a command line option to add "normal" peer addresses (`boot_node` option) at startup. This can be used if something is wrong with the DNS seeds, for example.
I think this PR partially resolves https://github.com/mintlayer/mintlayer-core/issues/621.

### DIFF
--- a/dns_server/src/crawler_p2p/crawler/tests/mod.rs
+++ b/dns_server/src/crawler_p2p/crawler/tests/mod.rs
@@ -129,13 +129,13 @@ fn randomized(#[case] seed: Seed) {
         })
         .collect::<Vec<_>>();
 
-    let added_count = rng.gen_range(0..5);
-    let added_nodes = nodes.choose_multiple(&mut rng, added_count).cloned().collect();
+    let reserved_count = rng.gen_range(0..5);
+    let reserved_nodes = nodes.choose_multiple(&mut rng, reserved_count).cloned().collect();
 
     let loaded_count = rng.gen_range(0..10);
     let loaded_nodes = nodes.choose_multiple(&mut rng, loaded_count).cloned().collect();
 
-    let mut crawler = test_crawler(loaded_nodes, added_nodes);
+    let mut crawler = test_crawler(loaded_nodes, reserved_nodes);
 
     for _ in 0..rng.gen_range(0..100000) {
         crawler.timer(Duration::from_secs(rng.gen_range(0..100)), &mut rng);
@@ -253,7 +253,7 @@ fn long_offline(#[case] seed: Seed) {
     );
     assert!(crawler.persistent.contains(&loaded_node));
 
-    // Reachable and added nodes are offline for two month
+    // Reachable and reserved nodes are offline for two month
     for _ in 0..24 * 60 {
         crawler.timer(Duration::from_secs(3600), &mut rng);
         for address in crawler.pending_connects.clone() {
--- a/p2p/src/interface/p2p_interface_impl_delegation.rs
+++ b/p2p/src/interface/p2p_interface_impl_delegation.rs
@@ -42,4 +42,12 @@ impl<T: Deref<Target = dyn P2pInterface> + DerefMut<Target = dyn P2pInterface> +
     async fn get_connected_peers(&self) -> crate::Result<Vec<ConnectedPeer>> {
         self.deref().get_connected_peers().await
     }
+
+    async fn add_reserved_node(&mut self, addr: String) -> crate::Result<()> {
+        self.deref_mut().add_reserved_node(addr).await
+    }
+
+    async fn remove_reserved_node(&mut self, addr: String) -> crate::Result<()> {
+        self.deref_mut().remove_reserved_node(addr).await
+    }
 }
--- a/p2p/src/peer_manager/peerdb/address_data/mod.rs
+++ b/p2p/src/peer_manager/peerdb/address_data/mod.rs
@@ -57,23 +57,23 @@ pub enum AddressStateTransitionTo {
 pub struct AddressData {
     state: AddressState,
 
-    user_added: ConstValue<bool>,
+    reserved: ConstValue<bool>,
 }
 
 impl AddressData {
-    pub fn new(was_reachable: bool, user_added: bool, now: Instant) -> Self {
+    pub fn new(was_reachable: bool, reserved: bool, now: Instant) -> Self {
         AddressData {
             state: AddressState::Disconnected {
                 was_reachable,
                 fail_count: 0,
                 disconnected_at: now,
             },
-            user_added: user_added.into(),
+            reserved: reserved.into(),
         }
     }
 
-    pub fn user_added(&self) -> bool {
-        *self.user_added
+    pub fn reserved(&self) -> bool {
+        *self.reserved
     }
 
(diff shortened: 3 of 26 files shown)


dns_server/src/crawler_p2p/crawler/tests/mod.rs
```diff
@@ -129,13 +129,13 @@ fn randomized(#[case] seed: Seed) {
         })
         .collect::<Vec<_>>();
 
-    let added_count = rng.gen_range(0..5);
-    let added_nodes = nodes.choose_multiple(&mut rng, added_count).cloned().collect();
+    let reserved_count = rng.gen_range(0..5);
```
TheQuantumPhysicist, line 132: It's always better to randomize this 5 and that 10 and create a model that covers them both with the relation between them. This applies to all magic numbers.

p2p/src/interface/p2p_interface_impl_delegation.rs
```diff
@@ -42,4 +42,12 @@ impl<T: Deref<Target = dyn P2pInterface> + DerefMut<Target = dyn P2pInterface> +
     async fn get_connected_peers(&self) -> crate::Result<Vec<ConnectedPeer>> {
         self.deref().get_connected_peers().await
     }
+
+    async fn add_reserved_node(&mut self, addr: String) -> crate::Result<()> {
```
TheQuantumPhysicist, line 46: What's the point of the `connect` function if we have this?

p2p/src/peer_manager/peerdb/address_data/tests.rs
```diff
@@ -0,0 +1,58 @@
...
+
+            let is_valid_transition = match transition {
+                AddressStateTransitionTo::Connected => !address_data.is_connected(),
+                AddressStateTransitionTo::Disconnected => address_data.is_connected(),
+                AddressStateTransitionTo::ConnectionFailed => !address_data.is_connected(),
+                AddressStateTransitionTo::SetReserved => true,
+                AddressStateTransitionTo::UnsetReserved => true,
+            };
+
+            if is_valid_transition {
```
TheQuantumPhysicist, line 53: What if it's not valid? No asserts? Nothing?

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -125,30 +142,53 @@ where
...
-        connected_outbound: &BTreeSet<A>,
+        connected_outbound_count: usize,
     ) -> Vec<A> {
+        let now = Instant::now();
         let count = MAX_OUTBOUND_CONNECTIONS
             .saturating_sub(pending_outbound.len())
-            .saturating_sub(connected_outbound.len());
+            .saturating_sub(connected_outbound_count);
 
         // TODO: Ignore banned addresses
```
TheQuantumPhysicist, line 156: That sounds like a too-important bit to ignore for later and just throw in as an "unimportant TODO".

TheQuantumPhysicist (approved):
Looks fine. I'm just worried that there are way too many TODOs.

p2p/src/interface/p2p_interface_impl_delegation.rs
TheQuantumPhysicist: Now, thinking about it, connect can be useful to just give a "one shot"... but then we should make this distinction clear.

p2p/src/peer_manager/peerdb/address_data/mod.rs
```diff
@@ -47,30 +46,47 @@ pub enum AddressState {
     },
 }
 
+#[derive(Copy, Clone, Debug)]
+// Update `ALL_TRANSITIONS` if a new transition is added!
```
stanislav-tkach, line 50: [`EnumIter` from the `strum` crate](https://docs.rs/strum_macros/0.24.3/strum_macros/derive.EnumIter.html) can be used instead.

stanislav-tkach (approved):

dns_server/src/crawler_p2p/crawler/tests/mod.rs
pavel-kokolemin: Sorry, I'm not sure what you mean. I just wanted to allow testing of corner cases and this should be covered here.

p2p/src/interface/p2p_interface_impl_delegation.rs
pavel-kokolemin: I updated comments.

p2p/src/peer_manager/peerdb/address_data/tests.rs
pavel-kokolemin: Nothing, the check is only needed to generate sane randomized input.

p2p/src/peer_manager/peerdb/mod.rs
pavel-kokolemin: I added the check for banned addresses.

p2p/src/peer_manager/peerdb/address_data/mod.rs
pavel-kokolemin: Right, thanks. It's only for tests so I think we can keep it.

dns_server/src/crawler_p2p/crawler/tests/mod.rs
TheQuantumPhysicist: Instead of 5, use a random number. Instead of 10, use a random number. Define the boundaries of the random number so it covers a wide range over time.

pavel-kokolemin: Could you please write how it should look? I think it would be easier for me to understand.

TheQuantumPhysicist: As agreed in the call. This is OK.

p2p/src/interface/p2p_interface_impl_delegation.rs
pavel-kokolemin: Here are the updated comments:

```
    /// Try to connect to a remote node (just once).
    /// For persistent connections `add_reserved_node` should be used.
    #[method(name = "connect")]
    async fn connect(&self, addr: String) -> rpc::Result<()>;

    /// Add the address to the reserved nodes list.
    /// The node will try to keep connections open to all reserved peers.
    #[method(name = "add_reserved_node")]
    async fn add_reserved_node(&self, addr: String) -> rpc::Result<()>;

    /// Remove the address from the reserved nodes list.
    /// Existing connection to the peer is not closed.
    #[method(name = "remove_reserved_node")]
    async fn remove_reserved_node(&self, addr: String) -> rpc::Result<()>;
```

PR #734, opened 2023-03-02
Spend stake pool output in block reward

azarovh:
In our PoS in order to produce a block you have to spend `StakePool` output in a block reward (which is also called a kernel) that will produce reward output with a total amount of staked coins + subsidy + fees.
I introduced a new output type specifically for this case: `SpendStakePool`. It will be output for block reward and also can be reused as a kernel to produce more blocks.

General rules for spending stake outputs are as follows:
- `StakePool` and `SpendStakePool` cannot be spent in a transaction. Additional functionality of decommissioning will be added later.
- `SpendStakePool` output can only be produced from `StakePool` or other `SpendStakePool` inputs.
- Though `SpendStakePool` is time-locked it can be used to produce a block.
- For now, block reward can have a single input and single output. 
- Burning outputs in a reward is forbidden.

P.S. there was an issue with randomness in `PoSData` so I added additional field to the struct please pay attention. Basically, we used single vrf proof to verify both randomnesses of the sealed epoch and current block.


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -391,55 +392,56 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
 
     fn check_block_reward_maturity_settings(&self, block: &Block) -> Result<(), CheckBlockError> {
         let required = block.consensus_data().reward_maturity_distance(self.chain_config);
+        let timelock_check_fn = |tl: &OutputTimeLock| -> Result<(), CheckBlockError> {
```
TheQuantumPhysicist, line 395: If you're not capturing any variables, just move this to a separate function.

common/src/chain/transaction/output.rs
```diff
@@ -41,26 +41,33 @@ pub enum OutputPurpose {
...
-    StakePool(Box<StakePoolData>),
-    #[codec(index = 3)]
     Burn,
+    /// Output type that is used to create a stake pool
+    #[codec(index = 3)]
+    StakePool(Box<StakePoolData>),
+    /// Output type that represents spending of a stake pool output in a block reward
+    /// in order to produce a block
+    #[codec(index = 4)]
+    SpendStakePool(Box<StakePoolData>, OutputTimeLock),
```
TheQuantumPhysicist, line 51: I wouldn't call this "Spend"... Maybe something like `StakeToProduceBlock`? I'm open to suggestions here.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -391,55 +392,56 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
-                },
-                common::chain::OutputPurpose::StakePool(_) => {
+                OutputPurpose::LockThenTransfer(_, tl) => timelock_check_fn(tl)?,
+                OutputPurpose::StakePool(_) => {
                     return Err(CheckBlockError::InvalidBlockRewardOutputType(
                         block.get_id(),
                     ))
                 }
-                common::chain::OutputPurpose::Burn => {
+                OutputPurpose::SpendStakePool(_, tl) => timelock_check_fn(tl)?,
```
TheQuantumPhysicist, line 443: When staking, the limitation to spending shouldn't be applied to staking, but to decomissioning.

Meaning: If you stay, you can still stake with no limitation. But when you decommission, then you must way for a period of time.

common/src/chain/transaction/output.rs
TheQuantumPhysicist: Just to be clear, "spending" is the opposite of having an output available.

azarovh: > Just to be clear, "spending" is the opposite of having an output available.

I'd say that burning is the opposite but spending is always using an output and producing another. So here I spend StakePool and produce SpendStakePool.

But in general I agree that the name can be improved. I searched for the verb that better describes the intention. Maybe the purpose of the output is something like `UtiliseStakePool`/`ApplyStakePool`/`RecycleStakePool`?

azarovh: In `StakeToProduceBlock` haven't we already staked when a pool was created? The meaning of stake verb is a bit vague to me

chainstate/src/detail/chainstateref/mod.rs
azarovh: right

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -391,55 +392,45 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
-                        return Err(CheckBlockError::InvalidBlockRewardMaturityTimelockType(
-                            block.get_id(),
-                        ))
-                    }
-                    common::chain::timelock::OutputTimeLock::ForBlockCount(c) => {
-                        let cs: i64 = (*c)
-                            .try_into()
-                            .map_err(|_| {
-                                CheckBlockError::InvalidBlockRewardMaturityDistanceValue(
+                OutputPurpose::LockThenTransfer(_, tl) => {
```
TheQuantumPhysicist, line 398: You could've still used a separate function... no problem in that. I don't know why you reverted it.

azarovh: Originally I moved the code to a closure because it was called twice, but without it, I don't see the point in separate function it's not that big. If you think it's a good thing to do I will.

common/src/chain/block/consensus_data.rs
```diff
@@ -50,30 +50,43 @@ impl ConsensusData {
...
+
+    /// VRF data used for calculating hash below the target.
+    /// It represents random seed generated based on the randomness of the sealed epoch.
+    vrf_data_from_sealed_epoch: VRFReturn,
+
+    /// VRF data used for calculating randomness for the current block.
+    /// It represents random seed generated based on the randomness of the previous block.
+    vrf_data_from_prev_block: VRFReturn,
+
+    target: Compact,
```
TheQuantumPhysicist, line 71: `compact_target`?

chainstate/src/detail/chainstateref/mod.rs
TheQuantumPhysicist: If there are things around it, I think it's a good thing.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -364,12 +364,72 @@ where
...
+    fn check_stake_outputs_in_reward(
+        &self,
+        block: &WithId<Block>,
+    ) -> Result<(), ConnectTransactionError> {
+        match block.consensus_data() {
+            ConsensusData::None | ConsensusData::PoW(_) => Ok(()),
+            ConsensusData::PoS(_) => {
+                let block_reward_transactable = block.block_reward_transactable();
+
+                let kernel_input =
```
TheQuantumPhysicist, line 376: Isn't there a function, get_kernel_output, that does this for you? Why repeating the logic here?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -559,14 +621,50 @@ where
...
+                | OutputPurpose::Burn
+                | OutputPurpose::StakePool(_) => false,
+                OutputPurpose::SpendStakePool(_) => true,
+            });
+
+        ensure!(
+            !attempt_to_use_spend_stake_pool,
+            ConnectTransactionError::AttemptToSpendStakedCoins
+        );
+
```
TheQuantumPhysicist, line 658: Where's the part where we increase the balance of the staking pool?

There's also the question of distributing rewards. We should have a discussion about this.

common/src/chain/block/consensus_data.rs
```diff
@@ -50,30 +50,43 @@ impl ConsensusData {
...
-    vrf_data: VRFReturn,
-    bits: Compact,
+
+    /// VRF data used for calculating hash below the target.
+    /// It represents random seed generated based on the randomness of the sealed epoch.
+    vrf_data_from_sealed_epoch: VRFReturn,
+
+    /// VRF data used for calculating randomness for the current block.
+    /// It represents random seed generated based on the randomness of the previous block.
+    vrf_data_from_prev_block: VRFReturn,
```
TheQuantumPhysicist, line 69: This is wrong. There should be only one vrf_data.

common/src/chain/transaction/output.rs
```diff
@@ -41,26 +41,33 @@ pub enum OutputPurpose {
...
-    StakePool(Box<StakePoolData>),
-    #[codec(index = 3)]
     Burn,
+    /// Output type that is used to create a stake pool
+    #[codec(index = 3)]
+    StakePool(Box<StakePoolData>),
+    /// Output type that represents spending of a stake pool output in a block reward
+    /// in order to produce a block
+    #[codec(index = 4)]
+    SpendStakePool(Box<StakePoolData>),
```
TheQuantumPhysicist, line 51: If you're gonna keep `StakePoolData` as the contents of the arm, what's exactly the point of creating a new arm?

consensus/src/pos/mod.rs
```diff
@@ -124,15 +133,13 @@ where
...
-    let kernel_output = get_kernel_output(pos_data, utxos_view)?;
-
-    let epoch_index =
-        chain_config.epoch_index_from_height(&prev_block_index.block_height().next_height());
-
-    let random_seed = randomness_of_sealed_epoch(chain_config, epoch_index, block_index_handle)?;
+    let current_height = prev_block_index.block_height().next_height();
+    let random_seed = randomness_of_sealed_epoch(chain_config, current_height, block_index_handle)?;
 
+    let kernel_output = get_kernel_output(pos_data, utxos_view)?;
```
TheQuantumPhysicist, line 139: I don't think this will work. We have to think about this better.

The problem is that we're checking this at the wrong stage. This should be checked when connecting blocks. The reason is that this depends on the history of the blockchain, and anything that depends on the history should be strictly done when connecting the block. Otherwise we'll potentially mischaracterize spends as double-spends when there are forks.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: It's not there. I realized how the reward should be distributed only yesterday after a discussion with you. Will have to add it

common/src/chain/transaction/output.rs
azarovh: good question. Added the type to separate logic of processing the outputs. But now agree that it seems like possible with only one type

common/src/chain/block/consensus_data.rs
azarovh: I don't see how it's possible with only one vrf data. It is used in 2 use cases: sealed epoch randomness + vrf data to mine a block; prev block randomness + vrf data to create a randomness for the new block which will become sealed randomness at some point.

consensus/src/pos/mod.rs
azarovh: not sure that I understand what you mean. Kernel output is required to get the vrf pub key and calculate hash to compare with target. So basically the whole PoS consensus depends on the history.

common/src/chain/transaction/output.rs
TheQuantumPhysicist: Like I mentioned before, the benefit of using a different type is saving space in the blockchain by having only the minimal relevant information to stake.

consensus/src/pos/mod.rs
TheQuantumPhysicist: This is the paradox I mentioned in the meeting today.

common/src/chain/transaction/output.rs
azarovh: I see. I could use something like `ProduceBlockFromStake(OutPoint)`. Thus it will reference to the stake data and save space on every block. I'll give it a try, do you agree?

TheQuantumPhysicist: I don't think it's enough. You need a little more information, like the stake pool id, at least. Maybe you can strangle that from the input? I don't know. Give it a shot.

azarovh: I mean using OutPoint I can read StakePool output and get all the information from it

azarovh: Apparently having only `OutPoint` is not enough to retrieve information to verify transactions in `TransactionVerifier` because the transaction index is optional. So I have to put at least some amount of information into the output. 

I'd propose not to proceed with this at the moment and just clone `StakePoolData` because I'm not sure which data fields are going to be used to implement all the operations we want. I think it's better to place a TODO and optimize it later.

TheQuantumPhysicist: I don't mind, but I don't understand the logic. You could copy the struct and then remove those you don't need.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: done, for now all reward goes to a pool

common/src/chain/block/consensus_data.rs
azarovh: done, from now on a single vrf is used. I reconsidered the code and removed `PreconnectData` and related logic. Every block generates randomness based on a sealed epoch. And the randomness of the last block is used as a randomness of an epoch.

consensus/src/pos/mod.rs
azarovh: yeah, it doesn't work already. No reorg can be performed because of this - tried to add `pos_processing_tests::check_pool_balance_after_reorg` test. I believe it's the next thing to do

TheQuantumPhysicist: Reorgs can be done as long as there's no conflict. The bigger issue is that we won't be able to distinguish between a true double-spend attempt and an honest attempt to add a block, and an honest reorg that causes respending relative to a previous state.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -915,15 +891,19 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
...
+        let tip_height = tip_index.block_height();
         epoch_seal::update_epoch_seal(
             &mut self.db_tx,
             self.chain_config,
-            epoch_seal::BlockStateEvent::Connect(tip_index.block_height()),
+            epoch_seal::BlockStateEvent::Connect(tip_height),
         )?;
+        let tip = self
+            .get_block_from_index(tip_index)?
+            .ok_or(BlockError::BlockAtHeightNotFound(tip_height))?;
```
TheQuantumPhysicist, line 902: This error is more like an invariant is broken, because we already connected that block so it should exist already. Can you create a new error with something like `InvariantBrokenBlockNotFoundAfterConnect`?

TheQuantumPhysicist: Perhaps we should change this to pass the block to both functions when connecting the tip and post-connecting... it's very inefficient to do it this one.

Maybe just throw in a TODO at the connect loop, under the comment "Connect the new chain"

consensus/src/pos/mod.rs
```diff
@@ -34,36 +34,24 @@ use crate::pos::{error::ConsensusPoSError, kernel::get_kernel_output};
...
     epoch_index: EpochIndex,
-    random_seed: &H256,
+    random_seed: &PoSRandomness,
     pos_data: &PoSData,
     kernel_output: &TxOutput,
     spender_block_header: &BlockHeader,
     pos_accounting_view: &P,
 ) -> Result<(), ConsensusPoSError> {
-    let target: Uint256 = (*pos_data.bits())
+    let target: Uint256 = (*pos_data.target())
```
TheQuantumPhysicist, line 43: The retarget algorithm is missing. With the current implementation, the block producer can put any target they want. There should be an algorithm that enforces that every target in the current block follows a given recipe.

chainstate/types/src/pos_randomness.rs
```diff
@@ -47,21 +45,13 @@ impl PoSRandomness {
     }
```
TheQuantumPhysicist, line 46: H256 implements the Copy trait... no reason to deprive PoSRandomness of that.

consensus/src/pos/mod.rs
azarovh: yes, I know, thought about adding it separately

common/src/chain/config/mod.rs
```diff
@@ -215,6 +215,18 @@ impl ChainConfig {
         self.is_last_block_in_epoch(height) && current_epoch_index >= sealed_epoch_distance_from_tip
     }
 
+    pub fn sealed_epoch_index(&self, height: &BlockHeight) -> Option<EpochIndex> {
```
TheQuantumPhysicist, line 218: Please write tests for this function. The arithmetic can easily be done wrong with bad consequences.

pos_accounting/src/pool/operations.rs
```diff
@@ -65,6 +64,12 @@ pub struct SpendFromShareUndo {
     pub(crate) amount: Amount,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode)]
+pub struct IncreasePoolBalanceUndo {
+    pub(crate) pool_id: PoolId,
+    pub(crate) amount_to_add: Amount,
```
TheQuantumPhysicist, line 70: amount_added

pos_accounting/src/pool/operations.rs
```diff
@@ -73,6 +78,7 @@ pub enum PoSAccountingUndo {
     CreateDelegationId(CreateDelegationIdUndo),
     DelegateStaking(DelegateStakingUndo),
     SpendFromShare(SpendFromShareUndo),
+    IncreasePoolBalance(IncreasePoolBalanceUndo),
```
TheQuantumPhysicist, line 81: Just for completeness, keep in mind that in the future we will need a `DecreasePoolbalance` thing too.

I thought of combining them, but then I changed my mind. Perhaps keeping them separated makes testing easier.

chainstate/test-suite/src/tests/pos_processing_tests.rs
```diff
@@ -419,17 +484,451 @@ fn pos_invalid_pool_id(#[case] seed: Seed) {
...
+            .unwrap()
+            .unwrap();
+    let total_subsidy =
+        tf.chainstate.get_chain_config().block_subsidy_at_height(&BlockHeight::from(1)) * 3;
+    let initially_staked = Amount::from_atoms(1);
+    assert_eq!(
+        (total_subsidy.unwrap() + initially_staked).unwrap(),
+        res_pool_balance
+    );
 }
```
TheQuantumPhysicist, line 934: Please also add a test for using bogus targets to have an easy difficulty.

common/src/chain/block/consensus_data.rs
TheQuantumPhysicist: The issue is that target is H256. This is kind of ambiguous.

pos_accounting/src/pool/operations.rs
azarovh: Do you mean explicit Decrease and not undo of Increase? Curious what for? Penalties for misbehavior?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -559,14 +626,50 @@ where
         let input0_getter =
             || tx.inputs().get(0).ok_or(ConnectTransactionError::MissingOutputOrSpent);
 
+        let attempt_to_spend_stake = tx
+            .inputs()
+            .iter()
+            .map(|i| {
```
TheQuantumPhysicist, line 632: Please call `i` `input` instead.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -559,14 +626,50 @@ where
         let input0_getter =
             || tx.inputs().get(0).ok_or(ConnectTransactionError::MissingOutputOrSpent);
 
+        let attempt_to_spend_stake = tx
```
TheQuantumPhysicist, line 629: Instead of these adhoc checks, I'd rather have universal rules that apply to output type, that can here be directly utilized. I feel this isn't clean because it solves a general problem for one specific type. Namely: "A transaction is attempting to spend something it isn't supposed to do, no matter what". A cleaner solution would create a rule that can be tested independently and applied here. 

This applies to this and the next check.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -403,9 +463,15 @@ where
...
 
-        let max_allowed_outputs_total =
-            amount_sum!(inputs_total, block_subsidy_at_height.0, total_fees.0)
-                .ok_or_else(|| ConnectTransactionError::RewardAdditionError(block.get_id()))?;
+        let max_allowed_outputs_total = match block.header().consensus_data() {
+            ConsensusData::None | ConsensusData::PoW(_) => {
+                amount_sum!(inputs_total, block_subsidy_at_height.0, total_fees.0)
+            }
+            ConsensusData::PoS(_) => {
+                amount_sum!(inputs_total, total_fees.0)
```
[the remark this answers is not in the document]
azarovh: correct, the subsidy goes to the pool balance and not output

TheQuantumPhysicist:
Great job! Looking much better. We're getting closer. I'm done with this review round.

pos_accounting/src/pool/operations.rs
TheQuantumPhysicist: Cancelling delegation.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I'm not sure I understand the reasoning. There will be duplication, but this isn't a valid proposal.

pos_accounting/src/pool/operations.rs
azarovh: we have `PoSAccountingOperations::spend_share_from_delegation_id` for that

chainstate/tx-verifier/src/transaction_verifier/mod.rs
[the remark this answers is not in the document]
azarovh: My understanding was that subsidy goes to the pool balance, this way it can be later distributed among staker and delegators according to the rules. So only on decommission or "spend_delegation" will this amount be added to the output. 
Meaning right now `StakePool/ProduceBlock` outputs will be `original amount + fees` and `DecomissionPool` becomes `original amount + accumulated fees + accumulated subsidy`. Later we can apply formula that accounts delegators.
Otherwise if added to the output how can it be later distributed? Isn't it a double spend?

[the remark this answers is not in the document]
azarovh: @TheQuantumPhysicist please take a look at the comment above

common/src/chain/config/mod.rs
azarovh: I'll do these tests when I add target calculation for PoS

chainstate/test-suite/src/tests/pos_processing_tests.rs
azarovh: I'll do these tests when I add target calculation for PoS

common/src/chain/transaction/output.rs
```diff
@@ -41,26 +41,33 @@ pub enum OutputPurpose {
...
-    StakePool(Box<StakePoolData>),
-    #[codec(index = 3)]
     Burn,
+    /// Output type that is used to create a stake pool
+    #[codec(index = 3)]
+    StakePool(Box<StakePoolData>),
+    /// Output type that represents spending of a stake pool output in a block reward
+    /// in order to produce a block
+    #[codec(index = 4)]
+    ProduceBlockFromStake(Box<StakePoolData>),
```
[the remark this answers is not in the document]
azarovh: done

chainstate/tx-verifier/src/transaction_verifier/mod.rs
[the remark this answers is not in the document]
azarovh: just for the history, we decided to revert this change and leave subsidy in the utxo for the consistency reasons and easy access on the decommission

common/src/chain/config/mod.rs
```diff
@@ -437,4 +451,50 @@ mod tests {
...
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(0), None)]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(1), None)]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(2), None)]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(3), None)]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(4), None)]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(5), Some(0))]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(6), Some(0))]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(7), Some(1))]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(8), Some(1))]
+    #[case(NonZeroU64::new(2).unwrap(), 2, BlockHeight::from(9), Some(2))]
```
TheQuantumPhysicist, line 487: Please add higher values.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: Why was this marked resolved?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -547,6 +620,50 @@ where
...
+            ConnectTransactionError::AttemptToSpendStakedCoins
+        );
+        Ok(())
+    }
+
+    fn check_no_produce_block_outputs_in_tx(
+        &self,
+        tx: &Transaction,
+    ) -> Result<(), ConnectTransactionError> {
+        let attempt_to_use_probuce_block =
```
TheQuantumPhysicist, line 651: probuce

TheQuantumPhysicist (approved):
Looking good, besides the comments. Please look into them.

The transaction verifier is becoming huge and complex. There must be an active effort to split it to smaller modules that can be individually tested. I have to be frank and say that I'm not happy with the way things are going in it. It's important in security applications to be able to separate smaller units, isolate them, test them. This is how we reduce the number of possible states and cover them all. The bigger a module, the harder it's to cover all possible states.

This effort to reduce the number of state should be carried after remove OutputValue from TxOutput. I will add it to #697.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: misunderstood your idea, please take a look now

TheQuantumPhysicist (approved):

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -0,0 +1,86 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common::chain::{OutputPurpose, Transaction, TxOutput};
+
+use super::error::ConnectTransactionError;
+
+/// Not all `OutputPurposes` can be used in a transaction.
+/// For example spending `ProduceBlockFromStake` and `StakePool` in a tx is not supported
```
TheQuantumPhysicist, line 21: I would say "not allowed". It's not like we're gonna allow it in the future :smile:

azarovh: I think we will if the output type is DecommissionPool

Issue #718, opened 2023-02-21
Race in `SyncManager`


pavel-kokolemin:
It looks like there is a race when `SyncManager` gets `SyncingEvent::Message` before `SyncControlEvent::Connected`. This is possible because messages are sent directly from the backend, but connected/disconnected messages are sent from `PeerManager`.
It would be nice to fix this problem, because otherwise we might get into a situation where some nodes don't respond to valid header list requests. Also, banned peers could send requests before being disconnected by `PeerManager`.

PR #736, opened 2023-03-03
Fix race when peers connect

pavel-kokolemin:
Peers would not be allowed to read messages from the network socket until they are accepted as valid by `PeerManager`.
Notifications to `BlockSyncManager` about new accepted peers are sent from the backend (not from `PeerManager`) to avoid race conditions between accepting a new peer and starting to send messages.
This should fix https://github.com/mintlayer/mintlayer-core/issues/718.

### DIFF
--- a/p2p/src/net/default_backend/backend.rs
+++ b/p2p/src/net/default_backend/backend.rs
@@ -250,7 +250,7 @@ where
 
                 // Handle commands.
                 command = self.cmd_rx.recv() => {
-                    self.handle_command(command.ok_or(P2pError::ChannelClosed)?).await?;
+                    self.handle_command(command.ok_or(P2pError::ChannelClosed)?);
                 },
                 // Process pending commands
                 callback = self.command_queue.select_next_some(), if !self.command_queue.is_empty() => {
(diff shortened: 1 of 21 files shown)


p2p/src/net/default_backend/backend.rs
```diff
@@ -58,6 +58,9 @@ struct PeerContext {
     /// Note that sending may fail unexpectedly if the connection is closed!
     /// Do not propagate ChannelClosed error to the higher level, handle it locally!
     tx: mpsc::UnboundedSender<Event>,
+
+    /// True if the peer was accepted by PeerManager and SyncManager was notified
+    was_accepted: bool,
```
TheQuantumPhysicist, line 63: While boolean looks to be valid, but it's the wrong representation of this value. We should use a flag, where once it's set, it cannot be unset. Isn't this what we're trying to represent, here?

p2p/src/net/default_backend/backend.rs
```diff
@@ -523,63 +570,64 @@ where
...
 
                     boxed_cb(move |this| this.handle_connect_res(address, connection_res))
                 }
-                .boxed()
+                .boxed();
+
+                self.command_queue.push(backend_task);
             }
-            Command::Disconnect { peer_id } => async move {
-                boxed_cb(move |this: &mut Self| {
```
TheQuantumPhysicist, line 549: Why are we removing these?

TheQuantumPhysicist (approved):
Nice work. Please see what you can do about my comments and wait for Stanislav to review.

p2p/src/net/default_backend/backend.rs
pavel-kokolemin: Yes, that's right. But do we have such type? I can't find anything.

TheQuantumPhysicist: Either create it, or throw in a TODO, explaining this problem.

If you're going with a TODO, please explain the problem fully. We have a problem of very short and not-understandable TODO some times.

p2p/src/net/default_backend/backend.rs
pavel-kokolemin: This code was added when the channels were bounded and sending required `await`. But after later refactoring, when we switched to unbounded channels, this code became just overhead.
For example, for each received message, we first construct a future, push it into the `FuturesUnordered` list, and this future is immediately polled on a next event loop iteration.

p2p/src/net/default_backend/backend.rs
TheQuantumPhysicist: That's understandable. It's a best-effort technique.

p2p/src/net/default_backend/peer.rs
```diff
@@ -201,16 +201,19 @@ where
             }
         }
 
+        let mut was_accepted = false;
```
stanislav-tkach, line 204: This variable can probably be of the `SetFlag` type.

p2p/src/net/default_backend/backend.rs
```diff
@@ -523,63 +571,64 @@ where
...
-        self.command_queue.push(backend_task);
-
-        Ok(())
+    fn send_sync_event(sync_tx: &mpsc::UnboundedSender<SyncingEvent>, event: SyncingEvent) {
+        // SyncManager should always be active and so sending to a closed `conn_tx` is not a backend's problem, just log the error.
+        // NOTE: `sync_tx` is not connected in some PeerManager tests.
+        let res = sync_tx.send(event);
+        if res.is_err() {
+            log::error!("sending sync event from the backend failed unexpectedly");
+        }
```
stanislav-tkach, line 631: Why not propagate an error here?

stanislav-tkach (approved):

p2p/src/net/default_backend/peer.rs
pavel-kokolemin: Sure, I will update it next time, thanks.

p2p/src/net/default_backend/backend.rs
pavel-kokolemin: I think the code will be better if we ignore errors from closed channels (where it's possible).
If BlockSyncManager's channel is closed, then either the node is shutting down or BlockSyncManager has unexpectedly stopped.
If the node is shutting down, then the backend's even loop will terminate on the next iterations (once both BlockSyncManager and PeerManager are destroyed).
And if BlockSyncManager has stopped unexpectedly, then this should be handled appropriately where BlockSyncManager was started (I added TODO for this):
```
    tokio::spawn(async move {
        sync::BlockSyncManager::<T>::new(
            chain_config,
            p2p_config,
            sync,
            chainstate_handle,
            rx_p2p_sync,
            tx_peer_manager,
        )
        .run()
        .await
        .tap_err(|err| log::error!("SyncManager failed: {err}"))
    });
```

PR #740, opened 2023-03-04
Add socks5 support

pavel-kokolemin:


p2p/src/config.rs
```diff
@@ -66,6 +66,8 @@ impl From<NodeType> for BTreeSet<PubSubTopic> {
 pub struct P2pConfig {
     /// Address to bind P2P to.
     pub bind_addresses: Vec<String>,
+    /// SOCKS5 proxy.
+    pub proxy: Option<String>,
```
TheQuantumPhysicist, line 70: Please call this `socks5_proxy`... we may support other types of proxies in the future.

I see you called this just `proxy` in many please... please rename them all to `socks5_proxy` where it's exclusive for socks5 proxy.

TheQuantumPhysicist (approved):
Thank you. Very nice work! I'm very proud we made that adapter design, which made adding this proxy very easy.

I have two concerns:

1. What if the user chose multiple proxys? Do we want to allow that?

2. This lacks tests. Testing this isn't going to be easy, but it's neither impossible nor a priority. A typical way to test this is to create a forwarding proxy with a socks interface that can test that things sent through the proxy are delivered to the other end.

But again, not a priority. Adding TODOs or creating an issue is good enough for now.

p2p/src/config.rs
pavel-kokolemin: Sure, updated.

pavel-kokolemin:
> I'm very proud we made that adapter design, which made adding this proxy very easy.

Right, it worked well. Thank you for the clean design.

> What if the user chose multiple proxys? Do we want to allow that?

Good question. I just checked and the node fails to start if multiple proxies are specified on the command line:
```
error: the argument '--p2p-socks5-proxy <P2P_SOCKS5_PROXY>' cannot be used multiple times
```
Supporting multiple proxies or proxy chains should not be very difficult, but I don't think we should bother with that (Bitcoin Core does not support it).

> This lacks tests.
... Adding TODOs or creating an issue is good enough for now.

TODO added.
And I tested the node with socks5 server (`ssh -D`) and it worked as expected.

I also switched from `SocketAddr` to `String` so users can use host names (for example, `localhost:9050`)

stanislav-tkach (approved):

PR #742, opened 2023-03-05
Spawn tasks for peers in the sync manager

stanislav-tkach:


p2p/src/sync/mod.rs
```diff
@@ -177,27 +149,41 @@ where
...
+    /// Stops the tasks of the given peer by closing the corresponding channel.
     fn unregister_peer(&mut self, peer: PeerId) {
         log::debug!("Unregister peer {peer} from sync manager");
 
-        // Remove the queued block responses associated with the disconnected peer.
-        self.blocks_queue.retain(|(p, _)| p != &peer);
-
         if self.peers.remove(&peer).is_some() {
             log::warn!("Unregistering unknown peer: {peer}");
         }
```
pavel-kokolemin, line 189: I just noticed that there is a bug: `is_some` is used instead of `is_none`.

pavel-kokolemin:
Nice work!

I think as a next step we can add a pair of bounded channels between the peer loop in the backend and the peer loop in the sync manager (separate for each connected peer) and start sending sync messages directly.
The peers won't be able to push too many messages and overload sync manager, and the sync manager won't be able to push too many messages before they are sent over the network.
We will need to think about potential deadlocks with this design, but this should probably solve most of our problems with unbounded channels in p2p.

p2p/src/sync/peer.rs
```diff
@@ -357,4 +428,36 @@ where
...
+        if headers.len() > *self.p2p_config.max_request_blocks_count {
+            self.known_headers = headers.split_off(*self.p2p_config.max_request_blocks_count);
+        }
+
+        let block_ids: Vec<_> = headers.into_iter().map(|h| h.get_id()).collect();
+        self.message_sender.send((
+            self.id(),
+            SyncMessage::BlockListRequest(BlockListRequest::new(block_ids.clone())),
+        ))?;
+        self.requested_blocks.extend(block_ids);
```
pavel-kokolemin, line 447: Maybe we should first check which blocks have already been requested?
Otherwise we might request some blocks twice and the peer will be penalized because of this check:

```
        if self.requested_blocks.take(&block.get_id()).is_none() {
            return Err(P2pError::ProtocolError(ProtocolError::UnexpectedMessage(
                "block response",
            )));
        }
```

p2p/src/sync/peer.rs
```diff
@@ -291,13 +362,13 @@ where
...
             return Ok(());
         }
 
         let header_ = header.clone();
         self.chainstate_handle.call(|c| c.preliminary_header_check(header_)).await??;
-        self.request_blocks(peer, vec![header])
+        self.request_blocks(vec![header])
     }
 
     /// Handles a result of message processing.
```
pavel-kokolemin, line 374: This comment is probably outdated (fatal errors in the peer loop won't stop the whole p2p subsystem).

pavel-kokolemin (approved):

p2p/src/sync/mod.rs
```diff
@@ -118,42 +110,34 @@ where
...
+
+                // Resend messages from peers to the backend.
+                message = self.peer_receiver.recv() => {
+                    let (peer, message) = message.ok_or(P2pError::ChannelClosed)?;
+                    self.messaging_handle.send_message(peer, message)?;
+                },
+
+                event = self.messaging_handle.poll_next() => {
+                    self.handle_peer_event(event?)?;
+                },
```
pavel-kokolemin, line 134: `handle_peer_event` will return an error if some peer's channel is disconnected.
So at this line we will stop the sync manager's even loop if some peer's loop is stopped (for any reason).
Looks like it's OK right for now. But maybe it would be better to ignore the error here if the processing for some peer has been stopped unexpectedly?

p2p/src/sync/mod.rs
```diff
@@ -118,42 +110,34 @@ where
         log::info!("Starting SyncManager");
 
         let mut new_tip_receiver = self.subscribe_to_new_tip().await?;
-        self.is_initial_block_download =
-            self.chainstate_handle.call(|c| c.is_initial_block_download()).await??;
+        self.is_initial_block_download.store(
+            self.chainstate_handle.call(|c| c.is_initial_block_download()).await??,
+            Ordering::SeqCst,
```
TheQuantumPhysicist, line 115: `Ordering::Release`?

p2p/src/sync/mod.rs
TheQuantumPhysicist: I would prefer this to be handled in this PR correctly.

stanislav-tkach: Sure, fixed.

p2p/src/sync/mod.rs
```diff
@@ -221,58 +224,35 @@ where
...
-            .send_message(
-                peer,
-                SyncMessage::HeaderListRequest(HeaderListRequest::new(locator)),
-            )
-            .map(|_| ())
-    }
+        let peer_channel = match self.peers.get(&peer) {
+            Some(c) => c,
+            None => {
+                log::warn!("Received a message from unknown peer ({peer}): {event:?}");
```
TheQuantumPhysicist, line 247: I'm wondering whether we should make this an `expect()`...

p2p/src/sync/mod.rs
```diff
@@ -173,40 +157,59 @@ where
...
+    /// Stops the task of the given peer by closing the corresponding channel.
     fn unregister_peer(&mut self, peer: PeerId) {
         log::debug!("Unregister peer {peer} from sync manager");
 
-        // Remove the queued block responses associated with the disconnected peer.
-        self.blocks_queue.retain(|(p, _)| p != &peer);
-
-        if self.peers.remove(&peer).is_some() {
+        if self.peers.remove(&peer).is_none() {
             log::warn!("Unregistering unknown peer: {peer}");
```
TheQuantumPhysicist, line 199: Another part where I think maybe an expect is appropriate?

I'm really not sure. But maybe we should start with that and see what happens. We can throw in a TODO to keep watching it.

p2p/src/sync/mod.rs
```diff
@@ -173,40 +157,59 @@ where
...
         }
     }
 
     /// Announces the header of a new block to peers.
     async fn handle_new_tip(&mut self, block_id: Id<Block>) -> Result<()> {
-        if self.is_initial_block_download {
-            self.is_initial_block_download =
-                self.chainstate_handle.call(|c| c.is_initial_block_download()).await??;
+        if self.is_initial_block_download.load(Ordering::SeqCst) {
```
TheQuantumPhysicist, line 205: `Ordering::Acquire`?

p2p/src/sync/mod.rs
```diff
@@ -173,40 +157,59 @@ where
...
+            self.is_initial_block_download.store(
+                self.chainstate_handle.call(|c| c.is_initial_block_download()).await??,
+                Ordering::Release,
+            );
         }
 
-        if self.is_initial_block_download {
+        if self.is_initial_block_download.load(Ordering::SeqCst) {
```
TheQuantumPhysicist, line 212: Also `Ordering::Acquire`?

TheQuantumPhysicist: I would even use `Ordering::Relaxed` for this, because at this point, the worst is missing one block from this condition. The best is removing all synchronization primitives that will be done on every new tip.

TheQuantumPhysicist: I think calling this twice is a waste... and is bad practice. You could do this:

```
        let is_initial_block_download = if self.is_initial_block_download.load(Ordering::Relaxed) {
            let chainstate_ibd =
                self.chainstate_handle.call(|c| c.is_initial_block_download()).await??;
            self.is_initial_block_download.store(chainstate_ibd, Ordering::Release);
            chainstate_ibd
        } else {
            false
        };

        if is_initial_block_download {
            return Ok(());
        }
```

p2p/src/sync/peer.rs
```diff
@@ -13,56 +13,157 @@
...
+                    self.handle_result(res).await?;
+                }
             }
-            SyncMessage::BlockListRequest(r) => {
-                self.handle_block_request(peer, r.into_block_ids()).await
+        }
+    }
+
+    async fn request_headers(&mut self) -> Result<()> {
+        let locator = self.chainstate_handle.call(|this| this.get_locator()).await??;
```
TheQuantumPhysicist, line 132: Btw, one point to keep in mind that a locator is not an absolute value. I think it should also be possible to start from some point. Please check how bitcoin does it when negotiating headers. Something for another PR.

p2p/src/sync/peer.rs
```diff
@@ -13,56 +13,157 @@
...
+        }
+    }
+
+    /// Returns an identifier of the peer associated with this task.
+    pub fn id(&self) -> PeerId {
+        *self.id
+    }
+
+    pub async fn run(&mut self) -> Result<Void> {
+        self.request_headers().await?;
```
TheQuantumPhysicist, line 114: Is this the right place to put this? I doubt it. It's OK temporarily though.

I think requesting headers should be a part of a negotiation cycle that ends with both nodes being synced.

p2p/src/sync/peer.rs
```diff
@@ -13,56 +13,157 @@
...
+
+        loop {
+            tokio::select! {
+                event = self.events_receiver.recv() => {
+                    let event = event.ok_or(P2pError::ChannelClosed)?;
+                    self.handle_event(event).await?;
+                },
+
+                block = async { self.blocks_queue.pop_front().expect("The block queue is empty") }, if !self.blocks_queue.is_empty() => {
```
TheQuantumPhysicist, line 123: `block_to_send_to_peer = ...`

p2p/src/sync/peer.rs
```diff
@@ -13,56 +13,157 @@
...
+                    let res = self.send_block(block).await;
+                    self.handle_result(res).await?;
```
TheQuantumPhysicist, line 125: This is some weird logic I don't understand. Why are we handling both sending and receiving blocks to the peer with the same function?

When a peer sends us a block, that can be an attack, no question there.
But when we're sending a block, this can never be an attack.

These aren't the same.

p2p/src/sync/peer.rs
TheQuantumPhysicist: `self.requested_blocks` is a BTreeSet... I don't think this is possible. Unless I'm missing the point.

TheQuantumPhysicist (approved):
Looking good. Please take my comments into consideration.

p2p/src/sync/peer.rs
```diff
@@ -133,24 +225,15 @@ where
             })
             .await??;
 
-        peer_state.num_blocks_to_send += block_ids.len();
-        self.blocks_queue.extend(block_ids.into_iter().map(|id| (peer, id)));
+        self.blocks_queue.extend(block_ids.into_iter());
```
TheQuantumPhysicist, line 228: What happens if we receive the same request twice? `self.blocks_queue` is a `VecDeque`.

p2p/src/sync/peer.rs
stanislav-tkach: I think it can be done this way:
```rust
headers.retain(|h| self.requested_blocks.contains(&h.get_id()));
```

p2p/src/sync/mod.rs
stanislav-tkach: I think that assertion (expect) here will be appropriate here because the `unregister_peer` function should never be called twice, so it indicates an error elsewhere. I'm not so sure about the previous comment though.

TheQuantumPhysicist: It's always this difficult balance between "willing to crash the node if something goes wrong", and "possibility to miss the presence of an error".

Things like this, if we decide to keep them and log them, should be labeled "CRITICAL" in the logs to be noticeable as this should "never happen".

p2p/src/sync/peer.rs
stanislav-tkach: https://github.com/mintlayer/mintlayer-core/issues/747

p2p/src/sync/peer.rs
stanislav-tkach: In that case the node sends the requested blocks multiple times, but I'm not sure if a simple check would make a difference here. After all, a malicious peer can request different blocks trying to DoS our node.

Now I'm thinking if we should only allow consecutive block requests. So asking for the blocks with `1, 2, 3, 4, 5` indexes is OK, but `1, 3, 5` isn't. I'm not sure if it is a too severe restriction though.

TheQuantumPhysicist: A more reasonable check is whether they already have it. We should know.

TheQuantumPhysicist: Also keep in mind that bandwidth dos doesn't make sense, because we don't disallow unknown messages in the interest of upgrades, and it's easy to create a new node and just keep pulling blocks.

stanislav-tkach: Fixed.

PR #744, opened 2023-03-06
Add mock time RPC

pavel-kokolemin:
This PR adds the mock time RPC and command line option for functional tests.
Making mocked instant time work consistently complicated the code, so I removed it from the time getter (as it was before my recent changes). Small backward and any forward system clock corrections (for example after suspend) should not be a problem for the node to work properly.

### DIFF
--- a/node/src/mock_time.rs
+++ b/node/src/mock_time.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2022 RBB S.r.l
+// Copyright (c) 2023 RBB S.r.l
 // opensource@mintlayer.org
 // SPDX-License-Identifier: MIT
 // Licensed under the MIT License;
@@ -13,21 +13,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Top-level node runner as a library
+use common::chain::config::ChainType;
 
-mod config_files;
-mod mock_time;
-mod options;
-mod regtest_options;
-mod rpc;
-mod runner;
-
-pub type Error = anyhow::Error;
-
-pub use config_files::{NodeConfigFile, NodeTypeConfigFile, StorageBackendConfigFile};
-pub use options::{Command, Options, RunOptions};
-pub use runner::{initialize, run};
-
-pub fn init_logging(_opts: &Options) {
-    logging::init_logging::<&std::path::Path>(None)
+pub fn set_mock_time(chain_type: ChainType, time: u64) -> Result<(), crate::Error> {
+    anyhow::ensure!(
+        chain_type == ChainType::Regtest,
+        "Mock time allowed on regtest chain only"
+    );
+    anyhow::ensure!(time != 0, "Mock time cannot be 0");
+    common::primitives::time::set(common::primitives::time::duration_from_int(time))?;
+    Ok(())
 }
--- a/node/src/mock_time.rs
+++ b/node/src/mock_time.rs
@@ -20,7 +20,6 @@ pub fn set_mock_time(chain_type: ChainType, time: u64) -> Result<(), crate::Erro
         chain_type == ChainType::Regtest,
         "Mock time allowed on regtest chain only"
     );
-    anyhow::ensure!(time != 0, "Mock time cannot be 0");
     common::primitives::time::set(common::primitives::time::duration_from_int(time))?;
     Ok(())
 }
--- a/node/src/options.rs
+++ b/node/src/options.rs
@@ -62,6 +62,7 @@ pub struct RunOptions {
 
     /// Mock time.
     #[clap(long)]
+    #[arg(hide = true)]
     pub mock_time: Option<u64>,
 
     /// The number of maximum attempts to process a block.
--- a/node/src/options.rs
+++ b/node/src/options.rs
@@ -60,6 +60,10 @@ pub struct RunOptions {
     #[clap(long)]
     pub node_type: Option<NodeTypeConfigFile>,
 
+    /// Mock time.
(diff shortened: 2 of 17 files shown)


node/src/options.rs
```diff
@@ -60,6 +60,10 @@ pub struct RunOptions {
     #[clap(long)]
     pub node_type: Option<NodeTypeConfigFile>,
 
+    /// Mock time.
+    #[clap(long)]
+    pub mock_time: Option<u64>,
```
TheQuantumPhysicist, line 65: I don't get it. Why do we need this? Can't we set the time with rpc?

node/src/rpc.rs
```diff
@@ -47,8 +55,13 @@ impl NodeRpcServer for NodeRpc {
     fn version(&self) -> rpc::Result<String> {
         Ok(env!("CARGO_PKG_VERSION").into())
     }
+
+    fn set_mock_time(&self, time: u64) -> rpc::Result<()> {
+        crate::mock_time::set_mock_time(self.chain_type, time)?;
```
TheQuantumPhysicist, line 60: Please allow this only on regtest.

TheQuantumPhysicist: Oh, I noticed it's enforced in the function. Never mind.

node/src/mock_time.rs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+use common::chain::config::ChainType;
+
+pub fn set_mock_time(chain_type: ChainType, time: u64) -> Result<(), crate::Error> {
+    anyhow::ensure!(
+        chain_type == ChainType::Regtest,
+        "Mock time allowed on regtest chain only"
+    );
+    anyhow::ensure!(time != 0, "Mock time cannot be 0");
```
TheQuantumPhysicist, line 23: time == 0 means we revert to standard system time. It's not forbidden.

node/src/rpc.rs
```diff
@@ -47,8 +55,13 @@ impl NodeRpcServer for NodeRpc {
...
+        Ok(())
+    }
 }
 
-pub fn init(shutdown_trigger: ShutdownTrigger) -> rpc::Methods {
-    NodeRpc::new(shutdown_trigger).into_rpc().into()
+pub fn init(shutdown_trigger: ShutdownTrigger, chain_type: ChainType) -> rpc::Methods {
```
TheQuantumPhysicist, line 65: Perhaps it's better to pass an `Arc<ChainConfig>` here... no need to be so conservative about it.

node/src/runner.rs
```diff
@@ -201,6 +205,10 @@ async fn start(
     run_options: &RunOptions,
     chain_config: ChainConfig,
 ) -> Result<()> {
+    if let Some(mock_time) = run_options.mock_time {
+        set_mock_time(*chain_config.chain_type(), mock_time)?;
```
TheQuantumPhysicist, line 209: Please... no. This isn't necessary. Use RPC for this. Unless I'm missing something?

What's the benefit of setting a constant time value on startup?

node/tests/cli.rs
```diff
@@ -115,6 +115,7 @@ fn read_config_override_values() {
     let options = RunOptions {
         storage_backend: Some(backend_type.clone()),
         node_type: Some(node_type),
+        mock_time: None,
```
TheQuantumPhysicist, line 118: No test for this?

node/src/options.rs
pavel-kokolemin: It can be used to start the node in a "paused" state with a fixed mock time. I think it's nice to have when you need full control over the mock time. Without it, there is a potential race when some parts of the code see real system time and others see mocked time (depending on when the RPC call is made).
Bitcoin Core has "-mocktime" command line option for this:
https://github.com/bitcoin/bitcoin/blob/master/src/init.cpp#L987

TheQuantumPhysicist: I see.

I kind of imagined the way to do this would be by setting a "ready-set-go" kind of program, where an RPC program would fire up the program after a set of preparations.

I'm OK with this for now... though I wish there's a cleaner way to do this.

pavel-kokolemin:
> > This PR adds the mock time RPC and command line option for functional tests.
> > Making mocked instant time work consistently complicated the code, so I removed it from the time getter (as it was before my recent changes). Small backward and any forward system clock corrections (for example after suspend) should not be a problem for the node to work properly.
> 
> Could you elaborate on the problems you faced when having both instant and time?

I was unable to find a clean way for the instant time to change smoothly when the mock time is set/unset. It's not a problem if the mock time is used for the whole duration of the program run (when the mock time is set from command line at start and is never reset to 0 from RPC). But once set/reset is allowed at any moment, the mocked instant time becomes erratic. It's possible to overcome this with convoluted time arithmetic, but I decided it wasn't worth it.

node/src/mock_time.rs
pavel-kokolemin: I see. Fixed, thanks.

node/src/rpc.rs
pavel-kokolemin: Sure, updated.

node/tests/cli.rs
pavel-kokolemin: I'm not sure what to test here. `mock_time` is only present in `RunOptions`.

TheQuantumPhysicist (approved):

muursh (approved):

stanislav-tkach (approved):

PR #748, opened 2023-03-08
Functional p2p ping test

pavel-kokolemin:
I added a basic p2p functional test (based on the ping test from bitcoin core).
For it to work a p2p connection from the python is needed (to keep things simple, noise encryption is disabled there).
Bitcoin tests use the built-in python `struct` module for parsing network messages, where you serialize/deserialize values like this:

```
    def deserialize(self, f, *, with_time=True):
        """Deserialize from addrv1 format (pre-BIP155)"""
        if with_time:
            # VERSION messages serialize CAddress objects without time
            self.time = struct.unpack("<I", f.read(4))[0]
        self.nServices = struct.unpack("<Q", f.read(8))[0]
        # We only support IPv4 which means skip 12 bytes and read the next 4 as IPv4 address.
        f.read(12)
        self.net = self.NET_IPV4
        self.ip = socket.inet_ntoa(f.read(4))
        self.port = struct.unpack(">H", f.read(2))[0]
```

It should be possible to use the same approach here, but I decided to try `py-scale-codec` instead.
First you declare the binary format like this:

```
    custom_types = {
        "types": {
            "SemVer": {
                "type": "struct",
                "type_mapping": [
                    ["major", "u8"],
                    ["minor", "u8"],
                    ["patch", "u16"],
                ]
            },

            "PubSubTopic": {
                "type": "enum",
                "type_mapping": [
                    ["Transactions", "Null"],
                    ["Blocks", "Null"],
                    ["PeerAddresses", "Null"],
                ]
            },

            "HandshakeHello": {
                "type": "struct",
                "type_mapping": [
                    ["version", "SemVer"],
                    ["network", "[u8; 4]"],
                    ["subscriptions", "BTreeSet<PubSubTopic>"],
                    ["receiver_address", "Option<PeerAddress>"],
                    ["handshake_nonce", "u64"],
                ]
            },
        }
    }
```

And then you can encode and decode values like this:

```
    message = {
        "handshake": {
            "Hello": {
                "version": {
                    "major": 0,
                    "minor": 1,
                    "patch": 0,
                },
                "network": [0xaa, 0xbb, 0xcc, 0xdd],
                "subscriptions": ["Transactions", "Blocks", "PeerAddresses"],
                "receiver_address": None,
                "handshake_nonce": 123,
            }
        }
    }
    obj = scalecodec.base.RuntimeConfiguration().create_scale_object("Message")
    data = obj.encode(message)
```

`py-scale-codec` ensures that all fields are present.
But I did not find an easy way to load/install `py-scale-codec` on demand, so I copied its sources to our repo (about 160 kb without some extra stuff).

I noticed that in many places we use u32 and u64 to represent time as a command line arguments. I replaced them with a specialized `RpcTime` type (it represents time internally as a float number of seconds).

Some debug log checks are currently disabled in the functional tests. I think for this to work we would need to replace `env_logger` with something that can write log messages to files.


node/src/options.rs
```diff
@@ -103,23 +109,23 @@ pub struct RunOptions {
 
     /// The p2p timeout value in seconds.
     #[clap(long)]
-    pub p2p_outbound_connection_timeout: Option<NonZeroU64>,
+    pub p2p_outbound_connection_timeout: Option<RpcTime>,
```
TheQuantumPhysicist, line 112: I don't think using RpcTime for these variables is right. Time is an absolute value, and timeout is a difference in time. These are, by coincidence, the same type. But they shouldn't be the same type while being used.

In fact, the NonZeroU64 is there for a reason. Now this is broken.

p2p/src/interface/types.rs
```diff
@@ -26,4 +27,13 @@ pub struct ConnectedPeer {
     pub inbound: bool,
 
     pub ban_score: u32,
+
+    /// Time spent waiting for a current ping response
+    pub ping_wait: Option<RpcTime>,
```
TheQuantumPhysicist, line 32: Same... I don't understand why a "wait" is measured in absolute time... this is all over the place.

p2p/src/lib.rs
```diff
@@ -203,7 +214,36 @@ pub async fn make_p2p<S: PeerDbStorage + 'static>(
...
+        ensure!(
+            *chain_config.chain_type() == ChainType::Regtest,
+            P2pError::InvalidConfigurationValue(
+                "P2P encryption can only be disabled on the regtest network".to_owned()
+            )
+        );
+        ensure!(
+            p2p_config.socks5_proxy.is_none(),
+            P2pError::InvalidConfigurationValue(
+                "SOCKS5 proxy support is not implemented for plaintext".to_owned()
```
TheQuantumPhysicist, line 227: plaintext? I'm not sure I understand what this means.

TheQuantumPhysicist: Oh... you mean unencrypted.

Please call it unencrypted.

test/functional/base58/__init__.py
```diff
@@ -0,0 +1,159 @@
+'''Base58 encoding
```
TheQuantumPhysicist, line 1: To be clear, we never use base58... all our data encoding uses bech32.

TheQuantumPhysicist:
Very nice work! I think we're gearing up to have great tests!

Please address my comments, and I'll take another look tomorrow morning.

common/src/primitives/time/rpc.rs
```diff
@@ -0,0 +1,81 @@
...
+type RpcTimeInternal = f64;
+
+/// Newtype for RPC timestamps and durations, expressed as a floating point number of seconds
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct RpcTime(RpcTimeInternal);
+
+#[derive(Debug, thiserror::Error)]
+pub enum RpcTimeError {
+    #[error("Invalid value")]
+    InvalidValue,
```
stanislav-tkach, line 29: Perhaps it would be better to change it to `InvalidValue(String)` for a more detailed error reporting?

pavel-kokolemin:
I fixed some problems with the failed p2p tests.
Most of them were races when time changed, but some parts of the code did not react to that. I added sync points where needed.

There was also an interesting problem with the `discovered_node_channel` test.
There are 3 nodes and they are all expected to discover each other. It so happens that nodes 2 and 3 try to connect to each other at the same time. When both connections are made, both peer managers see that there are duplicate connections from the same peer address (because in `MpscChannelTransport` addresses are single identifiers). And once both connections are closed as duplicates, the peers are no longer connected and the test fails. This only happened with `MpscChannelTransport` because it has no ports, unlike TCP based transports. To fix this, I decided that it would be better to add ports to `MpscChannelTransport` to make it behave more like TCP.

node/src/options.rs
pavel-kokolemin: I guess the name `RpcTime` was not appropriate. I changed it to `RpcDuration`. It is used where the `Duration` type is expected (but where `Duration` itself cannot be used). In any case, it should be better than the currently used `u32` and `u64`.

p2p/src/interface/types.rs
pavel-kokolemin: Should be OK now.

p2p/src/lib.rs
pavel-kokolemin: Sure, updated.

test/functional/base58/__init__.py
pavel-kokolemin: I see. It's only needed for `scalecodec` to work.

common/src/primitives/time/rpc.rs
pavel-kokolemin: Updated, thanks.

common/src/primitives/time/rpc.rs
```diff
@@ -0,0 +1,84 @@
...
+impl From<Duration> for RpcDuration {
+    fn from(value: Duration) -> Self {
+        Self(value.as_secs_f64())
+    }
+}
+
+impl From<RpcDuration> for Duration {
+    fn from(value: RpcDuration) -> Self {
+        // This won't panic because the internal value is guaranteed to be valid
+        Duration::from_secs_f64(value.0)
```
TheQuantumPhysicist, line 60: Why from_secs_f64 instead of from_secs?

We should never have to deal with floating-point numbers.

common/src/primitives/time/rpc.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+impl FromStr for RpcDuration {
+    type Err = RpcTimeError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let value: f64 = s
+            .parse()
+            .map_err(|e| RpcTimeError::InvalidValue(format!("invalid time value: {s}: {e}")))?;
+        let time = Duration::try_from_secs_f64(value)
```
TheQuantumPhysicist, line 45: I see f64 here too.

Please make sure we never use floating-point numbers at all. We only do integer arithmetic.

common/src/primitives/time/rpc.rs
```diff
@@ -0,0 +1,84 @@
...
+
+type RpcDurationInternal = f64;
+
+/// Newtype for time duration, expressed as a floating point number of seconds
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct RpcDuration(RpcDurationInternal);
+
+#[derive(Debug, thiserror::Error)]
+pub enum RpcTimeError {
+    #[error("Invalid value")]
```
TheQuantumPhysicist, line 28: Please print the value in the error. Otherwise in a chain of errors it'll be opaque.

node/src/config_files/chainstate/mod.rs
```diff
@@ -35,7 +34,7 @@ pub struct ChainstateConfigFile {
     ///
     /// The initial block download is finished if the difference between the current time and the
     /// tip time is less than this value.
-    pub max_tip_age: Option<u64>,
+    pub max_tip_age: Option<RpcDuration>,
```
TheQuantumPhysicist, line 37: My problem with this is that ChainstateConfig has nothing to do with Rpc. This doesn't make sense except for the coincidental relationship that both use an integer (and notice the disaster if we're using f64 here, which is another story). The time in a configuration file or configuration input shouldn't be quantified by some arbitrary units unclear. Given that this is crossing the boundaies of the software and type-safety is not possible, the correct way to do this is to use an integer (like u64, like it was), and clarify what units it uses. Then, do the conversion to something that's type-safe that'll deal with this correctly.

node/src/config_files/p2p.rs
```diff
@@ -64,14 +67,14 @@ pub struct P2pConfigFile {
     pub max_inbound_connections: Option<usize>,
     /// The score threshold after which a peer is banned.
     pub ban_threshold: Option<u32>,
-    /// Duration of bans in seconds.
-    pub ban_duration: Option<u64>,
-    /// The outbound connection timeout value in seconds.
-    pub outbound_connection_timeout: Option<NonZeroU64>,
+    /// Duration of bans.
+    pub ban_duration: Option<RpcDuration>,
```
TheQuantumPhysicist, line 71: Same discussion applies here about units of time. No units of time should be done in Rpc units. We use seconds or milliseconds (whatever is correct and consistent), and then convert to something sensible based on types.

This is the evil part of derive in Rust... now we have floating point numbers in our configuration without even noticing.

TheQuantumPhysicist (commented):
Here's what I don't understand in this PR. We have a module that's specific to time. The module provides the units we need for time and time processing. Suddenly, we have an extra treatment of time, extra type, and zero processing.

Isn't better to define two ditinct types for time, like `Seconds` and `Milliseconds` in the time module, where we define how these get converted from numbers, and then use these in configuration, instead of this arbitrary RpcDuration unit?

common/src/primitives/time/rpc.rs
pavel-kokolemin: `RpcDuration` removed.

common/src/primitives/time/rpc.rs
pavel-kokolemin: OK, sure.

common/src/primitives/time/rpc.rs
pavel-kokolemin: Removed.

node/src/config_files/chainstate/mod.rs
pavel-kokolemin: Sure, reverted.

pavel-kokolemin:
> Here's what I don't understand in this PR...

`RpcDuration` removed.

node/src/options.rs
```diff
@@ -60,7 +60,7 @@ pub struct RunOptions {
     #[clap(long)]
     pub node_type: Option<NodeTypeConfigFile>,
 
-    /// Mock time.
+    /// Mock time used to initialize the node time at startup, in seconds.
```
TheQuantumPhysicist, line 63: (valid only for regtest)

stanislav-tkach (approved):

PR #751, opened 2023-03-10
Add normal derivations to Secp256k1 keys

erasmospunk:


crypto/src/key/hdkd/child_number.rs
```diff
@@ -29,35 +30,78 @@ pub struct ChildNumber(DerivationType);
...
+
+    pub fn plus_one(&self) -> Result<Self, DerivationError> {
+        match self.0 {
+            DerivationType::Normal(i) => Ok(ChildNumber(DerivationType::Normal(i.plus_one()?))),
+            DerivationType::Hardened(i) => Ok(ChildNumber(DerivationType::Hardened(i.plus_one()?))),
+        }
+    }
+}
+
+impl Encode for ChildNumber {
```
TheQuantumPhysicist, line 91: Please write tests for Encoding/Decoding that are specific to ChildNumber in this module. Also, with ChildNumber becoming more complex, we should test all its functionality. Please add unit tests for the life cycle of ChildNumber.

TheQuantumPhysicist: To be clear, always in encoding/decoding tests, we test:

1. Back and forth encoding with random values
2. Decoding from selected values, to ensure that the underlying format is stable

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -116,9 +121,12 @@ impl Derivable for Secp256k1ExtendedPrivateKey {
...
         let secp_key = self.private_key.data;
 
-        // We only support hard derivations
-        mac.update(&[0u8]);
-        mac.update(&secp_key[..]);
+        if num.is_hardened() {
+            mac.update(&[0u8]);
+            mac.update(&secp_key[..]);
+        } else {
+            mac.update(&secp_key.public_key(SECP256K1).serialize());
```
TheQuantumPhysicist, line 128: Please add a comment on what's going on here.

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -135,15 +143,28 @@ impl Derivable for Secp256k1ExtendedPrivateKey {
...
+    fn get_derivation_path(&self) -> DerivationPath {
+        self.derivation_path.clone()
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode)]
 pub struct Secp256k1ExtendedPublicKey {
+    /// The derivation path of leading to this key
+    pub derivation_path: DerivationPath,
     /// Chain code
```
TheQuantumPhysicist, line 168: Useless comment. Please elaborate.

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -135,15 +143,28 @@ impl Derivable for Secp256k1ExtendedPrivateKey {
...
     pub chain_code: ChainCode,
     /// Public key
```
TheQuantumPhysicist, line 170: Also useless comment

crypto/src/key/secp256k1/mod.rs
```diff
@@ -306,19 +310,19 @@ mod test {
...
-            ("dff1d77f2a671c5f36183726db2341be58feae1da2deced843240f7b502ba659", "243f6a8885a308d313198a2e03707344a4093822299f31d0082efa98ec4e6c89", "4a298dacae57395a15d0795ddbfd1dcb564da82b0f269bc70a74f8220429ba1d69e89b4c5564d00349106b8497785dd7d1d713a8ae82b32fa79d5f7fc407d39b", false),
-            ("dff1d77f2a671c5f36183726db2341be58feae1da2deced843240f7b502ba659", "243f6a8885a308d313198a2e03707344a4093822299f31d0082efa98ec4e6c89", "fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f69e89b4c5564d00349106b8497785dd7d1d713a8ae82b32fa79d5f7fc407d39b", false),
-            ("dff1d77f2a671c5f36183726db2341be58feae1da2deced843240f7b502ba659", "243f6a8885a308d313198a2e03707344a4093822299f31d0082efa98ec4e6c89", "6cff5c3ba86c69ea4b7376f31a9bcb4f74c1976089b2d9963da2e5543e177769fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141", false),
+            ("02f9308a019258c31049344f85f89d5229b531c845836f99b08601f113bce036f9", "0000000000000000000000000000000000000000000000000000000000000000", "e907831f80848d1069a5371b402410364bdf1c5f8307b0084c55f1ce2dca821525f66a4a85ea8b71e482a74f382d2ce5ebeee8fdb2172f477df4900d310536c0", true),
```
TheQuantumPhysicist, line 313: You can use rstest for these test vectors. See here: https://github.com/mintlayer/mintlayer-core/blob/aa57f591b2179b33f88c66956e558a4f908f60fd/common/src/primitives/merkle/proof/multi/tests.rs#L448

TheQuantumPhysicist:
Looking good! Please address the comments and I'll look again. It's Friday night and I don't wanna approve something this late :smile:

Please use rstest for all these loops with test vectors... it's cleaner and when there's an error, easier to track.

crypto/src/key/hdkd/derivable.rs
```diff
@@ -25,21 +25,44 @@ pub enum DerivationError {
...
+    CannotDeriveHardenedKeyFromPublicKey(ChildNumber),
 }
 
 pub trait Derivable: Sized {
-    /// Derive a child private key given a derivation path
+    /// Derive a child private key given a derivation path. The derivation path must include
+    /// the path of this key. For example:
+    /// - If this (self) key has the path m/1/2/3
+    /// - Then the requested path should be m/1/2/3/<rest/of/the/path>
     fn derive_path(self, path: &DerivationPath) -> Result<Self, DerivationError> {
```
iljakuklic, line 45: Given the comment, maybe this could be called `derive_absolute_path`. To disambiguate from a hypothetical function that appends a bunch of components to the existing path.

iljakuklic (commented):
Looks fine to me. Some recommendations/comments:

crypto/src/key/hdkd/child_number.rs
TheQuantumPhysicist: I'm sorry, I still don't see these tests.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -16,29 +16,71 @@
...
-    fn from(path: Vec<ChildNumber>) -> Self {
-        DerivationPath(path)
+impl Decode for DerivationPath {
+    fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+        let size = input.read_byte()?;
+        let mut path = Vec::with_capacity(size as usize);
+
+        for _ in 0..size {
+            path.push(<ChildNumber>::decode(input)?);
+        }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: If we're to use the default encoding underneath, I actually would wonder why we don't use `WrapperTypeEncode` and `WrapperTypeDecode`.

crypto/src/key/hdkd/child_number.rs
```diff
@@ -98,6 +149,68 @@ impl fmt::Display for ChildNumber {
...
+    #[trace]
+    #[case(0, false)]
+    #[case(1, false)]
+    #[case(1234567, false)]
+    #[case(u32::MAX & (!0x80000000 - 1), false)]
+    #[case(u32::MAX & !0x80000000, false)]
+    #[case(u32::MAX & (!0x80000000 + 1), true)]
+    #[case(u32::MAX - 1, true)]
+    #[case(u32::MAX, true)]
+    fn create_child_number(#[case] encoded_num: u32, #[case] is_hardened: bool) {
```
TheQuantumPhysicist, line 187: Please another test where a random value is encoded, decoded, and then we ensure that it's the same. The random value should be treated in the same way we do our property-based testing (like we discussed the other day).

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -157,12 +178,56 @@ impl Secp256k1ExtendedPublicKey {
...
+            .public_key
+            .pubkey_data
+            .add_exp_tweak(SECP256K1, &key_part.into())
+            .map_err(|_| DerivationError::KeyDerivationError)?
+            .into();
+
+        let derivation_path = {
+            let mut child_path = self.get_derivation_path().into_vec();
+            child_path.push(num);
+            child_path.try_into()?
```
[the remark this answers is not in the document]
erasmospunk: Originally I wanted `get_derivation_path` to return a reference but had issues because of `MLRistrettoPrivateKey` that we will remove soon. So I did some work that makes this method return a reference.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -193,40 +237,79 @@ mod tests {
...
+        // Assert encoding
+        assert_encoded_eq(&path, path_encoded_hex);
+        // Assert decoding
+        let decoded_path: DerivationPath = decode_from_hex(path_encoded_hex);
+        assert_eq!(path_string, decoded_path.to_string());
+        assert_eq!(path, decoded_path);
+    }
+
+    #[test]
+    fn len_and_is_empty_() {
```
TheQuantumPhysicist, line 309: Why extra `_`?

crypto/src/key/hdkd/derivation_path.rs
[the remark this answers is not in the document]
iljakuklic: > This custom encoding is based on the [SLIP32 specification](https://github.com/satoshilabs/slips/blob/master/slip-0032.md). I recommend sticking with this one so that third party tools can easily support the Mintlayer exported keys.

That sounds sensible to me.

crypto/src/key/secp256k1/mod.rs
```diff
@@ -303,30 +304,30 @@ mod test {
...
-
-        for (pk, msg_hash, sig, is_valid) in test_vec {
-            let pk = Secp256k1PublicKey::from_bytes(&hex::decode(pk).unwrap()).unwrap();
-            let sig =
-                secp256k1::schnorr::Signature::from_slice(&hex::decode(sig).unwrap()).unwrap();
-            let msg_hash = secp256k1::Message::from_slice(&hex::decode(msg_hash).unwrap()).unwrap();
-            assert_eq!(pk.verify_message_hashed(&sig, &msg_hash), is_valid);
-        }
+    #[rstest]
+    #[trace]
```
TheQuantumPhysicist, line 308: You don't need trace here because this isn't random. And this trace actually applies only to the first case anyway. Tracing every case needs trace on every line of every case.

crypto/src/key/secp256k1/mod.rs
```diff
@@ -219,46 +225,41 @@ mod test {
...
-        ];
-
-        for (sk_hex, pk_hex) in sk_pk_hex {
-            let sk = Secp256k1PrivateKey::from_bytes(&hex::decode(sk_hex).unwrap()).unwrap();
-            let pk = Secp256k1PublicKey::from_bytes(&hex::decode(pk_hex).unwrap()).unwrap();
-
-            let secp = secp256k1::Secp256k1::new();
-            assert_eq!(pk.as_native(), &sk.as_native().x_only_public_key(&secp).0);
+    #[rstest]
+    #[trace]
```
TheQuantumPhysicist, line 229: Same thing about trace

PR #753, opened 2023-03-10
Fix: Check PoW target in block

TheQuantumPhysicist:
I just noticed this. There's no place where this check is done. Basically a block producer can put any target they want.

This glosses on the issue that we discussed before, on whether we want to remove the target altogether. But this doesn't look nice for PoW, since block PoW should allow being checked independently, without the history.

### DIFF
--- a/consensus/src/pow/work.rs
+++ b/consensus/src/pow/work.rs
@@ -49,11 +49,18 @@ pub fn check_proof_of_work(
 pub fn check_pow_consensus<H: BlockIndexHandle>(
     chain_config: &ChainConfig,
     header: &BlockHeader,
+    block_pow_data: &PoWData,
     pow_status: &PoWStatus,
     block_index_handle: &H,
 ) -> Result<(), ConsensusPoWError> {
     let work_required =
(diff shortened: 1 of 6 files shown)


consensus/src/pow/work.rs
```diff
@@ -49,11 +49,18 @@ pub fn check_proof_of_work(
...
     header: &BlockHeader,
+    block_pow_data: &PoWData,
     pow_status: &PoWStatus,
     block_index_handle: &H,
 ) -> Result<(), ConsensusPoWError> {
     let work_required =
         calculate_work_required(chain_config, header, pow_status, block_index_handle)?;
+
+    // TODO: add test for a block with invalid target
+    if work_required != block_pow_data.bits() {
```
azarovh, line 60: Was also curious why do have to store target in the header. Couldn't it be calculated on the fly whenever required?

azarovh (approved):

consensus/src/pow/work.rs
TheQuantumPhysicist: See the comment on the PR

azarovh: sorry, missed the description

PR #754, opened 2023-03-10
 Relay transactions

stanislav-tkach:
Closes https://github.com/mintlayer/mintlayer-core/issues/735.


mempool/src/pool/mod.rs
```diff
@@ -636,8 +636,9 @@ where
     pub async fn add_transaction(&mut self, tx: SignedTransaction) -> Result<(), Error> {
         let conflicts = self.validate_transaction(&tx).await?;
         self.store.drop_conflicts(conflicts);
-        self.finalize_tx(tx).await?;
+        self.finalize_tx(tx.clone()).await?;
         self.store.assert_valid();
+        self.events_controller.broadcast(MempoolEvent::NewTransaction(tx));
```
TheQuantumPhysicist, line 641: This is wrong.

There seems to be a misunderstanding on why a NewTransaction event would have to exist, even though, speaking from a design point of you, it's the wrong thing to do. Transactions never emanate from the mempool. They end up there. They either come from peers or they come from rpc. When we submit a transaction through RPC, we are required to verify the validity of the transaction as if we're receiving it from a peer (to avoid being banned by peers) AND we are required to add it to the mempool (so that it can become part of a block) AND we're required to broadcast it to other peers (so that if we fail in adding it to a block, someone else will). Given that the first two requirements are done by the mempool, one would be tempted to think that the mempool should also do the broadcasting, but this is wrong. In fact, seeing this, I'm betting that the commig that has "deadlock" in the label was because of this.

So, two separate events have to happen in the RPC call:
1. Transaction should be added to the mempool (where it's also verified against the best chain)
2. Broadcast the transaction to the network.

It seems to me it's more reasonable to put both of these in the `submit_transaction` function of rpc, instead of handing transactions to the mempool and then get the mempool to broadcast them again.

mempool/src/pool/mod.rs
```diff
@@ -636,7 +636,7 @@ where
     pub async fn add_transaction(&mut self, tx: SignedTransaction) -> Result<(), Error> {
         let conflicts = self.validate_transaction(&tx).await?;
         self.store.drop_conflicts(conflicts);
-        self.finalize_tx(tx).await?;
+        self.finalize_tx(tx.clone()).await?;
```
TheQuantumPhysicist, line 639: I don't think this is necessary

mempool/src/rpc.rs
```diff
@@ -21,6 +21,7 @@ trait MempoolRpc {
     fn dummy(&self) -> rpc::Result<String>;
 }
 
+#[async_trait::async_trait]
```
TheQuantumPhysicist, line 24: Not necessary

p2p/src/sync/tests/tx_announcement.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[tokio::test]
+#[should_panic = "Received a message from unknown peer"]
+async fn nonexistent_peer() {
+    let mut handle = SyncManagerHandle::builder().build().await;
+
+    let peer = PeerId::new();
+
+    let tx = Transaction::new(0x00, vec![], vec![], 0x01).unwrap();
+    let tx = SignedTransaction::new(tx.clone(), vec![]).unwrap();
```
TheQuantumPhysicist, line 41: No need to clone here or in any of the tests below.

I don't know why clippy isn't catching this.

p2p/src/rpc.rs
```diff
@@ -48,44 +51,54 @@ trait P2pRpc {
     /// Existing connection to the peer is not closed.
     #[method(name = "remove_reserved_node")]
     async fn remove_reserved_node(&self, addr: String) -> rpc::Result<()>;
+
+    /// Submits a transaction to mempool.
```
TheQuantumPhysicist, line 55: , and if it is valid, broadcasts it to the network.

p2p/src/sync/peer.rs
```diff
@@ -398,8 +410,17 @@ where
         };
 
         match error {
+            // Due to the fact that p2p is split into several tasks, it is possible to send a
+            // request/response after a peer is disconnected, but before receiving the disconnect
+            // event. Therefore this error can be safely ignored.
+            P2pError::PeerError(PeerError::PeerDoesntExist) => Ok(()),
+            P2pError::MempoolError(
```
TheQuantumPhysicist, line 417: There has to be a score associated with transactions from the mempool. If this doesn't exist not and it's not feasible, please create a TODO and an issue.

TheQuantumPhysicist (approved):
Looking good. The only big issue I see is the handling of errors from the mempool. I added a comment about it.

Please get Pavel to review this too.

mempool/src/pool/mod.rs
stanislav-tkach: Sure, it is an artifact from the previous implementation attempt. I have even removed it locally, but forgot to push the final commit before marking the pull request as ready. Sorry, fixed.

pavel-kokolemin (approved):

PR #755, opened 2023-03-11
Enable RPC for blockprod, along with ability to create blocks

alfiedotwtf:
This new RPC call allows the caller to create blocks. Currently it only accepts signed transactions, but https://github.com/mintlayer/mintlayer-core/issues/756 will allow accepting TXIDs that refer to mempool transactions.


alfiedotwtf:
Once PoW is done, tests can then be added to test this RPC endpoint along with testing the PoW code.

TheQuantumPhysicist:
> Once PoW is done, tests can then be added to test this RPC endpoint along with testing the PoW code.

PoW works now. No issues with it.

alfiedotwtf:
Clippy didn't like this, even though I added an exemption. I think I'll split this fix so that it's a smooth and simple merge before this one

alfiedotwtf:
> > Once PoW is done, tests can then be added to test this RPC endpoint along with testing the PoW code.
> 
> PoW works now. No issues with it.

Sorry, I meant once my PoW changes to the block maker were done.

> tests can then be added

Included in this list, I'll update the functional tests to use the new `generate_block` RPC call.

blockprod/src/detail/block_maker.rs
```diff
@@ -94,17 +97,28 @@ impl BlockMaker {
...
     ) -> Result<Block, BlockProductionError> {
-        // TODO: this isn't efficient. We have to create the header first, then see if it obeys consensus rules, then construct the full block
+        // TODO: Instead of the following static value, look at
+        // self.chain_config to get the correct consensus data
+        let consensus_data = ConsensusData::None;
+
+        // TODO: instead of the following static value, look at
+        // self.chain_config for the current block reward, then send
+        // it to self.reward_destination
+        let block_reward = BlockReward::new(vec![]);
```
TheQuantumPhysicist, line 110: This isn't really too hard to do, but very beneficial.

blockprod/src/detail/builder.rs
```diff
@@ -177,6 +182,7 @@ impl PerpetualBlockBuilder {
     }
 
     /// Subscribe to events from the mempool
+
```
TheQuantumPhysicist, line 185: Nitpick

blockprod/src/detail/mod.rs
```diff
@@ -73,6 +83,57 @@ impl BlockProductionInterface for BlockProduction {
...
+
+    async fn generate_block(
+        &self,
+        reward_destination: Destination,
+        signed_transactions: Vec<SignedTransaction>,
+    ) -> Result<Id<Block>, BlockProductionError> {
+        let (best_block_id, best_block_height, accumulator) = self
+            .chainstate_handle
+            .call(|this| {
+                // Stay with DefaultTxaccumulator as we are being
```
TheQuantumPhysicist, line 95: DefaultTxAccumulator

blockprod/src/detail/mod.rs
```diff
@@ -73,6 +83,57 @@ impl BlockProductionInterface for BlockProduction {
...
+            reward_destination,
+            best_block_id,
+            best_block_height,
+            dummy_rx,
+        );
+
+        // TODO: Stop any current block makers before making this new
+        // block, and then restart if enabled when done. Otherwise we
+        // will be doing work for no benefit.
+        Ok(block_maker.make_block(best_block_id, &accumulator)?.get_id())
```
TheQuantumPhysicist, line 135: I think the better way to handle this is to change the argument of make_block to take an enum, which can either be a Vec of transactions or an accumulator. The way it's done now is kind of twisted for no good reason; especially that this function is made for testing, so we need behavior that is decoupled from the behavior of the accumulator.

blockprod/src/detail/mod.rs
```diff
@@ -73,6 +83,57 @@ impl BlockProductionInterface for BlockProduction {
...
+            self.chainstate_handle.clone(),
+            self.mempool_handle.clone(),
+            self.time_getter.clone(),
+            reward_destination,
+            best_block_id,
+            best_block_height,
+            dummy_rx,
+        );
+
+        // TODO: Stop any current block makers before making this new
```
TheQuantumPhysicist, line 132: I wouldn't put it like this, because the operation of a block maker is more complex than this. Usually, there shouldn't be block makers running if you call generate.

blockprod/src/rpc.rs
```diff
@@ -38,6 +54,37 @@ impl BlockProductionRpcServer for super::BlockProductionHandle {
...
+    ) -> rpc::Result<Id<Block>> {
+        let reward_destination_data =
+            hex::decode(reward_destination_hex).map_err(rpc::Error::to_call_error)?;
+
+        let reward_destination = Destination::decode(&mut &reward_destination_data[..])
+            .map_err(rpc::Error::to_call_error)?;
+
+        let mut signed_transactions = vec![];
+
+        for transaction_hex in transactions_hex {
```
TheQuantumPhysicist, line 71: Please use map, then collect to `Result<Transaction, Error>` then use the try operator `?`.

TheQuantumPhysicist (approved):

alfiedotwtf:
Note: Things left to do:

- Reward destination for both RPC and local blocks needs to be taken into account
- Proper error handling (ignore the current error handling as it's just a filler at the moment or non-existant)
- unit tests
- TXID parameter input to RPC (#756 )
- A handful of TODOs in the commit changes
- Moving PoW mining to it's own executor
- User-modifiable configuration for PoW mining

blockprod/src/detail/block_maker.rs
```diff
@@ -92,21 +93,30 @@ impl BlockMaker {
...
-            BlockReward::new(vec![]), // TODO: define consensus and rewards through NetworkUpgrades
-        )?;
-        Ok(block)
+        self.chainstate_handle
+            .call({
+                let current_tip_id = self.current_tip_id;
+                let current_tip_height = self.current_tip_height;
+                let time_getter = self.time_getter.clone();
+                move |this| {
+                    this.generate_block(
```
TheQuantumPhysicist, line 106: Why are we generating the block in the chainstate thread? Isn't this the job of the block maker thread?

alfiedotwtf: After talking tonight with Sam, I'll be moving the generation code back into `blockprod` and will try to refactor out what's needed for consensus calculations via chainstate queries rather than trying to directly access the index *within* chainstate.

blockprod/src/detail/mod.rs
alfiedotwtf: Now that `generate_block()` is async, Accumulators can't be passed into the function because they are not `Sync`. Also, we're now passing in by value.

PR #757, opened 2023-03-11
Appease Clippy's default 7 max function parameters limit

alfiedotwtf:
The current value seems like an arbitrary limit, and I need it increased for a blockprod PR to pass clippy.

### DIFF
--- a/do_checks.sh
+++ b/do_checks.sh
@@ -3,5 +3,5 @@ cargo fmt --check &&
 PYTHON=$(which python || which python3)
 $PYTHON .github/scripts/codecheck.py &&
 cargo deny check --hide-inclusion-graph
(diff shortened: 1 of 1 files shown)


muursh (approved):

TheQuantumPhysicist:
I wouldn't make this a rule. This isn't the right way to go. Usually this is an exception that we do when we have no choice. It's not something we want to allow by default.

Look at it this way: When a clippy exception is made, it triggers a discussion on whether there's a better way to fix this problem.

alfiedotwtf:
Yep, no problem. I figured that by making this PR, it would start that discussion :)

TheQuantumPhysicist:
You can add that exception for the place where you have that problem. There are directives for single functions.

alfiedotwtf:
Closing this as I'll inline it instead.

PR #759, opened 2023-03-13
Handle SIGTERM on *nix

pavel-kokolemin:


iljakuklic (approved):

muursh (approved):

subsystem/src/manager.rs
```diff
@@ -226,15 +226,31 @@ impl Manager {
...
             "ctrl-c",
             |mut call_rq: CallRequest<()>, mut shutdown_rq| async move {
+                // Gracefully handle SIGTERM on *nix
+                #[cfg(unix)]
+                let term_signal = async move {
+                    let mut signal =
+                        tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
+                            .expect("Signal handler failed unexpectedly");
+                    signal.recv().await;
+                };
```
iljakuklic, line 236: I think the `async` scope and the final `await` cancel out, don't they?

pavel-kokolemin: Right, but this does not work:
```
let term_signal = {
    let mut signal =
        tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())
            .expect("Signal handler failed unexpectedly");
    signal.recv()
};
```

The error:
```
`signal` does not live long enough
borrowed value does not live long enough
```

pavel-kokolemin: And I can't move `tokio::signal::unix` outside of `#[cfg(unix)]`-only block.

TheQuantumPhysicist (approved):

subsystem/src/manager.rs
iljakuklic: ok, thanks for explanation

PR #761, opened 2023-03-14
Add remaining checks to p2p ping (functional tests)

pavel-kokolemin:
Fix for debug output grep was trivial - just use file path created by `NamedTemporaryFile`:
```
    def start(self, extra_args=None, *, cwd=None, stdout=None, stderr=None, **kwargs):
...
        if stderr is None:
            stderr = tempfile.NamedTemporaryFile(dir=self.stderr_dir, delete=False)
            # Save the temporary file name to be able grep the log later
            self.stderr_file = stderr.name
```

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -788,8 +788,8 @@ where
         );
     }
 
-    fn handle_ping_response(&mut self, peer: PeerId, nonce: u64) {
-        if let Some(peer) = self.peers.get_mut(&peer) {
+    fn handle_ping_response(&mut self, peer_id: PeerId, nonce: u64) {
+        if let Some(peer) = self.peers.get_mut(&peer_id) {
             if let Some(sent_ping) = peer.sent_ping.as_mut() {
                 if sent_ping.nonce == nonce {
                     // Correct reply received, clear pending request and update ping times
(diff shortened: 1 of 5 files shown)


p2p/src/peer_manager/mod.rs
```diff
@@ -808,14 +808,10 @@ where
                     peer.ping_last = Some(ping_time_last);
                     peer.ping_min = Some(ping_time_min);
                 } else {
-                    log::debug!(
-                        "wrong nonce in ping response: {}, expected: {}",
-                        nonce,
-                        sent_ping.nonce
-                    );
+                    log::debug!("wrong nonce in ping response from peer {}", peer_id,);
```
TheQuantumPhysicist, line 811: I suggest keeping the nonce, and just moving the nonce value after the text you're searching for. This way the assert passes in the tests and you can still have the extra information in the logs.

TheQuantumPhysicist (approved):

pavel-kokolemin:
> My only wish is that we introduce some smart search in the logging such that we don't search the whole log file, but we can set a "starting from here" model so that things that are emitted in the past are not found multiple times.

This is already implemented in the bitcoin testing framework.
When you call `assert_debug_log` it first remembers the current log file size, then executes what was requested, and then searches the log file for new text.

This is why `assert_debug_log` is called using `with`:
```
        with self.nodes[0].assert_debug_log(['ping check: dead peer detected: 1']):
            self.mock_forward(TIMEOUT_INTERVAL)
```

stanislav-tkach (approved):

PR #767, opened 2023-03-16
Add token bucket based rate limiter

pavel-kokolemin:
This should prevent malicious peers from overloading the network with bogus addresses.

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -56,7 +56,7 @@ use crate::{
         peer_address::{PeerAddress, PeerAddressIp4, PeerAddressIp6},
         peer_id::PeerId,
     },
-    utils::oneshot_nofail,
+    utils::{oneshot_nofail, rate_limiter::RateLimiter},
 };
 
 use self::{
@@ -75,6 +75,14 @@ const PEER_MGR_HEARTBEAT_INTERVAL_MAX: Duration = Duration::from_secs(30);
 
 /// How many addresses are allowed to be sent
 const MAX_ADDRESS_COUNT: usize = 1000;
+
+/// The maximum rate of address announcements the node will process from a peer (value as in Bitcoin Core).
+const MAX_ADDR_RATE_PER_SECOND: f64 = 0.1;
+/// Bucket size used to rate limit address announcements from a peer.
+/// Use a non-zero value to allow peers to send their own addresses immediately after connection.
+const ADDR_RATE_INITIAL_SIZE: u32 = 10;
+/// Bucket size used to rate limit address announcements from a peer.
+const ADDR_RATE_BUCKET_SIZE: u32 = 10;
 
 /// To how many peers resend received address
 const PEER_ADDRESS_RESEND_COUNT: usize = 2;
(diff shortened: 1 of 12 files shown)


p2p/src/peer_manager/mod.rs
```diff
@@ -698,17 +714,20 @@ where
...
-            self.peers
+            let peer = self
+                .peers
                 .get_mut(&peer)
-                .expect("peer sending AnnounceAddrRequest must be known")
-                .announced_addresses
-                .insert(address.clone());
+                .expect("peer sending AnnounceAddrRequest must be known");
+            if !peer.address_rate_limiter.accept(self.time_getter.get_time()) {
+                log::debug!("Address announcment is dropped because of rate limiting");
```
TheQuantumPhysicist, line 726: announcement

p2p/src/utils/rate_limiter/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// Check if the new request is within the allowed rate at the current time (updating the state)
+    ///
+    /// # Arguments
+    /// `now` - Current time
+    pub fn accept(&mut self, now: Duration) -> bool {
+        let seconds = now.checked_sub(self.last_time).unwrap_or_default().as_secs_f64();
+        self.last_time = now;
+        self.tokens = f64::min(self.tokens + self.rate * seconds, self.bucket.into());
+        // Use a value slightly less than 1.0 to account for f64 rounding errors (makes unit testing easier)
```
TheQuantumPhysicist, line 59: And people wonder why I hate floating point numbers.

pavel-kokolemin: I have tried using integral arithmetic here, but the code gets pretty ugly.

stanislav-tkach (approved):

p2p/src/utils/rate_limiter/mod.rs
TheQuantumPhysicist: It's OK. As long as accuracy and reproducibility isn't a requirement. Here, it's OK. In consensus, hell no!

p2p/src/utils/rate_limiter/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+use std::time::Duration;
+
+/// Token bucket based rate limiter
+///
+/// See https://en.wikipedia.org/wiki/Token_bucket
+#[derive(Debug)]
+pub struct RateLimiter {
+    rate: f64,
+    tokens: f64,
```
TheQuantumPhysicist, line 26: Why are tokens a floating point number? It's understandable that the rate is a floating point number. But the tokens? Can't we just scale up to make the minimum unit == 1?

TheQuantumPhysicist (approved):

p2p/src/utils/rate_limiter/mod.rs
pavel-kokolemin: I'm not sure how it can be done. The problem is that both `seconds` and `rate` here are floats:
```
self.tokens = f64::min(self.tokens + self.rate * seconds, self.bucket.into());
```
And so `tokens` must be float to be accurate.
It's implemented same way in Bitcoin Core: https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L360

TheQuantumPhysicist: I don't know. It sounds appropriate to be an integer. But it's OK.

Just please double-check NaN cases and similar are safe before merging this.

pavel-kokolemin: It should be OK because there is only multiplication and all values can be trusted.

muursh (approved):

PR #769, opened 2023-03-17
Try to fix the problem with stuck sync

pavel-kokolemin:
Functional tests started to fail after some recent changes.
Here is one such failed run: https://github.com/mintlayer/mintlayer-core/actions/runs/4442342875
The logs show this error:
```
Adjusting the '2' peer score by 20: ProtocolError(UnexpectedMessage("Peer requested already known block"))
```

It looks like the test fails this way:
1. Two new blocks submitted on Node0.
2. Node1 connects and downloads both blocks in initial block download.
3. Node0 sends new block announcement to all connected peers (this may happen with some delay because block announcements are sent from a different even loop).
4. Node1 receives this announcement and tries to download this already known block.
5. Node0 stops syncing because of the "Peer requested already known block" error.

To fix this, I added a check to drop block announcements for already known blocks.

### DIFF
--- a/p2p/src/error.rs
+++ b/p2p/src/error.rs
@@ -14,7 +14,10 @@
 // limitations under the License.
 
 use chainstate::ban_score::BanScore;
-use common::primitives::semver::SemVer;
+use common::{
+    chain::Block,
+    primitives::{semver::SemVer, Id},
+};
 use thiserror::Error;
 
 /// Errors related to invalid data/peer information that results in connection getting closed
@@ -25,10 +28,6 @@ pub enum ProtocolError {
     DifferentNetwork([u8; 4], [u8; 4]),
     #[error("Peer has an unsupported version. Our version {0}, their version {1}")]
     InvalidVersion(SemVer, SemVer),
-    // TODO: This error is very generic and probably should be replaced with several different ones,
-    // because it has a ban score of 100 and in many cases it is too harsh.
-    #[error("Peer sent an invalid message")]
-    InvalidMessage(&'static str),
     #[error("Peer is unresponsive")]
     Unresponsive,
     #[error("Locator size ({0}) exceeds allowed limit ({1})")]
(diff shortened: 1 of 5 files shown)


stanislav-tkach (approved):

TheQuantumPhysicist (approved):
Besides the little things I mentioned, looking good.

muursh (approved):
Works for me

p2p/src/sync/peer.rs
```diff
@@ -360,13 +369,27 @@ where
...
 
         if !self.requested_blocks.is_empty() {
             // We will download this block as part of syncing anyway.
             return Ok(());
         }
 
+        // Do not request the block if it is already known
+        if self
+            .chainstate_handle
+            .call(move |c| c.get_gen_block_index(&block_id.into()))
```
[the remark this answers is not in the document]
pavel-kokolemin: Right, thanks.

p2p/src/error.rs
```diff
@@ -28,7 +28,7 @@ pub enum ProtocolError {
     // TODO: This error is very generic and probably should be replaced with several different ones,
     // because it has a ban score of 100 and in many cases it is too harsh.
     #[error("Peer sent an invalid message")]
-    InvalidMessage,
+    InvalidMessage(&'static str),
```
TheQuantumPhysicist, line 31: Why not just a String?

pavel-kokolemin: Sure, updated.

p2p/src/error.rs
```diff
@@ -188,14 +195,17 @@ impl BanScore for ProtocolError {
...
             ProtocolError::LocatorSizeExceeded(_, _) => 20,
             ProtocolError::BlocksRequestLimitExceeded(_, _) => 20,
             ProtocolError::HeadersLimitExceeded(_, _) => 20,
-            ProtocolError::UnknownBlockRequested => 20,
+            ProtocolError::UnknownBlockRequested(_) => 20,
+            ProtocolError::DuplicatedBlockRequest(_) => 20,
             ProtocolError::DisconnectedHeaders => 20,
             ProtocolError::UnexpectedMessage(_) => 20,
+            ProtocolError::ZeroBlocksInRequest => 20,
+            ProtocolError::HandshakeExpected => 20,
```
TheQuantumPhysicist, line 207: Is 20 intended in all of these? It sounds OK for ZeroBlocks... but the others, the handshake thing sounds bad. Also limit breaking sounds bad too. Maybe they deserve 100?

TheQuantumPhysicist (approved):

PR #771, opened 2023-03-17
Random (IDE lints) fixes

stanislav-tkach:
After seeing the https://github.com/mintlayer/mintlayer-core/pull/768 pull request, I have decided to try the analysis embedded in my IDE (CLion with intellij-rust plugin). Surprisingly it hasn't found many typos or grammar errors, but it gives some helpful lints instead. For example "using assert_eq/assert_ne" , "redundant semicolon" or "unnecessary qualified path". The most important thing that it found was the unused [history_iteration.rs](https://github.com/mintlayer/mintlayer-core/blob/master/chainstate/test-suite/src/tests/history_iteration.rs) file.

### DIFF
--- a/common/src/chain/config/builder.rs
+++ b/common/src/chain/config/builder.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::emission_schedule::{self, *};
+use super::emission_schedule::*;
 use super::{create_mainnet_genesis, create_unit_test_genesis, ChainConfig, ChainType};
 
 use crate::chain::{
@@ -65,8 +65,8 @@ impl ChainType {
 #[derive(Clone)]
 enum EmissionScheduleInit {
     Mainnet,
-    Table(emission_schedule::EmissionScheduleTabular),
-    Fn(std::sync::Arc<emission_schedule::EmissionScheduleFn>),
+    Table(EmissionScheduleTabular),
+    Fn(Arc<EmissionScheduleFn>),
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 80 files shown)


utils/src/blockuntilzero.rs
```diff
@@ -56,7 +56,6 @@ where
         self.value.load(Ordering::Acquire)
     }
 
-    #[must_use = "CountTracker is useless without holding its object"]
```
stanislav-tkach, line 59: This isn't an error: the `CountTracker` type already has the `must_use` attribute, so this is redundant.

utils/src/counttracker.rs
```diff
@@ -33,7 +33,6 @@ where
     <T as Atomic>::Type: One,
     <T as Atomic>::Type: Zero,
 {
-    #[must_use = "CountTracker is useless without holding its object"]
```
stanislav-tkach, line 36: Same here.

common/src/chain/config/builder.rs
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::emission_schedule::{self, *};
+use super::emission_schedule::*;
```
TheQuantumPhysicist, line 16: I'm a big opponent of this kind of imports... very confusing and makes it hard to read files.

TheQuantumPhysicist (approved):

common/src/chain/config/builder.rs
stanislav-tkach: I agree, fixed.

pavel-kokolemin (approved):

muursh (approved):

PR #773, opened 2023-03-20
Mine PoS blocks in tests

azarovh:
This is a fix for the problem with PoS unit tests discovered by Sam.
I added a mining function but only for `pos_processing tests`.  Not sure that it's useful elsewhere at this point.

### DIFF
--- a/chainstate/types/src/pos_randomness.rs
+++ b/chainstate/types/src/pos_randomness.rs
@@ -15,7 +15,7 @@
 
 use common::{
     chain::{
-        block::{consensus_data::PoSData, BlockHeader},
+        block::{consensus_data::PoSData, timestamp::BlockTimestamp},
         config::EpochIndex,
         Block, ChainConfig,
     },
@@ -47,7 +47,7 @@ impl PoSRandomness {
 
     pub fn from_block(
         epoch_index: EpochIndex,
-        header: &BlockHeader,
+        block_timestamp: BlockTimestamp,
         seal_randomness: &PoSRandomness,
         pos_data: &PoSData,
         vrf_pub_key: &VRFPublicKey,
@@ -57,7 +57,7 @@ impl PoSRandomness {
             &seal_randomness.value(),
             pos_data.vrf_data(),
             vrf_pub_key,
-            header,
+            block_timestamp,
         )?;
 
         Ok(Self::new(hash))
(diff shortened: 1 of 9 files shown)


chainstate/types/src/pos_randomness.rs
```diff
@@ -47,7 +47,7 @@ impl PoSRandomness {
 
     pub fn from_block(
         epoch_index: EpochIndex,
-        header: &BlockHeader,
+        block_timestamp: BlockTimestamp,
```
TheQuantumPhysicist, line 50: The idea of passing the header is that we can choose "whatever we need from the header", since the function is `from_block`.

common/src/chain/block/timestamp.rs
```diff
@@ -54,6 +54,14 @@ impl BlockTimestamp {
     }
 }
 
+impl From<Duration> for BlockTimestamp {
```
TheQuantumPhysicist, line 57: Please remove this conversion and use `from_duration_since_epoch()` instead. This can be ambiguous as the starting time isn't clear.

TheQuantumPhysicist (approved):
Looking good. Please check my comments.

chainstate/types/src/pos_randomness.rs
azarovh: We only require the timestamp from a header to calculate randomness. It's a more flexible and clear dependency because this way we can mine a block without the block itself by calculating `PoSData` and `BlockTimestamp` that will pass the check. 
Please take a look at `mine` function and how it's used - I think it's better this way.

common/src/chain/block/timestamp.rs
azarovh: I believe `from_duration_since_epoch` is misleading because nothing prevents me from passing any `Duration` unrelated to the epoch. I was thinking about removing these 2 functions leaving only `From` conversion as a universal one.

chainstate/types/src/pos_randomness.rs
TheQuantumPhysicist: Alrighty... thought it'll make it harder in the future to include anything else.

common/src/chain/block/timestamp.rs
azarovh: ok, reverted

PR #774, opened 2023-03-20
Basic Wasm bindings for the cryptography library

TheQuantumPhysicist:
This a proof of concept that is meant to be a starting point where we can export various functions to the web wallet and other resources through wasm.

### DIFF
--- a/wasm_crypto/index.html
+++ b/wasm_crypto/index.html
@@ -1,0 +1,85 @@
+<html>
+  <head>
+    <meta content="text/html;charset=utf-8" http-equiv="Content-Type"/>
+  </head>
+  <body>
+    <!-- Note the usage of `type=module` here as this is an ES6 module -->
+    <script type="module">
+      // Use ES module import syntax to import functionality from the module
+      // that we have compiled.
+      //
+      // Note that the `default` import is an initialization function which
+      // will "boot" the module and make it ready to use. Currently browsers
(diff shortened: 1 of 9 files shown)


wasm_crypto/index.html
```diff
@@ -0,0 +1,85 @@
+<html>
```
azarovh, line 1: More like a point to discuss, but maybe it would be better to create some kind of `crypto_binding.js` instead of a web page. Then init and call required functions there and execute .js file from console. This way we can later integrate it into our CI and maintain them easier, considering that number of bindings will grow.

azarovh (approved):

wasm-crypto/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+getrandom = { version = "0.2", features = ["js"] }
+wasm-bindgen = "0.2"
+thiserror.workspace = true
+
+[dev-dependencies]
+rstest = "0.16"
+test-utils = {path = '../test-utils'}
+
+[profile.release]
+lto = true
```
stanislav-tkach, line 27: I'm curious why lto is used for this module.

stanislav-tkach (approved):

wasm-crypto/Cargo.toml
TheQuantumPhysicist: It's recommended for wasm to make the size of the binary smaller.

PR #776, opened 2023-03-21
Add bloom filters to p2p

pavel-kokolemin:
Bitcoin Core uses rolling bloom filters in many places:
```
CRollingBloomFilter m_tx_inventory_known_filter GUARDED_BY(m_tx_inventory_mutex){50000, 0.000001};
m_addr_known = std::make_unique<CRollingBloomFilter>(5000, 0.001);
CRollingBloomFilter m_recently_announced_invs = CRollingBloomFilter{INVENTORY_MAX_RECENT_RELAY, 0.000001};
CRollingBloomFilter m_recent_rejects GUARDED_BY(::cs_main){120'000, 0.000'001};
CRollingBloomFilter m_recent_confirmed_transactions GUARDED_BY(m_recent_confirmed_transactions_mutex){48'000, 0.000'001};
```
It would be useful to have something like this.
There are a lot of rust crates that implement bloom filters, but I have not been able to find anything that works like `CRollingBloomFilter` (it remembers the last N items inserted, with some false positive probability).
But it should be possible to do something like that with regular bloom filters, inserting items into subfilters and dropping old subfilters when max number of items are inserted. That's how our `RollingBloomFilter` is implemented.

For reference, here is the speed of the `RollingBloom` implementation from Bitcoin Core (tested with `bench_bitcoin`):
```
|               ns/op |                op/s |    err% |          ins/op |          cyc/op |    IPC |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|----------------:|-------:|---------------:|--------:|----------:|:----------
|              426.07 |        2,347,045.04 |    1.0% |        3,370.13 |        1,145.04 |  2.943 |         347.87 |    0.0% |      0.01 | `RollingBloom`
```
And here is the speed of our `RollingBloomFilter` implementation, as reported by `criterion`:
```
RollingBloomFilter      time:   [672.04 ns 674.12 ns 676.25 ns]                                
```
The same item count and fpp value were used (120000 and 0.000001). The speed is comparable.

Memory usage:
Bitcoin: 1294000 bytes
Our implementation: 698440 bytes
I'm not sure why bitcoin's implementation uses more memory.

I also tested the speed of other Bloom and Cuckoo filter implementations:
```
probabilistic-collections = "0.7.0"
fastbloom-rs = "0.5.3"
bloom = "0.3.2"
bloomy = "1.2.0"
cuckoofilter = "0.5.0"
bloom-filters = "0.1.2"
bloomfilter = "1.0.9"
```
And the results were about the same (about 500..600 ns per operation).

The only exception was `fastbloom-rs`, which was much faster (about 230 ns), but it can't be used without modifications.
I decided to use the `probabilistic-collections` crate (MIT license) because its code looks cleaner.
If ~650 ns is too slow, we could switch to `fastbloom-rs` (we would have to copy its sources into our repo, although it's only 130 lines if you remove the extra stuff).
I also tested `probabilistic-collections` with `xxhash` (the hash used by `fastbloom-rs`), but the speedup was not very big.


p2p/src/utils/bloom_filters/rolling_bloom_filter.rs
```diff
@@ -0,0 +1,105 @@
...
+            size_subfilter: size_per_filter,
+            fpp_subfilter: fpp_per_filter,
+        }
+    }
+
+    /// Add item to the rolling bloom filter.
+    pub fn insert(&mut self, item: &T, rng: &mut impl Rng) {
+        self.subfilters[self.subfilter_index].insert(item);
+        self.count_subfilter += 1;
+        if self.count_subfilter == self.size_subfilter {
```
TheQuantumPhysicist, line 86: Add a line: `debug_assert!(self.count_subfilter <= self.size_subfilter);`

p2p/src/utils/bloom_filters/rolling_bloom_filter.rs
```diff
@@ -0,0 +1,105 @@
...
+            subfilter_index: 0,
+            count_subfilter: 0,
+            size_subfilter: size_per_filter,
+            fpp_subfilter: fpp_per_filter,
+        }
+    }
+
+    /// Add item to the rolling bloom filter.
+    pub fn insert(&mut self, item: &T, rng: &mut impl Rng) {
+        self.subfilters[self.subfilter_index].insert(item);
```
TheQuantumPhysicist, line 84: This implementation is so C++... I think there are ways to improve it and make it more rusty. I'll have to think how this should be done. But off the top of my head, using a reference that switches between the elements of subfilters seems the right way to go.

p2p/src/utils/bloom_filters/rolling_bloom_filter.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.count_subfilter += 1;
+        if self.count_subfilter == self.size_subfilter {
+            // The maximum number of items in the current subfilter has been reached.
+            // Create a new subfilter with new seeds to get new false positives.
+            self.subfilter_index = (self.subfilter_index + 1) % 3;
+            self.subfilters[self.subfilter_index] =
+                BloomFilter::new(self.size_subfilter, self.fpp_subfilter, rng);
+            self.count_subfilter = 0;
```
TheQuantumPhysicist, line 92: If count_subfilter is common among all bloom filters, and the filters run in a cyclic way... what's the difference between having a single filter with 3x the size and having 3 filters like this?

In fact, if size_subfilter is multiple of 3, then only 1 subfilter will be constantly reset and the others will be ignored and never reset. Am I missing something?

TheQuantumPhysicist (changes requested):
I took a quick first look. Let's have a discussion about this.

p2p/src/utils/bloom_filters/rolling_bloom_filter.rs
pavel-kokolemin: I switched from array to vector, where the newest subfilter is always the first. I think this makes the code a bit easier to read. As a bonus, new subfilters are created as needed, using less memory right after construction.
The speed difference is not noticeable (less than 3%).

pavel-kokolemin:
Note that I have added a new dependency on `crypto` to the `utils` crate (and a dev dependency on `test_utils`).

utils/src/bloom_filters/rolling_bloom_filter/cyclic_subfilter.rs
```diff
@@ -0,0 +1,44 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::VecDeque;
+
+pub struct CyclicFilter<T, const SUBFILTER_COUNT: usize> {
```
TheQuantumPhysicist, line 18: Please add docs to this and add tests that will test the individual behavior here.

utils/src/bloom_filters/rolling_bloom_filter/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+            subfilter_inserted_max,
+            fpp_subfilter,
+        }
+    }
+
+    /// Add item to the rolling bloom filter.
+    pub fn insert(&mut self, item: &T, rng: &mut impl Rng) {
+        debug_assert!(self.subfilter_inserted_count < self.subfilter_inserted_max);
+        // Insert the element into the newest subfilter
+        self.subfilters.get_current_mut().insert(item);
```
TheQuantumPhysicist, line 98: If all you need to do is insert, then `get_current_mut` isn't necessary. Just create a function `insert` in `CyclicFilter`

utils/src/bloom_filters/rolling_bloom_filter/cyclic_subfilter.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    pub fn roll_filters(&mut self, new: T) {
+        if self.list.len() == SUBFILTER_COUNT {
+            self.list.pop_front();
+        }
+        // Create a new subfilter with new seeds to get new false positives
+        self.list.push_back(new);
+    }
+
+    pub fn get_all(&self) -> impl Iterator<Item = &T> {
```
TheQuantumPhysicist, line 41: `iter`?

utils/benches/benches.rs
```diff
@@ -0,0 +1,41 @@
...
+    let mut count = 0u32;
+    c.bench_function("RollingBloomFilter", |b| {
+        b.iter(|| {
+            count += 1;
+
+            data[0..4].copy_from_slice(&count.to_le_bytes());
+            list.insert(&data, &mut rng);
+
+            data[0..4].copy_from_slice(&count.to_be_bytes());
+            list.contains(&data);
```
stanislav-tkach, line 35:
```suggestion
            list.contains(&data)
```
I doubt it will affect the results, but the return value is passed to `black_box`.

stanislav-tkach (approved):

utils/src/bloom_filters/rolling_bloom_filter/cyclic_subfilter.rs
```diff
@@ -0,0 +1,44 @@
...
+
+impl<T, const SUBFILTER_COUNT: usize> CyclicFilter<T, SUBFILTER_COUNT> {
+    pub fn new(new: T) -> Self {
+        let mut list = VecDeque::new();
+        list.push_back(new);
+        CyclicFilter { list }
+    }
+
+    pub fn get_current_mut(&mut self) -> &mut T {
+        self.list.back_mut().expect("")
```
iljakuklic, line 30: Missing expect message

utils/src/bloom_filters/rolling_bloom_filter/cyclic_subfilter.rs
```diff
@@ -0,0 +1,44 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::VecDeque;
+
+pub struct CyclicFilter<T, const SUBFILTER_COUNT: usize> {
+    list: VecDeque<T>,
+}
```
iljakuklic, line 20: Is it possible/reasonable to use `GenericArray` here, since you already know the max size?

utils/src/bloom_filters/rolling_bloom_filter/cyclic_subfilter.rs
iljakuklic, line 41: Maybe call this `.iter()`?

iljakuklic (approved):
Looks good. A couple of minor comments below:

utils/src/bloom_filters/rolling_bloom_filter/cyclic_subfilter.rs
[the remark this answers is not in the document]
pavel-kokolemin: Right, sorry. Anyway, I changed it back from `VecDeque` to `Vec` because it's a tiny bit faster (about 2%). And `expect` was no longer needed.

utils/src/bloom_filters/rolling_bloom_filter/mod.rs
pavel-kokolemin: This is not possible because `CyclicFilter` is now more like a generic container and doesn't know anything about bloom filters.

utils/benches/benches.rs
pavel-kokolemin: Nice catch, thanks. I tried with and without it, and fortunately there is no difference (otherwise all my previous benchmarks would not be accurate). I added an explicit call to `black_box` which should be more noticeable.

utils/src/bloom_filters/tests.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    let mut fp = 0;
+    for i in 400..10400 {
+        if filter.contains(&i) {
+            fp += 1;
+        }
+    }
+    // Expect about 100 false positive
+    assert!(fp > 10 && fp < 200, "invalid fp value: {fp}");
```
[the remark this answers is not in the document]
pavel-kokolemin: Good question. I tested in a loop to get a more accurate range and then increased it. So there should be no unexpected failures now, thanks. BTW, Bitcoin Core uses a fixed seed in this case.

utils/src/bloom_filters/rolling_bloom_filter/cyclic_subfilter.rs
pavel-kokolemin: Probably not, because I've switched to `Vec` now (to make the filter a bit faster and use less memory when it's just created). And before that I used a regular array, so `GenericArray` is not really needed here.

pavel-kokolemin:
@TheQuantumPhysicist asked to revert to using arrays (like it was done before).

utils/src/bloom_filters/rolling_bloom_filter/cyclic_array.rs
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    /// Returns the latest inserted element
+    pub fn get_last_mut(&mut self) -> &mut T {
+        &mut self.list[self.current_index]
+    }
+
+    /// Inserts a new item, dropping one older item
+    pub fn push(&mut self, new: T) {
```
TheQuantumPhysicist, line 36: maybe call this `roll_filter`?

TheQuantumPhysicist (approved):

utils/src/bloom_filters/rolling_bloom_filter/cyclic_array.rs
pavel-kokolemin: I have renamed this helper to `CyclicArray` because it now looks more like a generic container. So I think `push` is appropriate.

muursh (approved):

iljakuklic (approved):

PR #777, opened 2023-03-21
Add the `get_locator_from_height` function

stanislav-tkach:


stanislav-tkach:
I have found only a minor improvement: to send the locator starting from the block preceding the tip, so the `best_known_block` field is initialized even for the peers that are fully synced with our node.

p2p/src/sync/peer.rs
```diff
@@ -277,11 +283,21 @@ where
...
 
-        // We are in sync with this peer.
         if headers.is_empty() {
             return Ok(());
         }
 
+        if headers.len() == 1 {
+            let block_id = headers.first().expect("Headers shouldn't be empty").block_id();
+            if let Some(index) =
+                self.chainstate_handle.call(move |c| c.get_block_index(&block_id)).await??
```
TheQuantumPhysicist, line 293: Nitpick: I would put this in its own line

stanislav-tkach:
After a discussion I have reverted the changes in the sync manager and updated the pull request description. The header exchange will be updated in the following pull request that will also remove the distinction between requests/responses and announcements because all of these are just messages.

TheQuantumPhysicist (approved):

azarovh (approved):

PR #778, opened 2023-03-22
Send p2p services as a bit set

pavel-kokolemin:
Sending `BTreeSet<PubSubTopic>` in handshakes is not a very good idea because it's not possible to add new `PubSubTopic` values without breaking old clients. Also, `BTreeSet` is serialized as a vector of values and this is not very good for security (it's possible to encode a very long vector with duplicate values and it won't be detected as an error).
Let's switch to sending a bit set instead (encoded as `u32'). I also renamed `PubSubTopic` to `Service`.


p2p/src/net/types/services.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub const TRANSACTIONS: Service = Service(1 << 0);
+
+    /// Blocks
+    pub const BLOCKS: Service = Service(1 << 1);
+
+    /// Peer address announcements from new nodes joining the network
+    pub const PEER_ADDRESSES: Service = Service(1 << 2);
+}
+
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Encode, Decode, Default)]
```
TheQuantumPhysicist, line 32: There's a problem with this being default... what does it mean? It's meaningless and is even invalid. What is a node that has zero services?!

p2p/src/net/types/services.rs
```diff
@@ -0,0 +1,49 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Encode, Decode, Default)]
+pub struct Service(u32);
```
TheQuantumPhysicist, line 19: Why not an enum? Something like

```
pub enum PublicFlags {
    Transactions = 0x01,
    Blocks = 0x02,
    PeerAddresses = 0x04,
}
```

Then you can define how services are concatenated.

There's no particular problem with what you did, but I just find this nicer. Maybe I'm missing something.

TheQuantumPhysicist: The only issue I see with the current implementation is that it's possible to have two "arms" with the same value.

p2p/src/net/types/services.rs
```diff
@@ -0,0 +1,49 @@
...
+        self.0 & service.0 != 0
+    }
+}
+
+impl From<&[Service]> for Services {
+    fn from(services: &[Service]) -> Self {
+        let mut result = 0;
+        for service in services {
+            result |= service.0
+        }
```
TheQuantumPhysicist, line 46: How about:

```
let result = services.iter().fold(0, |so_far, current| so_far | current.0);
```

Always prefer functional solutions.

TheQuantumPhysicist (approved):
Looking good! Thank you for doing this. I have complained about this multiple times because of the attacks you mentioned, but then forgot about it.

p2p/src/net/types/services.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// Blocks
+    pub const BLOCKS: Service = Service(1 << 1);
+
+    /// Peer address announcements from new nodes joining the network
+    pub const PEER_ADDRESSES: Service = Service(1 << 2);
+}
+
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Encode, Decode, Default)]
+pub struct Services(u32);
```
TheQuantumPhysicist, line 33: As dumb as it sounds... can we make this u64?

p2p/src/net/types/services.rs
pavel-kokolemin: Right, removed. Thanks.

p2p/src/net/types/services.rs
pavel-kokolemin: Thanks for the hint, that is much better.

p2p/src/net/types/services.rs
```diff
@@ -0,0 +1,54 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+#[derive(Eq, PartialEq, Clone, Copy, Debug)]
+pub enum PublicFlag {
```
TheQuantumPhysicist, line 19: Please call this `Service`... PublicFlag is a copy mistake :D

p2p/src/net/types/services.rs
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+#[derive(Eq, PartialEq, Clone, Copy, Debug)]
```
TheQuantumPhysicist, line 18: I think you should add here `#[repr(u64)]` so that the underlying integers are u64, not u32.

p2p/src/config.rs
```diff
@@ -47,17 +47,15 @@ pub enum NodeType {
     Inactive,
```
TheQuantumPhysicist, line 47: It would be great if we could remove this

pavel-kokolemin: It's used in the tests, so I added the `cfg` attribute.

p2p/src/net/types/services.rs
pavel-kokolemin: I checked a large flag value and it works normally without it. Added it anyway.

p2p/src/net/default_backend/types.rs
```diff
@@ -24,29 +22,17 @@ use crate::{
...
-    },
-    AnnounceData {
-        topic: PubSubTopic,
-        message: Vec<u8>,
-    },
+    Connect { address: A },
+    Accept { peer_id: PeerId },
+    Disconnect { peer_id: PeerId },
+    SendMessage { peer: PeerId, message: Message },
+    AnnounceData { topic: Service, message: Vec<u8> },
```
stanislav-tkach, line 35:
```suggestion
    AnnounceData { service: Service, message: Vec<u8> },
```
?

stanislav-tkach (approved):

PR #779, opened 2023-03-22
Remove the package section from wasm-crypto

stanislav-tkach:
> warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/s/code/mintlayer-core-2/wasm-crypto/Cargo.toml
workspace: /Users/s/code/mintlayer-core-2/Cargo.toml

### DIFF
--- a/wasm-crypto/Cargo.toml
+++ b/wasm-crypto/Cargo.toml
@@ -22,6 +22,3 @@ thiserror.workspace = true
 [dev-dependencies]
 rstest.workspace = true
 test-utils = {path = '../test-utils'}
-
-[profile.release]
-lto = true


TheQuantumPhysicist (approved):

stanislav-tkach:
I have tried several approaches to enable LTO for the `wasm-crypto` crate only, but I didn't succeed:
- Per target profiles aren't supported (yet?): https://github.com/rust-lang/cargo/issues/4897
- Package specific overrides (`[profile.release.package.wasm-crypto]`) cannot specify the panic, lto, or rpath settings: https://doc.rust-lang.org/cargo/reference/profiles.html#overrides
- Crate specific [configuration](https://doc.rust-lang.org/cargo/reference/config.html) only works when the build is started from that crate directory and is ignored otherwise.

PR #781, opened 2023-03-22
Generic merkle tree in its own crate

TheQuantumPhysicist:
So I decided to bite the bullet and do this, even though there are problems with this due to derive's imprefection. I had to implement some traits manually, which pose the risk that changing member variables may lead to forgetting to update the trait implementations for things like PartialEq.


merkletree/src/merkle/proof/multi/mod.rs
```diff
@@ -292,7 +307,7 @@ impl MultiProofHashes {
         // Note: This can be made more efficient by marking "hashed" nodes and skipping them,
         // but we don't care about performance here. We care more about security.
         for (leaf_index, leaf_hash) in &leaves {
-            let mut hash = *leaf_hash;
+            let mut hash = leaf_hash.clone();
```
iljakuklic, line 310: It may be possible to replace some of these `.clone()`s with moves

merkletree/src/internal.rs
```diff
@@ -0,0 +1,73 @@
...
+// pub type HashAlgoStream = crypto::hash::Blake2b32Stream;
+
+impl From<GenericArray<u8, typenum::U32>> for HashedData {
+    fn from(val: GenericArray<u8, typenum::U32>) -> Self {
+        Self(val.into())
+    }
+}
+
+impl PairHasher for HashAlgo {
+    type Type = HashedData;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: The issue is that it's not only used in MerkleTree. It's also used all over the place. Having a generic simply decouples the MerkleTree type from the hashing function. Though I have to honestly say that I can't see how to do it otherwise without referring to MerkleTree every time I need to hash something (like I did before with default_hash and so on).

merkletree/src/merkle/proof/multi/mod.rs
TheQuantumPhysicist: I gave it a shot yesterday. Didn't work out. The only place where this can be done is at the end of the loop because the value is left just to do the `==`... but the problem there is I have to mess around and do the equality before the clone, which is not clean and I'm afraid it could cause other issues in the future. I'm hoping the compiler could be smart about this.

merkletree/src/rand_tools.rs
```diff
@@ -0,0 +1,51 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub use rand::{CryptoRng, Rng, SeedableRng};
+use rand_chacha::ChaChaRng;
+use std::{num::ParseIntError, str::FromStr};
+
+#[derive(Debug, Copy, Clone)]
+pub struct Seed(pub u64);
```
azarovh, line 21: maybe a TODO stating that these are duplicates with types in tests_utils that some days should be resolved?

azarovh (approved):

merkletree/src/rand_tools.rs
TheQuantumPhysicist: Not really. This is OK. It doesn't have to be resolved. We may move this crate outside of mintlayer-core.

Issue #518, opened 2022-10-28
p2p constants refactoring


stanislav-tkach:
We have some duplicated constants in libp2p and mock backends and most of these constants should be moved to the p2p config anyway.

TheQuantumPhysicist:
@stanislav-tkach Is this over since we don't use libp2p anymore?

stanislav-tkach:
I think there are still constants present that should be moved in the config. Additionally some of them are duplicated in different crates. For example, the `HEADER_LIMIT` is present in both chainstate and p2p.

PR #782, opened 2023-03-23
Use one config parameter instead of two constants (message size and announcement size)

stanislav-tkach:
Closes https://github.com/mintlayer/mintlayer-core/issues/518.


p2p/src/net/default_backend/transport/impls/channel.rs
```diff
@@ -212,10 +212,10 @@ mod tests {
         let peer_stream = peer_res.unwrap();
 
         let message = Message::BlockListRequest(BlockListRequest::new(vec![]));
-        let mut peer_stream = BufferedTranscoder::new(peer_stream);
+        let mut peer_stream = BufferedTranscoder::new(peer_stream, 10 * 1024 * 1024);
```
TheQuantumPhysicist, line 215: Magic numbers

p2p/src/net/default_backend/transport/impls/channel.rs
```diff
@@ -212,10 +212,10 @@ mod tests {
...
         peer_stream.send(message.clone()).await.unwrap();
 
-        let mut server_stream = BufferedTranscoder::new(server_stream);
+        let mut server_stream = BufferedTranscoder::new(server_stream, 10 * 1024 * 1024);
```
TheQuantumPhysicist, line 218: Magic numbers

p2p/src/net/default_backend/transport/impls/stream_adapter/wrapped_transport/tests.rs
```diff
@@ -221,12 +221,12 @@ async fn send_2_reqs() {
     let mut rng = test_utils::random::make_seedable_rng(test_utils::random::Seed::from_entropy());
     let id: Id<Block> = H256::random_using(&mut rng).into();
     let message_2 = Message::BlockListRequest(BlockListRequest::new(vec![id]));
-    let mut peer_stream = BufferedTranscoder::new(peer_stream);
+    let mut peer_stream = BufferedTranscoder::new(peer_stream, 10 * 1024 * 1024);
```
TheQuantumPhysicist, line 224: Magic numbers

p2p/src/net/default_backend/transport/impls/stream_adapter/wrapped_transport/tests.rs
```diff
@@ -221,12 +221,12 @@ async fn send_2_reqs() {
...
     peer_stream.send(message_1.clone()).await.unwrap();
 
     peer_stream.send(message_2.clone()).await.unwrap();
 
-    let mut server_stream = BufferedTranscoder::new(server_stream);
+    let mut server_stream = BufferedTranscoder::new(server_stream, 10 * 1024 * 1024);
```
TheQuantumPhysicist, line 229: Magic numbers

p2p/src/net/default_backend/transport/impls/tcp.rs
```diff
@@ -180,6 +180,8 @@ mod tests {
     use super::*;
     use crate::net::default_backend::{transport::BufferedTranscoder, types::Message};
 
+    const MAX_MESSAGE_SIZE: usize = 10 * 1024 * 1024;
```
TheQuantumPhysicist, line 183: Better than having such a constant is using random values with constaints.

TheQuantumPhysicist (approved):
Please randomize as much as possible in tests. Otherwise, all good.

p2p/src/net/default_backend/backend.rs
```diff
@@ -243,13 +242,13 @@ where
         announcement: Announcement,
     ) -> crate::Result<()> {
         let size = announcement.encode().len();
-        if size > ANNOUNCEMENT_MAX_SIZE {
+        if size > *self.p2p_config.max_message_size {
```
pavel-kokolemin, line 245: I think this check is useless now. We already check the message size in `EncoderDecoder` and so this would always be `false`.

p2p/src/net/default_backend/mod.rs
```diff
@@ -203,10 +212,10 @@ impl<T: TransportSocket> MessagingService for MessagingHandle<T> {
 
     fn make_announcement(&mut self, announcement: Announcement) -> crate::Result<()> {
         let message = announcement.encode();
-        if message.len() > ANNOUNCEMENT_MAX_SIZE {
+        if message.len() > *self.p2p_config.max_message_size {
```
pavel-kokolemin, line 215: I think there are some problems with this check:
- Checking the encoded `Announcement` size does not guarantee that the final message size will be less than `max_message_size` (very unlikely to be happen in practice).
- We don't check the message size in `send_message` where it is more likely to fail.
- We should never send such large announcements to trigger this.

Maybe it would be better to remove this check?

iljakuklic (approved):
Looks good. Mostly checked the coding. I don't know enough to comment on how it fits into the wider context so not making a statement about the overall design.

p2p/src/net/default_backend/mod.rs
stanislav-tkach: I agree with your reasoning, so I have removed useless checks and added a few new tests.

muursh (approved):

p2p/src/net/default_backend/mod.rs
```diff
@@ -202,19 +211,12 @@ impl<T: TransportSocket> MessagingService for MessagingHandle<T> {
     }
 
     fn make_announcement(&mut self, announcement: Announcement) -> crate::Result<()> {
-        let message = announcement.encode();
-        if message.len() > ANNOUNCEMENT_MAX_SIZE {
```
TheQuantumPhysicist, line 206: Why remove this check? It doesn't hurt. It can be an assert at least.

TheQuantumPhysicist (approved):

pavel-kokolemin (approved):

p2p/src/net/default_backend/mod.rs
stanislav-tkach: The check will be performed in `EncoderDecoder` anyway, I'm not sure if it makes sense to perform the same check on several layers.

PR #783, opened 2023-03-23
mempool: Transaction verifier integration

iljakuklic:
Very messy at the moment. Lots of error handling incomplete.

TODO:
- [x] Remove dead code
- [x] Change mempool initialization interface to prevent deadlocks
- [x] Categorize mempool errors better (not critical, maybe for a future PR)
- [x] Error handling in `ChainstateHandle` impls
- [x] Error handling in `tx_verifier_impl`
- [x] Review which newly introduced impls are needed
- [x] Improve chainstate tip check in mempool
- [x] Check which `MempoolInterface` methods still need to be `async` (probably none)
- [x] <s>Use `chainstate::Fee` in mempool instead of having a custom type</s> (not doing now, chainstate `Fee` not expressive enough)
- [x] Revisit the RPC mechanism for submitting tx

### DIFF
--- a/p2p/src/sync/tests/helpers.rs
+++ b/p2p/src/sync/tests/helpers.rs
@@ -99,7 +99,7 @@ impl SyncManagerHandle {
         let (error_sender, error_receiver) = mpsc::unbounded_channel();
         let sync_manager_handle = tokio::spawn(async move {
             let e = sync.run().await.unwrap_err();
-            error_sender.send(e).unwrap();
+            let _ = error_sender.send(e);
         });
 
         Self {
@@ -312,9 +312,9 @@ impl MessagingService for MessagingHandleMock {
 #[async_trait]
 impl SyncingEventReceiver for SyncingEventReceiverMock {
     async fn poll_next(&mut self) -> Result<SyncingEvent> {
-        Ok(time::timeout(LONG_TIMEOUT, self.events_receiver.recv())
+        time::timeout(LONG_TIMEOUT, self.events_receiver.recv())
             .await
             .expect("Failed to receive event in time")
-            .unwrap())
+            .ok_or(P2pError::ChannelClosed)
     }
 }


chainstate/src/interface/chainstate_interface.rs
```diff
@@ -166,4 +166,31 @@ pub trait ChainstateInterface: Send {
 
     /// Returns true if the initial block download isn't finished yet.
     fn is_initial_block_download(&self) -> Result<bool, ChainstateError>;
+
+    fn pool_exists(&self, pool_id: PoolId) -> Result<bool, ChainstateError>;
```
TheQuantumPhysicist, line 170: staking_pool_exists()

Pool alone is confusing without context. Same for all other instances here.

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -166,4 +166,31 @@ pub trait ChainstateInterface: Send {
...
+    fn get_pool_balance(&self, pool_id: PoolId) -> Result<Option<Amount>, ChainstateError>;
+
+    fn get_pool_data(&self, pool_id: PoolId) -> Result<Option<PoolData>, ChainstateError>;
+
+    fn get_pool_delegations_shares(
+        &self,
+        pool_id: PoolId,
+    ) -> Result<Option<std::collections::BTreeMap<DelegationId, Amount>>, ChainstateError>;
+
+    fn get_delegation_balance(
```
TheQuantumPhysicist, line 181: Same, staking delegation

mempool/src/interface/mempool_interface_impl.rs
```diff
@@ -66,59 +71,67 @@ impl<M: GetMemoryUsage + Sync + Send + 'static> MempoolInterfaceImpl<M> {
...
+        self,
         mut call_rq: CallRequest<dyn MempoolInterface>,
         mut shut_rq: ShutdownRequest,
     ) {
-        let mut chainstate_events_rx = self
-            .subscribe_to_chainstate_events()
-            .await
-            .log_err()
-            .expect("chainstate event subscription");
+        log::trace!("Starting mempool");
```
TheQuantumPhysicist, line 80: I would make this info

mempool/src/pool/tx_verifier/chainstate_handle.rs
```diff
@@ -0,0 +1,243 @@
...
+impl From<Error> for crate::error::Error {
+    fn from(e: Error) -> Self {
+        match e {
+            Error::Chainstate(e) => crate::error::Error::TxValidationError(e.into()),
+            Error::Subsystem(_) => crate::error::Error::SubsystemFailure,
+        }
+    }
+}
+
+/// A wrapper over handle to chanistate
```
TheQuantumPhysicist, line 82: chainstate

mempool/src/interface/mempool_interface_impl.rs
```diff
@@ -66,59 +71,67 @@ impl<M: GetMemoryUsage + Sync + Send + 'static> MempoolInterfaceImpl<M> {
...
+        let mut mempool = Mempool::new(
+            self.chain_config,
+            self.chainstate_handle,
+            self.time_getter,
+            self.memory_usage_estimator,
+        );
+
+        log::trace!("Subscribing to chaninstate events");
```
TheQuantumPhysicist, line 88: chainstate

mempool/src/pool/tx_verifier/chainstate_handle.rs
```diff
@@ -0,0 +1,243 @@
...
+pub struct ChainstateHandle<C>(BlockingHandle<C>);
+
+impl<C> ChainstateHandle<C> {
+    pub fn new(inner: subsystem::Handle<C>) -> Self {
+        Self(BlockingHandle::new(inner))
+    }
+}
+
+impl<C: 'static + Send> ChainstateHandle<C> {
```
TheQuantumPhysicist, line 91: Too many statics all over the place... can we define a lifetime and make static an input instead of it being the default?

mempool/src/pool/tx_verifier/chainstate_handle.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// Chainstate subsystem call. We assume chainstate is alive here.
+    pub fn call<R: 'static + Send>(
+        &self,
+        func: impl 'static + Send + FnOnce(&C) -> Result<R, ChainstateError>,
```
TheQuantumPhysicist, line 95: The challenge is that some of these statics are contravariant... which is OK. So trying to make everything the same lifetime will be too constrained and it won't work.

test/functional/mempool_submit_tx.py
```diff
@@ -0,0 +1,92 @@
...
+        encoded_tx = signed_tx_obj.encode(signed_tx).to_hex()[2:]
+        self.log.debug("Encoded transaction: {}".format(encoded_tx))
+
+        try:
+            node.p2p_submit_transaction(encoded_tx)
+        except JSONRPCException as err:
+            self.log.debug("Error message: {}".format(err))
+            assert 'Transaction has no inputs' in str(err)
+        else:
+            raise AssertionError('Expected the tx submission to fail')
```
TheQuantumPhysicist, line 62: You can use `assert_raises_rpc_error()` instead of manually doing all this

test/functional/mempool_submit_tx.py
```diff
@@ -0,0 +1,92 @@
...
+            'time_lock': 0,
+        }
+        signed_tx = {
+            'transaction': tx,
+            'signatures': [witness],
+        }
+        encoded_tx = signed_tx_obj.encode(signed_tx).to_hex()[2:]
+        self.log.debug('Encoded transaction: {}'.format(encoded_tx))
+
+        node.p2p_submit_transaction(encoded_tx)
```
TheQuantumPhysicist, line 88: Please also add a test for `contains_transaction` rpc function to see that the transaction is now in the mempool.

TheQuantumPhysicist (approved):
Looking good.

I'm wondering whether making the mempool async is a good idea, given that every test now needs a multi-threaded flavor to work.

mempool/src/pool/store.rs
```diff
@@ -95,6 +95,14 @@ pub struct MempoolStore {
     // We keep the information of which outpoints are spent by entries currently in the mempool.
     // This allows us to recognize conflicts (double-spends) and handle them
     pub spender_txs: BTreeMap<OutPoint, Id<Transaction>>,
+
+    // Track transactions by internal unique sequence number
+    pub txs_by_seq_no: BTreeMap<usize, Id<Transaction>>,
+
+    pub seq_nos_by_tx: BTreeMap<Id<Transaction>, usize>,
```
TheQuantumPhysicist, line 102: Please add more documentation

mempool/src/pool/mod.rs
```diff
@@ -263,77 +269,78 @@ where
...
+            .connect_transaction(
+                &TransactionSourceForConnect::Mempool {
+                    current_best: &current_best,
+                },
+                &tx,
+                &BlockTimestamp::from_duration_since_epoch(self.clock.get_time()),
+            )?
+            .ok_or(TxValidationError::FeeNotDetermined)?;
 
+        let tx = TxWithFee::new_with_fee(tx.clone(), fee.into());
```
TheQuantumPhysicist, line 317: Why clone if tx is passed by value to this function?

mempool/src/pool/mod.rs
```diff
@@ -263,77 +269,78 @@ where
...
+        self.pays_minimum_relay_fees(&tx)?;
         self.pays_minimum_mempool_fee(&tx)?;
 
-        Ok(conflicts)
-    }
+        let conflicts = if ENABLE_RBF {
+            self.rbf_checks(&tx)?
+        } else {
+            // Wihtout RBF enabled, any conflicting transaction results in an error
```
TheQuantumPhysicist, line 324: Without

test/functional/mempool_submit_tx.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017-2021 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Mempool transaction submission test
+
+Check that:
+* Valid transaciton is accepted by the mempool and included in a block as
```
TheQuantumPhysicist, line 8: transaction

chainstate/src/interface/chainstate_interface.rs
iljakuklic: Good point, will update all staking-related methods exposed in chainstate.

mempool/src/pool/mod.rs
```diff
@@ -263,77 +269,78 @@ where
         // We don't have a notion of MAX_MONEY like Bitcoin Core does, so we also don't have a
         // `MoneyRange` check like the one found in Bitcoin Core's `CheckTransaction`
 
-        if tx.transaction().inputs().is_empty() {
-            return Err(TxValidationError::NoInputs);
-        }
+        // TODO: The following three checks may each better be discarded or moved to tx verifier
+        ensure!(
+            !tx.transaction().inputs().is_empty(),
+            TxValidationError::NoInputs,
```
azarovh, line 275: This case is already handled in verifier and it will claim that tx attempts to print money. Also in PR #796 I explicitly added such a rule for inputs.

mempool/src/pool/mod.rs
```diff
@@ -263,77 +269,78 @@ where
...
+        );
 
-        if tx.transaction().outputs().is_empty() {
-            return Err(TxValidationError::NoOutputs);
-        }
+        ensure!(
+            !tx.transaction().outputs().is_empty(),
+            TxValidationError::NoOutputs,
```
azarovh, line 280: Isn't it ok to burn utxo this way?
If not I'll add such an error to `input_output_poliecy.rs` and we can get rid of this check here

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -166,4 +166,31 @@ pub trait ChainstateInterface: Send {
 
     /// Returns true if the initial block download isn't finished yet.
     fn is_initial_block_download(&self) -> Result<bool, ChainstateError>;
+
+    fn pool_exists(&self, pool_id: PoolId) -> Result<bool, ChainstateError>;
+
+    fn get_pool_balance(&self, pool_id: PoolId) -> Result<Option<Amount>, ChainstateError>;
```
azarovh, line 172: What if we make `PoSAccountingView` a  supertrait of `ChainstateInterface`?

mempool/src/pool/mod.rs
```diff
@@ -710,19 +727,26 @@ where
...
-{
-    let mut uniq = BTreeSet::new();
-    iter.into_iter().any(move |x| !uniq.insert(x))
+        self.rolling_fee_rate.write().set_block_since_last_rolling_fee_bump(true);
+
+        // Take all the mempool previous transactions
+        let old_store = std::mem::replace(&mut self.store, MempoolStore::new());
+
+        // Discard the old tx verifier and replace it with a fresh one
+        self.tx_verifier = tx_verifier::create(
```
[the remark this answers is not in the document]
iljakuklic: I need to keep the tx verifier state (such as utxo set) around across the invocations of `add_transaction`. New transactions are added to mempool subject to the latest state implied by the mempool contents, and the long-lived tx verifier keeps track of that. Hope I did not miss something.

chainstate/src/interface/chainstate_interface.rs
iljakuklic: I am open to this possibility. However, as @TheQuantumPhysicist noted above, chainstate interface already contains many methods so it would be better to be more explicit with naming, e.g. `pool_exists` -> `stake_pool_exists`. In that case, the `PoSAccountingView` would have to be renamed too. Keeping them separate allows us to keep the more concise names that go with the more specialized interface.

Ultimatley I don't have a strong preference either way. If we add `PoSAccountingView` as a supertrait, we should consider whether the same should be done with `UtxosView` as well.

mempool/src/pool/mod.rs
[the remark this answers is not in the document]
azarovh: > I need to keep the tx verifier state (such as utxo set) around across the invocations of add_transaction.

If you create a verifier on the stack before the loop you will have it. But also you won't need to bother with flushing and discarding verifiers. 

Unless you need the same verifier across calls to `new_tip_set`, but I don't see it.

mempool/src/pool/mod.rs
iljakuklic: > Isn't it ok to burn utxo this way? If not I'll add such an error to `input_output_poliecy.rs` and we can get rid of this check here

Yea I am not sure. I am inclined to allow this but I don't claim I can work out all the security implications. Just as an aside, this does not burn the coins but rather pays them to the miner as a fee (even tokens when paying any token as a fee is implemented).

There's also distinction between consensus rules and mempool policy. The node may choose to not accept a transaction in the mempool but consider it valid if it appears in a block.

mempool/src/pool/mod.rs
iljakuklic: Cool will remove it from here then

mempool/src/pool/tx_verifier/chainstate_handle.rs
iljakuklic: This is a chainstate handle which is currently only instantiated with `Box<dyn ChainstateInterface>`. There may be merit in simplifying the code by just using the `Box<dyn ChainstateInterface>` type directly without generics. We don't lose the ability to mock since `dyn ChainstateInterface` is still there. Thoughts?

iljakuklic:
> I'm wondering whether making the mempool async is a good idea, given that every test now needs a multi-threaded flavor to work.

This is not a consequence of making it `async`. To the contrary, a lot of `async` stuff has been removed. We now require multithreading scheduler because the blocking chainstate handle does *not* use async so calls may have to be issued to a separate thread to avoid a deadlock.

mempool/src/pool/mod.rs
```diff
@@ -263,77 +272,78 @@ where
...
-                    Err(TxValidationError::OutPointNotFound {
-                        outpoint: input.outpoint().clone(),
-                        spending_tx_id: tx.transaction().get_id(),
-                    })
-                },
-            )
+        let delta = tx_verifier.consume()?;
+
+        let final_tip = chainstate_handle.call(|c| c.get_best_block_id())??;
+        ensure!(tip == final_tip, TxValidationError::TipMoved);
```
pavel-kokolemin, line 338: I don't like the way the new tip is processed here, because valid transactions may be rejected unexpectedly. Isn't it better to try again instead?

mempool/src/pool/mod.rs
TheQuantumPhysicist: I think this should be a decision coming from the tx-verifier.

We can have mempool rules, but then a question is whether want these rules done in a separate, clear function that clarifies this difference.

mempool/src/pool/tx_verifier/chainstate_handle.rs
TheQuantumPhysicist: Well, any Deref<ChainstateInterface> implements ChainstateInterface... I don't think mocks will be affected. So with my wild imagination I'm thining we could generalize this further and use generic lifetimes instead of 'static, because with 'static only that Box will work, so there's a question to the benefit of this.

It sounds worse to just use Box<dyn ChainstateInterface> directly, but when we put everything in context, it will look like that we should either generalize lifetimes or just use that Box directly, unless there are plans in the future to change something relevant and having this kind of generic type with non-generic lifetime helps with that.

I don't know what the right answer is... but it doesn't seem right as it's now.

pavel-kokolemin:
> ... the blocking chainstate handle does _not_ use async so calls may have to be issued to a separate thread to avoid a deadlock.

But why change that? Why introduce a different behavior (blocking sync code) into the codebase?

mempool/src/pool/mod.rs
iljakuklic: I see two possibilities here:

1. Retry adding the transaction if tip moved, with a limited number of attempts.
2. Change the tx verifier in mempool to pre-fetch all data needed to validate a transaction in one `call` to get a consistent view.

In the long run, we should probably go for (2) but I think (1) will do in the meantime.

iljakuklic:
> > ... the blocking chainstate handle does _not_ use async so calls may have to be issued to a separate thread to avoid a deadlock.
> 
> But why change that? Why introduce a different behavior (blocking sync code) into the codebase?

@TheQuantumPhysicist insisted on transaction verifier traits not being async for reasons I did not quite understand.

iljakuklic:
> > > > ... the blocking chainstate handle does _not_ use async so calls may have to be issued to a separate thread to avoid a deadlock.
> > > 
> > > 
> > > But why change that? Why introduce a different behavior (blocking sync code) into the codebase?
> > 
> > 
> > @TheQuantumPhysicist insisted on transaction verifier traits not being async for reasons I did not quite understand.
> 
> The problem is that chainstate block verification must be consistent, provably so. We cannot split the operation of verifying blocks because that will make it impossible to prove that the state is consistent when verifying blocks. This can be seen in the process_block function in chainstate. We can argue day and night that this is easily done by making everything async, but remember that we're using a real [database whose transaction, once started, will lock a mutex](https://github.com/mintlayer/mintlayer-core/blob/9089d04918dceab1dfd72888a2dcfa3da22440c1/chainstate/src/detail/mod.rs#L282), which will easily lead to dead-locks. This is why chainstate will always be synchronous.

I'm still not quite sure how avoiding async helps compared to using threads. The operation is still interleaved, and deadlocks are still possible in very much the same way. What am I missing? I don't see how this changes the nature of the concurrency issues we're facing. I can see how it may be intrusive to update substantial amount of code to support `async` but that's a separate problem.

mempool/src/pool/tx_verifier/chainstate_handle.rs
iljakuklic: As far as this `call` method is concerned, I only use it to implement the `tx_verifier` traits. It is sufficient for that use case. I made it private so it's clear we don't need it to be as general as possible, just general enough to deal with stuff that uses it locally.

On the broader question of `ChainstateHandle`, that too I consider an implementation detail. It lives in a private submodule of `mempool::tx_verifier` so it is not accessible from outside and mempool only accesses it through the transaction verifier traits. So at least for now, it is merely an implementation detail. Do you envision it being more generally useful?

mempool/src/pool/mod.rs
iljakuklic: > > We can have mempool rules, but then a question is whether want these rules done in a separate, clear function that clarifies this difference.
> 
> Having a separate function for this is on my todo list

Added some basic separation

mempool/src/pool/tx_verifier/chainstate_handle.rs
TheQuantumPhysicist: That's fair. It's just that the intent from the design is unclear. I'm OK with it... not a big deal.

mempool/src/pool/mod.rs
iljakuklic: > I see two possibilities here:
> 
> 1. Retry adding the transaction if tip moved, with a limited number of attempts.
> 2. Change the tx verifier in mempool to pre-fetch all data needed to validate a transaction in one `call` to get a consistent view.
> 
> In the long run, we should probably go for (2) but I think (1) will do in the meantime.

Implemented approach (1) for the time of being

TheQuantumPhysicist:
> I'm still not quite sure how avoiding async helps compared to using threads. The operation is still interleaved, and deadlocks are still possible in very much the same way. What am I missing? I don't see how this changes the nature of the concurrency issues we're facing. I can see how it may be intrusive to update substantial amount of code to support `async` but that's a separate problem

I don't know whether I'm missing the point, and please don't misunderstand my impatience. I'm just really confused. 

First, I'm not suggesting threads with regards to this problem. I don't know why this is coming up. I'm talking specifically about async in the middle of tx-verifier (and chainstate) (i.e., using the pre-defined thread-pool of the runtime executor).

But here's the more important point: Why is everything blocking/synchronous? Because you can't shard the state while updating a block processing. For example: You can't update the block index data, and not update the utxo set. That will basically lead to an invalid state. So, from the beginning of the mutable storage database transaction that is meant to update the state of chainstate to its end, nothing will be able to read anything from chainstate storage (unless cached in memory) because if it's possible, it will be an invalid state.

Now, does that mean that I have proof that it's inconceivable to have a manually managed sharded state of the storage where we can have everything async and inconsistent and yet read correct results because we do virtual locking in the database state instead of doing it using database transactions? Of course not. This is not what I'm saying. But I'm saying that we designed a simple, reliable and understandable set of rules on how to use chainstate that's based on ACID properties of the database. I don't even think it's of any importance to answer that question because that's fairytale world that doesn't contribute a solution to this problem. Of course, given infinite time, we can do anything. But even when I worked in large corporations, it was known how nasty it's to manually shard the database to achieve perfect partitioning of a system.

The reason I'm confused is, we made a design decision, and this design decision has limitations, and for some reason it's being critiqued as a bad idea for some idealistic reason where we're preferring complexity over correctness **in chainstate**. Somehow it looks like you guys have a better design in mind that was never brought up. As you guys know, I'm very open to hearing ideas and implementing new solutions to problems. Please feel free to drop them here or in a meeting.

mempool/src/pool/tx_verifier/chainstate_handle.rs
iljakuklic: I removed the generics here. I don't see this being used with multitude of parameters, and it is a bit confusing with the generics. We can re-introduce them if needed.

TheQuantumPhysicist:
Thank you, Pavel!

Is this good to be merged?

iljakuklic:
> Thank you, Pavel!
> 
> Is this good to be merged?

Will have to resolve some conflicts first. I've seen some failures even with the latest update but can't reproduce them anymore. Hopefully it's fine with a clean build.

muursh (approved):

stanislav-tkach (approved):

PR #784, opened 2023-03-24
WIP: PoS retargeting

azarovh:


chainstate/test-suite/src/tests/helpers/pos.rs
```diff
@@ -0,0 +1,103 @@
...
+        match tf.chainstate.get_chain_config().net_upgrade().consensus_status(block_height) {
+            RequiredConsensus::PoS(status) => status,
+            RequiredConsensus::PoW(_)
+            | RequiredConsensus::DSA
+            | RequiredConsensus::IgnoreConsensus => {
+                panic!("Invalid consensus")
+            }
+        };
+
+    let tmp_block = tf.make_block_builder().build();
```
TheQuantumPhysicist, line 91: Please don't call variables tmp. Everything on the stack is temporary.

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,477 @@
...
+fn calculate_new_target(
+    pos_config: &PoSChainConfig,
+    prev_target: &Uint256,
+    average_block_time: u64,
+) -> Result<Compact, ConsensusPoSError> {
+    let average_block_time = Uint256::from_u64(average_block_time);
+    let target_block_time = Uint256::from_u64(pos_config.target_block_time().as_secs());
+
+    // TODO: limiting factor (mintlayer/mintlayer-core#787)
+    let new_target = *prev_target / target_block_time * average_block_time;
```
TheQuantumPhysicist, line 79: Please assert that target_block_time is not zero.

common/src/chain/pos.rs
```diff
@@ -24,3 +29,56 @@ pub type PoolId = Id<Pool>;
 #[derive(Eq, PartialEq, TypeName)]
 pub enum Delegation {}
 pub type DelegationId = Id<Delegation>;
+
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
+pub struct PoSChainConfig {
+    /// The lowest possible difficulty
+    target_limit: Uint256,
+    /// Time interval between the blocks targeted by the difficulty adjustment algorithm
+    target_block_time: Duration,
```
TheQuantumPhysicist, line 38: Now that I think about it, there's a concern that data will be lost since we use this strictly as seconds. Someone can do `from_millis(500)`, which will return zero for seconds.

We need this to be strictly seconds.

common/src/chain/upgrades/netupgrade.rs
```diff
@@ -54,25 +55,29 @@ pub trait Activate {
...
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd)]
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
 pub enum RequiredConsensus {
-    // Either genesis or previous block was not PoW
     PoW(PoWStatus),
-    PoS,
+    PoS(PoSStatus),
     DSA,
     IgnoreConsensus,
 }
```
TheQuantumPhysicist, line 83: These two should probably go.

common/src/primitives/compact.rs
```diff
@@ -136,19 +136,4 @@ mod tests {
         err_conversion(0x01fedcba);
         err_conversion(!0x00800000); // overflow
     }
-    /*
-
-    #[test]
-    fn test_min_difficulty() {
```
TheQuantumPhysicist, line 142: Can we reenable this?

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,477 @@
...
+    let prev_block_id = match block_header.prev_block_id().classify(chain_config) {
+        GenBlockId::Genesis(_) => return Ok(pos_config.target_limit().into()),
+        GenBlockId::Block(id) => id,
+    };
+    let prev_block_index = block_index_handle
+        .get_block_index(&prev_block_id)?
+        .ok_or_else(|| ConsensusPoSError::PrevBlockIndexNotFound(block_header.get_id()))?;
+
+    // check if prev block is a net upgrade threshold
+    match chain_config.net_upgrade().consensus_status(prev_block_index.block_height()) {
```
TheQuantumPhysicist, line 109: I don't think this check is necessary. 

At the beginning of the function, the initial difficulty can be retrieved from PoS status. Then, if it's NOT initial, then just take the differences and calculate the average block time, and then use the last target... as usual. No need for special treatments.

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,477 @@
...
+        }
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn calculate_new_target_too_easy(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+        let config = PoSChainConfig::new(Uint256::ONE, Duration::from_secs(1), 1.into(), 1);
+        let prev_target = rng.gen_range(2..u64::MAX);
```
TheQuantumPhysicist, line 310: Why not random Uint256? You can easily do that by going from H256.

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,477 @@
...
+        .filter(|block_index| net_version_range.contains(&block_index.block_height()))
+        .map(|block_index| block_index.block_timestamp().as_int_seconds())
+        .collect::<Vec<_>>();
+
+    // At least 2 items are required to calculate difference
+    debug_assert!(block_times.len() >= 2);
+
+    // Block times are taken from history so they are sorted backwards
+    let block_diffs = block_times
+        .iter()
```
TheQuantumPhysicist, line 61: into_iter since this is consuming `block_times`.

common/src/chain/pos.rs
```diff
@@ -24,3 +29,56 @@ pub type PoolId = Id<Pool>;
...
+
+    pub fn target_block_time(&self) -> Duration {
+        self.target_block_time
+    }
+
+    pub fn reward_maturity_distance(&self) -> BlockDistance {
+        self.reward_maturity_distance
+    }
+
+    pub fn blocks_count_to_average(&self) -> usize {
```
TheQuantumPhysicist, line 72: Maybe `block_count_to_average_for_blocktime`? It's kind of unclear here.

Same for member variable.

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,477 @@
...
+        .tuple_windows::<(&u64, &u64)>()
+        .map(|t| t.0 - t.1)
+        .collect::<Vec<_>>();
```
TheQuantumPhysicist, line 64: Same here. No need to collect.

TheQuantumPhysicist: When going through this, please ensure you write tests for all possible invalid sizes.

chainstate/test-suite/src/tests/helpers/pos.rs
```diff
@@ -0,0 +1,103 @@
...
+use consensus::ConsensusPoSError;
+use crypto::vrf::{VRFPrivateKey, VRFPublicKey};
+
+use super::block_index_handle_impl::TestBlockIndexHandle;
+
+#[allow(clippy::too_many_arguments)]
+pub fn pos_mine(
+    initial_timestamp: BlockTimestamp,
+    kernel_outpoint: OutPoint,
+    vrf_pk: &VRFPublicKey,
```
TheQuantumPhysicist, line 37: Nitpick: You can produce the public key from the private key. 

Maybe this is irrelevant if this is part of the test where an invalid case is tested as well.

TheQuantumPhysicist (approved):
Looking good. Please check the comments I made.

common/src/chain/upgrades/netupgrade.rs
azarovh: `IgnoreConsensus ` seems pretty useful for unit tests. Otherwise, lot's of boilerplate code has to be written

common/src/primitives/compact.rs
azarovh: I think the test is not correct: it assumes that roundtrip conversion is possible between Compact <-> UInt256.

consensus/src/pos/target.rs
azarovh: I want it to be always bigger than initial target which is set to 1

common/src/primitives/compact.rs
TheQuantumPhysicist: I think it should be possible for minimum difficulty.

consensus/src/pos/target.rs
TheQuantumPhysicist: This is resolved but I don't see the assert...

common/src/chain/upgrades/netupgrade.rs
TheQuantumPhysicist: I'm wondering whether we should make the updates generic over this enum, such that a dangerous value like `IgnoreConsensus` can't fall into mainnet. Please create an issue for this.

consensus/src/pos/target.rs
azarovh: I removed one collect out of 2. Not sure how to achieve it in a clean way with iterators.

common/src/primitives/compact.rs
azarovh: hmmm, not sure the test doesn't pass and I thought it should:
```
Uint256::MAX => Compact(0x2100FFFF)
Compact(0x2100FFFF) => Uint256(0xFFFF0000, 0x0, 0x0, 0x0)
```

common/src/chain/upgrades/netupgrade.rs
azarovh: what exactly do you mean by "make the updates generic over this enum" ?

azarovh: #792

TheQuantumPhysicist (approved):

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,567 @@
...
+    let target_block_time = Uint256::from_u64(pos_config.target_block_time().get());
+
+    // TODO: limiting factor (mintlayer/mintlayer-core#787)
+    let new_target = *prev_target / target_block_time * average_block_time;
+
+    if new_target > pos_config.target_limit() {
+        Ok(Compact::from(pos_config.target_limit()))
+    } else {
+        Ok(Compact::from(new_target))
+    }
```
iljakuklic, line 83: Looks like a job for `std::cmp::min`

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,567 @@
...
+fn calculate_new_target(
+    pos_config: &PoSChainConfig,
+    prev_target: &Uint256,
+    average_block_time: u64,
+) -> Result<Compact, ConsensusPoSError> {
+    let average_block_time = Uint256::from_u64(average_block_time);
+    let target_block_time = Uint256::from_u64(pos_config.target_block_time().get());
+
+    // TODO: limiting factor (mintlayer/mintlayer-core#787)
+    let new_target = *prev_target / target_block_time * average_block_time;
```
iljakuklic, line 77: We need to be careful about the operation order here and in `calculate_average_block_time`. This is due to integral division rounding.

When using real numbers, the following give the same result for `a` and `b`:
`a = x / y * z`
`b = x * z / y`

In integer arithmetic, consider `x = 10, y = 11, z = 8`. Then `a = (10 / 11) * 8 = 0 * 8 = 0` whereas `b = (10 * 8) / 11 = 80 / 11 = 7`. For reference, the real valued result is about 7.2727 (= 7 + 3/11) for both expressions. So if `prev_target` is only less than target block time (even by tiny bit), `new_target` will be zero. (Unless the conversion to `Uint256` acts as a fixed point decimal number with extra precision.)

In general, the expression used to calculate `b` is more precise. However, calculating the `x * z` subexpression is more likely to overflow so it may be a good idea to perform it on type of twice the size depending of what sort of range the inputs can be (I see it's `Uint256`, so that should be fine).

We need to be extra careful here also because it's consensus critical code and will be very difficult to change once the thing is running.

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,567 @@
...
+        .expect("NetUpgrade must've been initialized");
+
+    // Get timestamps from the history but make sure they belong to the same consensus version.
+    // Then calculate differences of adjacent elements.
+    let block_diffs = history_iter
+        .take(pos_config.block_count_to_average_for_blocktime())
+        .filter(|block_index| net_version_range.contains(&block_index.block_height()))
+        .map(|block_index| block_index.block_timestamp().as_int_seconds())
+        .tuple_windows::<(u64, u64)>()
+        .map(|t| t.0 - t.1);
```
iljakuklic, line 57: I think the `tuple_windows` and `map` will, when finally summed, give you just the difference between the first and the last element. Could it be implemented easily without taking differences of adjacent elements? Maybe it does not matter much, and I don't immediately see a nice solution that works on iterators.

`(x0 - x1) + (x1 - x2) + (x2 - x3) + ... (xk - xn)` = `x0 + (-x1 + x1) + (-x2 + x2) + ... - xn` = `x0 + 0 + 0 + ... - xn` = `x0 - xn`

iljakuklic (commented):
Adding some comments. Please pay extra attention the comments mentioning integer arithmetic

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,567 @@
...
+
+    let (sum, count) = block_diffs.fold((0u64, 0u64), |(sum, count), curr| (sum + curr, count + 1));
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I wouldn't prefer this for simplicity and to avoid potential overflows.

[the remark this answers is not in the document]
iljakuklic: Maybe we should establish what sort of range of values do we expect in normal operation and under stress. Then work out how much error accumulates as the result and decide whether the precision loss is tolerable. The approach with single division guarantees the result will be same as calculating with unlimited precision and truncating the decimal part at the end. When using multiple division and feeding them into subsequent operations, the error may spiral out fairly quickly.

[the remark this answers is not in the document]
TheQuantumPhysicist: Well, prev_target can be Uint256::MAX.

consensus/src/pos/target.rs
iljakuklic: Ok, if `prev_target` is consistently a couple of orders of magnitude larger a number than `target_block_time`, this is probably not too bad.

chainstate/test-suite/src/tests/pos_retargeting_tests.rs
```diff
@@ -0,0 +1,212 @@
...
+            initial_block_time,
+            OutPoint::new(best_block_outputs.keys().next().unwrap().clone(), 0),
+            &vrf_sk,
+            chainstate_types::pos_randomness::PoSRandomness::new(sealed_epoch_randomness),
+            pool_id,
+            STAKED_BALANCE,
+            current_epoch_index,
+            new_target,
+        )
+        .expect("should be able to mine");
```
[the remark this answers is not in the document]
azarovh: Not 0 but pretty low. To be specific they equal to getting a random VRF hash above target (Uint256(0xffff0000..0000) for 1000 times in a row

consensus/src/pos/target.rs
azarovh: I will change the order and handle potential overflow

chainstate/test-suite/src/tests/history_iteration.rs
```diff
@@ -0,0 +1,82 @@
...
+        ///// test history iterator - start from an invalid non-existing block id
+        {
+            let db_tx = tf.storage.transaction_ro().unwrap();
+            let block_index_handle =
+                TestBlockIndexHandle::new(db_tx, tf.chainstate.get_chain_config().as_ref());
+            let mut iter =
+                BlockIndexHistoryIterator::new(Id::new(H256::zero()), &block_index_handle);
+
+            assert!(iter.next().is_none());
+            assert!(iter.next().is_none());
```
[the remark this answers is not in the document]
TheQuantumPhysicist: @azarovh Please add `impl<'a, H: BlockIndexHandle> FusedIterator for BlockIndexHistoryIterator<'a, H> {}` to BlockIndexHistoryIterator module somewhere under the struct.

iljakuklic (approved):
Looks good. One more discussion point below:

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,566 @@
...
+    prev_target: &Uint256,
+    average_block_time: u64,
+) -> Result<Compact, ConsensusPoSError> {
+    let average_block_time = Uint512::from_u64(average_block_time);
+    let target_block_time = Uint512::from_u64(pos_config.target_block_time().get());
+    let prev_target: Uint512 = (*prev_target).into();
+
+    // TODO: limiting factor (mintlayer/mintlayer-core#787)
+    let new_target = prev_target * average_block_time / target_block_time;
+    let new_target = Uint256::try_from(new_target)?;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: There's a min difficulty to protect from this.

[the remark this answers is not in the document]
iljakuklic: Ah I see now. In that case it makes sense to me to saturate if `new_target` happens to be over `Uint256::MAX`, rather than return an error. Since it is going to be "saturated" at the min difficulty target level anyway.

stanislav-tkach (approved):

consensus/src/pos/target.rs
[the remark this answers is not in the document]
azarovh: sounds reasonable, done

consensus/src/pos/error.rs
```diff
@@ -47,4 +48,12 @@ pub enum ConsensusPoSError {
     PoSAccountingError(#[from] pos_accounting::Error),
     #[error("PoS randomness error: `{0}`")]
     RandomnessError(#[from] PoSRandomnessError),
+    #[error("Invalid target value: `{0:?}`")]
+    InvalidTarget(Compact),
+    #[error("Decoding bits of block failed: `{0:?}`")]
+    DecodingBitsFailed(Compact),
+    #[error("Failed to conver target type: `{0:?}`")]
```
TheQuantumPhysicist, line 55: conver

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,575 @@
...
+        .unwrap_or(BlockHeight::zero()),
+    );
+
+    let time_span_start = block_index_handle
+        .get_ancestor(block_index, timespan_start_height)?
+        .block_timestamp()
+        .as_int_seconds();
+    let current_block_time = block_index.block_timestamp().as_int_seconds();
+
+    let blocks_in_timespan: i64 = (block_index.block_height() - timespan_start_height)
```
TheQuantumPhysicist, line 61: Please make this an error, not an expect. Then write an error log message starting with CRITICAL like I usually do.

The reason is that this depends on the invariant that time is always monotonic, which is generally difficult to uphold accross the whole application if decisions are made in the future to change this.

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,575 @@
...
+        .expect("cannot be negative")
+        .into();
+
+    let average = (current_block_time - time_span_start) / blocks_in_timespan as u64;
```
TheQuantumPhysicist, line 65: The previous comment should've gone here. Please make this a checked_sub and make it an error if it isn't positive, then write an error message that starts with CRITICAL. 

Reason if it's missed: This depends on the constraint that this is monotonic, which can change in the future.

azarovh: Should it be Err or expect?

consensus/src/pos/target.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    Ok(average)
+}
+
+fn calculate_new_target(
+    pos_config: &PoSChainConfig,
+    prev_target: &Uint256,
+    average_block_time: u64,
```
TheQuantumPhysicist, line 73: I think this should be called `block_time`... it's an assumption that it's an "average". It depends on other function calls.

consensus/src/pos/target.rs
TheQuantumPhysicist: Error + log error `CRITICAL:` in the beginning of the message

TheQuantumPhysicist (approved):

PR #785, opened 2023-03-25
Fix the get_locator_from_height test

stanislav-tkach:


TheQuantumPhysicist (commented):
You don't need to use floating point numbers and deal with their instability issues. Little understanding of bitshifting and trailing zeros is more than enough to do all the checks needed.

Basically log2 of an integer is its binary representation's trailing zeros.

We 100% avoid using floating point numbers.

TheQuantumPhysicist:
I fixed the tests myself and pushed.

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #789, opened 2023-03-28
Initialize consensus data

alfiedotwtf:
This change moves the consensus fetching calls into closures

I'll apply some of the changes in #755 on top of this PR.

Note: the chainstate subsystem call in `block_maker` is only used to read in data for PoW mining, and not the mining itself. This way, PoW mining will stays within the `blockprod` subsystem

### DIFF
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -96,12 +96,12 @@ impl BlockMaker {
     pub async fn make_block(
         &self,
         current_tip_id: Id<Block>,
-        transactions: Transactions,
+        transactions: &[SignedTransaction],
     ) -> Result<Block, BlockProductionError> {
         // TODO: this isn't efficient. We have to create the header first, then see if it obeys consensus rules, then construct the full block
         let current_time = self.time_getter.get_time();
         let mut block = Block::new(
-            transactions.transactions().to_vec(),
+            transactions.to_vec(),
             current_tip_id.into(),
             BlockTimestamp::from_duration_since_epoch(current_time),
             common::chain::block::ConsensusData::None,
@@ -194,10 +194,10 @@ impl BlockMaker {
     /// 2. A new tip is now on chainstate, indicating that there's no point in continuing to mine/stake at that tip
     pub async fn run(&mut self) -> Result<(), BlockProductionError> {
         let accumulator = self.collect_transactions().await?;
+        let transactions = accumulator.transactions();
 
         // TODO: do we want to introduce a separate executor for this loop to avoid starving other tasks?
         loop {
-            let transactions = Transactions::from(&accumulator);
             let block = self.make_block(self.current_tip_id, transactions).await?;
 
             match self.attempt_submit_new_block(block).await? {
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -113,7 +113,7 @@ impl BlockMaker {
             .chainstate_handle
             .call({
                 let chain_config = self.chain_config.clone();
-                let current_tip_height = self.current_tip_height.clone();
+                let current_tip_height = self.current_tip_height;
 
                 move |this| -> Result<Block, ConsensusVerificationError> {
                     let get_block_index = |&prev_block_id: &Id<Block>| {
@@ -123,15 +123,18 @@ impl BlockMaker {
                     };
 
                     let get_ancestor = |block_index: &BlockIndex, ancestor_height: BlockHeight| {
-                        this.get_ancestor(&block_index.clone().into_gen_block_index(), ancestor_height)
-                            .map_err(|_| {
-                                PropertyQueryError::GetAncestorError(
-                                    GetAncestorError::InvalidAncestorHeight {
-                                        block_height: block_index.block_height(),
-                                        ancestor_height: ancestor_height,
-                                    },
-                                )
-                            })
+                        this.get_ancestor(
+                            &block_index.clone().into_gen_block_index(),
+                            ancestor_height,
+                        )
+                        .map_err(|_| {
+                            PropertyQueryError::GetAncestorError(
+                                GetAncestorError::InvalidAncestorHeight {
+                                    block_height: block_index.block_height(),
+                                    ancestor_height,
+                                },
+                            )
+                        })
                     };
 
                     consensus::initialize_consensus_data(
@@ -146,7 +149,7 @@ impl BlockMaker {
                 }
             })
             .await?
-            .map_err(|err| BlockProductionError::FailedConsensusInitialization(err))?;
+            .map_err(BlockProductionError::FailedConsensusInitialization)?;
 
         // TODO: consensus::finalize_consensus_data(&block)
 
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -101,7 +101,7 @@ impl BlockMaker {
         // TODO: this isn't efficient. We have to create the header first, then see if it obeys consensus rules, then construct the full block
         let current_time = self.time_getter.get_time();
         let mut block = Block::new(
-            transactions.transactions().clone(),
+            transactions.transactions().to_vec(),
             current_tip_id.into(),
             BlockTimestamp::from_duration_since_epoch(current_time),
             common::chain::block::ConsensusData::None,
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -20,7 +20,7 @@ use chainstate_types::{BlockIndex, GetAncestorError};
 use common::{
     chain::{
         block::{timestamp::BlockTimestamp, BlockReward},
-        Block, ChainConfig, Transactions,
+        Block, ChainConfig, SignedTransaction,
     },
     primitives::{BlockHeight, Id, Idable},
     time_getter::TimeGetter,
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -25,7 +25,6 @@ use common::{
     primitives::{BlockHeight, Id, Idable},
     time_getter::TimeGetter,
 };
-use consensus::ConsensusVerificationError;
 use logging::log;
 use mempool::{
     tx_accumulator::{DefaultTxAccumulator, TransactionAccumulator},
@@ -109,13 +108,14 @@ impl BlockMaker {
             BlockReward::new(vec![]), // TODO: define consensus and rewards through NetworkUpgrades
         )?;
 
-        block = self
+        let consensus_data = self
             .chainstate_handle
             .call({
                 let chain_config = self.chain_config.clone();
                 let current_tip_height = self.current_tip_height;
+                let header = block.header().clone();
 
-                move |this| -> Result<Block, ConsensusVerificationError> {
+                move |this| {
                     let get_block_index = |&prev_block_id: &Id<Block>| {
                         this.get_block_index(&prev_block_id).map_err(|_| {
                             PropertyQueryError::PrevBlockIndexNotFound(prev_block_id.into())
@@ -137,21 +137,21 @@ impl BlockMaker {
                         })
                     };
 
-                    consensus::initialize_consensus_data(
+                    consensus::generate_consensus_data(
                         &chain_config,
-                        &mut block,
+                        &header,
                         current_tip_height,
                         get_block_index,
                         get_ancestor,
-                    )?;
-
-                    Ok(block)
+                    )
                 }
             })
             .await?
             .map_err(BlockProductionError::FailedConsensusInitialization)?;
 
-        // TODO: consensus::finalize_consensus_data(&block)
+        block.update_consensus_data(consensus_data);
+
+        // TODO: consensus::finalize_consensus_data(&block, consunsus_data)
 
         Ok(block)
     }
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -15,15 +15,17 @@
 
 use std::sync::Arc;
 
-use chainstate::ChainstateHandle;
+use chainstate::{ChainstateHandle, PropertyQueryError};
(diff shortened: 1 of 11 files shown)


consensus/src/lib.rs
```diff
@@ -15,15 +15,54 @@
...
+    pow::{calculate_work_required, check_proof_of_work, mine, ConsensusPoWError},
     validator::validate_consensus,
 };
 
-mod error;
-mod validator;
+#[allow(unreachable_code)]
+pub fn initialize_consensus_data<F, G>(
+    chain_config: &ChainConfig,
+    block: &mut Block,
```
azarovh, line 40: I think it would be better not to pass mut Block here at all. At this point it's not fully constructed which cannot be known from this function's point of view.

Rather this function should accept `height` and `prev_block_id` which are actually utilized here. And return `ConsensusData` as a result. I think it's possible but arguments of few functions need to be changed

blockprod/src/detail/block_maker.rs
```diff
@@ -92,20 +94,65 @@ impl BlockMaker {
...
             current_tip_id.into(),
             BlockTimestamp::from_duration_since_epoch(current_time),
             common::chain::block::ConsensusData::None,
             BlockReward::new(vec![]), // TODO: define consensus and rewards through NetworkUpgrades
         )?;
+
+        block = self
+            .chainstate_handle
+            .call({
+                let chain_config = self.chain_config.clone();
```
azarovh, line 115: why clone?

consensus/src/lib.rs
alfiedotwtf: The `initialize_consensus_data()` function will be used along with a new `finalize_consensus_data()` function, which is where mining will be performed.

But I guess `initialize_consensus_data()` could instead be something like `generate_consensus_data()` which returned `ConsensusData` and then `block.update_consensus_data()` could be moved to outside of the function... I just wanted to hide away the details and push them into the `consensus` crate. However I think you're right - returning `ConsensusData` is a better approach.

As for mining in `finalize_consensus_data()`, I'm still thinking this should be hidden away inside the `consensus` crate, but maybe you'll have a better approach once I push a PR soon.

blockprod/src/detail/block_maker.rs
alfiedotwtf: This one is a clone, because otherwise it will complain that it's being moved into the closure

consensus/src/lib.rs
alfiedotwtf: Ok, this turned out nice 👍

azarovh (approved):

blockprod/src/detail/block_maker.rs
TheQuantumPhysicist: Usually we use `Arc::clone(&chain_config)` to signify that this isn't a deep clone. The only ugly thing about it is that you have to import the `Arc` module if it's not already imported.

common/src/chain/transaction/mod.rs
```diff
@@ -143,6 +143,30 @@ impl Transaction {
...
+}
+
+impl Transactions {
+    pub fn new(transactions: &[SignedTransaction]) -> Transactions {
+        Transactions {
+            transactions: transactions.to_vec(),
+        }
+    }
+
+    pub fn transactions(&self) -> &Vec<SignedTransaction> {
```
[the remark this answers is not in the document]
alfiedotwtf: Ah! Of course. For some reason I always use Vec. Would be nice if there was a clippy for this

blockprod/src/detail/block_maker.rs
alfiedotwtf: Ah, ok I see now. Yep, I'll change it to follow the standard

consensus/src/lib.rs
alfiedotwtf: I had a look and I'm not sure about this one. By using BlockHeader, it's showing at the interface level what the meaning of that parameter is. Otherwise, we would need 3 individual fields to be passed in separately, and then we lose that grouping meaning. Don't know, maybe I'm wrong here?

common/src/chain/transaction/mod.rs
```diff
@@ -143,6 +143,30 @@ impl Transaction {
     }
 }
 
+pub struct Transactions {
```
TheQuantumPhysicist, line 146: What's the value of having this separate struct instead of just passing a Vec?

blockprod/src/detail/block_maker.rs
```diff
@@ -92,20 +93,66 @@ impl BlockMaker {
...
         &self,
         current_tip_id: Id<Block>,
-        accumulator: &dyn TransactionAccumulator,
+        transactions: Transactions,
     ) -> Result<Block, BlockProductionError> {
         // TODO: this isn't efficient. We have to create the header first, then see if it obeys consensus rules, then construct the full block
         let current_time = self.time_getter.get_time();
-        let block = Block::new(
-            accumulator.transactions().clone(),
+        let mut block = Block::new(
```
TheQuantumPhysicist, line 103: This can be made more efficient relatively easily, which you can make in another PR if you wish. Instead of copying everything into this block every time, you can create a new struct, say `BlockTemplate` (?), which has references to all the components. Since it's using references, you won't need to copy anything into it (except for little things that don't matter). Then, once you're done mining it, you have a method in it called `into_block(self)`, which will consume it and create a block. This way you'll copy the transactions only once (and any other very expensive part).

blockprod/src/detail/block_maker.rs
```diff
@@ -92,20 +93,66 @@ impl BlockMaker {
...
+                        get_ancestor,
+                    )
+                }
+            })
+            .await?
+            .map_err(BlockProductionError::FailedConsensusInitialization)?;
+
+        block.update_consensus_data(consensus_data);
+
+        // TODO: consensus::finalize_consensus_data(&block, consunsus_data)
```
TheQuantumPhysicist, line 154: consunsus_data

blockprod/src/detail/block_maker.rs
```diff
@@ -92,20 +93,66 @@ impl BlockMaker {
...
+                        current_tip_height,
+                        get_block_index,
+                        get_ancestor,
+                    )
+                }
+            })
+            .await?
+            .map_err(BlockProductionError::FailedConsensusInitialization)?;
+
+        block.update_consensus_data(consensus_data);
```
TheQuantumPhysicist, line 152: I'm confused... where's the mining process?

common/src/chain/transaction/mod.rs
```diff
@@ -143,6 +143,30 @@ impl Transaction {
...
+            transactions: transactions.to_vec(),
+        }
+    }
+
+    pub fn transactions(&self) -> &[SignedTransaction] {
+        &self.transactions
+    }
+}
+
+impl From<&[SignedTransaction]> for Transactions {
```
TheQuantumPhysicist, line 162: There's very little value in this `From` impl... isn't it easier to just call `to_vec()` anyway?

TheQuantumPhysicist (commented):
Done with the review. There are some things I don't understand. Please check the comments.

blockprod/src/detail/block_maker.rs
alfiedotwtf: Yep, sounds good.

blockprod/src/detail/block_maker.rs
alfiedotwtf: This PR was only going to contain the initialization of ConsensusData, and I was going to put the mining in the RPC PR, but you're right - initialization and finalization makes more sense to go together. I'll add it here too

common/src/chain/transaction/mod.rs
alfiedotwtf: I think I misunderstood this comment and ended up creating this new type https://github.com/mintlayer/mintlayer-core/pull/755#discussion_r1134585835

Yep, sounds good. I'll simplify this by passing in Vec.

TheQuantumPhysicist: The idea from that comment is that there are two ways where we need to pass transactions to be included in a block:

1. Through an accumulator from the mempool
2. Through RPC, where we provide a Vec of transactions

Both should be one parameter unified through an enum, where choice one is used in normal operations, and choice two is used in RPC

common/src/chain/transaction/mod.rs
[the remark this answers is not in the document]
alfiedotwtf: I've gotten rid of the Transaction abstraction, so this comment can be resolved.

common/src/chain/transaction/mod.rs
alfiedotwtf: While I was reverting the `Transactions` struct, it wouldn't compile because I've made `make_block()` async which now makes `accumulator` possibly outlive the function. The reason for making it async is that `make_block()` will now contain all the mining work (via `finalize_consensus_data()`) by both methods (mempool and RPC), and can soon be taken out of the Tokio executor (not implemented yet) so that it doesn't hold up the rest of the system.

alfiedotwtf: Also, to simplify things further, I've stripped away the `Transaction` and the need for any enum by going straight for an array slice:

```
    pub async fn make_block(                  
        &self,                                
        current_tip_id: Id<Block>,            
        transactions: &[SignedTransaction],   
    ) -> Result<Block, BlockProductionError> {
```

blockprod/src/detail/block_maker.rs
alfiedotwtf: Mining has now been added in this PR via `finalize_consensus_data()`.

blockprod/src/detail/block_maker.rs
alfiedotwtf: I'm not sure this is needed anymore. As `finalize_consensus_data()` has been added into `make_block()`, the result of the function should be a finalized block. So we shouldn't be need to try again any longer so any copying will be done only once.

common/src/chain/transaction/mod.rs
alfiedotwtf: Ok, this ended up not working once I added the RPC call because the created block refers to the transactions that live longer. I'm going to commit my current rebased RPC changes to #755 and come back to this later - I don't want to be bogged down be one issue.

consensus/src/lib.rs
```diff
@@ -15,15 +15,76 @@
...
+        RequiredConsensus::IgnoreConsensus => Ok(()),
+        RequiredConsensus::DSA | RequiredConsensus::PoS => unimplemented!(),
+        RequiredConsensus::PoW(_) => {
+            match block.consensus_data() {
+                ConsensusData::None => Ok(()),
+                ConsensusData::PoS(_) => unimplemented!(),
+                ConsensusData::PoW(pow_data) => {
+                    // TODO: replace the static value below with one
+                    // from BlockProductionConfig
+                    mine(block, 0xffffffff, pow_data.bits())
```
TheQuantumPhysicist, line 82: Why not u128::MAX for max nonce?

blockprod/src/detail/block_maker.rs
```diff
@@ -92,20 +93,70 @@ impl BlockMaker {
...
+
+        consensus::finalize_consensus_data(
+            &self.chain_config,
+            &mut block,
+            self.current_tip_height,
```
TheQuantumPhysicist, line 157: It makes more sense to pass the new block height instead of incrementing it inside.

consensus/src/lib.rs
alfiedotwtf: I just put a number there while developing it, but thought we'd want the user to configure it. But sure, we can do this too, but if MAX, should we instead just turn it into an endless loop?

blockprod/src/detail/block_maker.rs
alfiedotwtf: Aha, gotcha.

consensus/src/lib.rs
TheQuantumPhysicist: This shouldn't really be configurable. Just the MAX value is more than enough.

blockprod/src/detail/block_maker.rs
```diff
@@ -92,20 +93,70 @@ impl BlockMaker {
         Ok(returned_accumulator)
     }
 
-    pub fn make_block(
+    pub async fn make_block(
         &self,
         current_tip_id: Id<Block>,
-        accumulator: &dyn TransactionAccumulator,
+        transactions: &[SignedTransaction],
```
TheQuantumPhysicist, line 99: Rule of thumb: If you pass a reference then have to clone it, then make the parameter by value. This way you give the option to the caller to clone it consume it.

alfiedotwtf:
I've rebased #755 onto this PR, so I'm going to close this PR and move the latest changes needed there.

blockprod/src/detail/block_maker.rs
```diff
@@ -147,10 +198,11 @@ impl BlockMaker {
     /// 2. A new tip is now on chainstate, indicating that there's no point in continuing to mine/stake at that tip
     pub async fn run(&mut self) -> Result<(), BlockProductionError> {
         let accumulator = self.collect_transactions().await?;
+        let transactions = accumulator.transactions();
 
         // TODO: do we want to introduce a separate executor for this loop to avoid starving other tasks?
         loop {
-            let block = self.make_block(self.current_tip_id, &*accumulator)?;
+            let block = self.make_block(self.current_tip_id, transactions).await?;
```
TheQuantumPhysicist, line 205: Keep in mind that the way it's done now, mining will *never* exit unless mining is finished, which is practically impossible. So cancelling won't work.

Notice that the cancel function works assuming that the make_block function doesn't really take a long time to finish.

TheQuantumPhysicist (approved):
Besides the comments, looking good. The cancelling problem has to be addressed sooner or later. Please at least write a TODO about it. Also please check the other comments.

blockprod/src/detail/block_maker.rs
alfiedotwtf: Ah, yes this was why I got confused about u128::MAX. You're right... I'll have to think about a nice way to solve this.

PR #791, opened 2023-03-28
Improve transactions relaying

stanislav-tkach:


mocks/src/mempool.rs
```diff
@@ -99,6 +99,13 @@ impl MempoolInterface for MempoolInterfaceMock {
         }
     }
 
+    async fn transaction(
+        &self,
+        _id: &Id<Transaction>,
+    ) -> Result<Option<&SignedTransaction>, Error> {
+        Ok(None)
```
TheQuantumPhysicist, line 106: I'm not sure why this impl for the mock exists... but this doesn't look right.

stanislav-tkach: As far as I can see, this mock is only used inside the mempool tests, so for now a proper implementation isn't needed. I have added the `todo!` macro here.

stanislav-tkach:
I have added a few more checks and additional tests. As far as I can see, only one improvement needs to be done: the timer should be added to punish the peers that don't respond to the transaction requests.

mocks/src/mempool.rs
```diff
@@ -99,6 +99,10 @@ impl MempoolInterface for MempoolInterfaceMock {
         }
     }
 
+    async fn transaction(&self, _id: &Id<Transaction>) -> Result<Option<SignedTransaction>, Error> {
+        todo!("Add tests")
```
TheQuantumPhysicist, line 103: I think CI will complain about this. Use `unimplemented!()`

p2p/src/sync/peer.rs
```diff
@@ -369,6 +383,38 @@ where
...
+                .send_message(self.id(), SyncMessage::TransactionResponse(tx))?;
+        }
+
+        Ok(())
+    }
+
+    async fn handle_transaction_response(&mut self, tx: SignedTransaction) -> Result<()> {
+        let id = tx.serialized_hash();
+
+        if self.announced_transactions.take(&id).is_none() {
```
TheQuantumPhysicist, line 408: Isn't it better to implement this as a push/pop queue instead of using this unnecessary invariant?

TheQuantumPhysicist (approved):

p2p/src/sync/peer.rs
stanislav-tkach: I'm not sure. What if we receive responses in a different order or don't receive one of them at all? With the current implementation we will still be able to obtain other transactions.

TheQuantumPhysicist: Receiving responses in a different order should have zero bearing on the final outcome. We cannot guarantee that anyway due to both multithreading and network latency.

pavel-kokolemin (approved):

PR #794, opened 2023-03-29
Add utxos to signature hash to allow not having full transactions when verifying before signing

### LINKED ISSUE #303: Add output data of inputs in signature hash
We have to consider using the taproot way of signature hash and add the output data of inputs in the commitment. Discussion [here](https://bitcoin.stackexchange.com/questions/113782/segwit-includes-the-input-amount-in-the-signaturehash-what-possible-attack-can).

TheQuantumPhysicist:
When creating transactions, we usually need to download the full transactions of the inputs to verify the information provided by the signing software. This becomes an issue when internet connection isn't available or is weak, because transactions can be very large (like those coming from exchanges with way too many outputs). A solution to this problem is including all utxos of the current transaction's inputs in signature hash so that the wallet/software in question can't lie about the amounts claimed to be spent.

Closes #303

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -492,6 +492,17 @@ where
             None => return Ok(()),
         };
 
+        let inputs_utxos = inputs
+            .iter()
+            .map(|input| {
+                let outpoint = input.outpoint();
+                self.utxo_cache
+                    .utxo(outpoint)
+                    .ok_or(ConnectTransactionError::MissingOutputOrSpent)
+                    .map(|utxo| utxo.take_output())
+            })
+            .collect::<Result<Vec<_>, ConnectTransactionError>>()?;
+
         for (input_idx, input) in inputs.iter().enumerate() {
             let outpoint = input.outpoint();
             let utxo = self
@@ -502,8 +513,10 @@ where
             // TODO: see if a different treatment should be done for different output purposes
             // TODO: ensure that signature verification is tested in the test-suite, they seem to be tested only internally
             match utxo.output().purpose().destination() {
-                Some(d) => verify_signature(self.chain_config.as_ref(), d, tx, input_idx)
-                    .map_err(ConnectTransactionError::SignatureVerificationFailed)?,
+                Some(d) => {
+                    verify_signature(self.chain_config.as_ref(), d, tx, &inputs_utxos, input_idx)
+                        .map_err(ConnectTransactionError::SignatureVerificationFailed)?
+                }
                 None => return Err(ConnectTransactionError::AttemptToSpendBurnedAmount),
             }
         }
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -513,10 +513,14 @@ where
             // TODO: see if a different treatment should be done for different output purposes
             // TODO: ensure that signature verification is tested in the test-suite, they seem to be tested only internally
             match utxo.output().purpose().destination() {
-                Some(d) => {
-                    verify_signature(self.chain_config.as_ref(), d, tx, &inputs_utxos, input_idx)
-                        .map_err(ConnectTransactionError::SignatureVerificationFailed)?
-                }
+                Some(d) => verify_signature(
+                    self.chain_config.as_ref(),
+                    d,
+                    tx,
+                    &inputs_utxos.iter().collect::<Vec<_>>(),
+                    input_idx,
+                )
+                .map_err(ConnectTransactionError::SignatureVerificationFailed)?,
                 None => return Err(ConnectTransactionError::AttemptToSpendBurnedAmount),
             }
         }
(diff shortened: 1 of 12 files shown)


pavel-kokolemin (approved):

common/src/chain/transaction/signature/mod.rs
```diff
@@ -266,6 +268,7 @@ impl Transactable for SignedTransaction {
 
 fn stream_signature_hash<T: Signable>(
     tx: &T,
+    inputs_utxos: &[&TxOutput],
```
azarovh, line 271: wouldn't `&[TxOutput]` be simpler to work with?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -502,8 +513,14 @@ where
             // TODO: see if a different treatment should be done for different output purposes
             // TODO: ensure that signature verification is tested in the test-suite, they seem to be tested only internally
             match utxo.output().purpose().destination() {
-                Some(d) => verify_signature(self.chain_config.as_ref(), d, tx, input_idx)
-                    .map_err(ConnectTransactionError::SignatureVerificationFailed)?,
+                Some(d) => verify_signature(
+                    self.chain_config.as_ref(),
+                    d,
+                    tx,
+                    &inputs_utxos.iter().collect::<Vec<_>>(),
```
azarovh, line 520: then here simply `&inputs` can be passed

common/src/chain/transaction/signature/mod.rs
```diff
@@ -294,8 +297,15 @@ fn stream_signature_hash<T: Signable>(
...
 
-    // TODO: consider doing just like taproot, and hash in all outputs that come from the outpoints of inputs,
-    //       this would be a good solution to avoid having to download full transactions to verify inputs
+    // Include utxos of the inputs to make it possible to verify the inputs scripts and amounts without downloading the full transactions
+    if inputs.len() != inputs_utxos.len() {
+        return Err(TransactionSigError::InvalidUtxoCountVsInputs(
+            inputs_utxos.len(),
+            inputs.len(),
+        ));
+    } else {
```
azarovh, line 306: minor but else is unnecessary. Also, do we prefer `ensure!` for such situations?

azarovh (approved):

common/src/chain/transaction/signature/mod.rs
TheQuantumPhysicist: Yes, but this removes the necessity to copy values, since these outputs aren't generally in one Vec but from multiple transactions.

azarovh: ahh, I see. It's just in all places these outputs are already collected. But ok

pavel-kokolemin: `TxOutput` is only 24 bytes, I don't think it's worth the troubles to collect references here (and we would need to keep those transactions/UTXO somewhere).

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: That won't work. This is a Vec of values. We want a slice of refs.

azarovh: I mean that would work in combination with my prev comment

common/src/chain/transaction/signature/mod.rs
TheQuantumPhysicist: Done

common/src/chain/transaction/signature/mod.rs
TheQuantumPhysicist: The reason why I preferred to do this, besides the size issue, which is minor for this single function, is that this is called through a series of functions and all the other parameters are by reference. I didn't want to break that pattern.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: It was like that before. I changed it to slice of refs. I'll resolve this and move that discussion there.

common/src/chain/transaction/signature/mod.rs
```diff
@@ -294,8 +297,15 @@ fn stream_signature_hash<T: Signable>(
...
+        hash_encoded_to(&inputs_utxos, stream);
```
iljakuklic, line 307: I don't think this will play nicely with the likes of `SIGHASH_ANYONECANPAY`. This basically always commits to all inputs, regardless of the flag.

From wiki:
> The SIGHASH_ANYONECANPAY modifier can be combined with the above three modes. When set, only that input is signed and the other inputs can be anything.

Maybe the spent output commitment could be a part of tx input sighash calculation, so it shares the ANYONECANPAY logic + you get utxo being spent and the corresponding input nicely next to each other.

iljakuklic (commented):
Please address the comment I made below, potentially fairly important. Looks good otherwise.

common/src/chain/transaction/signature/mod.rs
TheQuantumPhysicist: This isn't compatible with anyonecanpay. Neither was the previous version. That'll will be done later.

PR #795, opened 2023-03-29
Enable scale code as a feature in merkletree crate

TheQuantumPhysicist:
Another step to move the crate outside of mintlayer-core

### DIFF
--- a/merkletree/Cargo.toml
+++ b/merkletree/Cargo.toml
@@ -6,8 +6,7 @@ edition.workspace = true
 rust-version.workspace = true
 
 [features]
-scale-codec-encode = ["dep:parity-scale-codec"]
-scale-codec-decode = ["dep:parity-scale-codec"]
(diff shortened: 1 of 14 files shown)


merkletree/Cargo.toml
```diff
@@ -5,9 +5,14 @@ version.workspace = true
 edition.workspace = true
 rust-version.workspace = true
 
+[features]
+scale-codec-encode = ["dep:parity-scale-codec"]
+scale-codec-decode = ["dep:parity-scale-codec"]
```
iljakuklic, line 10: It's maybe a bit too granular to have this as two separate features. Either one will pull in the dependency, and enabling both does not add any new deps. When crated optionally depend on `serde`, both encoding and decoding are guarded by single feature. Do you have a specific reason to do it like this?

iljakuklic (approved):

stanislav-tkach (approved):

PR #796, opened 2023-03-30
Pool decommissioning

azarovh:
Added new `TxOutput::DecommissionPool` type and the code that supports decommissioning.

The main challenge for this PR was to handle new rules for input/output combinations in `input_output_policy.rs`. Please pay attention there. I think I've covered all of them, but the tests still don't prove there are no false positives. If you know how to improve the testing please let me know.


chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        Ok((pool_id, undo))
+    }
+
+    pub fn decommission_pool(
+        &mut self,
+        tx_source: TransactionSource,
+        pool_id: PoolId,
+    ) -> Result<PoSAccountingUndo, ConnectTransactionError> {
+        let mut temp_delta = PoSAccountingDelta::new(&self.accounting_delta);
```
TheQuantumPhysicist, line 99: Again, please don't use `temp` in variable names. All stack variables, by definition, are temporary.

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -139,5 +138,40 @@ pub fn check_timelocks<
         }
     }
 
+    let outputs = match tx.outputs() {
+        Some(outputs) => outputs,
+        None => return Ok(()),
```
TheQuantumPhysicist, line 143: This is really dangerous and wrong. Here's why:

This function isn't only for this match and relevant checks and doesn't strictly depend on it. The problem here is with the following scenario: Imagine someone in the future adding more functionality after the for loop below for something other than the outputs. What's gonna happen is basically that his functionality will be limited by whether the outputs are Some or None. This is certainly not the desired behavior by default.

The correct way to do this should not limit the contiuation of the function UNLESS the whole function is about this check. Meaning: You can either create another function that will do the check AND run the loop and return if unnecessary (like you did in this match), or you limit what you did here to a smaller control block that doesn't affect the whole function in case someone adds more things to be done in the future.

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
```diff
@@ -0,0 +1,166 @@
...
+use common::{
+    chain::{stakelock::StakePoolData, OutPoint, PoolId},
+    primitives::Amount,
+};
+use pos_accounting::{
+    DeltaMergeUndo, FlushablePoSAccountingView, PoSAccountingDelta, PoSAccountingDeltaData,
+    PoSAccountingOperations, PoSAccountingUndo, PoSAccountingView,
+};
+
+/// Adapter over `PosAccountingDelta` that mimics `PoSAccountingOperations`.
```
TheQuantumPhysicist, line 30: Why not just implement that trait here?

azarovh: I need an additional argument `TransactionSource` for every function to accumulate deltas for a specific block.

TheQuantumPhysicist: You can put it in the constructor of this struct

azarovh: Won't work. Nothing prevents the verifier to be used on multiple blocks, so I just pass it on every call

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,773 @@
...
-    utils::ensure!(
-        can_be_spent,
-        ConnectTransactionError::AttemptToSpendInvalidOutputType
-    );
     Ok(())
 }
 
-fn check_outputs_can_be_used_in_tx(tx: &Transaction) -> Result<(), ConnectTransactionError> {
-    let are_outputs_valid = tx.outputs().iter().all(is_valid_output_for_tx);
+#[rustfmt::skip]
```
TheQuantumPhysicist, line 168: If you want this to keep the OR lines at the end, I think it's a rule we can add to rustfmt

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
```diff
@@ -0,0 +1,166 @@
...
+        PoSAccountingDeltaData,
+        BTreeMap<TransactionSource, PoSAccountingDeltaData>,
+    ) {
+        (
+            self.accounting_delta.consume(),
+            self.accounting_block_deltas,
+        )
+    }
+
+    pub fn get_accounting_delta(&self) -> &PoSAccountingDelta<P> {
```
TheQuantumPhysicist, line 61: accounting_delta()

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
TheQuantumPhysicist: You can still create an intermediary object that refers to the delta that holds the tx_source.

Meaning: PoSAccountingDeltaAdapter doesn't need to be owned by tx-verifier, but it can be created on demand with a reference to the delta object + contain the tx-source (which is light-weight). This way, you can still implement the trait to ensure that it respects the rules of that trait.

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
```diff
@@ -0,0 +1,166 @@
...
+            pool_data.margin_ratio_per_thousand(),
+            pool_data.cost_per_epoch(),
+        )?;
+
+        let new_delta_data = temp_delta.consume();
+        self.accounting_delta.merge_with_delta(new_delta_data.clone())?;
+        self.accounting_block_deltas
+            .entry(tx_source)
+            .or_default()
+            .merge_with_delta(new_delta_data)?;
```
TheQuantumPhysicist, line 89: If this is repeated all over the place, please take it out to a single function.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,773 @@
...
-}
-
-fn check_inputs_can_be_used_in_tx(
+/// Not all `TxOutput` combinations can be used in a transaction.
+pub fn check_tx_inputs_outputs_purposes(
     tx: &Transaction,
     utxo_view: &impl utxo::UtxosView,
 ) -> Result<(), ConnectTransactionError> {
-    let can_be_spent = tx
+    let inputs = tx
```
TheQuantumPhysicist, line 116: Please call this inputs_utxos

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,773 @@
...
-        .collect::<Result<Vec<_>, _>>()?
-        .iter()
-        .all(|utxo| is_valid_input_for_tx(utxo.output()));
+        .collect::<Result<Vec<_>, _>>()?;
+
+    match inputs.as_slice() {
+        // no inputs
+        [] => return Err(ConnectTransactionError::MissingTxInputs),
+        // single input
+        [input] => match tx.outputs() {
```
TheQuantumPhysicist, line 132: What I don't understand here is... what's the value of distinguishing between one input or many? There's no special circumstance where one input is the right way to do things (except for staking). In fact, most of the time the number of inputs is irrelevant.

But then let's assume this is the right way to do it... this just means there has to be (by definition), duplicate code, which is what I'm seeing at the bottom of both arms of one or more inputs... well, why aren't these taking outside to a private function?

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
azarovh: Wanted this adapter to encapsulate the state as well so that there was no way to modify deltas except through this adapter. Which means that it has to hold the state. 
I could split this type in two: one that stores the state with a method that creates a lightweight accessor with references and implements the trait.

TheQuantumPhysicist:
Looking good. A few issues should be addressed, then I'll look into it again. Cheers!

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
azarovh: The rules for one and many are different. For example, it's ok to have `StakePool` -> `DecommissionPool`, but it's not ok to have `StakePool` + `Transfer` -> `DecommissionPool`.
So I come up with a solution to explicitly distinguish the cases.

Regarding duplications I see one place that can be optimised, will do

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
TheQuantumPhysicist: Having something that implements the trait is a necessity to maintain consistency. Please give it a shot.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
azarovh: good point, added a rule that allows aligning `|` in match arms:
https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#match_arm_leading_pipes

Besides, I wanted the pair to fit in a single line, but since you removed OutputPurpose it's ok now.

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -139,5 +138,38 @@ pub fn check_timelocks<
         }
     }
 
+    if let Some(outputs) = tx.outputs() {
```
TheQuantumPhysicist, line 141: Didn't we discuss this and agree that this should be done in check_block because the history is not required?

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,778 @@
...
-        can_be_spent,
-        ConnectTransactionError::AttemptToSpendInvalidOutputType
-    );
     Ok(())
 }
 
-fn check_outputs_can_be_used_in_tx(tx: &Transaction) -> Result<(), ConnectTransactionError> {
-    let are_outputs_valid = tx.outputs().iter().all(is_valid_output_for_tx);
+#[allow(clippy::unnested_or_patterns)]
+fn is_valid_one_to_one_combination(input: &TxOutput, output: &TxOutput) -> bool {
```
TheQuantumPhysicist, line 157: input -> input_utxo

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,778 @@
...
+    let valid_outputs = are_outputs_valid_for_tx(outputs);
+    ensure!(
+        valid_outputs,
+        ConnectTransactionError::InvalidOutputTypeInTx
     );
     Ok(())
 }
 
-// TODO(Gosha): add tests once decommissioning is supported, because the rules will significantly change
+fn are_inputs_valid_for_tx(inputs: &[TxOutput]) -> bool {
```
TheQuantumPhysicist, line 211: inputs_utxos

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,778 @@
...
+    #[case(produce_block(), decommission_pool(),  Ok(()))]
+    /*-----------------------------------------------------------------------------------------------*/
+    #[case(decommission_pool(), transfer(),           Ok(()))]
+    #[case(decommission_pool(), burn(),               Ok(()))]
+    #[case(decommission_pool(), lock_then_transfer(), Ok(()))]
+    #[case(decommission_pool(), stake_pool(),         Ok(()))]
+    #[case(decommission_pool(), produce_block(),      Err(ConnectTransactionError::InvalidOutputTypeInTx))]
+    #[case(decommission_pool(), decommission_pool(),  Err(ConnectTransactionError::InvalidOutputTypeInTx))]
+    fn tx_one_to_one(
+        #[case] input: TxOutput,
```
TheQuantumPhysicist, line 440: input_utxo

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,778 @@
...
+    #[case(produce_block(), decommission_pool(),  Err(ConnectTransactionError::InvalidOutputTypeInReward))]
+    /*-----------------------------------------------------------------------------------------------*/
+    #[case(decommission_pool(), transfer(),           Err(ConnectTransactionError::InvalidInputTypeInReward))]
+    #[case(decommission_pool(), burn(),               Err(ConnectTransactionError::InvalidInputTypeInReward))]
+    #[case(decommission_pool(), lock_then_transfer(), Err(ConnectTransactionError::InvalidInputTypeInReward))]
+    #[case(decommission_pool(), stake_pool(),         Err(ConnectTransactionError::InvalidInputTypeInReward))]
+    #[case(decommission_pool(), produce_block(),      Err(ConnectTransactionError::InvalidInputTypeInReward))]
+    #[case(decommission_pool(), decommission_pool(),  Err(ConnectTransactionError::InvalidInputTypeInReward))]
+    fn reward_one_to_one(
+        #[case] input: TxOutput,
```
TheQuantumPhysicist, line 662: input_utxo

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,778 @@
...
+
+        let tx = Transaction::new(0, vec![outpoint.into()], vec![output], 0).unwrap();
+        assert_eq!(result, check_tx_inputs_outputs_purposes(&tx, &utxo_db));
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn tx_one_to_many(#[case] seed: Seed) {
+        let check = |source_inputs: &[TxOutput], source_outputs: &[TxOutput], expected_result| {
```
TheQuantumPhysicist, line 458: inputs_utxos

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn undo(&mut self, undo: PoSAccountingUndo) -> Result<(), pos_accounting::Error> {
+        let mut delta = PoSAccountingDelta::new(&self.adapter.accounting_delta);
+
+        delta.undo(undo)?;
+
+        self.merge_delta(delta.consume())
+    }
+}
```
TheQuantumPhysicist, line 182: TODO: unit tests?

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,778 @@
...
-    );
     Ok(())
 }
 
-fn check_outputs_can_be_used_in_tx(tx: &Transaction) -> Result<(), ConnectTransactionError> {
-    let are_outputs_valid = tx.outputs().iter().all(is_valid_output_for_tx);
+#[allow(clippy::unnested_or_patterns)]
+fn is_valid_one_to_one_combination(input: &TxOutput, output: &TxOutput) -> bool {
+    match (input, output) {
+        | (TxOutput::Transfer(_, _), TxOutput::Transfer(_, _))
```
TheQuantumPhysicist, line 159: So the leading OR has no effect? Wow this is very confusing!

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
azarovh: right, sorry totally forgot that

azarovh: ok so I moved it to check_block, but it requires me to provide the function with current block height. Because maturity setting is taken from netupgrade

chainstate/src/detail/mod.rs
```diff
@@ -430,7 +435,17 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+        let best_block_id = chainstate_ref
+            .get_best_block_id()
+            .map_err(BlockError::BestBlockLoadError)
+            .log_err()?;
+        let best_block_height = chainstate_ref
+            .get_block_height_in_main_chain(&best_block_id)
+            .map_err(BlockError::BestBlockLoadError)
+            .log_err()?
+            .expect("best block height doesn't exist");
+        chainstate_ref.check_block(&block, best_block_height.next_height()).log_err()?;
```
TheQuantumPhysicist, line 448: `best_block_height.next_height()` here is wrong. This implies that the current best block is always the previous block of the block you're checking. 

You should get the block height of the previous block, then do `next_height()` on that. Not the best block.

chainstate/src/detail/mod.rs
```diff
@@ -285,9 +285,14 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
                 .log_err()?;
+            let best_block_height = chainstate_ref
+                .get_block_height_in_main_chain(&best_block_id)
+                .map_err(BlockError::BestBlockLoadError)
+                .log_err()?
+                .expect("best block height doesn't exist");
 
             chainstate_ref
-                .check_block(&block)
+                .check_block(&block, best_block_height.next_height())
```
TheQuantumPhysicist, line 295: Same here. This is wrong.

chainstate/src/detail/mod.rs
TheQuantumPhysicist: A good question is: How didn't the tests catch this mistake?!

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
TheQuantumPhysicist: That's OK.

chainstate/src/detail/mod.rs
azarovh: > A good question is: How didn't the tests catch this mistake?!

It requires a very specific test to catch. But at the moment we support only producing PoS blocks from the tip. I'll add the test but leave it ignored for now.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -552,19 +555,68 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+                match output {
+                    TxOutput::Transfer(_, _)
+                    | TxOutput::Burn(_)
+                    | TxOutput::StakePool(_)
+                    | TxOutput::ProduceBlockFromStake(_, _, _)
+                    | TxOutput::LockThenTransfer(_, _, _) => {}
+                    TxOutput::DecommissionPool(_, _, _, timelock) => match timelock {
+                        OutputTimeLock::ForBlockCount(c) => {
+                            let cs: i64 = (*c).try_into().map_err(|_| {
+                                CheckBlockTransactionsError::InvalidDecommissionMaturityDistanceValue
```
TheQuantumPhysicist, line 574: Please write the error in the same format as InvalidBlockRewardMaturityDistance. Always convey as much information as possible.

TheQuantumPhysicist (approved):

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -13,75 +13,778 @@
...
+                ))?
+                .iter()
+                .all(|output| match output {
+                    TxOutput::LockThenTransfer(_, _, _) => true,
+                    TxOutput::Transfer(_, _)
+                    | TxOutput::Burn(_)
+                    | TxOutput::StakePool(_)
+                    | TxOutput::ProduceBlockFromStake(_, _, _)
+                    | TxOutput::DecommissionPool(_, _, _, _) => false,
+                });
```
iljakuklic, line 101: Are there supposed to be any restrictions on the time lock duration?

iljakuklic (approved):

chainstate/src/detail/mod.rs
```diff
@@ -286,8 +286,14 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
                 .map_err(BlockError::BestBlockLoadError)
                 .log_err()?;
 
+            let prev_block_height = chainstate_ref
+                .get_block_height_in_main_chain(&block.prev_block_id())
+                .map_err(BlockError::BlockLoadError)
+                .log_err()?
+                .ok_or(BlockError::PrevBlockNotFound)?;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I actually have been thinking about this for a while in addition to signing blocks. That's coming later.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
azarovh: Output timelocks are checked in `ChainstateRef::check_outputs_timelock`

chainstate/tx-verifier/src/transaction_verifier/accounting_delta_adapter.rs
```diff
@@ -0,0 +1,183 @@
...
+        unimplemented!()
+    }
+
+    fn spend_share_from_delegation_id(
+        &mut self,
+        _delegation_id: common::chain::DelegationId,
+        _amount: Amount,
+    ) -> Result<PoSAccountingUndo, pos_accounting::Error> {
+        unimplemented!()
+    }
```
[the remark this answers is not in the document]
azarovh: I'll need these methods in the following PR for delegation. So I decided just to implement it.

PR #798, opened 2023-03-31
Merge OutputPurpose with TxOutput to make output value part of the arms in preparation for more complex output types

TheQuantumPhysicist:
We don't always need value to be present for every output type. Hence, it makes more sense to make it part of the enum instead of existent in every type.


alfiedotwtf (approved):
That's a nice simplification.

common/src/chain/transaction/output/stakelock.rs
```diff
@@ -33,13 +34,15 @@ pub struct StakePoolData {
 
 impl StakePoolData {
     pub fn new(
+        value: Amount,
```
iljakuklic, line 37: Maybe there should be a type distinction between Mlt coins and other token amounts. There is already a `Mlt` type somewhere in chain config. We can then distinguish which output types accept arbitrary tokens vs Mlts only more granulary.

iljakuklic (approved):

common/src/chain/transaction/output/stakelock.rs
TheQuantumPhysicist: This is exactly what's done in this PR. For staking, you can only stake MLT. Hence, it's an amount. For other outputs, it's OutputValue enum, so can be anything.

stanislav-tkach (approved):

chainstate/test-suite/src/tests/stake_pool_tests.rs
```diff
@@ -241,80 +231,6 @@ fn stake_pool_and_transfer_tokens_same_tx(#[case] seed: Seed) {
...
-            ))
-            .build();
-        let tx1_id = tx1.transaction().get_id();
-
-        let result = tf.make_block_builder().with_transactions(vec![tx0, tx1]).build_and_process();
-
-        assert_eq!(
-            result.unwrap_err(),
-            ChainstateError::ProcessBlockError(BlockError::StateUpdateFailed(
-                ConnectTransactionError::TokenOutputInPoSAccountingOperation(tx1_id)
```
azarovh, line 312: I think this error is obsolete now

common/src/chain/transaction/output/mod.rs
```diff
@@ -41,71 +41,64 @@ pub enum Destination {
...
     /// Output type that represents spending of a stake pool output in a block reward
     /// in order to produce a block
     #[codec(index = 4)]
-    ProduceBlockFromStake(Destination, PoolId),
+    ProduceBlockFromStake(Amount, Destination, PoolId),
 }
 
-impl OutputPurpose {
+impl TxOutput {
+    // TODO: this has to go
```
azarovh, line 62: could you explain why?

common/src/chain/transaction/signature/tests/sign_and_mutate.rs
```diff
@@ -1080,17 +1080,23 @@ fn add_output(_rng: &mut impl Rng, tx: &SignedTransactionWithUtxo) -> SignedTran
...
-    updater.outputs[0] = TxOutput::new(
-        match updater.outputs[0].value() {
-            OutputValue::Coin(coin) => {
-                OutputValue::Coin((*coin + Amount::from_atoms(100)).unwrap())
-            }
-            OutputValue::Token(asset) => OutputValue::Token(asset.clone()),
-        },
-        updater.outputs[0].purpose().clone(),
-    );
+    // Should failed due to change in output value
```
azarovh, line 1092: "Should fail"

azarovh (approved):

common/src/chain/transaction/output/stakelock.rs
iljakuklic: Well `Amount` is also used in `OutputValue` to represent token amount (see `TokenTransfer`), so the separation is not as tight as it could be

chainstate/test-suite/src/tests/stake_pool_tests.rs
TheQuantumPhysicist: Actually this wouldn't work anyway due to accounting, before or now. I'll remove the whole test.

TheQuantumPhysicist: You were talking specifically about the error, I'm guessing. I just noticed that I removed the test earlier.

common/src/chain/transaction/output/mod.rs
TheQuantumPhysicist: Because it makes assumptions about destination. A destination (and whether it exists, and whether there's more than one) depends on the type of the output. The correct way it to match this when needed, not make assumptions about what destination means, which can easily lead to misinterpretations.

common/src/chain/transaction/output/stakelock.rs
TheQuantumPhysicist: This is, again, the same issue we're having with timestamps, where they're also just numbers that can't be interpreted without assumptions when coming from bytes. Unfortunately, it's an unsolvable problem. A number is just a number as long as it touches bytes. Without context, we can't deserialize it into a type without making an assumption. This is why an `Amount` is just an amount until it gets into an output. Only then, we can make an assumption on what that `Amount` really means; i.e., whether it's a coin or a token.

I'm not saying it's impossible to find a trick to do all this... but it's, at least, not so obvious that it can be done given the current constraints.

TheQuantumPhysicist: To be clear about the `Mlt` type, it's not a good choice to involve the blockchain operation with how many decimals we're using. This is the job of the wallets and user-touching stuff. It doesn't really add value unless we involve the context of ChainConfig.

TheQuantumPhysicist: You can't really benefit much from this, and here's why: Amount, itself, is already type-safe (and cannot be added to timestamps, for example). However, we can have any number of tokens defined at runtime, and they're not distinguishable at compile-time. So, the best we can do is distinguish between ML and other tokens. But tokens among each other will still have one amount type at best.

Alrighty. How about adding a generic parameter to Amount to make that distinction? We can make a default generic type, and specialize it for Amount where appropriate. Sounds good? This will have the added benefit that all math ops will remain valid, instead of having to reintroduce them, which to be honest is a big problem with extra types on things that support math.

iljakuklic: > Alrighty. How about adding a generic parameter to Amount to make that distinction? We can make a default generic type tag, and specialize it for ML where appropriate. Sounds good? This will have the added benefit that all math ops will remain valid, instead of having to reintroduce them, which to be honest is a big problem with extra types on things that support math.

That sounds reasonable.

Regarding tokens not being distinguishible from each other at compile time, I think it can be partially mitigated by a type that enforces the check at runtime. I.e. instead of having plain amount, you have amount + token ID. Arithmetic operations and similar check you are adding same kind of tokens, and return a `Result` as appropriate. Also such type would probably implement `PartialEq`, `PartialOrd` but not `Eq`, `Ord`.

PR #800, opened 2023-03-31
Fix anyonecanpay in sighash and other single input commitments

### LINKED ISSUE #171: Deal with `OP_CODESEPARATOR`
Have to decide on behaviour of the instruction in Scripts. Most likely just disable it.

TheQuantumPhysicist:
We also remove OP_CODESEPARATOR todo comment

Closes #171


common/src/chain/transaction/signature/sighash/hashable.rs
```diff
@@ -0,0 +1,242 @@
...
+                    let inputs_utxos = self.inputs_utxos;
+                    hash_encoded_to(&(inputs_utxos.len() as u32), stream);
+                    for input_utxo in inputs_utxos {
+                        hash_encoded_to(&input_utxo, stream);
+                    }
+                }
+            }
+            sighashtype::InputsMode::AnyoneCanPay => {
+                hash_encoded_to(&target_input.outpoint(), stream);
+            }
```
iljakuklic, line 123: Not adding the spent UTXO in this case?

common/src/chain/transaction/signature/sighash/hashable.rs
```diff
@@ -0,0 +1,242 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct SignatureHashableInputs<'a> {
+    inputs: &'a [TxInput],
+    /// Include utxos of the inputs to make it possible to verify the inputs scripts and amounts without downloading the full transactions
+    inputs_utxos: &'a [&'a TxOutput],
+}
```
iljakuklic, line 63: Not sure if it's an improvement but using `&[(TxInput, &TxOutput)]` enforces the same number of inputs and outputs by construction. Enforcing that in the constructor as done now is fine too.

iljakuklic (approved):
Cool stuff

common/src/chain/transaction/signature/sighash/hashable.rs
TheQuantumPhysicist: Good question. I'm still not sure about this, to be honest. The main complaint about this problem (that required adding the utxo) was nullified in segwit with an attack that made it possible to manipulate the user due to the presence of two outputs with similar amounts (and possibly destinations). So the question is, would it even help here? I don't know.

common/src/chain/transaction/signature/sighash/hashable.rs
TheQuantumPhysicist: Your solution is nicer, but the only issue with it is that usually inputs come in a Vec straight from a transaction, which will mean that creating this Vec/Slice of tuples will require rearranging the inputs every time we need a sighash, since it's not natural to have this format of an array.

common/src/chain/transaction/signature/sighash/hashable.rs
TheQuantumPhysicist: Seems like bitcoin does it, so I'll do it too. Nothing to lose here.

https://github.com/bitcoin/bitcoin/blob/a0d37d1d230b837ebb0ec8f609885ab89280936c/src/script/interpreter.cpp#L1534

PR #803, opened 2023-04-02
Create wallet RPC communication with the node

TheQuantumPhysicist:
The wallet needs to be able to get blocks and submit keys to the node for various purposes. The first attempt at creating the RPC client implementation is in this PR.


alfiedotwtf (approved):

wallet/wallet-node-client/tests/call_tests.rs
```diff
@@ -0,0 +1,145 @@
...
+    let rpc_bind_address = rpc_subsys.http_address().cloned().unwrap();
+
+    let _rpc = manager.add_subsystem("rpc-test", rpc_subsys);
+
+    tokio::spawn(async move { manager.main().await });
+
+    (shutdown_trigger, chainstate_subsys, rpc_bind_address)
+}
+
+// TODO: why do we need multi_thread? Otherwise, rpc calls block forever.
```
pavel-kokolemin, line 82: This is because you are using `ureq`, which uses the blocking IO API. And you should not use blocking IO within the async runtime.

wallet/wallet-node-client/src/rpc_client/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+        let block_id_hex = block_id.hex_encode();
+
+        let req_data = ureq::json!({
+            "method": "chainstate_get_block",
+            "jsonrpc": "2.0",
+            "id": REQ_ID.fetch_add(1, std::sync::atomic::Ordering::SeqCst),
+            "params": ureq::json!({
+                "id": block_id_hex,
+            }),
+        });
```
pavel-kokolemin, line 97: I think we should not keep RPC in sync manually. Why not use generated `jsonrpsee` client instead?

wallet/wallet-node-client/tests/call_tests.rs
```diff
@@ -0,0 +1,145 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn wallet_rpc_communication() {
+    let chain_config = Arc::new(common::chain::config::create_unit_test_config());
+
+    let (_shutdown_trigger, chainstate_handle, rpc_bind_address) =
+        start_subsystems(chain_config.clone(), "127.0.0.1:0".to_string()).await;
+
+    std::thread::sleep(std::time::Duration::from_secs(2));
```
erasmospunk, line 90: Why does it need to sleep for 2 seconds here? Isn't there a better way of handling it?

TheQuantumPhysicist: I don't think it does. I was playing with it and trying stuff in the tests. I'll remove it.

wallet/wallet-node-client/src/rpc_client/mod.rs
TheQuantumPhysicist: I agree... but besides that I don't know how to do that, there's a problem with whether we're gonna even keep jsonrpcsee due to lack of authentication. Until that's resolved, I don't really care how this works.

wallet/wallet-node-client/tests/call_tests.rs
TheQuantumPhysicist: Yeah... I'll have to find another library to do this. Unfortunately this is the best library and most supported I could find.

stanislav-tkach (approved):

pavel-kokolemin (approved):

muursh (approved):
looks fine to me

PR #805, opened 2023-04-02
Workaround for race in shutdown RPC call

pavel-kokolemin:
There was an unexpected failed functional test run: https://github.com/mintlayer/mintlayer-core/actions/runs/4582639337/attempts/1
After reading the logs and some experimentation, I'm pretty sure this happened because the `node_shutdown` RPC call can sometimes return an error. The reason is a race between processing the `jsonrpsee` response and shutting down the process. It looks like this:

```
$ curl -v -XPOST -H"Content-Type: application/json" http://localhost:3030 -d '{"jsonrpc":"2.0","id":1,"method":"node_shutdown","params":null}'                                                     
...
* Empty reply from server             
* Closing connection 0
curl: (52) Empty reply from server                                                                                                                                                                 
```

I could not find a proper fix for this problem (it looks like `jsonrpsee` does not support graceful shutdowns), so I added this workaround instead.

### DIFF
--- a/node/src/rpc.rs
+++ b/node/src/rpc.rs
@@ -15,7 +15,7 @@
 
 //! Node RPC methods
 
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 use chainstate_launcher::ChainConfig;
 use subsystem::manager::ShutdownTrigger;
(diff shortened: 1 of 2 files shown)


TheQuantumPhysicist (approved):
We're planning to move away from jsonrpsee's crappy server soon. If you could, please explain in the code where the person writing the new http wrapper should do the graceful shutdown.

stanislav-tkach (approved):

PR #807, opened 2023-04-03
Discussion: Reverse hex endianness for H256 and Uint256 hex strings

TheQuantumPhysicist:
### What does this PR do?

It basically creates a 1:1 mapping for hashes that encodes to bytes to encode the exact same way to hex.

### The issue we're trying to solve and the issues that may ensue

The primary issue under question here is that bitcoin [uses reversed order to print hashes in hex](https://bitcoin.stackexchange.com/questions/116730/why-does-bitcoin-core-print-sha256-hashes-uint256-bytes-in-reverse-order). Meaning, even though bitcoin serializes data in little endian, the hexadecimal value of hashes is always printed as big endian.

In Bitcoin Core, this is taken care of manually, where there's zero interfacing bitcoin hexadecimal representation and serialization representation. It's C++.

For us in Rust, this is a difficult situation. As was done in #806, we have created a 1:1 mapping from byte encoding and hex encoding. It's what makes sense and the right thing to do in a software written from scratch. However, this basically makes it impossible to read hexadecimal hashes directly from bitcoin without assumptions, making interfacing harder. 

If we choose to merge this PR, we will, in the future, most probably face issues integrating Bitcoin in our node. It may be as simple `H256::from_bitcoin_uint256_hex()`, or it may have deeper consequences. 

What do you think? Should we go ahead and merge this?


common/src/uint/impls.rs
```diff
@@ -691,11 +691,11 @@ mod tests {
     pub fn uint256_display_test() {
         assert_eq!(
             format!("{:?}", Uint256::from_u64(0xDEADBEEF)),
-            "0x00000000000000000000000000000000000000000000000000000000deadbeef"
+            "0xefbeadde00000000000000000000000000000000000000000000000000000000"
         );
```
iljakuklic, line 695: I think `Uint256` values should be printed in BE, so the printed string stays in the natural numeric notation. Same as `u64` and similar.

common/src/uint/impls.rs
```diff
@@ -542,8 +542,8 @@ macro_rules! construct_uint {
             fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
                 let &$name(ref data) = self;
                 write!(f, "0x")?;
-                for ch in data.iter().rev() {
-                    write!(f, "{:016x}", ch)?;
+                for ch in data {
+                    write!(f, "{:016x}", ch.to_be())?;
```
iljakuklic, line 546: I don't think [to_be](https://doc.rust-lang.org/std/primitive.u64.html#method.to_be) does what you want here. It is a platform-dependent operation that does nothing on a BE computer and swaps bytes on a LE computer (causing it to print essentially a completely different number). So different string will be printed depending on endianness of the machine the code is running on, which does not sound right.

iljakuklic (commented):
I am in favour of the idea in general.

Below started some discussion regarding printing of the `Uint256` type:

common/src/uint/impls.rs
TheQuantumPhysicist: You're saying that it's machine dependent, as if it's a problem. I really don't understand. Don't we want a specific output format regardless of the machine type? Whether it's big or little endian, it'll be formatted as big endian. What's wrong with that?

common/src/uint/impls.rs
TheQuantumPhysicist: The problem here is that rust treats both to_string and display similarly and Display implements ToString automatically... which can easily lead to confusion.

Imagine a dev working, not worrying about endianness. They want a function to convert to hex. They try to_string, and it writes in hex. They use it, shrug it off and contiue. Big mistake if not caught in a PR.

How do we prevent big mistakes? We make them impossible. 

The way I think is: There should be exclusively one way to print things, so that people don't mix multiple ways up.

common/src/uint/impls.rs
pavel-kokolemin: I think `u64::to_be` will not work correctly on big endian (which is no-op there) and `u64::swap_bytes` should be used instead.

common/src/uint/impls.rs
pavel-kokolemin: I also think it would be better to keep printing our `UintXXX` types in the natural order (like it was before) for consistency with built-in types.

common/src/uint/impls.rs
iljakuklic: > You're saying that it's machine dependent, as if it's a problem. I really don't understand. Don't we want a specific output format regardless of the machine type? Whether it's big or little endian, it'll be formatted as big endian. What's wrong with that?

The default formatting routine already takes care of that. `write!(f, "{:016x}", ch)` already writes the number as a 64-bit number regardless of the platform. So if you explicitly convert `to_be`, the conversion happens in effect twice, so you end up printing in native order which is platform dependent.

common/src/uint/impls.rs
iljakuklic: > Imagine a dev working, not worrying about endianness. They want a function to convert to hex. They try to_string, and it writes in hex. They use it, shrug it off and contiue. Big mistake if not caught in a PR.

I think we can get both behave consistently as long as the conversion methods between `Uint256` and the likes of `H256`, `[u8; 32]` and such consistently perform a BE conversion.

common/src/uint/impls.rs
TheQuantumPhysicist: I reverted this change as per our discussion.

pavel-kokolemin (approved):

iljakuklic (approved):
Looks fine to me. A small comment below.

chainstate/types/src/block_index.rs
```diff
@@ -27,15 +26,15 @@ pub struct BlockIndex {
     block_id: Id<Block>,
     block_header: BlockHeader,
     some_ancestor: Id<GenBlock>,
-    chain_trust: Uint256,
+    chain_trust: H256,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: That's fair. I'll go for it.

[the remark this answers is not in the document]
TheQuantumPhysicist: Done.

iljakuklic (approved):

PR #808, opened 2023-04-04
Add network protocol version to p2p connections

pavel-kokolemin:
I think having a separate field to negotiate the p2p network protocol will make future upgrades easier.
Bitcoin Core uses something like that too: https://github.com/bitcoin/bitcoin/blob/master/src/version.h

### DIFF
--- a/p2p/src/error.rs
+++ b/p2p/src/error.rs
@@ -22,14 +22,18 @@ use common::{
 };
 use mempool::error::Error as MempoolError;
 
+use crate::protocol::NetworkProtocol;
+
 /// Errors related to invalid data/peer information that results in connection getting closed
 /// and the peer getting banned.
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum ProtocolError {
+    #[error("Peer has an unsupported network protocol: {0:?}")]
+    UnsupportedProtocol(NetworkProtocol),
     #[error("Peer is in different network. Our network {0:?}, their network {1:?}")]
     DifferentNetwork([u8; 4], [u8; 4]),
     #[error("Peer has an unsupported version. Our version {0}, their version {1}")]
-    InvalidVersion(SemVer, SemVer),
+    UnsupportedVersion(SemVer, SemVer),
     #[error("Peer is unresponsive")]
     Unresponsive,
     #[error("Locator size ({0}) exceeds allowed limit ({1})")]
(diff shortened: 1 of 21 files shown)


p2p/src/net/default_backend/backend.rs
```diff
@@ -479,17 +483,19 @@ where
...
+                version,
                 receiver_address,
                 handshake_nonce,
-                user_agent,
             } => self.create_peer(
                 peer_id,
                 handshake_nonce,
                 PeerInfo {
                     peer_id,
+                    protocol,
```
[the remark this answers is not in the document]
pavel-kokolemin: Right, the version field is enough. But using a separate filed for the protocol version should be a bit easier to use. There will probably be many node releases, but much fewer protocol upgrades. And this is a very important part, so making it explicit is better.
Also, what if we have third party clients in our p2p network? Why should they have to emulate versions of our node? And which version should they use?
Bitcoin Core also uses two fields for this. The protocol version is sent over the network in the handshake, and the application version is sent in the user agent string. Both values are closely related, but only one is used to decide how to communicate with the peer (the protocol version).

p2p/tests/block_announcement.rs
```diff
@@ -150,15 +151,13 @@ where
     ));
 }
 
-// TODO: Implement announcements resending in partially connected networks.
```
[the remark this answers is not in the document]
pavel-kokolemin: Sure, I just removed the extra TODO copies.

p2p/src/net/default_backend/tests.rs
```diff
@@ -66,7 +66,7 @@ where
     {
         assert_eq!(address, conn2.local_addresses()[0]);
         assert_eq!(peer_info.network, *config.magic_bytes());
-        assert_eq!(peer_info.version, SemVer::new(0, 1, 0));
+        assert_eq!(peer_info.version, *config.version());
```
TheQuantumPhysicist, line 69: The version should be coming from Cargo... not from config. But maybe this is something for another day.

You can write a TODO in ChainConfig at the version variable stating this.

p2p/src/error.rs
```diff
@@ -22,14 +22,18 @@ use common::{
...
 /// and the peer getting banned.
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum ProtocolError {
+    #[error("Peer has an unsupported network protocol: {0:?}")]
+    UnsupportedProtocol(NetworkProtocol),
     #[error("Peer is in different network. Our network {0:?}, their network {1:?}")]
     DifferentNetwork([u8; 4], [u8; 4]),
     #[error("Peer has an unsupported version. Our version {0}, their version {1}")]
-    InvalidVersion(SemVer, SemVer),
+    UnsupportedVersion(SemVer, SemVer),
```
TheQuantumPhysicist, line 36: Is this really usable?

TheQuantumPhysicist (approved):

p2p/src/error.rs
pavel-kokolemin: Right, removed, thanks.

p2p/src/net/default_backend/tests.rs
pavel-kokolemin: I updated `Builder` to use `CARGO_PKG_VERSION`. Is this OK?
```
...
            version: SemVer::try_from(env!("CARGO_PKG_VERSION"))
                .expect("invalid CARGO_PKG_VERSION value"),
```

stanislav-tkach (approved):

PR #809, opened 2023-04-04
Generalize error handling in tx-verifier-related traits

iljakuklic:
Make error handling of some traits more general. This is to prepare the traits to be callable from a different subsystem where more error states are possible.

Sadly this resulted in loss of error information granularity in some places (see `TODO`s in the code), as a temporary workaround to keep the scope creep of the PR in check. Could be addressed in subsequent development.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/error.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/error.rs
@@ -104,7 +104,8 @@ pub enum ConnectTransactionError {
     #[error("Data of pool {0} not found")]
     PoolDataNotFound(PoolId),
 
-    // TODO The following should contain more granullar inner error information
+    // TODO The following should contain more granular inner error information
+    //      https://github.com/mintlayer/mintlayer-core/issues/811
     #[error("Fetching undo data failed")]
     UndoFetchFailure,
     #[error("Some transaction verifier storage error")]
--- a/chainstate/tx-verifier/src/transaction_verifier/error.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/error.rs
@@ -107,6 +107,8 @@ pub enum ConnectTransactionError {
     // TODO The following should contain more granullar inner error information
     #[error("Fetching undo data failed")]
     UndoFetchFailure,
+    #[error("Some transaction verifier storage error")]
+    TxVerifierStorage,
 }
 
 impl From<chainstate_storage::Error> for ConnectTransactionError {
--- a/chainstate/tx-verifier/src/transaction_verifier/error.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/error.rs
@@ -103,6 +103,10 @@ pub enum ConnectTransactionError {
     InvalidOutputTypeInReward(Id<Block>),
     #[error("Data of pool {0} not found")]
     PoolDataNotFound(PoolId),
+
+    // TODO The following should contain more granullar inner error information
+    #[error("Fetching undo data failed")]
+    UndoFetchFailure,
 }
 
 impl From<chainstate_storage::Error> for ConnectTransactionError {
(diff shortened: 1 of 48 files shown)


chainstate/tx-verifier/src/transaction_verifier/error.rs
```diff
@@ -103,6 +103,12 @@ pub enum ConnectTransactionError {
     InvalidOutputTypeInReward(Id<Block>),
     #[error("Data of pool {0} not found")]
     PoolDataNotFound(PoolId),
+
+    // TODO The following should contain more granullar inner error information
```
TheQuantumPhysicist, line 107: granular

chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs
```diff
@@ -98,38 +99,39 @@ impl<C, S: TransactionVerifierStorageRef, U: UtxosView, A: PoSAccountingView>
...
 
-    fn get_undo_data(
-        &self,
-        id: Id<Block>,
-    ) -> Result<Option<UtxosBlockUndo>, storage_result::Error> {
+    fn get_undo_data(&self, id: Id<Block>) -> Result<Option<UtxosBlockUndo>, Self::Error> {
         match self.utxo_block_undo.data().get(&TransactionSource::Chain(id)) {
             Some(v) => Ok(Some(v.undo.clone())),
-            None => self.storage.get_undo_data(id),
+            None => self.storage.get_undo_data(id).map_err(|_| todo!("PR")),
```
TheQuantumPhysicist, line 126: ?

consensus/src/pos/error.rs
```diff
@@ -60,4 +60,8 @@ pub enum ConsensusPoSError {
     InvalidTargetBlockTime,
     #[error("CRITICAL: Block time must be monotonic")]
     InvariantBrokenNotMonotonicBlockTime,
+
+    // TODO the following error should include the corresponding error from UtxosView
```
TheQuantumPhysicist, line 64: In those TODOs like this one, can you please reference the PR number so that anyone seeing this would understand the context?

TheQuantumPhysicist: Also maybe create an issue and put more details there.

iljakuklic: Will do

chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs
iljakuklic: Must have overlooked this one, will fix

TheQuantumPhysicist (approved):
Looking good besides the minor comments. Most important thing is to link those TODO comments to an issue + the PR so that we can find the problem and fix it later and not be lost in the code.

chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs
TheQuantumPhysicist: If you add a comment `// TODO(PR)` instead then CI won't pass unless you fix it. If you want to do it with this format that you chose `todo!("PR")`, please add it to `codecheck.py` so that it helps you not to forget.

iljakuklic: > What's confusing to me is that it has not been caught by `codecheck.py`: [relevant part](https://github.com/mintlayer/mintlayer-core/blob/eedfdae4eb196966908b072e97d39870b8317cb1/.github/scripts/codecheck.py#L141).

I see now, it ignores `todo!` with an argument

iljakuklic: Maybe we could use clippy to catch this, will be more reliable than grepping code: https://rust-lang.github.io/rust-clippy/stable/index.html#todo

TheQuantumPhysicist: If clippy can do it, then it's better.

azarovh (approved):

muursh (approved):

stanislav-tkach (approved):

PR #810, opened 2023-04-04
Use different default data dirs

pavel-kokolemin:
We use the same default data directories for all networks, which is wrong.
For example, if you run mainnet first and then regtest, you will get this confusing error:
```
Mintlayer node launch failed: Initialization error: Not at genesis but block at height 1 not available

Caused by:
    Not at genesis but block at height 1 not available
```
Let's change that and use different default data directories for different networks (like it's done in Bitcoin Core).

I also added file locking to prevent users from starting multiple node instances using the same data dir.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -29,6 +29,7 @@ serde = { workspace = true, features = ["derive"] }
 toml = "0.7"
 directories = "4.0"
 paste = "1.0"
+fs4 = "0.6.3"
 
 [dev-dependencies]
 crypto = { path = "../crypto" }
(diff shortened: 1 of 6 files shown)


node/Cargo.toml
```diff
@@ -29,6 +29,7 @@ serde = { workspace = true, features = ["derive"] }
 toml = "0.7"
 directories = "4.0"
 paste = "1.0"
+fs4 = "0.6.3"
```
TheQuantumPhysicist, line 32: 0.6

muursh (approved):

stanislav-tkach (approved):

iljakuklic (approved):

PR #812, opened 2023-04-04
Block production refactoring and fixing of some issues + introduce a thread-pool for heavy lifting

TheQuantumPhysicist:
Until a proper solution for exhaustion of async is found, this shall be a good solution.

One thing is left in this PR, which is actually doing the block solving in the thread-pool. Everything is in place. I think this PR qualifies for reviews though.

### DIFF
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::sync::Arc;
+use std::sync::{atomic::AtomicBool, Arc};
 
 use chainstate::{ChainstateHandle, PropertyQueryError};
 use chainstate_types::{BlockIndex, GetAncestorError};
@@ -152,12 +152,14 @@ impl BlockMaker {
     pub fn solve_block(
         &self,
         mut block_header: BlockHeader,
+        stop_flag: Arc<AtomicBool>,
     ) -> Result<BlockHeader, BlockProductionError> {
         // TODO: use a separate executor for this loop to avoid starving tokio tasks
         consensus::finalize_consensus_data(
             &self.chain_config,
             &mut block_header,
             self.current_tip_height,
+            stop_flag,
         )?;
 
         Ok(block_header)
@@ -227,7 +229,7 @@ impl BlockMaker {
                 consensus_data,
             );
 
-            let block_header = self.solve_block(block_header)?;
+            let block_header = self.solve_block(block_header, Arc::new(false.into()))?;
 
             let block = Block::new_from_header(block_header, block_body.clone())?;
 
--- a/blockprod/src/detail/block_maker.rs
+++ b/blockprod/src/detail/block_maker.rs
@@ -146,21 +146,15 @@ impl BlockMaker {
         Ok(consensus_data)
     }
 
-    pub async fn generate_block(
+    pub fn generate_block(
         &self,
+        consensus_data: ConsensusData,
         transactions: Vec<SignedTransaction>,
     ) -> Result<Block, BlockProductionError> {
         // TODO: instead of the following static value, look at
         // self.chain_config for the current block reward, then send
         // it to self.reward_destination
         let block_reward = BlockReward::new(vec![]);
-
-        let consensus_data = self
-            .pull_consensus_data(
-                self.current_tip_id,
-                BlockTimestamp::from_duration_since_epoch(self.time_getter.get_time()),
-            )
-            .await?;
 
         let mut block = Block::new(
             transactions,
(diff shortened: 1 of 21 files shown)


alfiedotwtf:
I've got a few comments on this PR, but after rebasing some code on top of this PR, unless I'm mistaken I starting to think that we may need to replace all the `tokio::spawn()` calls in `blockprod/` with `tokio::spawn_blocking()`. I'm still going through a few more tests, and then I'll start commenting here.

azarovh (approved):
looks good to me

blockprod/src/detail/block_maker.rs
```diff
@@ -97,29 +103,16 @@ impl BlockMaker {
...
-            transactions,
-            self.current_tip_id,
-            BlockTimestamp::from_duration_since_epoch(self.time_getter.get_time()),
-            ConsensusData::None,
-            block_reward,
-        )?;
-
+        prev_block_id: Id<GenBlock>,
+        block_timestamp: BlockTimestamp,
+    ) -> Result<ConsensusData, BlockProductionError> {
```
alfiedotwtf, line 110: Nice. This is what I mentioned early - I was having trouble finalizing the block in a separate thread because `transactions` can't be safely sent across threads. By only using `prev_block_id` to calculate the consensus data, you've removed the need for me to derive `Sync` everywhere, which is what I was going to start this morning.

blockprod/src/detail/block_maker.rs
```diff
@@ -155,16 +149,24 @@ impl BlockMaker {
...
+        Ok(consensus_data)
+    }
 
+    pub fn solve_block(
+        chain_config: Arc<ChainConfig>,
+        mut block_header: BlockHeader,
+        current_tip_height: BlockHeight,
+        stop_flag: Arc<AtomicBool>,
+    ) -> Result<BlockHeader, BlockProductionError> {
         // TODO: use a separate executor for this loop to avoid starving tokio tasks
```
alfiedotwtf, line 161: I've used this PR to move `solve_block()` in a different thread, but I've done it inside `run()` instead of here in `solve_block()`. The reasoning is that since `run()` is `async`, we either need to wait to join in from `run()` meaning that `run()` won't be able to progress, or we spawn new threads per iteration, which is wrong.

So this comment can probably be dropped, but I'll do it in my PR.

blockprod/src/detail/block_maker.rs
```diff
@@ -207,8 +209,40 @@ impl BlockMaker {
...
+            let consensus_data = self.pull_consensus_data(self.current_tip_id, timestamp).await?;
+
+            let block_header = BlockHeader::new(
+                self.current_tip_id,
+                tx_merkle_root,
+                witness_merkle_root,
+                timestamp,
+                consensus_data,
+            );
+
```
alfiedotwtf, line 237: The above three assignments don't change per iteration, so I've moved them in my PR.

common/src/chain/block/mod.rs
```diff
@@ -132,6 +134,32 @@ impl Block {
         Ok(block)
     }
 
+    pub fn new_from_header(
+        header: BlockHeader,
```
alfiedotwtf, line 138: Should this be renamed given that both `header` and `body` is supplied?

alfiedotwtf:
As how the code is now, `stop_flag` is not being `store(true...)` anywhere, and so PoW mining will continue uninterrupted until nonce exhaustion. This is why in #815 I moved to `10_000`, so that when mining fails up to this limit, we return and then iterate once again in the outer loop. this would be a good thing because now the header will be rebuilt with a new timestamp whereas in this PR the timestamp will be fixed at all times.

blockprod/src/detail/block_maker.rs
```diff
@@ -207,46 +209,120 @@ impl BlockMaker {
...
+            // In this we store the successfully mined block, if any.
+            // This is necessary because a no futures can be awaited inside `select!` macro
+            // TODO: is there a better way to get the value from the `select!` macro below instead of mutating this value?
+            let mut block_to_submit = None;
+
+            // We wait for one of the following to happen:
+            // 1. `self.block_maker_rx` to tell us that the chainstate has a new tip, or that we should stop for some other reason
+            // 2. the mining thread to finish, and tell us whether it was successful or not,
+            //    in which case we can attempt to submit the block to chainstate
+            crossbeam_channel::select! {
```
alfiedotwtf, line 266: Not sure we want to introduce more blocking code within an async function. I know that the current mining is blocking already, but I'm not sure if we gain anything here.

blockprod/src/detail/block_maker.rs
```diff
@@ -207,46 +209,120 @@ impl BlockMaker {
...
+            let block_header = BlockHeader::new(
+                self.current_tip_id,
+                tx_merkle_root,
+                witness_merkle_root,
+                timestamp,
+                consensus_data,
+            );
+
+            is_send(block_header.clone());
+
```
alfiedotwtf, line 239: Not sure why this is necessary here.

alfiedotwtf:
I know #815 isn't ideal, but it does solve a lot of the issues in the PR, though I'm at the moment not sure how this PR or #815 integrates with PoS... so maybe I'm not seeing your point of view yet.

I'll review the Wallet RPC PR and start looking at PoS, so maybe this PR will make more sense in context.

stanislav-tkach (approved):

TheQuantumPhysicist:
I've done huge changes to this PR, so I'm gonna close this and reopen it with a new goal when I'm done.

PR #813, opened 2023-04-05
Use clippy to find the leftover `todo!()`s

iljakuklic:
As per discussion in #809.

Had to change some existing code to make it pass, the idea is to establish the following convention:

* Use `todo!` for short-term todos that are supposed to be fixed in the current PR.
* Use `unimplemented!` for stuff that is not implemented and will stay unimplemented beyond the current PR.

### DIFF
--- a/.github/scripts/codecheck.py
+++ b/.github/scripts/codecheck.py
@@ -139,7 +139,7 @@ def check_todos():
 
         with open(path) as file:
             file_data = file.read()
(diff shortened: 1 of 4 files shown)


iljakuklic:
This is what you get if there are still `todo!`s in the code:
https://github.com/mintlayer/mintlayer-core/actions/runs/4616271887

pavel-kokolemin (approved):

azarovh (approved):

TheQuantumPhysicist (approved):

stanislav-tkach (approved):

PR #814, opened 2023-04-05
Add NotFound to TransactionResponse (p2p)

pavel-kokolemin:
We don't send a response when a peer has requested a transaction that is no longer in the mempool. This can happen when a new block is mined and the recently announced transaction is removed from the mempool. As a result, the `announced_transactions` set will grow and the peer may be banned later.

### DIFF
--- a/p2p/src/message.rs
+++ b/p2p/src/message.rs
@@ -32,7 +32,7 @@ pub enum SyncMessage {
     HeaderListResponse(HeaderListResponse),
     BlockResponse(BlockResponse),
     TransactionRequest(Id<Transaction>),
-    TransactionResponse(SignedTransaction),
+    TransactionResponse(TransactionResponse),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -133,6 +133,14 @@ impl BlockResponse {
     pub fn into_block(self) -> Block {
         *self.block
     }
+}
+
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
+pub enum TransactionResponse {
+    #[codec(index = 0)]
+    NotFound(Id<Transaction>),
+    #[codec(index = 1)]
+    Found(SignedTransaction),
 }
 
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
--- a/p2p/src/sync/peer.rs
+++ b/p2p/src/sync/peer.rs
@@ -29,7 +29,7 @@ use void::Void;
 use chainstate::chainstate_interface::ChainstateInterface;
 use chainstate::{ban_score::BanScore, BlockError, BlockSource, ChainstateError, Locator};
 use common::{
-    chain::{block::BlockHeader, Block, SignedTransaction, Transaction},
+    chain::{block::BlockHeader, Block, Transaction},
     primitives::{BlockHeight, Id, Idable},
 };
 use logging::log;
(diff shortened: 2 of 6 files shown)


p2p/src/message.rs
```diff
@@ -135,6 +135,12 @@ impl BlockResponse {
     }
 }
 
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
+pub enum TransactionResponse {
+    NotFound(Id<Transaction>),
```
TheQuantumPhysicist, line 140: Please add index values for enum arms

p2p/src/sync/peer.rs
```diff
@@ -390,29 +390,44 @@ where
...
+        self.messaging_handle
+            .send_message(self.id(), SyncMessage::TransactionResponse(res))?;
 
         Ok(())
     }
 
-    async fn handle_transaction_response(&mut self, tx: SignedTransaction) -> Result<()> {
-        let id = tx.serialized_hash();
+    async fn handle_transaction_response(&mut self, resp: TransactionResponse) -> Result<()> {
+        match resp {
```
stanislav-tkach, line 409: The repetition inside match bothers me. 🙃 Perhaps the following would be better?..
```rust
let (id, tx) = match resp {
    TransactionResponse::NotFound(id) => (id, None),
    TransactionResponse::Found(tx) => (tx.serialized_hash(), tx),
};

if self.announced_transactions.take(&id).is_none() {
    return Err(P2pError::ProtocolError(ProtocolError::UnexpectedMessage(
        "Unexpected transaction response",
    )));
}

if let Some(tx) = tx {
    self.mempool_handle.call_async_mut(|m| m.add_transaction(tx)).await??;
    self.messaging_handle.make_announcement(Announcement::Transaction(id))
}
```

stanislav-tkach (approved):

p2p/src/sync/peer.rs
TheQuantumPhysicist: Btw, we should never use serialized_hash. Always only use get_id.

TheQuantumPhysicist: Btw, we should never use serialized_hash. Always only use get_id.a

TheQuantumPhysicist (approved):

p2p/src/sync/peer.rs
stanislav-tkach: @TheQuantumPhysicist Perhaps I'm missing something, but as far as I can see, `SignedTransaction` doesn't implement the `Idable` trait.

TheQuantumPhysicist: Take it from `Transaction` within it. This glosses on the issue of transaction malleability.

pavel-kokolemin: I updated all places in p2p where `serialized_hash` was used. And I changed the type of `SignedTransaction::serialized_hash` to catch such errors:
```
-    pub fn serialized_hash(&self) -> Id<Transaction> {
+    pub fn serialized_hash(&self) -> Id<SignedTransaction> {
```
Is it OK now?

TheQuantumPhysicist: No. Please don't change SignedTransaction

So something like tx.transaction().get_id()

pavel-kokolemin: OK, I reverted changing `SignedTransaction::serialized_hash` output type.

PR #815, opened 2023-04-05
Move block solving to a non-async threadpool

alfiedotwtf:


consensus/src/lib.rs
```diff
@@ -32,6 +32,8 @@ use common::{
     primitives::{BlockHeight, Id},
 };
 
+const POW_MAX_NONCE: u128 = 10_000;
```
TheQuantumPhysicist, line 35: Wrong. There should not be a limit.

alfiedotwtf: We talked about this the other day, saying that we should have a limit so that we're not unnecessarily checking `stop_flag` in each iteration.

Also, now that I think about it, if we never stop unless nonce exhaustion or the stop flag is set, then `block_header`'s timestamp will be fixed the whole time

PR #816, opened 2023-04-05
Unify announcements and responces processing (p2p sync manager)

stanislav-tkach:


pavel-kokolemin (approved):

p2p/src/message.rs
```diff
@@ -145,10 +150,17 @@ pub struct PingResponse {
     pub nonce: u64,
 }
 
-#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Announcement {
```
TheQuantumPhysicist, line 154: For the love of god... can we just get done with the work "Announcement" forever and ever and celebrate that grepping it will result in nothing? Isn't the curse of libp2p ever gonna be gone?!

TheQuantumPhysicist (approved):

p2p/src/message.rs
stanislav-tkach: This can be done easily, but I like this separation:
```rust
trait MessagingService {
    fn send_message(&mut self, peer: PeerId, message: SyncMessage) -> crate::Result<()>;
    fn make_announcement(&mut self, announcement: Announcement) -> crate::Result<()>;
}
```
This way we can separate the messages that can be broadcast. Perhaps this method can be renamed to something like that:
```rust
fn broadcast_message(&mut self, message: SyncBroadcastMessage) -> crate::Result<()>;
```
What do you think?

TheQuantumPhysicist: I prefer. I'm strictly against making it sound like announcements are something special. They're all messages that get the same treatment.

stanislav-tkach: I have remove this enum and renamed the `make_announcement` method.

Issue #787, opened 2023-03-27
Limit difficulty swings for PoS retargeting


azarovh:
Difficulty change should be limited in a way to prevent high spikes.
Bitcoin for example limits timespan by factor of 4 (https://github.com/bitcoin/bitcoin/blob/master/src/pow.cpp#L56). 

For us, this approach might not be good enough because we adjust difficulty on every block, so the limiting function should be more smooth as well.

alfiedotwtf:
I think we already do this in `common/src/chain/pow.rs:max_retarget_factor()`.

TheQuantumPhysicist:
> I think we already do this in `common/src/chain/pow.rs:max_retarget_factor()`.

This issue is meant for PoS, not PoW.

PR #817, opened 2023-04-06
Limit target swings by factor of 4

azarovh:
This is a solution that bitcoin core uses. 
For us, this limiting function might not be smooth enough but it's better than nothing for now.

#787


TheQuantumPhysicist (approved):
Keep in mind that this is completely useless because in bitcoin the retarget happens for 2016 blocks, not every block. This is 2000x the allowed swings in bitcoin per block.

stanislav-tkach (approved):

muursh (approved):

iljakuklic (approved):

Issue #192, opened 2022-05-24
Add RPC authentication


iljakuklic:
At the moment, anyone able to establish a connection is able to send RPC command. Some form of authentication should be added.

TheQuantumPhysicist:
Keep in mind also that we want to support cookie authentication. A cookie is basically a temporary authentication file generated if no username/password are specified for RPC. On launch, the node creates the file `.cookie` in the data directory, which has a random username and password in the format `username:password` (completely generated at random) in one line. 

When functional tests need to use the node it just reads the cookie file and parses it, or when a user wants to run a quick RPC command with curl, then they just pipe `$(cat ~/.mintlayer/.cookie)` to their command and it just works without caring what the username/password are, since http rpc typically uses `username:password` as the format of inline authentication.

This file is created on every launch, and is only accepted as means for authentication if no username and password for RPC are specified.

TheQuantumPhysicist:
It seems that jsonrpsee doesn't support authentication at all.

While jsonrpsee doesn't support authentication, the way to do it is by adding an authentication layer, that acts as a reverse-proxy, using the tower framework. An example of this is the [authentication layer](https://docs.rs/tower-http/latest/tower_http/auth/struct.AddAuthorizationLayer.html). 

The problem with this approach is that we, kind of, have to reimplement the http/websocket servers ourselves. Currently, the way it works is:

```
socket port -> jsonrpsee -> function handler
```

But it has to become:

```
socket port -> authentication layer -> jsonrpsee -> function handler
```

The reason we call it "reimplement" is that this is basically [what jsonrpsee did in their source code](https://github.com/paritytech/jsonrpsee/blob/18ee2122ec84392c915bfe26bea73c30f41a7a6d/server/src/server.rs). We can see that the authentication layer is just a specialization of that. We have to implement the socket and everything around it, and jsonrpsee will become just a function handler. At this point, I'm just wondering whether we should just use some web framework from scratch like ActixWeb and do this whole thing ourselves.

Pros:
- Full control to add all the features we need
- Less bloat and less fragmentation in the code

Cons:
- More time to do this right and more code to manage
- We may have to reimplement the macro magic of jsonrpsee to generate function mapping.

Two more things will make this whole problem much more complex, and we should consider this while thinking of a solution:

1. End-to-end encryption for RPC, if we want to implement that and not rely on the user's discretion
[...]

PR #819, opened 2023-04-07
Add authentication to RPC

pavel-kokolemin:
To keep things simple, the password is specified in plaintext. This can be problematic because it allows the password to be read from the config file or from the process list (at least on Linux). I think Bitcoin Core is migrating to hashed passwords for this reason. We can add this later.

This should resolve https://github.com/mintlayer/mintlayer-core/issues/192

### DIFF
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -16,6 +16,7 @@ crypto = { path = "../crypto/" }
 
 # External dependencies
 anyhow = "1.0"
+thiserror.workspace = true
 async-trait.workspace = true
 jsonrpsee = { workspace = true, features = ["full"] }
 tower-http = { version = "0.3", features = ["auth"] }
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -26,3 +26,4 @@ base64 = "0.21"
 
 [dev-dependencies]
 tokio.workspace = true
+rstest.workspace = true
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -12,11 +12,17 @@ rust-version.workspace = true
 subsystem = { path = "../subsystem" }
 logging = { path = "../logging" }
 utils = { path = "../utils/" }
+crypto = { path = "../crypto/" }
 
 # External dependencies
 anyhow = "1.0"
 async-trait.workspace = true
 jsonrpsee = { workspace = true, features = ["full"] }
+tower-http = { version = "0.3", features = ["auth"] }
+tower = { version = "0.4", features = ["util"] }
+hyper = "0.14"
+http = "0.2"
+base64 = "0.21"
 
 [dev-dependencies]
 tokio.workspace = true
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -24,5 +24,7 @@ tower = { version = "0.4", features = ["util"] }
 tower-http = { version = "0.3", features = ["auth"] }
 
 [dev-dependencies]
+test-utils = { path = "../test-utils" }
+
 rstest.workspace = true
 tokio.workspace = true
--- a/rpc/Cargo.toml
+++ b/rpc/Cargo.toml
@@ -8,23 +8,21 @@ rust-version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-# Local dependencies
-subsystem = { path = "../subsystem" }
-logging = { path = "../logging" }
-utils = { path = "../utils/" }
 crypto = { path = "../crypto/" }
+logging = { path = "../logging" }
+subsystem = { path = "../subsystem" }
+utils = { path = "../utils/" }
 
-# External dependencies
 anyhow = "1.0"
-thiserror.workspace = true
 async-trait.workspace = true
-jsonrpsee = { workspace = true, features = ["full"] }
-tower-http = { version = "0.3", features = ["auth"] }
-tower = { version = "0.4", features = ["util"] }
-hyper = "0.14"
-http = "0.2"
 base64 = "0.21"
+http = "0.2"
+hyper = "0.14"
+jsonrpsee = { workspace = true, features = ["full"] }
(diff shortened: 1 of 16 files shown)


rpc/src/rpc_creds.rs
```diff
@@ -0,0 +1,122 @@
...
+        use std::os::unix::prelude::OpenOptionsExt;
+        options.mode(0o600);
+    }
+
+    options.create(true).write(true).open(path)?.write_all(data.as_bytes())
+}
+
+impl RpcCreds {
+    pub fn new(
+        data_dir: &Path,
```
TheQuantumPhysicist, line 59: `impl AsRef<Path>`

rpc/src/rpc_creds.rs
```diff
@@ -0,0 +1,122 @@
...
+        username: Option<&str>,
```
TheQuantumPhysicist, line 60: `impl AsRef<str>`

rpc/src/rpc_creds.rs
```diff
@@ -0,0 +1,122 @@
...
+                    .with_context(|| format!("failed to create cookie file {cookie_file:?}"))?;
+
+                Ok(Self {
+                    username,
+                    password,
+                    cookie_file: Some(cookie_file),
+                })
+            }
+
+            _ => anyhow::bail!("both RPC username and password must be set"),
```
TheQuantumPhysicist, line 100: both RPC username and password must be set or unset, together

rpc/src/rpc_creds.rs
```diff
@@ -0,0 +1,122 @@
...
+    options.create(true).write(true).open(path)?.write_all(data.as_bytes())
+}
+
+impl RpcCreds {
+    pub fn new(
+        data_dir: &Path,
+        username: Option<&str>,
+        password: Option<&str>,
+        cookie_file: Option<&str>,
+    ) -> anyhow::Result<Self> {
```
TheQuantumPhysicist, line 63: Please use `thiserror`. We don't use anyhow in libraries.

rpc/src/rpc_creds.rs
```diff
@@ -0,0 +1,133 @@
...
+    #[error("Both RPC username and password must be set or unset, together")]
+    InvalidUsernamePasswordConfig,
+}
+
+impl RpcCreds {
+    pub fn new(
+        data_dir: impl AsRef<Path>,
+        username: Option<impl AsRef<str>>,
+        password: Option<impl AsRef<str>>,
+        cookie_file: Option<&str>,
```
TheQuantumPhysicist, line 73: Is there a reason this is excluded from impl AsRef?

rpc/Cargo.toml
```diff
@@ -12,11 +12,19 @@ rust-version.workspace = true
 subsystem = { path = "../subsystem" }
 logging = { path = "../logging" }
 utils = { path = "../utils/" }
+crypto = { path = "../crypto/" }
```
TheQuantumPhysicist, line 15: Nitpick: Please try to maintain alphabetical ordering of crates

rpc/src/rpc_auth/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::num::NonZeroUsize;
+
+use anyhow::anyhow;
```
TheQuantumPhysicist, line 18: There's a huge anyhow infection all over the place... why?!

rpc/src/lib.rs
```diff
@@ -286,4 +289,86 @@ mod tests {
...
+        anyhow::ensure!(response == "version1");
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn rpc_server_auth() {
+        const GOOD_USERNAME: &str = "correct_username";
+        const GOOD_PASSWORD: &str = "correct_password";
+        const BAD_USERNAME: &str = "wrong_username";
+        const BAD_PASSWORD: &str = "wrong_password";
```
TheQuantumPhysicist, line 340: Please randomize these

TheQuantumPhysicist (approved):
Please fix the issues in the comments, then feel free to merge. Thank you for doing this.

rpc/src/rpc_auth/mod.rs
TheQuantumPhysicist: Actually I meant we should be using thiserror

rpc/src/rpc_auth/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+            .map_err(CheckError::InvalidUtf8Value)?;
+        let (username, password) =
+            username_password.split_once(':').ok_or(CheckError::ColonNotFound)?;
+        let password_valid = verify_password(
+            password.as_bytes(),
+            self.password_hash.clone(),
+            SliceEqualityCheckMethod::TimingResistant,
+        )
+        .map_err(CheckError::KdfError)?;
+        Ok(username == self.username && password_valid)
```
iljakuklic, line 103: Is this a standard method of doing password authentication? If so, is there a pre-made library to implement it rather than parsing headers manually?

iljakuklic (approved):
Good stuff

stanislav-tkach (approved):

rpc/src/rpc_auth/mod.rs
pavel-kokolemin: `RequireAuthorizationLayer::basic` could be used instead, but I decided to write my own version to be able to easily switch to hashed passwords later (when the node does not know the actual passwords).

iljakuklic: > `RequireAuthorizationLayer::basic` could be used instead, but I decided to write my own version to be able to easily switch to hashed passwords later (when the node does not know the actual passwords).

That's fair. Not a big deal.

PR #820, opened 2023-04-08
Remove perpetual block builder and block maker + use a dedicated thread-pool for mining/staking blocks

TheQuantumPhysicist:
Given that proof of stake needs to provide a (possibly) new key for every new block to be produced, the perpetual block builder is useless unless it provides a two-way communication with the wallet, which is unnecessary. The more reasonable strategy for staking is to get the wallet to communicate with the node and issue commands for creating blocks by providing the staking key.

Besides all that, there's lots of code duplication in generate_block.

This is all now gone.

In addition, we use a dedicated thread-pool for solving blocks in order to avoid starving the async thread-pool.


alfiedotwtf:
First off, this is a nice change - it's a whole lot simpler now!

As this is such a big change to previous behaviour, I'm going to outline my summary of the changes for a) other people who want to review, and b) so that my understanding can be corrected when necessary...

Old code:

- Mempool events
  - Start block production if at new tip
     - Exit if currently disabled
     - Collect transactions
     - Loop
       - Get consensus data
       - Solve block
       - Attempt to submit new block
         - Exit if submitted successfully
       - Try reading events
         - Exit if new tip
         - Exit if told to 

- Chainstate events
  - Stop block production if at new tip

- Blockprod events
  - Stop
    - Stop block production
  - Start
    - Enable block production
    - Start block production
  - Drop
    - Stop block production

New code:

- Blockprod events
  - Stop Job
    - Trigger cancel signal on a specific block production job
  - Stop All
     - Loop through all current jobs
      - Trigger cancel signal to stop block production job
  - Generate Block
    - Get start tip
    - Create block production job
      - Create Job Key
      - Create cancel signal
    - Loop
      - Get consensus data
      - Get current tip
      - Exit if current tip != start tip
      - Collect transactions (from mempool or from caller)
      - Create block header
      - Spawn thread
        - Solve block header
        - Send block header back
      - Select! events
        - Cancel signal
          - Stop solving block headers
          - Exit with Cancelled error
        - Block header
          - Loop back if no header
          - Return new block

Given that the `blockprod` subsystem is no longer subscribing to Chainstate or the Mempool events, I'm taking that as `generate_block()` is now a one-shot rather than a perpetual miner i.e call it to generate a single block and then exit (as you've mentioned before in meetings)? 

Thinking about this more, we may still be able to get the perpetual block builder for free by triggering `generate_block()` inside the mempool when there's a new tip (I think I made a TODO comment in the mempool code a while back).

More comments coming...

blockprod/src/detail/mod.rs
```diff
@@ -13,32 +13,70 @@
...
+    },
+    primitives::{BlockHeight, Id},
     time_getter::TimeGetter,
 };
-use mempool::MempoolHandle;
-use tokio::sync::mpsc;
-
-use crate::{
-    detail::block_maker::BlockMaker, interface::BlockProductionInterface, BlockProductionError,
+use futures::channel::oneshot;
```
alfiedotwtf, line 34: As we're already using Tokio, `tokio::sync::oneshot` is a drop-in a 99% drop-in replacement for `futures::channel::oneshot`

blockprod/src/rpc.rs
```diff
@@ -22,54 +22,59 @@ use common::{
...
+    async fn stop_job(&self, job_id: JobKey) -> rpc::Result<()> {
+        handle_error(self.call_mut(move |this| this.stop_job(job_id)).await)
     }
 
     async fn generate_block(
         &self,
         reward_destination_hex: String,
-        transactions_hex: Vec<String>,
+        transactions_hex: Option<Vec<String>>,
     ) -> rpc::Result<String> {
```
alfiedotwtf, line 62: Currently `generate_block()` works as "call function -> block is returned", however with the introduction to `JobKey`, we should probably refactor this API - possibly something like splitting this into multiple functions as there's currently no way for the caller to know what its JobKey is so that it can call `stop_job()`. I was thinking something like:

1. `generate_block()` as is for PoW calls
2. `enqueue_block_generation()` so that PoS can get a JobKey of its pending job
3. `poll_block_generation()` so that PoS can check/get results of its pending job

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +85,291 @@ impl BlockProduction {
...
+            }
 
-        tokio::select! {
-            msg = recv => match msg.expect("Block Builder timed out").expect("Error reading from Block Builder") {
-                BlockBuilderControlCommand::Start => {},
-                _ => panic!("Invalid message received from Block Builder"),
+            tokio::select! {
+                _ = &mut cancel_receiver => {
+                    // TODO: test cancellations
+                    stop_flag.store(true, std::sync::atomic::Ordering::SeqCst);
```
alfiedotwtf, line 312: Maybe my understanding is incorrect, but I thought that when you use `SeqCst` ordering in one place, a happens-before relationship is only created when *all* accesses to that same variable are `SeqCst` too... however there is a `Relaxed` access elsewhere in the code.

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +85,291 @@ impl BlockProduction {
...
+        consensus::finalize_consensus_data(
+            &chain_config,
+            &mut block_header,
+            current_tip_height,
+            stop_flag,
+        )?;
 
-    use super::*;
-    use crate::tests::setup_blockprod_test;
+        Ok(block_header)
```
alfiedotwtf, line 202: As `finalize_consensus_data()` internally doesn't check the return of `mine()`'s success, the spawned thread could send invalid block headers back to the main thread.

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +85,291 @@ impl BlockProduction {
...
+                        .send(block_header)
+                        .expect("Failed to send block header back to main thread");
+                });
+            }
 
-        tokio::select! {
-            msg = recv => match msg.expect("Block Builder timed out").expect("Error reading from Block Builder") {
-                BlockBuilderControlCommand::Start => {},
-                _ => panic!("Invalid message received from Block Builder"),
+            tokio::select! {
```
alfiedotwtf, line 309: We currently don't subscribe to chainstate events, so we will never `stop_flag.store(true...)` on a new chainstate `NewTip` event while solving blocks. And so the only time we currently check for a new tip is when we go around the loop again.

blockprod/src/lib.rs
```diff
@@ -19,15 +19,15 @@ use std::sync::Arc;
...
-use detail::{builder::PerpetualBlockBuilder, BlockProduction};
-use interface::BlockProductionInterface;
+use detail::{BlockProduction, JobKey};
+use interface::blockprod_interface::BlockProductionInterface;
 use mempool::MempoolHandle;
 use subsystem::subsystem::CallError;
-use tokio::sync::mpsc;
 
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
 pub enum BlockProductionError {
```
alfiedotwtf, line 33: We can probably delete a lot of these as they're no longer being used

blockprod/src/rpc.rs
```diff
@@ -22,54 +22,59 @@ use common::{
 
 use serialization::{hex::HexDecode, hex::HexEncode};
 
-use crate::BlockProductionError;
+use crate::{detail::JobKey, BlockProductionError};
 use subsystem::subsystem::CallError;
 
 #[rpc::rpc(server, namespace = "blockprod")]
 trait BlockProductionRpc {
```
alfiedotwtf, line 29: I'm not sure if it's captured anywhere, but we may later want to restrict access to these RPC calls to prevent abuse from unauthed users.

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +85,291 @@ impl BlockProduction {
...
+
+    pub fn stop_all_jobs(&mut self) {
+        let mut all_jobs = Vec::new();
+        while let Some((key, handle)) = self.all_jobs.pop_first() {
+            all_jobs.push((key, handle));
+        }
+
+        log::info!("Cancelling {} jobs", all_jobs.len());
+
+        for (key, handle) in all_jobs.drain(..) {
```
alfiedotwtf, line 113: Unless I'm mistaken, the `pop_first()` call in the above `while()` loop will also drain the hashmap, so there's no need to recreate it in order to drain.

blockprod/src/rpc.rs
```diff
@@ -22,54 +22,59 @@ use common::{
...
-    async fn stop(&self) -> rpc::Result<()>;
+    #[method(name = "stop_all")]
+    async fn stop_all(&self) -> rpc::Result<()>;
 
-    /// Start block production on the next chance (when new tip is available)
-    #[method(name = "start")]
-    async fn start(&self) -> rpc::Result<()>;
+    // Stop a specific job
+    #[method(name = "stop_job")]
+    async fn stop_job(&self, job_id: JobKey) -> rpc::Result<()>;
```
alfiedotwtf, line 36: Should `JobKey` be a `String` that we parse on input instead?

alfiedotwtf:
Apart from my above comments, which could be mistaken, I'm really liking this lightweight rewrite.

alfiedotwtf:
Bah! I wanted to approve my review but instead converted this is "ready for review". Sorry for the notifications

alfiedotwtf (approved):

blockprod/src/detail/mod.rs
TheQuantumPhysicist: We do need to cancel on NewTip. It's just not implemented.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: The right thing to do here is acquire-release, but since it's a one-time thing, it doesn't matter.

Usually we got for relaxed when performance is an issue.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: Fixed

blockprod/src/detail/mod.rs
TheQuantumPhysicist: Done

blockprod/src/lib.rs
TheQuantumPhysicist: I removed those that aren't used and kept ChainstateChannelClosed because we need it for NewTip

blockprod/src/rpc.rs
TheQuantumPhysicist: This is something to be fixed. I did the bare minimum. Please feel free to improve this design.

blockprod/src/rpc.rs
TheQuantumPhysicist: It can be some serialized hex. That's fine. Eventually it'll contain a key too.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: I reverted that change because it broke the tests. Feel free to change it and fix the tests that come with it.

alfiedotwtf: Cool, no problem.

blockprod/src/rpc.rs
alfiedotwtf: Since it will change for PoS, this can be left as it is now.

blockprod/src/detail/mod.rs
alfiedotwtf: Cool, I'll add it to my TODO list.

blockprod/src/lib.rs
alfiedotwtf: Ah of course. Ok, this can be cleaned up later.

alfiedotwtf:
There's minor changes that can be cherry picked from #838.

The obvious solution to seeing chainstate `NewTip` notifications was to use `tokio::sync::broadcast`, however you need to set a maximum number of subscribers on creation and then it will allocate buffers which was not ideal.

So then I was going to switch to `tokio::sync::watch`, but it looks like it will only wake up one of the listening threads.

Pushing #838 so they're not delayed by chainstate event processing.

blockprod/src/interface/blockprod_interface.rs
```diff
@@ -0,0 +1,37 @@
...
+    /// When called, the Block Builder will cancel all current attempts to create blocks
+    /// and won't attempt to do it again for new tips in chainstate or mempool
+    /// Call start() to enable again
+    fn stop_all(&mut self) -> Result<(), BlockProductionError>;
+
+    fn stop_job(&mut self, job_id: JobKey) -> Result<(), BlockProductionError>;
+
+    /// Generate a block with the given transactions to the specified reward destination
+    /// If transactions are None, the block will be generated with available transactions in the mempool
+    /// If submit_to_chainstate is true, the block will be submitted to the chainstate if successfully created
```
azarovh, line 31: `submit_to_chainstate` argument is missing

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +85,566 @@ impl BlockProduction {
...
+
+        let timestamp = BlockTimestamp::from_duration_since_epoch(self.time_getter().get_time());
+
+        let (_, tip_at_start) = self.pull_consensus_data(timestamp).await?;
+
+        {
+            // define the job and insert it into the map of all jobs
+            let job_key = Self::make_job_key(&tip_at_start);
+
+            #[allow(clippy::map_entry)]
```
azarovh, line 228: use `match self.all_jobs.entry(&job_key){...}`?

azarovh (approved):
Few minor comments

blockprod/src/detail/mod.rs
TheQuantumPhysicist: It requires copying the job key. That's why I didn't do it.

blockprod/src/interface/blockprod_interface.rs
TheQuantumPhysicist: Fixed. Thanks.

blockprod/src/detail/mod.rs
azarovh: Using `Entry` should be the same because `VacantEntry::insert` doesn't require a key at all

azarovh: oh right, thanks

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +85,566 @@ impl BlockProduction {
...
-        transactions: Vec<SignedTransaction>,
-    ) -> Result<Block, BlockProductionError> {
-        let (current_tip_id, current_tip_height) = self
+    ) -> Result<Box<dyn TransactionAccumulator>, BlockProductionError> {
+        let max_block_size = self.chain_config.max_block_size_from_txs();
+        let returned_accumulator = self
+            .mempool_handle
+            .call_async(move |mempool| {
+                mempool.collect_txs(Box::new(DefaultTxAccumulator::new(max_block_size)))
+            })
```
iljakuklic, line 137: I think this can be converted to a plain `call` (without async).

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +85,566 @@ impl BlockProduction {
...
+        Ok(consensus_data)
+    }
+
+    fn solve_block(
+        chain_config: Arc<ChainConfig>,
+        mut block_header: BlockHeader,
+        current_tip_height: BlockHeight,
+        stop_flag: Arc<AtomicBool>,
+    ) -> Result<BlockHeader, BlockProductionError> {
+        // TODO: use a separate executor for this loop to avoid starving tokio tasks
```
iljakuklic, line 194: You can use https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html. The task will be allocated to a thread that does not starve other tasks.

iljakuklic (approved):
Looks good. A couple of minor comments below.
Have not yet seen the follow up PR so some of the comments may be irrelevant.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: We're using SlavePool. This is an outdated comment.

blockprod/src/interface/blockprod_interface.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::{detail::JobKey, BlockProductionError};
+
+#[async_trait::async_trait]
+pub trait BlockProductionInterface: Send {
+    /// When called, the Block Builder will cancel all current attempts to create blocks
+    /// and won't attempt to do it again for new tips in chainstate or mempool
+    /// Call start() to enable again
+    fn stop_all(&mut self) -> Result<(), BlockProductionError>;
+
+    fn stop_job(&mut self, job_id: JobKey) -> Result<(), BlockProductionError>;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Alfie is on this in his PR.

blockprod/src/rpc.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Typo, I'd say

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +67,516 @@ impl BlockProduction {
...
+        let timestamp = BlockTimestamp::from_duration_since_epoch(self.time_getter().get_time());
+
+        let (_, tip_at_start) = self.pull_consensus_data(timestamp).await?;
+
+        let (job_key, mut cancel_receiver) =
+            self.job_manager.add_job(tip_at_start.block_id()).await?;
+
+        // At the end of this function, the job has to be removed
+        // Once the job key is used, we swap it with None... this is all temporary.
+        // The the docs in the function self.stop_job_in_scope()
```
alfiedotwtf, line 208: The the

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +67,516 @@ impl BlockProduction {
...
+                        .expect("Failed to send block header back to main thread");
+                });
+            }
+
+            tokio::select! {
+                _ = cancel_receiver.recv() => {
+                    stop_flag.store(true, std::sync::atomic::Ordering::SeqCst);
+
+                    // This can fail if the mining thread has already finished
+                    let _ended = ended_receiver.recv();
```
alfiedotwtf, line 297: I was concered what would happen if we never reach send because of some error in the spawned thread because this `recv()` blocks, but luckily if there's a disconnect then `recv()` returns an error right away and so we won't end up blocking forever.

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +67,516 @@ impl BlockProduction {
...
+            assert!(
+                !job_key_destroyed.load(std::sync::atomic::Ordering::SeqCst),
+                "Must be true as it was done already"
+            );
+            let result_receiver = self.job_manager.stop_job(job_key.clone());
+            let _stop_result = result_receiver.await;
+        }
+        // We consume the value so that it doesn't happen again
+        job_key_destroyed.store(true, std::sync::atomic::Ordering::SeqCst);
+    }
```
alfiedotwtf, line 129: I've gone over this several times, and I think because you're calling this in both branches in the below `select!`, then this should be ok.

alfiedotwtf:
I've been thinking about how to do RAII...

So what we want to do is call `self.job_manager.stop_job(job_key)` from a destructor, but we can't do that because we have to pass in `self` which won't compile because it will have to move. I was thinking - if we break encapsulation just for this one instance, i.e don't pass in `self` but pass in `stop_job_sender` instead, then we can stop the job from the destructor while getting RAII and still be able to compile.

Unless I'm mistaken, here's what seems to be working for me:

```
    let (job_key, stop_job_sender, mut cancel_receiver) =                          
        self.job_manager.add_job(tip_at_start.block_id()).await?;

    let _job_remove_checker = OnceDestructor::new({
        let job_key = job_key.clone();
        || {                         
            let (result_sender, result_receiver) = oneshot::channel();             
            _ = stop_job_sender.send((Some(job_key), result_sender));            
            tokio::spawn(async { result_receiver.await });  
        }
    });                                                                        
```

Now, things to note:
  - I modified add_job() to also return `self.stop_job_sender`
  - Instead of spawning, we can possible do `result_receiver.blocking_recv()`

At least this can be a starting point for ideas that may end up working.

blockprod/src/detail/job_manager/jobs_container.rs
```diff
@@ -0,0 +1,126 @@
...
+        let stop_results =
+            taken_jobs.into_iter().map(|j| j.1.cancel_sender.send(())).collect::<Vec<_>>();
+
+        let send_fail_count = stop_results.into_iter().filter_map(|r| r.ok()).count();
+        if send_fail_count > 0 {
+            log::info!("Sending stop jobs for block production failed for {send_fail_count}");
+        }
+
+        count
+    }
```
alfiedotwtf, line 108: Should this return `count - send_fail_count` instead? Maybe not.

azarovh (approved):
lgtm, a few minor comments

blockprod/src/detail/mod.rs
TheQuantumPhysicist: Yes, it works now fine because it's always called before returning. But again, this is naive C way of doing things. It cannot be guaranteed in every case, which can cause memory leaks. Until we have a better solution... we'll see.

blockprod/src/detail/job_manager/jobs_container.rs
```diff
@@ -0,0 +1,126 @@
...
+            if let Err(e) = result_sender.send(Ok(job_key)) {
+                log::error!("Error sending new job event: {e:?}");
+            }
+        }
+    }
+
+    pub fn handle_chainstate_event(&mut self, new_tip_id: Id<GenBlock>) {
+        let mut jobs_to_stop: Vec<JobKey> = vec![];
+
+        for (job_key, _) in self.jobs.iter() {
```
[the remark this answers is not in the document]
alfiedotwtf: Sure. I tend to prototype using iterative, and then later cleanup into functional (when I remember to do so)

blockprod/src/detail/job_manager/jobs_container.rs
```diff
@@ -0,0 +1,126 @@
...
+                true
+            }
+        }
+    }
+
+    fn stop_all(&mut self) -> usize {
+        let taken_jobs = std::mem::take(&mut self.jobs);
+        let count = taken_jobs.len();
+        let stop_results =
+            taken_jobs.into_iter().map(|j| j.1.cancel_sender.send(())).collect::<Vec<_>>();
```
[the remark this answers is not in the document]
alfiedotwtf: I've changed this to `into_values()`.

blockprod/src/detail/mod.rs
```diff
@@ -47,158 +67,516 @@ impl BlockProduction {
...
-            builder_tx,
+            prepare_thread_pool(1),
         )
-        .expect("Error initializing Block Builder");
+        .expect("Error initializing blockprod");
 
-        block_production.stop().expect("Error stopping Block Builder");
+        let (_other_job_key, _other_job_cancel_receiver) = block_production
+            .job_manager
+            .add_job(Id::new(H256::random_using(&mut make_pseudo_rng())) as Id<GenBlock>)
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yes. We should be using rstest.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: Done

consensus/src/pow/work.rs
```diff
@@ -81,41 +92,44 @@ pub fn check_pow_consensus<H: BlockIndexHandle>(
...
+    Stopped,
+}
+
+impl MiningResult {
+    pub fn is_success(&self) -> bool {
+        *self == Self::Success
+    }
+}
+
+pub fn calculate_work_required<G>(
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Generally speaking, this doesn't have to be called only from that function.

blockprod/src/rpc.rs
```diff
@@ -22,54 +22,59 @@ use common::{
...
-            .map_err(rpc::Error::to_call_error)?;
+        let signed_transactions = match transactions_hex {
+            Some(txs) => Some(
+                txs.into_iter()
+                    .map(SignedTransaction::hex_decode_all)
+                    .collect::<Result<Vec<_>, _>>()
+                    .map_err(rpc::Error::to_call_error)?,
+            ),
+            None => None,
+        };
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I tried it... not a big fan. Harder to read.

blockprod/src/detail/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Fixed

PR #821, opened 2023-04-08
Enforce keeping build version absent in Cargo.toml files

TheQuantumPhysicist:


.github/scripts/codecheck.py
```diff
@@ -83,6 +94,74 @@ def check_workspace_and_package_versions_equal():
...
+    Both dependencies and dev-dependencies have the same structure, so we check the versions the same way for both
+    root_node is the root node of the Cargo.toml file
+    file_path is the path of the Cargo.toml file, for logging
+    '''
+
+    res = True
+
+    # list of crates, whose version can have build version
+    exempted_crates = [
+       # 'ctor'    # left here as an example, remove if you ever add one crated that is exempted
```
muursh, line 108:
```suggestion
       # 'ctor'    # left here as an example, remove if you ever add one crate that is exempt
```

.github/scripts/codecheck.py
```diff
@@ -83,6 +94,74 @@ def check_workspace_and_package_versions_equal():
...
+    ]
+
+    if dependencies_name in root_node:
+        deps = root_node[dependencies_name]
+        for dep in deps:
+            # skip exempted crates
```
muursh, line 114: I'm not convinced by exempted here (or anywhere in this) but I'm struggling to care too much

muursh (approved):

.github/scripts/codecheck.py
TheQuantumPhysicist: I think both are correct, one is passive and the other is an adjective.

azarovh (approved):

.github/scripts/codecheck.py
```diff
@@ -83,6 +94,74 @@ def check_workspace_and_package_versions_equal():
...
+            if dep in exempted_crates:
+                continue
+
+            # versions that look late `tokio = { version = "1.2.3" }`
```
stanislav-tkach, line 118: look like?

.github/scripts/codecheck.py
```diff
@@ -83,6 +94,74 @@ def check_workspace_and_package_versions_equal():
...
+            if 'version' in deps[dep]:
+                version = deps[dep]['version']
+                if len(version.split('.')) > 2:
+                    print("In {}, {} Version '{}' in '{}' has build version".format(dependencies_name, dep, version, file_path))
+                    res = False
+
+            # versions that look late `tokio = "1.2.3"`
```
stanislav-tkach, line 125: Same here.

stanislav-tkach (approved):
I was a little bit confused by the "build version" here because as far as I know it is called `MAJOR.MINOR.PATCH`.

iljakuklic (approved):

.github/scripts/codecheck.py
TheQuantumPhysicist: 3 AM coding :-)

PR #822, opened 2023-04-08
Update cargo

TheQuantumPhysicist:
Unfortunately this has caused an issue in rpc's subsystem shutdown in tests. I don't know why, but I disabled the shutdown as it's unnecessary.

I also had to do modifications to the syn library for parsing Rust code for the Tagged struct, which is used to version blocks and transactions.

### DIFF
--- a/rpc/src/lib.rs
+++ b/rpc/src/lib.rs
@@ -236,8 +236,7 @@ mod tests {
         let response: Result<u64> = client.request("some_subsystem_add", rpc_params!(2, 5)).await;
         assert_eq!(response.unwrap(), 7);
 
-        // TODO: after a cargo update, this test freezes on shutdown, so it's commented out
-        // subsystem::Subsystem::shutdown(rpc).await;
+        subsystem::Subsystem::shutdown(rpc).await;
         Ok(())
     }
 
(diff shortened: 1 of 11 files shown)


muursh (approved):

stanislav-tkach (approved):

azarovh (approved):

rpc/src/lib.rs
```diff
@@ -236,7 +236,8 @@ mod tests {
         let response: Result<u64> = client.request("some_subsystem_add", rpc_params!(2, 5)).await;
         assert_eq!(response.unwrap(), 7);
 
-        subsystem::Subsystem::shutdown(rpc).await;
+        // TODO: after a cargo update, this test freezes on shutdown, so it's commented out
+        // subsystem::Subsystem::shutdown(rpc).await;
```
iljakuklic, line 240: I'm not sure what the underlying reason for this is. However, it is notable that the test creates the RPC sysbsystem without adding it to the sybsystem manager. That is not a supported way of using subsystems. I'd suggest creating a manager at the start of the test and accessing RPC through it. That way the shutdown will be handled in the usual way when the manager terminates.

iljakuklic (approved):

rpc/src/lib.rs
TheQuantumPhysicist: Should we care and fix it? Or just keep it and move on?

pavel-kokolemin: Looks like it's a known issue: https://github.com/paritytech/jsonrpsee/pull/1062
The master branch is fixed (I just checked our tests with it), but there is no new release yet. So we can keep using tokio 1.26 for now or switch to the jsonrpsee master branch.

iljakuklic (approved):

PR #824, opened 2023-04-10
Change default directory on Windows and macOS

pavel-kokolemin:
I noticed that the node creates its own files in the home directory by default. I don't think this is a very good style, especially on Windows and macOS. Let's do it like Bitcoin Core and use these directories:

Windows: C:\Users\Username\AppData\Roaming\Mintlayer
macOS: ~/Library/Application Support/Mintlayer
Unix-like: ~/.mintlayer


muursh (approved):
Fine by me. Seems sensible

iljakuklic (approved):

stanislav-tkach (approved):

TheQuantumPhysicist (approved):

PR #825, opened 2023-04-11
Minor updates in address list handling (p2p)

pavel-kokolemin:
Process address list requests just once and only from inbound peers (same way it's done in Bitcoin Core).

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -37,7 +37,7 @@ use common::{
     time_getter::TimeGetter,
 };
 use logging::log;
-use utils::{bloom_filters::rolling_bloom_filter::RollingBloomFilter, ensure};
+use utils::{bloom_filters::rolling_bloom_filter::RollingBloomFilter, ensure, set_flag::SetFlag};
 
 use crate::{
     config::P2pConfig,
@@ -463,6 +463,18 @@ where
         }
     }
 
+    /// Should we load addresses from this peer?
+    fn load_addresses_from(role: Role) -> bool {
+        // Load addresses only from outbound peers, like it's done in Bitcoin Core
+        role == Role::Outbound
+    }
+
+    /// Should we send addresses to this peer if it requests them?
+    fn send_addresses_to(role: Role) -> bool {
+        // Send addresses only to inbound peers, like it's done in Bitcoin Core
+        role == Role::Inbound
+    }
+
     /// Try accept new connection
     ///
     /// The event is received from the networking backend and it's either a result of an incoming
@@ -487,11 +499,7 @@ where
             self.subscribed_to_peer_addresses.insert(info.peer_id);
         }
 
-        // For security reasons, the address list is only requested from outbound peers.
-        // (inbound peers are generally less trustworthy than outbound peers).
-        let expect_addr_list_response = role == Role::Outbound;
-
-        if expect_addr_list_response {
+        if Self::load_addresses_from(role) {
             Self::send_peer_message(
                 &mut self.peer_connectivity_handle,
                 peer_id,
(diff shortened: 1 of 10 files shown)


stanislav-tkach (approved):

p2p/src/peer_manager/peer_context.rs
```diff
@@ -50,8 +50,11 @@ pub struct PeerContext<A> {
     /// Min ping time
     pub ping_min: Option<Duration>,
 
-    /// Whether an address list request was sent and no response was received.
-    ///
+    /// Whether to process an address list request from this peer.
+    /// If false, the request is ignored.
+    pub expect_addr_list_request: bool,
```
TheQuantumPhysicist, line 55: Shouldn't we use SetFlag for this?

p2p/src/peer_manager/mod.rs
```diff
@@ -790,7 +795,14 @@ where
         }
     }
 
-    fn handle_add_list_request(&mut self, peer: PeerId) {
+    fn handle_add_list_request(&mut self, peer_id: PeerId) {
+        let peer = self.peers.get_mut(&peer_id).expect("peer must be known");
+        if !peer.expect_addr_list_request {
+            log::debug!("Ignore unexpected address list request from peer {peer_id}");
+            return;
```
TheQuantumPhysicist, line 802: No punishment?

pavel-kokolemin: Yes, the same way it's done in Bitcoin Core. I don't see any risks with that.

p2p/src/peer_manager/mod.rs
```diff
@@ -790,7 +795,14 @@ where
...
     }
 
-    fn handle_add_list_request(&mut self, peer: PeerId) {
+    fn handle_add_list_request(&mut self, peer_id: PeerId) {
+        let peer = self.peers.get_mut(&peer_id).expect("peer must be known");
+        if *peer.ignore_addr_list_request {
+            log::warn!("Ignore unexpected address list request from peer {peer_id}");
+            return;
+        }
+        peer.ignore_addr_list_request.set();
```
TheQuantumPhysicist, line 804: Two thing I can thing to improve this:

1. This can be isolated into its own function
2. SetFlag can have a test_and_set method that both returns true and sets the value similar to the way C++ does it for atomics: https://cplusplus.com/reference/atomic/atomic_flag/test_and_set/

If No. 2 is implemented, No. 1 isn't necessary. The problem with its current form is that it's kind of weird how it's done, where we check then manually set a value. The whole degrees-of-freedom discussion can be brought up again.

utils/src/set_flag.rs
```diff
@@ -32,3 +32,9 @@ impl Deref for SetFlag {
         &self.0
     }
 }
+
+impl From<bool> for SetFlag {
+    fn from(value: bool) -> Self {
```
TheQuantumPhysicist, line 37: This is problematic. There should be only one way to initialize this, then one way to set it, such that it's not used twice. This can ruin setting the value due to confusion.

In other words: This should always be initialized unset, then with a call becomes set.

p2p/src/peer_manager/peer_context.rs
```diff
@@ -50,10 +50,11 @@ pub struct PeerContext<A> {
...
 
-    /// Whether an address list request was sent and no response was received.
-    ///
-    /// It is used to score peers that send unsolicited address list responses.
-    pub expect_addr_list_response: bool,
+    /// If set, address list requests are ignored from this peer
+    pub ignore_addr_list_request: SetFlag,
+
+    /// If set, address list responses are ignored and the peer is scored
+    pub ignore_addr_list_response: SetFlag,
```
TheQuantumPhysicist, line 57: Can we add tests to ensure the behavior is correct?

TheQuantumPhysicist (approved):

PR #828, opened 2023-04-12
Randomize reconnect delays (p2p)

pavel-kokolemin:
Right now p2p reconnect logic uses fixed delays. This can cause issues if there is something wrong with the host's Internet connection. If all peers are disconnected at the same time, then later reconnect attempts will be highly correlated in time. Applying some randomization should help. Sampling from the exponential distribution is used for this (Bitcoin Core uses it).
Exponential distribution sampler can be found in the `rand_distr` crate. I decided to not to introduce an extra dependency just for this. At first I copied this function from Bitcoin Core:
```
    double unscaled = -std::log1p(GetRand(uint64_t{1} << 48) * -0.0000000000000035527136788 /* -1/2^48 */);
```
But then I noticed a better looking variant in the `rand_distr` crate:
```
pub fn exponential_rand(rng: &mut impl Rng) -> f64 {
    -rng.gen::<f64>().ln()
}
```

### DIFF
--- a/p2p/src/peer_manager/peerdb/address_data/mod.rs
+++ b/p2p/src/peer_manager/peerdb/address_data/mod.rs
@@ -17,12 +17,21 @@ use std::time::Duration;
 
 use crypto::random::Rng;
 
+/// Maximum delay between reconnection attempts to reserved nodes
+const MAX_DELAY_RESERVED: Duration = Duration::from_secs(360);
+
+/// Maximum delay between reconnection attempts to previously reachable nodes
+const MAX_DELAY_REACHABLE: Duration = Duration::from_secs(3600);
+
 /// When the node drops the unreachable node address. Used for negative caching.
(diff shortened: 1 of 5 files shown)


TheQuantumPhysicist (approved):
Looking good. But can you please add tests for the connection behavior?

pavel-kokolemin:
I think there is a problem with using large delays between reconnection attempts to once reachable nodes (currently 24 hours).
Imagine a situation where the network is small and has only 10 reachable nodes and the host's internet connection is down for some reason.
In such a case, after some time, all peers will be marked as offline and `fail_count` will grow. And when the Internet connection is restored, it may take a long time before the node tries to reconnect to any peer (24 hours / 10 ~ 2.4 hours). To avoid this, I have lowered the maximum reconnect delay to reachable nodes to 1 hour.

stanislav-tkach (approved):

PR #833, opened 2023-04-14
Remove amount from TxOutput::ProduceBlockFromStake

azarovh:
The high-level motivation for this PR is to move staked coins arithmetics from utxo model to accounting. Basically once staked the manipulations like reward distribution to staker and delegators can be performed inside accounting only.

### DIFF
--- a/common/src/chain/transaction/output/mod.rs
+++ b/common/src/chain/transaction/output/mod.rs
@@ -55,7 +55,7 @@ pub enum TxOutput {
     /// Output type that represents spending of a stake pool output in a block reward
     /// in order to produce a block
     #[codec(index = 4)]
-    ProduceBlockFromStake(Amount, Destination, PoolId),
+    ProduceBlockFromStake(Destination, PoolId),
     #[codec(index = 5)]
     DecommissionPool(Amount, Destination, PoolId, OutputTimeLock),
 }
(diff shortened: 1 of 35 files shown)


common/src/chain/transaction/output/mod.rs
```diff
@@ -79,21 +79,21 @@ impl TxOutput {
...
-            | TxOutput::ProduceBlockFromStake(_, _, _) => false,
+            | TxOutput::ProduceBlockFromStake(_, _) => false,
             TxOutput::Burn(_) => true,
         }
     }
 }
 
 impl TxOutput {
-    pub fn value(&self) -> OutputValue {
+    pub fn value(&self) -> Option<OutputValue> {
```
azarovh, line 89: This change bothers me because making it `Option` doesn't provide enough information for the outside code on how to treat it. Is None equivalent to No-op, zero or an Err?

Particularly in `mempool/try_get_fee.rs` it seems unclear what to do with it and very plausible just to ignore it with unclear consequences for the fee calculation. 

I tend to think that this method should go and it will be impossible to get output value without the context (i.e. without pos_accounting information)

azarovh: @TheQuantumPhysicist

TheQuantumPhysicist: Yep. This should go. Just like the destination. I just kept it because I didn't want to introduce functionality change. Now it's time to remove it.

TheQuantumPhysicist (approved):
Looking good. I don't see anything exceptional to even review.

Btw, there's a typo "transacitons" somewhere. Please fix it.

muursh (approved):

stanislav-tkach (approved):

PR #835, opened 2023-04-15
Fix `p2p_syncing_test.py` test failure

pavel-kokolemin:
Do not reconnect to peers that have been manually disconnected (from RPC commands).

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -368,6 +368,8 @@ where
     /// The decision to close the connection is made either by the user via RPC
     /// or by the [`PeerManager::heartbeat()`] function which has decided to cull
     /// this connection in favor of another potential connection.
+    ///
+    /// If the `response` channel is not empty, the peer is marked as disconnected by the user and no reconnect attempts are made.
     fn disconnect(
         &mut self,
         peer_id: PeerId,
@@ -618,15 +620,22 @@ where
                 peer.address
             );
 
-            if let Some(Some(response)) = self.pending_disconnects.remove(&peer_id) {
+            let resp_ch = self.pending_disconnects.remove(&peer_id).flatten();
+
+            if peer.role == Role::Outbound {
+                // If `resp_ch` is some, the peer is disconnected after the RPC command
+                if resp_ch.is_some() {
+                    self.peerdb.outbound_peer_disconnected_by_user(peer.address);
+                } else {
+                    self.peerdb.outbound_peer_disconnected(peer.address);
+                }
+            }
+
+            if let Some(response) = resp_ch {
                 response.send(Ok(()));
             }
 
             self.subscribed_to_peer_addresses.remove(&peer_id);
-
-            if peer.role == Role::Outbound {
-                self.peerdb.outbound_peer_disconnected(peer.address);
-            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


p2p/src/peer_manager/peerdb/address_data/mod.rs
```diff
@@ -105,7 +110,8 @@ impl AddressData {
                 fail_count: _,
                 next_connect_after,
                 was_reachable: _,
-            } => now >= next_connect_after,
+                disconnected_by_user,
+            } => now >= next_connect_after && !disconnected_by_user,
```
iljakuklic, line 114: A brief comment why this condition was introduced would be nice

iljakuklic (approved):

p2p/src/peer_manager/peerdb/address_data/mod.rs
```diff
@@ -47,6 +47,8 @@ pub enum AddressState {
 
         /// Next time connect to the peer
         next_connect_after: Duration,
+
+        disconnected_by_user: bool,
```
TheQuantumPhysicist, line 51: Please add documentation explaining why this is introduced and how it helps.

p2p/src/peer_manager/peerdb/address_data/mod.rs
```diff
@@ -152,7 +160,11 @@ impl AddressData {
             MAX_DELAY_REACHABLE
         };
 
-        std::cmp::min(fail_count * Duration::from_secs(60), max_delay)
+        // 10, 20, 40, 80... seconds
+        std::cmp::min(
+            2u32.saturating_pow(fail_count) * Duration::from_secs(10),
```
TheQuantumPhysicist, line 165: This number can easily overflow... fail_count only has to be > 32 for that to happen, with or without multiplication... please guard the operations with whatever necessary to ensure this can't happen.

p2p/src/peer_manager/mod.rs
```diff
@@ -618,15 +618,21 @@ where
...
+            let resp_ch = self.pending_disconnects.remove(&peer_id).flatten();
+
+            if peer.role == Role::Outbound {
+                if resp_ch.is_some() {
+                    self.peerdb.outbound_peer_disconnected_by_user(peer.address);
+                } else {
+                    self.peerdb.outbound_peer_disconnected(peer.address);
+                }
+            }
+
```
TheQuantumPhysicist, line 630: Can you please explain the logic here further?

p2p/src/peer_manager/peerdb/address_data/mod.rs
pavel-kokolemin: It should not overflow because seconds in `Duration` are represented as u64. Replaced by `saturated_mul` anyway.

p2p/src/peer_manager/peerdb/address_data/mod.rs
```diff
@@ -47,6 +47,10 @@ pub enum AddressState {
 
         /// Next time connect to the peer
         next_connect_after: Duration,
+
+        /// It's a flag for outbound peers that have been manually disconnected
```
TheQuantumPhysicist, line 51: Disconnect is done by requesting to do it. This flag ensures that p2p will not try to reconnect to the target node.

TheQuantumPhysicist (approved):
After adding tests, like we discussed, feel free to merge.

stanislav-tkach (approved):

PR #836, opened 2023-04-16
Use handles for wallet-node calls

TheQuantumPhysicist:
If we're gonna introduce a GUI node, the wallet should be part of it, since GUI can handle parallelism very easily. This PR makes the wallet communicate with the node directly without the need for RPC. It adds another NodeInterface implementation that uses subsystem handles instead of RPC calls.


pavel-kokolemin:
I don't like this. It's more code that isn't really needed. We can start the node inside the GUI process if we want (although I'm not sure that's a good idea) and still use RPC. And less code means fewer bugs.

TheQuantumPhysicist:
> I don't like this. It's more code that isn't really needed. We can start the node inside the GUI process if we want (although I'm not sure that's a good idea) and still use RPC. And less code means less bugs. Also, using RPC from the GUI allows us to connect to remote nodes, which is nice to have.

It's true that more code is worse. Granted. But there's an implicit rule that's more important than this: Do not make your code abstractions create unnecessary physical interfaces. For example, if you have abstractions that open files, ports, or any kind of "physical" abstractions, it's a bad idea to keep them just to save your abstractions when unnecessary. The reason for that, besides performance, is that it now starts to increase your error surface. 

Besides this, notice that the "extra code" is barely extra code. It's barely direct calls to chainstate through subsystems. There's no new logic, really.

pavel-kokolemin (approved):

stanislav-tkach (approved):

PR #838, opened 2023-04-17
Minor fixes for #820 PR

alfiedotwtf:


blockprod/src/detail/mod.rs
```diff
@@ -526,7 +592,7 @@ mod tests {
         );
 
         assert!(
-            other_job_cancel_receiver.try_recv().unwrap().is_none(),
+            other_job_cancel_receiver.try_recv().is_err(),
```
TheQuantumPhysicist, line 595: I think is_err is not equivalent to unwrap().is_none(). The TryRecvError can be either Closed or Empty, and it makes a difference which is the current state.

alfiedotwtf: Ah of course. Ok, will update

blockprod/src/detail/mod.rs
```diff
@@ -306,33 +310,95 @@ impl BlockProduction {
...
-                    // TODO: test cancellations
-                    stop_flag.store(true, std::sync::atomic::Ordering::SeqCst);
-                    return Err(BlockProductionError::Cancelled);
+            'consume_chainstate_events: loop {
+                tokio::select! {
+                    _ = chainstate_receiver.changed() => {
+                        // TODO: test broadcasts
+                        self.broadcast_new_tip(*chainstate_receiver.borrow());
+                        continue 'consume_chainstate_events;
+                    }
```
TheQuantumPhysicist, line 319: This is not really the appropriate place for this. Receiving chainstate events shouldn't be done in generate_block, but in a possible event loop related to block production itself.

So what you're doing is: ChainstateEvent -> generate block -> broadcast_new_tip -> generate_block -> stop
What should be done is: ChainstateEvent -> BlockProduction -> broadcast_new_tip -> generate_block -> stop

There are many reasons for why this should be done like this, including circularity of the current logic. I think the confusion here is that you need an event loop... well... start a new task that receives chainstate events! That's how you create an event loop.

So once you connect chainstate's new tip with the function broadcast_new_tip, your job is basically done, because generate_block already has a cancel_receiver that solves that problem to you. Notice that part of the design flaw here is that you have to get through the `select!` macro twice.

alfiedotwtf: Updated to move into its own `run()` function called by the node runner

blockprod/src/detail/mod.rs
```diff
@@ -66,7 +69,141 @@ pub struct JobKey {
...
-    cancel_signal: oneshot::Sender<()>,
+    cancel_sender: UnboundedSender<()>,
+}
+
+pub struct JobManager {
+    chainstate_handle: ChainstateHandle,
+    new_job_sender: UnboundedSender<(Id<GenBlock>, UnboundedSender<()>)>,
+    stop_job_sender: UnboundedSender<(Option<JobKey>, oneshot::Sender<bool>)>,
+    all_jobs: BTreeMap<JobKey, JobHandle>,
+}
```
TheQuantumPhysicist, line 80: When you're done with your impl, please move this to `job.rs`

alfiedotwtf: Yep cool, once it was all going, moving it was the last step.

blockprod/src/detail/mod.rs
```diff
@@ -66,7 +69,141 @@ pub struct JobKey {
...
+
+                    #[allow(clippy::map_entry)]
+                    if !all_jobs.contains_key(&job_key) {
+                        all_jobs.insert(
+                            job_key,
+                            JobHandle {
+                                cancel_sender: new_job.1,
+                            },
+                        );
+                    }
```
TheQuantumPhysicist, line 145: You can return an error in a oneshot passed inside the new_job_receiver channel if the job is already there.

blockprod/src/detail/mod.rs
```diff
@@ -66,7 +69,141 @@ pub struct JobKey {
 }
 
 pub struct JobHandle {
-    cancel_signal: oneshot::Sender<()>,
+    cancel_sender: UnboundedSender<()>,
+}
+
+pub struct JobManager {
```
TheQuantumPhysicist, line 75: Perhaps you can implement the drop trait of the JobManager to stop everything.

blockprod/src/detail/mod.rs
```diff
@@ -66,7 +69,141 @@ pub struct JobKey {
...
+    }
+
+    fn run(
+        &mut self,
+        mut chainstate_receiver: UnboundedReceiver<Id<GenBlock>>,
+        mut new_job_receiver: UnboundedReceiver<(Id<GenBlock>, UnboundedSender<()>)>,
+        mut stop_job_receiver: UnboundedReceiver<(Option<JobKey>, oneshot::Sender<bool>)>,
+    ) {
+        tokio::spawn(async move {
+            let mut all_jobs: BTreeMap<JobKey, JobHandle> = BTreeMap::new();
```
TheQuantumPhysicist, line 112: While this works for the time being, it'll probably cause an issue later when you try to stop everything on exit. But it's a good start.

blockprod/src/detail/mod.rs
alfiedotwtf: Yep, sounds good. Just calling `stop_job(None)` in `drop()` should do the trick

blockprod/src/rpc.rs
```diff
@@ -22,18 +22,18 @@ use common::{
...
 trait BlockProductionRpc {
     /// Stop block production
     #[method(name = "stop_all")]
-    async fn stop_all(&self) -> rpc::Result<()>;
+    async fn stop_all(&self) -> rpc::Result<usize>;
 
     // Stop a specific job
     #[method(name = "stop_job")]
-    async fn stop_job(&self, job_id: JobKey) -> rpc::Result<()>;
+    async fn stop_job(&self, job_id: String) -> rpc::Result<bool>;
```
iljakuklic, line 36: Maybe also something to query if given job is running? Is that possible and useful?

blockprod/src/lib.rs
```diff
@@ -145,9 +149,14 @@ mod tests {
...
             blockprod
-                .call_mut(move |this: &mut Box<dyn BlockProductionInterface>| {
-                    assert!(this.stop_all().is_ok(), "Failed to stop non-existent jobs");
-                    shutdown.initiate();
+                .call_async_mut(move |this| {
+                    Box::pin(async move {
+                        assert!(
+                            this.stop_all().await == Ok(0),
+                            "Failed to stop non-existent jobs"
+                        );
```
iljakuklic, line 157: Not too keen on side effects inside an assert. I'd prefer to run stop_all in a seperate step and let the assert just deal with the result value. Also `assert_eq!` can be used here.

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+        if let Err(e) = self.shutdown_sender.send(result_sender) {
+            log::info!("Error sending shutdown during job manager drop: {e:?}");
+            return;
+        }
+
+        tokio::spawn(async move {
+            tokio::select! {
+                _ = result_receiver => {},
+            }
+        });
```
iljakuklic, line 356: What's the purpose of this extra task? Would just closing the receiver work too?

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+                jobs.insert(job_key.clone(), JobHandle { cancel_sender });
+
+                if let Err(e) = result_sender.send(Ok(job_key)) {
+                    log::info!("Error sending new job event: {e:?}");
+                }
+            }
+        }
+    }
+
+    pub async fn stop_job(&mut self, job_key: Option<JobKey>) -> Result<usize, JobManagerError> {
```
iljakuklic, line 282: Not a fan of the `Option` argument. The call `man.stop_job(None)` reads like you're not stopping any job.

I can see two possible ways to improve it:
* Either make this a private method (something like `stop_job_internal`) and have `stop_job` and `stop_jobs` wrappers, or
* Define a custom enum in place of `Option`, something like `enum JobSelector { Exact(JobKey), All }`

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+
+        if self.stop_job_sender.send((job_key, result_sender)).is_err() {
+            return Err(JobManagerError::FailedToSendStopJobEvent);
+        }
+
+        tokio::select! {
+            result = result_receiver => match result {
+                Ok(v) => Ok(v),
+                Err(_) => Err(JobManagerError::FailedToStopJobs),
+            }
```
iljakuklic, line 293: Maybe this? There are other occurrences of this pattern. (not tested)
```suggestion
            result = result_receiver => result.map_err(|_| JobManagerError::FailedToStopJobs),
```

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+                .call_mut(|this| {
+                    let subscribe_func =
+                        Arc::new(
+                            move |chainstate_event: ChainstateEvent| match chainstate_event {
+                                ChainstateEvent::NewTip(block_id, _) => {
+                                    if let Err(e) = chainstate_sender.send(block_id.into()) {
+                                        log::error!(
+                                            "Chainstate subscriber failed to send new tip: {e:?}"
+                                        )
+                                    }
```
iljakuklic, line 176: We have `log_err_pfx` in the `utils` crate for this. More occurrences below. Your call whether it's an improvement or not.

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+            for (job_key, _) in jobs.iter() {
+                if job_key.current_tip_id() != new_tip_id {
+                    jobs_to_stop.push(job_key.clone());
+                }
+            }
+
+            for job_key in jobs_to_stop {
+                if let Some(handle) = jobs.remove(&job_key) {
+                    if let Err(e) = handle.cancel_sender.send(()) {
+                        log::info!("Error sendin cancel job event: {e:?}");
```
iljakuklic, line 200:
```suggestion
                        log::info!("Error sending cancel job event: {e:?}");
```

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+        if self.new_job_sender.send((block_id, cancel_sender, result_sender)).is_err() {
+            return Err(JobManagerError::FailedToSendNewJobEvent);
+        }
+
+        tokio::select! {
+            result = result_receiver => match result {
+                Ok(v) => match v {
+                    Ok(v) => Ok((v, cancel_receiver)),
+                    Err(e) => Err(e),
+                },
```
iljakuklic, line 250: Possibly something like
```suggestion
                Ok(v) => v.map(move |v| (v, cancel_receiver)),
```

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    #[allow(clippy::type_complexity)]
+    fn new_job_handler(
+        jobs: &mut BTreeMap<JobKey, JobHandle>,
+        event: Option<(
+            Id<GenBlock>,
+            UnboundedSender<()>,
+            oneshot::Sender<Result<JobKey, JobManagerError>>,
+        )>,
```
iljakuklic, line 263: Maybe package up the inner tuple into a struct?

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+            match job_key {
+                Some(job_key) => {
+                    if let Some(job_handle) = jobs.remove(&job_key) {
+                        stop_jobs.push((job_key, job_handle));
+                    }
+                }
+                None => {
+                    while let Some((job_key, job_handle)) = jobs.pop_first() {
+                        stop_jobs.push((job_key, job_handle));
+                    }
```
iljakuklic, line 313: You can `std::mem::take` the `jobs` object to take ownership and then `.into_iter().collect()`

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+                    log::info!("Cancelling {} jobs", stop_jobs.len());
+                }
+            }
+
+            let stop_count = stop_jobs.len();
+
+            for (job_key, job_handle) in stop_jobs.drain(..) {
+                let _ = job_handle.cancel_sender.send(());
+                log::info!("Stopped mining job for tip {}", job_key.current_tip_id());
+            }
```
iljakuklic, line 324: Wondering if this should wait for the jobs to actually stop?

iljakuklic (commented):
Some mostly minor comments below:

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    pub async fn stop_job(&mut self, job_key: Option<JobKey>) -> Result<usize, JobManagerError> {
+        let (result_sender, result_receiver) = oneshot::channel();
+
+        if self.stop_job_sender.send((job_key, result_sender)).is_err() {
+            return Err(JobManagerError::FailedToSendStopJobEvent);
+        }
+
+        tokio::select! {
```
TheQuantumPhysicist, line 289: You don't need a `select!` for result_receiver. It's a future. You can await on it directly.

blockprod/src/detail/job_manager.rs
TheQuantumPhysicist: Perhaps the easiest way to achieve this is to return a receiver of a oneshot that'll confirm when the job is stopped, which can be done asynchronously. The question is how much cascading effect this will have all the way down to the generator of blocks.

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+            }
+        }
+    }
+
+    fn shutdown_handler(
+        jobs: &mut BTreeMap<JobKey, JobHandle>,
+        event: Option<oneshot::Sender<usize>>,
+    ) {
+        if let Some(result_sender) = event {
+            log::info!("Stopping job manager");
```
TheQuantumPhysicist, line 337: "Stopping block production job manager"

Remember, this is global logging.

blockprod/src/detail/job_manager.rs
TheQuantumPhysicist: I'm guessing here... I think the purpose is to await on the result_receiver so that it eventually closes. The right way to do this, assuming having the task is OK, like I mentioned in another comment, is to await on result_receiver because it's a future, which is what `select!` will do. Besides that, the problem here is that fn drop is not async fn, which means you can't await. So Alfie found the only way to do this is by spawning an async task to do this, hoping that before the program shuts down, all tasks will be collected. Unlikely, but there's no better way I see.

To be fair, in a world where async makes sense, unlike what rust async has become, a blocking await would be the right way to go... but I think blocking-await can cause deadlocks... so here we are. I'm always afraid of suggesting blocking actions in async.

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+}
+
+impl Drop for JobManager {
+    fn drop(&mut self) {
+        let (result_sender, result_receiver) = oneshot::channel();
+
+        if let Err(e) = self.shutdown_sender.send(result_sender) {
+            log::info!("Error sending shutdown during job manager drop: {e:?}");
```
TheQuantumPhysicist, line 348: error

blockprod/src/detail/mod.rs
```diff
@@ -202,41 +171,19 @@ impl BlockProduction {
...
-                    JobHandle {
-                        cancel_signal: cancel_sender,
-                    },
-                );
-            } else {
-                return Err(BlockProductionError::JobAlreadyExists(job_key));
-            }
-        }
+        let (_job_key, mut cancel_receiver) =
+            self.job_manager.new_job(tip_at_start.block_id()).await?;
```
TheQuantumPhysicist, line 186: I think this function is incomplete. By the end of this function, this key should be removed from the job_manager.

There are multiple ways to do this. 

There's the naive C way of doing this, where we call this function before returning, and hope no one in the future will change the function to return before and cause a memory leak.

There's the smart way to do this, where we create an RAII wrapper, which will work on deleting the job from the queue automatically when this scope finishes.

TheQuantumPhysicist:
Very nice. Getting very close to a good version to merge.

blockprod/src/detail/job_manager.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    pub async fn stop_all_jobs(&mut self) -> Result<usize, JobManagerError> {
+        self._stop_job(None).await
+    }
+
+    pub async fn stop_job(&mut self, job_key: JobKey) -> Result<usize, JobManagerError> {
+        self._stop_job(Some(job_key)).await
+    }
+
+    pub async fn _stop_job(&mut self, job_key: Option<JobKey>) -> Result<usize, JobManagerError> {
```
TheQuantumPhysicist, line 262: 1. I think this should be private
2. Please call it `stop_job_internal`

blockprod/src/detail/mod.rs
alfiedotwtf: This was not as straight forward as I hoped, as noticed in a comment below...

blockprod/src/rpc.rs
alfiedotwtf: Sounds good, but I've held off until PoS is implemented, because that's the one that's going to drive the API for this possible call

blockprod/src/lib.rs
alfiedotwtf: Sounds fair. I've moved this out but forgot to clean up the others. I'll clean these all up soon.

blockprod/src/detail/job_manager.rs
alfiedotwtf: Yep. As above, this wasn't as straight forward so had to jump through some hoops. Sam's doing some changes now, but if he hasn't worked out how to do RAII, I have a drop-esque work around that seems to work as I wanted.

blockprod/src/detail/job_manager.rs
alfiedotwtf: Yep, you're right. I've opted for the first (but forgot to make it `private`...

blockprod/src/detail/job_manager.rs
alfiedotwtf: Yep, nice. I tried to do that the other day in some places but for some reason kept getting `select!` parsing issues, but I tried again today and it seems to be working. So this is now done.

blockprod/src/detail/job_manager.rs
alfiedotwtf: Nice. I've updated everywhere where it worked.

blockprod/src/detail/job_manager.rs
alfiedotwtf: I recently switched to MX Cherry Green switches, and I'm missing a lot a ',', 'o', and 'g'. This has been fixed, and am now trying to type harder to avoid misses.

blockprod/src/detail/job_manager.rs
alfiedotwtf: Nice. This is now done.

blockprod/src/detail/job_manager.rs
alfiedotwtf: I've instead created a new type, and removed all the clippy workarounds

blockprod/src/detail/job_manager.rs
alfiedotwtf: This actually worked. Cool :)

blockprod/src/detail/job_manager.rs
alfiedotwtf: Still not sure what to do here for this one

blockprod/src/detail/job_manager.rs
alfiedotwtf: This was a forrest from the trees moment. Yeah, that's obvious now. `select!` has been removed everywhere when there's only a single branch

blockprod/src/detail/job_manager.rs
alfiedotwtf: Sorry, not sure what this comment meant. FWIW, we want to log that's we're in shutdown, so globally logging this isn't an issue

blockprod/src/detail/job_manager.rs
alfiedotwtf: Yep, you're right. I've updated everywhere where `info!` should have been `error!`

blockprod/src/detail/job_manager.rs
alfiedotwtf: Ah, I see now

TheQuantumPhysicist:
I'll merge this one into #820. Let's move the discussion there, since most of the issues here were solved.

PR #839, opened 2023-04-17
Added wallet key pools, lookahead and improved the code structe

erasmospunk:


crypto/src/key/mod.rs
```diff
@@ -48,7 +48,7 @@ pub struct PrivateKey {
     key: PrivateKeyHolder,
 }
 
-#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Decode, Encode)]
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Hash, Clone, Decode, Encode)]
```
TheQuantumPhysicist, line 51: We don't use `Hash` at all to avoid confusion with proper cryptography hashes. In practice, there's barely any need for this. If you need a HashSet or HashMap, please use BTreeSet or BTreeMap instead.

crypto/src/key/key_holder.rs
```diff
@@ -25,7 +25,7 @@ pub enum PrivateKeyHolder {
     Secp256k1Schnorr(Secp256k1PrivateKey),
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Decode, Encode)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Decode, Encode)]
```
TheQuantumPhysicist, line 28: Same here. Remove Hash please.

crypto/src/key/secp256k1/mod.rs
```diff
@@ -96,7 +96,7 @@ impl Secp256k1PrivateKey {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
```
TheQuantumPhysicist, line 99: Remove Hash

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -54,6 +54,23 @@ impl DerivationPath {
     pub fn is_root(&self) -> bool {
         self.is_empty()
     }
+
+    /// Get the difference of this path and a sub path.
+    pub fn get_super_path_diff(&self, sub_path: &DerivationPath) -> Option<&[ChildNumber]> {
+        let super_path_vec = self.as_vec();
```
TheQuantumPhysicist, line 60: Not related to this PR, but I just noticed this now. Isn't it better to make the `as_vec` method an `as_slice` and return `&[ChildNumber]` instead?

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -134,6 +151,20 @@ mod tests {
     use super::*;
     use test_utils::{assert_encoded_eq, decode_from_hex};
 
+    #[test]
+    fn path_diff() {
```
TheQuantumPhysicist, line 155: Please, without me having to ask for this, tests can go in two ways:

1. Exhaustive manually (something like parse_derivation_path() test below)
2. Exhaustive through randomization/propery-based testing, like most of the tests in mintlayer-core.

I'm frustrated.

TheQuantumPhysicist: To be clear, this isn't just bad because it breaks the bylaws of the code, but it also omits success/error cases + omits corner cases (empty vecs, etc).

wallet/types/src/account_id.rs
```diff
@@ -0,0 +1,62 @@
...
+            item_id,
+        }
+    }
+
+    pub fn into_item_id(self) -> Id {
+        self.item_id
+    }
+}
+
+impl<ID: Encode> HasPrefix<AccountId> for AccountPrefixedId<ID> {}
```
TheQuantumPhysicist, line 57: Id, as per rust style guidelines.

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,94 @@
...
+        self.last_used = new_last_used;
+    }
+
+    /// Get the last index issued to the user
+    pub fn get_last_issued(&self) -> Option<ChildNumber> {
+        self.last_issued
+    }
+
+    /// Set the last index issued to the user
+    pub fn set_last_issued(&mut self, new_last_issued: Option<ChildNumber>) {
```
TheQuantumPhysicist, line 99: If the only way we use this is by incrementing the value, why does this struct expose a free functionality to do anything with the number? Why not just increment it here, which guarantees the behavior to external users of this module?

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,94 @@
...
+        }
+    }
+
+    /// Get the last index used in the blockchain
+    pub fn get_last_used(&self) -> Option<ChildNumber> {
+        self.last_used
+    }
+
+    /// Set the last index used in the blockchain
+    pub fn set_last_used(&mut self, new_last_used: Option<ChildNumber>) {
```
TheQuantumPhysicist, line 89: Same here. There are fixed rules to how the value is set. Why don't we incorporate that logic in here instead of making this struct act like another layer of storage with set/get?

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,94 @@
...
+    pub fn get_deterministic_index(&self) -> ChildNumber {
+        match self {
+            ReceiveFunds => Self::RECEIVE_FUNDS_INDEX,
+            Change => Self::CHANGE_INDEX,
+        }
+    }
+}
+
+impl TryFrom<ChildNumber> for KeyPurpose {
+    type Error = ChildNumber;
```
TheQuantumPhysicist, line 49: Error is ChildNumber? This is wrong. Errors should implement `std::error::Error` trait, which this does not.

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,94 @@
...
+
+    fn try_from(num: ChildNumber) -> Result<Self, Self::Error> {
+        match num.get_index() {
+            0 => Ok(ReceiveFunds),
+            1 => Ok(Change),
+            _ => Err(num),
+        }
```
TheQuantumPhysicist, line 56: Can you please explain the logic here? I'm having trouble understanding why this is like this.

wallet/types/src/account_id.rs
```diff
@@ -0,0 +1,62 @@
...
+        let pk = pub_key.clone().into_public_key();
+        let pkh = PublicKeyHash::from(&pk);
+        AccountId(pkh)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
+pub struct AccountPrefixedId<Id> {
+    account_id: AccountId,
+    item_id: Id,
```
TheQuantumPhysicist, line 41: Please add documentation to what these are

TheQuantumPhysicist:
Looking OK. Though please check the comments. I'll do another review round later.

wallet/types/src/account_id.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// This is a composite id that combines a prefix account id and a generic item id suffix.
+/// It is useful for storing key/values that belong to different accounts and are stored in the
+/// same map.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
+pub struct AccountPrefixedId<Id> {
+    /// The account id is the prefix and implements HasPrefix for this struct
+    account_id: AccountId,
+    /// The generic item id. This could be anything, like `Id<Transaction` or `OutPoint`
```
TheQuantumPhysicist, line 45: Id<Transaction> typo

crypto/src/key/hdkd/u31.rs
```diff
@@ -30,6 +30,10 @@ impl U31 {
         (result, msb)
     }
 
+    pub const fn into_u32(self) -> u32 {
+        self.0
```
TheQuantumPhysicist, line 34: You could call into_u32() in the From implementation below.

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -163,7 +163,7 @@ impl Derivable for Secp256k1ExtendedPrivateKey {
             .into();
 
         let derivation_path = {
-            let mut child_path = self.derivation_path.as_vec().clone();
+            let mut child_path = self.derivation_path.as_slice().to_vec();
```
TheQuantumPhysicist, line 166: Since you're consuming `self` here, why not use `into_vec()` intead of a copy of that vec through the slice?

wallet/src/key_chain/account_key_chain.rs
```diff
@@ -0,0 +1,262 @@
...
+use wallet_types::{AccountId, AccountInfo, DeterministicAccountInfo};
+
+#[allow(dead_code)] // TODO remove
+/// This key chain contains a pool of pre-generated keys and addresses for the usage in a wallet
+pub struct AccountKeyChain {
+    /// The specific chain this KeyChain is based on, this will affect the address format
+    chain_config: Arc<ChainConfig>,
+
+    /// The account key from which all the addresses are derived
+    account_pubkey: ExtendedPublicKey,
```
TheQuantumPhysicist, line 38: ConstValue?

wallet/src/key_chain/account_key_chain.rs
```diff
@@ -0,0 +1,262 @@
...
+            self.account_pubkey.clone(),
+            self.lookahead_size,
+        ))
+    }
+
+    pub fn get_lookahead_size(&self) -> u32 {
+        self.lookahead_size
+    }
+
+    pub fn set_lookahead_size<B: Backend>(
```
TheQuantumPhysicist, line 235: Shouldn't this now be constant?

My understanding is that bitcoin core wouldn't allow this because you can't guarantee that you haven't used pre-generated addresses when you increase the gap.

wallet/src/key_chain/account_key_chain.rs
erasmospunk: Is `ConstValue` a marker type? Because it is still possible to change the field if the struct is mut.

TheQuantumPhysicist: Yes. It's for devs to realize that a value is meant to be constant over its lifetime.

crypto/src/key/secp256k1/extended_keys.rs
```diff
@@ -249,7 +249,7 @@ impl Derivable for Secp256k1ExtendedPublicKey {
...
```
TheQuantumPhysicist, line 252: into_vec()

crypto/src/lib.rs
```diff
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+extern crate core;
```
TheQuantumPhysicist, line 16: extern crate is unnecessary. This is very old rust.

wallet/src/key_chain/master_key_chain.rs
```diff
@@ -0,0 +1,144 @@
...
+        &self,
+        db_tx: &mut StoreTxRw<B>,
+        account_index: ChildNumber,
+    ) -> KeyChainResult<AccountKeyChain> {
+        AccountKeyChain::new_from_root_key(
+            self.chain_config.clone(),
+            db_tx,
+            &self.root_key,
+            account_index,
+            LOOKAHEAD_SIZE,
```
TheQuantumPhysicist, line 121: Does it make sense to make this an initialization parameter instead of just a constant?

wallet/src/account.rs
```diff
@@ -129,59 +129,110 @@ impl Account {
...
     }
 
+    /// Remove transaction outputs from the UTXO set of the account
+    fn remove_from_utxos<B: Backend>(
+        &mut self,
+        db_tx: &mut StoreTxRw<B>,
+        wallet_tx: &WalletTx,
+    ) -> WalletResult<()> {
+        let tx = wallet_tx.get_tx();
+        for i in 0..tx.outputs().len() {
```
TheQuantumPhysicist, line 178: Enumerating yourself is a bad idea as it's prone to off-by-one error. Usually the better solution is something like `tx.outputs().iter().enumerate()`, which will return a pair of the value and its index, which then you can use.

wallet/src/account.rs
```diff
@@ -129,59 +129,110 @@ impl Account {
...
         Ok(())
     }
 
+    /// Remove transaction outputs from the UTXO set of the account
+    fn remove_from_utxos<B: Backend>(
+        &mut self,
+        db_tx: &mut StoreTxRw<B>,
+        wallet_tx: &WalletTx,
+    ) -> WalletResult<()> {
+        let tx = wallet_tx.get_tx();
```
TheQuantumPhysicist, line 177: As an optimization, I would make this `WithId<Transaction>` since `get_id()` is called multiple times.

wallet/src/account.rs
```diff
@@ -129,59 +129,110 @@ impl Account {
...
+        for i in 0..tx.outputs().len() {
+            let outpoint = OutPoint::new(tx.get_id().into(), i as u32);
+            self.utxo.remove(&outpoint);
```
TheQuantumPhysicist, line 180: Do we want to assert that "something was removed"? A `debug_assert!` might help. The function `remove()` returns `Some()` when something is removed.

TheQuantumPhysicist:
I think besides the look-ahead issue this is looking better and better.

wallet/src/key_chain/account_key_chain.rs
TheQuantumPhysicist: But we don't handle the case of reducing the size properly, right? Why not modify that test to make the assumption that this doesn't change?

If the value won't change, then there's no need to test that it'll ever change.

TheQuantumPhysicist (approved):

wallet/src/key_chain/leaf_key_chain.rs
```diff
@@ -0,0 +1,517 @@
...
+        }
+
+        let mut receiving_public_keys = BTreeMap::new();
+        let mut change_public_keys = BTreeMap::new();
+        for (pubkey_id, xpub) in db_tx.get_public_keys(id)? {
+            let (purpose, key_index) = get_purpose_and_index(&pubkey_id.into_item_id())?;
+            let old_value = match purpose {
+                KeyPurpose::ReceiveFunds => receiving_public_keys.insert(key_index, xpub),
+                KeyPurpose::Change => change_public_keys.insert(key_index, xpub),
+            };
```
iljakuklic, line 148: Here's a random idea, feel free to run with it, modify or ignore as you please.

I noticed you often have a pairs of things (here `BTreeMap` with public keys) where one of the pairs is used with `ReceiveFunds` purpose and the other is used for `Change` purpose. You could define a helper container that packages these together and gives you convenient access to the object corresponding to given purpose without having to `match` every time.

Rough outline:
```rust
struct WithPurpose<T> { // TODO: name?
    receive: T,
    change: T,
}

impl<T> WithPurpose<T> {
    // implementations obvious, omitted
    fn new(receive: T, change: T) -> Self;
    fn ref_for(&self, purpose: KeyPurpose) -> &T;
    fn mut_for(&mut self, purpose: KeyPurpose) -> &mut T;
    // Optional, useful combined with `flat_map`:
    fn into_iter(&mut self) -> impl Iterator<Item = T> + ExactSizeIterator + '_;
    fn iter(self) -> impl Iterator<Item = &T> + ExactSizeIterator;
    fn iter_mut(&mut self) -> impl Iterator<Item = &mut T> + ExactSizeIterator + '_;
}
```

The above code could then be something like this, given `public_keys: WithPurpose<BTreeMap<_, _>>`:
```rust
let old_value = public_keys.mut_for(purpose).insert(key_index, xpub);
```

Also the return value of this whole function (`load_leaf_keys`) could be `WithPurpose<LeafKeyChain>` instead of `(LeafKeyChain, LeafKeyChain)`.

wallet/src/key_chain/leaf_key_chain.rs
```diff
@@ -0,0 +1,517 @@
...
+        lookahead_size: u32,
+    ) -> KeyChainResult<()> {
+        // Find how many keys to derive
+        let (starting_index, up_to_index) = match self.get_last_derived_index() {
+            None => (0u32, lookahead_size),
+            Some(last_derived_index) => {
+                let start_index = last_derived_index.increment()?.get_index();
+                let up_to_index = match self.usage_state.get_last_used() {
+                    None => lookahead_size,
+                    Some(last_used) => last_used.get_index() + lookahead_size + 1,
```
iljakuklic, line 370: Hmm, why `+1`?

wallet/src/key_chain/leaf_key_chain.rs
```diff
@@ -0,0 +1,517 @@
...
+        &mut self,
+        db_tx: &mut StoreTxRw<B>,
+        lookahead_size: u32,
+    ) -> KeyChainResult<()> {
+        // Find how many keys to derive
+        let (starting_index, up_to_index) = match self.get_last_derived_index() {
+            None => (0u32, lookahead_size),
+            Some(last_derived_index) => {
+                let start_index = last_derived_index.increment()?.get_index();
+                let up_to_index = match self.usage_state.get_last_used() {
```
iljakuklic, line 368: You can use `self.get_last_used()` and remove the dead code annotation from it.

wallet/src/key_chain/leaf_key_chain.rs
```diff
@@ -0,0 +1,517 @@
...
+            &AccountKeyPurposeId::new(self.account_id.clone(), self.purpose),
+            &self.usage_state,
+        )?)
+    }
+
+    /// Get a new issued index and check that it is a valid one i.e. not exceeding the lookahead
+    fn get_new_issued_index(&self, lookahead_size: u32) -> KeyChainResult<ChildNumber> {
+        // TODO consider last used index as well
+        let new_issued_index = {
+            match self.usage_state.get_last_issued() {
```
iljakuklic, line 259: Again, you can use `self.get_last_issued()` and remove the dead code annotation.

iljakuklic (approved):
Looks good. A number of minor comments and suggestions below:

wallet/src/key_chain/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! # BIP39 key chain
+//! The KeyChain struct holds and constantly derives keys for the wallet addresses
+//! It uses the following derivation scheme:
+//!
+//! m/44'/19788'/<account_number>'/<key_purpose>'/<key_index>'
```
pavel-kokolemin, line 20: The comment says that `key_purpose` and `key_index` are hardened keys, but in the code they are normal keys.

wallet/src/key_chain/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+    KeyNotRoot,
+}
+
+/// Result type used for the key chain
+type KeyChainResult<T> = Result<T, KeyChainError>;
+
+/// Create a deterministic path for an account identified by the `account_index`
+fn make_account_path(chain_config: &ChainConfig, account_index: ChildNumber) -> DerivationPath {
+    // The path is m/44'/<coin_type>'/<account_index>'
+    let path = vec![BIP44_PATH, chain_config.bip44_coin_type(), account_index];
```
pavel-kokolemin, line 101: Maybe we have a check here that the `account_index` is really hardened?

wallet/src/account.rs
```diff
@@ -129,59 +130,111 @@ impl Account {
...
-        // TODO implement
-        true
+    fn is_mine_or_watched(&self, txo: &TxOutput) -> bool {
+        let destination = match txo {
+            TxOutput::Transfer(_, d) => Some(d),
+            TxOutput::LockThenTransfer(_, d, _) => Some(d),
+            TxOutput::Burn(_) => None,
+            TxOutput::StakePool(_) => None,
+            TxOutput::ProduceBlockFromStake(_, _) => None,
+            TxOutput::DecommissionPool(_, _, _, _) => None,
```
pavel-kokolemin, line 203: What about `DecommissionPool` outputs? I think they should be checked as well.

wallet/src/key_chain/leaf_key_chain.rs
```diff
@@ -0,0 +1,517 @@
...
+
+    /// Derive up `lookahead_size` keys starting from the last used index. If the gap from the last
+    /// used key to the last derived key is already `lookahead_size`, this method has no effect
+    pub fn top_up<B: Backend>(
+        &mut self,
+        db_tx: &mut StoreTxRw<B>,
+        lookahead_size: u32,
+    ) -> KeyChainResult<()> {
+        // Find how many keys to derive
+        let (starting_index, up_to_index) = match self.get_last_derived_index() {
```
pavel-kokolemin, line 364: Maybe it would be better to rewrite it like this?
```
        let starting_index = match self.get_last_derived_index() {
            Some(last_derived_index) => last_derived_index.increment()?.get_index(),
            None => 0,
        };
        let up_to_index = match self.usage_state.get_last_used() {
            Some(last_used) => last_used.get_index() + lookahead_size + 1,
            None => lookahead_size,
        };
```

wallet/src/key_chain/leaf_key_chain.rs
```diff
@@ -0,0 +1,517 @@
...
+        db_tx: &mut StoreTxRw<B>,
+        extended_public_key: &ExtendedPublicKey,
+        lookahead_size: u32,
+    ) -> KeyChainResult<bool> {
+        // Check if public key is in the key pool
+        if self.is_pubkey_mine_in_key_pool(extended_public_key) {
+            // Get the key index of the public key, this should always be Some
+            let key_index = extended_public_key.get_derivation_path().as_slice().last()
+                .expect("The provided public key belongs to this key chain, hence it should always have a key index");
+            self.usage_state.set_last_used(Some(*key_index));
```
pavel-kokolemin, line 463: This may override the last used to a lower value, which is probably not a good idea.

TheQuantumPhysicist: This is actually one of the reasons I have been against having explicit way to set_last_used instead of something that increments + checks based on the internal state.

pavel-kokolemin (approved):

stanislav-tkach (approved):

wallet/src/key_chain/leaf_key_chain.rs
pavel-kokolemin: I think +1 is needed here (it's now replaced by the `increment()` call). For example, if we the last used index is 5 and 20 unused addresses are needed. This will be `[6..25]` range. But because we exclude `up_to_index` when we iterate (`for i in starting_index..up_to_index`) +1 is used.

wallet/src/account.rs
pavel-kokolemin: I updated it to also look at the `DecommissionPool` outputs.

wallet/src/account.rs
```diff
@@ -129,59 +130,111 @@ impl Account {
...
+    /// Return true if this transaction output is can be spent by this account or if it is being
+    /// watched.
     #[allow(dead_code)] // TODO remove
-    fn is_mine_or_watched(&self, _txo: &TxOutput) -> bool {
-        // TODO implement
-        true
+    fn is_mine_or_watched(&self, txo: &TxOutput) -> bool {
+        let destination = match txo {
+            TxOutput::Transfer(_, d) => Some(d),
+            TxOutput::LockThenTransfer(_, d, _) => Some(d),
```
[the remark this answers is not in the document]
pavel-kokolemin: Updated, thanks.

TheQuantumPhysicist:
I think is good to go, unless you're waiting for someone else's review.

crypto/src/key/hdkd/derivation_path.rs
```diff
@@ -54,6 +54,16 @@ impl DerivationPath {
...
     }
+
+    /// Get the difference of this path and a sub path.
+    pub fn get_super_path_diff(&self, sub_path: &DerivationPath) -> Option<&[ChildNumber]> {
+        if let Some(diff_path) = self.as_slice().strip_prefix(sub_path.as_slice()) {
+            if !diff_path.is_empty() {
+                return Some(diff_path);
+            }
+        }
+        None
```
iljakuklic, line 65: Is there a problem with just returning `Some(&[])` if the paths are the same? Looks like a good neutral element with respect to path concatenation.

If not, would something like this work?
```suggestion
        self.as_slice().strip_prefix(sub_path.as_slice()).filter(|p| !p.is_empty())
```

wallet/src/key_chain/account_key_chain.rs
```diff
@@ -0,0 +1,269 @@
...
+use utils::const_value::ConstValue;
+use wallet_storage::{StoreTxRo, StoreTxRw, WalletStorageRead, WalletStorageWrite};
+use wallet_types::keys::KeyPurpose;
+use wallet_types::{AccountId, AccountInfo, DeterministicAccountInfo};
+
+/// This key chain contains a pool of pre-generated keys and addresses for the usage in a wallet
+pub struct AccountKeyChain {
+    #[allow(dead_code)] // TODO remove
+    /// The specific chain this KeyChain is based on, this will affect the address format
+    chain_config: Arc<ChainConfig>,
```
iljakuklic, line 37: Is the dead code annotation still required?

wallet/src/key_chain/leaf_key_chain.rs
```diff
@@ -0,0 +1,506 @@
...
+            Some(last_used_index) => {
+                new_issued_index > last_used_index.get_index() + lookahead_size
+            }
+        };
+
+        if lookahead_exceeded {
+            Err(KeyChainError::LookAheadExceeded)
+        } else {
+            Ok(())
+        }
```
iljakuklic, line 282: Nitpick: Would `ensure!` be an improvement here?

wallet/src/key_chain/master_key_chain.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        // The current format supports a single root key
+        if root_keys.len() != 1 {
+            return Err(KeyChainError::OnlyOneRootKeyIsSupported);
+        }
+
+        let (_, key_content) =
+            root_keys.pop_first().expect("Should not fail because it contains 1 key/value");
+
+        let root_key = key_content.into_key();
```
iljakuklic, line 102: Nitpick: I think `Itertools::exactly_one` could make this part a bit more concise.

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,109 @@
...
+
+    /// Get the last index used in the blockchain
+    pub fn get_last_used(&self) -> Option<ChildNumber> {
+        self.last_used
+    }
+
+    /// Increments the last index used in the blockchain until up_to_last_used. This has no effect
+    /// if the up_to_last_used is smaller than the self value
+    pub fn increment_up_to_last_used(&mut self, up_to_last_used: ChildNumber) {
+        if self.last_used.is_none() || self.last_used.unwrap() < up_to_last_used {
```
iljakuklic, line 100: Clippy complains about unwrap here

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,109 @@
...
+
+    /// Get the last index issued to the user
+    pub fn get_last_issued(&self) -> Option<ChildNumber> {
+        self.last_issued
+    }
+
+    /// Increments the last index issued in the blockchain until up_to_last_issued.
+    /// This has no effect if the up_to_last_issued is smaller than the self value
+    pub fn increment_up_to_last_issued(&mut self, up_to_last_issued: ChildNumber) {
+        if self.last_issued.is_none() || self.last_issued.unwrap() < up_to_last_issued {
```
iljakuklic, line 113: ...and here

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,109 @@
...
+        Self {
+            last_used,
+            last_issued,
+        }
+    }
+
+    /// Get the last index used in the blockchain
+    pub fn get_last_used(&self) -> Option<ChildNumber> {
+        self.last_used
+    }
```
iljakuklic, line 95: I'd prefer this to be called just `last_used` without `get_`. More consistent with (most of) rest of the code base where `get_` indicates that some work other than returning a field may be performed (e.g. `Idable::get_id` may calculate hash).

wallet/types/src/keys.rs
```diff
@@ -13,9 +13,109 @@
...
+    pub fn increment_up_to_last_used(&mut self, up_to_last_used: ChildNumber) {
+        if self.last_used.is_none() || self.last_used.unwrap() < up_to_last_used {
+            self.last_used = Some(up_to_last_used);
+        }
+    }
+
+    /// Get the last index issued to the user
+    pub fn get_last_issued(&self) -> Option<ChildNumber> {
+        self.last_issued
+    }
```
iljakuklic, line 108: Same as above regarding the `get_` prefix.

iljakuklic (approved):
Looks good. I left some extra comments but all quite nitpicky and safe to ignore.

pavel-kokolemin:
> Looks good. I left some extra comments but all quite nitpicky and safe to ignore.

Thanks. I will fix these in the next PR.

PR #841, opened 2023-04-17
Add basic wallet cli binary

pavel-kokolemin:
It can be used like this:
Start the node:
```cargo run --bin node -- regtest```

Start the wallet-cli:
```cargo run --bin wallet-cli -- --network regtest```

### CODE REVIEW

File: dns_server/src/config.rs [hunk 1]
```diff
@@ -57,14 +57,6 @@ pub struct DnsServerConfig {
...
-
 fn default_data_dir() -> String {
-    BaseDirs::new()
-        .expect("BaseDirs build failed")
-        .data_dir()
-        .join(DEFAULT_DATA_DIR_NAME)
-        .to_str()
-        .expect("expected valid default data dir path")
-        .to_owned()
+    default_data_dir_common().to_str().expect("Valid UTF-8 expected").to_owned()
```
TheQuantumPhysicist, line 61: Is this always true? Windows uses UTF-16 in paths and registry entries.

wallet/wallet-cli/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum WalletCliError {
+    #[error("RPC error: {0}")]
+    RpcError(node_comm::rpc_client::NodeRpcError),
+    #[error("Wallet error: {0}")]
+    WalletError(wallet::wallet::WalletError),
+    #[error("Console IO error: {0}")]
+    ConsoleIoError(std::io::Error),
+    #[error("Cookie file {0} read error: {1}")]
+    CookieFileReadError(PathBuf, std::io::Error),
```
TheQuantumPhysicist, line 39: When I run the CLI, I get this error in a panic form:

```
Error: CookieFileReadError("/home/samer/.mintlayer/mainnet/.cookie", Os { code: 2, kind: NotFound, message: "No such file or directory" })
```

which is reasonable from a technical point of view. But we should have a nicer message to the user. Consider a user who doesn't really know what a cookie is or how that works. We should have a way to detect whether the node is running, and not attempt connection (or anything related) unless it exists.

wallet/wallet-cli/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum WalletCliError {
+    #[error("RPC error: {0}")]
+    RpcError(node_comm::rpc_client::NodeRpcError),
+    #[error("Wallet error: {0}")]
+    WalletError(wallet::wallet::WalletError),
+    #[error("Console IO error: {0}")]
+    ConsoleIoError(std::io::Error),
+    #[error("Cookie file {0} read error: {1}")]
```
TheQuantumPhysicist, line 38: `RPC authentication cookie-file {0} isn't found. Error: {1}`

wallet/wallet-cli/src/config.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// RPC address
+    #[clap(long)]
+    pub rpc_address: Option<SocketAddr>,
+
+    /// Path to the RPC cookie file. If not set, the value is read from the default cookie file location.
+    #[clap(long)]
+    pub rpc_cookie_file: Option<String>,
+
+    /// RPC username
```
TheQuantumPhysicist, line 58: RPC username (either provide a username and password, or use a cookie file. You cannot use both)

wallet/wallet-cli/src/config.rs
```diff
@@ -0,0 +1,152 @@
...
+    #[clap(long)]
+    pub rpc_username: Option<String>,
+
+    /// RPC password
```
TheQuantumPhysicist, line 62: RPC password (either provide a username and password, or use a cookie file. You cannot use both)

wallet/wallet-cli/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+        None => load_wallet(chain_config, db)?,
+    };
+
+    let rpc_client = make_rpc_client(
+        config.rpc_address,
+        Some((&config.rpc_username, &config.rpc_password)),
+    )
+    .await
+    .map_err(WalletCliError::RpcError)?;
+    println!("Best block id: {:?}", rpc_client.get_best_block_id().await);
```
TheQuantumPhysicist, line 91: Please use only hex, not plain Debug

TheQuantumPhysicist: Please use a custom macro for printing. Not `println!` directly. This will help in wrapping anything we do with printing or change the function we use to print.

wallet/wallet-cli/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    let mut history = CliHistory::default();
+
+    loop {
```
TheQuantumPhysicist, line 95: Please put this in a function... something like `start_cli_repl()`.

dns_server/src/config.rs
pavel-kokolemin: Good question. I looked a bit and it seems that Rust on Windows uses WTF-8 encoding internally and `to_str` succeeds most of the time (so it's not much different from on other OSs).
Anyway, I switched to `PathBuf` here.

wallet/wallet-cli/src/main.rs
pavel-kokolemin: It looks like this now:

wallet-cli launch failed: RPC authentication cookie-file /home/username/.mintlayer/regtest/.cookie read error: No such file or directory (os error 2). Please make sure the node is started.

wallet/wallet-cli/src/main.rs
pavel-kokolemin: "isn't found" is not always accurate. There can also be permission errors. I changed it to this:
```
RPC authentication cookie-file {0} read error: {1}. Please make sure the node is started.
```

pavel-kokolemin:
I removed the "create-wallet" subcommand. The wallet-cli now automatically offers to create or import a wallet at startup.

wallet/wallet-cli/src/main.rs
TheQuantumPhysicist: Alright. I can live with that for now.

wallet/wallet-cli/src/main.rs
TheQuantumPhysicist: The node does generate a cookie file no matter what, right? Because I remember in bitcoin you can choose whether you wanna generate a cookie file. I even remember that if you provide a username and password to the node, a cookie file won't be generated. What logic do we use?

If we have similar discrepancy, we may wanna do some test command first to ensure that the node is reachable. Besides, not being able to connect to the node isn't necessarily a crime. The connection can be lost at any moment and that should be OK.

wallet/src/wallet.rs
```diff
@@ -117,6 +117,15 @@ impl<B: storage::Backend> Wallet<B> {
     }
 }
 
+pub fn generate_new_mnemonic() -> bip39::Mnemonic {
+    let mut rng = crypto::random::make_true_rng();
+    let mut data = [0u8; 32];
+    rng.fill(&mut data);
+    let res = bip39::Mnemonic::from_entropy(&data).expect("should not fail");
```
TheQuantumPhysicist, line 124: Please add a TODO for the language. We can support other languages for free.

TheQuantumPhysicist: Well, frankly, it's not that hard to do. You can add a parameter to this function for language, then just pass English... then we do the CLI implementation for that later.

wallet/wallet-cli/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::sync::Arc;
+
+use clap::Parser;
+use dialoguer::theme::ColorfulTheme;
+use errors::WalletCliError;
+use node_comm::make_rpc_client;
+use output::OutputContext;
+use wallet::{Wallet, WalletError};
+
+type DefWallet = Wallet<wallet_storage::DefaultBackend>;
```
TheQuantumPhysicist, line 33: DefaultWallet?

Not a big fan of some abbreviations. Let's call it a nitpick :D

wallet/wallet-cli/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    let db = wallet::wallet::open_or_create_wallet_file(&config.wallet_file)
+        .map_err(WalletCliError::WalletError)?;
+
+    let theme = ColorfulTheme::default();
+
+    let wallet = match Wallet::load_wallet(Arc::clone(&chain_config), Arc::clone(&db)) {
+        Ok(wallet) => wallet,
+        Err(WalletError::WalletNotInitialized) => {
+            wallet_init::new_wallet(output, chain_config, db, &theme)?
+        }
```
TheQuantumPhysicist, line 51: We have to be careful and not overwrite the user wallet.

If you see how Monero does this, usually you can open a wallet at a specific path, and if the wallet isn't found, the user is asked to create a wallet at that location.

wallet/wallet-cli/src/config.rs
```diff
@@ -0,0 +1,150 @@
...
+            Network::Regtest => ChainType::Regtest,
+            Network::Signet => ChainType::Signet,
+        };
+
+        let data_dir = prepare_data_dir(
+            || default_data_dir_for_chain(chain_type.name()).join(DEFAULT_WALLETS_DIR),
+            &wallets_dir,
+        )
+        .map_err(WalletCliError::PrepareData)?;
+        let wallet_file = data_dir.join(wallet_name.unwrap_or(DEFAULT_WALLET_NAME.into()));
```
TheQuantumPhysicist, line 121: The wallet data should not be related to the node data dir. The way it should be seen, is that the node is a stateful entity that keeps running no matter what, and an unbounded number of wallets can connect to it. The files of the wallets have nothing to do with the node. They can be on removable drives, on a network, or whatever.

We have a design here different that what bitcoin core does, which is something that has caused issues for Bitcoin. I prefer the way Monero does this, where the wallet is completely separate from the wallet, and the only thing that they make them communicate is the RPC interface, which is for us not necessarily just the RPC interface, but the NodeInterface trait in general.

wallet/wallet-cli/src/config.rs
```diff
@@ -0,0 +1,150 @@
...
+                load_cookie(default_data_dir_for_chain(chain_type.name()).join(COOKIE_FILENAME))?
+            }
+            (Some(cookie_path), None, None) => load_cookie(cookie_path)?,
+            (None, Some(username), Some(password)) => (username, password),
+            _ => {
+                return Err(WalletCliError::InvalidConfig(
+                    "Invalid RPC cookie/username/password combination".to_owned(),
+                ))
+            }
+        };
```
TheQuantumPhysicist, line 138: Let's forget the part of the wallet file for now, but the wallet file is also part of this discussion.

Something to keep in mind, is that the wallet initialization from a utilization point of view shouldn't be in the wallet-cli crate, but should be in a separate crate, which can be used by any wallet implementation we create in the future. Imagine in the future we create a GUI wallet. Everything should be common between those two, and the only difference is the configuration we input into those two. 

So what should be in the wallet CLI is only everything related to CLI specifically, not to the wallet. Like prompts, messages, commands, etc. Things like initialization, creation, etc, should all go to a separate crate that is used in this crate. In the future, once we establish the GUI wallet, we just import that crate and we use it, and spend time only on designing the GUI.

wallet/wallet-cli/src/wallet_init.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn from(val: ImportMnemonic) -> Self {
+        match val {
+            ImportMnemonic::Generate => "Generate new mnemonic",
+            ImportMnemonic::Import => "Import mnemonic",
+            ImportMnemonic::Cancel => "Cancel",
+        }
+    }
+}
+
+pub fn new_wallet(
```
TheQuantumPhysicist, line 43: This function, and the whole file, to be honest, is a good example of something that should go to that new crate.

Let's call `wallet-controller`.

wallet/wallet-node-client/src/rpc_client/mod.rs
```diff
@@ -39,10 +41,29 @@ pub struct NodeRpcClient {
     http_client: HttpClient,
 }
 
+fn get_headers(username_password: Option<(&str, &str)>) -> http::HeaderMap {
```
TheQuantumPhysicist, line 44: I would call this `make_http_header_with_auth`

wallet/wallet-cli/src/commands/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+use crate::{cli_println, errors::WalletCliError, output::OutputContext, DefWallet};
+
+#[derive(Debug, Parser)]
+#[clap(rename_all = "lower")]
+pub enum WalletCommands {
+    /// Returns the current best block hash
+    BestBlock,
+
+    /// Returns the current block height
+    BlockHeight,
```
TheQuantumPhysicist, line 30: BestBlockHeight

wallet/wallet-cli/src/commands/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+    Rescan,
+
+    /// Quit the REPL
+    Exit,
+
+    /// Print history
+    History,
+
+    /// Clear screen
+    Clear,
```
TheQuantumPhysicist, line 45: ClearScreen

wallet/wallet-cli/src/repl/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    ";
+
+    // Strip out name/version
+    const APPLET_TEMPLATE: &str = "\
+        {about-with-newline}\n\
+        {usage-heading}\n    {usage}\n\
+        \n\
+        {all-args}{after-help}\
+    ";
+
```
TheQuantumPhysicist, line 51: How difficult is it to support `help <cmd>`, where we can add additional information about the arguments, and such?

pavel-kokolemin: It already works.

wallet/wallet-cli/src/repl/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    let repl_command = Command::new("repl")
+        .multicall(true)
+        .arg_required_else_help(true)
+        .subcommand_required(true)
+        .subcommand_value_name("APPLET")
+        .subcommand_help_heading("APPLETS")
```
TheQuantumPhysicist, line 57: How about Commands?

wallet/wallet-cli/src/config.rs
pavel-kokolemin: Sure, I understand that. But we still need a default directory to store the wallet files somewhere. I decided to use  the "wallets" directory in the default data dir (for example, "$HOME/.mintlayer/wallets" on Linux). It does not matter if the node is started or not on the same host.

wallet/wallet-cli/src/config.rs
pavel-kokolemin: Okay, thanks. I will keep that in mind.

wallet/wallet-cli/src/wallet_init.rs
pavel-kokolemin: > This function, and the whole file, to be honest, is a good example of something that should go to that new crate.

I made small changes there and now `wallet_init.rs` has only CLI related functionality. 

> Let's call wallet-controller.

Do we really need another crate? Can't we use the root `wallet` crate for this?

wallet/wallet-cli/src/repl/mod.rs
pavel-kokolemin: Sorry, what about it? I think we have everything we need.

wallet/wallet-cli/src/repl/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+        .get_subcommands()
+        .map(|command| command.get_name().to_owned())
+        .chain(std::iter::once("help".to_owned()))
+        .collect::<Vec<_>>();
+
+    let completer = Box::new(DefaultCompleter::new_with_wordlen(commands.clone(), 0));
+
+    let mut line_editor = Reedline::create()
+        .with_history(history)
+        .with_completer(completer)
```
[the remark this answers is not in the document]
pavel-kokolemin: Right, thanks. That should be fixed now.

wallet/wallet-cli/src/config.rs
TheQuantumPhysicist: Please, no default dir/path for the wallet. This is wrong.

wallet/wallet-cli/src/wallet_init.rs
TheQuantumPhysicist: I do prefer creating a separate crate because using the wallet is not like using the wallet primitives. Some parametrization is being done, which is application-specific, which should be separate.

wallet/wallet-cli/src/repl/mod.rs
TheQuantumPhysicist: Can we add extra information to the returned message that only appears when we `help <cmd>`?

wallet/wallet-cli/src/repl/mod.rs
TheQuantumPhysicist: I mean the name "applets" is weird

wallet/wallet-cli/src/main.rs
pavel-kokolemin: The node won't generate a cookie file if username/password is specified.

wallet/wallet-cli/src/repl/mod.rs
pavel-kokolemin: Good question. I just tried and it can be done like this:
```
    /// Submit a block to be included in the chain
    ///
    /// More information about block submits.
    /// More information about block submits.
    ///
    /// Even more information about block submits.
    /// Even more information about block submits.
    /// Even more information about block submits.
    /// Even more information about block submits.
    SubmitBlock {
        /// Hex encoded block
        block: String,
    },
```

```
Wallet> help submitblock
Submit a block to be included in the chain

More information about block submits. More information about block submits.

Even more information about block submits. Even more information about block submits. Even more information about block submits. Even more information about block submits.

Usage:
    submitblock <BLOCK>

Arguments:
  <BLOCK>
          Hex encoded block

Options:
  -h, --help
          Print help (see a summary with '-h')
```

And the short help works like before:
```
Wallet> help 
Commands:
  bestblock        Returns the current best block hash
  bestblockheight  Returns the current best block height
  submitblock      Submit a block to be included in the chain
  rescan           Rescan
  exit             Quit the REPL
  history          Print history
  clear            Clear screen
  clearhistory     Clear history
  help             Print this message or the help of the given subcommand(s)
```

TheQuantumPhysicist: Very nice! Please add something as an example and we'll complete the docs later.

wallet/wallet-cli/src/commands/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /// Clear screen
+    #[clap(name = "clear")]
+    ClearScreen,
+
+    /// Clear history
+    ClearHistory,
+}
+
+pub async fn handle_wallet_command(
```
TheQuantumPhysicist, line 93: This function has to be part of the controller. Not in the sense that it has to be moved, but in the sense that the controller has to expose function calls (maybe through a trait?) with all the functionality of the wallet that a "client" (whether GUI or not) should be able to do. Then, this function will parse that command here in this crate, in this function, and call that function in the controller to do the task.

In other words: This crate should not even know that there's something called "NodeInterface". So node-comm should not be here.

TheQuantumPhysicist: Btw, this is called "model-view-controller" design.

So that trait would be the model, and this crate would be the view. Hopefully I'm not missing something :smile:

pavel-kokolemin: I added a simple `Controller` for this. For now it's just a wrapper for `NodeInterface`, but it will be extended later to do more sophisticated stuff.

TheQuantumPhysicist (approved):

PR #842, opened 2023-04-17
Wallet send request

erasmospunk:


wallet/src/account.rs
```diff
@@ -81,6 +81,22 @@ impl Account {
         })
     }
 
+    pub fn complete_and_add_send_request<B: Backend>(
```
TheQuantumPhysicist, line 84: Please explain what "complete" means

wallet/src/account.rs
```diff
@@ -81,6 +81,22 @@ impl Account {
...
+        self.complete_send_request(&mut request)?;
+        self.add_transaction(db_tx, request.transaction.clone(), TxState::InMempool)?;
+        Ok(request.transaction)
+    }
+
+    fn complete_send_request(&mut self, _request: &mut SendRequest) -> WalletResult<()> {
+        // TODO implement this
+        println!("Send request ");
+        Ok(())
+    }
```
TheQuantumPhysicist, line 98: Can we encourage more separation and move these outside this struct? The last thing we want is modules/structs that are 1000 lines long.

crypto/src/key/secp256k1/mod.rs
```diff
@@ -96,7 +96,7 @@ impl Secp256k1PrivateKey {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
```
TheQuantumPhysicist, line 99: No Hash, please.

crypto/src/key/mod.rs
```diff
@@ -48,7 +48,7 @@ pub struct PrivateKey {
     key: PrivateKeyHolder,
 }
 
-#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Decode, Encode)]
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Hash, Clone, Decode, Encode)]
```
TheQuantumPhysicist, line 51: For the love of God...

erasmospunk:
@TheQuantumPhysicist please pull this branch https://github.com/erasmospunk/mintlayer-core/tree/wallet-send-request

I made a mistake with rebasing and this is not the correct branch.

pavel-kokolemin:
I think this should be ready for initial review. It's now possible to generate addresses, view wallet balance and send balance to addresses from the CLI wallet. I will work on tests in the meantime.

wallet/src/account/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+                .map(|(outpoint, utxo)| (outpoint.into_item_id(), utxo))
+                .collect();
+
+            // TODO: Call coin selector
+
+            request.fill_inputs(utxos)?;
+        }
+
+        let (input_coin_amount, input_tokens_amounts) =
+            Self::calculate_output_amounts(request.utxos().iter())?;
```
azarovh, line 99: Just thinking out loud: should we consider in the future reusing this functionality from TransactionVerifier? It could be beneficial in 2 ways: avoid code duplication and unifying interfaces because it will require implementing `UtxoView` and `PoSAccountingView`. `UtxoView` and `UtxoCache` are good abstractions to replace the current approach with BtreeMap

TheQuantumPhysicist (commented):
I'm not done reviewing but some times github holds my comments hostage until I submit a review without me starting a formal review.

wallet/src/account/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+        db_tx.del_utxo(&account_utxo_id)?;
+        Ok(())
+    }
+
+    fn is_available_for_spending(&self, _txo: &TxOutput) -> bool {
+        // TODO implement
+        true
+    }
+
+    fn get_tx_output_destination(txo: &TxOutput) -> Option<&Destination> {
```
azarovh, line 282: Proper destination getter is implemented in `chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs`
I'd place a todo here and return something like `UnsupportedTransactionOutput`

wallet/src/account/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+        db_tx: &mut StoreTxRw<B>,
+        outpoint: &OutPoint,
+    ) -> WalletResult<()> {
+        let account_utxo_id = AccountOutPointId::new(self.get_account_id(), outpoint.clone());
+        db_tx.del_utxo(&account_utxo_id)?;
+        Ok(())
+    }
+
+    fn is_available_for_spending(&self, _txo: &TxOutput) -> bool {
+        // TODO implement
```
TheQuantumPhysicist, line 278: Shouldn't we just call `self.is_mine_or_watched(txo)` here?

wallet/src/send_request/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    pub fn outputs(&self) -> &[TxOutput] {
+        &self.outputs
+    }
+
+    pub fn utxos(&self) -> &[TxOutput] {
+        &self.utxos
+    }
+
+    pub fn fill_inputs(
```
TheQuantumPhysicist, line 94: `with_inputs`, and return Self?

wallet/src/send_request/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+        utxos: BTreeMap<OutPoint, Utxo>,
+    ) -> Result<(), TransactionCreationError> {
+        for (outpoint, utxo) in utxos.into_iter() {
+            self.inputs.push(TxInput::new(outpoint.tx_id(), outpoint.output_index()));
+            self.utxos.push(utxo.take_output());
+        }
+        Ok(())
+    }
+
+    pub fn get_transaction(&self) -> Result<Transaction, TransactionCreationError> {
```
TheQuantumPhysicist, line 105: `into_transaction()`, and consume self, instead of cloning in the fucntion?

wallet/wallet-node-client/src/node_traits.rs
```diff
@@ -38,6 +38,11 @@ pub trait NodeInterface {
         first_block: Id<GenBlock>,
         second_block: Id<GenBlock>,
     ) -> Result<Option<(Id<GenBlock>, BlockHeight)>, Self::Error>;
+    async fn generate_block(
+        &self,
+        reward_destination_hex: String,
+        transactions_hex: Option<Vec<String>>,
```
TheQuantumPhysicist, line 44: One thing that isn't done correctly in this trait and should be fixed later (or feel free to fix it now if you wish), is the assumption that everything has to be serialized. The serialization should be implemented in the RPC implementation, not in this trait. This is because if we use handles directly, then we won't have to serialize anything.

crypto/src/key/hdkd/u31.rs
```diff
@@ -34,6 +36,14 @@ impl U31 {
         self.0
     }
 
+    pub const fn from_u32(value: u32) -> Option<Self> {
+        if value & MSB_BIT == 0 {
```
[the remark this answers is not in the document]
azarovh: missed that it's const, sorry

wallet/src/account/mod.rs
azarovh: To clarify a bit: I suggest returning error for the match arms that require accounting info, not the whole function

wallet/src/wallet/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+        self.for_account_ro(account_index, |account, db_tx| account.get_balance(db_tx))
+    }
+
+    pub fn get_new_address(&mut self, account_index: U31) -> WalletResult<Address> {
+        self.for_account_rw(account_index, |account, db_tx| {
+            account.get_new_address(db_tx, KeyPurpose::ReceiveFunds)
+        })
+    }
+
+    pub fn send_to_address(
```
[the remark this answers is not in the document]
pavel-kokolemin: I updated to `Vec<TxOutput>` which should be more flexible.

wallet/src/account/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+            for (output_index, output) in block.block_reward().outputs().iter().enumerate() {
+                let outpoint =
+                    OutPoint::new(OutPointSourceId::BlockReward(block_id), output_index as u32);
+                self.add_utxo_if_own(db_tx, output, outpoint, utxo_source.clone())?
+            }
+
+            for signed_tx in block.transactions() {
+                let tx_id = signed_tx.transaction().get_id();
+                for input in signed_tx.inputs().iter() {
+                    self.del_utxo_if_own(db_tx, input.outpoint())?
```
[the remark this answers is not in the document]
pavel-kokolemin: Sure, but do we need to spend time on this now? I don't see why we would need it in the near future.

wallet/src/account/mod.rs
pavel-kokolemin: I'm not sure about this idea. I would prefer to keep wallet's code separately.

wallet/src/account/mod.rs
pavel-kokolemin: Sorry, I don't quite understand what you mean. I think this code will change and we could leave it as is for now.

wallet/src/send_request/mod.rs
pavel-kokolemin: I changed it to `into_transaction_and_utxos` because UTXOs are also needed.

wallet/wallet-node-client/src/node_traits.rs
pavel-kokolemin: OK, sure, but let's do this in other PR.

pavel-kokolemin: Should be fixed here - https://github.com/mintlayer/mintlayer-core/pull/898

serialization/src/hex_encoded.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Wrapper that serializes objects as hex encoded string for `serde`
+#[derive(Debug, Clone)]
+pub struct HexEncoded<T>(T);
+
+impl<T> HexEncoded<T> {
+    pub fn new(value: T) -> Self {
+        Self(value)
+    }
+
+    pub fn into_inner(self) -> T {
```
TheQuantumPhysicist, line 29: I would call this `take`

wallet/src/send_request/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    pub fn outputs(&self) -> &[TxOutput] {
+        &self.outputs
+    }
+
+    pub fn utxos(&self) -> &[TxOutput] {
+        &self.utxos
+    }
+
+    pub fn with_inputs(mut self, utxos: BTreeMap<OutPoint, Utxo>) -> Self {
```
TheQuantumPhysicist, line 92: `BTreeMap<OutPoint, Utxo>` can be changed to something like `impl IntoIterator<Item = (Outpoint, Utxo)>`.

wallet/src/send_request/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+        Self {
+            flags: transaction.flags(),
+            utxos,
+            inputs: transaction.inputs().to_vec(),
+            outputs: transaction.outputs().to_vec(),
+            lock_time: transaction.lock_time(),
+        }
+    }
+
+    pub fn with_outputs(mut self, mut outputs: Vec<TxOutput>) -> Self {
```
TheQuantumPhysicist, line 75: This can also use IntoIterator, and then we just chain iterators in the impl.

wallet/src/send_request/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+        Self {
+            flags: 0,
+            utxos: Vec::new(),
+            inputs: Vec::new(),
+            outputs: Vec::new(),
+            lock_time: 0,
+        }
+    }
+
+    pub fn from_transaction(transaction: Transaction, utxos: Vec<TxOutput>) -> Self {
```
TheQuantumPhysicist, line 65: In my opinion, this is an example where `Vec<>` is OK, because we just move it inside the struct, as is.

wallet/src/wallet/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub fn get_new_address(&mut self, account_index: U31) -> WalletResult<Address> {
+        self.for_account_rw(account_index, |account, db_tx| {
+            account.get_new_address(db_tx, KeyPurpose::ReceiveFunds)
+        })
+    }
+
+    pub fn create_transaction_to_addresses(
+        &mut self,
+        account_index: U31,
+        outputs: Vec<TxOutput>,
```
TheQuantumPhysicist, line 258: IntoIterator

crypto/src/key/hdkd/u31.rs
```diff
@@ -15,11 +15,13 @@
...
 use std::{fmt::Display, str::FromStr};
 
+use serialization::{Decode, Encode};
+
 use super::derivable::DerivationError;
 
 pub const MSB_BIT: u32 = 0x80000000;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd, Encode)]
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I see it's encoded correctly now, assuming I'm not mistaken with the correct encoding. Did you change your mind?

Just to be clear, having multiple ways to encode one thing is a no-no. It just creates confusion and problems in the future.

wallet/src/account/mod.rs
TheQuantumPhysicist: Retrieving the destination from an output is not as simple as it looks. It depends on the task at hand. It's simple for normal Transfer outputs, but when staking is involved, for example, it becomes more complex. I think Gosha means that there's a facility now for how to do that (I created that a few days ago). So, you can use that in order to handle this problem correctly, whether through using the already-provided implementations, or adding your own for your own wallet-related purposes.

wallet/src/account/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: The problem is that this kind of question can only be answered after suffering. I definitely don't have the foresight to see the consequences of ignoring something that's done for years... but we tend to avoid reinventing the wheel unless we have to and we find a good reason. So, it may be my ignorance playing in here, but I would prefer we do this correctly to avoid problems. 

If you'd like to do this later in the name of getting something working, I don't mind at all. Just please create an issue, write a TODO, and explain the problem at hand.

wallet/src/account/mod.rs
TheQuantumPhysicist: To be fair, there's tons of code that's implemented in the TxVerifier and Gosha has a fair point. You're just starting with the wallet and UTXOs, Pavel, and you'll have to reinvent everything. Perhaps it's a good idea to try to reuse code when we can. Again, as usual, not right now for this particular problem, but in general we want to avoid code duplication.

wallet/types/src/wallet_tx.rs
```diff
@@ -15,45 +15,37 @@
 
 use serialization::{Decode, Encode};
 
-use common::chain::{GenBlock, Transaction};
+use common::chain::{Block, Transaction};
 use common::primitives::id::WithId;
-use common::primitives::Id;
-
-#[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
-pub enum TxState {
```
TheQuantumPhysicist, line 23: Why is this being removed? This is really important information for the user... to know whether their transaction is confirmed or has a problem.

wallet/storage/src/lib.rs
```diff
@@ -42,7 +42,7 @@ pub trait WalletStorageRead {
     fn get_utxo_set(&self, account_id: &AccountId) -> Result<BTreeMap<AccountOutPointId, Utxo>>;
     fn get_transaction(&self, id: &AccountTxId) -> Result<Option<WalletTx>>;
     fn get_transactions(&self, account_id: &AccountId) -> Result<BTreeMap<AccountTxId, WalletTx>>;
-    fn get_account(&self, id: &AccountId) -> Result<Option<AccountInfo>>;
+    fn get_account_infos(&self) -> crate::Result<BTreeMap<AccountId, AccountInfo>>;
```
TheQuantumPhysicist, line 45: info doesn't have a plural in English. It's just info.

wallet/src/wallet/tests.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    let wallet =
+        Wallet::new_wallet(Arc::clone(&chain_config), Arc::clone(&db), MNEMONIC, None).unwrap();
+    drop(wallet);
+
+    let _wallet = Wallet::load_wallet(chain_config, db).unwrap();
+}
+
+#[test]
+fn wallet_balance_genesis() {
```
TheQuantumPhysicist, line 120: Many of these tests require randomization...

wallet/src/wallet/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    InvalidAddress(String, PublicKeyHashError),
+}
+
+/// Result type used for the wallet
+pub type WalletResult<T> = Result<T, WalletError>;
+
+pub struct Wallet<B: storage::Backend> {
+    chain_config: Arc<ChainConfig>,
+    db: Arc<Store<B>>,
+    // key_chain: MasterKeyChain<B>,
```
TheQuantumPhysicist, line 87: Commented code

wallet/src/key_chain/leaf_key_chain/mod.rs
```diff
@@ -418,6 +419,16 @@ impl LeafKeyChain {
         false
     }
 
+    pub fn is_destination_mine(&self, dest: &Destination) -> bool {
+        match dest {
+            Destination::Address(pkh) => self.is_public_key_hash_mine(pkh),
+            Destination::PublicKey(pk) => self.is_public_key_mine(pk),
+            Destination::AnyoneCanSpend
+            | Destination::ScriptHash(_)
+            | Destination::ClassicMultisig(_) => false,
```
TheQuantumPhysicist, line 428: This can also have another public key hash function

TheQuantumPhysicist: I'm not sure what steps we should take to support this... but whether now or later, this should have some TODO on it at least.

TheQuantumPhysicist:
I'm done with my first round. I'll take another round once you're done. Cheers!

wallet/src/key_chain/leaf_key_chain/mod.rs
pavel-kokolemin: TODO added.

wallet/types/src/wallet_tx.rs
pavel-kokolemin: Reverted, thanks.

serialization/src/hex_encoded.rs
pavel-kokolemin: OK, renamed.
But I think the `take` name is usually used with mutable references:
https://doc.rust-lang.org/std/mem/fn.take.html
https://doc.rust-lang.org/std/option/enum.Option.html#method.take
And `into_inner` when `self` is consumed:
https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.into_inner

crypto/src/key/hdkd/u31.rs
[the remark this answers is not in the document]
pavel-kokolemin: Please look one more time. Both `U31` and `ChildNumber` are now serialized as little-endian u32 for consistency. The `DerivationPath` encoding remains the same (where a vector of `ChildNumber` is serialized as a list of big-endian u32s).

wallet/src/account/mod.rs
pavel-kokolemin: I think it's too early to spend a lot of time on that. Once we support staking/decommissioning, we could think about that. I added a few TODOs.

wallet/src/account/mod.rs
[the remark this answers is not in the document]
pavel-kokolemin: I added TODO.

serialization/src/hex_encoded.rs
TheQuantumPhysicist: I see. Actually we do use `take` in WithId too. If we want to change this we have to fix it there too.

crypto/src/key/hdkd/u31.rs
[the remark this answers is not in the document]
pavel-kokolemin: Serialization has been removed from `ChildNumber` to avoid possible confusion. U31 is serialized the same way as u32 (little-endian).

wallet/src/key_chain/account_key_chain/mod.rs
```diff
@@ -173,17 +183,44 @@ impl AccountKeyChain {
         }
     }
 
+    pub fn get_private_key_for_destination(
+        &self,
+        destination: &Destination,
+    ) -> KeyChainResult<ExtendedPrivateKey> {
```
TheQuantumPhysicist, line 189: Can we return `KeyChainResult<Option<ExtendedPrivateKey>>` to distinguish between errors and not finding the key?

wallet/src/wallet/tests.rs
TheQuantumPhysicist: Please add a TODO for this

wallet/types/src/wallet_tx.rs
```diff
@@ -49,7 +47,7 @@ impl WalletTx {
         WalletTx { tx, state }
     }
 
-    pub fn get_tx(&self) -> &WithId<Transaction> {
+    pub fn tx(&self) -> &WithId<Transaction> {
```
TheQuantumPhysicist, line 50: How about `state()` for the other function too?

TheQuantumPhysicist (approved):
Looking good! Please fix the few issues that are left, then feel free to merge.

PR #846, opened 2023-04-20
mempool: Error refactoring + ban score

### LINKED ISSUE #770: Implement a proper ban score for mempool errors
Not every mempool TxValidationError error means that the transaction was malicious, double-spends may happen due to a minor reorg and a peer shouldn't be punished in such cases. Perhaps this can be done by implementing the `BanScore` trait for the `TxValidationError` error.

iljakuklic:
* Categorize errors, clean up some error-related code
* Remove some unused error enum arms
* Introduce mempool ban score
* Integrate mempool ban score with p2p

Fixes #770


mempool/src/error/ban_score.rs
```diff
@@ -0,0 +1,300 @@
...
+            E::InvariantErrorDelegationCreationFailedIdAlreadyExists => 0,
+            E::AttemptedDecommissionNonexistingPoolBalance => 0,
+            E::AttemptedDecommissionNonexistingPoolData => 0,
+            E::DelegationCreationFailedPoolDoesNotExist => 0,
+            E::DelegateToNonexistingId => 0,
+            E::DelegateToNonexistingPool => 0,
+            E::InvariantErrorPoolCreationReversalFailedBalanceNotFound => 0,
+            E::InvariantErrorPoolCreationReversalFailedDataNotFound => 0,
+            E::InvariantErrorPoolCreationReversalFailedAmountChanged => 0,
+            E::InvariantErrorDelegationShareNotFound => 0,
```
iljakuklic, line 270: What do you think about banning a peer when these occur @azarovh? (See the comment at the top)

TheQuantumPhysicist: Two things:

1. Peers that are out of sync are not supposed to relay transactions
2. There's a general problem that has to be handled for PoS, which is that stakes on different branches can appear as double-spent. That will come later.

I'm not sure about these cases specifically and can't do a deep-check it's 3:30 AM here. But I can say that there's some tolerance in some cases based on how bitcoin does this because we cannot guarantee that all peers are at the same tip all the time; there may be slight discrepancy due to network latency or temporary forks.

mempool/src/pool/tests/utils.rs
```diff
@@ -77,3 +77,66 @@ pub fn estimate_tx_size(num_inputs: usize, num_outputs: usize) -> usize {
     );
     result
 }
+
+// TODO this calculation is already done in ChainState, reuse it
```
TheQuantumPhysicist, line 81: Actually the calculation comes straight from the transaction verifier, which is available now.

mempool/src/pool/try_get_fee.rs
```diff
@@ -1,80 +0,0 @@
...
-    async fn try_get_fee(&self, tx: &SignedTransaction) -> Result<Fee, TxValidationError>;
-}
-
-#[async_trait::async_trait]
-impl<M> TryGetFee for Mempool<M>
-where
-    M: GetMemoryUsage + Send + Sync,
-{
-    // TODO this calculation is already done in ChainState, reuse it
-    async fn try_get_fee(&self, tx: &SignedTransaction) -> Result<Fee, TxValidationError> {
```
azarovh, line 33: nice! I've been recently in this function, but didn't notice that it is used only in tests and can be removed. I think it make sense now to also remove `ChainstateInterface::get_inputs_outpoints_coin_amount` and `ChainstateInterface::get_outputs_coin_amount` that were used here

azarovh (approved):

mempool/src/pool/try_get_fee.rs
iljakuklic: I removed `ChainstateInterface::get_outputs_coin_amount`, it's no longer used anywhere. However, `ChainstateInterface::get_inputs_outpoints_coin_amount` is still used in the test helper function implementation. Any more suggestions?

mempool/src/error/ban_score.rs
iljakuklic: Will leave this as a todo for now and do proper research on how to handle cases where peers may be out of sync. Will be included in a future PR.

TheQuantumPhysicist: No problem. Please open an issue referencing, just in case.

stanislav-tkach (approved):

mempool/src/pool/try_get_fee.rs
azarovh: no all good

PR #850, opened 2023-04-23
Separate node executable from common functionality in preparation for a GUI

TheQuantumPhysicist:
We're preparing to have a GUI for the node (and potentially the wallet).

I extracted the executable part of the node into a separate crate, as a GUI will have a different launcher that will use the new "node-lib".


pavel-kokolemin (approved):

iljakuklic (approved):
I am fine with this change.

However there are some comments:
* I don't think `node-tui` is an accurate name as it's not a user interface for the node. It's the node daemon executable itself.
* I don't quite understand how will it help with the GUI. Is the GUI supposed to embed the whole node in it? I'd expect it to communicate with the node through RPC or something like that.
* The original `node` had both library and executable targets. It was already possible to use it as a library. What is the motivation for creating a separate crate?

TheQuantumPhysicist:
> There are several unknown lint: `clippy::extra_unused_type_parameters` in unchanged files. Looks strange, is it ok?

Old version of rust doesn't see it.

PR #851, opened 2023-04-24
Add a basic block scan to wallets

pavel-kokolemin:


wallet/wallet-controller/src/lib.rs
```diff
@@ -36,14 +38,27 @@ pub enum ControllerError {
 
 pub struct Controller<T> {
     rpc_client: T,
-    _wallet: DefaultWallet,
+    _wallet: Arc<Mutex<DefaultWallet>>,
```
TheQuantumPhysicist, line 41: Why do we need `Arc<Mutex>` here? I would really not have to do this early on instead of making this technical debt.

wallet/wallet-cli/src/commands/mod.rs
```diff
@@ -29,6 +32,18 @@ pub enum WalletCommands {
...
     BestBlockHeight,
 
+    /// Get a block hash at height
+    BlockHash {
+        /// Block height
+        height: BlockHeight,
+    },
+
+    /// Get a block by its hash
+    Block {
```
TheQuantumPhysicist, line 42: GetBlock

wallet/wallet-controller/src/lib.rs
```diff
@@ -149,16 +164,21 @@ impl<T: NodeInterface> Controller<T> {
...
-pub type HandlesController = Controller<WalletHandlesClient>;
+impl<T> Drop for Controller<T> {
+    fn drop(&mut self) {
+        // TODO: Wait for the sync task to stop
+        self.sync_handle.abort();
+    }
+}
 
 pub async fn make_rpc_controller(
+    chain_config: Arc<ChainConfig>,
```
TheQuantumPhysicist, line 175: What's the point of passing chain_config and returning it?

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/// Sync the wallet state (blocks) from the node.
+/// Returns true if the wallet state has changed and false otherwise.
+async fn sync_blocks<T: NodeInterface>(
+    chain_config: &ChainConfig,
+    rpc_client: &mut T,
+    wallet: Arc<Mutex<DefaultWallet>>,
+) -> Result<bool, SyncError> {
+    // TODO: Make it more efficient: download blocks concurrently and send fewer requests
+    let wallet_block_height = match wallet
```
TheQuantumPhysicist, line 39: The height shouldn't be used to determine if there's a change. You should instead store the best block id. It's always possible to switch between two different blockchain states while maintaining the same height.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+            .map_err(|e| SyncError::UnexpectedRpcError(e.to_string()))?
+        {
+            Some(id) => id,
+            None => {
+                return Err(SyncError::UnexpectedRpcError(format!(
+                    "Node block height is {node_block_height} but block at height {wallet_block_height} is None"
+                )));
+            }
+        };
+        let wallet_block_id = match wallet
```
TheQuantumPhysicist, line 74: I see what you're doing down there... you're doing that manually. You shouldn't. 

First, again, you're depending on the height. That's wrong.

Second, there's a function for doing that in chainstate (finding the common ancestor). Perhaps you should add a chainstate interface function and call it, which takes two arbitrary block ids, and returns the common ancestor.

wallet/wallet-controller/src/lib.rs
pavel-kokolemin: Sorry, what do you mean? `chain_config` is passed to the `Controller::new` call.

wallet/wallet-controller/src/sync/mod.rs
pavel-kokolemin: Okay, thanks. This should be fixed now.

wallet/wallet-controller/src/sync/mod.rs
pavel-kokolemin: I added `get_last_common_height` and use it to sync the wallet. Block ids are used to determine if the node state has changed.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+/// Sync the wallet state (blocks) from the node.
+/// Returns true if the wallet state has changed and false otherwise.
+async fn sync_blocks<T: NodeInterface>(
+    chain_config: &ChainConfig,
+    rpc_client: &mut T,
+    wallet: Arc<Mutex<DefaultWallet>>,
+) -> Result<bool, SyncError> {
+    // TODO: Make it more efficient: download blocks concurrently and send fewer requests
+    let wallet_block_height = match wallet
+        .lock()
```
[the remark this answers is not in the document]
pavel-kokolemin: Thanks, should be done now.

wallet/wallet-controller/src/lib.rs
pavel-kokolemin: `Mutex` is not used now.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -0,0 +1,172 @@
...
+                    SyncEvent::Reset {
+                        block_id,
+                        block_height,
+                    } => {
+                        wallet_block_id = *block_id;
+                        wallet_block_height = *block_height;
+                    }
+                    SyncEvent::NewBlock { block } => {
+                        wallet_block_id = block.header().block_id().into();
+                        wallet_block_height = wallet_block_height.next_height();
```
TheQuantumPhysicist, line 135: Updating the state locally, as discussed, is a bad idea. The problem here is that if anything goes wrong in the processing or sending, this event loop will be perpetually running for ever, sending blocks the wallet cannot recognize, causing more trouble, because nothing can stop it, ever, other than shutting the program down.

pavel-kokolemin: Right, thanks. Should be fixed now. Block sync is started from the event loop now.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+
+pub type BlockFetchResult<T> = Result<FetchedBlock, FetchBlockError<T>>;
+
+pub async fn run_state_sync<T: NodeInterface>(state_tx: mpsc::Sender<ChainInfo>, rpc_client: T) {
+    let mut last_state = None;
+    while !state_tx.is_closed() {
+        let state_res = rpc_client.chainstate().await;
+        match state_res {
+            Ok(state) => {
+                if last_state.as_ref() != Some(&state) {
```
TheQuantumPhysicist, line 59: It's better to just compare the tip hash. It's wasteful to have to compare everything.

pavel-kokolemin: This check is negligible compared to the cost of the RPC call itself.

chainstate/src/interface/chainstate_interface_impl.rs
```diff
@@ -286,6 +286,28 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> ChainstateInterfa
             .map_err(ChainstateError::FailedToReadProperty)
     }
 
+    fn last_common_block(
```
TheQuantumPhysicist, line 289: last_common_ancestor

wallet/wallet-cli/src/main.rs
```diff
@@ -107,22 +109,37 @@ async fn run(output: &ConsoleContext) -> Result<(), WalletCliError> {
...
+
+    let output_copy = output.clone();
+    let repl_command = repl::get_repl_command();
+    let create_line_editor =
+        repl::create_line_editor(repl_command.clone(), output, &data_dir, vi_mode)?;
+
+    std::thread::spawn(move || {
+        // Run a blocking loop in a separate thread
+        repl::run(create_line_editor, repl_command, event_tx, &output_copy);
+    });
```
TheQuantumPhysicist, line 131: You should join this at the end of the program. We don't need more issues like p2p.

wallet/wallet-cli/src/main.rs
```diff
@@ -107,22 +109,37 @@ async fn run(output: &ConsoleContext) -> Result<(), WalletCliError> {
...
+
+    even_loop::run(&mut controller, event_rx).await;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Another thing, the controller is only used here. You can put both this and the controller in one scope for better isolation.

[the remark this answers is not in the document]
TheQuantumPhysicist: Also I would call that cli_event_loop... for clarity.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+    BlockNotFound(Id<Block>),
+    #[error("Invalid prev block id: {0}, expected: {1}")]
+    InvalidPrevBlockId(Id<GenBlock>, Id<GenBlock>),
+}
+
+pub type BlockFetchResult<T> = Result<FetchedBlock, FetchBlockError<T>>;
+
+pub async fn run_state_sync<T: NodeInterface>(state_tx: mpsc::Sender<ChainInfo>, rpc_client: T) {
+    let mut last_state = None;
+    while !state_tx.is_closed() {
```
TheQuantumPhysicist, line 55: Nitpick: Please add an empty line here

chainstate/src/interface/chainstate_interface_impl.rs
pavel-kokolemin: We already have `last_common_ancestor`:
```
    fn last_common_ancestor(
        &self,
        first_block_index: &GenBlockIndex,
        second_block_index: &GenBlockIndex,
    ) -> Result<GenBlockIndex, ChainstateError> {
```
(I can't use it directly because we don't have `GenBlockIndex` in the wallet controller).

wallet/wallet-controller/src/lib.rs
```diff
@@ -17,148 +17,279 @@
...
         self.rpc_client
             .p2p_remove_reserved_node(address)
             .await
-            .map_err(|e| ControllerError::RpcError(e.to_string()))
+            .map_err(ControllerError::RpcError)
+    }
+
+    fn handle_node_state_change(&mut self, chain_info: ChainInfo) {
+        log::info!(
+            "Node chain info updated, best block height: {}, best block id: {}",
```
TheQuantumPhysicist, line 177: chainstate

wallet/wallet-controller/src/lib.rs
```diff
@@ -17,148 +17,281 @@
...
+                chain_info_opt = self.node_state_rx.recv() => {
+                    // Channel is always open because [run_tip_sync] does not return
+                    self.handle_node_state_change(chain_info_opt.expect("Channel must be open"));
+                }
+                sync_result = Self::recv_block_fetch_result(&mut self.block_fetch_task) => {
+                    self.handle_block_fetch_result(sync_result);
+                }
+            }
+        }
     }
```
TheQuantumPhysicist, line 275: It would be great if all these went into a separate file that can be tested in isolation...

chainstate/src/interface/chainstate_interface_impl.rs
TheQuantumPhysicist: last_common_ancestor_by_id?

wallet/wallet-controller/src/sync/mod.rs
TheQuantumPhysicist: Chainstate info can, in the future, contain things like relative time or whatever. It's not really safe to do it this way.

TheQuantumPhysicist:
The design now is fine. I hope I didn't miss anything. The only issue I see is that it's too convoluted with zero testing. This concerns me.

We should be able to test this system using mocks or similar. In the context of the main problem we're solving in this PR, the test would show that the basic behavior works. Meaning, when the chainstate returns a different block than stored in the sync event loop, block pulling will keep running until it's equal to the state. This can be achieved by defining the proper traits that can be mocked. Currently, everything uses concrete types. This can be changed to something more general, but it needs some thought. I'm happy to have a call to discuss some possible designs that makes it testable.

wallet/wallet-controller/src/lib.rs
```diff
@@ -17,148 +17,281 @@
...
+use sync::{BlockFetchResult, FetchedBlock};
+use tokio::{sync::mpsc, task::JoinHandle};
 use wallet::DefaultWallet;
 
 #[derive(thiserror::Error, Debug)]
-pub enum ControllerError {
+pub enum ControllerError<T: NodeInterface> {
     #[error("RPC error: {0}")]
-    RpcError(String),
+    RpcError(T::Error),
```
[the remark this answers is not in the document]
pavel-kokolemin: `NodeInterfaceCallError` is too verbose IMO, I changed to `NodeCallError`.

wallet/wallet-controller/src/sync/mod.rs
pavel-kokolemin: Okay, updated, thought I still don't see what could possibly go wrong if we have up-to-date `ChainInfo`.

wallet/wallet-controller/src/lib.rs
```diff
@@ -17,148 +17,279 @@
...
+        // Wait until the node has enough block height.
+        // Block sync may not work correctly otherwise.
+        if node_block_id == wallet_block_id || node_block_height < wallet_block_height {
+            return;
+        }
+
+        let chain_config = Arc::clone(&self.chain_config);
+        let mut rpc_client = self.rpc_client.clone();
+
+        self.block_fetch_task = Some(tokio::spawn(async move {
```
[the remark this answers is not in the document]
pavel-kokolemin: It must be stored internally to make the `run` function cancel safe.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -0,0 +1,362 @@
...
+    error_delay: Duration,
+}
+
+impl BlockSyncingConfig {
+    pub fn new() -> Self {
+        Self {
+            normal_delay: Duration::from_secs(1),
+            error_delay: Duration::from_secs(10),
+        }
+    }
```
TheQuantumPhysicist, line 109: Default impl?

wallet/wallet-controller/src/sync/tests/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    fn new(chain_config: &ChainConfig, new_tip_tx: mpsc::UnboundedSender<Id<Block>>) -> Self {
+        Self {
+            genesis_id: chain_config.genesis_block_id(),
+            blocks: Vec::new(),
+            new_tip_tx,
+        }
+    }
+
+    fn get_best_block_id(&self) -> Id<GenBlock> {
+        self.blocks.last().cloned().map_or(self.genesis_id, Into::into)
```
TheQuantumPhysicist, line 46: No tests for reorgs?

TheQuantumPhysicist (approved):
Looking good!

wallet/wallet-controller/src/lib.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Now it's OK since it's isolated in the sync struct.

wallet/wallet-controller/src/sync/tests/mod.rs
pavel-kokolemin: Reorgs are tested in all tests.

PR #852, opened 2023-04-24
Fix/raii_job_removal_in_block_prod

TheQuantumPhysicist:


blockprod/src/detail/mod.rs
```diff
@@ -189,27 +173,38 @@ impl BlockProduction {
...
 
         // At the end of this function, the job has to be removed
-        // Once the job key is used, we swap it with None... this is all temporary.
-        // The docs in the function self.stop_job_in_scope()
-        let job_key_destroyed = AtomicBool::new(false);
-        let _job_remove_checker = OnceDestructor::new(|| {
-            assert!(job_key_destroyed.load(std::sync::atomic::Ordering::SeqCst));
-        });
+        let (job_remover_func, end_confirm_receiver) = self.job_manager.make_job_stopper_function();
+        let _job_remover = OnceDestructor::new(move || job_remover_func(job_key));
```
alfiedotwtf, line 207: Can these two lines be put inside `add_job()`?

Also, instead of returning the block as is, to reduce the amount of things that are returned and reduce API complexity, maybe `end_confirm_receiver` should just be `block_receiver` and so the below code of:

```
        let (block, end_receiver) = self.produce_block(reward_destination, transactions).await?;
        let _finished = end_receiver.await;
        Ok(block)
```

instead becomes something like:

```
        let block_receiver = self.produce_block(reward_destination, transactions).await?;
        let block = block_receiver.await;
        Ok(block)
```

or the uglier but shorter:

```
        let block = self.produce_block(reward_destination, transactions).await?.await;
        Ok(block)
```

blockprod/src/interface/blockprod_interface_impl.rs
```diff
@@ -42,6 +42,8 @@ impl BlockProductionInterface for BlockProduction {
             None => crate::detail::TransactionsSource::Mempool,
         };
 
-        self.generate_block(reward_destination, transactions).await
+        let (block, end_receiver) = self.produce_block(reward_destination, transactions).await?;
+        let _finished = end_receiver.await;
```
alfiedotwtf, line 46: Maybe this should be `end_receiver.await?;` just in case the other end closes/errors out

alfiedotwtf (approved):

blockprod/src/detail/mod.rs
TheQuantumPhysicist: We can do this but then we'll need to create a custom future. I'm not up for this, tbh. Right now it behaves correctly and that's the concern I wanted to fix.

Feel free to add a TODO for this.

blockprod/src/interface/blockprod_interface_impl.rs
TheQuantumPhysicist: The only error that can happen is if the channel is closed. We don't care about that. I'll add a comment about it.

muursh (approved):

PR #853, opened 2023-04-24
First basic GUI for the mintlayer-core node

TheQuantumPhysicist:
Basic GUI design with a model-view-controller paradigm. The controller will be used to receive events and data from the node, and will also be used to send commands to node.

The plan is to have the controller subscribe to events and retrieve information and set local variables, which in turn update the GUI values. The controller also has the capability to send commands directly to the node through subsystems.

Right now, the GUI basically initializes the node on startup properly with async tasks, and shuts the node down properly when the window is closed.


node-lib/src/runner.rs
```diff
@@ -124,35 +127,49 @@ pub async fn initialize(
...
     Ok(manager)
 }
 
 /// Processes options and potentially runs the node.
-pub async fn run(options: Options) -> Result<()> {
-    match options.command {
+pub async fn run(
+    options: Options,
+    node_controller_sender: Option<oneshot::Sender<NodeController>>,
+) -> Result<subsystem::Manager> {
```
iljakuklic, line 162: It seems to me `run` is no longer an appropriate name here, since it does not invoke `Manager::main` but returns the manager instead. Maybe something like `setup`?

node-gui/src/backend_controller.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+impl NodeBackendController {
+    pub async fn initialize() -> anyhow::Result<NodeBackendController> {
+        let (node_controller_sender, node_controller_receiver) = oneshot::channel();
+
+        if std::env::var("RUST_LOG").is_err() {
+            std::env::set_var("RUST_LOG", "info");
+        }
```
iljakuklic, line 42: Variable not set is not the only error condition here. It may be appropriate to handle the `NotUnicode` variant differently. (I am inclined to say a log entry would be nice but that is a chicken and egg problem)

https://doc.rust-lang.org/std/env/enum.VarError.html

node-gui/src/backend_controller.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// If the shutdown was already triggered, returns None.
+    pub fn trigger_shutdown(&mut self) -> Option<tokio::task::JoinHandle<()>> {
+        if self.manager_join_handle.is_none() {
+            // We shutdown and join only once, so this being None means we took the handle already
+            logging::log::warn!("Shutdown already requested.");
+            return None;
+        }
+        logging::log::info!("Starting shutdown process...");
+
+        self.controller.shutdown_trigger.clone().initiate();
```
iljakuklic, line 80: @stanislav-tkach changed the `initiate` method to consume the trigger. However, with weak receiver being used (#856), this may no longer be necessary. Would allow us to avoid the clone here.

Thoughts? @stanislav-tkach @pavel-kokolemin

iljakuklic (approved):
Looks good. Some minor comments below:

node-gui/src/backend_controller.rs
TheQuantumPhysicist: Yeah, I know. I saw the enum in there, but frankly I don't care. If someone setup a wrong value for the env var, I think it's OK to set it to a default value in the scope of the program.

node-gui/src/backend_controller.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        let mut join_handle = None;
+        std::mem::swap(&mut self.manager_join_handle, &mut join_handle);
+
+        join_handle
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Oh... I seem to have confused it with some other take or replace function.

Done.

node-lib/src/runner.rs
TheQuantumPhysicist: That's fair. Done.

node-gui/src/backend_controller.rs
stanislav-tkach: I don't mind reverting the `initiate` method signature to taking `&self`.

TheQuantumPhysicist: I can revert it in this PR. Should I?

pavel-kokolemin (approved):

node-gui/src/backend_controller.rs
pavel-kokolemin: `&self` would be a bit cleaner, but it's probably not that important. I'm fine with either way.

TheQuantumPhysicist: I'm ignoring this for now. I don't really care that much.

iljakuklic: Probably a wash in the end. We can leave it as is, it does not make all that much sense to initiate shutdown multiple times.

PR #854, opened 2023-04-25
Distribute reward to delegators

azarovh:
Added new module `reward_distribution` that distributes subsidy + fees among staker and delegators.

I had to do small refactoring in order to pass `Fee` to `connect_block_reward`. I removed `connect_transactable` function as it doesn't do much but rather complicates things by generalizing code that should be separated. As a result `BlockTransactableRef` was also removed. I think it's much simpler and cleaner now.


chainstate/tx-verifier/src/transaction_verifier/transferred_amount_check.rs
```diff
@@ -278,13 +279,21 @@ pub fn check_transferred_amount_in_reward<U: UtxosView, P: PoSAccountingView>(
...
-            .ok_or_else(|| ConnectTransactionError::RewardAdditionError(block_id))?;
+    let max_allowed_outputs_total = match consensus_data {
+        ConsensusData::None | ConsensusData::PoW(_) => {
+            amount_sum!(inputs_total, block_subsidy_at_height.0, total_fees.0)
+                .ok_or_else(|| ConnectTransactionError::RewardAdditionError(block_id))?
+        }
+        ConsensusData::PoS(_) => {
+            // TODO: it doesn't actually make much sense to check amounts for PoS, because reward doesn't
+            // transfer anything in terms of utxos.
+            inputs_total
```
TheQuantumPhysicist, line 290: Are there tests for this behavior? Let's make sure this is tested so that it cannot be abused in a way we can't imagine. When you're writing this test, make sure all holes are closed.

chainstate/test-suite/src/tests/helpers/pos.rs
```diff
@@ -113,7 +113,7 @@ pub fn create_stake_pool_data(
         anyonecanspend_address(),
         vrf_pk,
         destination,
-        0,
+        1000,
```
TheQuantumPhysicist, line 116: Doesn't look like good randomization of test data.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,338 @@
...
+    chain::{Block, PoolId},
+    primitives::{Amount, Id},
+    Uint256,
+};
+use pos_accounting::{AccountingBlockRewardUndo, PoSAccountingOperations, PoSAccountingView};
+
+use super::accounting_delta_adapter::PoSAccountingDeltaAdapter;
+
+/// Distribute reward among the staker and delegators
+pub fn distribute_reward<P: PoSAccountingView>(
```
TheQuantumPhysicist, line 28: distribute_pos_reward

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,338 @@
...
+    let total_staker_reward = (total_reward - pool_data.cost_per_block())
+        .and_then(|v| (v * pool_data.margin_ratio_per_thousand().into()).and_then(|v| v / 1000))
+        .and_then(|v| v + pool_data.cost_per_block())
+        .ok_or(ConnectTransactionError::StakerRewardCalculationFailed(
+            block_id,
+        ))?;
+    let total_delegators_reward = (total_reward - total_staker_reward).ok_or(
+        ConnectTransactionError::DelegatorRewardCalculationFailed(block_id),
+    )?;
+    let total_delegators_reward = Uint256::from_amount(total_delegators_reward);
```
TheQuantumPhysicist, line 50: I do understand why you're doing this conversion, but I have to say that doing math operations with Uint256 is playing with fire, because none of the operations are checked for overflow/underflow problems.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,338 @@
...
+                delegation_shares.values().try_fold(Amount::ZERO, |acc, v| {
+                    (acc + *v).ok_or(ConnectTransactionError::DelegatorsRewardSumFailed(block_id))
+                })?;
+            let total_delegators_balance = Uint256::from_amount(total_delegators_balance);
+
+            delegation_shares
+                .iter()
+                .map(|(delegation_id, balance)| {
+                    let balance = Uint256::from_amount(*balance);
+                    let reward = total_delegators_reward * balance / total_delegators_balance;
```
TheQuantumPhysicist, line 70: I have a feeling you're doing this and thinking that you're "covered with Uint256 because the precision is twice the multiplicands"... but you're not doing this right. In this line, first, `balance / total_delegators_balance` will happen first, and then the result will be multiplied by `total_delegators_reward`. If this is what you really intended, then I gotta ask why you converted to Uint256 in the first place.

azarovh: Why division should happen first? 
I thought that operations are left to right as per: https://doc.rust-lang.org/reference/expressions.html#expression-precedence

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,338 @@
...
+                    let reward: u128 = reward.try_into().map_err(|_| {
+                        ConnectTransactionError::DelegatorRewardCalculationFailed(block_id)
+                    })?;
+
+                    accounting_adapter
+                        .operations(tx_source)
+                        .delegate_staking(*delegation_id, Amount::from_atoms(reward))
```
TheQuantumPhysicist, line 77: Keep in mind that since you're using division and fractional arithmetic with integers, it's not always correct that the sum of all amounts is equal to the total.

An easy example is dividing 4 among shares of 1:1:1. That's basically 4/3+4/3+4/3 = 3 which is not 4, which means you'll have 1 unit undistributed. Did you account for that?

I'm not saying you have to hand it to someone, but you have to define the protocol parameters, by which such "burning" is acceptable. You're basically burning MLT for this. 

What happens if the amounts are so small that you got some zeros?

Usually the better way to do this (from an integer arithmetic mathematical point of view) is to do the distribution in memory in some data structure, confirm the constraints, and then write it to your cache/deltas. This conforms to the read-modify-write paradigm. Though the cost here is a temporary allocation of a relatively large data structure in memory.

You can't have it both ways. Either pre-calculate, do your constraints verification and ensure that the amounts are sane, OR, have this kind of vulnerabilities where integer arithmetic policy is unclear.

The alternative is some ugly, mutable data structure that changes while you're writing on disk. I don't prefer that.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: Multiplying uint128 shouldn't ever overflow uint256 and doing division should never underflow. So as long as we are doing only this set of operations we are safe, aren't we?

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: Division always happens first in arithmetic. I'll double check to see why rust is claiming to do it like this.

azarovh: `println!("{}", 3 * 3 / 2);` gives me 4 not 3

azarovh: I'll add parentheses anyway to avoid confusion in the future

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: Don't forget division by zero.

Generally speaking we ensure such mistakes won't happen by guarding operations with runtime checks. That's all I'm saying here. If someone decides to change the formula later, they may not notice the assumptions you made.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: That's better. 

Weird. I could swear it's not like that in Mathemtica and C++, but now I have to recheck my understanding of this.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: Do you suggest another approach instead of burning or is it the implementation that bothers you?

For me, it's either we figure out what to do with the remainder (burn, tip owner, give to a delegator) or play with fixed-point balances.

TheQuantumPhysicist: Basically what I'm saying is that with such algorithm we should be careful because of corner cases. I suggest moving the numeric distribution function to a separate function that can be tested and where corner cases are verified.

Maybe giving the remainder to the staker is OK. I don't see another way.

chainstate/test-suite/src/tests/helpers/pos.rs
azarovh: I believe it will just complecate tests. In all the use cases we either don't care or want 100% to the owner

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: Split into smaller functions and added check for zero devision

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: Now all the remainder goes to the owner. Also if there are delegations but their balance is 0 the reward also goes to the owner. Tests included.

common/src/primitives/per_thousand.rs
```diff
@@ -0,0 +1,65 @@
...
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug, Encode, Decode)]
+pub struct PerThousand(#[codec(compact)] u16);
+
+impl PerThousand {
+    pub fn new(value: u16) -> Option<Self> {
+        if value <= 1000 {
```
TheQuantumPhysicist, line 23: Besides that there are issues with this definition of PerThousand, based on your design this should probably be `value < 1000`.

common/src/primitives/per_thousand.rs
```diff
@@ -0,0 +1,65 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Debug, Encode, Decode)]
```
TheQuantumPhysicist, line 18: Note about type-safety: Whenever Decode exists, you basically have a de-facto method to circumvent whatever type-safety rules you created. Meaning: If you think now that `PerThousand` can never have a `value >= 1000`, think again. This can now be circumvented through `Decode`.

In other words: Never rely on type-safety alone if you'll allow Decode.

You can see this problem being solved in the multisig implementation I created a monto or two ago, for example. It's the same issue.

IF you want to do this, then you have to implement `Decode` yourself and make decoding a `value >= 1000` an error.

chainstate/storage/src/schema.rs
```diff
@@ -69,6 +69,8 @@ storage::decl_schema! {
         pub DBAccountingPoolDataTip: Map<PoolId, PoolData>,
         /// Store for tip accounting pool balances
         pub DBAccountingPoolBalancesTip: Map<PoolId, Amount>,
+        /// Store for tip accounting pool owner balances
+        pub DBAccountingPoolOwnerBalancesTip: Map<PoolId, Amount>,
```
TheQuantumPhysicist, line 73: So at the lowest level, this probably goes here... why not include this value in PoolData?

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+fn calculate_pool_owner_reward(total_reward: Amount, pool_data: &PoolData) -> Option<Amount> {
+    (total_reward - pool_data.cost_per_block())
+        .and_then(|v| v * pool_data.margin_ratio_per_thousand().value().into())
+        .and_then(|v| v / 1000)
+        .and_then(|v| v + pool_data.cost_per_block())
+}
+
+/// The reward is distributed among delegations proportionally to their balance
+fn distribute_delegations_pos_reward<P: PoSAccountingView>(
+    accounting_adapter: &mut PoSAccountingDeltaAdapter<P>,
```
TheQuantumPhysicist, line 90: Can we here instead pass a `&mut impl PoSAccountingOperationImpl`? Same for the whole file. This file doesn't need to know anything about the adapter.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+        .get_pool_delegations_shares(pool_id)?
+        .map(|delegation_shares| {
+            let total_delegations_balance =
+                delegation_shares.values().try_fold(Amount::ZERO, |acc, v| {
+                    (acc + *v).ok_or(ConnectTransactionError::DelegationsRewardSumFailed(
+                        block_id, pool_id,
+                    ))
+                })?;
+
+            if total_delegations_balance != Amount::ZERO {
```
TheQuantumPhysicist, line 106: Nitpick: I always prefer > 0 compared to != 0 in such cases.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+        })
+        .transpose()
+}
+
+fn calculate_rewards_per_delegation(
+    delegation_shares: &BTreeMap<DelegationId, Amount>,
+    block_id: Id<Block>,
+    total_delegations_balance: Amount,
+    total_delegations_reward: Amount,
+) -> Result<Vec<(DelegationId, u128)>, ConnectTransactionError> {
```
TheQuantumPhysicist, line 153: If you want to return u128, then please use the `UnsignedIntType` in the amount module.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+    let total_delegations_balance = Uint256::from_amount(total_delegations_balance);
```
TheQuantumPhysicist, line 154: Since we're doing the arithmetic here, let's make it more robust. Please add a static_assertion from that station_assertions crate where we assert that the Amount size is 16 bytes. This way, if that ever changes in the future, we know that Uint256 should be updated. You can use `static_assertions::assert_eq_size()`

chainstate/storage/src/schema.rs
azarovh: hmm, right, didn't see that option. Will do

common/src/uint/impls.rs
```diff
@@ -601,6 +601,18 @@ pub enum UintConversionError {
     ConversionOverflow,
 }
 
+impl TryFrom<Uint256> for u128 {
+    type Error = UintConversionError;
+
+    fn try_from(n: Uint256) -> Result<Self, UintConversionError> {
+        if n > Uint256::from_u128(u128::MAX) {
```
TheQuantumPhysicist, line 608: I noticed there are no tests for this function... can you please add some?

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: I agree but the problem is that I need `&mut impl PoSAccountingOperations` and `& impl PoSAccountingView` and I can't have both at the same time from a single object

common/src/primitives/per_thousand.rs
azarovh: will do, thanks for pointing out

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: Why not? Either use a generic explicitly (which is cleaner in this case, since impl is meant for simple things), or something like `&mut impl (PoSAccountingOperations + PoSAccountingView)` will work.

TheQuantumPhysicist: To be clear, what wouldn't work is `dyn`. You can only add auto traits to `dyn` with `+`.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+        .unwrap_or_default()
+        .into_iter()
+        .chain(vec![increase_pool_balance_undo].into_iter())
+        .collect();
+
+    Ok(AccountingBlockRewardUndo::new(undos))
+}
+
+fn calculate_pool_owner_reward(total_reward: Amount, pool_data: &PoolData) -> Option<Amount> {
+    (total_reward - pool_data.cost_per_block())
```
TheQuantumPhysicist, line 82: Given this, we should probably limit the cost_per_block in pool creation.

TheQuantumPhysicist: Please create an issue explaining the problem. We can tackle this later.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: It violates the borrows checker because I have single `PoSAccountingDelta` which implements `PoSAccountingView` and which is a parent to `PoSAccountingOperationImpl`

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+    let total_delegations_balance = Uint256::from_amount(total_delegations_balance);
+    let total_delegations_reward = Uint256::from_amount(total_delegations_reward);
+    delegation_shares
+        .iter()
+        .map(
+            |(delegation_id, balance)| -> Result<_, ConnectTransactionError> {
+                let balance = Uint256::from_amount(*balance);
+                let reward = (total_delegations_reward * balance) / total_delegations_balance;
+                let reward: u128 = reward.try_into().map_err(|_| {
+                    ConnectTransactionError::DelegationRewardOverflow(*delegation_id, block_id)
```
TheQuantumPhysicist, line 163: I would also add the relevant amounts to the error.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+    total_delegations_reward: Amount,
+) -> Result<Vec<(DelegationId, u128)>, ConnectTransactionError> {
+    let total_delegations_balance = Uint256::from_amount(total_delegations_balance);
+    let total_delegations_reward = Uint256::from_amount(total_delegations_reward);
+    delegation_shares
+        .iter()
+        .map(
+            |(delegation_id, balance)| -> Result<_, ConnectTransactionError> {
+                let balance = Uint256::from_amount(*balance);
+                let reward = (total_delegations_reward * balance) / total_delegations_balance;
```
TheQuantumPhysicist, line 161: Please check that `total_delegations_balance > 0`.

TheQuantumPhysicist: Return an error if so. Test for it as well.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+                            .delegate_staking(*delegation_id, Amount::from_atoms(*reward))
+                            .map_err(ConnectTransactionError::PoSAccountingError)
+                    })
+                    .collect::<Result<Vec<_>, _>>()?;
+
+                Ok(distribute_remainder_undo
+                    .unwrap_or_default()
+                    .into_iter()
+                    .chain(delegation_undos.into_iter())
+                    .collect())
```
TheQuantumPhysicist, line 136: Not only in the selected code, but in the whole function there are many things that can be made iterators. Now I this seems to be the easiest part to just use iterators without collecting to Vec. You can probably use FallibleIterator for that. It's used one in TransactionVerifier.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+            } else {
+                // if total balance of all delegations is 0 then give the reward to the pool's owner
+                let increase_pool_balance_undo = accounting_adapter
+                    .operations(TransactionSource::Chain(block_id))
+                    .increase_pool_owner_balance(pool_id, total_delegations_reward)?;
+                Ok(vec![increase_pool_balance_undo])
```
TheQuantumPhysicist, line 142: Is having this if/else an optimization? I think in general it shouldn't make a difference. You'll have zeros that will still be valid.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+        (total_delegations_reward - total_delegations_reward_distributed).ok_or(
+            ConnectTransactionError::DistributedDelegationsRewardExceedTotal(
+                pool_id,
+                block_id,
+                total_delegations_reward_distributed,
+                total_delegations_reward,
+            ),
+        )?;
+
+    if delegations_reward_remainder != Amount::ZERO {
```
TheQuantumPhysicist, line 199: Why not just an empty vector? The sum of that is zero as well.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,575 @@
...
+                    .map(|(delegation_id, reward)| {
+                        accounting_adapter
+                            .operations(TransactionSource::Chain(block_id))
+                            .delegate_staking(*delegation_id, Amount::from_atoms(*reward))
+                            .map_err(ConnectTransactionError::PoSAccountingError)
+                    })
+                    .collect::<Result<Vec<_>, _>>()?;
+
+                Ok(distribute_remainder_undo
+                    .unwrap_or_default()
```
TheQuantumPhysicist, line 133: This should probably go... unclear why this distinction between empty vector and None exists.

TheQuantumPhysicist:
Another round done. Things are looking better and better, closer to the final version. Thank you. Please consider my comments and continue.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: done #868

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: These are different execution paths: I give delegations reward to the owner if there are no delegators or their total balance is 0

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: I changed this place, but others are trickier indeed: couldn't make it work in a relatively short time, how bad do we want it?

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: It would just perform useless operation and produce useless undo. Maybe we should ignore 0 cases inside `pos_accounting`

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: That's OK. Don't worry about it.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: Isn't this how it has to go anyway if there are no delegators?

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: Then please make it with a subset/superset relationship. Not different branches.

chainstate/src/interface/chainstate_interface_impl.rs
```diff
@@ -560,17 +560,18 @@ fn get_output_coin_amount(
         }
         TxOutput::StakePool(data) => Some(data.value()),
         TxOutput::ProduceBlockFromStake(_, pool_id) => {
```
TheQuantumPhysicist, line 562: This isn't specifically for this PR, but this whole function (and the one calling it) seems to be playing with fire. There's no "output amount" for a coin. This is dangerous. If we want to allow such behavior, why didn't we allow something like `value()` in TxOutput?

common/src/primitives/per_thousand.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[case(Seed::from_entropy())]
+    fn test_encode_decode(#[case] seed: Seed) {
+        let mut rng = test_utils::random::make_seedable_rng(seed);
+
+        let encoded_valid = PerThousand::new(rng.gen_range(0..=1000)).unwrap().encode();
+        PerThousand::decode(&mut encoded_valid.as_slice()).unwrap();
+
+        let encoded_invalid = rng.gen_range(1001..=u16::MAX).encode();
+        PerThousand::decode(&mut encoded_invalid.as_slice()).unwrap_err();
+    }
```
TheQuantumPhysicist, line 93: Also include select values tests, to ensure that a change in encoding format will cause an error.

mempool/src/error/ban_score.rs
```diff
@@ -159,13 +159,20 @@ impl MempoolBanScore for ConnectTransactionError {
...
-            ConnectTransactionError::PoolBalanceNotFound(_) => 0,
+            ConnectTransactionError::PoolOwnerBalanceNotFound(_) => 0,
             ConnectTransactionError::PoolDataNotFound(_) => 0,
             ConnectTransactionError::StorageError(_) => 0,
             ConnectTransactionError::UndoFetchFailure => 0,
             ConnectTransactionError::TxVerifierStorage => 0,
             ConnectTransactionError::MissingTxUndo(_) => 0,
             ConnectTransactionError::MissingMempoolTxsUndo => 0,
+            ConnectTransactionError::PoolOwnerRewardCalculationFailed(_, _) => 0,
+            ConnectTransactionError::PoolOwnerRewardCannotExceedTotalReward(_, _, _, _) => 100,
```
TheQuantumPhysicist, line 170: This seems to ignore the title up there, in the comment

pos_accounting/src/pool/operations.rs
```diff
@@ -92,7 +92,7 @@ pub trait PoSAccountingOperations {
 
     fn decommission_pool(&mut self, pool_id: PoolId) -> Result<PoSAccountingUndo, Error>;
 
-    fn increase_pool_balance(
+    fn increase_pool_pledge_amount(
```
TheQuantumPhysicist, line 95: I'm confused about this. The pledge amount and the pool balance are two different things. When I said add the staker balance to the pool data, I meant as another field. We need to do this distinction. On one hand, if we have no total balance of the pool, how are we gonna determine the reward? By looping over all shares? That's O(N). Bad. If we don't have the owner balance, then we can't know how much to pay him on decomission.

chainstate/test-suite/src/tests/helpers/pos.rs
TheQuantumPhysicist: In that case, please call the function "create_stake_pool_data_with_all_reward_to_owner".

chainstate/tx-verifier/Cargo.toml
```diff
@@ -17,6 +17,7 @@ pos_accounting = {path = '../../pos_accounting'}
 utils = {path = '../../utils'}
 utxo = {path = '../../utxo'}
 
+static_assertions.workspace = true
```
TheQuantumPhysicist, line 20: This is nitpicky, but we're trying to follow a specfic way here.

workspace elements alone, versions alone, and crates from ".." alone. Each of those groups should be sorted by name.

pos_accounting/src/pool/operations.rs
TheQuantumPhysicist: Let's have a call about this tomorrow.

azarovh: pool_balance is a separate field no changes here. It is a sum of pledge_amount and all delegation balances. Every time `increase_pool_pledge_amount` is called it changes `pledge_amount` inside pool_data and also updates pool_balance.

So right now getting total pool balance and pledged amount should be O(logN) - at least for in_memory with BTreeMap. At the same time getting total delegations balance is O(N) because we need to sum them up every time.

chainstate/src/interface/chainstate_interface_impl.rs
azarovh: Agree, this function has to go. It's only used in mempool tests and was introduced to calculate fees, but is obsolete now.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: hmm, no. We subtract a fixed fee and a margin. But if MPT is less than 1000 the rest would be lost. 

That's how it is with the current approach. Alternatively, we could replace MPT with a reward proportional to the pledge amount, like it is with delegators.  Then it would be as you say.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: Do you mean like allow actually calling `increase_pool_pledge_amount(0)` with no-op as an undo? In other work hide this behavior inside `pos_accounting`?

TheQuantumPhysicist: OK. So we have no Option anymore. All good.

pos_accounting/src/pool/operations.rs
TheQuantumPhysicist: As discussed, we'll have tests that ensure that the balance distribution is correct.

The misunderstanding here is that we increase the individual balances (for delegators), including the pledge.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: As discussed I'll refactor the flow of the distribution to avoid else clause and make it more straightforward, effectively trying to call `increase_pledge_amount` only once

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,637 @@
...
+    Ok((delegation_undos, delegations_reward_remainder))
+}
+
+fn calculate_rewards_per_delegation(
+    delegation_shares: &BTreeMap<DelegationId, Amount>,
+    pool_id: PoolId,
+    total_delegations_amount: Amount,
+    total_delegations_reward_amount: Amount,
+) -> Result<Vec<(DelegationId, Amount)>, ConnectTransactionError> {
+    assert_eq_size!(Amount, Uint128);
```
TheQuantumPhysicist, line 170: Why Uint128 not u128?

Also please add a comment explaining what this assert is valuable here. Something like "this condition is necessary to ensure that the multiplication of balance and rewards won't overflow; if this is to change, please ensure that the output of the operations below has twice as many bits".

Though I would really prefer to have a runtime check for this... but apparently Uint256 is infected with C/C++ issues.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,637 @@
...
+        .collect();
+
+    Ok(AccountingBlockRewardUndo::new(undos))
+}
+
+fn calculate_pool_owner_reward(total_reward: Amount, pool_data: &PoolData) -> Option<Amount> {
+    (total_reward - pool_data.cost_per_block())
+        .and_then(|v| v * pool_data.margin_ratio_per_thousand().value().into())
+        .and_then(|v| v / 1000)
+        .and_then(|v| v + pool_data.cost_per_block())
```
TheQuantumPhysicist, line 113: Please add a check at the beginning of this function that cost_per_block < total reward. I can see this being used as an attack vector. Also some tests for this would be nice.

TheQuantumPhysicist: I'd also add a check, always, making sure the output of this function is < total reward.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,637 @@
...
+                } else {
+                    // If total balance of all delegations is 0 then give the reward to the pool's owner
+                    (Vec::new(), total_delegations_reward)
+                }
+            }
+            // If no delegations then give the reward to the pool's owner
+            None => (Vec::new(), total_delegations_reward),
+        }
+    } else {
+        // Do nothing if no delegations
```
TheQuantumPhysicist, line 91: "no delegations rewards"

based on the if condition up there

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,637 @@
...
+        (Vec::new(), Amount::ZERO)
+    };
+
+    let total_owner_reward = (pool_owner_reward + unallocated_reward)
+        .ok_or(ConnectTransactionError::RewardAdditionError(block_id))?;
+    let increase_pool_balance_undo = accounting_adapter
+        .operations(TransactionSource::Chain(block_id))
+        .increase_pool_pledge_amount(pool_id, total_owner_reward)?;
+
```
TheQuantumPhysicist, line 100: I guess at this point we can add a check, something like pool_owner_reward + unallocated_reward + allocated reward = total reward.

I see it's challenging because we may need to keep track of the distributed rewards... if you can easily do that, please go for it. Otherwise, skip it.

Basically I'm trying to enforce as many checks as humanly possible, with tons of tests ensuring that nothing can go in this distribution. The last thing we want is someone finding a way to print money by staking.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,637 @@
...
+    let total_delegations_balance = Uint256::from_amount(total_delegations_amount);
+    let total_delegations_reward = Uint256::from_amount(total_delegations_reward_amount);
+    delegation_shares
+        .iter()
+        .map(
+            |(delegation_id, balance_amount)| -> Result<_, ConnectTransactionError> {
+                let balance = Uint256::from_amount(*balance_amount);
+                let reward = (total_delegations_reward * balance) / total_delegations_balance;
+                let reward: UnsignedIntType = reward.try_into().map_err(|_| {
```
TheQuantumPhysicist, line 184: It's better to make this `amount::UnsignedIntType` to make it readable that this is about amount.

iljakuklic (approved):
Looks good to me. Few minor comments below.

muursh (approved):
I have no issue with any of the logic in here. Looks fine to me

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: > cost_per_block < total reward

not sure that this check is required, because subtraction is already a checked operation. Or do you want it to be super explicit?

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,637 @@
...
+
+    let total_owner_reward = (pool_owner_reward + unallocated_reward)
+        .ok_or(ConnectTransactionError::RewardAdditionError(block_id))?;
+    let increase_pool_balance_undo = accounting_adapter
+        .operations(TransactionSource::Chain(block_id))
+        .increase_pool_pledge_amount(pool_id, total_owner_reward)?;
+
+    let undos = delegation_undos
+        .into_iter()
+        .chain(vec![increase_pool_balance_undo].into_iter())
```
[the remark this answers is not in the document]
azarovh: cool didn't know that

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
pavel-kokolemin: I see this has already been discussed and added as a TODO. But why not send the entire reward to the owner if `cost_per_block` is greater than `total_reward`? This seems more reasonable than rejecting such blocks.

chainstate/src/detail/ban_score.rs
```diff
@@ -357,6 +364,7 @@ impl BanScore for pos_accounting::Error {
             E::PoolBalanceSubtractionError => 100,
             E::DelegationSharesAdditionError => 100,
             E::DelegationSharesSubtractionError => 100,
+            E::InvariantErrorDelegationSharesMoreThanPoolBalance => 100,
```
pavel-kokolemin, line 367: This error is not used in the code.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -0,0 +1,637 @@
...
+    )?;
+
+    let total_delegations_reward = (total_reward - pool_owner_reward).ok_or(
+        ConnectTransactionError::PoolOwnerRewardCannotExceedTotalReward(
+            block_id,
+            pool_id,
+            pool_owner_reward,
+            total_reward,
+        ),
+    )?;
```
pavel-kokolemin, line 57: Do we really need `PoolOwnerRewardCannotExceedTotalReward` error? It can't happen, so maybe it's better to use `expect` instead?

common/src/primitives/per_thousand.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    pub fn value(&self) -> u16 {
+        self.0
+    }
+}
+
+impl Encode for PerThousand {
+    fn encode_to<T: Output + ?Sized>(&self, dest: &mut T) {
+        dest.write(&self.0.to_be_bytes());
```
pavel-kokolemin, line 37: Why `to_be_bytes` is used? I think for consistency we should use default (little-endian) encoding (`self.0.encode_to(dest)`). Decoding would be a bit easier too:
```
Self::new(u16::decode(input)?).ok_or(serialization::Error::from(
            "PerThousand deserialization failed",
        ))
```

TheQuantumPhysicist: I actually agree. In fact, I don't see a need to implement Encode at all manually.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
TheQuantumPhysicist: It's just paranoia. Getting an error that we can handle is better than crashing. Things can go wrong.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: you are right it's not that straightforward. Suggest skipping

common/src/primitives/per_thousand.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    pub fn value(&self) -> u16 {
+        self.0
+    }
+}
+
+impl Decode for PerThousand {
+    fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
+        Self::new(u16::decode(input)?).ok_or(serialization::Error::from(
+            "PerThousand deserialization failed",
```
TheQuantumPhysicist, line 43: Since you're decoding this to u16 first, can you add the invalid value in the error?

TheQuantumPhysicist (approved):
OK. I think we're in a good shape now. Cheers!

common/src/primitives/per_thousand.rs
azarovh: It seems that I can't since `serialization::Error` has a description of type `&'static str`

TheQuantumPhysicist: I see. Never mind.

PR #855, opened 2023-04-26
mempool: Update TODOs

iljakuklic:
Update the comments to reflect what we learned about peer punishment. In Bitcoin, peers are not punished for sending invalid transactions if the transaction validity depends on the current chain state.

### DIFF
--- a/mempool/src/error/ban_score.rs
+++ b/mempool/src/error/ban_score.rs
@@ -257,6 +257,9 @@ impl MempoolBanScore for pos_accounting::Error {
             E::DelegateToNonexistingId => 0,
             E::DelegateToNonexistingPool => 0,
 
+            // Accounting error has to be inspected further
+            E::AccountingError(err) => err.mempool_ban_score(),
+
             // Internal invariant errors
             E::InvariantErrorPoolBalanceAlreadyExists => 0,
             E::InvariantErrorPoolDataAlreadyExists => 0,
@@ -267,7 +270,6 @@ impl MempoolBanScore for pos_accounting::Error {
             E::InvariantErrorDelegationShareNotFound => 0,
 
             // These signify an invalid transaction
-            E::AccountingError(_) => 100,
             E::AdditionError => 100,
             E::SubError => 100,
             E::DelegationBalanceAdditionError => 100,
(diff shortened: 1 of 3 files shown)


mempool/src/error/ban_score.rs
```diff
@@ -254,16 +251,16 @@ impl MempoolBanScore for pos_accounting::Error {
...
-            // TODO: Can we ban the peer in these cases anyway? Some will be only triggered
-            // after an epoch has elapsed which should be enough time for peer to be up-to-date
-            E::InvariantErrorPoolBalanceAlreadyExists => 0,
-            E::InvariantErrorPoolDataAlreadyExists => 0,
-            E::InvariantErrorDelegationCreationFailedIdAlreadyExists => 0,
             E::AttemptedDecommissionNonexistingPoolBalance => 0,
             E::AttemptedDecommissionNonexistingPoolData => 0,
             E::DelegationCreationFailedPoolDoesNotExist => 0,
             E::DelegateToNonexistingId => 0,
             E::DelegateToNonexistingPool => 0,
```
iljakuklic, line 258: @azarovh Could you check these are all indeed dependent on the current state? Or could we use any of these to ban the peer?

azarovh: I believe these are correct. 

One thing is that `E::AccountingError(_) => 100` shouldn't be generalized. I think there are some errors inside that depend on a state like `DataCreatedMultipleTimes` or `ModifyNonexistingData` and so on. They are most likely overlapped by higher level `pos_accounting` errors but still.

azarovh (approved):

TheQuantumPhysicist (approved):

muursh (approved):

mempool/src/error/ban_score.rs
iljakuklic: > One thing is that `E::AccountingError(_) => 100` shouldn't be generalized. I think there are some errors inside that depend on a state like `DataCreatedMultipleTimes` or `ModifyNonexistingData` and so on. They are most likely overlapped by higher level `pos_accounting` errors but still.

Added `BanScore` for `accounting::Error` (see commit 6334000c50fa7407a08ac3c5163d381225e8095f). Can you take a look whether it seems sensible?

mempool/src/error/ban_score.rs
```diff
@@ -298,3 +297,29 @@ impl MempoolBanScore for pos_accounting::Error {
...
+    fn mempool_ban_score(&self) -> u32 {
+        match self {
+            // These should not happen with valid transactions
+            accounting::Error::ArithmeticErrorToUnsignedFailed => 100,
+            accounting::Error::ArithmeticErrorToSignedFailed => 100,
+            accounting::Error::ArithmeticErrorSumToUnsignedFailed => 100,
+            accounting::Error::ArithmeticErrorSumToSignedFailed => 100,
+
+            // These depend on the current state which may be out of sync
+            accounting::Error::ArithmeticErrorDeltaAdditionFailed => 0,
```
azarovh, line 311: I'd make it 100 as well so no one tries to overflow

iljakuklic: done

stanislav-tkach (approved):

azarovh (approved):

PR #856, opened 2023-04-27
Fix GUI shutdown

pavel-kokolemin:
There is exactly the same problem that @stanislav-tkach had with p2p tests shutdown. Making one more `ShutdownTrigger` prevents proper shutdown detection. As a quick fix I added `WeakShutdownTrigger` which uses `tokio::sync::mpsc::WeakSender` instead of `tokio::sync::mpsc::Sender`.

### DIFF
--- a/subsystem/src/manager.rs
+++ b/subsystem/src/manager.rs
@@ -370,15 +370,14 @@ impl ShutdownTrigger {
     /// Initiate shutdown
     pub fn initiate(&self) {
         use mpsc::error::TrySendError as E;
-        if let Some(sender) = self.0.upgrade() {
-            sender.try_send(()).unwrap_or_else(|err| match err {
-                E::Full(_) => {
-                    log::info!("Shutdown requested but the system is already shutting down")
-                }
-                E::Closed(_) => log::info!("Shutdown requested but the system is already down"),
-            })
-        } else {
-            log::info!("Shutdown requested but the system is already down")
+        match self.0.upgrade().map(|s| s.try_send(())) {
+            None | Some(Err(E::Closed(_))) => {
+                log::info!("Shutdown requested but the system is already down")
+            }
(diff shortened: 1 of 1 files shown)


iljakuklic:
Would it be sensible to change the normal shutdown trigger to be the weak version instead of having two variants? I don't see why this should not be the default. Using the strong one, as we've seen, is rather error prone and really blocking the shutdown until shutdown triggers are dropped seems pretty useless.

pavel-kokolemin:
> Would it be sensible to change the normal shutdown trigger to be the weak version instead of having two variants? 

Good idea. Updated, thanks.

subsystem/src/manager.rs
```diff
@@ -364,16 +364,22 @@ impl Manager {
...
+        if let Some(sender) = self.0.upgrade() {
+            sender.try_send(()).unwrap_or_else(|err| match err {
+                E::Full(_) => {
+                    log::info!("Shutdown requested but the system is already shutting down")
+                }
+                E::Closed(_) => log::info!("Shutdown requested but the system is already down"),
+            })
+        } else {
+            log::info!("Shutdown requested but the system is already down")
+        }
```
stanislav-tkach, line 382: I'm not sure this is better, but it allows not to repeat the message:
```suggestion
match self.0.upgrade().map(|s| s.try_send(())) {
            None | Some(Err(E::Closed(_))) => {
                log::info!("Shutdown requested but the system is already down")
            }
            Some(Err(E::Full(_))) => {
                log::info!("Shutdown requested but the system is already shutting down")
            }
            Some(Ok(())) => {}
        }
```

stanislav-tkach (approved):
👍

iljakuklic (approved):

TheQuantumPhysicist (approved):

PR #857, opened 2023-04-27
Fix logs printing in wallet-cli

pavel-kokolemin:
Use `reedline::ExternalPrinter` to print log output without mangling reedline console input.


pavel-kokolemin:
`wallet-cli` will need some updates. I will include this changes as a part of other PR.

PR #861, opened 2023-04-28
Add Proof-of-Stake block production

alfiedotwtf:


consensus/src/pow/work.rs
```diff
@@ -106,15 +106,15 @@ impl MiningResult {
...
 }
 
-pub fn calculate_work_required<G>(
+pub fn calculate_work_required<F>(
     chain_config: &ChainConfig,
     prev_block_index: &GenBlockIndex,
     block_timestamp: BlockTimestamp,
     pow_status: &PoWStatus,
-    get_ancestor: G,
+    get_ancestor: F,
```
TheQuantumPhysicist, line 114: Is there a reason you're naming this from F to G? Btw it's G for Get or Getter.

alfiedotwtf: I faintly remember there were originally two functions - F and G, and F was removed. If not for this function, for another one, and so I was just "resetting the counter" so to speak. But you're rith - G for Getter works too, so I'll leave as is and drop that commit.

blockprod/src/detail/mod.rs
```diff
@@ -141,14 +143,22 @@ impl BlockProduction {
                         })
                     };
 
+                    let block_height = best_block_index.block_height().next_height();
+                    let sealed_epoch_randomness = this.get_epoch_data(block_height);
+                    // TODO this doesn't exist, thinking of adding to chainstate trait
```
alfiedotwtf, line 148: This is what I was referring to, about adding to the chainstate trait (read only)

consensus/src/lib.rs
```diff
@@ -50,11 +63,101 @@ pub enum ConsensusCreationError {
...
+    PoW,
+    PoS(PoSFinalizeBlockInputData),
+}
+
+#[derive(Debug, Clone, Encode, Decode)]
+pub struct PoSFinalizeBlockInputData {
+    vrf_private_key: VRFPrivateKey,
+    epoch_index: EpochIndex,
+    sealed_epoch_randomness: PoSRandomness,
+    pool_balance: Amount,
```
TheQuantumPhysicist, line 117: Why not pull the balance from the database, from accounting?

alfiedotwtf: Yep, that's what I'll be doing. I just left this as one of the last TODO items rather than being any issue

consensus/src/pos/target.rs
```diff
@@ -92,13 +94,41 @@ fn calculate_new_target(
...
+pub fn calculate_target_required_from_block_index<F>(
+    chain_config: &ChainConfig,
+    pos_status: &PoSStatus,
+    prev_gen_block_index: &GenBlockIndex,
+    get_ancestor: F,
+) -> Result<Compact, ConsensusPoSError>
+where
+    F: Fn(&BlockIndex, BlockHeight) -> Result<GenBlockIndex, PropertyQueryError>,
+{
+    let pos_config = match pos_status {
```
azarovh, line 106: seems like pos_status can be checked in internal function

azarovh: nevermind, pos_config is required

consensus/src/lib.rs
```diff
@@ -50,11 +66,113 @@ pub enum ConsensusCreationError {
...
+    PoS(Box<PoSGenerateBlockInputData>),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub struct PoSGenerateBlockInputData {
+    stake_private_key: PrivateKey,
+    vrf_private_key: VRFPrivateKey,
+    pool_id: PoolId,
+    kernel_input: TxInput,
+    kernel_witness: InputWitness,
```
TheQuantumPhysicist, line 89: Having a witness here doesn't make sense. Isn't the private key here to produce that witness?

TheQuantumPhysicist:
Why not write some functional tests for this?

consensus/src/pos/mod.rs
```diff
@@ -175,3 +194,60 @@ where
...
+    finalize_pos_data: PoSFinalizeBlockInputData,
+    stop_flag: Arc<AtomicBool>,
+) -> Result<StakeResult, ConsensusPoSError> {
+    let sealed_epoch_randomness = finalize_pos_data.sealed_epoch_randomness();
+    let vrf_sk = finalize_pos_data.vrf_private_key();
+    let vrf_pk = finalize_pos_data.vrf_public_key();
+
+    let mut current_timestamp = block_header.timestamp();
+
+    let end_timestamp = match current_timestamp.add_int_seconds(60) {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: No. It should use the max_future_block_time_offset from ChainConfig. The only factor that affects time is its difference compared to previous block.

So the range should be from previous block up to current time + max_future_block_time_offset.

consensus/src/pow/work.rs
```diff
@@ -255,7 +254,7 @@ pub fn mine(
             return Ok(MiningResult::Success);
         }
 
-        if stop_flag.load(std::sync::atomic::Ordering::Relaxed) {
+        if stop_flag.load(std::sync::atomic::Ordering::SeqCst) {
```
TheQuantumPhysicist, line 257: This can be relaxed. In fact, this is an excellent example for using relaxed memory order. The question is: What's the worst thing that will happen if this is relaxed? The answer is: It'll be delayed a few cycles and stop eventually? Who cares.

consensus/Cargo.toml
```diff
@@ -14,6 +14,7 @@ serialization = { path = "../serialization" }
 utils = { path = "../utils" }
 utxo = { path = '../utxo' }
 
+parity-scale-codec.workspace = true
```
TheQuantumPhysicist, line 17: Normally you shouldn't have to add this. Just adding the serialization crate should be enough.

test/functional/feature_lmdb_backend_test.py
```diff
@@ -40,10 +40,10 @@ def run_test(self):
         assert_equal(self.block_height(), 0)
```
TheQuantumPhysicist, line 41: So I'm writing this comment again here so that it's not lost.

Please try to write functional tests for PoS staking. There are primitives that can be generated (VRF keys, private keys, etc), for which some RPC functions have to be created. In case you're stuck, let's talk about this.

consensus/src/lib.rs
```diff
@@ -50,11 +66,113 @@ pub enum ConsensusCreationError {
...
+    #[error("Staking error")]
+    StakingError(#[from] ConsensusPoSError),
+    #[error("Staking failed")]
+    StakingFailed,
+    #[error("Staking stopped")]
+    StakingStopped,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub enum GenerateBlockInputData {
```
TheQuantumPhysicist, line 78: Please add Encode/Decode indexes for enum arms

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -325,6 +325,13 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         Ok(self.db_tx.get_block_reward(block_index).log_err()?)
     }
 
+    pub fn get_epoch_data(
+        &self,
+        epoch_index: u64,
```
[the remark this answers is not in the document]
alfiedotwtf: I forgot about that, thanks. Done.

chainstate/types/src/error.rs
```diff
@@ -30,6 +30,8 @@ pub enum PropertyQueryError {
     BestBlockIndexNotFound,
     #[error("Block not found {0}")]
     BlockNotFound(Id<Block>),
+    #[error("Epoch data not found")]
+    EpochDataNotFound,
```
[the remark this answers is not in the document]
alfiedotwtf: I've extended this out to `block_height` so we can see the original value

consensus/src/pos/mod.rs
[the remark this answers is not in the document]
alfiedotwtf: I wasn't too sure about the logic on this one. Thanks, done.

consensus/src/pos/target.rs
```diff
@@ -92,13 +94,41 @@ fn calculate_new_target(
...
+}
+
+pub fn calculate_target_required<F>(
     chain_config: &ChainConfig,
     pos_status: &PoSStatus,
     prev_block_id: Id<GenBlock>,
     block_index_handle: &impl BlockIndexHandle,
-) -> Result<Compact, ConsensusPoSError> {
-    // check if current block is a net upgrade threshold
+    get_ancestor: F,
```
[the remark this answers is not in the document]
alfiedotwtf: Ah nice. I've removed it from the API when we have a `BlockIndexHandle`

consensus/src/lib.rs
alfiedotwtf: Done.

I wonder if in the future we could add something to `do_checks` that detects codec enums and warns when they don't have an index?

consensus/Cargo.toml
alfiedotwtf: I thought so too, but if removed then it throws errors during compilation.

blockprod/src/detail/mod.rs
```diff
@@ -141,14 +149,70 @@ impl BlockProduction {
                         })
                     };
 
+                    let block_height = best_block_index.block_height().next_height();
+                    let sealed_epoch_index = chain_config.sealed_epoch_index(&block_height);
+
+                    let sealed_epoch_randomness = sealed_epoch_index
+                        .map(|index| this.get_epoch_data(index))
```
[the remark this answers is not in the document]
alfiedotwtf: `sealed_epoch_index` can still be `None` so that currently won't work.

blockprod/src/detail/mod.rs
```diff
@@ -141,14 +149,70 @@ impl BlockProduction {
...
+                        sealed_epoch_randomness.clone(),
+                        input_data.clone(),
                         block_timestamp,
-                        best_block_index.block_height().next_height(),
+                        block_height,
                         get_ancestor,
                     );
-                    consensus_data.map(|cons_data| (cons_data, best_block_index))
+
+                    let finalize_block_data = match input_data {
```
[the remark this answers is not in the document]
alfiedotwtf: I'm not sure this is a problem, as the `match input_data` will only evaluate the larger block when we're PoS.

blockprod/src/detail/mod.rs
```diff
@@ -141,14 +149,70 @@ impl BlockProduction {
...
+                    let sealed_epoch_randomness = sealed_epoch_index
+                        .map(|index| this.get_epoch_data(index))
+                        .transpose()
+                        .map_err(|_| {
+                            ConsensusPoSError::PropertyQueryError(
+                                PropertyQueryError::EpochDataNotFound,
+                            )
+                        })?
+                        .flatten()
+                        .or_else(|| Some(EpochData::new(PoSRandomness::at_genesis(&chain_config))));
```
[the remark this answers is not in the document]
azarovh: I'd place a todo here, we have the same one in `consensus/src/pos/mod.rs`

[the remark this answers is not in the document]
alfiedotwtf: I think there's a few places where `get_epoch_data()` is called, and if no data is found then we default to `PoSRandomness::at_genesis()` rather than error. Maybe this has changed, or it's time to change this in all occurrences?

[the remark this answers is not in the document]
azarovh: yes, we default to genesis but this is only ok unless we seal an epoch, after that it should be an error. Right now there are todos for that all over the code so I think it should be placed here as well.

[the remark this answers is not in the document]
alfiedotwtf: TODO added.

consensus/src/lib.rs
alfiedotwtf: I've removed `kernel_witness` and have updated all of the kernel signing code.

consensus/src/lib.rs
```diff
@@ -50,11 +68,133 @@ pub enum ConsensusCreationError {
...
+        PublicKey::from_private_key(&self.stake_private_key)
+    }
+
+    pub fn vrf_private_key(&self) -> &VRFPrivateKey {
+        &self.vrf_private_key
+    }
+}
+
+#[derive(Debug, Clone, Encode, Decode)]
+pub enum FinalizeBlockInputData {
```
TheQuantumPhysicist, line 123: Please add documentation explaining what this struct is used for. Also the one below. There's now a few structs and they can be confusing for readers.

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -131,6 +132,11 @@ pub trait ChainstateInterface: Send {
         &self,
         block_index: &BlockIndex,
     ) -> Result<Option<BlockReward>, ChainstateError>;
+
+    // Returns epoch data for the given epoch index.
```
TheQuantumPhysicist, line 136: Three slashes for documentation of functions

blockprod/src/detail/mod.rs
```diff
@@ -180,15 +278,17 @@ impl BlockProduction {
...
     ) -> Result<(Block, oneshot::Receiver<usize>), BlockProductionError> {
-        self.produce_block_with_custom_id(reward_destination, transactions_source, None)
+        self.produce_block_with_custom_id(input_data, reward_destination, transactions_source, None)
             .await
     }
 
     async fn produce_block_with_custom_id(
         &self,
+        input_data: Option<GenerateBlockInputData>,
         _reward_destination: Destination,
```
azarovh, line 292: should `reward_destination` be used in PoW?

blockprod/src/detail/mod.rs
```diff
@@ -141,14 +149,102 @@ impl BlockProduction {
...
+                        // TODO: no epoch_data means either that no
+                        // epoch has been created yet, or that the
+                        // data is actually missing
+                        .or_else(|| Some(EpochData::new(PoSRandomness::at_genesis(&chain_config))));
+
                     let consensus_data = consensus::generate_consensus_data(
                         &chain_config,
                         &best_block_index,
+                        sealed_epoch_randomness.clone(),
+                        input_data.clone(),
```
azarovh, line 173: can this clone be avoided if calculate `finalize_block_data` before?

blockprod/src/detail/mod.rs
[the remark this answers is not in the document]
azarovh: agree, maybe move it to a separate function to make it more readable?

consensus/src/pos/error.rs
```diff
@@ -30,6 +30,8 @@ pub enum ConsensusPoSError {
     PropertyQueryError(#[from] chainstate_types::PropertyQueryError),
     #[error("Stake kernel hash failed to meet the target requirement")]
     StakeKernelHashTooHigh,
+    #[error("Epoch data not provided")]
+    NoEpochData,
```
azarovh, line 34: provide epoch index

consensus/src/lib.rs
```diff
@@ -50,11 +68,133 @@ pub enum ConsensusCreationError {
...
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub enum GenerateBlockInputData {
+    #[codec(index = 0)]
+    PoW,
+    #[codec(index = 1)]
+    PoS(Box<PoSGenerateBlockInputData>),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub struct PoSGenerateBlockInputData {
```
azarovh, line 88: I think it's better to move all PoS related types and functionality somewhere into `pos` submodule to make lib file clean

consensus/src/lib.rs
```diff
@@ -64,7 +204,73 @@ where
...
+
+            let target_required = calculate_target_required_from_block_index(
+                chain_config,
+                &pos_status,
+                prev_block_index,
+                get_ancestor,
+            )?;
+
+            Ok(ConsensusData::PoS(Box::new(PoSData::new(
+                pos_input_data.kernel_inputs().clone(),
```
azarovh, line 267: can you calculate pool_id beforehand and avoid cloning kernel?

azarovh (commented):
couldn't finish it in one set, will continue tomorrow

TheQuantumPhysicist (approved):
I've been thinking about whether we want to merge this, and I think we should, because we want the wallet to be prepared for staking.

So, I'm approving this. Please fix the issues at hand then feel free to merge. Based on that, we'll get the wallet, at least, to start doing something useful with staking.

test-rpc-functions/src/rpc.rs
```diff
@@ -0,0 +1,207 @@
...
+        )
+        .map_err(RpcTestFunctionsError::from);
+
+        let vrf_output = handle_error(vrf_output)?;
+
+        Ok(vrf_output.hex_encode())
+    }
+}
+
+// fn handle_call_error<T>(e: Result<Result<T, RpcTestFunctionsError>, CallError>) -> rpc::Result<T> {
```
azarovh, line 201: commented code

azarovh (approved):

consensus/src/lib.rs
alfiedotwtf: No problem. Done.

chainstate/src/interface/chainstate_interface.rs
alfiedotwtf: Fixed too.

blockprod/src/detail/mod.rs
alfiedotwtf: I tried moving things around, but unfortunately it won't work because copies of both `sealed_epoch_randomnness` and `input_data` are needed in more than one place (unless I'm mistaken).

consensus/src/pos/error.rs
alfiedotwtf: This error here means that epoch data wasn't provided by the caller of the function (as it's an `Option`), and not that epoch data could not be found, so I don't think a height or index here will help.

consensus/src/lib.rs
alfiedotwtf: Nice. I've also shuffled a bit of code around into a separate function to make things a bit clearer.

consensus/src/lib.rs
alfiedotwtf: Here, `pool_id` is just an accessor so there's no calculations being done. As for `kernel_inputs`, that accessor returns a ref which needs to be passed on - so it will either need to be `clone`d or we can keep in inside an `Option` and then `take()` it as a last step, but then that adds more steps each time it's accessed to see if it's empty.

test-rpc-functions/src/rpc.rs
alfiedotwtf: Fixed.

alfiedotwtf: Cool, thanks. Once @azarovh finishes his review (and of the updated fixes), I'll merge and push tests to a new PR.

alfiedotwtf:
Cool, thanks. Once @azarovh finishes his review (and of the updated fixes), I'll merge and push tests to a new PR.

blockprod/src/detail/mod.rs
alfiedotwtf: Yep, this is what I wasn't too sure of. Currently there's a `Destination` parameter in the RPC call, but was thinking this may not be needed for PoS now, so it could be pushed into the `PoW` input data... I'll work out the mechanics as I make tests

consensus/src/lib.rs
```diff
@@ -79,23 +132,145 @@ where
     }
 }
 
+#[allow(clippy::too_many_arguments)]
+fn generate_pos_consensus_data<G>(
```
azarovh, line 136: this function should go to pos as well

consensus/src/pos/input_data.rs
```diff
@@ -0,0 +1,151 @@
...
+        PublicKey::from_private_key(&self.stake_private_key)
+    }
+
+    pub fn vrf_private_key(&self) -> &VRFPrivateKey {
+        &self.vrf_private_key
+    }
+}
+
+#[derive(Debug, Clone, Encode, Decode)]
+/// Input needed to finalize PoS consensus data
```
TheQuantumPhysicist, line 81: Please put the comments above derive macros.

Also, please explain with more information. Who provides this information and in what circumstance (same for the other struct).

consensus/src/lib.rs
```diff
@@ -50,11 +68,33 @@ pub enum ConsensusCreationError {
...
+#[derive(Debug, Clone, Encode, Decode)]
+pub enum FinalizeBlockInputData {
+    PoW,
+    PoS(PoSFinalizeBlockInputData),
 }
 
 pub fn generate_consensus_data<G>(
     chain_config: &ChainConfig,
     prev_block_index: &GenBlockIndex,
+    sealed_epoch_randomness: Option<EpochData>,
```
azarovh, line 96: can randomness be part of InputData where all consensus specific data is stored? It's slightly confusing to see it in this generic function

blockprod/src/detail/mod.rs
```diff
@@ -141,14 +149,102 @@ impl BlockProduction {
...
                         block_timestamp,
-                        best_block_index.block_height().next_height(),
+                        block_height,
                         get_ancestor,
-                    );
-                    consensus_data.map(|cons_data| (cons_data, best_block_index))
+                    )?;
+
+                    let finalize_block_data = match input_data {
```
azarovh, line 179: I think this match deserves to be a function

azarovh (approved):
looks good

consensus/src/pos/input_data.rs
alfiedotwtf: I've expanded here, which should hopefully be what wallet RPC devs to know what to pass in. Let me know if any section needs further detail

consensus/src/lib.rs
alfiedotwtf: Nice. Done.

blockprod/src/detail/mod.rs
alfiedotwtf: Yeah, that's one of the places I wanted to clean up later, but while I'm in this code, I might as well add it.

I originally moved it into `/consensus/` but when fixing up the imports I realised that `chainstate` would need to be imported for `ChainstateInterface`, which would up as a circular dependancy... so instead I moved it out of that function to a helper fuction below it.

consensus/src/pos/error.rs
alfiedotwtf: The reason why it's optional is because it's passed to the general `generate_finalize_block_data()` function, and then only looks at it epoch data if we're wanting PoS.

It's as if we need two types of parameter enums - one for user input over the wire, and another for internal data. This can be done, but I think that may be a bigger change, which we can probably put off for another PR?

consensus/src/lib.rs
azarovh: my concern is that randomness is only valid for PoS and should be somehow encapsulated. Maybe we can get back to this later I won't insist. 

Btw name is randomness but the type is EpochData.

consensus/src/pos/error.rs
azarovh: ok

PR #863, opened 2023-04-28
Shutdown subsystems in the reverse creation order

stanislav-tkach:


subsystem/src/manager.rs
```diff
@@ -29,13 +29,13 @@ pub struct ManagerConfig {
...
     /// Shutdown timeout. Set to `None` for no (i.e. unlimited) timeout.
-    shutdown_timeout: Option<Duration>,
+    shutdown_timeout_per_subsystem: Option<Duration>,
 }
 
 impl ManagerConfig {
     /// Default shutdown timeout.
     const DEFAULT_SHUTDOWN_TIMEOUT: Option<Duration> = if cfg!(all(feature = "time", not(loom))) {
-        Some(Duration::from_secs(20))
+        Some(Duration::from_secs(5))
```
TheQuantumPhysicist, line 38: I would make this 30 seconds or even more. Otherwise if there's an issue we won't notice it. This should never happen!

TheQuantumPhysicist (approved):
Looking good! I don't see any obvious issue. Please wait for others to review.

TheQuantumPhysicist:
The sybsystem panic test froze for me, btw. Something is wrong, apparently.

TheQuantumPhysicist (approved):

subsystem/src/manager.rs
```diff
@@ -308,60 +282,84 @@ impl Manager {
...
         log::info!("Manager {} shutting down", self.name);
+        // Drop the receiver in order to prevent blocking of subsystems.
+        drop(self.shutting_down_rx);
 
-        // Order all the remaining subsystems to shut down.
-        let _ = self.shutdown_request_tx.send(());
+        // Shut down the subsystems in the reverse order of creation.
+        for (name, handle, shutdown_tx) in subsystems.into_iter().rev() {
+            if let Err(()) = shutdown_tx.send(()) {
+                log::warn!("Manager {}: {name} subsystem is already down", self.name);
```
alfiedotwtf, line 305: Should this `continue` rather than fall through as this subsystem has already shutdown?

alfiedotwtf (approved):

pavel-kokolemin (approved):

iljakuklic (approved):
Looks good.

One thing I wonder about is why did you choose to have a timeout per-subsystem? It means the total time limit for shutdown is dependent on the number of subsystems we have. I think one timeout for the whole process makes more sense

stanislav-tkach:
@iljakuklic I see your point, but I still think that timeout per subsystem has its own benefits. For example, with a single shutdown timeout the first subsystem can waste all of the time. Also it is easier to implement. 🙃

subsystem/src/manager.rs
stanislav-tkach: I thought about that, but joining the handle allows to print a potential error message.

TheQuantumPhysicist:
If everyone is happy with this PR and no other changes will be done, let's get it merged. No need to wait much longer.

PR #864, opened 2023-04-29
mempool: Transactions disconnected during reorg to be put back into mempool

iljakuklic:
Transactions that are disconnected during a reorg and are not present in the new fork are flushed back into the mempool so they have a chance to be added back into the chain later.

TODO: Need to properly wire up chainstate event subscription which apparently has not yet been done.

### DIFF
--- a/common/src/chain/block/block_v1.rs
+++ b/common/src/chain/block/block_v1.rs
@@ -89,8 +89,12 @@ impl BlockV1 {
         self.header.timestamp()
     }
 
-    pub fn transactions(&self) -> &Vec<SignedTransaction> {
+    pub fn transactions(&self) -> &[SignedTransaction] {
         &self.body.transactions
+    }
+
+    pub fn into_transactions(self) -> Vec<SignedTransaction> {
+        self.body.transactions
     }
 
     pub fn prev_block_id(&self) -> &Id<crate::chain::GenBlock> {
--- a/mempool/src/pool/reorg.rs
+++ b/mempool/src/pool/reorg.rs
@@ -92,7 +92,7 @@ fn find_disconnected_txs(
         return Ok(Vec::new());
     }
 
-    // Collect IDs of trancsactions included in the newly connected chain
+    // Collect IDs of transactions included in the newly connected chain
     let mut connected_txs = BTreeSet::new();
     for_each_block(chainstate, new_tip_id, common_id, |block| {
         connected_txs.extend(block.transactions().iter().map(|t| t.transaction().get_id()))
@@ -102,9 +102,9 @@ fn find_disconnected_txs(
     let mut disconnected_txs = Vec::new();
     for_each_block(chainstate, old_tip_id, common_id, |block| {
         // We iterate blocks in the reverse order, so for consistent transaction input/output
-        // dependencies, take transacitons in reverse too,
+        // dependencies, take transactions in reverse too,
         let txns = block.transactions().iter().rev();
-        // We disregerd transactions that are re-added on the newly connected chain
+        // We disregard transactions that are re-added on the newly connected chain
         let txns = txns.filter(|txn| !connected_txs.contains(&txn.transaction().get_id()));
         disconnected_txs.extend(txns.cloned())
     })?;
--- a/mempool/src/pool/reorg.rs
+++ b/mempool/src/pool/reorg.rs
@@ -46,7 +46,7 @@ pub enum ReorgError {
 }
 
 /// Collect blocks between the given two points
-fn collect_blocks<C: ChainstateInterface + ?Sized>(
+fn collect_blocks<C: ChainstateInterface>(
     chainstate: &C,
     mut curr_id: Id<GenBlock>,
     stop_id: Id<GenBlock>,
@@ -68,14 +68,15 @@ fn collect_blocks<C: ChainstateInterface + ?Sized>(
 }
 
 /// Blocks affected by a reorg
-struct ReorgInfo {
+struct ReorgData {
+    // List of connected / disconnected blocks, both in reverse chronological order
     disconnected: Vec<Block>,
     connected: Vec<Block>,
 }
 
-impl ReorgInfo {
+impl ReorgData {
     /// Extract blocks that have been disconnected and connected from the chainstate.
-    fn from_chainstate<C: ChainstateInterface + ?Sized>(
+    fn from_chainstate<C: ChainstateInterface>(
         chainstate: &C,
         old_tip_id: Id<GenBlock>,
         new_tip_id: Id<GenBlock>,
@@ -119,8 +120,8 @@ fn fetch_disconnected_txs<M>(
     let old_tip = mempool.tx_verifier.get_best_block_for_utxos().map_err(|_| ReorgError::OldTip)?;
     mempool
         .blocking_chainstate_handle()
-        .call(move |c| ReorgInfo::from_chainstate(c.as_ref(), old_tip, new_tip.into()))?
-        .map(ReorgInfo::into_disconnected_transactions)
+        .call(move |c| ReorgData::from_chainstate(c, old_tip, new_tip.into()))?
+        .map(ReorgData::into_disconnected_transactions)
 }
 
 pub fn handle_new_tip<M: GetMemoryUsage>(mempool: &mut Mempool<M>, new_tip: Id<Block>) {
--- a/mempool/src/pool/reorg.rs
+++ b/mempool/src/pool/reorg.rs
@@ -1,0 +1,162 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Support for updating the mempool upon a reorg
+
+use std::{collections::BTreeSet, mem};
+
+use chainstate::chainstate_interface::ChainstateInterface;
+use common::{
+    chain::{Block, GenBlock, SignedTransaction},
+    primitives::{Id, Idable},
+};
+use logging::log;
+use utils::{shallow_clone::ShallowClone, tap_error_log::LogError};
+use utxo::UtxosStorageRead;
+
+use crate::{
+    get_memory_usage::GetMemoryUsage,
+    pool::{tx_verifier, Mempool, MempoolStore},
+};
+
+/// An error that can happen in mempool on chain reorg
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+pub enum ReorgError {
+    #[error(transparent)]
+    Chainstate(#[from] chainstate::ChainstateError),
+    #[error("Could not find the previous tip")]
+    OldTip,
+    #[error("Could not find the previous tip index")]
+    OldTipIndex,
+    #[error("Could not find the new tip index")]
+    NewTipIndex,
+    #[error("Block {0:?} not found while traversing history")]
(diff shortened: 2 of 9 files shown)


mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+    let mut connected_txs = BTreeSet::new();
+    for_each_block(chainstate, new_tip_id, common_id, |block| {
+        connected_txs.extend(block.transactions().iter().map(|t| t.transaction().get_id()))
+    })?;
+
+    // Collect txns that have been removed from the blockchain and not added on the new fork
+    let mut disconnected_txs = Vec::new();
+    for_each_block(chainstate, old_tip_id, common_id, |block| {
+        // We iterate blocks in the reverse order, so for consistent transaction input/output
+        // dependencies, take transacitons in reverse too,
```
TheQuantumPhysicist, line 105: transactions

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+        let common_index = chainstate.last_common_ancestor(&old_index.into(), &new_index.into())?;
+        common_index.block_id()
+    };
+
+    // Short circuit the expensive processing below if no blocks are being disconnected
+    if old_tip_id == common_id {
+        return Ok(Vec::new());
+    }
+
+    // Collect IDs of trancsactions included in the newly connected chain
```
TheQuantumPhysicist, line 95: transactions

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+        connected_txs.extend(block.transactions().iter().map(|t| t.transaction().get_id()))
+    })?;
+
+    // Collect txns that have been removed from the blockchain and not added on the new fork
+    let mut disconnected_txs = Vec::new();
+    for_each_block(chainstate, old_tip_id, common_id, |block| {
+        // We iterate blocks in the reverse order, so for consistent transaction input/output
+        // dependencies, take transacitons in reverse too,
+        let txns = block.transactions().iter().rev();
+        // We disregerd transactions that are re-added on the newly connected chain
```
TheQuantumPhysicist, line 107: disregard

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+    NewTipIndex,
+    #[error("Block {0:?} not found while traversing history")]
+    BlockNotFound(Id<Block>),
+    #[error("Chainstate call: {0}")]
+    ChainstateCall(#[from] subsystem::subsystem::CallError),
+}
+
+/// Iterate over blocks following the parent link. Apply given function to each block.
+fn for_each_block(
+    chainstate: &dyn ChainstateInterface,
```
TheQuantumPhysicist, line 53: Please avoid dyn when possible. Let's just use impl ChainstateInterface (or generic explicitly).

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+    };
+
+    // Short circuit the expensive processing below if no blocks are being disconnected
+    if old_tip_id == common_id {
+        return Ok(Vec::new());
+    }
+
+    // Collect IDs of trancsactions included in the newly connected chain
+    let mut connected_txs = BTreeSet::new();
+    for_each_block(chainstate, new_tip_id, common_id, |block| {
```
TheQuantumPhysicist, line 97: This is wrong for two reasons:

1. The order of transactions matters within a single block, because transactions can have outputs in the same block. Comingling transactions in a BTreeSet will certainly not yield the correct order.
2. The blocks that are being connected should be done in an ascending order, being later blocks should be able to spend earlier blocks

TheQuantumPhysicist: There should be tests that proves that such cases are covered.

TheQuantumPhysicist: I'll delete this comment and rethink it as I may be wrong with what's going on here.

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+    })?;
+
+    // Collect txns that have been removed from the blockchain and not added on the new fork
+    let mut disconnected_txs = Vec::new();
+    for_each_block(chainstate, old_tip_id, common_id, |block| {
+        // We iterate blocks in the reverse order, so for consistent transaction input/output
+        // dependencies, take transacitons in reverse too,
+        let txns = block.transactions().iter().rev();
+        // We disregerd transactions that are re-added on the newly connected chain
+        let txns = txns.filter(|txn| !connected_txs.contains(&txn.transaction().get_id()));
```
TheQuantumPhysicist, line 108: I'm not too sure why this filter is being done. The presence of a transaction in two branches should be an issue. It's completely valid. In fact, this is very common. When two miners/stakers compete to create a block, usually both of them same the same mempool and share a large fraction of the available transactions. Meaning: A transaction being disconnected on one branch and reconnected in the new branch should be OK.

TheQuantumPhysicist: I guess what's meant here is that the same transaction is already in chainstate... maybe this is OK. 

Geez this process is too complex... it's supposed to be simple and composable... but I see you're trying to do an optimization by not doing it "block by block"...

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+    new_tip: Id<Block>,
+) -> Result<Vec<SignedTransaction>, ReorgError> {
+    let chainstate = mempool.blocking_chainstate_handle();
+
+    let old_tip = mempool.tx_verifier.get_best_block_for_utxos().map_err(|_| ReorgError::OldTip)?;
+    if old_tip == new_tip {
+        return Ok(Vec::new());
+    }
+
+    chainstate.call(move |c| find_disconnected_txs(c.as_ref(), old_tip, new_tip))?
```
TheQuantumPhysicist, line 126: Isn't it better to reduce the load on chainstate by loading all the relevant blocks to the mempool thread and then process them one by one?

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let mut disconnected_txs = fetch_disconnected_txs(mempool, new_tip)
+        .log_err_pfx("Fetching disconnected transactions on a reorg")
+        .unwrap_or_default();
+
+    // Take all the mempool previous transactions
+    let old_store = mem::replace(&mut mempool.store, MempoolStore::new());
+
+    // Discard the old tx verifier and replace it with a fresh one
+    mempool.tx_verifier = tx_verifier::create(
```
TheQuantumPhysicist, line 140: I would create a method to do this instead of manually reassigning the new verifier in the interest of extensibility in the future.

TheQuantumPhysicist:
```
fn mempool.reset_tx_verifier() {
    ...
}
```

TheQuantumPhysicist (approved):
I wish the implementation could be done in a simpler way... not that I can provide a concrete suggestion. The only thing I can suggest is that the reorg process should be done on block-by-block basis, which can simplify this a lot.

Other than that, looks good. 

One thing I'd like to suggest, is adding tests that have transactions that spend each other in a row (both in a block and multiple blocks).

mempool/src/pool/reorg.rs
TheQuantumPhysicist: One thing I've been thinking about... to be clear this doesn't necessarily mean that the transactions you're adding to the mempool will be accepted. This is because there could be a conflict. A conflict is defined as two different transactions spending same output (basically what double-spend does). If every branch has a different transaction spending a different output, then adding transactions will not work. Having this in mind, we gotta ask whether this complicated processing is necessary... who cares? We could just add the transactions from the disconnected branch to the transaction verifier, and if the transaction already exists in chainstate (by merit of a double-spend), then it will be rejected anyway. How much value does it add to attempt to collect ids, filter, etc? My bet is none.

mempool/src/pool/reorg.rs
iljakuklic: Yea, I had impl previously but then it's a private function that is only used once with `&dyn ChainstateInterface` so I thought it's a bit excessive to generalise it. Can put the `impl` back, not a big deal

mempool/src/pool/reorg.rs
iljakuklic: Yea this is basically meant to be an optimisation. It avoids checking transactions that would have been rejected anyway. The idea is that collecting IDs is proportional to the depth of the reorg while looking up stuff in the utxo set is probably rather expensive as it can grow very large over time (proportional to the entire history worst case). In addition to that, the "hit rate" of this optimization is likely to be very high.

It's hard to know how it works out in the end without ptoper benchmarking. I can imagine the difference starts being significant only once the UTXO set reaches certain critical size. I think bitcoin does something like this.

Also I don't think it's particularly complicated, just collecting some IDs and filtering them out...

TheQuantumPhysicist: One thing that could be done to make it easier to read, is create function that does the collection and another function that does the filtering, and add documentation to them explaining the story.

mempool/src/pool/reorg.rs
iljakuklic: Done

mempool/src/pool/reorg.rs
iljakuklic: Attempted to streamline this a bit. Let me know if it looks better.

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,153 @@
...
+            .get_block(curr_block_id)?
+            .ok_or_else(|| ReorgError::BlockNotFound(curr_block_id))?;
+        curr_id = block.prev_block_id();
+        result.push(block);
+    }
+    Ok(result)
+}
+
+/// Blocks affected by a reorg
+struct ReorgInfo {
```
TheQuantumPhysicist, line 71: ReorgData?

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,153 @@
...
+    disconnected: Vec<Block>,
+    connected: Vec<Block>,
```
TheQuantumPhysicist, line 73: Please add a comment about the order of these blocks. It seems that both are in reverse chronological order (higher height first)

mempool/src/pool/reorg.rs
TheQuantumPhysicist: Much nicer now! Thank you!

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl ReorgInfo {
+    /// Extract blocks that have been disconnected and connected from the chainstate.
+    fn from_chainstate<C: ChainstateInterface + ?Sized>(
```
TheQuantumPhysicist, line 78: I don't think that `?Sized` is necessary at all. By default all out subsystem traits that implement deref also implement the trait itself through deref, which means that even if we have an unsized trait, we can just pass it by reference.

common/src/chain/block/block_v1.rs
```diff
@@ -89,10 +89,14 @@ impl BlockV1 {
         self.header.timestamp()
     }
 
-    pub fn transactions(&self) -> &Vec<SignedTransaction> {
+    pub fn transactions(&self) -> &[SignedTransaction] {
         &self.body.transactions
     }
 
+    pub fn into_transactions(self) -> Vec<SignedTransaction> {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: You know what, ignore this comment. We can keep this and add more later if needed.

mempool/src/pool/reorg.rs
```diff
@@ -0,0 +1,153 @@
...
+        })
+    }
+
+    /// Get transactions that have been disconnected and not reconnected
+    fn into_disconnected_transactions(self) -> impl Iterator<Item = SignedTransaction> {
+        let connected_txs: BTreeSet<_> = self
+            .connected
+            .into_iter()
+            .flat_map(|block| {
+                block.into_transactions().into_iter().map(|tx| tx.transaction().get_id())
```
TheQuantumPhysicist, line 104: We can go nuts with rayon here and parallize this. Not necessary though in this PR.

TheQuantumPhysicist (approved):
Besides the few comments I made, looking good now!

azarovh (approved):

PR #865, opened 2023-04-30
Add main menu and tabs to the main window and define a composable hierarchy for widgets

TheQuantumPhysicist:
~While the code compiles and runs, there [seems to be a bug in iced](https://github.com/iced-rs/iced_aw/issues/118). I'm not sure how to proceed with this, but I'm gonna merge this for now and then see later how we can proceed.~

~Please review this code as if the GUI is behaving correctly.~

Issue is fixed. Please perform a regular review.


TheQuantumPhysicist:
The bug can be circumvented. I did that. I'll push after a rebase.

azarovh (approved):
looks good

muursh (approved):
No complaints here

PR #866, opened 2023-05-01
Use in-memory sqlite database in tests

pavel-kokolemin:
Our tests became much slower after the SQLite backend tests were added. This can be fixed by using in-memory databases. It looks like we don't do anything with the DB files themselves (don't try to reopen them, for example), so this should be acceptable.

Here are my test times when I run `nextest run --all --no-fail-fast`:
Before:
```
Summary [  22.921s] 1773 tests run: 1773 passed, 14 skipped
```
After:
```
Summary [  15.927s] 1773 tests run: 1773 passed, 14 skipped
```

### DIFF
--- a/storage/sqlite/src/lib.rs
+++ b/storage/sqlite/src/lib.rs
@@ -218,6 +218,22 @@ impl backend::BackendImpl for SqliteImpl {
 }
 
 #[derive(Eq, PartialEq, Clone, Debug)]
+pub struct Options {
+    /// If enabled, sets synchronous pragma to OFF, see <https://www.sqlite.org/pragma.html#pragma_synchronous>.
+    /// It should normally only be used in unit tests.
+    pub disable_fsync: bool,
+}
+
+#[allow(clippy::derivable_impls)]
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            disable_fsync: false,
+        }
+    }
+}
+
+#[derive(Eq, PartialEq, Clone, Debug)]
 enum SqliteStorageMode {
     InMemory,
     File(PathBuf),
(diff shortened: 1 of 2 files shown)


iljakuklic:
Alternatively, if it's deemed useful to test sqlite with proper database files, the number of proptest checks in backend-test-suite could be made configurable. We could then adjust it to get a balance between coverage and performance we want.

Or even both: small number of cases with on-disk database and higher coverage with in-memory one...

TheQuantumPhysicist:
> It looks like we don't do anything with the DB files themselves (don't try to reopen them, for example)

This actually concerns me. We should have tests that cover closing and reopening a database. 

Besides that, I'm not sure how to deal with this. On one hand, you're totally right with your assessment, Pavel. On the other hand, testing the files is useful, because this is the lowest level of our implementation + files implementation.

pavel-kokolemin:
I updated the PR to run tests with `PRAGMA schema.synchronous = OFF` instead.
We still use files and the tests are still much faster:
```
     Summary [  15.155s] 1780 tests run: 1780 passed, 14 skipped
```

TheQuantumPhysicist (approved):
Very nice! Didn't know sqlite supports turning off synchronous writes. You have my blessing :smile:

storage/sqlite/src/lib.rs
```diff
@@ -226,22 +242,29 @@ enum SqliteStorageMode {
...
     /// New Sqlite database backend
     pub fn new<P: AsRef<Path>>(path: P) -> Self {
         Self {
             backend: SqliteStorageMode::File(path.as_ref().to_path_buf()),
+            options: Default::default(),
         }
     }
 
+    pub fn with_options(self, options: Options) -> Self {
+        Self { options, ..self }
```
TheQuantumPhysicist, line 265: Please don't use `..`

muursh (approved):

PR #867, opened 2023-05-02
Add non-interactive mode to wallet-cli

pavel-kokolemin:
CLI wallet has been updated to be able run in non-interactive mode (when TTY is not available).

It can be used by sending commands to standard input:
```
$ echo -n "newwallet /tmp/Wallet1 \n closewallet \n nodeversion" | cargo run --bin wallet-cli -- --network regtest
New wallet created successfully
Your mnemonic: speed slot tunnel inch faculty wheel lizard artefact hat obscure apart own hire nephew dove office copper update tragic worth unusual merge step fat
Please write it somewhere safe to be able to restore your wallet.
Success
0.1.0
```

Or by specifying file with commands as a command line argument:
```
$ cat commands.sh
nodeversion
connectedpeers
peercount

$ cargo run --bin wallet-cli -- --network regtest --commands-file commands.sh
0.1.0
[]
0
```
Empty lines and lines starting with '#' are ignored.

To keep things simple, I have removed interactive wallet file name prompt. This makes CLI wallet less user-friendly, but more flexible. This should be acceptable since we will also have the GUI wallet.

BTW, CLI wallet looks like a real shell now. It's even possible to use it with shebang scripts:
```
$ cat ./run.sh
#!/usr/bin/env -S /home/admin0/rep/mintlayer/mintlayer-core/target/debug/wallet-cli --network regtest --commands-file
nodeversion
connectedpeers
peercount

$ ./run.sh
0.1.0
[]
0
```


wallet/wallet-cli/src/repl/non_interactive/log.rs
```diff
@@ -13,23 +13,6 @@
...
-    let index = dialoguer::Select::with_theme(theme)
-        .with_prompt(prompt)
-        .default(0)
-        .items(&texts)
-        .interact_opt()
-        .map_err(WalletCliError::ConsoleIoError)?
-        .ok_or(WalletCliError::Cancelled)?;
-    Ok(items[index].clone())
+pub fn init() {
+    env_logger::builder().init();
```
TheQuantumPhysicist, line 17: Logging must strictly come from the logging crate.

wallet/wallet-controller/src/lib.rs
```diff
@@ -39,34 +44,41 @@ const BLOCK_SYNC_ENABLED: bool = false;
...
+    WalletError(wallet::wallet::WalletError),
 }
 
 pub struct Controller<T: NodeInterface> {
+    chain_config: Arc<ChainConfig>,
+
     rpc_client: T,
 
-    wallet: DefaultWallet,
+    wds: Vec<WalletData<T>>,
```
TheQuantumPhysicist, line 58: Why multiple wallets like this?

People who want to open multiple wallets should run the software multiple times. That simplifies things a lot.

wallet/wallet-controller/src/lib.rs
```diff
@@ -175,11 +187,82 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+    ) -> Result<(), ControllerError<T>> {
+        utils::ensure!(
+            !file_path.as_ref().exists(),
+            ControllerError::WalletFileError(
+                file_path.as_ref().to_owned(),
+                "File already exists".to_owned()
+            )
+        );
+
+        let db = wallet::wallet::open_or_create_wallet_file(file_path)
```
TheQuantumPhysicist, line 204: I would prefer to have a distinction between open and create so that errors can't happen. 

Of course this depends on the context. but then if you're calling the funciton "create_wallet", then it shouldn't be possible to overwrite a wallet.

wallet/wallet-controller/src/lib.rs
```diff
@@ -175,11 +187,82 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+    pub fn open_wallet(&mut self, file_path: impl AsRef<Path>) -> Result<(), ControllerError<T>> {
+        utils::ensure!(
+            file_path.as_ref().exists(),
+            ControllerError::WalletFileError(
+                file_path.as_ref().to_owned(),
+                "File does not exist".to_owned()
+            )
+        );
+
+        let db = wallet::wallet::open_or_create_wallet_file(file_path)
```
TheQuantumPhysicist, line 227: Same here... open shouldn't create a wallet.

wallet/wallet-controller/src/lib.rs
TheQuantumPhysicist: Even in the GUI, the handle for opening a wallet should be accessed multiple times to open multiple wallets.

For example, if there's a function `open_or_create_wallet`, for every wallet the user wants to open, that function is called, a controller is returned, and that controller is what's used in the GUI to communicate with the wallet.

wallet/wallet-controller/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 chainstate = { path = "../../chainstate" }
 common = { path = "../../common" }
 crypto = { path = "../../crypto" }
+futures.workspace = true
```
TheQuantumPhysicist, line 12: The workspace dependencies are grouped together (below in this file)

wallet/wallet-controller/src/lib.rs
pavel-kokolemin: We check that the file exists.
Should I move the check to the wallet crate and split `open_or_create_wallet_file` to two functions?

wallet/wallet-controller/src/lib.rs
pavel-kokolemin: Okay, reverted.

wallet/wallet-controller/src/lib.rs
TheQuantumPhysicist: For CLI it makes sense to open or create. I'm worried we're gonna have a problem in the GUI where someone tries to create a wallet, then they have an issue where we overwrite their wallet or something.

So, this is OK for the CLI, but for the GUI there has to be a strong distinction, since people usually select wallets manually through a file dialog.

wallet/wallet-cli/src/repl/non_interactive/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+                    Ok(Some(command)) => {
+                        let res = super::run_command_blocking(&event_tx, command);
+
+                        match res {
+                            Ok(cmd) => match cmd {
+                                ConsoleCommand::Print(text) => {
+                                    cli_println!(output, "{}", text);
+                                }
+                                ConsoleCommand::ClearScreen => {
+                                    cli_println!(output, "Not supported");
```
TheQuantumPhysicist, line 48: I would break and return an error when using unsupported commands in non-interactive mode.

wallet/wallet-cli/src/repl/non_interactive/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+                                }
+                                ConsoleCommand::PrintHistory => {
+                                    cli_println!(output, "Not supported");
```
TheQuantumPhysicist, line 51: When you said you removed history in the meeting, did you mean only for non-interactive mode? Because that's probably OK.

wallet/wallet-cli/src/main.rs
```diff
@@ -21,28 +21,31 @@ mod commands;
...
 const COOKIE_FILENAME: &str = ".cookie";
 
-async fn run(output: &ConsoleContext) -> Result<(), WalletCliError> {
-    logging::init_logging::<&std::path::Path>(None);
+enum Mode {
+    Interactive {
+        printer: reedline::ExternalPrinter<String>,
+    },
+    NonInteractive,
+    Commands(std::fs::File),
```
TheQuantumPhysicist, line 45: `CommandsList(Vec<String>)`.

This makes it possible to run unit tests with this. Then the only difference is that you do an std::fs::read_lines() or similar and input here.

The idea here is we limit having interaction with files directly as much as possible. Dealing with the file system invites all kinds of problems.

wallet/wallet-cli/src/repl/non_interactive/mod.rs
pavel-kokolemin: No, I meant for interactive mode. History is now not persistent between application runs (unless the history file location is specified on the command line). Non-interactive does not use/need history.

pavel-kokolemin: But I will change that and make sure that only safe history is stored there.

TheQuantumPhysicist: Maybe also see if it's easy to have a shortcut like shift+del to delete a history entry.

wallet/wallet-controller/src/lib.rs
pavel-kokolemin: I agree. But UIs should use the `wallet-controller` crate to manage wallets, and there are two separate functions specifically for that: `create_wallet` and `open_wallet`. `create_wallet` checks that the file does not exist and `open_wallet` checks that the wallet file exists. So I don't see a problem here.

wallet/wallet-cli/src/repl/non_interactive/mod.rs
pavel-kokolemin: I think breaking on unsupported commands should be part of more robust error handling. I updated wallet-cli to break on a first error in non-interactive mode (can be also enabled/disabled from the command line).

wallet/wallet-cli/src/main.rs
```diff
@@ -21,28 +21,31 @@ mod commands;
...
-use dialoguer::theme::ColorfulTheme;
+use crossterm::tty::IsTty;
 use errors::WalletCliError;
 use tokio::sync::mpsc;
-use utils::default_data_dir::{default_data_dir_for_chain, prepare_data_dir};
-use wallet::Wallet;
+use utils::default_data_dir::default_data_dir_for_chain;
 use wallet_controller::cookie::load_cookie;
 
 const COOKIE_FILENAME: &str = ".cookie";
```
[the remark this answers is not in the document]
pavel-kokolemin: Yes, it should. But I did not want to add dependency to the rpc crate. I moved some cookie related functions to the utils crate.

wallet/wallet-cli/src/main.rs
pavel-kokolemin: Updated. I still need to do IO to read from `stdin`, but this is now hidden behind generic `Iterator`:
```
pub fn run(
    lines: impl Iterator<Item = String>,
    output: &ConsoleContext,
    event_tx: mpsc::UnboundedSender<Event>,
    exit_on_error: bool,
) -> Result<(), WalletCliError> {
```

TheQuantumPhysicist: Cool. No problem there (I'll look at that part indepndently).

Though please call the arm "CommandsList", not "Commands".

wallet/wallet-cli/src/repl/non_interactive/mod.rs
TheQuantumPhysicist: Yep. In non-interactive mode should break and exit. But in interactive mode it should just print the error. Though I'm making this comment because this file is in non-interactive.

wallet/wallet-controller/src/lib.rs
TheQuantumPhysicist: Alrighty. Sounds good.

wallet/wallet-cli/src/repl/non_interactive/mod.rs
TheQuantumPhysicist: To be clear, I'm OK with history being persistent just in individual runs. That's acceptable.

pavel-kokolemin:
I added a basic integration test for the CLI wallet. It creates a temporary directory, spawns the node and CLI wallet, and executes test commands (all in one process). 

I had a problem with this check and disabled it for now:
```
assert!(cfg!(not(feature = "testing_utils")));
```
I will debug why `testing_utils` is enabled later.

wallet/wallet-cli/src/repl/non_interactive/mod.rs
pavel-kokolemin: OK, great. It's done then. And if someone wants persistent history between runs, we have a command-line option for that.

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+                line_editor.print_history().expect("Should not fail normally");
+            }
+            Ok(Some(ConsoleCommand::Exit)) => return Ok(()),
+
+            Ok(None) => {}
+
+            Err(err) => {
+                if exit_on_error {
+                    return Err(err);
+                }
```
iljakuklic, line 171: Should the error be printed out even if `exit_on_error` is enabled? Or does it happen somewhere else?

iljakuklic (approved):
Very nice! A few minor nitpicks below.

node-lib/src/runner.rs
```diff
@@ -136,17 +138,28 @@ pub async fn initialize(
...
-        let _rpc = manager.add_subsystem("rpc", rpc?);
-    }
+        let rpc = rpc.await?;
+        rpc_http_address = rpc.http_address().cloned();
+        rpc_websocket_address = rpc.websocket_address().cloned();
+        let _rpc = manager.add_subsystem("rpc", rpc);
+    } else {
+        rpc_http_address = None;
+        rpc_websocket_address = None;
+    };
```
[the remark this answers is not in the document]
pavel-kokolemin: I usually prefer this way too. But in this particular case it does not look good.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -13,18 +13,39 @@
...
 #[derive(Debug, Parser)]
 #[clap(rename_all = "lower")]
 pub enum WalletCommand {
+    /// Create new wallet
+    CreateWallet {
+        /// File path
+        wallet_path: PathBuf,
+
+        // Mnemonic
+        mnemonic: Option<String>,
```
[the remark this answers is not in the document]
pavel-kokolemin: Updated, thanks. Other help messages are still missing.

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
pavel-kokolemin: Yes, the error is printed from main in this case.

pavel-kokolemin:
There was an annoying problem with the `testing_utils` feature flag in p2p. It was being enabled/disabled depending on the build command used. As a result, the node would sometimes crash unexpectedly due to this p2p startup check:
```
assert!(cfg!(not(feature = "testing_utils")));
```
Also, a lot of code recompilation was happening after changing the flag combination (for example starting the node after running the tests).
To fix all this, I did a small refactoring and removed this feature flag from p2p.

I also disabled this test, because it only passed because of the above assert check:
```
#[test]
fn no_args() {
    Command::new(BIN_NAME).assert().failure();
}```

File: node-lib/src/runner.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (iljakuklic) [reply node-lib/src/runner.rs:148]: Ok, no bother.

File: wallet/wallet-cli-lib/tests/basic.rs [hunk 13]
```diff
@@ -0,0 +1,111 @@
...
+};
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn basic_wallet_cli() {
+    // logging::init_logging::<std::path::PathBuf>(None);
+
+    let test_root = test_utils::test_root!("wallet-cli-tests").unwrap();
+    let test_dir = test_root.fresh_test_dir("basic_wallet_cli");
+
+    let commands_file = test_dir.as_ref().join("commands.txt");
```
TheQuantumPhysicist, line 29: Why not use the vector input that was developed instead of depending on a file in a test? The whole point of that vector is to minimize file I/O to the point where our abstraction layer is so good that we basically are testing reading a file if we use a file.

pavel-kokolemin: Minimizing IO is probably a bit pointless since we start the whole node here. Anyway, I switched to using a generic console implementation to abstract console IO.

pavel-kokolemin:
PR has been updated to start the in-memory node (RPC interface is still used).

wallet/wallet-cli-lib/tests/basic.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn wallet_cli_file() {
+    let test = CliTestFramework::setup().await;
+
+    // Use dir name with spaces to make sure quoting works as expected
+    let file_name = test
+        .test_root
+        .fresh_test_dir("wallet dir")
```
TheQuantumPhysicist, line 219: Is that a space in the name? Maybe that's a bad idea.

wallet/wallet-cli-lib/tests/basic.rs
```diff
@@ -0,0 +1,234 @@
...
+        .join("wallet1")
+        .to_str()
+        .unwrap()
+        .to_owned();
+
+    let output = test.run(&[&format!("createwallet \"{file_name}\"")]).await;
+    assert_eq!(output.len(), 1);
+    assert!(output[0].starts_with("New wallet created successfully\n"));
+
+    let output = test.run(&[&format!("openwallet \"{file_name}\""), "closewallet"]).await;
```
TheQuantumPhysicist, line 230: If I understand this correctly, please add a comment "Start the wallet, open it, then close it, then shutdown".

TheQuantumPhysicist (approved):
Besides the minor things I mentioned, I think this is OK. Cheers!

PR #869, opened 2023-05-02
Subscription to events in the GUI

TheQuantumPhysicist:
Subscribe to events and update GUI elements from the results

### DIFF
--- a/node-gui/src/main_window/main_widget/tabs/summary.rs
+++ b/node-gui/src/main_window/main_widget/tabs/summary.rs
@@ -26,7 +26,10 @@ use iced::{
 };
 use iced_aw::{tab_bar::TabLabel, Grid};
 use subsystem::Handle;
-use tokio::sync::mpsc::UnboundedSender;
+use tokio::sync::{
+    mpsc::{UnboundedReceiver, UnboundedSender},
+    Mutex,
+};
 use utils::tap_error_log::LogError;
 
 use crate::backend_controller::NodeBackendController;
@@ -49,8 +52,9 @@ pub enum WidgetDataUpdate {
 
 #[derive(Clone)]
 pub struct RegisteredSubscriptions {
-    chainstate_event_receiver:
-        Arc<tokio::sync::Mutex<tokio::sync::mpsc::UnboundedReceiver<(Id<GenBlock>, BlockHeight)>>>,
+    // Unfortunately, GUI messages must support Clone, and async channel receivers are not Clone, so we need this (ugly) Arc.
+    #[allow(clippy::type_complexity)]
+    chainstate_event_receiver: Arc<Mutex<UnboundedReceiver<(Id<GenBlock>, BlockHeight)>>>,
 }
 
 impl Debug for RegisteredSubscriptions {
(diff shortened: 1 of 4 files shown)


node-gui/src/main_window/main_widget/tabs/summary.rs
```diff
@@ -13,34 +13,144 @@
...
+pub enum WidgetDataUpdate {
+    TipUpdated((Id<GenBlock>, BlockHeight)),
     NoOp,
 }
 
+#[derive(Clone)]
+pub struct RegisteredSubscriptions {
+    // Unfortunately, GUI messages must support Clone, and async channel receivers are not Clone, so we need this (ugly) Arc.
+    #[allow(clippy::type_complexity)]
+    chainstate_event_receiver: Arc<Mutex<UnboundedReceiver<(Id<GenBlock>, BlockHeight)>>>,
```
pavel-kokolemin, line 57: Maybe it would be better to run a persistent async event loop in the background, where we can process requests from the UI and events from the node? We need such a loop to sync wallets anyway. Once we have it, we won't need to use non-POD messages.

pavel-kokolemin (approved):

node-gui/src/main_window/main_widget/tabs/summary.rs
TheQuantumPhysicist: The loop for the GUI has to be based on something like this, because the update function is what updates the GUI. If we want to update the GUI, we have to issue a command through update, which will suspend until there's change in the state of the backend. At first it was an independent event loop like you're suggesting, but then later I realized it's not really possible to have a separate event loop for the GUI. We have to, somehow, connect events together through update and iced::Command.

Btw, none of this stops having event loops in other subsystems. This is just for the the GUI view. So the wallet can still have its event loop, but updating the GUI will always have something like this... unless I still am a noob in this GUI library and there's a better way I don't know :smile:

pavel-kokolemin: I see. I don't have an answer for how it can be done with `iced`. But I'm pretty sure it should be possible, and the the code will be cleaner if we have a background event loop for async processing. Please let me know if I should look into this.

stanislav-tkach (approved):

node-gui/src/main_window/main_widget/tabs/summary.rs
TheQuantumPhysicist: I would let you take this over... but you're doing more important things :smile:

iljakuklic: I don't know how to wire it up exactly with the node controller and such but there's a concept called subscription that is supposed to be possible to use as a sort of bridge between async code and iced.

The [iced::subscription::channel](https://docs.rs/iced/0.9.0/iced/subscription/fn.channel.html) function seems vaguely relevant, and there's also the [websocket example](https://github.com/iced-rs/iced/tree/0.9/examples/websocket) that demonstrates listening to external events.

TheQuantumPhysicist: I'll look into that more. Thanks.

PR #870, opened 2023-05-03
Implement the p2p shutdown

stanislav-tkach:


p2p/src/lib.rs
```diff
@@ -119,21 +139,31 @@ where
             time_getter.clone(),
             peerdb_storage,
         )?;
-        tokio::spawn(async move {
-            // TODO: Shutdown p2p if PeerManager unexpectedly quits
-            peer_manager.run().await.tap_err(|err| log::error!("PeerManager failed: {err}"))
+        let shutdown_ = Arc::clone(&shutdown);
+        let peer_manager_task = tokio::spawn(async move {
+            match peer_manager.run().await {
+                Ok(_) => {}
```
pavel-kokolemin, line 145: Please use `unreachable!` for consistency.

pavel-kokolemin (approved):

p2p/src/interface/p2p_interface.rs
```diff
@@ -33,3 +32,8 @@ pub trait P2pInterface: Send + Sync {
 
     async fn submit_transaction(&mut self, tx: SignedTransaction) -> crate::Result<()>;
 }
+
+#[async_trait::async_trait]
+pub trait P2pSubsystemInterface: 'static {
+    async fn run(self, call: CallRequest<dyn P2pInterface>, shutdown: ShutdownRequest);
+}
```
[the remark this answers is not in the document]
pavel-kokolemin: Looks like it's not that easy. I guess no need to spend time on it. Sorry!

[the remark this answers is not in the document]
stanislav-tkach: @pavel-kokolemin Don't worry, the current solution isn't really optimal.

@TheQuantumPhysicist We discussed this with Pavel and he proposed a slightly different solution: to extend the p2p interface with the shutdown method. Something like this:
```rust
trait P2pInterface: Send + Sync {
    async fn do_shutdown(self);
}

#[async_trait::async_trait]
impl subsystem::Subsystem for Box<dyn P2pInterface> {
    async fn shutdown(self) {
        self.do_shutdown().await;
    }
}
```
I think it is somewhat better, but I still don't like the idea of the shutdown method in the interface.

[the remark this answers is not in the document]
TheQuantumPhysicist: We shouldn't have a shutdown method in the p2p interface... that's just wrong. Shutdown events should strictly come from subsystem. Now I have to say I'm not familiar with the current predicament so I don't understand why this is even being discussed.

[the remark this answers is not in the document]
TheQuantumPhysicist: I guess what I'm asking is: Why are we not using concrete methods and are using interfaces instead for run and shutdown?!

p2p/src/net/default_backend/mod.rs
```diff
@@ -131,18 +144,27 @@ impl<T: TransportSocket> NetworkingService for DefaultNetworkingService<T> {
...
+                shutdown_,
+                shutdown_receiver,
             );
 
-            // TODO: Shutdown p2p if the backend unexpectedly quits
-            if let Err(err) = backend.run().await {
-                log::error!("failed to run backend: {err}");
+            match backend.run().await {
+                Ok(_) => unreachable!(),
+                Err(P2pError::ChannelClosed) if shutdown.load(Ordering::Acquire) => {
```
TheQuantumPhysicist, line 153: Please use SeqCst as long as performance doesn't matter.

p2p/src/net/default_backend/mod.rs
```diff
@@ -131,18 +144,27 @@ impl<T: TransportSocket> NetworkingService for DefaultNetworkingService<T> {
...
+                    log::info!("Backend is shut down");
+                }
+                Err(e) => {
+                    shutdown.store(true, Ordering::Release);
```
TheQuantumPhysicist, line 157: Same here

p2p/src/sync/mod.rs
```diff
@@ -116,15 +116,19 @@ where
 
         let mut new_tip_receiver = self.subscribe_to_new_tip().await?;
         self.is_initial_block_download.store(
-            self.chainstate_handle.call(|c| c.is_initial_block_download()).await??,
+            self.chainstate_handle
+                .call(|c| c.is_initial_block_download())
+                .await
+                // This shouldn't fail unless the chainstate subsystem is down.
```
TheQuantumPhysicist, line 122: "which shouldn't happen since subsystems are shutdown in reverse order"

p2p/src/sync/peer.rs
```diff
@@ -141,7 +140,15 @@ where
         *self.id
     }
 
-    pub async fn run(&mut self) -> Result<Void> {
+    pub async fn run(&mut self) {
+        match self.main_loop().await {
+            // The unexpect "channel closed" error will be handled by the sync manager.
```
TheQuantumPhysicist, line 145: unexpected?

p2p/Cargo.toml
```diff
@@ -45,13 +45,15 @@ chainstate-storage = { path = "../chainstate/storage" }
 chainstate-test-framework = { path = "../chainstate/test-framework" }
 consensus = { path = "../consensus" }
 crypto = { path = "../crypto/" }
+storage-lmdb = { path = "../storage/lmdb" }
```
TheQuantumPhysicist, line 48: This has to go

subsystem/src/manager.rs
```diff
@@ -30,9 +30,9 @@ use crate::subsystem::{CallRequest, Handle, ShutdownRequest, Subsystem, Subsyste
 /// Manager configuration options.
 pub struct ManagerConfig {
     /// Subsystem manager name
-    name: &'static str,
+    pub name: &'static str,
     /// Shutdown timeout. Set to `None` for no (i.e. unlimited) timeout.
-    shutdown_timeout_per_subsystem: Option<Duration>,
+    pub shutdown_timeout_per_subsystem: Option<Duration>,
```
TheQuantumPhysicist, line 35: Please remove pub

p2p/src/interface/p2p_interface.rs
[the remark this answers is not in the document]
stanislav-tkach: The `P2p` struct is generic over the networking service (transport) and the concrete transport is chosen based on runtime parameters, so it is impossible to return it from a function:
```rust
pub async fn make_p2p<S: PeerDbStorage + 'static>(
    p2p_config: Arc<P2pConfig>,
    chainstate_handle: subsystem::Handle<Box<dyn chainstate_interface::ChainstateInterface>>,
    mempool_handle: MempoolHandle,
    time_getter: TimeGetter,
    peerdb_storage: S,
) -> Result<Box<dyn P2pInterface>> {
    if let Some(true) = p2p_config.disable_noise {
        let transport = make_p2p_transport_unencrypted();

        let p2p = P2p::<P2pNetworkingServiceUnencrypted>::new(
            transport,
            bind_addresses,
            chain_config,
            p2p_config,
            chainstate_handle,
            mempool_handle,
            time_getter,
            peerdb_storage,
        )
        .await?;

        Ok(Box::new(p2p))
    } else if let Some(socks5_proxy) = &p2p_config.socks5_proxy {
        let transport = make_p2p_transport_socks5_proxy(socks5_proxy);

        let p2p = P2p::<P2pNetworkingServiceSocks5Proxy>::new(
            transport,
            bind_addresses,
            chain_config,
            p2p_config,
            chainstate_handle,
            mempool_handle,
            time_getter,
            peerdb_storage,
        )
        .await?;

        Ok(Box::new(p2p))
    } else {
        let transport = make_p2p_transport();

        let p2p = P2p::<P2pNetworkingService>::new(
            transport,
            bind_addresses,
            chain_config,
            p2p_config,
            chainstate_handle,
            mempool_handle,
            time_getter,
            peerdb_storage,
        )
        .await?;

        Ok(Box::new(p2p))
    }
}
```

Therefore the interface (`Box<dyn P2pInterface>`) is returned. As a result, the `Subsystem` trait is implemented for the interface instead of the `P2p` type.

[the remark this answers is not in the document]
TheQuantumPhysicist: So in a call we clarified that this trait isn't necessary, and the implementee struct can just be returned instead.

I hope the same would be done for the mempool. I just realized it's done that way there as well.

TheQuantumPhysicist (approved):

PR #874, opened 2023-05-04
Add block signature to block header

TheQuantumPhysicist:
As a first step, we add the signature field of the block.

The processing comes with proof of stake. It's not done yet. An issue will be opened for that.

#875

### DIFF
--- a/common/src/chain/block/signed_block_header.rs
+++ b/common/src/chain/block/signed_block_header.rs
@@ -13,7 +13,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crypto::key::Signature;
 use serialization::{Decode, Encode};
 use typename::TypeName;
 
@@ -28,15 +27,33 @@ pub enum SignedHeaderError {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeName)]
+pub struct BlockHeaderSignatureData {
+    // TODO: determine the proper types for these fields.
+    //       So far, the plan is to make the public key match
+    //       the kernel input destination and hence it should
+    //       be able to verify it too.
+    public_key: Vec<u8>,
+    signature: Vec<u8>,
+}
+
+impl BlockHeaderSignatureData {
+    pub fn new(public_key: Vec<u8>, signature: Vec<u8>) -> Self {
+        Self {
+            public_key,
+            signature,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeName)]
 pub enum BlockHeaderSignature {
     None,
-    PoSBlock(Signature),
+    PoSBlock(BlockHeaderSignatureData),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeName, serialization::Tagged)]
 pub struct SignedBlockHeader {
     block_header: BlockHeader,
-    // TODO: Do we want to include the public key to make it possible to cross-check with the spent kernel?
     signature: BlockHeaderSignature,
 }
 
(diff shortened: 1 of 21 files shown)


common/src/chain/block/signed_block_header.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn new(signature: BlockHeaderSignature, block_header: BlockHeader) -> Self {
+        Self {
+            signature,
+            block_header,
+        }
+    }
+
+    pub fn signature(&self) -> &BlockHeaderSignature {
+        &self.signature
+    }
```
alfiedotwtf, line 47: Probably want a `sign(pk)` method to sign headers currently without a signature

alfiedotwtf (approved):

azarovh (approved):

stanislav-tkach (approved):

common/src/chain/block/signed_block_header.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    pub fn header(&self) -> &BlockHeader {
+        &self.block_header
+    }
+
+    pub fn header_mut(&mut self) -> &mut BlockHeader {
+        &mut self.block_header
+    }
```
iljakuklic, line 55: I am somewhat uncomfortable with `header_mut` in a signed header. Giving a mutable access to the header changes the header hash which will invalidate any signature that may already be present.

I'd suggest:
* Either document the potential signature integrity violation (possibly even in the method name), or
* remove this method and use `into_block_header`, modify the header, and reconstruct signed header afterwards instead.

iljakuklic (commented):
Here are some initial thoughts. Some potentially important.

common/src/chain/block/signed_block_header.rs
iljakuklic: Another alternative is to return `Option<&mut BlockHeader>` and return a `Some(_)` only if the header signature is `None`.

common/src/chain/block/signed_block_header.rs
```diff
@@ -0,0 +1,88 @@
...
+        self.block_header
+    }
+
+    pub fn consensus_data(&self) -> &ConsensusData {
+        &self.header().consensus_data
+    }
+
+    pub fn block_id(&self) -> Id<Block> {
+        self.header().get_id()
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Block malleabiility doesn't matter like transaction malleability. Having multiple signatures for the same block (in theory) shouldn't really matter as long as they're all valid. The question is always: What's the worst thing that can happen?

For transactions it matters because people track transactions by their id, and it can ruin some applications to see that their transaction didn't go through, while it really did.

common/src/chain/block/signed_block_header.rs
TheQuantumPhysicist: Not a bad idea. I'll go ahead and do it.

common/src/chain/block/signed_block_header.rs
TheQuantumPhysicist: Maybe later. The interface is still not clear.

common/src/chain/block/signed_block_header.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeName)]
+pub enum BlockHeaderSignature {
+    None,
+    PoSBlock(Signature),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeName, serialization::Tagged)]
+pub struct SignedBlockHeader {
+    block_header: BlockHeader,
+    // TODO: Do we want to include the public key to make it possible to cross-check with the spent kernel?
```
muursh, line 33: I'd say yes.

muursh (approved):

iljakuklic (approved):

common/src/chain/block/signed_block_header.rs
TheQuantumPhysicist: Alright. I'll add it and add some notes about it. Doing this properly will be some work.

PR #876, opened 2023-05-05
PoSAccounting simulation test

### LINKED ISSUE #568: Add simulation test for PoSAccountingOperations
Test must perform random operations and check invariants. For example that owner balance + all delegations balance = pool balance

azarovh:
Add tests that perform random operations and check storage consistency on every iteration.

Closes #568

### DIFF
--- a/pos_accounting/src/pool/storage/operator_impls.rs
+++ b/pos_accounting/src/pool/storage/operator_impls.rs
@@ -62,6 +62,7 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingOperations
             pool_id,
             PoSAccountingUndo::CreatePool(CreatePoolUndo {
                 pool_id,
+                pledge_amount: pool_data.pledge_amount(),
                 data_undo: PoolDataUndo::Data(Box::new(pool_data)),
             }),
         ))
@@ -73,11 +74,17 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingOperations
             .get_pool_data(pool_id)?
             .ok_or(Error::AttemptedDecommissionNonexistingPoolData)?;
 
+        let pool_balance = self
+            .store
+            .get_pool_balance(pool_id)?
+            .ok_or(Error::AttemptedDecommissionNonexistingPoolBalance)?;
+
         self.store.del_pool_balance(pool_id)?;
         self.store.del_pool_data(pool_id)?;
 
         Ok(PoSAccountingUndo::DecommissionPool(DecommissionPoolUndo {
             pool_id,
+            pool_balance,
             data_undo: PoolDataUndo::Data(Box::new(pool_data)),
         }))
     }
@@ -105,6 +112,7 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingOperations
         Ok(PoSAccountingUndo::IncreasePledgeAmount(
             IncreasePledgeAmountUndo {
                 pool_id,
+                amount_added: amount_to_add,
                 data_undo: PoolDataUndo::Data(Box::new(pool_data)),
             },
         ))
@@ -174,7 +182,10 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingOperations
 
         self.sub_delegation_from_pool_share(pool_id, delegation_id, amount)?;
 
-        self.sub_balance_from_pool(pool_id, amount)?;
+        // it's possible that the pool was decommissioned
+        if self.pool_exists(pool_id)? {
+            self.sub_balance_from_pool(pool_id, amount)?;
+        }
 
         self.sub_from_delegation_balance(delegation_id, amount)?;
 
@@ -200,16 +211,14 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingOperations
 
 impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingDB<S, T> {
     fn undo_create_pool(&mut self, undo: CreatePoolUndo) -> Result<(), Error> {
-        let amount = self.store.get_pool_balance(undo.pool_id)?;
-
-        let data_undo = match undo.data_undo {
+        match undo.data_undo {
             PoolDataUndo::Data(v) => v,
             PoolDataUndo::DataDelta(_) => panic!("incompatible PoolDataUndo supplied"),
         };
 
-        match amount {
+        match self.store.get_pool_balance(undo.pool_id)? {
             Some(amount) => {
-                if amount != data_undo.pledge_amount() {
+                if amount != undo.pledge_amount {
                     return Err(Error::InvariantErrorPoolCreationReversalFailedAmountChanged);
                 }
             }
@@ -240,7 +249,7 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingDB<S, T> {
             return Err(Error::InvariantErrorDecommissionUndoFailedPoolDataAlreadyExists);
         }
 
-        self.store.set_pool_balance(undo.pool_id, data_undo.pledge_amount())?;
+        self.store.set_pool_balance(undo.pool_id, undo.pool_balance)?;
         self.store.set_pool_data(undo.pool_id, &data_undo)?;
 
         Ok(())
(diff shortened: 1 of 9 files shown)


pos_accounting/src/storage/in_memory.rs
```diff
@@ -59,6 +59,69 @@ impl InMemoryPoSAccounting {
...
+        // pool_balance and pool_data must contain the same keys
+        assert_eq!(self.pool_balances.len(), self.pool_data.len());
+        let pool_balance_and_data_consistant =
+            self.pool_balances.keys().all(|key| self.pool_data.contains_key(key));
+        assert!(pool_balance_and_data_consistant);
+
+        // delegation_balance and delegation_data must contain the same keys
+        //
+        // Note: delegation_balances and delegation_data can have different length
+        // because zero balances are removed.
```
azarovh, line 89: Initially, we decided not to store zero balances in deltas and storage because it seemed redundant. But maybe we should reconsider this for consistency reasons as it would be easier to check the state. Thoughts?

pos_accounting/src/pool/storage/operator_impls.rs
```diff
@@ -301,7 +310,10 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingDB<S, T> {
 
         self.add_to_delegation_balance(undo_data.delegation_id, undo_data.amount)?;
 
-        self.add_balance_to_pool(pool_id, undo_data.amount)?;
+        // it's possible that the pool was decommissioned
+        if self.pool_exists(pool_id)? {
+            self.add_balance_to_pool(pool_id, undo_data.amount)?;
+        }
```
iljakuklic, line 316: Is it legal to have a non-existing pool here? I.e. should there be an `else` with an error?
This is replicated in multiple places, could the code somehow be shared?

pos_accounting/src/storage/in_memory.rs
```diff
@@ -59,6 +59,69 @@ impl InMemoryPoSAccounting {
...
+        // delegation balances and delegation shares must contain the same amounts
+        assert_eq!(
+            self.delegation_balances.len(),
+            self.pool_delegation_shares.len()
+        );
+        let delegation_shares_and_balances_consistent = self
+            .pool_delegation_shares
+            .iter()
+            .all(|((_, key), balance)| self.delegation_balances.get(key) == Some(balance));
+        assert!(delegation_shares_and_balances_consistent);
```
iljakuklic, line 123: Wondering if there's a value in using `.for_each(... assert_eq!` instead of `.all(... ==`. You get information about what failed rather than that something failed.

iljakuklic (approved):
Nice. Some minor comments below.

pos_accounting/src/pool/storage/operator_impls.rs
azarovh: It is legal. The reason is that decommissioning pool removes only pool balance and data. Pool's owner cannot affect delegations, so they are kept in the db until delegation owner decides to spend it.

pos_accounting/src/storage/in_memory.rs
azarovh: makes sense, will do

pos_accounting/src/pool/storage/operator_impls.rs
```diff
@@ -301,7 +310,10 @@ impl<S: PoSAccountingStorageWrite<T>, T: StorageTag> PoSAccountingDB<S, T> {
 
         self.add_to_delegation_balance(undo_data.delegation_id, undo_data.amount)?;
 
-        self.add_balance_to_pool(pool_id, undo_data.amount)?;
+        // it's possible that the pool was decommissioned
```
TheQuantumPhysicist, line 313: Decommissioning the pool should not prevent people from withdrawing their staked amounts.

pos_accounting/src/storage/in_memory.rs
TheQuantumPhysicist: What's wrong with "if it's not availabe, it's zero"?

azarovh: It's impossible to know if it's "not available because it's zero" or "not available because of some error"

TheQuantumPhysicist: The problem there is that we'll have these zeroes forever. Not a good deal.

What you're talking about is not an error, but an inconsistency. 

So we have to store tons of useless data because we can't do proper arithmetic. I think we can do proper arithmetic checks to ensure that non-existent is zero.

pos_accounting/src/pool/storage/operator_impls.rs
TheQuantumPhysicist: I'm trying to think if there's a corner case that isn't covered where this can be exploited.

pos_accounting/src/pool/tests/simulation_tests.rs
```diff
@@ -0,0 +1,210 @@
...
+fn perform_random_operation(
+    rng: &mut (impl Rng + CryptoRng),
+    op: &mut (impl PoSAccountingOperations + PoSAccountingView),
+    undos: &mut Vec<PoSAccountingUndo>,
+    random_pool: Option<PoolId>,
+    random_delegation: Option<(DelegationId, DelegationData)>,
+    random_delegation_balance: Option<(DelegationId, Amount)>,
+) {
+    match rng.gen_range(0..11) {
+        0 | 1 => {
```
TheQuantumPhysicist, line 151: Please label every arm with a comment on what's going on inside of it.

pos_accounting/src/pool/tests/simulation_tests.rs
```diff
@@ -0,0 +1,210 @@
...
+            let (_, undo) = op.create_pool(&input0_outpoint, pool_data).unwrap();
+            undos.push(undo);
+        }
+        2 => {
+            if let Some(pool_id) = random_pool {
+                let undo = op.decommission_pool(pool_id).unwrap();
+                undos.push(undo);
+            }
+        }
+        3 | 4 => {
```
TheQuantumPhysicist, line 165: You can write `3..=4`... which is more readable.

pos_accounting/src/pool/tests/simulation_tests.rs
```diff
@@ -0,0 +1,210 @@
...
+
+fn perform_random_operation(
+    rng: &mut (impl Rng + CryptoRng),
+    op: &mut (impl PoSAccountingOperations + PoSAccountingView),
+    undos: &mut Vec<PoSAccountingUndo>,
+    random_pool: Option<PoolId>,
+    random_delegation: Option<(DelegationId, DelegationData)>,
+    random_delegation_balance: Option<(DelegationId, Amount)>,
+) {
+    match rng.gen_range(0..11) {
```
TheQuantumPhysicist, line 150: One of my concerns that I've been thinking about is, what if we, in the future, add new operations? How are we gonna remember to add them to the tests?

Here's a solution: There's a crate named `variant_count`. With it, you can do:
```
assert_eq!(PoSAccountingUndo::VARIANT_COUNT, 6);
```

This way, if new ops are added, this assert will fire. With this assert, you can write a comment/message explaining what's going on and why.

TheQuantumPhysicist: This match can be done in a much nicer way... using proper arms and probabilities... but I'm not gonna fixate on that now.

TheQuantumPhysicist (approved):
I'm approving this, but I'm concerned about one thing: Popping things randomly from database basically means order doesn't matter. But is this true?

pos_accounting/src/pool/tests/simulation_tests.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+fn get_random_pool_id(rng: &mut impl Rng, storage: &InMemoryPoSAccounting) -> Option<PoolId> {
+    let all_pool_data = storage.all_pool_data();
+    if !all_pool_data.is_empty() {
+        let (pool_id, _) = all_pool_data.iter().nth(rng.gen_range(0..all_pool_data.len())).unwrap();
+        Some(*pool_id)
+    } else {
+        None
+    }
```
iljakuklic, line 45: Not sure it's much of an improvement but it seems you can use [bool::then](https://github.com/Cameri/nostream) here and a few other functions in this file:
```rust
(!all_pool_data.is_empty()).then(|| all_pool_data.iter().blah...())
```

iljakuklic (approved):
+1

TheQuantumPhysicist:
>> Not sure that I understand your concern, isn't it what happens in production? Users doing random actions. Btw everything is random but `undo` is always for the latest action.  

Random, but they're in order because of utxos.

PR #877, opened 2023-05-05
Mempool reorg test

iljakuklic:
Fairly hacky at places but works. Also added some extra information to reported errors that turned out to be useful for debugging.


test/functional/mempool_basic_reorg.py
```diff
@@ -0,0 +1,160 @@
...
+        # Prepare three transactions, each spending the previous one in sequence
+        (tx1, tx1_id) = make_tx([ reward_input(genesis_id) ], [ 1_000_000 ] )
+        self.log.debug("Encoded tx1 {}: {}".format(tx1_id, tx1))
+        (tx2, tx2_id) = make_tx([ tx_input(tx1_id) ], [ 900_000 ] )
+        self.log.debug("Encoded tx2 {}: {}".format(tx2_id, tx2))
+        (tx3, tx3_id) = make_tx([ tx_input(tx2_id) ], [ 800_000 ] )
+        self.log.debug("Encoded tx3 {}: {}".format(tx3_id, tx3))
+
+        # Submit the first transaction
+        node.p2p_submit_transaction(tx1)
```
TheQuantumPhysicist, line 115: The difference between submitting a transaction through p2p and submitting the transaction directly in the mempool, is that the latter doesn't broadcast. This seems to be superfluous.

test/functional/mempool_basic_reorg.py
```diff
@@ -0,0 +1,160 @@
...
+        assert not node.mempool_contains_tx(tx3_id)
+
+        # Submit the other two transactions
+        node.p2p_submit_transaction(tx2)
+        node.p2p_submit_transaction(tx3)
+        assert not node.mempool_contains_tx(tx1_id)
+        assert node.mempool_contains_tx(tx2_id)
+        assert node.mempool_contains_tx(tx3_id)
+
+        # Submit a block with the other two transacitons
```
TheQuantumPhysicist, line 136: transactions

TheQuantumPhysicist (approved):

test/functional/mempool_basic_reorg.py
iljakuklic: The RPC for that was not available so I added it and changed the mempool functional tests to use it.

muursh (approved):

azarovh (approved):

iljakuklic:
Added an RPC to query the current tip from mempool. It is used in the test to check whether mempool is fully up to date with chainstate. That should make the test way more robust.

stanislav-tkach (approved):

PR #878, opened 2023-05-06
Try to fix p2p tests

pavel-kokolemin:
Try to fix a shutdown deadlock in p2p tests where mempool is making a blocking call but other subsystems are already stopped.


stanislav-tkach (approved):
👍

TheQuantumPhysicist (approved):

PR #880, opened 2023-05-07
Enable block signature verification

### LINKED ISSUE #834: Signing blocks
We should pursue signing blocks. Just signing inputs in PoS is not enough as some components (like, say, block rewards) can be mutated.

### LINKED ISSUE #875: Block signature signature verification
Based on #874 Two open questions:

1. The block verification should be done, in line with proof of stake.
2. Do we want to include the public key of the signature? This has the pro that we can verify the header (and hence the block) easily and add more trust, given that the same key that signs the stake kernel is the key signing the header. Hence, if that's done, then the only verification we need is that the output is valid, which is doable in other ways.

TheQuantumPhysicist:
Closes #834
Closes #875

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
@@ -1,3 +1,18 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use common::chain::{Destination, PoolId, TxOutput};
 use pos_accounting::PoSAccountingView;
 
--- a/chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
@@ -1,0 +1,91 @@
+use common::chain::{Destination, TxOutput};
+use pos_accounting::PoSAccountingView;
+
+use crate::error::ConnectTransactionError;
+
+use super::accounting_delta_adapter::PoSAccountingDeltaAdapter;
+
+pub type SignatureDestinationGetterFn<'a> =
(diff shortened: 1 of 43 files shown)


chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
```diff
@@ -0,0 +1,140 @@
...
+                        // Spending an output of a pool creation transaction is only allowed in a
+                        // context of a transaction (as opposed to block reward) only if this pool
+                        // is being decommissioned.
+                        // If this rule is being invalidated, it will be detected in other parts
+                        // of the code.
+                        Ok(pool_data.decommission_key().clone())
+                    }
+                    TxOutput::ProduceBlockFromStake(_, pool_id) => {
+                        // The only way we can spend this output in a transaction
+                        // (as opposed to block reward), is if we're decommissioning a mempool
```
azarovh, line 81: typo: a pool

azarovh (approved):

chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
```diff
@@ -0,0 +1,140 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use common::chain::{Destination, PoolId, TxOutput};
+use pos_accounting::PoSAccountingView;
+
+use super::accounting_delta_adapter::PoSAccountingDeltaAdapter;
+
+pub type SignatureDestinationGetterFn<'a> =
+    dyn Fn(&TxOutput) -> Result<Destination, SignatureDestinationGetterError> + 'a;
```
iljakuklic, line 22: Seems to me it should be possible to return a reference to destination here, it is always cloned in the implementation below as far as I can see. The lifetimes may be non-trivial though so likely not worth it.

iljakuklic (approved):
+1. No serious objections.

chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
TheQuantumPhysicist: Nope. It's not possible. I already tried that yesterday. There's a destination that's pulled from the database through accounting.

clippy.toml
```diff
@@ -1,3 +1,4 @@
 disallowed-methods = []
 disallowed-types = []
 enforced-import-renames = []
+large-error-threshold = 256
```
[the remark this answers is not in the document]
TheQuantumPhysicist: The problem with boxing errors is that you can't match with them anymore. There's a `box` qualifier for that but it's nightly. I don't see how else to do this.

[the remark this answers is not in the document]
iljakuklic: Hmm ok.

chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
iljakuklic: Hmm I see so `Cow` at best, not worth it.

PR #884, opened 2023-05-10
Salvaged send-request branch

TheQuantumPhysicist:


pavel-kokolemin:
Thanks. Looks like there is only one relevant commit here: [Added a send request functionality and signing to the wallet](https://github.com/mintlayer/mintlayer-core/pull/884/commits/710661988ddda031744fa43d5993f2d3feb219a2)
I moved it here - https://github.com/mintlayer/mintlayer-core/pull/842

PR #885, opened 2023-05-10
New sync manager tests and some fixes

pavel-kokolemin:


p2p/src/sync/peer.rs
```diff
@@ -149,8 +151,8 @@ where
     }
 
     async fn main_loop(&mut self) -> Result<()> {
-        let mut stalling_interval = tokio::time::interval(*self.p2p_config.sync_stalling_timeout);
-        stalling_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+        let mut periodic_check = tokio::time::interval(*self.p2p_config.sync_stalling_timeout);
```
TheQuantumPhysicist, line 154: I find stalling_interval (or `peer_stalling_interval`) a more appropriate name. Is there a reason we're renaming this?

Generic names like `period_check` are very bad as they cannot be understood in generic contexts.

p2p/src/sync/peer.rs
```diff
@@ -256,24 +262,39 @@ where
...
                     }
                 }
+
                 Result::<_>::Ok(())
             })
             .await??;
 
+        // A peer can ignore the headers request if it is in the initial block download state.
+        // Assume this is the case if it asks us for blocks.
+        self.last_activity = None;
```
TheQuantumPhysicist, line 296: In the definition of `last_activity` in the struct, please write why it can be None. Without context it's really hard to guess what's going on. I'd even prefer an enum to handle all that...

PR #886, opened 2023-05-10
Create a facility for creating inclusion proofs of transactions in blocks

TheQuantumPhysicist:
Also some clean up and refactoring

### DIFF
--- a/common/src/chain/block/block_body/merkle_proxy.rs
+++ b/common/src/chain/block/block_body/merkle_proxy.rs
@@ -1,0 +1,171 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use merkletree::{
+    proof::single::{SingleProofHashes, SingleProofNodes},
+    tree::MerkleTree,
+};
+
+use crate::primitives::H256;
+
+use super::{
+    block_merkle::{calculate_tx_merkle_tree, calculate_witness_merkle_tree},
+    merkle_tools::MerkleHasher,
+    BlockBody, BlockMerkleTreeError,
+};
+
+#[must_use]
+pub struct BlockBodyMerkleProxy {
+    witness_tree: WrappedWitnessMerkleTree,
+    tree: WrappedMerkleTree,
+}
+
+impl BlockBodyMerkleProxy {
+    pub fn new(body: &BlockBody) -> Result<Self, BlockMerkleTreeError> {
+        Ok(Self {
+            witness_tree: calculate_witness_merkle_tree(body)?.into(),
+            tree: calculate_tx_merkle_tree(body)?.into(),
+        })
+    }
+
+    pub fn witness_merkle_tree(&self) -> &WrappedWitnessMerkleTree {
+        &self.witness_tree
+    }
+
+    pub fn merkle_tree(&self) -> &WrappedMerkleTree {
+        &self.tree
+    }
+}
+
+/// This struct wrapper, and the other one, are an attempt to create minimal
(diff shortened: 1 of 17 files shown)


common/src/chain/block/block_body/merkle_proxy.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+/// This struct wrapper, and the other one, are an attempt to create minimal
+/// type safety to avoid confusing the two merkle trees, with and without
+/// the witness.
+#[must_use]
+pub struct WrappedMerkleTree {
+    merkle_tree: MerkleTree<H256, MerkleHasher>,
+}
```
iljakuklic, line 120: Looks like this struct and `WrappedWitnessMerkleTree` are identical, including methods. How about having one generic structure `MerkleTreeOf<Something>` where `Something` is just a type tag, stored in `PhantomData`, specifying what sort of data the tree refers to. Then we can have two distinct types `MerkleTreeOf<Transaction>` and `MerkleTreeOf<Witness>` which prevents from them being mixed up. Methods can be implemented in one `impl` block generic over the tag type.

iljakuklic (approved):
Looks good. Some suggestions below.

common/src/chain/block/block_body/merkle_proxy.rs
```diff
@@ -0,0 +1,171 @@
...
+    block_merkle::{calculate_tx_merkle_tree, calculate_witness_merkle_tree},
+    merkle_tools::MerkleHasher,
+    BlockBody, BlockMerkleTreeError,
+};
+
+#[must_use]
+pub struct BlockBodyMerkleProxy {
+    witness_tree: WrappedWitnessMerkleTree,
+    tree: WrappedMerkleTree,
+}
```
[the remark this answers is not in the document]
TheQuantumPhysicist: It used to take a reference to the body and cache the result on call... but then I fell into mutability issues that couldn't be solved without nasty RefCells and copies, so I decided to just evaluate the trees on construction. I was OK with that given that in all verification cases (so far) we always have to evaluate both.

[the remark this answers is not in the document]
TheQuantumPhysicist: About the reasoning, without such a structure, there will be 4 functions that all sound the same... high probability of errors are there. So I thought I'd create such a structure where it's less likely to confuse things.

common/src/chain/block/block_body/merkle_proxy.rs
TheQuantumPhysicist: That's reasonable. But not with `Witness` and `Transaction`. I'll create custom tag-structs.

TheQuantumPhysicist: OK... now I remember why I didn't do that... I wanted the method names to be different so that a copy-paste error doesn't happen...

iljakuklic: > OK... now I remember why I didn't do that... I wanted the method names to be different so that a copy-paste error doesn't happen...

You can sort of do it by introducing separate `impl` blocks for `MerkleTreeOf<Witness>` and `MerkleTreeOf<Transaction>` (or whatever names you end up choosing). That is in addition to the generic `impl` blocks for methods that are shared. You can use the generic block for shared implementation of the methods with different names if needed, just keep it private.

TheQuantumPhysicist: Alrighty. I'll give that a shot. Cheers!

common/src/chain/block/block_body/merkle_proxy.rs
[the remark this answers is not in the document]
iljakuklic: Sorry I may have been a bit unclear. My concern in the original comment was mainly with the naming rather than the mechanism itself. Not sure where the concept of "proxy" comes in. Not that I have any constructive alternatives...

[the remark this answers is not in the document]
TheQuantumPhysicist: > Not that I have any constructive alternatives...

Neither do I... which is why I kept it like this. It's still a proxy in a remote way, given that a user of the struct doesn't see the internals. It doesn't bother me that much anyway. So, if you have a better name, please go ahead and suggest it :smile:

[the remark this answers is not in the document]
muursh: Meh the name's fine for me

muursh (approved):

common/src/chain/block/block_body/merkle_proxy.rs
TheQuantumPhysicist: Done.

Issue #893, opened 2023-05-16
Reconsider TxOutput types for decommissioning and spending delegation shares


azarovh:
Right now to decommission we have to create a tx `StakePool` -> `DecommissionPool`. Though it might seem natural it creates several problems:

1. the necessity to check that `pool_id` is consistent across input/output;
2. the number of combinations of input/output types increases dramatically;
3. `DecommissionPool`  has exactly the same functionality as `LockThenTransfer`  and looks redundant when later used like `DecommissionPool`  -> `Transfer`

The first problem has already snicked undetected into our codebase. So what if we decommission with `StakePool`  -> `LockThenTransfer`?

The same approach can be applied to delegations that will be used in txs like `DelegateStaking`  -> `DelegateStaking`  + `LockThenTransfer`  or `DelegateStaking`  -> `LockThenTransfer`

PR #894, opened 2023-05-16
Delegation creation and spending

### LINKED ISSUE #873: Clean up delegation ids from db
We need a way to clean up unused delegation ids. 
It seems that the best way is to automatically remove them when the pool is already decommissioned and the delegation balance is 0.

azarovh:
This PR introduces 3 new TxOutput types: `CreateDelegationId`, `DelegateStaking` and `SpendShareFromDelegation`.

Here I followed the same design approach as with create/decommission pool but I think we should reconsider it: #893

Some implementation details to pay attention to: 
- `SpendShareFromDelegation` doesn't perform any accounting operation as actual balance decrease is done when spending `DelegateStaking` utxo
- `SpendShareFromDelegation` is timelocked by the value from the chain config
- `DelegateStacking` output can only be spent with the key provided by `CreateDelegationId::spend_destination`
- `DelegateStacking` can be spent partially by producing another `DelegateStacking` for a change 
- Operations of creating delegation and adding balance to it can't be done in a single tx
- There is a TODO to fix mismatched delegation id in input/output (but I think it should be fixed with #893)
- If a pool has been decommissioned and the delegation balance is 0 then delegation id is removed

Closes #873


chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -326,27 +368,89 @@ where
...
+                        )
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            }
+            None => None,
+        };
```
iljakuklic, line 442: I am wondering if the cascade of conditionals here can be simplified somewhat with combinators like `map`, `filter`, `flat_map` and such. The issue I see is the error handling which would force the use of `transpose` and similar making it potentially less readable.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -591,36 +597,54 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+                        self.chain_config.as_ref().spend_share_maturity_distance(block_height);
+                    check(timelock, required, tx.transaction().get_id())
+                }
+                TxOutput::DecommissionPool(_, _, _, timelock) => {
+                    let required = self
+                        .chain_config
+                        .as_ref()
+                        .decommission_pool_maturity_distance(block_height);
+                    check(timelock, required, tx.transaction().get_id())
+                }
```
iljakuklic, line 647: Seems this checks the time lock field adheres to various policies imposed on it by the consensus, not that the actual time lock is satisfied. It confused me a bit at first. Maybe a more expressive name for the function would be better. Something like `check_time_lock_values` or `check_time_lock_minimal_distances` or something like that. Not that I have a great suggestion...

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -311,6 +443,23 @@ mod tests {
         )
     }
 
+    fn create_delegation() -> TxOutput {
+        TxOutput::CreateDelegationId(Destination::AnyoneCanSpend, PoolId::new(H256::zero()))
+    }
+
+    fn delegate_staking() -> TxOutput {
+        TxOutput::DelegateStaking(Amount::ZERO, DelegationId::new(H256::zero()))
+    }
```
iljakuklic, line 452: What do these `H256::zero()` mean here? If it's a valid value here, maybe a special type would be appropriate here, something that behaves like `Option` but the `None` value is internally represented by `H256::zero`?

iljakuklic (commented):
Looking forward for the simplifying follow up. Some initial comments below.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: I tried to play with `map` and `bool::then` but didn't like it. You are right that error handling here becomes a mess

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
azarovh: The actual values don't matter here so I just stubbed them with the simplest. IO policy only cares about types combinations and not values inside

chainstate/test-suite/src/tests/delegation_tests.rs
```diff
@@ -0,0 +1,458 @@
...
+
+        let delegation_spend_destination =
+            Destination::PublicKey(PrivateKey::new_from_rng(&mut rng, KeyKind::Secp256k1Schnorr).1);
+
+        let tx = TransactionBuilder::new()
+            .add_input(transfer_outpoint.into(), empty_witness(&mut rng))
+            .add_output(TxOutput::CreateDelegationId(
+                delegation_spend_destination.clone(),
+                pool_id,
+            ))
```
TheQuantumPhysicist, line 135: Please add an output for change.

chainstate/test-suite/src/tests/delegation_tests.rs
```diff
@@ -0,0 +1,458 @@
...
+            OutPoint::new(OutPointSourceId::Transaction(tx.transaction().get_id()), 1);
+        let transfer_outpoint2 =
+            OutPoint::new(OutPointSourceId::Transaction(tx.transaction().get_id()), 2);
+
+        tf.make_block_builder().add_transaction(tx).build_and_process().unwrap();
+
+        // Try to create 2 delegations in 1 transaction
+        let tx = TransactionBuilder::new()
+            .add_input(transfer_outpoint1.into(), empty_witness(&mut rng))
+            .add_input(transfer_outpoint2.into(), empty_witness(&mut rng))
```
TheQuantumPhysicist, line 230: Why two?

chainstate/test-suite/src/tests/delegation_tests.rs
```diff
@@ -0,0 +1,458 @@
...
+            .add_output(TxOutput::DecommissionPool(
+                Amount::from_atoms(1),
+                Destination::AnyoneCanSpend,
+                pool_id,
+                OutputTimeLock::ForBlockCount(1),
+            ))
+            .build();
+        tf.make_block_builder().add_transaction(tx).build_and_process().unwrap();
+
+        // Spend all delegation share
```
TheQuantumPhysicist, line 431: Please do this in two steps at least. I would love seeing a randomized test that spends chunks until it's empty, then try one more time after that and confirm that it fails.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -159,40 +165,139 @@ pub fn check_tx_inputs_outputs_purposes(
...
+        let valid_outputs = are_outputs_valid_for_tx(outputs);
+        ensure!(
+            valid_outputs,
+            ConnectTransactionError::InvalidOutputTypeInTx
+        );
+    }
+    Ok(())
+}
+
+// single DelegateStaking input; zero or one DelegateStakingOutput + any number of SpendShareFromDelegation
```
TheQuantumPhysicist, line 252: Why do we allow zero? That basically burns the bridge, unless the balance there is zero.

TheQuantumPhysicist: ... balance is zero + pool is decommissioned.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -316,8 +325,41 @@ where
         tx: &Transaction,
     ) -> Result<(), ConnectTransactionError> {
         let input0 = tx.inputs().get(0).ok_or(ConnectTransactionError::MissingOutputOrSpent)?;
+        let mut check_for_delegation_cleanup: Option<DelegationId> = None;
+
+        let inputs_undos = tx
```
TheQuantumPhysicist, line 330: Please explain in every large step what you're doing and why

common/src/chain/config/mod.rs
```diff
@@ -316,6 +316,24 @@ impl ChainConfig {
...
+                let pos_config = match &status {
+                    PoSStatus::Ongoing(config)
+                    | PoSStatus::Threshold {
+                        initial_difficulty: _,
+                        config,
+                    } => config,
+                };
+                pos_config.spend_share_maturity_distance()
+            }
+        }
```
TheQuantumPhysicist, line 334: If you need to specialize for tests, please change ChainConfig in the tests to reflect the values you want through the builder. Having special rules here is a bad idea. We reserve this for implementations or specific chain specifications.

pos_accounting/src/pool/tests/operations_tests.rs
```diff
@@ -340,3 +340,67 @@ fn spend_more_than_delegated(#[case] seed: Seed) {
...
+    {
+        let db = PoSAccountingDB::new(&mut storage);
+        let mut delta = PoSAccountingDelta::new(&db);
+
+        assert_eq!(
+            delta.delete_delegation_id(delegation_id).unwrap_err(),
+            Error::DelegationDeletionFailedPoolStillExists
+        );
+    }
+}
```
TheQuantumPhysicist, line 406: No valid case(s)?

pos_accounting/src/pool/tests/undo_tests.rs
```diff
@@ -15,8 +15,8 @@
 
 use std::collections::BTreeMap;
 
-use common::primitives::Amount;
-use crypto::random::{CryptoRng, Rng};
+use common::{chain::Destination, primitives::Amount};
+use crypto::random::{CryptoRng, Rng, RngCore};
 use rstest::rstest;
 use test_utils::random::{make_seedable_rng, Seed};
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I don't mind reexporting `crypto::random` through `test_utils::random`.

utxo/src/cache.rs
```diff
@@ -476,6 +451,19 @@ impl<P> FlushableUtxoView for UtxosCache<P> {
     }
 }
 
+fn can_be_spent(output: &TxOutput) -> bool {
+    match output {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: In principle yes, but I'd hesitate to do that because that's consensus-related. I'm not sure though.

wallet/types/src/keys.rs
```diff
@@ -97,7 +97,7 @@ impl KeychainUsageState {
     /// Increments the last index used in the blockchain until up_to_last_used. This has no effect
     /// if the up_to_last_used is smaller than the self value
     pub fn increment_up_to_last_used(&mut self, up_to_last_used: ChildNumber) {
-        if self.last_used.is_none() || self.last_used.unwrap() < up_to_last_used {
+        if self.last_used.is_none() || self.last_used.expect("must be present") < up_to_last_used {
```
TheQuantumPhysicist, line 100: I fixed this correctly in Alfie's PR... expect() is not really the right way to do this. Maybe we should pull it from there and put it in a PR of its own... now we have multiple PRs getting in trouble because of this.

TheQuantumPhysicist:
Looking fine. Though I'd appreciate more documentation into this. Also, perhaps combining this with the upcoming changes will make this easier for the eyes. I don't know which should come first.

azarovh:
> Also, perhaps combining this with the upcoming changes will make this easier for the eyes. I don't know which should come first.

I created a PR #902, you can take a look and if you like the approach in general I'll merge it and we can continue combined review here

TheQuantumPhysicist:
> > Also, perhaps combining this with the upcoming changes will make this easier for the eyes. I don't know which should come first.
> 
> I created a PR #902, you can take a look and if you like the approach in general I'll merge it and we can continue combined review here

Very well. I'll review that, then we merge it into this if it looks good. Cheers!

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
azarovh: this allows tx like `DelegateStaking` -> `LockThenTransfer` to completely spend the share without change

chainstate/test-suite/src/tests/delegation_tests.rs
azarovh: why? I don't need it for this test. Or do you suggest adding it as an example of how it works?

chainstate/src/detail/chainstateref/mod.rs
azarovh: renamed to `check_output_maturity_setting`

chainstate/test-suite/src/tests/delegation_tests.rs
azarovh: added 2 steps for now

common/src/chain/config/mod.rs
azarovh: This code specializes for consensus type and is not test-specific per se. Specific values can be set by providing PoSChainConfig struct when constructing ChainConfig if necessary.

P.S. I cleaned up this code a bit take a look at newer version

utxo/src/cache.rs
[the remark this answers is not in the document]
azarovh: I'd move it there once we have another place that will use this code

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -423,55 +423,36 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+                match output {
+                    TxOutput::LockThenTransfer(_, _, tl) => {
+                        let outpoint = OutPoint::new(block.get_id().into(), index as u32);
+                        let required =
+                            block.consensus_data().reward_maturity_distance(self.chain_config);
+                        tx_verifier::timelock_check::check_output_maturity_setting(
+                            tl, required, outpoint,
+                        )
+                        .map_err(CheckBlockError::BlockRewardMaturityError)
+                    }
```
TheQuantumPhysicist, line 442: I'm confused... are we allowing putting a LockThenTransfer in a BlockReward output?

azarovh: for PoW yes but not for PoS. The checks are in `input_output_policy.rs`
This particular function performs isolated output timelock  checks only.

azarovh: what I did is reused a function from tx_verifier

TheQuantumPhysicist: Honestly, I don't like this... I would prefer to have a separate function for each than this unexplainable mess...

If you want to check for maturity, then check for maturity. What bothers me here is that the function claims that it checks for the maturity, then we kind of return an error for other types too.

In my head, the right thing to do is: You match for consensus type, then you call a separate function that does the checking for that type alone, where one is PoW, and checks LockThenTransfer, and the other is PoS, and allows ProduceBlockFromStake with no limitations. This way, if we add new consensus types in the future (say, PoSv2), we can simply add another function with a separate check.

Mixing and matching like this isn't really correct and creates attack vectors in the future when we get more and more consensus types because the number of combinations can grow exponentially.

azarovh: ok I see

chainstate/test-suite/src/tests/delegation_tests.rs
TheQuantumPhysicist: As proof that change won't cause an issue. I would rather have it with and without.

chainstate/test-suite/src/tests/output_timelock.rs
```diff
@@ -61,7 +61,7 @@ fn output_lock_until_height(#[case] seed: Seed) {
         );
 
         // attempt to create the next block, and attempt to spend the locked output
-        assert_eq!(
+        assert!(matches!(
```
TheQuantumPhysicist, line 64: In all these tests, you do have the OutPoint, right? Why not put it in the assert instead of using matches? Please understand that using assert + matches makes debugging much, much harder.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -423,55 +426,49 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
-                    // The output can be reused in block reward right away
-                    Ok(())
+        block
+            .block_reward()
+            .outputs()
+            .iter()
+            .enumerate()
+            .try_for_each(|(index, output)| {
+                match block.consensus_data() {
+                    ConsensusData::None | ConsensusData::PoW(_) => match output {
```
TheQuantumPhysicist, line 437: Do we want to put block reward limitations on ConsensusData::None?

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
```diff
@@ -209,38 +300,55 @@ fn is_valid_any_to_any_combination_for_tx(
...
+            | TxOutput::DelegateStaking(..) => false,
+            TxOutput::CreateDelegationId(..) => true,
+        })
+        .count()
+        < 2;
+
+    valid_outputs_types && is_stake_pool_unique && is_create_delegation_unique
 }
 
 #[cfg(test)]
```
TheQuantumPhysicist, line 354: Can you please move these tests into a submodule?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -316,24 +347,97 @@ where
         tx: &Transaction,
     ) -> Result<(), ConnectTransactionError> {
         let input0 = tx.inputs().get(0).ok_or(ConnectTransactionError::MissingOutputOrSpent)?;
+        let mut check_for_delegation_cleanup: Option<DelegationId> = None;
 
-        let tx_undo = tx
+        // Process tx inputs in terms of pos accounting.
+        // Spending `CreateStakePool`, `ProduceBlockFromStake` or `DelegateStaking` utxos should result in either
+        // decommissioning a pool of spending share in accounting
```
TheQuantumPhysicist, line 354: or?

TheQuantumPhysicist (commented):
.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -343,10 +451,46 @@ where
...
-        if !tx_undo.is_empty() {
+        // delete delegation if the balance is 0 and the pool has been decommissioned
+        let delete_delegation_undo = match check_for_delegation_cleanup {
+            Some(delegation_id) => {
+                let accounting_view = self.accounting_delta_adapter.accounting_delta();
+                let delegation_balance =
+                    accounting_view.get_delegation_balance(delegation_id)?.ok_or(
+                        ConnectTransactionError::DelegationDataNotFound(delegation_id),
+                    )?;
+                if delegation_balance == Amount::ZERO {
```
TheQuantumPhysicist, line 462: Shouldn't we run this same check when a pool is decommissioned, for all the delegations?

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
TheQuantumPhysicist: I did it already.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy.rs
TheQuantumPhysicist: What happens if people attempt to burn the bridge. Meaning, they attempt to spend without the DelegateStaking output while they still have balance?

TheQuantumPhysicist:
Looking good. We have to talk about burning the delegation handle.

TheQuantumPhysicist:
@iljakuklic The simplifications have been merged into this PR, if your'e interested in reviewing.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: good point, thanks

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -315,25 +346,102 @@ where
...
+                        )))
+                    } else {
+                        let res = self
+                            .accounting_delta_adapter
+                            .operations(tx_source)
+                            .create_pool(input0.outpoint(), data.as_ref().clone().into())
+                            .map(|(_, undo)| undo)
+                            .map_err(ConnectTransactionError::PoSAccountingError);
+                        Some(res)
+                    }
```
iljakuklic, line 424: Tiny style nitpick: maybe factor out the `Some` out of the `if` expression:

```rust
if expected != pool_id {
    Some(Err(err_result))
} else {
    let res = result;
    Some(res)
}
```

change to:

```rust
let res = if expected != pool_id {
    Err(err_result)
} else {
    result
};
Some(res)
```

Also the double negation incurs a slight additional cognitive load, I'd go just for an equality condition and switch the branches around:

```rust
let res = if expected == pool_id {
    result
} else {
    Err(err_result)
};
Some(res)
```

Hope I did not make a mistake somewhere

chainstate/src/detail/chainstateref/mod.rs
azarovh: I'd say yes, but the maturity distance is 0

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/tests.rs
```diff
@@ -0,0 +1,721 @@
...
+        None => get_random_outputs_combination(rng, origin_outputs, number_of_outputs),
+    };
+
+    prepare_utxos_and_tx(rng, input_utxos, outputs)
+}
+
+#[rstest]
+    #[rustfmt::skip]
+    #[case(transfer(), transfer(),           Ok(()))]
+    #[case(transfer(), burn(),               Ok(()))]
```
iljakuklic, line 178: Why is the test and all the `#[case]` directives indented? I think the formatter would have fixed it were it not prevented by `#[rustfmt::skip]` but it would also screw up the vertical alignment of the expected results. I suspect a copy-paste error because the module was in the parent file directly before.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/tests.rs
```diff
@@ -0,0 +1,721 @@
...
+        &valid_inputs,
+        &invalid_outputs,
+        Err(ConnectTransactionError::InvalidOutputTypeInTx),
+    );
+}
+
+#[rstest]
+    #[rustfmt::skip]
+    #[case(transfer(), transfer(),           Err(ConnectTransactionError::InvalidInputTypeInReward))]
+    #[case(transfer(), burn(),               Err(ConnectTransactionError::InvalidInputTypeInReward))]
```
iljakuklic, line 531: Extra indentation again.

Btw I just find it's possible to instruct rustfmt to only skip formatting of certain attributes. May be worth giving a shot. https://github.com/rust-lang/rustfmt#tips

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+            TxOutput::Transfer(..)
+            | TxOutput::LockThenTransfer(..)
+            | TxOutput::Burn(..)
+            | TxOutput::CreateStakePool(..)
+            | TxOutput::ProduceBlockFromStake(..)
+            | TxOutput::DelegateStaking(..) => false,
+            TxOutput::CreateDelegationId(..) => true,
+        })
+        .count()
+        < 2;
```
iljakuklic, line 350: Instead of `.count() < 2`, you can use `.at_most_one().is_ok()`. That has the advantage of exiting early if the second matching item is hit. Also avoids the slightly weird formatting on the comparison.

iljakuklic: Also there could be some sort of early exit if one of the conditions fail. One hack to do it would be to turn the conditions into lambdas and change the last line to:

```rust
    valid_outputs_types() && is_stake_pool_unique() && is_create_delegation_unique()
```

Not sure about the readability of that though

iljakuklic: Similar points apply in `is_valid_delegation_spending`

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+    let inputs_utxos = tx
+        .inputs()
+        .iter()
+        .map(|input| {
+            utxo_view
+                .utxo(input.outpoint())
+                .map_err(|_| utxo::Error::ViewRead)?
+                .map(|u| u.output().clone())
+                .ok_or(ConnectTransactionError::MissingOutputOrSpent)
+        })
```
iljakuklic, line 129: This bit of code is repeated here and in `check_reward_inputs_outputs_purposes`. Factor out into a function?

iljakuklic (approved):
Looks sensible to me. Some minor, mostly stylistic, comments below.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/tests.rs
azarovh: Unfortunately "custom inner attributes are unstable"

iljakuklic: Oh, that's a shame. Thanks for checking.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: created an issue for that 
https://github.com/mintlayer/mintlayer-core/issues/909

azarovh:
The mechanics of delegation is messed up and doesn't allow to spend more than was originally delegated (basically leaving all reward inaccessible). The fix is quite massive so I'll do it in a separate branch and would like to merge this one because it's a pain to support it.

PR #895, opened 2023-05-16
Refactor code with more uses of Map's Entry API

OBorce:
Small improvements.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/amounts_map.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/amounts_map.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::BTreeMap;
+use std::collections::{btree_map::Entry, BTreeMap};
 
 use common::primitives::Amount;
 use fallible_iterator::{FallibleIterator, IntoFallibleIterator};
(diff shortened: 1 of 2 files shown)


accounting/src/delta/delta_amount_collection.rs
```diff
@@ -58,28 +61,29 @@ impl<K: Ord> DeltaAmountCollection<K> {
...
+                // if the new amount is zero, no need to have it at all since it has no effect
+                if new_bal == SignedAmount::ZERO {
+                    drop(current.remove_entry());
+                } else {
+                    *current.get_mut() = new_bal;
+                }
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(other_amount);
+            }
```
TheQuantumPhysicist, line 76: This seems to me more unnecessarily verbose.

If you'd like to get rid of that `Option<SignedAmount>` in `combine_signed_amount_delta`, just change `current.get()` to `*current.unwrap_or(&SignedAmount::ZERO)`... not a huge benefit though.

TheQuantumPhysicist (approved):
Please revert that one change I commented on (unless we agree otherwise), then feel free to merge.

accounting/src/delta/delta_amount_collection.rs
OBorce: It was more about replacing the get + remove/insert on the map than the Option but ok.

muursh (approved):

accounting/src/delta/delta_amount_collection.rs
OBorce: reverted

PR #896, opened 2023-05-16
Remove some extra cloning of data

OBorce:


mempool/src/pool/store.rs
```diff
@@ -583,17 +583,15 @@ impl TxMempoolEntry {
     }
 
     pub fn unconfirmed_ancestors_from_parents(
-        parents: BTreeSet<Id<Transaction>>,
+        parents: &BTreeSet<Id<Transaction>>,
         store: &MempoolStore,
     ) -> Result<Ancestors, MempoolPolicyError> {
-        let mut ancestors = BTreeSet::new();
+        let mut ancestors = parents.clone().into();
```
TheQuantumPhysicist, line 589: Why pass parents by reference then clone it? Generally speaking, it's always better to pass it by value so that the caller can decide whether they want to clone.

TheQuantumPhysicist (approved):
Besides that minor comment, looks good. Please fix it then feel free to merge.

mempool/src/pool/store.rs
OBorce: It is cloned for the ancestors, but the parents are then iterated over in the for loop below.
The clone is just to avoid adding the parents one by one in the for loop.
Also can't iterate over the ancestors as they will be modified inside the loop.

TheQuantumPhysicist: I think there's a misunderstanding. The `parents` are passed as a parameter, right? Instead of passing it by reference, just pass it by (mutable) value. Then you can remove this unnecessary clone.

muursh (approved):

PR #902, opened 2023-05-18
Remove DecommissionPool and SpendShareFromPool output types

### LINKED ISSUE #893: Reconsider TxOutput types for decommissioning and spending delegation shares

azarovh:
- Removed `DecommissionPool` and `SpendShareFromPool` from `TxOutput`. From now on `LockThenTransfer` should be used instead. It means that tx `CreateStakePool` -> `LockThenTransfer` will decommission pool and tx `DelegateStaking` -> `LockThenTransfer` will spend share from delegation
- Added `PoolId` to `TxOutput::CreateStakePool` in order to check that block reward cannot misuse pool in StakePool -> ProduceBlockFromStake situation when a staker has 2 pools to stake.
- `check_outputs_timelock` function was moved from ChaninstateRef to TxVerifier because now it requires history to check the lock in txs like `CreateStakePool` -> `LockThenTransfer`

Closes #893

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -21,7 +21,6 @@ mod input_output_policy;
 mod optional_tx_index_cache;
 mod reward_distribution;
 mod signature_check;
-mod timelock_check;
 mod token_issuance_cache;
 mod transferred_amount_check;
 mod tx_index_cache;
@@ -33,6 +32,7 @@ pub mod flush;
 pub mod hierarchy;
 pub mod signature_destination_getter;
 pub mod storage;
+pub mod timelock_check;
 
 mod tx_source;
 pub use tx_source::{TransactionSource, TransactionSourceForConnect};
@@ -567,8 +567,9 @@ where
             &self.storage,
             &self.chain_config,
             &self.utxo_cache,
-            tx_source,
             tx,
+            tx_source,
+            tx.transaction().get_id().into(),
             median_time_past,
         )?;
 
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -66,7 +66,7 @@ use common::{
         signature::Signable,
         signed_transaction::SignedTransaction,
         tokens::{get_tokens_issuance_count, TokenId},
-        Block, ChainConfig, DelegationId, GenBlock, OutPoint, OutPointSourceId, Transaction,
+        Block, ChainConfig, DelegationId, GenBlock, OutPointSourceId, Transaction,
         TxMainChainIndex, TxOutput,
     },
     primitives::{id::WithId, Amount, Id, Idable, H256},
@@ -212,7 +212,7 @@ where
                 TxOutput::Burn(v) => v.coin_amount(),
                 TxOutput::Transfer(_, _)
                 | TxOutput::LockThenTransfer(_, _, _)
-                | TxOutput::CreateStakePool(_)
+                | TxOutput::CreateStakePool(_, _)
                 | TxOutput::ProduceBlockFromStake(_, _)
                 | TxOutput::CreateDelegationId(_, _)
                 | TxOutput::DelegateStaking(_, _) => None,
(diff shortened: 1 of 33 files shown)


chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -89,13 +94,28 @@ where
...
+            | TxOutput::LockThenTransfer(_, _, _)
+            | TxOutput::Burn(_)
+            | TxOutput::CreateDelegationId(_, _) => None,
+            TxOutput::CreateStakePool(_, _) | TxOutput::ProduceBlockFromStake(_, _) => {
+                Some(OutputTimelockCheckRequired::DecommissioningMaturity)
+            }
+            TxOutput::DelegateStaking(_, _) => {
+                Some(OutputTimelockCheckRequired::DelegationSpendMaturity)
+            }
+        };
```
TheQuantumPhysicist, line 117: I don't like these mutations. I think this can be cleaned up in a better way where we don't mutate a value like this. Here's my suggestion:

1. Iter through inputs' outpoints, map, and pull the utxos, and then put them in a `Vec<(Input, Utxo)>`.
2. Run another iter with a map, and calculate `output_check_required`.
3. Do the other checks.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -342,14 +364,26 @@ where
...
+                            Some(res)
+                        }
+                        TxOutput::ProduceBlockFromStake(_, pool_id) => {
+                            let res = self
+                                .accounting_delta_adapter
+                                .operations(tx_source)
+                                .decommission_pool(*pool_id)
+                                .map_err(ConnectTransactionError::PoSAccountingError);
+                            Some(res)
+                        }
```
TheQuantumPhysicist, line 382: Please add comments explaining what you're doing here

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -133,5 +153,60 @@ where
...
+                    }
+                    OutputTimelockCheckRequired::DecommissioningMaturity => {
+                        let required =
+                            chain_config.as_ref().decommission_pool_maturity_distance(block_height);
+                        check_output_timelock(timelock, required)
+                    }
+                },
+            }?;
+        }
+    }
```
TheQuantumPhysicist, line 185: Unrelated blocks of code should either be prefixed with comments or moved into their own functions. We should avoid throwing separable code consecutively after one another... it just leads to confusion. Maybe create a function that does this check, which can be tested individual and has its own explanation of what's going on.

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -133,5 +153,60 @@ where
...
+
     Ok(())
 }
+
+fn check_output_timelock(
```
TheQuantumPhysicist, line 190: This function is kind of exactly done in `ChainstateRef::check_block_reward_timelock()`. Can we somehow unify them?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
TheQuantumPhysicist: To be clear, I don't understand why there's decommission here.

TheQuantumPhysicist (approved):
Besides the comments I mentioned, this looks fine. Let's merge this into the other PR and review them altogether.

PR #905, opened 2023-05-20
Basic QR code interface and support

TheQuantumPhysicist:


pavel-kokolemin (approved):

utils/src/qrcode.rs
```diff
@@ -0,0 +1,248 @@
...
+        let mut result = String::with_capacity(2 * self.side_length() * self.side_length());
+        let border: i32 = border_size as i32;
+        for y in -border..self.side_length() as i32 + border {
+            for x in -border..self.side_length() as i32 + border {
+                let c: char = if self.pixel_or_false(x as usize, y as usize) {
+                    filled_char
+                } else {
+                    empty_char
+                };
+                // We push twice because in a standard terminal, chars are rectangular,
```
azarovh, line 78: Shouldn't default implemetation return char per pixel and all these console specifics be in a separate ConsoleQrCode or something like that?

TheQuantumPhysicist: Maybe... the general case though is available in Vec representation and others. There's no generic string format other than this. This also assumes mono font and things like that. We could rename it to something related to console... that's the best we can do.

azarovh: Maybe it can be a utility function to keep this trait minimalistic. It's just surprising to see console pixel details in a default impl for a trait.

utils/src/qrcode.rs
```diff
@@ -0,0 +1,248 @@
...
+                // not square, with the height being twice the width (9x18 pixels).
+                // This makes the QR code pixels become squares.
+                result.push(c);
+                result.push(c);
```
iljakuklic, line 82: Maybe to make the displayed qr code a bit smaller, we can make use of the half-filled unicode chars "▀", "▄", and display two "pixels" per char vertically. That would be a fair bit more complex to implement though, so probably not worth the effort at the moment.

iljakuklic (approved):
Looks good. At some point, it would be nice to turn various submodules in `utils` into crates (can still live in the utils folder)

utils/src/qrcode.rs
TheQuantumPhysicist: I'll rename the function.

utils/src/qrcode.rs
TheQuantumPhysicist: Yep. That'd be nice. This suffices at the moment.

PR #910, opened 2023-05-23
Fix some TODOs and remove lock_time

TheQuantumPhysicist:
Please review the TODOs I eliminated. Many of them were removed as they're irrelevant or unimportant after having learned more about how everything works together.

### DIFF
--- a/chainstate/test-framework/src/transaction_builder.rs
+++ b/chainstate/test-framework/src/transaction_builder.rs
@@ -24,11 +24,10 @@ use rstest::rstest;
 
 /// The transaction builder.
 pub struct TransactionBuilder {
-    flags: u32,
+    flags: u128,
     inputs: Vec<TxInput>,
     outputs: Vec<TxOutput>,
     witnesses: Vec<InputWitness>,
-    lock_time: u32,
 }
 
 impl TransactionBuilder {
@@ -38,11 +37,10 @@ impl TransactionBuilder {
             inputs: Vec::new(),
             outputs: Vec::new(),
             witnesses: Vec::new(),
-            lock_time: 0,
         }
     }
 
-    pub fn with_flags(mut self, flags: u32) -> Self {
+    pub fn with_flags(mut self, flags: u128) -> Self {
         self.flags = flags;
         self
     }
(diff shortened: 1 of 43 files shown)


common/src/chain/transaction/signed_transaction.rs
```diff
@@ -92,11 +88,6 @@ impl SignedTransaction {
     }
 }
 
-// TODO enforce that inputs size is equal to signatures size when decoding
-
-// TODO make the SignedTransaction serialization ignore the size of the witness vec and just use the size of the inputs
-// NOTE: there might be difficulties there as Encode cannot fail. It may lead to accepting a panic there
```
iljakuklic, line 98: I think it would be nice if the `SignedTransaction` type maintained the invariant internally. Allows for early rejection of obviously malformed transactions before hitting `chainstate`.

iljakuklic (approved):

iljakuklic:
Btw functional tests need updating too

common/src/chain/transaction/signed_transaction.rs
TheQuantumPhysicist: I'm torn here... I'm hesitating to do this decode thing because I'm afraid we'd modify it in the future and forget to update the Decode impl. But maybe I should just get over myself and do it... not sure...

chainstate/test-framework/src/transaction_builder.rs
```diff
@@ -24,11 +24,10 @@ use rstest::rstest;
 
 /// The transaction builder.
 pub struct TransactionBuilder {
-    flags: u32,
+    flags: u128,
```
pavel-kokolemin, line 27: Isn't it better to use compact encoding here? We could save 15 bytes when the flag is 0.

TheQuantumPhysicist: That's exactly what I'm doing if you look at the previous commit. However, there's an issue in Python I'm still working on. Even going from master, changing this to compact and changing Python's TransactionV1 flags scalecodec schema to Compact<u32> makes the mempool tests fail because the transaction IDs are different... still trying to understand wth is going on.

pavel-kokolemin (approved):

common/src/chain/transaction/signed_transaction.rs
TheQuantumPhysicist: Done.

PR #911, opened 2023-05-24
Transaction inputs from accounts

azarovh:
This PR introduces the ability to spend not only from utxos but also from accounts. This is achieved by making `TxInput` an enum of `Utxo` | `Account`.

TODOs:

- [x] verify input/output policy with inputs from accounts;
- [x] check transferred amounts;
- [x] handle transaction fee
- [x] revisit timelock checks;
- [x] handle nonce in chainstate: verify, provide access from db;
- [x] sign tx inputs from accounts; verify signature;
- [x] transaction index
- [x] track double-spends in mempool via txinput nonce;
- [x] revisit tests, add new;

P.S. last commit contains renames in tons of files, I can potentially make it as a separate PR to simplify reviewing process

### DIFF
--- a/chainstate/test-suite/src/tests/pos_processing_tests.rs
+++ b/chainstate/test-suite/src/tests/pos_processing_tests.rs
@@ -46,9 +46,9 @@ use common::{
         stakelock::StakePoolData,
         timelock::OutputTimeLock,
         tokens::OutputValue,
-        AccountOutPoint, AccountSpending, Block, ConsensusUpgrade, Destination, GenBlock, Genesis,
-        NetUpgrades, OutPointSourceId, PoSChainConfig, PoolId, SignedTransaction, TxInput,
-        TxOutput, UpgradeVersion, UtxoOutPoint,
+        AccountNonce, AccountOutPoint, AccountSpending, Block, ConsensusUpgrade, Destination,
+        GenBlock, Genesis, NetUpgrades, OutPointSourceId, PoSChainConfig, PoolId,
+        SignedTransaction, TxInput, TxOutput, UpgradeVersion, UtxoOutPoint,
     },
     primitives::{per_thousand::PerThousand, Amount, BlockHeight, Id, Idable, H256},
     Uint256,
@@ -2098,7 +2098,7 @@ fn spend_from_delegation_with_reward(#[case] seed: Seed) {
 
     let amount_to_withdraw = Amount::from_atoms(rng.gen_range(1..amount_to_delegate.into_atoms()));
     let tx_input_spend_from_delegation = AccountOutPoint::new(
-        0,
+        AccountNonce::new(0),
         AccountSpending::Delegation(delegation_id, amount_to_withdraw),
     );
     let tx = TransactionBuilder::new()
@@ -2162,7 +2162,7 @@ fn spend_from_delegation_with_reward(#[case] seed: Seed) {
     {
         let delegation_balance_overspend = (delegation_balance + Amount::from_atoms(1)).unwrap();
         let tx_input_spend_from_delegation = AccountOutPoint::new(
-            1,
+            AccountNonce::new(1),
             AccountSpending::Delegation(delegation_id, delegation_balance_overspend),
         );
         let tx = TransactionBuilder::new()
@@ -2198,7 +2198,7 @@ fn spend_from_delegation_with_reward(#[case] seed: Seed) {
     }
 
     let tx_input_spend_from_delegation = AccountOutPoint::new(
-        1,
+        AccountNonce::new(1),
         AccountSpending::Delegation(delegation_id, delegation_balance),
     );
     let tx = TransactionBuilder::new()
--- a/chainstate/test-suite/src/tests/pos_processing_tests.rs
+++ b/chainstate/test-suite/src/tests/pos_processing_tests.rs
@@ -46,8 +46,9 @@ use common::{
         stakelock::StakePoolData,
         timelock::OutputTimeLock,
         tokens::OutputValue,
-        Block, ConsensusUpgrade, Destination, GenBlock, Genesis, NetUpgrades, OutPoint,
-        OutPointSourceId, PoSChainConfig, PoolId, SignedTransaction, TxOutput, UpgradeVersion,
+        AccountInput, AccountType, Block, ConsensusUpgrade, Destination, GenBlock, Genesis,
+        NetUpgrades, OutPoint, OutPointSourceId, PoSChainConfig, PoolId, SignedTransaction,
+        TxInput, TxOutput, UpgradeVersion,
     },
     primitives::{per_thousand::PerThousand, Amount, BlockHeight, Id, Idable, H256},
     Uint256,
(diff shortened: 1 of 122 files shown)


common/src/chain/transaction/input.rs
```diff
@@ -122,25 +123,39 @@ impl OutPoint {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
-pub struct TxInput {
-    outpoint: OutPoint,
+pub enum Account {
```
TheQuantumPhysicist, line 126: AccountType

common/src/chain/transaction/input.rs
```diff
@@ -122,25 +123,39 @@ impl OutPoint {
...
+    Pool(PoolId),
```
TheQuantumPhysicist, line 127: We don't need this for pool right now... we may change it later

common/src/chain/transaction/input.rs
```diff
@@ -122,25 +123,39 @@ impl OutPoint {
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+pub struct AccountOutPoint {
+    nonce: u128,
+    account: Account,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+pub enum TxInput {
+    Utxo(OutPoint),
```
TheQuantumPhysicist, line 139: OutPoint -> UtxoOutPoint

common/src/chain/transaction/input.rs
```diff
@@ -122,25 +123,39 @@ impl OutPoint {
...
+    Accounting(AccountOutPoint),
```
TheQuantumPhysicist, line 140: Account(AccountOutPoint)

common/src/chain/transaction/input.rs
```diff
@@ -122,25 +123,39 @@ impl OutPoint {
...
-pub struct TxInput {
-    outpoint: OutPoint,
+pub enum Account {
+    Pool(PoolId),
+    Delegation(DelegationId),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+pub struct AccountOutPoint {
+    nonce: u128,
```
TheQuantumPhysicist, line 133: Compact encoding

azarovh:
Looking at the code again I'm starting to doubt whether making `TxInput` an enum instead of `OutPoint` itself was a good idea.
It could be like this: 
``` rust
enum Outpoint {
    Utxo(UtxoOutPoint),
    Account(AccountOutPoint),
}
```
I see that OutPoint is heavily used in `mempool` and `wallet`,  so when accounts are integrated there we will have to either use `TxInput` everywhere or invent an enum for outpoints. Not sure that `TxInput` is good for use cases outside of `Transaction` context. Thoughts?

common/src/chain/transaction/account_outpoint.rs
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::chain::DelegationId;
+use crate::primitives::Amount;
+use serialization::{Decode, Encode};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+pub enum AccountType {
+    Delegation(DelegationId),
```
TheQuantumPhysicist, line 22: Please add an explicit encode index

common/src/chain/transaction/account_outpoint.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/// Type of OutPoint that represents spending from an account
+#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+pub struct AccountOutPoint {
+    /// An incremental value that represents sequential number of spending from an account.
+    /// It's equivalent to the nonce in Ethereum and helps preserving order of transactions and avoid double-spends.
```
TheQuantumPhysicist, line 29: I think the correct term is "transaction replay" instead of double-spend

chainstate/test-suite/src/tests/pos_processing_tests.rs
```diff
@@ -1743,3 +1745,293 @@ fn pos_stake_testnet_genesis(#[case] seed: Seed) {
...
+        .add_output(TxOutput::DelegateStaking(amount_to_delegate, delegation_id))
+        .build();
+
+    tf.make_block_builder()
+        .with_transactions(vec![tx1, tx2])
+        .build_and_process()
+        .unwrap();
+    tf.progress_time_seconds_since_epoch(target_block_time);
+
+    // Process block_2 and distibute reward
```
TheQuantumPhysicist, line 1830: distibute

test/functional/test_framework/__init__.py
```diff
@@ -58,14 +58,22 @@ def init_mintlayer_types():
...
             },
 
-            "TxInput": {
+            "OutPoint": {
                 "type": "struct",
                 "type_mapping": [
                     ["id", "OutPointSourceId"],
                     ["index", "u32"],
                 ]
             },
```
TheQuantumPhysicist, line 68: There's a "Nono" instead of None typo in this file. Please fix it.

utxo/src/cache.rs
```diff
@@ -163,18 +165,22 @@ impl<P: UtxosView> UtxosCache<P> {
...
-        for (tx_in, utxo) in tx.inputs().iter().zip(tx_undo.into_inner().into_iter()) {
-            self.add_utxo(tx_in.outpoint(), utxo, false)?;
-        }
-        Ok(())
+        tx.inputs()
+            .iter()
+            .zip(tx_undo.into_inner().into_iter())
+            .filter_map(|(tx_in, utxo)| match tx_in {
+                TxInput::Utxo(outpoint) => Some((outpoint, utxo)),
+                TxInput::Account(_) => None,
```
TheQuantumPhysicist, line 181: I'm still not done with the review. But do we have this covered? Just keeping this here as a reminder. We need good tests for reversibility.

azarovh: nope and it doesn't work correctly, I'm on it

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -112,6 +113,8 @@ pub struct TransactionVerifier<C, S, U, A> {
 
     accounting_delta_adapter: PoSAccountingDeltaAdapter<A>,
     accounting_block_undo: AccountingBlockUndoCache,
+
+    account_nonce: BTreeMap<AccountType, Option<u128>>,
```
TheQuantumPhysicist, line 117: If the Option in the value is used to indicate deletion (as from my understanding in the code), please use CachedOperation instead of Option.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -96,6 +96,7 @@ pub struct TransactionVerifierDelta {
     accounting_delta: PoSAccountingDeltaData,
     accounting_delta_undo: BTreeMap<TransactionSource, AccountingBlockUndoEntry>,
     accounting_block_deltas: BTreeMap<TransactionSource, PoSAccountingDeltaData>,
+    account_nonce: BTreeMap<AccountType, Option<u128>>,
```
TheQuantumPhysicist, line 99: CachedOperation

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -341,100 +347,145 @@ where
...
+        &mut self,
+        tx_source: TransactionSource,
+        account_input: &AccountOutPoint,
+    ) -> Result<PoSAccountingUndo, ConnectTransactionError> {
+        let account = *account_input.account();
+        // Check that account nonce increments previous value
+        let current_nonce = self
+            .get_account_nonce_count(account)
+            .map_err(|_| ConnectTransactionError::TxVerifierStorage)?;
+        let expected_nonce = current_nonce.map_or(0, |nonce| nonce + 1);
```
TheQuantumPhysicist, line 360: Why define another variable with a new name?

pavel-kokolemin:
> I see that OutPoint is heavily used in mempool and wallet, so when accounts are integrated there we will have to either use TxInput everywhere or invent an enum for outpoints. Not sure that TxInput is good for use cases outside of Transaction context. Thoughts?

I think it's fine and there's no need to change it.

chainstate/tx-verifier/src/transaction_verifier/error.rs
```diff
@@ -143,10 +145,18 @@ pub enum ConnectTransactionError {
...
+    #[error("Nonce is not incremental: {0:?}")]
+    NonceIsNotIncremental(AccountType),
+    #[error("Nonce is not found: {0:?}")]
+    MissingTransactionNonce(AccountType),
     #[error(
         "Transaction {0} has not enough pledge to create a stake pool: giver {1:?}, required {2:?}"
     )]
     NotEnoughPledgeToCreateStakePool(Id<Transaction>, Amount, Amount),
+    #[error("Attempt to create stake pool from accounting inputs")]
+    AttemptToCreateStakePoolFromAccounts,
```
TheQuantumPhysicist, line 157: More rules... more complexity.

I'll come back to this later when we're done with the VIU problem.

chainstate/tx-verifier/src/transaction_verifier/signature_check.rs
```diff
@@ -40,24 +40,25 @@ where
...
-            let outpoint = input.outpoint();
-            utxo_view
+        .map(|input| match input {
+            TxInput::Utxo(outpoint) => utxo_view
                 .utxo(outpoint)
                 .map_err(|_| utxo::Error::ViewRead)?
                 .ok_or(ConnectTransactionError::MissingOutputOrSpent)
-                .map(|utxo| utxo.take_output())
+                .map(|utxo| Some(utxo.take_output())),
+            TxInput::Account(_) => Ok(None),
```
TheQuantumPhysicist, line 49: No signature verification for accounts?

azarovh: This code just collects utxos for verification, all inputs are checked below

common/src/chain/transaction/signature/mod.rs
```diff
@@ -188,7 +188,7 @@ fn verify_standard_input_signature<T: Transactable>(
     outpoint_destination: &Destination,
     witness: &StandardInputSignature,
     tx: &T,
-    inputs_utxos: &[&TxOutput],
+    inputs_utxos: &[Option<&TxOutput>],
```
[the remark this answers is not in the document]
azarovh: we don't actually need to, because acount input already contains amount

TheQuantumPhysicist (commented):
Done with first round. Please fix the comments and I'll come back to it.

utxo/src/cache.rs
azarovh: fixed disconnect for utxo, added test for that

TheQuantumPhysicist (approved):

iljakuklic (commented):
Looks good. Some comments below. Also I have some concerns about the overall complexity levels the transaction verifier is reaching.

common/src/chain/transaction/utxo_outpoint.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+}
+
+impl Ord for OutPointSourceId {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        let id = self.outpoint_source_id_as_monolithic_tuple();
+        let other_id = other.outpoint_source_id_as_monolithic_tuple();
+        id.cmp(&other_id)
+    }
+}
```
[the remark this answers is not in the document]
azarovh: I'd say yes but I just moved this code so maybe there were some reasoning. Sam?

common/src/chain/transaction/signature/sighash/mod.rs
```diff
@@ -40,7 +40,7 @@ fn hash_encoded_if_some<T: Encode>(val: &Option<T>, stream: &mut DefaultHashAlgo
 
 fn stream_signature_hash<T: Signable>(
     tx: &T,
-    inputs_utxos: &[&TxOutput],
+    inputs_utxos: &[Option<&TxOutput>],
```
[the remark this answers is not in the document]
iljakuklic: Ok. I might have gotten confused a bit. The `TxOuptut` here identifies what's being spent. In case of accounts, the account ID identifies what's being spent. I guess the account ID is already being committed to in the input, so that's fine.

However it also means that the signature commits to the account nonce for other inputs. I am not sure if that's the right thing to do, quite hard to work out all the implications. Does Ethereum do it? Or should the nonce be in witness? My instinct is that the current approach is correct because it basically emulates a sequence of single-input single-ouptut UTXOs but not 100% convinced. Thoughts? @TheQuantumPhysicist

wallet/src/account/output_cache/mod.rs
```diff
@@ -69,21 +76,28 @@ impl OutputCache {
...
         if let Some(tx) = tx_opt {
             for input in tx.inputs() {
-                self.consumed.remove(input.outpoint());
+                match input {
+                    TxInput::Utxo(outpoint) => {
+                        self.consumed.remove(outpoint);
+                    }
+                    TxInput::Account(_, _) => {
+                        unimplemented!()
+                    }
```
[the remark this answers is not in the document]
iljakuklic: Ok. Maybe put a string in the `unimplemented!()` macro for documentation / better crash diagnosis

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -171,14 +160,38 @@ where
...
+                        }
+                    }
+                }
+                TxInput::Account(account_input) => match account_input.account() {
+                    AccountSpending::Delegation(_, _) => {
+                        Some(OutputTimelockCheckRequired::DelegationSpendMaturity)
+                    }
+                },
+            },
+        );
```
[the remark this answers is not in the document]
azarovh: good point, added collection of all requirements and an error in case there are multiple

common/src/chain/transaction/signature/sighash/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I can see an issue that a long living transaction won't be possible with this approach, but I don't see another issue with it. To be fair, ethereum has this issue too. The nonce can be used to invalidate or stall transactions.

Ethereum has one nonce per transaction, because the nonce binds to the "from" field, and the signature commits to all that.

While the nonce being in the witness sounds nice, I can see it introducing all kinds of complications. On one hand, the nonce is part of the signature, not part of the spendee; that's because, intiuitively, signing something should increment the nonce, in a perfect world. On the other hand, even if the nonce is in the witness, it should also be part of the witness commitment. So we'll have a circular problem where we have part of the witness data that should be "constrained" by the witness signature. What a mess!

common/src/chain/transaction/utxo_outpoint.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: This is done out of caution because I couldn't verify this from the standard.

common/src/chain/transaction/signature/tests/sign_and_mutate.rs
```diff
@@ -992,12 +961,42 @@ fn add_input(_rng: &mut impl Rng, tx: &SignedTransactionWithUtxo) -> SignedTrans
...
+                TxInput::Account(AccountOutPoint::new(
+                    outpoint.nonce(),
+                    AccountSpending::Delegation(
+                        DelegationId::new(H256::random_using(rng)),
+                        Amount::from_atoms(rng.gen()),
+                    ),
+                ))
+            } else {
+                TxInput::Account(AccountOutPoint::new(
+                    AccountNonce::new(rng.gen()),
```
TheQuantumPhysicist, line 992: Keep in mind that there's a chance the rng provides the same value, and then the test fails.

common/src/chain/transaction/account_nonce.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn new(nonce: u64) -> Self {
+        Self(nonce)
+    }
+
+    pub fn value(&self) -> u64 {
+        self.0
+    }
+
+    pub fn increment(self) -> Option<Self> {
+        (self.0 < u64::MAX).then(|| AccountNonce::new(self.0 + 1))
```
TheQuantumPhysicist, line 34: How about `self.0.checked_add(1).map(AccountNonce::new)`

TheQuantumPhysicist (approved):
Please check the minor things I mentioned, add the skipping nonce we talked about, then feel free to merge if no one else has objections.

common/src/chain/transaction/utxo_outpoint.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: If there's a test that proves the order is correct, then it's OK to remove it... I know I'm overly paranoid... I could let it go.

common/src/chain/transaction/signature/tests/utils.rs
```diff
@@ -142,24 +157,36 @@ pub fn generate_and_sign_tx(
     chain_config: &ChainConfig,
     rng: &mut (impl Rng + CryptoRng),
     destination: &Destination,
-    inputs_utxos: &[&TxOutput],
+    inputs_utxos: &[Option<TxOutput>],
```
iljakuklic, line 160: Why not `&[Option<&TxOutput>]`?

EDIT: You know what, forget it, this is test utils...

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -171,20 +161,53 @@ where
...
+                },
+            })
+            .collect::<Vec<_>>();
+
+        // Note: theoretically there could be several inputs that impose timelock requirement
+        // but for now such transaction is considered invalid
+        ensure!(
+            output_checks_required.len() < 2,
+            ConnectTransactionError::InvalidInputTypeInTx
+        );
```
iljakuklic, line 203: Can use `at_most_one` from itertools here.

iljakuklic (approved):
Good enough to be merged imo.

utxo/src/cache.rs
```diff
@@ -140,16 +140,23 @@ impl<P: UtxosView> UtxosCache<P> {
...
+            .collect();
+
+        let utxos = tx
+            .inputs()
+            .iter()
+            .map(|input| match input {
+                TxInput::Utxo(outpoint) => self.spend_utxo(outpoint).map(Some),
+                TxInput::Account(_) => Ok(None),
+            })
+            .collect::<Result<Vec<_>, Error>>()?;
```
ImplOfAnImpl, line 159: Is this change only needed to avoid the intermediate `collect`? Asking this because I've also solved this in my mini-pr https://github.com/mintlayer/mintlayer-core/pull/944 via `itertools::process_results`:
```
let (sources, utxos) = itertools::process_results(
    tx.inputs().iter().map(|tx_in| {
        let utxo = self.spend_utxo(tx_in.outpoint())?;
        Ok((tx_in.outpoint().tx_id(), utxo))
    }),
    |iter| iter.unzip(),
)?;
```
Now I'm wondering which version I should keep.
(I suppose `process_results` should be better, because it iterates over the collection only once and it shouldn't impose any noticeable overhead by itself).

azarovh: The logic has changed. Now `input` has optional outpoint. So `sources` and `utxos` collections can have different length. I removed zipping and split the code into two loops to account for that.

PR #912, opened 2023-05-24
Fix block reward in block production

alfiedotwtf:
Before this PR, `Destination` was a parameter to the `generate_block()` RPC call. But as PoS blocks no longer need this parameter, I've move it into the PoW's `input_data` enum.

So from here, since we need to look at `ConsenusData` fields for PoS when generating the block reward, I've moved creating the `BlockReward` into `consensus/src/po[sw]/input_data.rs` for convenience.

### DIFF
--- a/blockprod/src/interface/blockprod_interface.rs
+++ b/blockprod/src/interface/blockprod_interface.rs
@@ -32,7 +32,7 @@ pub trait BlockProductionInterface: Send {
     /// generated with available transactions in the mempool
     async fn generate_block(
         &mut self,
-        input_data: Option<GenerateBlockInputData>,
+        input_data: GenerateBlockInputData,
         transactions: Option<Vec<SignedTransaction>>,
     ) -> Result<Block, BlockProductionError>;
 }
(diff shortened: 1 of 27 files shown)


blockprod/src/interface/blockprod_interface.rs
```diff
@@ -33,7 +33,6 @@ pub trait BlockProductionInterface: Send {
     async fn generate_block(
         &mut self,
         input_data: Option<GenerateBlockInputData>,
```
TheQuantumPhysicist, line 35: This shouldn't be optional anymore, right?

alfiedotwtf: I think it would have to be `None` for `RequiredConsensus::IgnoreConsensus` though

TheQuantumPhysicist: Well, then why not add a third arm in `GenerateBlockInputData` with index 0 to cover that? It'll have the same effect of `None`

alfiedotwtf: Of course. Cool, doing that now

TheQuantumPhysicist (approved):

pavel-kokolemin (approved):

PR #913, opened 2023-05-24
Wallet staking

pavel-kokolemin:


wallet/src/account/txo_cache/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn outpoints(&self) -> &BTreeSet<OutPoint> {
+        &self.consumed
+    }
+
+    pub fn add_block(&mut self, block_height: AccountBlockHeight, block: WalletBlock) {
+        for input in block.kernel_inputs().iter() {
+            self.consumed.insert(input.outpoint().clone());
+        }
+        self.blocks.insert(block_height, block);
+    }
```
alfiedotwtf, line 72: Should we also call `add_tx()` for every transaction within `block` here?

Edit: ah, looking down, we're adding kernel inputs. All good, ignore this comment.

alfiedotwtf (approved):

common/src/chain/config/builder.rs
```diff
@@ -55,7 +56,23 @@ impl ChainType {
...
+                let pos_config = crate::chain::create_unittest_pos_config();
+                let upgrades = vec![
+                    (
+                        BlockHeight::new(0),
+                        UpgradeVersion::ConsensusUpgrade(ConsensusUpgrade::IgnoreConsensus),
+                    ),
+                    (
+                        BlockHeight::new(1),
+                        UpgradeVersion::ConsensusUpgrade(ConsensusUpgrade::PoS {
+                            initial_difficulty: crate::chain::pow::limit(ChainType::Testnet).into(),
```
alfiedotwtf, line 69: I just noticed that this is `pow`. Should we instead set this in `crate::chain::pos` too so that PoS isn't relying on something defined in PoW code?

blockprod/src/detail/mod.rs
```diff
@@ -210,42 +214,30 @@ impl BlockProduction {
         Ok(consensus_data)
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn generate_finalize_block_data(
         chain_config: &ChainConfig,
         chainstate_handle: &(dyn ChainstateInterface),
         best_block_index: &GenBlockIndex,
         block_height: BlockHeight,
+        current_timestamp: BlockTimestamp,
```
TheQuantumPhysicist, line 223: Why can't we extract this from the best block index?

blockprod/src/detail/mod.rs
```diff
@@ -328,13 +321,33 @@ impl BlockProduction {
         let (job_remover_func, end_confirm_receiver) = self.job_manager.make_job_stopper_function();
         let _job_remover = OnceDestructor::new(move || job_remover_func(job_key));
 
+        let mut previous_attempt = timestamp_at_start;
```
TheQuantumPhysicist, line 324: "last_timestamp_used"

Also please explain above this the idea behind it. Something in the lines of "Proof of stake depends on timestamp when searching for a valid block. The timestamp starts from the best/previous block's timestamp + 1 second, and ends at the current time stamp + some distance in time defined by the blockchain. This variable keeps track of the latest time that was checked, so that it's not tested multiple times."

blockprod/src/detail/mod.rs
```diff
@@ -328,13 +321,33 @@ impl BlockProduction {
...
 
+            match previous_consensus_status {
+                RequiredConsensus::IgnoreConsensus | RequiredConsensus::PoW(_) => {}
+                RequiredConsensus::PoS(_) => {
+                    if previous_attempt == timestamp && timestamp != timestamp_at_start {
+                        sleep(Duration::from_secs(1)).await;
+                        continue;
+                    }
+                }
+            }
```
TheQuantumPhysicist, line 340: If I understand correctly, this will result in producing blocks blocking forever until a block is found. I'm not particularly against this, but I think this can be engineered in a better way.

My concern here is that something is missing in this calculation, and the abstraction is not properly done. Consider the following: Let's call the current timestamp at start T0, and the maximum timestamp in the future at start T1, and the timestamp of the previous block TB. The first scan goes over the range `[TB,T1]`. But then... it seems that with this setting, we'll save `T0+t` all the time, where `t` is the time elapsed since we started staking the first time, and we'll "try to stake" again repeatedly over a certain range. Now the problem is that there's a range that's always re-checked in this mechanism, and this range is `[T0+t, T1+t]`, while in reality, we want to check only `[T1,T1+t]`, because `[TB,T1]` has already been covered. So this solution, while it works, it kind creates an efficiency issue.

Not that we have to fix this right now... but this has to be addressed sooner or later. The fix should have an object that takes all these considerations into account and updates itself such that this doesn't happen.

blockprod/src/detail/mod.rs
```diff
@@ -328,13 +321,33 @@ impl BlockProduction {
         let (job_remover_func, end_confirm_receiver) = self.job_manager.make_job_stopper_function();
         let _job_remover = OnceDestructor::new(move || job_remover_func(job_key));
 
+        let mut previous_attempt = timestamp_at_start;
+        let mut previous_consensus_status =
```
TheQuantumPhysicist, line 325: Why do we need to keep updating this if we're targeting one, specific previous block?

common/src/chain/config/builder.rs
TheQuantumPhysicist: There should be a separate limit for pos, granted.

common/src/chain/config/mod.rs
```diff
@@ -452,6 +456,60 @@ fn create_mainnet_genesis() -> Genesis {
...
+    let decommission_pub_key = decode_hex::<PublicKey>(
+        "0002ea30f3bb179c58022dcf2f4fd2c88685695f9532d6a9dd071da8d7ac1fe91a7d",
+    );
+    let staker_pub_key = decode_hex::<PublicKey>(
+        "0002884adf48b0b32ab3d66e1a8b46576dfacca5dd25b66603650de792de4dd2e483",
+    );
+
+    let vrf_pub_key = decode_hex::<VRFPublicKey>(
+        "00d4b8138be57c1bddfff397ffc72ca0453131f8df364d2b6bf4a8c6558ed81d79",
+    );
```
TheQuantumPhysicist, line 487: It would be nice to have an unused function that generates these values so that we can generate them on the go to boot the testnet when needed.

common/src/chain/pos.rs
```diff
@@ -95,6 +95,17 @@ impl PoSChainConfig {
     }
 }
 
+pub fn create_testnet_pos_config() -> PoSChainConfig {
+    PoSChainConfig {
```
TheQuantumPhysicist, line 99: The limits on difficulty swings should be part of this struct

crypto/src/key/mod.rs
```diff
@@ -89,6 +89,13 @@ impl PrivateKey {
         };
         Ok(signature)
     }
+
+    /// Returns internal private key data (32-byte slice for Secp256k1Schnorr)
+    pub fn as_bytes(&self) -> &[u8] {
+        match &self.key {
+            PrivateKeyHolder::Secp256k1Schnorr(key) => key.as_bytes(),
+        }
+    }
```
TheQuantumPhysicist, line 98: Why? This is wrong. We should strictly use the Encode trait if we need to serialize. The key details should be opaque.

wallet/src/account/mod.rs
```diff
@@ -128,6 +147,93 @@ impl Account {
         Ok(tx)
     }
 
+    fn get_vrf_key(&self) -> WalletResult<(VRFPrivateKey, VRFPublicKey)> {
+        let vrf_key_path = make_path_to_vrf_key(&self.chain_config, self.account_index());
+        let private_key = self.key_chain.get_private_key_for_path(&vrf_key_path)?.private_key();
+        let vrf_keys =
+            VRFPrivateKey::new_from_bytes(private_key.as_bytes(), VRFKeyKind::Schnorrkel)
+                .expect("should not fail because private keys are always 32 bytes");
```
TheQuantumPhysicist, line 155: This transformation is incorrect and can lead to issues. Not all points on a curve are valid in another. We should allow storing VRF keys properly.

crypto/src/vrf/schnorrkel/mod.rs
```diff
@@ -103,6 +103,23 @@ impl SchnorrkelPrivateKey {
...
+        let keypair = mini_secret.expand_to_keypair(schnorrkel::ExpansionMode::Uniform);
+        Ok((
+            SchnorrkelPrivateKey {
+                key: keypair.secret.clone(),
+            },
+            SchnorrkelPublicKey {
+                key: keypair.public,
+            },
+        ))
+    }
```
TheQuantumPhysicist, line 121: This has to go.

crypto/src/vrf/mod.rs
```diff
@@ -87,6 +89,25 @@ impl VRFPrivateKey {
...
+                    VRFPrivateKey {
+                        key: VRFPrivateKeyHolder::Schnorrkel(k.0),
+                    },
+                    VRFPublicKey {
+                        pub_key: VRFPublicKeyHolder::Schnorrkel(k.1),
+                    },
+                ))
+            }
+        }
+    }
```
TheQuantumPhysicist, line 109: This has to goo too.

wallet/wallet-controller/src/lib.rs
```diff
@@ -125,7 +129,16 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
         self.wallet
-            .get_balance(DEFAULT_ACCOUNT_INDEX)
+            .get_balance(
+                DEFAULT_ACCOUNT_INDEX,
+                UtxoType::Transfer | UtxoType::LockThenTransfer,
+            )
+            .map_err(ControllerError::WalletError)
+    }
+
+    pub fn get_utxos(&self) -> Result<BTreeMap<OutPoint, TxOutput>, ControllerError<T>> {
```
TheQuantumPhysicist, line 139: get_all_utxos(), since we do have a filter?

Even better, keep the name, and pass UtxosTypes as parameter. There can be Default impl for UtxosTypes that's equivalent to UtxosTypes::ALL.

wallet/wallet-controller/src/lib.rs
```diff
@@ -151,9 +164,43 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+    }
+
     /// Sync the wallet block chain from the node.
     /// This function is cancel safe.
     pub async fn run_sync(&mut self) {
-        self.block_sync.run(&mut self.wallet).await;
+        self.block_sync.run(&mut self.wallet, None).await;
+    }
+
+    pub async fn sync_once(&mut self) -> Result<(), ControllerError<T>> {
```
TheQuantumPhysicist, line 197: Please add documentation to this function as to what it does.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -70,14 +69,7 @@ pub enum WalletCommand {
...
 
-    /// Generate a block with the given transactions to the specified
-    /// reward destination. If transactions are None, the block will be
-    /// generated with available transactions in the mempool
-    GenerateBlock {
-        reward_destination: HexEncoded<Destination>,
-
-        transactions: Option<Vec<HexEncoded<SignedTransaction>>>,
-    },
+    GenerateBlock,
```
TheQuantumPhysicist, line 72: No more docs?

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -295,16 +295,13 @@ pub async fn handle_wallet_command(
...
 
-        WalletCommand::GenerateBlock {
-            reward_destination: _,
-            transactions,
-        } => {
-            let transactions =
-                transactions.map(|txs| txs.into_iter().map(HexEncoded::take).collect());
-            let block = rpc_client
-                .generate_block(GenerateBlockInputData::None, transactions)
+        WalletCommand::GenerateBlock => {
```
TheQuantumPhysicist, line 298: Does the user have to keep calling this to create blocks one-by-one? I'm not done reviewing yet, but if yes, then we need some kind of an event loop that automates this. The user should start this once and then the blocks get generated automatically.

wallet/types/src/wallet_tx.rs
```diff
@@ -15,24 +15,15 @@
...
-    /// Unconfirmed transaction in the mempool
-    #[codec(index = 1)]
-    InMempool,
-    /// Conflicted transaction with a confirmed block
-    #[codec(index = 2)]
-    Conflicted(Id<GenBlock>),
-    /// Transaction that is not confirmed or conflicted and is not in the mempool.
-    #[codec(index = 3)]
-    Inactive,
+    Confirmed(BlockHeight),
```
TheQuantumPhysicist, line 26: Please explain why the other possibilities were removed. These are states that can happen.

wallet/types/src/wallet_block.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use serialization::{Decode, Encode};
+
+use common::{
+    chain::{block::ConsensusData, Block, GenBlock, Genesis, TxInput, TxOutput},
+    primitives::{BlockHeight, Id, Idable},
+};
+
+#[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
+pub struct WalletBlock {
```
TheQuantumPhysicist, line 24: Please add documentation, explaining why this struct even needs to exist and what problem it solves.

wallet/storage/src/lib.rs
```diff
@@ -69,8 +72,12 @@ pub trait WalletStorageRead {
 pub trait WalletStorageWrite: WalletStorageRead {
     /// Set storage version
     fn set_storage_version(&mut self, version: u32) -> Result<()>;
-    fn set_utxo(&mut self, outpoint: &AccountOutPointId, entry: Utxo) -> Result<()>;
-    fn del_utxo(&mut self, outpoint: &AccountOutPointId) -> Result<()>;
+    fn set_block(
+        &mut self,
+        block_height: &AccountBlockHeight,
+        block: &WalletBlock,
+    ) -> crate::Result<()>;
```
TheQuantumPhysicist, line 79: Why do we need to store the whole block in the wallet?!!!! That's crazy and should be avoided at all cost as it can blow up the db storage size! This information should strictly be stored in the node and retrieved from the node.

wallet/storage/src/internal/test.rs
```diff
@@ -35,36 +27,3 @@ fn storage_get_default_version_in_tx() {
         assert_eq!(vtx, vst, "Transaction and non-transaction inconsistency");
     })
 }
-
-#[cfg(not(loom))]
-#[rstest]
-#[trace]
-#[case(Seed::from_entropy())]
-fn read_write_utxo_in_db_transaction(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 43: Why are we deleting this test?

wallet/storage/src/lib.rs
TheQuantumPhysicist, line 79: Why can't we retrieve this information from the node? This can be too much for a wallet.

wallet/storage/src/internal/mod.rs
```diff
@@ -118,8 +118,8 @@ impl<B: storage::Backend> WalletStorageRead for Store<B> {
 impl<B: storage::Backend> WalletStorageWrite for Store<B> {
     delegate_to_transaction! {
         fn set_storage_version(&mut self, version: u32) -> crate::Result<()>;
-        fn set_utxo(&mut self, outpoint: &AccountOutPointId, entry: Utxo) -> crate::Result<()>;
-        fn del_utxo(&mut self, outpoint: &AccountOutPointId) -> crate::Result<()>;
+        fn set_block(&mut self, block_height: &AccountBlockHeight, block: &WalletBlock) -> crate::Result<()>;
```
TheQuantumPhysicist, line 121: This probably needs a better name. Just like `WalletBlock`.

wallet/src/account/txo_cache/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+use common::{
+    chain::{OutPoint, OutPointSourceId, TxOutput},
+    primitives::Idable,
+};
+use wallet_types::{
+    account_id::AccountBlockHeight, wallet_block::WalletBlock, AccountTxId, WalletTx,
+};
+
+pub struct TxoCache {
```
TheQuantumPhysicist, line 26: Please explain why this was created and what purpose it serves.

Also... why call it TxoCache instead of OutputCache?

wallet/src/account/mod.rs
```diff
@@ -128,6 +147,93 @@ impl Account {
...
+        )?;
+
+        let request = SendRequest::new().with_inputs(utxos).with_outputs([stake_output]);
+
+        self.process_send_request(db_tx, request)
+    }
+
+    pub fn get_pos_gen_block_data<B: storage::Backend>(
+        &self,
+        _db_tx: &StoreTxRo<B>,
```
TheQuantumPhysicist, line 198: Do we need this?

wallet/src/account/mod.rs
```diff
@@ -301,75 +379,155 @@ impl Account {
         })
     }
 
+    fn mark_outputs_as_used<B: storage::Backend>(
```
TheQuantumPhysicist, line 382: Used here probably doesn't make sense. Correct me if I'm wrong.

"Used" is for keys in the context of BIP-39, where a key is "used" if there's a transaction that it can spend on the blockchain.

"Spend", well... it's for outputs. We all know what it means.

I think there could be confusion in the naming here. Am I getting this right?

TheQuantumPhysicist:
Great work, guys! There are a few things we need to work on to finalize this.

common/src/chain/config/builder.rs
pavel-kokolemin: Should be fixed now.

common/src/chain/config/mod.rs
```diff
@@ -452,6 +457,60 @@ fn create_mainnet_genesis() -> Genesis {
...
+
+    let initial_pool = TxOutput::CreateStakePool(
+        H256::zero().into(),
+        Box::new(StakePoolData::new(
+            initial_pool_amount,
+            Destination::PublicKey(staker_pub_key),
+            vrf_pub_key,
+            Destination::PublicKey(decommission_pub_key),
+            PerThousand::new(10).expect("must be valid"),
+            (COIN * 100).expect("must be valid"),
```
azarovh, line 503: I think that entire reward should go to stake owner, so that no one would be tempted to delegate to this pool
```
 PerThousand::new(1000).expect("must be valid"),
 Amount::ZERO,
```

TheQuantumPhysicist: Why is delegating to this pool a problem?

azarovh: easy money - it's guarantee to be the winner for the first 2 epochs

TheQuantumPhysicist: I don't think that's true, because there's no special treatment for this pool.

azarovh: its existence already makes it special: it's the only pool that can mine 2 epochs. It's just looks safer to me and also better than arbitrary numbers. But if you think that's invalid in this case I won't insist.

crypto/src/key/mod.rs
pavel-kokolemin: This is used to get raw bytes from the private key (32 bytes in the case of Secp256k1Schnorr). And I needed exactly 32 bytes to derive the VRF keys. We have now added SHA3, so we can now use serialized bytes as well.

wallet/types/src/wallet_tx.rs
pavel-kokolemin: I removed them because they were unused currently. I have reverted this change.

wallet/wallet-cli-lib/src/commands/mod.rs
pavel-kokolemin: I added the `GenerateBlocks` command which takes the number of blocks to generate.

wallet/src/account/mod.rs
pavel-kokolemin: Right. I renamed to `mark_output_as_seen`. I hope it's better.

wallet/storage/src/lib.rs
pavel-kokolemin: We only store kernel inputs and reward outputs here (and only blocks that are relevant to the wallet). So it's not much different from storing `WalletTx`.

wallet/storage/src/internal/test.rs
pavel-kokolemin: Because it was testing UTXO DB store/load which was removed.

wallet/storage/src/internal/mod.rs
pavel-kokolemin: What should I use instead?

crypto/src/vrf/schnorrkel/mod.rs
pavel-kokolemin: I think we can keep this as it was discussed?

blockprod/src/detail/mod.rs
alfiedotwtf: I've cleaned up a few of the timestamp usages around blockprod, including proper naming i.e before, current didn't mean "now" sometimes.

blockprod/src/detail/mod.rs
alfiedotwtf: Yep, done.

blockprod/src/detail/mod.rs
alfiedotwtf: Yep, thanks. As above, I've redone a few timestamp calculations and I think my latest change fixes your comment.

The issue I had was that because we are starting at Genesis for the timestamp, tests were failing because the block timestamps were too old. As a temporary solution, I've temporarily increased `ChainstateConfig.max_tip_age` to a large number so that I don't block Pavel moving forward.

I tried to increase `max_tip_age` for only the functional tests, but it was taking too long working out how the `ChainstateConfigFile` was built for the tests, so I'll leave it for later.

blockprod/src/detail/mod.rs
alfiedotwtf: As above, this has now changed.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: I don't think this issue was addressed. Please add a comment about it. Feel free to discuss with me on Slack if it's not clear.

common/src/chain/config/mod.rs
TheQuantumPhysicist: Was this done? If yes, please write above them how to generate these.

crypto/src/key/mod.rs
TheQuantumPhysicist: This is now gone and encode is used.

wallet/src/account/mod.rs
```diff
@@ -301,109 +394,202 @@ impl Account {
...
+        for tx_id in revoked_txs {
+            db_tx.del_transaction(&tx_id)?;
+            self.output_cache.remove_tx(&tx_id);
         }
 
         Ok(())
     }
 
-    fn scan_genesis_block<B: storage::Backend>(
+    fn add_block_if_relevant<B: storage::Backend>(
```
TheQuantumPhysicist, line 500: Please add docs explaining what relevant means

wallet/wallet-cli-lib/src/commands/mod.rs
TheQuantumPhysicist: Please create a function for normal users like `start_staking`, which will perpetually run and keep creating blocks. This is how monero-simple-wallet works too, I believe, where you issue a command it'll just keep mining.

crypto/src/vrf/mod.rs
```diff
@@ -87,6 +89,25 @@ impl VRFPrivateKey {
         }
     }
 
+    pub fn new_from_bytes(
```
TheQuantumPhysicist, line 92: Please rename this to `new_using_random_bytes`, and add documentation above it saying `This function initializes the key using the bytes safely, but DOES NOT mean that the bytes will be used as the key. There could be an internal algorithm that does the initialization. The function expects a random sequence of bytes.`

I can't find a better way to protect from abusing this.

wallet/types/src/wallet_block.rs
```diff
@@ -0,0 +1,81 @@
...
+    block_id: Id<GenBlock>,
+
+    height: BlockHeight,
+
+    kernel_inputs: Vec<TxInput>,
+
+    reward: Vec<TxOutput>,
+}
+
+impl WalletBlock {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: An open question is: Why don't we make WalletTx an enum, which can be a regular transaction or a block reward?

chainstate/src/config.rs
```diff
@@ -31,7 +31,12 @@ make_config_setting!(
     )
 );
 make_config_setting!(TxIndexEnabled, bool, false);
-make_config_setting!(MaxTipAge, Duration, Duration::from_secs(60 * 60 * 24));
+make_config_setting!(
+    MaxTipAge,
+    Duration,
+    // TODO: temporary fix so that functional tests still work
```
TheQuantumPhysicist, line 37: If this is meant to be reverted before the PR is merged, please add TODO(PR). Otherwise, please explain in more detail what this fix does.

common/src/chain/config/mod.rs
pavel-kokolemin: I added a comment. The CLI wallet now has two new commands for this: newpublickey and getvrfpublickey.

chainstate/src/config.rs
alfiedotwtf: Updated to `TODO(PR)`.

blockprod/src/detail/mod.rs
alfiedotwtf: Thanks for showing me how incorrect this was. I've now updated it with an `AtomicU64` so that we continue from where we left off when we failed to produce a block.

Hopefully the updated code fixes this issue.

blockprod/src/detail/mod.rs
alfiedotwtf: I've removed the checking of consensus here since we should be doing a MAX block timestamp check regardless of the consensus we're currently configured for.

blockprod/src/detail/mod.rs
```diff
@@ -125,6 +126,7 @@ impl BlockProduction {
         &self,
         input_data: GenerateBlockInputData,
         block_timestamp: BlockTimestamp,
+        current_timestamp: BlockTimestamp,
```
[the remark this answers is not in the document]
alfiedotwtf: The latest changes should now be `[previous_timestamp+1,now+120]`.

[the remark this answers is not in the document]
alfiedotwtf: Also returning an error now once we hit `now+120` so that the `start_staking` RPC call doesn't block forever

blockprod/src/detail/mod.rs
alfiedotwtf: An issue I see here now from this change is that PoW block timestamps will be `prev+1` as we don't advance. Not sure what you want as a design decision for this, but an option would be to add back in the `block_timestamp++` for the loop, or to have PoW consensus use `current_timestamp i.e NOW` internally.

wallet/wallet-cli-lib/src/commands/mod.rs
pavel-kokolemin: `startstaking` and `stopstaking` commands added. The wallet-controller will now try to generate new blocks every second in the background.

wallet/types/src/wallet_block.rs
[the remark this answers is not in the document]
pavel-kokolemin: > Question: Why don't we make WalletTx an enum, which can be a regular transaction or a block reward?
Done, thanks.

wallet/src/key_chain/mod.rs
```diff
@@ -130,5 +150,13 @@ fn get_purpose_and_index(
     Ok((purpose, key_index))
 }
 
+/// Derive a VRF private key from a normal PrivateKey
+pub fn vrf_from_private_key(private_key: &PrivateKey) -> (VRFPrivateKey, VRFPublicKey) {
+    let bytes = private_key.encode();
+    let key_hash = crypto::hash::hash::<crypto::hash::Sha3_512, _>(bytes);
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Not that I'm an expert, but this is also what we do for all our hashing with blake2b. Generally speaking, truncation shouldn't matter in hashes. AFAIU, it's a general statement assuming the hash doesn't have any bias in its bytes.

wallet/wallet-controller/src/lib.rs
```diff
@@ -151,9 +184,96 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+                tokio::time::sleep(ERROR_DELAY).await;
+                continue;
+            }
+
+            if self.staking_started {
+                let generate_res = self.generate_block(None).await;
+
+                if let Ok(block) = generate_res {
+                    log::info!(
+                        "New block generated succesfully, block id: {}",
```
TheQuantumPhysicist, line 262: succesfully

blockprod/src/detail/mod.rs
```diff
@@ -254,12 +260,47 @@ impl BlockProduction {
...
+        // searching for a valid block, Proof of Stake can only vary
+        // the header timestamp. Its search space starts at the
+        // previous block's timestamp + 1 second, and ends at the
+        // current timestamp + some distance in time defined by the
+        // blockchain.
+        //
+        // This variable keeps track of the last timestamp that was
+        // attempted, and during Proof of Stake, will prevent
+        // searching over the same search space.
+        let last_epoch_used = {
```
TheQuantumPhysicist, line 273: I think we shouldn't call this epoch... it's confusing because we already have that word used elsewhere.

blockprod/src/detail/mod.rs
```diff
@@ -254,12 +260,47 @@ impl BlockProduction {
...
+            let tip_timestamp = tip_at_start.block_timestamp();
+
+            let tip_plus_one = tip_timestamp
+                .add_int_seconds(1)
+                .ok_or(ConsensusCreationError::TimestampOverflow(tip_timestamp, 1))?;
+
+            Arc::new(AtomicU64::new(tip_plus_one.as_int_seconds()))
+        };
+
+        let current_timestamp =
```
TheQuantumPhysicist, line 283: As a nitpick, since this is called only once, you can scope it inside whatever calculates `max_block_timestamp`

blockprod/src/detail/mod.rs
TheQuantumPhysicist: I just noticed that everything was renamed to epoch... why?

consensus/src/lib.rs
```diff
@@ -173,8 +179,9 @@ pub fn finalize_consensus_data(
     chain_config: &ChainConfig,
     block_header: &mut BlockHeader,
     block_height: BlockHeight,
+    block_epoch: &Arc<AtomicU64>,
```
TheQuantumPhysicist, line 182: Passing a reference of an Arc is bad practice. If you can simply call as_ref on the Arc and pass &AtomicU64. Same lifetime.

wallet/src/key_chain/mod.rs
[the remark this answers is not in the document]
muursh: I'd leave it for now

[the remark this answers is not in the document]
muursh: And open an issue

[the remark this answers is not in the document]
muursh: This whole thread can be ignored. I found the paper that was niggling me in the back of my mind and it doesn't apply to this. Ignore me.

wallet/wallet-controller/src/lib.rs
```diff
@@ -151,9 +184,96 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+        Ok(())
+    }
+
+    /// Synchronize the wallet to the current node tip height and return
+    pub async fn sync_once(&mut self) -> Result<(), ControllerError<T>> {
+        sync::sync_once(&self.chain_config, &self.rpc_client, &mut self.wallet).await?;
+        Ok(())
+    }
+
+    /// Synchronize the wallet continuously from the node's blockchain.
```
azarovh, line 245: continuously

azarovh (approved):
looks good to me

muursh (approved):

blockprod/src/detail/mod.rs
alfiedotwtf: I'm used to calling any integer that represents seconds since Epoch as the unit "epoch". I've renamed "epoch" to "timestamp_seconds".

blockprod/src/detail/mod.rs
alfiedotwtf: Yep, I like doing that. Nice catch.

consensus/src/lib.rs
alfiedotwtf: Thanks for pointing that out. Fixed.

blockprod/src/detail/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: We use now TimeGetter.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: This is gone now.

wallet/src/wallet/mod.rs
```diff
@@ -257,22 +253,57 @@ impl<B: storage::Backend> Wallet<B> {
...
+        account.get_pos_gen_block_data()
     }
 
     /// Returns the last scanned block hash and height.
     /// Returns genesis block when the wallet is just created.
     pub fn get_best_block(&self) -> WalletResult<(Id<GenBlock>, BlockHeight)> {
-        Ok((self.best_block_id, self.best_block_height))
+        // TODO: Scan all accounts
+        let account = self.accounts.values().next().ok_or(WalletError::WalletNotInitialized)?;
+        Ok(account.best_block())
```
OBorce, line 306: why was best block moved to per account, can they be different?

pavel-kokolemin: Yes, I decided that it would be better to keep it separate for different accounts. For example, the user can create a new account at a later time, and while the wallet is scanning the node, it should be possible to continue working with the old accounts. The wallet can be stopped at any time and the sync progress should be kept separate.
But this is for a the future, it's not currently implemented.

TheQuantumPhysicist (approved):
Let's merge this thing.

I have many reservations, but we can tackle them later.

wallet/src/wallet/mod.rs
OBorce: maybe I am not understanding it correctly, wouldn't creating a new account mean there are no transactions associated with it, so nothing to sync?
Or do you mean importing an existing account from a different root key?

pavel-kokolemin: We can't tell if an account has already been used on another node/wallet or not, and should scan all blocks starting at genesis. The root key can be the same, but the derivation path will be different.

OBorce: thanks, didn't think about that.

PR #914, opened 2023-05-24
Add password encryption to wallet for root keys

### LINKED ISSUE #799: Wallet implementation tasks
The following items need to be completed for the wallet to be usable.

- [x] Wallet daemon
- [x] RPC for wallet handling
- [x] Connect via RPC to node daemon
- [x] Scan blocks for transactions
- [x] Pool of pre-generated addresses
- [x] Send requests
- [x] Transaction construction
- [x] Transaction signing
- [ ] Key encryption implementation

OBorce:
- new optional Symmetric key is stored in the Wallet Store
- transactions are now split in 4 types ReadLocked, ReadUnlocked, WriteLocked and WriteUnlocked
where the Unlocked variants have access to the encrypted data in this case the root keys, and the Locked variants do not.
- each get/set operation of the root keys is now encrypted/decrypted according to the set symmetric key
- new util struct `MaybeEncrypted<T>` to help with encode+encrypt a T to bytes and decryption+decode_all from bytes back to T.
- new commands for the Wallet CLI for encrypting/unlocking/locking the private keys

TODO: 
- Argon2Config hardcoded values

Closes #799.

### DIFF
--- a/crypto/src/symkey/mod.rs
+++ b/crypto/src/symkey/mod.rs
@@ -122,7 +122,7 @@ mod test {
     }
 
     #[test]
-    fn constryct_key_from_slice_random_size() {
+    fn construct_key_from_slice_random_size() {
         let mut rng = make_true_rng();
         let bytes: Vec<u8> = (0..rng.gen_range(0..100)).map(|_| rng.gen::<u8>()).collect();
         let result =
--- a/crypto/src/symkey/mod.rs
+++ b/crypto/src/symkey/mod.rs
@@ -33,11 +33,17 @@ pub enum Error {
     WrongLenKeyBytes(usize, usize),
 }
 
-use self::chacha20poly1305::{Chacha20poly1305Key, KEY_LEN};
+use self::chacha20poly1305::Chacha20poly1305Key;
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone, Decode, Encode)]
 pub enum SymmetricKeyKind {
     XChacha20Poly1305,
+}
+
+pub const fn key_size(kind: SymmetricKeyKind) -> usize {
+    match kind {
+        SymmetricKeyKind::XChacha20Poly1305 => Chacha20poly1305Key::KEY_LEN,
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Decode, Encode)]
@@ -54,22 +60,20 @@ pub struct SymmetricKey {
 impl SymmetricKey {
     pub fn new<R: Rng + CryptoRng>(kind: SymmetricKeyKind, rng: &mut R) -> Self {
         let key = match kind {
-            SymmetricKeyKind::XChacha20Poly1305 => SymmetricKeyHolder::XChacha20Poly1305(
-                Chacha20poly1305Key::new_from_array(rng.gen::<[u8; KEY_LEN]>()),
-            ),
+            SymmetricKeyKind::XChacha20Poly1305 => {
+                SymmetricKeyHolder::XChacha20Poly1305(Chacha20poly1305Key::new_from_array(
+                    rng.gen::<[u8; Chacha20poly1305Key::KEY_LEN]>(),
+                ))
+            }
         };
         Self { key }
     }
 
     pub fn from_raw_key(kind: SymmetricKeyKind, bytes_slice: &[u8]) -> Result<Self, Error> {
         let key = match kind {
-            SymmetricKeyKind::XChacha20Poly1305 => {
-                SymmetricKeyHolder::XChacha20Poly1305(Chacha20poly1305Key::new_from_array(
-                    bytes_slice
-                        .try_into()
-                        .map_err(|_| Error::WrongLenKeyBytes(bytes_slice.len(), KEY_LEN))?,
-                ))
-            }
+            SymmetricKeyKind::XChacha20Poly1305 => SymmetricKeyHolder::XChacha20Poly1305(
+                Chacha20poly1305Key::new_from_bytes_slice(bytes_slice)?,
+            ),
         };
         Ok(Self { key })
     }
@@ -110,7 +114,7 @@ mod test {
     #[test]
     fn encode_then_decode_key_from_slice() {
         let mut rng = make_true_rng();
-        let bytes = rng.gen::<[u8; KEY_LEN]>();
+        let bytes = rng.gen::<[u8; Chacha20poly1305Key::KEY_LEN]>();
         let key = SymmetricKey::from_raw_key(SymmetricKeyKind::XChacha20Poly1305, &bytes).unwrap();
         let encoded = key.encode();
         let decoded = SymmetricKey::decode_all(&mut encoded.as_slice()).unwrap();
@@ -123,10 +127,16 @@ mod test {
         let bytes: Vec<u8> = (0..rng.gen_range(0..100)).map(|_| rng.gen::<u8>()).collect();
         let result =
             SymmetricKey::from_raw_key(SymmetricKeyKind::XChacha20Poly1305, bytes.as_slice());
-        if bytes.len() == KEY_LEN {
+        if bytes.len() == Chacha20poly1305Key::KEY_LEN {
             assert!(result.is_ok());
         } else {
-            assert_eq!(result, Err(Error::WrongLenKeyBytes(bytes.len(), KEY_LEN)));
+            assert_eq!(
+                result,
+                Err(Error::WrongLenKeyBytes(
+                    bytes.len(),
+                    Chacha20poly1305Key::KEY_LEN
+                ))
+            );
         }
     }
 
--- a/crypto/src/symkey/mod.rs
+++ b/crypto/src/symkey/mod.rs
@@ -29,9 +29,11 @@ pub enum Error {
     CipherTextTooShort(usize, usize),
     #[error("Decryption error: {0}")]
     DecryptionError(String),
+    #[error("Raw Key bytes wrong size: {0} expected {1}")]
+    WrongLenKeyBytes(usize, usize),
 }
 
-use self::chacha20poly1305::Chacha20poly1305Key;
+use self::chacha20poly1305::{Chacha20poly1305Key, KEY_LEN};
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone, Decode, Encode)]
 pub enum SymmetricKeyKind {
@@ -53,10 +55,23 @@ impl SymmetricKey {
     pub fn new<R: Rng + CryptoRng>(kind: SymmetricKeyKind, rng: &mut R) -> Self {
         let key = match kind {
             SymmetricKeyKind::XChacha20Poly1305 => SymmetricKeyHolder::XChacha20Poly1305(
-                Chacha20poly1305Key::new_from_array(rng.gen::<[u8; 32]>()),
+                Chacha20poly1305Key::new_from_array(rng.gen::<[u8; KEY_LEN]>()),
             ),
         };
         Self { key }
+    }
+
+    pub fn from_raw_key(kind: SymmetricKeyKind, bytes_slice: &[u8]) -> Result<Self, Error> {
+        let key = match kind {
+            SymmetricKeyKind::XChacha20Poly1305 => {
+                SymmetricKeyHolder::XChacha20Poly1305(Chacha20poly1305Key::new_from_array(
+                    bytes_slice
+                        .try_into()
+                        .map_err(|_| Error::WrongLenKeyBytes(bytes_slice.len(), KEY_LEN))?,
+                ))
+            }
+        };
+        Ok(Self { key })
     }
 
     pub fn encrypt<R: Rng + CryptoRng>(
@@ -91,6 +106,29 @@ mod test {
     use crate::random::make_true_rng;
 
     use super::*;
+
+    #[test]
+    fn encode_then_decode_key_from_slice() {
+        let mut rng = make_true_rng();
+        let bytes = rng.gen::<[u8; KEY_LEN]>();
+        let key = SymmetricKey::from_raw_key(SymmetricKeyKind::XChacha20Poly1305, &bytes).unwrap();
+        let encoded = key.encode();
+        let decoded = SymmetricKey::decode_all(&mut encoded.as_slice()).unwrap();
+        assert_eq!(key, decoded);
+    }
+
+    #[test]
+    fn constryct_key_from_slice_random_size() {
+        let mut rng = make_true_rng();
+        let bytes: Vec<u8> = (0..rng.gen_range(0..100)).map(|_| rng.gen::<u8>()).collect();
+        let result =
+            SymmetricKey::from_raw_key(SymmetricKeyKind::XChacha20Poly1305, bytes.as_slice());
+        if bytes.len() == KEY_LEN {
+            assert!(result.is_ok());
+        } else {
+            assert_eq!(result, Err(Error::WrongLenKeyBytes(bytes.len(), KEY_LEN)));
+        }
+    }
 
     #[test]
     fn encode_then_decode() {
--- a/utils/src/maybe_encrypted.rs
+++ b/utils/src/maybe_encrypted.rs
@@ -1,0 +1,177 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::marker::PhantomData;
+
+use crypto::{random::make_true_rng, symkey::SymmetricKey};
+use serialization::{Decode, DecodeAll, Encode};
+use zeroize::Zeroizing;
+
+/// A generic type that stores a value as a vector of bytes that are optionally encrypted
+#[derive(Encode, Decode)]
+pub struct MaybeEncrypted<T: Encode + Decode + DecodeAll> {
+    value: Vec<u8>,
+    _phantom: PhantomData<fn() -> T>,
+}
+
+impl<T: Decode + Encode + DecodeAll> MaybeEncrypted<T> {
+    /// Creates a new `MaybeEncrypted` instance with the given value and encryption key.
+    ///
+    /// If the encryption key is `Some`, the value will be encrypted using the key.
+    /// If the encryption key is `None`, the value will be stored as plain bytes.
+    pub fn new(value: &T, encryption_key: &Option<SymmetricKey>) -> Self {
+        match encryption_key {
+            Some(key) => Self::new_encrypted(value, key),
+            None => Self::new_plain(value),
+        }
+    }
+
+    fn new_plain(value: &T) -> Self {
+        Self {
+            value: value.encode(),
+            _phantom: Default::default(),
+        }
+    }
+
+    fn new_encrypted(value: &T, encryption_key: &SymmetricKey) -> Self {
+        Self {
+            value: encryption_key
+                .encrypt(
+                    Zeroizing::new(value.encode()).as_slice(),
+                    &mut make_true_rng(),
+                    None,
+                )
+                .expect("should not fail"),
+            _phantom: Default::default(),
+        }
+    }
+
+    /// Attempts to take the value from the `MaybeEncrypted` instance.
+    ///
+    /// If the encryption key is `Some`, the value will be decrypted using the key.
+    /// If the encryption key is `None`, the value is assumed to be plain bytes and decoded as is.
+    pub fn try_take(
+        self,
+        encryption_key: &Option<SymmetricKey>,
+    ) -> Result<T, crypto::symkey::Error> {
+        match encryption_key {
+            Some(key) => self.try_take_decrypt(key),
+            None => self.try_take_plain(),
+        }
+    }
+
+    fn try_take_plain(self) -> Result<T, crypto::symkey::Error> {
+        T::decode_all(&mut self.value.as_slice()).map_err(|_| {
+            crypto::symkey::Error::DecryptionError(
+                "Could not decode plain content probably it is encrypted".into(),
+            )
+        })
+    }
+
+    pub fn try_take_decrypt(
+        self,
+        encryption_key: &SymmetricKey,
+    ) -> Result<T, crypto::symkey::Error> {
+        encryption_key.decrypt(self.value.as_slice(), None).map(|decrypted_bytes| {
+            T::decode_all(&mut Zeroizing::new(decrypted_bytes).as_slice())
+                .expect("should have been correctly encoded")
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crypto::random::Rng;
+    use crypto::symkey::SymmetricKeyKind;
+    use rstest::rstest;
+    use test_utils::random::{make_seedable_rng, Seed};
+
+    use super::*;
+
+    #[test]
+    fn test_new_plain_and_back() {
+        let value = 42;
+        let maybe_encrypted = MaybeEncrypted::new(&value, &None);
+
+        assert_eq!(maybe_encrypted.try_take(&None).unwrap(), value);
+    }
+
+    #[test]
+    fn test_new_plain_and_back_plain() {
+        let value = 42;
+        let maybe_encrypted = MaybeEncrypted::new(&value, &None);
+
+        assert_eq!(maybe_encrypted.try_take_plain().unwrap(), value);
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn test_new_plain_and_back_decrypted_error(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+        let value = rng.gen::<u32>();
+        let maybe_encrypted = MaybeEncrypted::new(&value, &None);
+
+        let key = SymmetricKey::new(SymmetricKeyKind::XChacha20Poly1305, &mut rng);
+        let key = Some(key);
+
+        assert!(maybe_encrypted.try_take(&key).is_err());
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn test_new_encrypted_and_back(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+        let value = rng.gen::<u32>();
+        let key = SymmetricKey::new(SymmetricKeyKind::XChacha20Poly1305, &mut rng);
+        let key = Some(key);
+        let maybe_encrypted = MaybeEncrypted::new(&value, &key);
+
+        assert_eq!(maybe_encrypted.try_take(&key).unwrap(), value);
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn test_new_encrypted_and_back_no_key_error(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+        let value = rng.gen::<u32>();
+        let key = SymmetricKey::new(SymmetricKeyKind::XChacha20Poly1305, &mut rng);
+        let key = Some(key);
+        let maybe_encrypted = MaybeEncrypted::new(&value, &key);
+
+        assert!(maybe_encrypted.try_take(&None).is_err());
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn test_new_encrypted_and_back_different_key_error(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+        let value = rng.gen::<u32>();
+        let key = SymmetricKey::new(SymmetricKeyKind::XChacha20Poly1305, &mut rng);
+        let key = Some(key);
+        let maybe_encrypted = MaybeEncrypted::new(&value, &key);
+
+        let mut different_key = SymmetricKey::new(SymmetricKeyKind::XChacha20Poly1305, &mut rng);
+        while &different_key == key.as_ref().unwrap() {
+            different_key = SymmetricKey::new(SymmetricKeyKind::XChacha20Poly1305, &mut rng);
+        }
+        let different_key = Some(different_key);
+
+        assert!(maybe_encrypted.try_take(&different_key).is_err());
+    }
+}
--- a/utils/src/maybe_encrypted.rs
+++ b/utils/src/maybe_encrypted.rs
@@ -19,14 +19,25 @@ use crypto::{random::make_true_rng, symkey::SymmetricKey};
 use serialization::{Decode, DecodeAll, Encode};
 use zeroize::Zeroizing;
 
-/// A generic type that stores a value as a vector of bytes that are optionally encrypted
+use thiserror::Error;
+
+#[derive(Debug, PartialEq, Eq, Clone, Error)]
+pub enum MaybeEncryptedError {
+    #[error("Error in decryption: {0}")]
+    DecryptionError(#[from] crypto::symkey::Error),
+
+    #[error("Error in decoding: {0}")]
+    DecodingError(#[from] serialization::Error),
+}
+
+/// A generic helper type that is used to encode/decode and optionaly encrypt/decrypt a type T
 #[derive(Encode, Decode)]
-pub struct MaybeEncrypted<T: Encode + Decode + DecodeAll> {
+pub struct MaybeEncrypted<T> {
     value: Vec<u8>,
     _phantom: PhantomData<fn() -> T>,
 }
 
-impl<T: Decode + Encode + DecodeAll> MaybeEncrypted<T> {
+impl<T: Encode + DecodeAll> MaybeEncrypted<T> {
     /// Creates a new `MaybeEncrypted` instance with the given value and encryption key.
     ///
     /// If the encryption key is `Some`, the value will be encrypted using the key.
@@ -60,34 +71,35 @@ impl<T: Decode + Encode + DecodeAll> MaybeEncrypted<T> {
 
     /// Attempts to take the value from the `MaybeEncrypted` instance.
     ///
-    /// If the encryption key is `Some`, the value will be decrypted using the key.
+    /// # Arguments
+    ///
+    /// * `encryption_key` - An optional `SymmetricKey`
+    /// If the encryption key is `Some`, the value will be decrypted using the key before decoding.
     /// If the encryption key is `None`, the value is assumed to be plain bytes and decoded as is.
-    pub fn try_take(
-        self,
-        encryption_key: &Option<SymmetricKey>,
-    ) -> Result<T, crypto::symkey::Error> {
+    ///
+    /// # Returns
+    ///
+    /// Returns an `MaybeEncryptedError::DecryptionError` if the `encryption_key` is wrong
+    /// Returns an `MaybeEncryptedError::DecodingError` if the MaybeEncrypted object was decoded
+    /// from an invalid slice of bytes
+    pub fn try_take(self, encryption_key: &Option<SymmetricKey>) -> Result<T, MaybeEncryptedError> {
         match encryption_key {
-            Some(key) => self.try_take_decrypt(key),
+            Some(key) => self.try_decrypt_then_take(key),
             None => self.try_take_plain(),
         }
     }
 
-    fn try_take_plain(self) -> Result<T, crypto::symkey::Error> {
-        T::decode_all(&mut self.value.as_slice()).map_err(|_| {
-            crypto::symkey::Error::DecryptionError(
-                "Could not decode plain content probably it is encrypted".into(),
-            )
-        })
+    fn try_take_plain(self) -> Result<T, MaybeEncryptedError> {
+        T::decode_all(&mut self.value.as_slice()).map_err(MaybeEncryptedError::from)
     }
 
-    pub fn try_take_decrypt(
+    pub fn try_decrypt_then_take(
         self,
         encryption_key: &SymmetricKey,
-    ) -> Result<T, crypto::symkey::Error> {
-        encryption_key.decrypt(self.value.as_slice(), None).map(|decrypted_bytes| {
-            T::decode_all(&mut Zeroizing::new(decrypted_bytes).as_slice())
-                .expect("should have been correctly encoded")
-        })
+    ) -> Result<T, MaybeEncryptedError> {
+        let decrypted_bytes = encryption_key.decrypt(self.value.as_slice(), None)?;
+        T::decode_all(&mut Zeroizing::new(decrypted_bytes).as_slice())
+            .map_err(MaybeEncryptedError::from)
     }
 }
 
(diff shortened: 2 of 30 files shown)


crypto/src/symkey/mod.rs
```diff
@@ -59,6 +59,11 @@ impl SymmetricKey {
         Self { key }
     }
 
+    pub fn new_chacha20_from_bytes(bytes: [u8; KEY_LEN]) -> Self {
+        let key = SymmetricKeyHolder::XChacha20Poly1305(Chacha20poly1305Key::new_from_array(bytes));
+        Self { key }
+    }
```
TheQuantumPhysicist, line 65: You don't need this, unless I'm missing something. You have `Decode` implemented/derived for this struct. So, something like `SymmetricKey::decode_all(&bytes)` should work.

Notice you use `decode_all` to consume all the bytes given, while you sue `decode` to consume only what's needed to create the current object.

crypto/src/symkey/mod.rs
```diff
@@ -31,7 +31,7 @@ pub enum Error {
     DecryptionError(String),
 }
 
-use self::chacha20poly1305::Chacha20poly1305Key;
+use self::chacha20poly1305::{Chacha20poly1305Key, KEY_LEN};
```
TheQuantumPhysicist, line 34: This is also unnecessary. When decoding, the decode process will enforce having the correct key length.

crypto/src/symkey/mod.rs
OBorce: maybe the name is confusing, it is not meant for decoding from bytes, it is creating new key using those bytes as key, the bytes being the ones obtained from the password.

Reason was the other new method does not take bytes, it generates random ones.
The KEY_LEN was put as that is what the Chacha20poly1305Key::new_from_array() was taking, otherwise an error would need to be returned.

wallet/storage/src/internal/mod.rs
```diff
@@ -31,13 +37,58 @@ use wallet_types::{
     AccountTxId, KeychainUsageState, RootKeyContent, RootKeyId, WalletTx,
 };
 
+fn password_to_sym_key(password: String, salt: &[u8; 32]) -> SymmetricKey {
```
TheQuantumPhysicist, line 40: salt shouldn't be a parameter. It should be generated on the fly when it's needed to be used.

wallet/storage/src/internal/mod.rs
```diff
@@ -31,13 +37,58 @@ use wallet_types::{
...
+
 /// Store for wallet data, parametrized over the backend B
-pub struct Store<B: storage::Backend>(storage::Storage<B, Schema>);
+pub struct Store<B: storage::Backend>(storage::Storage<B, Schema>, Option<SymmetricKey>);
 
 impl<B: storage::Backend> Store<B> {
     /// Create a new wallet storage
-    pub fn new(backend: B) -> crate::Result<Self> {
-        let storage = Self(storage::Storage::new(backend).map_err(crate::Error::from)?);
+    pub fn new(backend: B, password: Option<String>) -> crate::Result<Self> {
```
TheQuantumPhysicist, line 59: I'm becoming more inclined to not put the password on startup at the constructor. This doesn't seem to be correct.

The way I see it is: The wallet open without a password, whether the private keys are encrypted or not. Since we're using xpub, we can derive all the public keys we need without a password. 

To encrypt/decrypt and similar functions, there should be functions in the wallet controller:

- `decrypt(password)`: Opens the wallet, and stores the key in memory, so that the private key can be used
- `encrypt(Option<password>)`: This function encrypts the current wallet, assuming it's decrypted in memory. `None` would remove the encryption.

The functions that require the private key will first check if the wallet is decrypted.

crypto/src/symkey/mod.rs
TheQuantumPhysicist: Thinking about this for a minute, I don't see a way around having an error. The problem is that this interface is generic and isn't only for chacha.

So, I recommend creating a method here: `from_raw_key(KeyKind, bytes_slice)`, and depending on the key kind, we enforce the slice size.

wallet/storage/src/internal/mod.rs
OBorce: it is generated on the first try, and on change password in the future. 

But for unlocking you must reuse the existing one otherwise the same password with different salt will give you a different key and we will be unable to unlock the wallet.

wallet/storage/src/internal/mod.rs
OBorce: Agreed, bitcoin core wallet also does that with lock/unlock. You can also use the wallet as watch only without any private keys.

The difference being they still store all private keys encrypted, and only keep the encryption keys (not sure why they can have multiple) to decrypt the specific private key before usage.
Unlock iterates the map of the private keys and only tries to see if it can decrypt them all and returns an error if not.

crypto/src/symkey/mod.rs
OBorce: if it is a runtime decision for which `KeyKind` to use yes, but if it is known at compile time which one to use wouldn't it be ok to have a `from_raw_key_specific_algo(bytes_slice)` so you avoid the error/using except("")?

TheQuantumPhysicist: It's not that big of a deal to have some error here... interface consistency is more important in this case, I'd say.

wallet/storage/src/internal/mod.rs
TheQuantumPhysicist: In the database you should store `KdfChallenge`, which contains the salt and the other parameters you need to reconstruct the hash. It also contains the facility to compare against a password using `hash_from_challenge()`.

wallet/storage/src/internal/mod.rs
TheQuantumPhysicist: Yep. Let's call it lock/unlock. That sounds clearer.

crypto/src/symkey/mod.rs
OBorce: changed to `from_raw_key(KeyKind, bytes_slice)` that returns an error for invalid key bytes length

wallet/storage/src/internal/mod.rs
OBorce: added 3 separate methods for encrypt/unlock/lock private_keys

wallet/storage/src/internal/mod.rs
OBorce: fixed, changed to storing the KdfChallenge instead of just the salt.

wallet/storage/src/internal/mod.rs
```diff
@@ -31,19 +39,133 @@ use wallet_types::{
...
+        };
+        tx.encrypt_root_keys(&sym_key)?;
+        tx.commit()?;
+
+        self.encryption_key = sym_key;
+
+        Ok(())
+    }
+
+    pub fn unlock_unlock_private_keys(&mut self, password: &Option<String>) -> crate::Result<()> {
```
OBorce, line 133: typo unlock_unlock

wallet/storage/src/internal/store_tx.rs
```diff
@@ -99,32 +142,89 @@ macro_rules! impl_read_ops {
...
+            ) -> crate::Result<()> {
+                self.storage
+                    .get::<db::DBRootKeys, _>()
+                    .prefix_iter_decoded(&())
+                    .map_err(crate::Error::from)
+                    .map(|mut item| {
+                        item.try_for_each(|(_, v)| {
+                            let decrypted_value = encryption_key
+                                .decrypt(v.as_slice(), None)
+                                .map_err(|_| crate::Error::WalletInvalidPassword())?;
```
iljakuklic, line 210: Not familiar with the encryption algorithm being used here. Does this have sufficient integrity checks to reliably detect the symmetric key used is the correct one?

wallet/storage/src/schema.rs
```diff
@@ -33,7 +33,7 @@ storage::decl_schema! {
         /// Store keychain usage states
         pub DBKeychainUsageStates: Map<AccountKeyPurposeId, KeychainUsageState>,
         /// Store for all the private keys in this wallet
-        pub DBRootKeys: Map<RootKeyId, RootKeyContent>,
+        pub DBRootKeys: Map<RootKeyId, Vec<u8>>,
```
iljakuklic, line 36: I'd still prefer some sort of wrapper type that communicates it's a (possibly encrypted) root key over a plain `Vec<u8>`.

wallet/storage/src/internal/mod.rs
```diff
@@ -31,19 +39,133 @@ use wallet_types::{
...
     /// Create a new wallet storage
     pub fn new(backend: B) -> crate::Result<Self> {
-        let storage = Self(storage::Storage::new(backend).map_err(crate::Error::from)?);
+        let storage: storage::Storage<B, Schema> =
+            storage::Storage::new(backend).map_err(crate::Error::from)?;
+
+        let mut storage = Self {
+            storage,
+            locked: true,
+            encryption_key: None,
```
iljakuklic, line 101: Should these two fields be rolled into one somehow?
Possibly using an enum, very rough sketch:
```rust
enum KeyStorageState {
    Decrypted(SymmKey), // Keys are available and are decrypted (or decryptable?) using given key
    Plain,              // Keys are available in plaintext, no need to decrypt
    Locked,             // Keys locked, may or may not be encrypted
}
```

Or is `locked = true && encryption_key = Some(_)` an admissible state?

iljakuklic (commented):
Quick question: How about password protecting the whole database file? Is that planned? Seems like currently, it only protects the private key entry in the db.

Some initial minor code suggestions below.

wallet/storage/src/internal/store_tx.rs
```diff
@@ -229,15 +331,62 @@ impl<'st, B: storage::Backend> WalletStorageWrite for StoreTxRw<'st, B> {
...
+                    (
+                        k,
+                        match new_encryption_key.as_ref() {
+                            Some(key) => key
+                                .encrypt(v.as_slice(), &mut make_true_rng(), None)
+                                .expect("should not fail"),
+                            None => v,
+                        },
+                    )
+                })
```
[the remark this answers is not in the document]
OBorce: fixed

wallet/storage/src/internal/mod.rs
OBorce: created an EncryptionState with only Locked and Unlocked(Option<SymmKey>), removed Plain as it was making the code more verbose.

wallet/storage/src/schema.rs
OBorce: Created a wrapper struct `MaybeEncrypted<T>` in util

wallet/storage/src/internal/store_tx.rs
OBorce: the algorithm is ChaCha20-Poly1305 which is an authenticated cipher, it will verify that the decryption is correct.

iljakuklic (commented):
Cool stuff. Some comments on the recent changes below:

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use crypto::{random::make_true_rng, symkey::SymmetricKey};
+use serialization::{Decode, DecodeAll, Encode};
+
+/// A generic type that stores a value as a vector of bytes that are optionally encrypted
+#[derive(Encode, Decode)]
+pub struct MaybeEncrypted<T: Encode + Decode + DecodeAll> {
+    value: Vec<u8>,
+    _phantom: PhantomData<T>,
+}
```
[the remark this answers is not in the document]
OBorce: fixed, didn't know that.

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    fn new_encrypted(value: &T, encryption_key: &SymmetricKey) -> Self {
+        Self {
+            value: encryption_key
+                .encrypt(value.encode().as_slice(), &mut make_true_rng(), None)
+                .expect("should not fail"),
+            _phantom: Default::default(),
+        }
+    }
```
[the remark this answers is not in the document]
OBorce: Wrapped the vectors returned from encode and decode in `Zeroizing`, which will move the vector inside the new struct and zeroize the memory on Drop.

wallet/storage/src/internal/store_tx.rs
```diff
@@ -335,54 +326,42 @@ impl<'st, B: storage::Backend> WalletStorageWrite for StoreTxRw<'st, B> {
...
+                .storage
+                .get::<db::DBRootKeys, _>()
+                .prefix_iter_decoded(&())
+                .map_err(crate::Error::from)
+                .map(|item| {
+                    item.map(|(k, v)| {
+                        let decrypted = v.try_take(key).expect("key was checked when unlocked");
+                        (k, MaybeEncrypted::new(&decrypted, new_encryption_key))
+                    })
+                })?
```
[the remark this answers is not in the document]
OBorce: correct, fixed

crypto/src/symkey/mod.rs
```diff
@@ -53,12 +55,25 @@ impl SymmetricKey {
...
     }
 
+    pub fn from_raw_key(kind: SymmetricKeyKind, bytes_slice: &[u8]) -> Result<Self, Error> {
+        let key = match kind {
+            SymmetricKeyKind::XChacha20Poly1305 => {
+                SymmetricKeyHolder::XChacha20Poly1305(Chacha20poly1305Key::new_from_array(
+                    bytes_slice
+                        .try_into()
+                        .map_err(|_| Error::WrongLenKeyBytes(bytes_slice.len(), KEY_LEN))?,
+                ))
```
TheQuantumPhysicist, line 71: This should be an implementation detail inside the chacha20poly1305 module.

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,177 @@
...
+        })
+    }
+
+    pub fn try_take_decrypt(
+        self,
+        encryption_key: &SymmetricKey,
+    ) -> Result<T, crypto::symkey::Error> {
+        encryption_key.decrypt(self.value.as_slice(), None).map(|decrypted_bytes| {
+            T::decode_all(&mut Zeroizing::new(decrypted_bytes).as_slice())
+                .expect("should have been correctly encoded")
```
TheQuantumPhysicist, line 89: You cannot enforce this invariant based on this design. Given that Decode is derived for MaybeEncrypted, is by implementing Decode yourself and ensuring that any decoded object holds the required invariant.

I recommend that you support decoding errors. Either way, with or without that, it's possible to have invalid objects in MaybeEncrypted.

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,177 @@
...
+            crypto::symkey::Error::DecryptionError(
+                "Could not decode plain content probably it is encrypted".into(),
+            )
+        })
+    }
+
+    pub fn try_take_decrypt(
+        self,
+        encryption_key: &SymmetricKey,
+    ) -> Result<T, crypto::symkey::Error> {
```
TheQuantumPhysicist, line 86: In the spirit of supporting decoding errors, it's probably a good idea to have a dedicated error type for this, using `thiserror`, like we always do. That can support decoding errors and encryption/decryption errors.

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    fn try_take_plain(self) -> Result<T, crypto::symkey::Error> {
+        T::decode_all(&mut self.value.as_slice()).map_err(|_| {
+            crypto::symkey::Error::DecryptionError(
+                "Could not decode plain content probably it is encrypted".into(),
+            )
+        })
+    }
+
+    pub fn try_take_decrypt(
```
TheQuantumPhysicist, line 83: I would call this `try_decrypt_then_take`

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,172 @@
...
+        self,
+        encryption_key: &SymmetricKey,
+    ) -> Result<T, crypto::symkey::Error> {
+        encryption_key
+            .decrypt(self.value.as_slice(), None)
+            .map(|decrypted_bytes| T::decode_all(&mut decrypted_bytes.as_slice()))?
+            .map_err(|_| {
+                crypto::symkey::Error::DecryptionError("Could not decode decrypted content".into())
+            })
```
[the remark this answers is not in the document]
iljakuklic: Apologies, `flat_map` for Option is called `and_then`, my bad.

wallet/wallet-controller/src/lib.rs
```diff
@@ -123,6 +123,18 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
         Ok(wallet)
     }
 
+    pub fn encrypt_wallet(&mut self, password: &Option<String>) -> Result<(), ControllerError<T>> {
```
TheQuantumPhysicist, line 126: Please add documentation to those three function, including what the parameters mean. Given the controller is the heighest level function in the wallet, it's important to make everything clear so that a user who doesn't have to deal with the details of the wallet would understand how to use it.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -49,6 +48,21 @@ pub enum WalletCommand {
...
     CloseWallet,
 
+    /// Encrypts the private keys with a new password, expects the wallet to be unlocked
+    EncryptPrivateKeys {
+        // Optional new password, if not specified will remove any existing encryption.
+        password: Option<String>,
+    },
+
+    // Unlocks the private keys for usage.
+    UnlockPrivateKeys {
```
TheQuantumPhysicist, line 58: Why would someone attempt to "unlock" the wallet without providing a password?

wallet/wallet-controller/src/lib.rs
```diff
@@ -123,6 +123,18 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+        self.wallet.encrypt_wallet(password).map_err(ControllerError::WalletError)
+    }
+
+    pub fn unlock_wallet(&mut self, password: &Option<String>) -> Result<(), ControllerError<T>> {
```
TheQuantumPhysicist, line 130: Same here... unlock without a password is kind of weird and needs explanation.

wallet/storage/Cargo.toml
```diff
@@ -15,9 +15,12 @@ storage = { path = '../../storage', features = ['inmemory'] }
 storage-sqlite = { path = "../../storage/sqlite" }
 utxo = { path = '../../utxo' }
 wallet-types = { path = '../types' }
+utils = { path = '../../utils' }
+
+# External dependencies
```
TheQuantumPhysicist, line 20: No need to label them.

wallet/storage/src/lib.rs
```diff
@@ -30,9 +30,19 @@ use wallet_types::{
     AccountTxId, KeychainUsageState, RootKeyContent, RootKeyId, WalletTx,
 };
 
+/// Wallet Errors
+#[derive(Debug, Ord, PartialOrd, PartialEq, Eq, Clone, thiserror::Error)]
+pub enum Error {
+    #[error("{0}")]
+    StorageError(#[from] storage::Error),
+    #[error("The wallet is locked")]
+    WalletLocked(),
```
TheQuantumPhysicist, line 39: Why the extra parenthesis?

wallet/storage/src/lib.rs
```diff
@@ -30,9 +30,19 @@ use wallet_types::{
...
+    #[error("Invalid wallet password")]
+    WalletInvalidPassword(),
```
TheQuantumPhysicist, line 41: Why the extra parenthesis?

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: It was in the case when there is no password, but I guess it doesn't make sense to lock it without a password as nothing is encrypted. 
I will change it to return an error if lock is called when there is no password already set, and remove the optional from the unlock.

wallet/src/key_chain/master_key_chain/mod.rs
```diff
@@ -74,50 +72,47 @@ impl MasterKeyChain {
...
 
-        Ok(MasterKeyChain {
-            chain_config,
-            root_key,
-        })
+        Ok(key)
+    }
+
+    /// Creates a Master key chain, checks the database for an existing one
+    pub fn existing_from_database(
```
TheQuantumPhysicist, line 90: `new_from_existing_database()`?

wallet/src/wallet/tests.rs
```diff
@@ -398,24 +431,120 @@ fn wallet_accounts_creation() {
...
+const NETWORK_FEE: u128 = 10000;
+
+fn gen_random_password(rng: &mut (impl Rng + CryptoRng)) -> Option<String> {
+    let new_password: String = (0..rng.gen_range(0..100)).map(|_| rng.gen::<char>()).collect();
+    if new_password.is_empty() {
+        Some(new_password)
+    } else {
+        None
+    }
 }
```
TheQuantumPhysicist, line 550: Please put helper functions and constants at the top of the module

wallet/src/wallet/tests.rs
```diff
@@ -398,24 +431,120 @@ fn wallet_accounts_creation() {
...
+    wallet.unlock_wallet(&password).unwrap();
+    wallet
+        .create_transaction_to_addresses(DEFAULT_ACCOUNT_INDEX, vec![new_output])
+        .unwrap();
+}
+
+const NETWORK_FEE: u128 = 10000;
+
+fn gen_random_password(rng: &mut (impl Rng + CryptoRng)) -> Option<String> {
+    let new_password: String = (0..rng.gen_range(0..100)).map(|_| rng.gen::<char>()).collect();
```
TheQuantumPhysicist, line 544: Is the password allowed to be empty? We have to be careful about that and define the behavior and not create a semi-predicate situation with empty meaining both no encryption and a specific encryption. Preventing empty passwords is probably a good idea.

wallet/storage/src/internal/mod.rs
```diff
@@ -31,19 +39,138 @@ use wallet_types::{
...
+    let config = KdfConfig::Argon2id {
+        // TODO: hardcoded values
+        config: Argon2Config::new(700, 16, 2),
+        hash_length: NonZeroUsize::new(32).expect("not 0"),
+        salt_length: NonZeroUsize::new(32).expect("not 0"),
+    };
+    let kdf_result = hash_password(&mut rng, config, password.as_bytes())
+        .map_err(|_| crate::Error::WalletInvalidPassword())?;
+    let KdfResult::Argon2id {
+        hashed_password, ..
```
TheQuantumPhysicist, line 73: Please don't use `..` at all. Generally speaking, we want the compiler to yell at us when we change something, so that everything can be thought of when invariants are broken.

wallet/storage/src/internal/store_tx.rs
```diff
@@ -52,13 +55,50 @@ mod well_known {
     }
 
     declare_entry!(StoreVersion: u32);
+    declare_entry!(EncryptionKeyKdfChallenge: KdfChallenge);
+}
+
+#[derive(PartialEq, Clone)]
+pub enum EncryptionState {
+    Locked,
+    Unlocked(Option<SymmetricKey>),
```
TheQuantumPhysicist, line 64: What does `Unlocked<None>` mean?

If it means Unlocked without encryption, can we make a third state, `Unencrypted` to cover that?

TheQuantumPhysicist: You know what, I thought about it. Keep it like this. Just please add some documentation explaining what `Unlocked(None)` means.

wallet/storage/src/internal/test.rs
```diff
@@ -68,3 +68,66 @@ fn read_write_utxo_in_db_transaction(#[case] seed: Seed) {
...
+    utils::concurrency::model(move || {
+        let mut rng = make_seedable_rng(seed);
+        let mut store = Store::new(DefaultBackend::new_in_memory()).unwrap();
+        let (xpriv_key, xpub_key) =
+            ExtendedPrivateKey::new_from_rng(&mut rng, ExtendedKeyKind::Secp256k1Schnorr);
+        let key_id = RootKeyId::from(xpub_key);
+        let key_content = RootKeyContent::from(xpriv_key);
+        store.set_root_key(&key_id, &key_content).unwrap();
+
+        // check it was writen correctly
```
TheQuantumPhysicist, line 85: written

wallet/storage/src/internal/store_tx.rs
```diff
@@ -52,13 +55,50 @@ mod well_known {
...
 }
 
 /// Read-only chainstate storage transaction
-pub struct StoreTxRo<'st, B: storage::Backend>(pub(super) storage::TransactionRo<'st, B, Schema>);
+pub struct StoreTxRo<'st, B: storage::Backend> {
+    pub(super) storage: storage::TransactionRo<'st, B, Schema>,
```
TheQuantumPhysicist, line 69: I don't think `pub(super)` is necessary here. Besides that, generally speaking, we consider `pub(X)` to be a bad practice that can be avoided with good structuring of modules. Sometimes it simplfies things.

One more thing: If there's something you're doing temporarily in the PR, please use either `TODO(PR)` or `FIXME`. Both will trigger a CI error through a python script (can be seen in `do_checks.sh`) to remind you of them.

TheQuantumPhysicist (approved):
Great work! I'm done with the first round. Tons of comments, I know. As discussed, please rebase this on Pavel's PR and let's see how the conflicts will go.

utils/src/maybe_encrypted.rs
OBorce: Added new MaybeEncryptedError with decoder or decryption errors.
In Unlock both errors are treated as invalid password, although decoding should not fail unless the DB was corrupted or modified.

utils/src/maybe_encrypted.rs
OBorce: done

utils/src/maybe_encrypted.rs
OBorce: renamed

wallet/wallet-controller/src/lib.rs
OBorce: done

wallet/wallet-controller/src/lib.rs
OBorce: removed Optional

wallet/storage/src/lib.rs
OBorce: removed

wallet/src/key_chain/master_key_chain/mod.rs
OBorce: renamed

wallet/src/wallet/tests.rs
```diff
@@ -168,13 +173,45 @@ fn wallet_balance_genesis() {
...
+    let mut wallet = Wallet::new_wallet(Arc::clone(&chain_config), db, MNEMONIC, None).unwrap();
+    wallet.create_account(DEFAULT_ACCOUNT_INDEX).unwrap();
+
+    let (coin_balance, _) = wallet.get_balance(DEFAULT_ACCOUNT_INDEX).unwrap();
+    assert_eq!(coin_balance, genesis_amount);
+
+    let password = gen_random_password(&mut rng);
+    wallet.encrypt_wallet(&password).unwrap();
+    wallet.lock_wallet();
+
```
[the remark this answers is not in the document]
OBorce: Added return of an error if the the wallet can't be locked in the case when there is no password set.

wallet/src/wallet/tests.rs
OBorce: moved

wallet/storage/src/internal/mod.rs
OBorce: fixed

wallet/storage/src/internal/store_tx.rs
OBorce: added comments

wallet/storage/src/internal/test.rs
OBorce: fixed typo

wallet/storage/src/internal/store_tx.rs
OBorce: that was like that from before, was not sure what was the use case,
Made the field private as it is not used from anywhere outside, and I don't think the internal storage should be touched on a transaction anyway.

wallet/storage/src/internal/mod.rs
```diff
@@ -31,19 +36,99 @@ use wallet_types::{
...
 };
 
+use self::store_tx::EncryptionState;
+
 /// Store for wallet data, parametrized over the backend B
-pub struct Store<B: storage::Backend>(storage::Storage<B, Schema>);
+pub struct Store<B: storage::Backend> {
+    storage: storage::Storage<B, Schema>,
+    encryption_state: EncryptionState,
+}
```
iljakuklic, line 45: Probably more for the future but wondering whether it would make sense to keep track of the encryption state at compilation time.

Rough sketch:

```rust
struct Locked;
struct Unlocked(Option<Key>);

struct Store<B, ES> { storage: Storage<B>, encryption: ES }
impl Store { fn new() -> Result<Either<Store<B, Locked>, Store<B, Unlocked>>>; }
impl Store<B, Locked> {
    fn unlock(self, passwd) -> Result<Store<B, Unlocked>>;
}
impl Store<B, Unlocked> {
    fn lock(self) -> Store<B, Locked>;
    // Other operations only applicable to unlocked storage
}
```

Thoughts? Is the usage too dyanmic for something like that to be useful?

iljakuklic (commented):
Some more minor suggestions/comments below:

wallet/wallet-controller/src/lib.rs
```diff
@@ -127,6 +127,41 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+
+    /// Unlocks the wallet using the specified password.
+    ///
+    /// # Arguments
+    ///
+    /// * `password` - A `String` representing the old password that was used to encrypt the wallet.
+    ///
+    /// # Returns
+    ///
+    /// This method returns an error the password is incorrect
```
TheQuantumPhysicist, line 151: if

wallet/wallet-controller/src/lib.rs
```diff
@@ -127,6 +127,41 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+    pub fn unlock_wallet(&mut self, password: &String) -> Result<(), ControllerError<T>> {
+        self.wallet.unlock_wallet(password).map_err(ControllerError::WalletError)
+    }
+
+    /// Locks the wallet.
```
TheQuantumPhysicist, line 156: Locks the wallet by making the encrypted private keys inaccessible

wallet/storage/src/internal/store_tx.rs
```diff
@@ -52,13 +55,53 @@ mod well_known {
     }
 
     declare_entry!(StoreVersion: u32);
+    declare_entry!(EncryptionKeyKdfChallenge: KdfChallenge);
+}
+
+#[derive(PartialEq, Clone)]
+pub enum EncryptionState {
+    // The DB is encrypted and we don't have the key to decrypt it
```
TheQuantumPhysicist, line 63: I wouldn't say "The DB"... I would say "The secret parts of the DB, such as private keys". The distinction is vital, because if the whole DB is encrypted, then nothing is accessible, which isn't true based on our design.

wallet/src/wallet/tests.rs
```diff
@@ -31,16 +31,25 @@ use common::{
...
 
 // TODO: Many of these tests require randomization...
 
 const MNEMONIC: &str =
     "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about";
 
+const NETWORK_FEE: u128 = 10000;
+
+fn gen_random_password(rng: &mut (impl Rng + CryptoRng)) -> String {
+    (0..rng.gen_range(1..100)).map(|_| rng.gen::<char>()).collect()
```
TheQuantumPhysicist, line 50: Given that we changed this to 1..100, do we have a test that verifies the behavior of an empty password?

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/// A generic type that stores a value as a vector of bytes that are optionally encrypted
+#[derive(Encode, Decode)]
+pub struct MaybeEncrypted<T: Encode + Decode + DecodeAll> {
+    value: Vec<u8>,
+    _phantom: PhantomData<fn() -> T>,
+}
+
+impl<T: Decode + Encode + DecodeAll> MaybeEncrypted<T> {
```
TheQuantumPhysicist, line 40: DecodeAll is implied when Decode is implemented. You can see the impl block of DecodeAll imposing Decode trait bound.

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,180 @@
...
+pub enum MaybeEncryptedError {
+    #[error("Error in decryption: {0}")]
+    DecryptionError(#[from] crypto::symkey::Error),
+
+    #[error("Error in decoding: {0}")]
+    DecodingError(#[from] serialization::Error),
+}
+
+/// A generic type that stores a value as a vector of bytes that are optionally encrypted
+#[derive(Encode, Decode)]
```
TheQuantumPhysicist, line 34: Just to be clear: Decode is derived, which means the construction rules can be circumvented.

Besides, DecodingError isn't used. I'm assuming you're not done with this yet?

wallet/storage/src/internal/mod.rs
iljakuklic: > This is nice, and is a common pattern in Rust. The problem is that the call decisions are dynamic from the user. There are wallet commands at the controller level to lock/unlock that eventually yield these calls at the low level.

You can also have the dynamic version by introducing something like `MaybeLocked`:
```rust
enum MaybeLocked { Locked | Unlocked(Option<Key>) }
impl Store<B, MaybeLocked> { fn check(self) -> Either<Store<B, Locked>, Store<B, Unlocked>>; }
```

That way:
* If the storage is `MaybeLocked`, you are forced to check.
* If you know it's unlocked, you can access it without checking.

Not suggesting this is a good idea, it all depends how is the store used. Just pointing out something like that is possible.

wallet/storage/src/internal/password.rs
```diff
@@ -0,0 +1,122 @@
...
+        hashed_password,
+        config: _,
+        salt: _,
+    } = &kdf_result;
+
+    let sym_key = SymmetricKey::from_raw_key(
+        SymmetricKeyKind::XChacha20Poly1305,
+        hashed_password.as_slice(),
+    )
+    .expect("must be correct size");
```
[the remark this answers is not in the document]
OBorce: That was discussed at the beginning with @TheQuantumPhysicist
https://github.com/mintlayer/mintlayer-core/pull/914#discussion_r1204496748

the problem is `SymmetricKey` can have multiple implementations that required different key sizes, to propagate the const generic key size we would need a specific `from_raw_key_specific_key_kind`, but the decision was not to have it.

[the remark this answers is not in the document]
iljakuklic: Ok, that's fair. Cheers.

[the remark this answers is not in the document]
iljakuklic: Ok, that's fair. Cheers.

wallet/storage/src/internal/mod.rs
OBorce: The Store is always decided at runtime if it is locked or unlocked. But I can modify the transaction objects to have separate types for `ReadOnlyLocked`, `ReadOnlyUnlocked`, `ReadWriteLocked`, `ReadWriteUnlocked`
where the Locked variants will not have any methods to operate on the encrypted things (private keys)
and the Unlocked variants will have them.

I also need to separate the actual locking/unlocking/encrypting methods to a private trait so that others outside the Store can't change the state through a transaction object.

wallet/wallet-controller/src/lib.rs
```diff
@@ -127,6 +127,41 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+    /// This method returns an error if the wallet is locked
+    pub fn encrypt_wallet(&mut self, password: &Option<String>) -> Result<(), ControllerError<T>> {
+        self.wallet.encrypt_wallet(password).map_err(ControllerError::WalletError)
+    }
+
+    /// Unlocks the wallet using the specified password.
+    ///
+    /// # Arguments
+    ///
+    /// * `password` - A `String` representing the old password that was used to encrypt the wallet.
```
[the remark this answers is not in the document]
OBorce: fixed

wallet/wallet-controller/src/lib.rs
OBorce: fixed

wallet/wallet-controller/src/lib.rs
```diff
@@ -127,6 +127,41 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
         Ok(wallet)
     }
 
+    /// Encrypts the wallet using the specified password, or removes the existing encryption if the password is `None`.
```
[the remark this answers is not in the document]
OBorce: fixed

wallet/storage/src/internal/store_tx.rs
OBorce: fixed

wallet/src/wallet/tests.rs
OBorce: added an error for empty passwords and 2 tests in the password and wallet modules

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,180 @@
...
+pub struct MaybeEncrypted<T: Encode + Decode + DecodeAll> {
```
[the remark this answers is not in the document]
OBorce: removed

utils/src/maybe_encrypted.rs
OBorce: removed Decode

utils/src/maybe_encrypted.rs
OBorce: Added a test for decoding failure. In the case of the DB I put a panic in the case of a decoding error as it means the DB has been corrupted.

utils/src/maybe_encrypted.rs
```diff
@@ -0,0 +1,217 @@
...
+#[derive(Debug, PartialEq, Eq, Clone, Error)]
+pub enum MaybeEncryptedError {
+    #[error("Error in decryption: {0}")]
+    DecryptionError(#[from] crypto::symkey::Error),
+
+    #[error("Error in decoding: {0}")]
+    DecodingError(#[from] serialization::Error),
+}
+
+/// A generic helper type that is used to encode/decode and optionaly encrypt/decrypt a type T
```
TheQuantumPhysicist, line 33: optionally

wallet/wallet-controller/src/lib.rs
```diff
@@ -132,6 +132,41 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+    ///
+    /// * `password` - A `String` representing the password that was used to encrypt the wallet.
+    ///
+    /// # Returns
+    ///
+    /// This method returns an error if the password is incorrect
+    pub fn unlock_wallet(&mut self, password: &String) -> Result<(), ControllerError<T>> {
+        self.wallet.unlock_wallet(password).map_err(ControllerError::WalletError)
+    }
+
```
TheQuantumPhysicist, line 160: There's a typo "succesfully" in this file. Can you please fix it?

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -52,6 +51,21 @@ pub enum WalletCommand {
     /// Close wallet file
     CloseWallet,
 
+    /// Encrypts the private keys with a new password, expects the wallet to be unlocked
+    EncryptPrivateKeys {
+        // Optional new password, if not specified will remove any existing encryption.
+        password: Option<String>,
+    },
```
TheQuantumPhysicist, line 58: I'm a little concerned about the usage here. While the API makes sense to have an `Option<String>` due to the 1:1 mapping, this doesn't make sense in a command line prompt interface. Basically, I'd expect if someone entrs "encrypt_private_keys" in a CLI with no arguemnts, to have an error "please enter the password". 

So while the API is OK, I'd rather have another CLI function `RemovePrivateKeysEncryption`, which does the `None` part, while `EncryptPrivateKeys` enforces providing a password as an argument.

crypto/src/symkey/mod.rs
```diff
@@ -92,6 +111,35 @@ mod test {
...
+        let mut rng = make_true_rng();
+        let bytes = rng.gen::<[u8; Chacha20poly1305Key::KEY_LEN]>();
+        let key = SymmetricKey::from_raw_key(SymmetricKeyKind::XChacha20Poly1305, &bytes).unwrap();
+        let encoded = key.encode();
+        let decoded = SymmetricKey::decode_all(&mut encoded.as_slice()).unwrap();
+        assert_eq!(key, decoded);
+    }
+
+    #[test]
+    fn constryct_key_from_slice_random_size() {
```
TheQuantumPhysicist, line 125: construct

TheQuantumPhysicist (approved):
Looking good! Thank you! Please fix the few minor issues I mentioned then feel free to merge.

utils/src/maybe_encrypted.rs
OBorce: fixed

wallet/wallet-controller/src/lib.rs
OBorce: fixed

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: added a new `RemovePrivateKeysEncryption` command to the wallet CLI, and made password requred for the `EncryptPrivateKeys` command

PR #918, opened 2023-05-25
Blockprod cleanup

alfiedotwtf:
Simplifying and tidying up areas that were too bulky, so it's easier to test

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -15,12 +15,10 @@
 
 pub mod job_manager;
 
-use crate::{
-    detail::job_manager::{JobKey, JobManager},
-    BlockProductionError,
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    mpsc, Arc,
 };
-
-use std::sync::{atomic::AtomicBool, mpsc, Arc};
 
 use chainstate::{chainstate_interface::ChainstateInterface, ChainstateHandle, PropertyQueryError};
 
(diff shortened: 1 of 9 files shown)


TheQuantumPhysicist (approved):

blockprod/src/detail/mod.rs
```diff
@@ -443,11 +346,129 @@ impl BlockProduction {
...
+    chainstate_handle: &(dyn ChainstateInterface),
+    best_block_index: &GenBlockIndex,
+    block_height: BlockHeight,
+    sealed_epoch_index: Option<EpochIndex>,
+    sealed_epoch_randomness: Option<PoSRandomness>,
+    input_data: GenerateBlockInputData,
+) -> Result<Option<FinalizeBlockInputData>, ConsensusPoSError> {
+    match input_data {
+        GenerateBlockInputData::PoS(pos_input_data) => {
+            let sealed_epoch_index = sealed_epoch_index.ok_or(ConsensusPoSError::NoEpochData)?;
```
TheQuantumPhysicist, line 419: Btw, this is wrong. It just caused an error for Pavel while trying to stake. Starting from nothing, this won't exist. So it should be tolerated.

TheQuantumPhysicist: For Pavel I just recommended using `unwrap_or(0)`. See if you can fix this in a way where this can be just skipped. I don't like the idea that randomness is separate from this. They both should either exist together or not, where genesis uses initial randomness.

Also another thing, there's no reason for randomness to be an Option. It should always exist. I see you have a TODO there though. But for this function, there should always be a source of randomness, whether it's from genesis or otherwise. There's no situation where randomness won't exist.

azarovh: we could add initial epoch data in `process_genesis`

TheQuantumPhysicist: That's a good idea, actually.

muursh (approved):

iljakuklic (approved):

alfiedotwtf:
I'm going to merge once this passes, and have rolled back changes from the staking branch. I realised those changes should remain in that branch as they haven't been reviewed yet.

PR #922, opened 2023-05-28
Add checkpoints support

TheQuantumPhysicist:


common/src/chain/config/checkpoints.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn from(checkpoints: BTreeMap<BlockHeight, Id<GenBlock>>) -> Self {
+        Self::new(checkpoints)
+    }
+}
+
+impl Checkpoints {
+    pub fn new(checkpoints: BTreeMap<BlockHeight, Id<GenBlock>>) -> Self {
+        checkpoints
+            .get(&BlockHeight::new(0))
+            .expect("Checkpoints must have genesis id at height 0");
```
azarovh, line 38: Maybe additionally classify the result to prove that it's genesis at height 0?

azarovh (approved):

common/src/chain/config/checkpoints.rs
TheQuantumPhysicist: Classify requires ChainConfig... if that were available, we wouldn't need this. Passing ChainConfig here isn't possible because it's a circular dependency. `Checkpoints` is part of ChainConfig.

azarovh: I see, then the message should be changed to "Checkpoints must have a genblock id at height 0" because we can't be sure that it's genesis. 

Also this check can be moved to ChainConfig to validate rules for checkpoints.

TheQuantumPhysicist: I don't see the difference with what you suggested in the message... maybe I'm just tired.

Also, checking is circular in chain config. There's no way to do it without setting the value.

azarovh: I mean that check will pass if there will be not genesis block at index 0. But the message suggests otherwise

TheQuantumPhysicist: That's something that can only be validated outside. In something like chainstate. Here, it's not possible unless we pass Genesis itself, which can be lied with too.

azarovh: I agree now that's why I just suggest changing the message only

PR #927, opened 2023-05-30
Wallet staking rebase

alfiedotwtf:
This PR is #913 rebased on top of `master`.

I've made this as a new PR so that @pavel-kokolemin PR doesn't get clobbered, and we'll want to separate fixes on #913 vs continuing to fix here while pulling in from `master`.


pavel-kokolemin:
Thanks, seems to work fine now. I will close this and use the old PR for now (which points to the same commint now).

PR #928, opened 2023-05-31
mempool: Initial orphan pool implementation

iljakuklic:
Introducing orphan pool into mempool.

Still missing (for follow up PRs):
* Tighter p2p integration: If a transaction is received from a peer, it may end up in orphan pool. In case it's later re-verified, we should:
  * Propagate the transaction to other peers if it's valid
  * Ban the peer it originated from if it's invalid, according to the normal mempool ban rules
* A stricter orphan pool acceptance criteria are possible.
  * Processing all outputs: For example, if a transaction has one input with missing UTXO and second one where signature is wrong, it will currently be accepted to orphan pool, because input processing stops at the first error. This transactions should be rejected and the originator peer punished instead.
  * The inputs connecting to other orphan pool transactions are currently not verified either.


chainstate/tx-verifier/src/transaction_verifier/error.rs
```diff
@@ -145,6 +145,66 @@ pub enum ConnectTransactionError {
     OutputTimelockError(#[from] timelock_check::OutputMaturityError),
 }
 
+impl ConnectTransactionError {
+    pub fn is_utxo_missing_error(&self) -> bool {
```
TheQuantumPhysicist, line 149: I would prefer this is moved to the orphan transaction part of the code because it's irrelevant here in the general context. We discourage generalizing such ideas because it can lead to negligence errors. A good example is a "value" of an output, which isn't always clear depending on the output type.

mempool/src/interface/mempool_interface.rs
```diff
@@ -23,17 +23,23 @@ use subsystem::{CallRequest, ShutdownRequest};
...
 pub trait MempoolInterface: Send + Sync {
     /// Add a transaction to mempool
-    fn add_transaction(&mut self, tx: SignedTransaction) -> Result<(), Error>;
+    fn add_transaction(&mut self, tx: SignedTransaction) -> Result<TxStatus, Error>;
 
     /// Get all transactions from mempool
     fn get_all(&self) -> Result<Vec<SignedTransaction>, Error>;
 
-    /// Get a specific transaction from the mempool
+    /// Get a specific transaction from the main mempool
```
TheQuantumPhysicist, line 31: Is there a definition somewhere to what "main mempool" means? If you mean that this excludes orphans, please write it explicitly. Instead of main mempool, we say "from the mempool, excluding orphans".

mempool/src/interface/mempool_interface.rs
```diff
@@ -23,17 +23,23 @@ use subsystem::{CallRequest, ShutdownRequest};
...
 
-    /// Check given transaction is contained in the mempool
+    /// Get a specific transaction from the orphan pool
+    fn orphan_transaction(&self, id: &Id<Transaction>) -> Result<Option<SignedTransaction>, Error>;
+
+    /// Check given transaction is contained in the main mempool
     fn contains_transaction(&self, tx: &Id<Transaction>) -> Result<bool, Error>;
 
+    /// Check given transaction is contained in the main mempool
+    fn contains_orphan_transaction(&self, tx: &Id<Transaction>) -> Result<bool, Error>;
```
TheQuantumPhysicist, line 41: I'm wondering whether it's of value to return a value vs return whether it exists...

mempool/src/pool/orphans/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+}
+
+/// Various indices and lookup maps over orphan transaction entries
+#[derive(Debug)]
+struct TxOrphanPoolMaps {
+    /// Translation from the transaction ID to internal ID
+    by_tx_id: BTreeMap<Id<Transaction>, InternalId>,
+
+    /// Transactions indexed by the insertion time. Useful for removing stale transacitons
```
TheQuantumPhysicist, line 59: transacitons

mempool/src/pool/orphans/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+#[derive(Debug)]
+pub struct TxOrphanPool {
+    /// Transactions in the orphan pool, indexed into by the internal ID
+    transactions: Vec<TxEntry>,
+
+    /// Maps used to index the set of transactions by various criteria
+    maps: TxOrphanPoolMaps,
+
+    /// Maximum orphan pool size in the number of transactions
+    transaction_count_limit: usize,
```
TheQuantumPhysicist, line 111: ConstValue?

mempool/src/interface/mempool_interface.rs
iljakuklic: It's being used in a number of places where the value is not required. Functions returning the value are available too.

mempool/src/pool/orphans/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+        let n_evicted = self.enforce_max_size(*self.transaction_count_limit);
+        n_expired + n_evicted
+    }
+
+    fn remove_expired(&mut self, cur_time: Time) -> usize {
+        // Remove all expired txns
+        let expiry = cur_time.saturating_sub(config::DEFAULT_ORPHAN_TX_EXPIRY_INTERVAL);
+
+        let mut n_expired = 0;
+        while let Some(entry) = self.maps.by_insertion_time.first().filter(|(t, _)| *t < expiry) {
```
TheQuantumPhysicist, line 236: I wonder whether we can use `take_while()` here... seems appropriate.

iljakuklic: I don't think so. The iterator would have to take a shared reference to the map but the map is also being modified in the loop body.

mempool/src/pool/mod.rs
```diff
@@ -647,49 +737,90 @@ impl<M: GetMemoryUsage> Mempool<M> {
...
 impl<M: GetMemoryUsage> Mempool<M> {
-    pub fn add_transaction(&mut self, tx: SignedTransaction) -> Result<(), Error> {
+    pub fn add_transaction(&mut self, tx: SignedTransaction) -> Result<TxStatus, Error> {
         let creation_time = self.clock.get_time();
-        let entry = TxEntry::new(tx, creation_time);
-        self.add_transaction_entry(entry)
+        self.add_transaction_and_descendants(TxEntry::new(tx, creation_time))
+    }
+
+    /// Add given transaction entry and potenrially its descendats sourced from the orphan pool
```
TheQuantumPhysicist, line 768: potenrially

descendats

mempool/src/pool/entry.rs
```diff
@@ -53,6 +53,10 @@ impl TxEntry {
     pub fn size(&self) -> usize {
         serialization::Encode::encoded_size(self.transaction())
     }
+
+    pub fn outpoints(&self) -> impl '_ + Iterator<Item = &OutPoint> + ExactSizeIterator {
+        self.transaction().inputs().iter().map(|i| i.outpoint())
```
TheQuantumPhysicist, line 58: In Gosha's incoming PR, this will not be possible. We'll have accounts too, which won't have outpoints, but account + nonce.

mempool/src/rpc.rs
```diff
@@ -40,7 +55,31 @@ impl MempoolRpcServer for super::MempoolHandle {
...
+            .await,
+        )?;
+
+        Ok(res.map(|(transaction, status)| GetTxResponse {
+            id: tx_id,
+            status,
+            transaction: HexEncoded::new(transaction),
+        }))
+    }
+
```
TheQuantumPhysicist, line 81: Putting this here for the lack of a better location.

Can you add functional tests that prove that the orphan pool functions normally in a node?

mempool/src/interface/mempool_interface.rs
```diff
@@ -23,17 +23,23 @@ use subsystem::{CallRequest, ShutdownRequest};
 
 pub trait MempoolInterface: Send + Sync {
     /// Add a transaction to mempool
-    fn add_transaction(&mut self, tx: SignedTransaction) -> Result<(), Error>;
+    fn add_transaction(&mut self, tx: SignedTransaction) -> Result<TxStatus, Error>;
```
TheQuantumPhysicist, line 26: Just to be clear because I read this in the PR description: If a transaction is not validated, we should never forward it. This is probably a big, sensitive problem that we should double-check what other blockchains do to mitigate this issue. 

I understand it's a trade-off between losing transactions and relaying invalid transactions. Relaying invalid transactions should be avoided at all cost, though. This is how I see, unless a better argument is provided.

TheQuantumPhysicist:
Looking good for the first round. Great work, thank you!

mempool/src/pool/mod.rs
```diff
@@ -78,6 +87,7 @@ pub struct Mempool<M> {
     memory_usage_estimator: M,
```
TheQuantumPhysicist, line 87: That allow(unused) can go, btw.

mempool/src/config.rs
```diff
@@ -36,3 +36,9 @@ pub const MAX_MEMPOOL_SIZE_BYTES: usize = 300_000_000;
 pub const DEFAULT_MEMPOOL_EXPIRY: Duration = Duration::new(336 * 60 * 60, 0);
```
TheQuantumPhysicist, line 36: There's a typo up there, "suring" instead of during.

mempool/src/interface/mempool_interface.rs
iljakuklic: I agree. The comment in the description is concerned with the scenario where a transaction is initially put in orphan pool so not fully validated but later another transaction comes which makes it possible to fully validate the first transaction. We can then forward it since it has been checked for issues (or punish the originating peer if it turns out to be invalid).

iljakuklic:
I removed the code that prevented obviously conflicting transactions from orphan pool. Due to missing inputs, it's not possible to tell which (if any) of the conflicts is the correct one. This is more in line with what Bitcoin does.

mempool/src/pool/tests/orphans.rs
```diff
@@ -0,0 +1,283 @@
...
+            })
+            .collect()
+    };
+
+    let all_tx_ids: Vec<_> = full_tx_sequence.iter().map(|tx| tx.transaction().get_id()).collect();
+
+    // Pick a subset of these transactions, taking each with 90% probability.
+    let tx_subseq_0: Vec<_> = full_tx_sequence
+        .iter()
+        .filter(|_| rng.gen_range(0..100) < 90)
```
OBorce, line 251: you can also use `rng.gen_bool(0.9)`

mempool/src/pool/tests/orphans.rs
```diff
@@ -0,0 +1,280 @@
...
+//                                   +-------+
+//
+// This is a simple diamond transaction graph but it covers many interesting cases.
+//
+#[rstest]
+#[trace]
+#[case::topological_order(
+    Seed::from_entropy(),
+    vec![(0, 0b0001, 0b0000), (1, 0b0011, 0b0000), (2, 0b0111, 0b0000), (3, 0b1111, 0b0000)],
+)]
```
iljakuklic, line 117: I noticed you can name test cases in rstest (the identifier right after `case::`). Pretty useful for documentation and running individual cases from the command line.

TheQuantumPhysicist: Nice! Good to know.

chainstate/tx-verifier/src/transaction_verifier/error.rs
TheQuantumPhysicist: I'm sorry, but this can encourage abusing it just due to its name. We have to be extremely careful with what other devs can do wrong. Please move this to the mempool if this is ought to be used in the mempool.

And just to be clear: It's not like these kinds of mistakes never happen. We had enough instances of people using the `value()` function and `destination()` methods of the output struct/enum incorrectly until we decided that they have to go, to force people to match on the enum, because most devs don't even understand the context and assume the simplest explanation possible from the name of the function. Then tons of clean up had to be done to fix all the context mistakes due to that.

TheQuantumPhysicist (approved):
Looking good. Though please remove that method on the error enum. Otherwise feel free to merge.

TheQuantumPhysicist:
Btw, it just occurred to me; I would really consider it being a good idea if you could wait until Gosha merges #911, because the mempool should support the account model, and I don't think Gosha (alone) can fix that. I think, at least, there should be some discussion on how to reconcile both the mempool orphans' pool and the accounting model.

p2p/src/sync/mod.rs
```diff
@@ -255,5 +255,23 @@ pub async fn subscribe_to_new_tip(
...
+    use mempool::TxStatus;
+    let tx_id = common::primitives::Idable::get_id(transaction.transaction());
+
+    match mempool.call_mut(|m| m.add_transaction(transaction)).await?? {
+        // Transaction accepted to local mempool, propagate it to the peers
+        TxStatus::InMempool => messaging.broadcast_message(SyncMessage::NewTransaction(tx_id)),
+        // We don't know whether the orphan is valid, don't propagate at this point
+        TxStatus::InOrphanPool => Ok(()),
+    }
+}
```
iljakuklic, line 274: Not being very familiar with the `p2p` module structure, not sure this is a sensible place to place this code. Any opinions? @pavel-kokolemin

pavel-kokolemin: > Not being very familiar with the p2p module structure, not sure this is a sensible place to place this code

Works fine for me, thanks.

iljakuklic:
Converting to draft to update for account-based transaction inputs.

iljakuklic:
Updated for the account-based orphans

mempool/src/pool/tests/utils.rs
```diff
@@ -25,11 +25,30 @@ pub use logging::log;
...
 pub use test_utils::{
     mock_time_getter::mocked_time_getter_seconds,
-    random::{make_seedable_rng, Seed},
+    random::{make_seedable_rng, Rng, Seed},
 };
 
 use super::*;
 
+impl TxStatus {
+    /// Fetch status of given instruction from mempool, doing some integriry checks
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Still unfixed

TheQuantumPhysicist:
I think besides that "integriry" typo, we're good to go.

TheQuantumPhysicist (approved):

mempool/src/pool/entry.rs
```diff
@@ -14,44 +14,108 @@
...
+    /// Dependency graph edges this entry requires
+    pub fn requires(&self) -> impl Iterator<Item = TxDependency> + '_ {
+        self.inputs_iter().filter_map(TxDependency::from_input_requires)
+    }
+
+    /// Dependency graph edges this entry provides
+    pub fn provides(&self) -> impl Iterator<Item = TxDependency> + '_ {
+        std::iter::once(TxDependency::Transaction(*self.tx_id()))
+            .chain(self.inputs_iter().filter_map(TxDependency::from_input_provides))
+    }
```
pavel-kokolemin, line 115: Should not `provides` look at the transaction outputs instead?

pavel-kokolemin (approved):

PR #935, opened 2023-06-07
Add StakePoolBalance command (wallet-cli)

pavel-kokolemin:


wallet/wallet-controller/src/sync/tests/mod.rs
```diff
@@ -139,6 +142,12 @@ impl NodeInterface for MockNode {
             .last_common_ancestor_by_id(&first_block, &second_block)
             .unwrap())
     }
+    async fn get_stake_pool_balance(
+        &self,
+        _pool_id: PoolId,
+    ) -> Result<Option<Amount>, Self::Error> {
+        unreachable!()
```
azarovh, line 149: Is it really unreachable or just unimplemented?

azarovh (approved):

wallet/wallet-controller/src/sync/tests/mod.rs
pavel-kokolemin: It's both. It's a mock node for testing the client sync implementation, and only a subset of `NodeInterface` is used here.

TheQuantumPhysicist (approved):

PR #936, opened 2023-06-07
Blockprod unit testing

alfiedotwtf:
Adding Proof of Stake testing to `blockprod`.

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -92,8 +92,7 @@ impl BlockProduction {
         &self.time_getter
     }
 
-    /// Private method used for unit testing
-    #[allow(dead_code)]
+    #[cfg(test)]
     fn set_job_manager(&mut self, job_manager_handle: JobManagerHandle) {
         self.job_manager_handle = job_manager_handle
     }
@@ -488,4 +487,4 @@ fn generate_finalize_block_data(
 }
 
 #[cfg(test)]
-pub mod tests;
+mod tests;
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -46,7 +46,7 @@ use utils::atomics::{RelAtomicBool, SynAtomicU64};
 use utils::once_destructor::OnceDestructor;
 
 use crate::{
-    detail::job_manager::{JobKey, JobManager},
+    detail::job_manager::{JobKey, JobManagerImpl, JobManagerHandle},
     BlockProductionError,
 };
 
@@ -62,7 +62,7 @@ pub struct BlockProduction {
     chainstate_handle: ChainstateHandle,
     mempool_handle: MempoolHandle,
     time_getter: TimeGetter,
-    job_manager: JobManager,
+    job_manager_handle: JobManagerHandle,
     mining_thread_pool: Arc<slave_pool::ThreadPool>,
 }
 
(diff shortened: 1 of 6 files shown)


alfiedotwtf:
As a last step, or to be done sometime after this PR as it's low priority - https://github.com/mintlayer/mintlayer-core/issues/957

blockprod/src/detail/tests.rs
```diff
@@ -185,26 +426,128 @@ async fn stop_job_non_existent_job(#[case] seed: Seed) {
...
+    mock_job_manager
+        .expect_stop_all_jobs()
+        .times(1)
+        .returning(move || Ok(return_value));
+
+    block_production.set_job_manager(mock_job_manager);
+
+    let result = block_production.stop_all_jobs().await;
+
+    assert!(matches!(result, Ok(_expected_value)));
```
azarovh, line 506: matches! is redundant

blockprod/src/detail/tests.rs
```diff
@@ -308,74 +649,47 @@ async fn produce_block_multiple_jobs(#[case] seed: Seed) {
...
+    block_production.set_job_manager(mock_job_manager);
 
-    let jobs_count = block_production.job_manager.get_job_count().await.unwrap();
-    assert_eq!(jobs_count, 0, "Jobs count is incorrect");
+    let mut rng = make_seedable_rng(seed);
+    let job_key = JobKey::new(None, Id::new(H256::random_using(&mut rng)) as Id<GenBlock>);
+
+    let result = block_production.stop_job(job_key).await;
+
+    assert!(matches!(result, Ok(true)));
```
azarovh, line 694: assert_eq

azarovh (approved):

blockprod/src/detail/mod.rs
```diff
@@ -474,4 +480,4 @@ fn generate_finalize_block_data(
 }
 
 #[cfg(test)]
-mod tests;
+pub mod tests;
```
TheQuantumPhysicist, line 483: This should never be done. Exposing test modules guarded by cfg(test) can lead to all kinds of problems, including the false perception that something is public, while it's not.

blockprod/src/detail/mod.rs
```diff
@@ -94,15 +94,21 @@ impl BlockProduction {
         &self.time_getter
     }
 
+    /// Private method used for unit testing
+    #[allow(dead_code)]
+    fn set_job_manager(&mut self, job_manager_handle: JobManagerHandle) {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Once you do that, you won't need to allow(dead_code)

blockprod/src/lib.rs
```diff
@@ -94,18 +94,72 @@ pub fn make_blockproduction(
...
+        vrf::{VRFKeyKind, VRFPrivateKey},
+    };
     use mempool::{MempoolHandle, MempoolSubsystemInterface};
     use subsystem::Manager;
+    use test_utils::random::{make_seedable_rng, Seed};
 
     use super::*;
 
-    pub fn setup_blockprod_test() -> (Manager, Arc<ChainConfig>, ChainstateHandle, MempoolHandle) {
+    const MIN_STAKE_POOL_PLEDGE: Amount = Amount::from_atoms(40_000 * Mlt::ATOMS_PER_MLT);
```
TheQuantumPhysicist, line 127: Please take this value from `ChainConfig::min_stake_pool_pledge`.

iljakuklic (approved):

blockprod/src/detail/tests.rs
alfiedotwtf: If I remove `matches!` here, `assert_eq!` will fail to compile. I've changed it into a vanilla `match`.

iljakuklic: Would it be solved by adding the `Eq` impl to the error type, if it's not there already?

alfiedotwtf: I thought about doing this, but didn't want to pollute the code and binary just for a single test. It's doing what it needs to without the `matches!` now, but if `Eq` is warranted, then I'm happy to add it.

alfiedotwtf: Cool, no problem.

PR #938, opened 2023-06-07
Add initial implementation of Utxo selector

OBorce:
- fees and weight are set to 0 for now

### DIFF
--- a/common/src/chain/tokens/tokens_utils.rs
+++ b/common/src/chain/tokens/tokens_utils.rs
@@ -15,7 +15,7 @@
 
 use super::TokenId;
 use crate::{
-    chain::{Transaction, TxInput, TxOutput},
+    chain::{Transaction, TxOutput},
     primitives::id::hash_encoded,
 };
 
@@ -23,10 +23,6 @@ pub fn token_id(tx: &Transaction) -> Option<TokenId> {
     Some(hash_encoded(tx.inputs().get(0)?))
 }
 
-pub fn token_id_from_input0(input0: &TxInput) -> TokenId {
-    hash_encoded(input0)
-}
-
 pub fn get_tokens_issuance_count(outputs: &[TxOutput]) -> usize {
-    outputs.iter().filter(|&output| output.is_token_or_nft_issuence()).count()
+    outputs.iter().filter(|&output| output.is_token_or_nft_issuance()).count()
 }
--- a/common/src/chain/tokens/tokens_utils.rs
+++ b/common/src/chain/tokens/tokens_utils.rs
@@ -13,9 +13,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::{OutputValue, TokenData, TokenId};
+use super::TokenId;
 use crate::{
-    chain::{Transaction, TxOutput},
+    chain::{Transaction, TxInput, TxOutput},
     primitives::id::hash_encoded,
 };
 
@@ -23,27 +23,10 @@ pub fn token_id(tx: &Transaction) -> Option<TokenId> {
     Some(hash_encoded(tx.inputs().get(0)?))
 }
 
-pub fn is_tokens_issuance(output_value: &OutputValue) -> bool {
-    match output_value {
-        OutputValue::Coin(_) => false,
-        OutputValue::Token(token_data) => match **token_data {
-            TokenData::TokenIssuance(_) | TokenData::NftIssuance(_) => true,
-            TokenData::TokenTransfer(_) => false,
-        },
-    }
+pub fn token_id_from_input0(input0: &TxInput) -> TokenId {
+    hash_encoded(input0)
 }
 
 pub fn get_tokens_issuance_count(outputs: &[TxOutput]) -> usize {
-    outputs
-        .iter()
-        .filter(|&output| match output {
-            TxOutput::Transfer(v, _) | TxOutput::LockThenTransfer(v, _, _) | TxOutput::Burn(v) => {
-                is_tokens_issuance(v)
-            }
-            TxOutput::CreateStakePool(_, _)
-            | TxOutput::ProduceBlockFromStake(_, _)
-            | TxOutput::CreateDelegationId(_, _)
-            | TxOutput::DelegateStaking(_, _) => false,
-        })
-        .count()
+    outputs.iter().filter(|&output| output.is_token_or_nft_issuence()).count()
 }
--- a/common/src/primitives/amount.rs
+++ b/common/src/primitives/amount.rs
@@ -143,12 +143,6 @@ impl Amount {
     }
 }
 
-impl Default for Amount {
-    fn default() -> Self {
-        Self::ZERO
-    }
-}
-
 impl std::ops::Add for Amount {
     type Output = Option<Self>;
 
--- a/common/src/primitives/amount.rs
+++ b/common/src/primitives/amount.rs
@@ -143,6 +143,12 @@ impl Amount {
     }
 }
 
+impl Default for Amount {
+    fn default() -> Self {
+        Self::ZERO
+    }
+}
+
 impl std::ops::Add for Amount {
     type Output = Option<Self>;
 
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -22,7 +22,7 @@ use crate::account::utxo_selector::{select_coins, OutputGroup};
 use crate::key_chain::{
     make_path_to_vrf_key, vrf_from_private_key, AccountKeyChain, KeyChainError,
 };
-use crate::send_request::{make_address_output, make_stake_output};
+use crate::send_request::{make_address_output, make_address_output_token, make_stake_output};
 use crate::{SendRequest, WalletError, WalletResult};
 use common::address::Address;
 use common::chain::signature::inputsig::standard_signature::StandardInputSignature;
@@ -52,6 +52,7 @@ use wallet_types::wallet_tx::{BlockData, TxData, TxState};
 use wallet_types::{AccountId, AccountInfo, AccountWalletTxId, KeyPurpose, WalletTx};
 
 use self::output_cache::OutputCache;
+use self::utxo_selector::PayFee;
 
 pub struct Account {
     chain_config: Arc<ChainConfig>,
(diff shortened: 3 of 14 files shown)


common/src/chain/tokens/tokens_utils.rs
```diff
@@ -13,37 +13,20 @@
...
 
-pub fn is_tokens_issuance(output_value: &OutputValue) -> bool {
-    match output_value {
-        OutputValue::Coin(_) => false,
-        OutputValue::Token(token_data) => match **token_data {
-            TokenData::TokenIssuance(_) | TokenData::NftIssuance(_) => true,
-            TokenData::TokenTransfer(_) => false,
-        },
-    }
+pub fn token_id_from_input0(input0: &TxInput) -> TokenId {
```
TheQuantumPhysicist, line 26: I think this should be a private function. Exposing it makes it possible to abuse it. The function `token_id` must be used instead.

common/src/primitives/amount.rs
```diff
@@ -143,6 +143,12 @@ impl Amount {
     }
 }
 
+impl Default for Amount {
+    fn default() -> Self {
+        Self::ZERO
+    }
+}
```
TheQuantumPhysicist, line 150: Please remove this, and use `Amount::ZERO`

wallet/src/account/utxo_selector/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+    primitives::{signed_amount::SignedAmount, Amount},
+};
+use crypto::random::{make_pseudo_rng, Rng, SliceRandom};
+
+const TOTAL_TRIES: u32 = 100_000;
+
+/// A group of UTXOs paid to the same output script.
+/// This helps reduce privacy leaks resulting from address reuse.
+#[derive(Clone)]
+pub struct OutputGroup {
```
TheQuantumPhysicist, line 32: Please move this to its own submodule

wallet/src/account/utxo_selector/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+    /// The list of UTXOs contained in this output group.
+    outputs: Vec<(UtxoOutPoint, TxOutput)>,
+    /// the total amount of the outputs in this group
+    value: Amount,
+    /// The fee to spend these UTXOs at the effective feerate.
+    fee: Amount,
+    /// The fee to spend these UTXOs at the long term feerate.
+    long_term_fee: Amount,
+    /// Total weight of the UTXOs in this group.
+    /// Non-witness data + witness data * Factor
```
TheQuantumPhysicist, line 42: We don't use any kind of weighting... this is segwit bitcoin stuff. We can replace this with a normal size, I believe.

wallet/src/account/utxo_selector/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+/// This helps reduce privacy leaks resulting from address reuse.
+#[derive(Clone)]
+pub struct OutputGroup {
+    /// The list of UTXOs contained in this output group.
+    outputs: Vec<(UtxoOutPoint, TxOutput)>,
+    /// the total amount of the outputs in this group
+    value: Amount,
+    /// The fee to spend these UTXOs at the effective feerate.
+    fee: Amount,
+    /// The fee to spend these UTXOs at the long term feerate.
```
TheQuantumPhysicist, line 39: The English here is a little broken "The fee to spend these UTXOs"... I'm not sure I understand what this means.

Same applies for the long_term_fee below.

wallet/src/account/utxo_selector/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+    long_term_fee: Amount,
+    /// Total weight of the UTXOs in this group.
+    /// Non-witness data + witness data * Factor
+    /// The weight metric was introduced to ensure that the larger witness data doesn't dominate the block size limit
+    weight: u32,
+}
+
+/// Should we pay fee with this currency or not in the case we pay the total fees with another
+/// currency
```
TheQuantumPhysicist, line 48: Please add ". Currency is a generalized term for coin or token.".

TheQuantumPhysicist (approved):
Looking good. Please fix the minor issues I mentioned and give others a chance to review this.

wallet/src/account/utxo_selector/mod.rs
OBorce: it refers to the amount of fee that we will need to pay if we include these Utxos in the transaction based on the size in bytes they will take up.
The normal fee should be based on the current fee rate (maybe there is a spike or dip in fee), and the long-term fee is the estimated fee rate overall.

OBorce: yes

common/src/chain/tokens/tokens_utils.rs
OBorce: fixed

wallet/src/account/utxo_selector/mod.rs
OBorce: moved to a submodule

wallet/src/account/utxo_selector/mod.rs
OBorce: changed the comment, it is still 0 for now

wallet/src/account/utxo_selector/mod.rs
TheQuantumPhysicist: I would then make the comments "The fee cost of these UTXOs..."

wallet/src/account/utxo_selector/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+            }
+
+            if target_reached {
+                break;
+            }
+        }
+    }
+}
+
+/// This is the Branch and Bound Coin Selection algorithm designed by Murch. It searches for an input
```
pavel-kokolemin, line 432: Maybe we should update the license notice, since some of the code comes from Bitcoin Core?

wallet/src/account/mod.rs
```diff
@@ -185,17 +272,19 @@ impl Account {
...
-            .get_utxos(UtxoType::Transfer.into())
-            .into_iter()
-            .map(|(outpoint, txo)| (outpoint, txo.clone()))
-            .collect::<Vec<(UtxoOutPoint, TxOutput)>>();
-
-        let input0 = utxos.get(0).ok_or(WalletError::NoUtxos)?;
-        let pool_id = pos_accounting::make_pool_id(&input0.0);
+        // the first UTXO is needed in advance to calculate pool_id, so just make a dummy Burn
+        // output and then replace it with the stake output
+        let request = SendRequest::new().with_outputs([TxOutput::Burn(OutputValue::Coin(amount))]);
```
pavel-kokolemin, line 277: The `Burn` output takes up less space than the `CreateStakePool` output, so once it's replaced, the network fee calculation may be off. So I think it would be better to use a `CreateStakePool` output with a dummy `pool_id`.

pavel-kokolemin (approved):
Nice work!

wallet/src/account/mod.rs
OBorce: noticed also that replacing the outputs later with the proper pool_id will remove any change output that might have been added. For now, I changed it to find and modify the output, and plan to fix it properly in another PR when the fees are actually calculated properly. What do you think @pavel-kokolemin ?

wallet/src/account/utxo_selector/mod.rs
OBorce: I added the fallowing section to the license notice of this file:
```
// Original code copyright (c) 2011-2022 The Bitcoin Core developers
// Distributed under the MIT software license, see the accompanying
// file COPYING or http://www.opensource.org/licenses/mit-license.php.
```

azarovh (approved):
Looks good to me. 
One comment on the general style: number of mutable variables and function arguments makes me uncomfortable because it feels like anything can be modified at any point and thus it’s harder to reason about the code.

wallet/src/account/utxo_selector/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+    let is_feerate_high = utxo_pool[0].fee > utxo_pool[0].long_term_fee;
+    let mut max_tx_weight_exceeded = false;
+
+    // Depth First search loop for choosing the UTXOs
+    let mut utxo_pool_index = 0;
+    for _ in 0..TOTAL_TRIES {
+        // Conditions for starting a backtrack
+        let backtrack = if
+        // Cannot possibly reach target with the amount remaining in the curr_available_value.
+        (curr_value + curr_available_value).expect("total has been check to not overflow") < selection_target
```
iljakuklic, line 499: `checked`

wallet/src/account/utxo_selector/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+            curr_value = (curr_value - utxo.get_effective_value(pay_fees)).expect("no underflow");
+            curr_waste = utxo
+                .fee
+                .into_signed()
+                .and_then(|fee| {
+                    utxo.long_term_fee.into_signed().and_then(|long_term_fee| fee - long_term_fee)
+                })
+                .and_then(|fee_difference| curr_waste - fee_difference)
+                .ok_or(UtxoSelectorError::AmountArithmeticError)?;
+            curr_selection_weight -= utxo.weight;
```
iljakuklic, line 564: I would say it would probably be nicer to abstract these into a separate struct that represents a partial solution. Any time a utxo is added to or removed from the partial solution, it updates all the metrics like the value, waste, weight etc. Maybe it would even be possible to reuse it in multiple solvers but not sure about that. It seems the main algorithm is made less clear by being interleaved with substantial amount of bookkeeping.

iljakuklic (commented):
Good stuff. A little late to the party but anyway:

wallet/src/account/mod.rs
pavel-kokolemin: Great, thanks.

wallet/src/account/mod.rs
```diff
@@ -110,54 +116,135 @@ impl Account {
...
+                .ok_or(WalletError::NotEnoughUtxo(input_amount, output_amount))?;
+            if change_amount > Amount::ZERO {
+                let change_address = self.get_new_address(db_tx, KeyPurpose::Change)?;
+                let change_output = match currency {
+                    Currency::Coin => make_address_output(change_address, change_amount)?,
+                    Currency::Token(token_id) => {
+                        make_address_output_token(change_address, change_amount, token_id)?
+                    }
+                };
+                request = request.with_outputs([change_output]);
```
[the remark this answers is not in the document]
OBorce: there is a TODO for that in the process_send_request function

OBorce:
Nice suggestions as always @iljakuklic, I will correct them in the next PR.

PR #939, opened 2023-06-07
Mempool memory usage tracker

iljakuklic:
Notes:
* This is built on top of #928
* Contains multiple lines of defences to prevent the users from forgetting to update the memory usage when objects tracked for memory usage are changed, created, or deleted.
* This is best effort. Not guaranteed to be accurate but should be reasonably close.
* Most of the substantive changes are in `mempool/src/pool/store/*.rs`

Todo:
* [x] Try to get rid of interior mutability in the `Tracker` type. Should be possible but failed to achieve so far
* [x] More tests

### DIFF
--- a/mempool/src/pool/store/mem_usage.rs
+++ b/mempool/src/pool/store/mem_usage.rs
@@ -1,0 +1,449 @@
+// Copyright (c) 2022 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Estimate and track memory usage taken by data structures.
+
+use std::{cmp, mem};
+
+use common::chain::{
+    signature::inputsig::InputWitness, stakelock::StakePoolData, SignedTransaction, TxInput,
+    TxOutput,
+};
+use logging::log;
+
+use super::TxMempoolEntry;
+
+/// Structure that stores the current memory usage and keeps track of its changes
+#[derive(Debug)]
+pub struct MemUsageTracker {
+    current_usage: usize,
+    peak_usage: usize,
+}
+
+impl MemUsageTracker {
+    pub fn new() -> Self {
+        Self {
+            current_usage: 0,
+            peak_usage: 0,
+        }
+    }
+
+    pub fn get_usage(&self) -> usize {
+        self.current_usage
+    }
+
+    fn add(&mut self, amount: usize) {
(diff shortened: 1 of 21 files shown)


mempool/src/pool/store/mem_usage.rs
```diff
@@ -0,0 +1,446 @@
...
+    signature::inputsig::InputWitness, stakelock::StakePoolData, SignedTransaction, TxInput,
+    TxOutput,
+};
+use logging::log;
+
+use super::TxMempoolEntry;
+
+/// Structure that stores the current memory usage and keeps track of its changes
+#[derive(Debug)]
+pub struct Tracker {
```
TheQuantumPhysicist, line 30: MemUsageTracker?

mempool/src/pool/store/mem_usage.rs
```diff
@@ -0,0 +1,446 @@
...
+}
+
+impl MemoryUsage for SignedTransaction {
+    /// Only data included indirectly (via pointers). The actual object usage is already contained
+    /// in the B-tree map usage.
+    fn indirect_memory_usage(&self) -> usize {
+        let ins = self.inputs().indirect_memory_usage();
+        let outs = self.outputs().indirect_memory_usage();
+        let sigs = self.signatures().indirect_memory_usage();
+        ins + outs + sigs
```
TheQuantumPhysicist, line 333: This looks like it needs a macro... though it's not a priority.

TheQuantumPhysicist (approved):
Looking good. Though I'm not sure there's an eviction algorithm implemented. If there's, please write some functional test for it.

mempool/src/pool/store/mem_usage.rs
TheQuantumPhysicist: Cool!

iljakuklic:
I have disabled the memory-based transaction eviction from this PR so it can be reviewed and merged separately. Memory consumption tracking happens and is being tested. However, transactions are not evicted on that basis to avoid the issues with transaction verifier being out of sync with mempool. While this allows an attacker to exhaust node memory, it is the case on master already, so it does not make things worse.

I continue working on the tx verifier stuff so this can be fixed asap. That I think is better addressed as a separate PR.

TheQuantumPhysicist (approved):
Based on the comments, I'm approving this again since nothing has changed.

alfiedotwtf (approved):

mempool/src/pool/store/mem_usage.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+}
+
+impl<T: MemoryUsage> MemoryUsage for Box<T> {
+    fn indirect_memory_usage(&self) -> usize {
+        total_memory_usage::<T>(self.as_ref())
+    }
+}
+
+impl<T> MemoryUsage for common::primitives::Id<T> {
```
azarovh, line 308: use `impl_no_indirect_memory_usage` for this?

azarovh (approved):
nice!

mempool/src/pool/store/mem_usage.rs
iljakuklic: It's a stupidly simple macro that does not handle generics. Not sure it's worth adding the support for generics just for this one thing.

azarovh: ah ok nevermind then

Issue #888, opened 2023-05-11
Remove the shutdown flag in the p2p


stanislav-tkach:
Currently the shutdown flag is needed because the p2p parts (tasks) are tightly coupled together and there is no easy way to organize a clear shutdown process. The tasks communicate using channels and when one of the tasks is shut down, it closes its end of a channel and that is considered an error by other tasks. The `ChannelClosed` error is used for that. When the shutdown flag is set such errors are basically ignored, otherwise an error is reported.

Ideally a graceful termination shouldn't require this flag and a shutdown order should be established. For example, backend can signal to the peer and the sync managers to shut down and wait for the corresponding tasks to be finished.

RGafiyatullin:
@discosultan , sorry I didn't notice you were the assignee for that issue. I didn't mean to hijack the issue.

discosultan:
@RGafiyatullin No worries at all, I only assigned myself yesterday :) will review your PR shortly. Thanks for the contribution!

TheQuantumPhysicist:
Related: #888

TheQuantumPhysicist:
Related: #943

PR #943, opened 2023-06-08
#888: P2P graceful shutdown

RGafiyatullin:
# *DISCLAIMER*

I haven't succeeded to run a test-net on my own.
I've just run the tests, and tried to launch and shutdown a single node.

# Intro

What started as just removing a simple AtomicBool-flag, turned into a research on the subject of how to better organise the communication and lifecycle management of concurrent activities of a subsystem.

So I can't say this PR addresses #888 exactly as its description requires. But still, may be worth a look.

# Notes

#### `void@1.0` — an unnecessary dependency

What this crate tries to deliver, is already brought by `std::convert::Infallible` or `futures::never::Never`.

#### NetworkingService, API change

* get rid of the `AtomicBool`-flag;
* be more lenient as to what to accept as the shutdown-signal;
* do not spawn, return a future instead (let the calling party decide how to compose the tasks).

#### Introduce `P2pError::Cancelled`

Signalize that the task terminated "against its will".

#### Usage of AtomicBool-flag in tests

I see a problem in the usage of the same instance of such flag by several instances of p2p-subsystem.
The flag may be removed with this PR, but the tests are to be reviewed after that.

# Conclusion

All over the project the tasks are spawned with little "discipline", namely:
* the dependencies between the tasks are not clear;
* there is no well-defined shutdown behaviour;
* multiple input-channels per single activity make it difficult to handle failures;
* sending `Arc`-ed closures to serialise access to some shared capability — I'd talk about that :) .


p2p/src/run_p2p.rs
```diff
@@ -0,0 +1,385 @@
...
+    use crate::Result;
+
+    type Killswitch = oneshot::Sender<Never>;
+
+    pub fn spawn_with_killswitch<F>(f: F) -> (BoxFuture<'static, Result<()>>, Killswitch)
+    where
+        F: Future<Output = Result<()>> + Send + 'static,
+    {
+        let (ks_tx, ks_rx) = killswitch_chan();
+        let jh = tokio::spawn(async move {
```
RGafiyatullin, line 296: I tried not spawning these activities, but instead "selecting" the collection of task-futures.

This did not work smoothly: the peer-manager and sync-manager do not get polled after the p2p-loop returns from the first `select_all`.

The networking-service backend is polled normally though.

Maybe somewhere in the peer-manager and sync-manager event-loops the waker call is "lost".

p2p/src/run_p2p.rs
```diff
@@ -0,0 +1,385 @@
...
+        assert_eq!(services.len(), shutdown_switches.len());
+
+        log::trace!("run-multiple [len: {}]", services.len());
+
+        let (outcome, idx, _rest) = futures::future::select_all(services.iter_mut()).await;
+
+        if let Err(reason) = &outcome {
+            log::warn!("Service[{}] shut down with error: {}", idx, reason);
+        }
+
```
RGafiyatullin, line 321: The idea is that the tasks with higher indices in the tasks-vec depend on those that have the lower indices.
Therefore if a task terminates, everything "to the right" of it has little chances to work correctly, so it is shut down harshly.

The tasks "to the left" are asked to shut down nicely sequentially.

p2p/src/run_p2p.rs
RGafiyatullin: Now that everything is spawned, it would probably make sense to use `JoinHandle::abort(&self)` for last-resort cancellation of a task :\

TheQuantumPhysicist:
Dear Roman:

Thank you very much for your contribution. We really appreciate it, and it's a great gesture and shows your expertise, and I'm actually impressed. 

Besides that, I would like to point to an issue with continuing with this. Usually we communicate by voice to settle discussions (especially complex discussions) for topics like this one. Discussing only on Github can be easily riddled with misunderstandings as not all people have the same skills in writing. I'm kind of worried how we can have such a complex discussion here, and hence, given that you're absolutely not required to work for Mintlayer for free, I'd like to ask @discosultan to continue working on this with his own discretion, and then arrive to any result that may be up for discussion. But as it stands, I don't really know how to continue with this PR in its current form since it's incomplete.

Does this make sense to you? If not, please let me know. Otherwise, @discosultan may "steal" (😄) some of your code.  

Best regards.

RGafiyatullin:
@TheQuantumPhysicist , @discosultan , I have no objections whatsoever.

Issue #947, opened 2023-06-09
.cookie file gets deleted when running a second node daemon for the same chain type


discosultan:
When starting a node daemon, it generates a `~/.mintlayer/<chain_type>/.cookie` file that contains the credentials for Basic Authentication used by JSON-RPC. Running a second node daemon with the same configuration will obviously fail for various reasons. For example, the port for JSON-RPC HTTP server is already in use. The problem is that as the second node fails to start, it also deletes the cookie file which is required for the first node to operate.

PR #949, opened 2023-06-09
Minor wallet updates

pavel-kokolemin:
Fix https://github.com/mintlayer/mintlayer-core/issues/947

### DIFF
--- a/subsystem/src/manager.rs
+++ b/subsystem/src/manager.rs
@@ -84,7 +84,7 @@ pub struct Manager {
     shutting_down_tx: mpsc::UnboundedSender<()>,
     shutting_down_rx: mpsc::UnboundedReceiver<()>,
 
-    // List of subsystem tasks.
+    // List of subsystem tasks
     subsystems: Vec<SubsystemInfo>,
 }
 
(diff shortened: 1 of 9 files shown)


subsystem/src/manager.rs
```diff
@@ -84,8 +84,11 @@ pub struct Manager {
     shutting_down_tx: mpsc::UnboundedSender<()>,
     shutting_down_rx: mpsc::UnboundedReceiver<()>,
 
-    // List of subsystem tasks.
+    // List of subsystem tasks
     subsystems: Vec<SubsystemInfo>,
+
+    // Stored handle to the lock file
+    lock_file: Option<File>,
```
TheQuantumPhysicist, line 91: I think this shouldn't be part of subsystems... but should be part of initialization.

TheQuantumPhysicist:
Besides the file lock problem, I don't have a problem with this PR. We should find a better place for it.

subsystem/src/manager.rs
pavel-kokolemin: OK, updated.

iljakuklic (approved):

TheQuantumPhysicist (approved):

PR #951, opened 2023-06-11
Implementation of a "simplified" wrapper for atomics

ImplOfAnImpl:
Here is an implementation of the wrapper for atomics that we've discussed earlier. 

The code has a couple of FIXMEs, those are basically my questions to reviewers. I'm not sure whether I should implement them, turn into proper TODOs or just delete them.

For now, I only used "simplified" atomics in a couple of places, mainly as an example (if this PR gets approved, I plan to change (most of) the other places too).
Note that the replaced AtomicU64 is the one that I've mentioned in our discussion - it used to use SeqCst on loads and Relaxed on stores, which didn't make much sense. I've replaced it with RelAtomicU64, which is the "relaxed" variant.

Edit: I forgot to mention, I also had to introduce traits for standard atomic types, similar to what the crate "atomic_traits" has, so that I could implement them for "loom" atomics too and then use them in my implementation of the wrapper.
Probably, we can get rid of the "atomic_traits" crate now. (Note though that the current version of my atomic traits doesn't support AtomicPtr. But I suppose it's not something we will miss).

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -15,10 +15,7 @@
 
 pub mod job_manager;
 
-use std::sync::{
-    atomic::{AtomicBool, AtomicU64, Ordering},
-    mpsc, Arc,
-};
+use std::sync::{mpsc, Arc};
 
 use chainstate::{chainstate_interface::ChainstateInterface, ChainstateHandle, PropertyQueryError};
 use chainstate_types::{
@@ -45,6 +42,7 @@ use mempool::{
     MempoolHandle,
 };
 use tokio::sync::oneshot;
+use utils::atomics::{RelAtomicBool, SynAtomicU64};
 use utils::once_destructor::OnceDestructor;
 
 use crate::{
@@ -245,7 +243,7 @@ impl BlockProduction {
         transactions_source: TransactionsSource,
         custom_id: Option<Vec<u8>>,
     ) -> Result<(Block, oneshot::Receiver<usize>), BlockProductionError> {
-        let stop_flag = Arc::new(AtomicBool::new(false));
+        let stop_flag = Arc::new(RelAtomicBool::new(false));
         let tip_at_start = self.pull_best_block_index().await?;
 
         let (job_key, mut cancel_receiver) =
@@ -274,7 +272,7 @@ impl BlockProduction {
                 .add_int_seconds(1)
                 .ok_or(ConsensusCreationError::TimestampOverflow(tip_timestamp, 1))?;
 
-            Arc::new(AtomicU64::new(tip_plus_one.as_int_seconds()))
+            Arc::new(SynAtomicU64::new(tip_plus_one.as_int_seconds()))
         };
 
         let max_block_timestamp = {
@@ -292,12 +290,11 @@ impl BlockProduction {
         loop {
             {
                 // If the last timestamp we tried on a block is larger than the max range allowed, no point in continuing
-                let last_used_block_timestamp = BlockTimestamp::from_int_seconds(
-                    last_timestamp_seconds_used.load(Ordering::SeqCst),
-                );
+                let last_used_block_timestamp =
+                    BlockTimestamp::from_int_seconds(last_timestamp_seconds_used.load());
 
                 if last_used_block_timestamp >= max_block_timestamp {
-                    stop_flag.store(true, Ordering::Relaxed);
+                    stop_flag.store(true);
                     return Err(BlockProductionError::TryAgainLater);
                 }
             }
@@ -350,7 +347,7 @@ impl BlockProduction {
 
             tokio::select! {
                 _ = cancel_receiver.recv() => {
-                    stop_flag.store(true, Ordering::Relaxed);
+                    stop_flag.store(true);
 
                     // This can fail if the mining thread has already finished
                     let _ended = ended_receiver.recv();
@@ -375,13 +372,25 @@ impl BlockProduction {
         }
     }
 
+    // TODO: here, `block_timestamp_seconds` is a scary thing because, by being Syn, it might
+    // imply that we perform thread synchronization through it. Which would be a bad thing
+    // to do, because thread synchronization via atomics is too low-level and non-trivial
+    // to implement correctly. Normally, it should be properly encapsulated, but here we
+    // share the variable across packages, passing it to `consensus::finalize_consensus_data`.
+    // So it's better to get rid of it ASAP.
+    // (Note that we don't really do any thread synchronization through it currently; we made
+    // it "Syn" just in case, for extra peace of mind.)
+    // One way of removing it would be to pass the initial value via a non-atomic parameter and
+    // return the updated value back; in `finalize_consensus_data` and its callees it can
+    // be done simply via the functions' return values. And here in `spawn_block_solver` we
+    // already have a one-shot `result_sender`, which may be used for that purpose.
     #[allow(clippy::too_many_arguments)]
     fn spawn_block_solver(
         &self,
         current_tip_index: &GenBlockIndex,
-        stop_flag: Arc<AtomicBool>,
+        stop_flag: Arc<RelAtomicBool>,
         block_body: &BlockBody,
-        block_timestamp_seconds: Arc<AtomicU64>,
+        block_timestamp_seconds: Arc<SynAtomicU64>,
         finalize_block_data: FinalizeBlockInputData,
         consensus_data: ConsensusData,
         ended_sender: mpsc::Sender<()>,
@@ -395,8 +404,7 @@ impl BlockProduction {
             let merkle_proxy =
                 block_body.merkle_tree_proxy().map_err(BlockCreationError::MerkleTreeError)?;
 
-            let block_timestamp =
-                BlockTimestamp::from_int_seconds(block_timestamp_seconds.load(Ordering::SeqCst));
+            let block_timestamp = BlockTimestamp::from_int_seconds(block_timestamp_seconds.load());
 
             let mut block_header = BlockHeader::new(
                 current_tip_index.block_id(),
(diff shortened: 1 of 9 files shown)


utils/src/atomics/simple_atomic.rs
```diff
@@ -0,0 +1,235 @@
...
+use crate::concurrency_impl::sync::atomic::Ordering;
+
+// FIXME: do we also need SimpleAtomic traits or the fact that the wrapper is generic is enough?
+
+/// A predefined set of orderings to use by the [Atomic] type.
+pub trait Orderings {
+    const ORD_LOAD: Ordering;
+    const ORD_STORE: Ordering;
+    const ORD_LOAD_STORE: Ordering;
+}
```
alfiedotwtf, line 50: Nice

alfiedotwtf (approved):

utils/src/atomics/atomic_traits.rs
```diff
@@ -0,0 +1,191 @@
...
+//! here too.
+//! We also omit deprecated functions, such as `compare_and_swap`, and don't support `AtomicPtr`.
+
+use crate::concurrency_impl::sync::atomic::{
+    AtomicBool, AtomicI16, AtomicI32, AtomicI64, AtomicI8, AtomicIsize, AtomicU16, AtomicU32,
+    AtomicU64, AtomicU8, AtomicUsize, Ordering,
+};
+
+/// A trait that contains basic atomic operations common for all atomic types.
+pub trait Atomic {
```
TheQuantumPhysicist, line 29: I'm a little surprised we're not using `atomic_traits::Atomic`... it's already in utils.

blockprod/src/detail/mod.rs
```diff
@@ -379,9 +376,9 @@ impl BlockProduction {
     fn spawn_block_solver(
         &self,
         current_tip_index: &GenBlockIndex,
-        stop_flag: Arc<AtomicBool>,
+        stop_flag: Arc<RelAtomicBool>,
         block_body: &BlockBody,
-        block_timestamp_seconds: Arc<AtomicU64>,
+        block_timestamp_seconds: Arc<RelAtomicU64>,
```
TheQuantumPhysicist, line 381: This has to be `SeqCst` or at least `AcqRel`.

consensus/src/lib.rs
```diff
@@ -179,8 +177,8 @@ pub fn finalize_consensus_data(
     chain_config: &ChainConfig,
     block_header: &mut BlockHeader,
     block_height: BlockHeight,
-    block_timestamp_seconds: Arc<AtomicU64>,
-    stop_flag: Arc<AtomicBool>,
+    block_timestamp_seconds: Arc<RelAtomicU64>,
```
TheQuantumPhysicist, line 180: Same, not relaxed.

utils/src/atomics/simple_atomic.rs
```diff
@@ -0,0 +1,235 @@
...
+        self.0.fetch_max(val, Ord::ORD_LOAD_STORE)
+    }
+
+    pub fn fetch_min(&self, val: T) -> T {
+        self.0.fetch_min(val, Ord::ORD_LOAD_STORE)
+    }
+}
+
+// FIXME: here the trait bounds could be simplified and relaxed if we introduced a separate trait
+// `ConstructibleFrom` and move the AtomicTrait's `new` function there. Is it worth the effort?
```
TheQuantumPhysicist, line 148: I would say no.

utils/src/atomics/simple_atomic.rs
```diff
@@ -0,0 +1,235 @@
...
+//! but, on the other hand, they must still agree with each other. Using a simplified atomic
+//! in this case basically encodes the purpose of the atomic in its type's name and makes
+//! the orderings a part of the module's interface.
+
+use crate::atomics::{atomic_traits::HasStdAtomic, AtomicNumTrait, AtomicTrait};
+use std::marker::PhantomData;
+
+use crate::concurrency_impl::sync::atomic::Ordering;
+
+// FIXME: do we also need SimpleAtomic traits or the fact that the wrapper is generic is enough?
```
TheQuantumPhysicist, line 43: It's already complicated as it's. We wrapped two possible synchronization models. That's good enough, I'd say.

TheQuantumPhysicist (approved):
Besides that wrong relaxed usage, I think this is fine.

utils/src/atomics/atomic_traits.rs
ImplOfAnImpl: Initially I was using them, but then I found out that they are not implemented for `loom`'s atomic types. So the alternative would be to abandon the idea of using `loom`'s atomics.

blockprod/src/detail/mod.rs
ImplOfAnImpl: I don't think it's really needed, because there is no synchronization on `block_timestamp_seconds` AKA `last_timestamp_seconds_used`. The actual synchronization points should be the call of `tokio::select!` (the `result_receiver` part) in `BlockProduction::produce_block_with_custom_id` and `result_sender.send(...)` in `BlockProduction::spawn_block_solver`. So the real synchronization is actually encapsulated inside `BlockProduction`, which is a good thing. And I would argue that making `block_timestamp_seconds/last_timestamp_seconds_used` a "SynAtomic" would introduce code smell, because it will suggest that we have a low-level synchronization primitive that is a part of a package's public interface.
But I definitely understand your concerns, because even if it's not used for synchronization, this fact is still very non-obvious. So, I don't argue about making that parameter a SynAtomic, but I also want at least to write a TODO to remove it completely, preferably ASAP. Ideally, Pavel should remove it because he added it relatively recently, but if he's busy and there are no objections, I could try to do it myself (but I'll have to ask some stupid questions in the process).

TheQuantumPhysicist: The problem, because of which I think we should use SynAtomic, is that the main thread (where `produce_block_with_custom_id` resides) should learn that the atomic value has been updated in the new thread. I would rather use an atomic fence, but it's not worth it. What I fear is, through the lack of synchronization, what will happen is that the `spawn_block_solver` will update the value of `last_timestamp_seconds_used`, and then the main thread won't notice due to the lack of enforcement of cache-reloads. The current implementation doesn't seem to take into account that we have to continue staking from where we left of, so `last_timestamp_seconds_used` should probably be updated in the main thread and returned to the request sender (rpc, for example). But again, one problem at a time. 

Like I said before: The fear is always that someone looks at the code at face value and changes something without us noticing. So making this SynAtomic is a good guarantee for future changes. Feel free to add a TODO for when we're done with everything related, to check this problem again.

TheQuantumPhysicist: Just to be clear: I don't really care if the synchronization will happen because some other part/primitive of the code has sufficient synchronization semantics. This feels to me like playing with fire and creates complex interdependencies, especially that depending on the program, it may be in another memory page, which may or may not synchronize. Having worked with emulators, I've seen how complex the memory model in CPUs is compared to the naive all-or-nothing mental model we usually have in mind. Yet I still assert my ignorance in this topic and I'll never learn enough about it. 

Having said that, `stop_flag` is a whole other discussion. Like I mentioned before, `stop_flag` is allowed to have all the delay it needs because the worst thing that can happen is a delay in response. On the other hand, lack of synchronization for `last_timestamp_seconds_used` can lead to incorrect value in the main thread. Almost impossible to see in real life, thanks to x86_64's toleraly AcqRel for everything, but still.

I really don't like these discussions to happen in text. It's a waste of time, leads to misunderstandings and mystifies nuance and given that it doesn't affect performance to use the more synchronizing option, I would rather pick whatever works now and move on. If you wish, let's have a call to have that discussion so that we don't spend 30 minutes writing text that will be buries in less than a week. Let's have a call whenever you wish.

alfiedotwtf:
> The current implementation doesn't seem to take into account that we have to continue staking from where we left of, so last_timestamp_seconds_used should probably be updated in the main thread and returned to the request sender (rpc, for example). But again, one problem at a time.

Pavel asked me to look into this yesterday, and I've been thinking about how to solve this without cluttering the API. What I'm thinking is to cache this time along with the staking key within the blockprod subsystem. On future requests, we carry on from where we left off, but If we receive a new tip or a timer on the cache item goes off, we remove it from the cache.

Thoughts?

utils/src/atomics/simple_atomic.rs
iljakuklic, line 50: Do we want this trait to be open to custom impls? May consider using the sealed trait pattern to prevent that.

utils/src/atomics/simple_atomic.rs
```diff
@@ -0,0 +1,235 @@
...
+impl<T, Ord> From<T> for Atomic<T, Ord>
+where
+    T: HasStdAtomic + Default,
+    <T as HasStdAtomic>::AtomicType: AtomicTrait<Type = T>,
+    Ord: Orderings,
+{
+    fn from(value: T) -> Self {
+        Self::new(value)
+    }
+}
```
iljakuklic, line 169: The `T: Default` constraint not needed here I think.

utils/src/atomics/simple_atomic.rs
```diff
@@ -0,0 +1,235 @@
...
+
+impl<T, Ord> std::fmt::Debug for Atomic<T, Ord>
+where
+    T: HasStdAtomic + Default + std::fmt::Debug,
```
iljakuklic, line 173: And here.

iljakuklic (approved):
Looks good as far as I'm concerned. Some minor discussion points below.

muursh (approved):

blockprod/src/detail/mod.rs
TheQuantumPhysicist: We came to the conclusion that we'll return the timestamp with the error and not use atomic. So the value will be moved back and forth instead of using an atomic.

PR #952, opened 2023-06-12
Load RPC cookie file on every request

pavel-kokolemin:
Currently we load the cookie file once. As a result, the wallet stops working properly when the node is restarted. I think we can load the cookie file every time an HTTP request is made instead.


p2p/src/sync/peer.rs
```diff
@@ -432,11 +432,20 @@ where
         let block = self.chainstate_handle.call(|c| c.preliminary_block_check(block)).await??;
         match self
             .chainstate_handle
-            .call_mut(|c| c.process_block(block, BlockSource::Peer))
+            .call_mut(|c| {
+                // If the block has already been processed, return the existing block index.
```
TheQuantumPhysicist, line 436: I would say "If the block already exists in the block tree"

p2p/src/sync/peer.rs
```diff
@@ -432,11 +432,20 @@ where
...
+                // It's used to prevent chainstate from printing "Block already exists" errors.
+                if let Some(block_index) = c.get_block_index(&block.get_id())? {
+                    log::debug!("Block ({}) already exists", block.get_id());
```
TheQuantumPhysicist, line 439: It's a good idea to add the context of this message, like "peer X sent a block that already exists". This helps in fine-tuning the process later.

p2p/src/sync/peer.rs
```diff
@@ -432,11 +432,20 @@ where
...
+                    return Ok(Some(block_index));
+                }
+                c.process_block(block, BlockSource::Peer)
+            })
             .await?
         {
             Ok(_) => Ok(()),
-            // It is OK to receive an already processed block.
+            // It is OK to receive an already processed block
+            // This should not happen now because of the `get_block_index` check above.
```
TheQuantumPhysicist, line 448: No need for "now". It sounds like it refers to this PR.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -110,7 +110,7 @@ pub async fn sync_once<T: NodeInterface>(
 
         log::info!(
             "Node chainstate updated, block height: {}, top block id: {}",
```
TheQuantumPhysicist, line 112: tip, not top.

rpc/src/lib.rs
```diff
@@ -198,20 +207,72 @@ impl subsystem::Subsystem for Rpc {
     }
 }
 
-/// Construct a basic authorization header for HTTP requests (when [username_password] is set)
-pub fn make_http_header_with_auth(username_password: Option<(&str, &str)>) -> http::HeaderMap {
+#[derive(Clone)]
+pub enum RpcAuthType {
```
TheQuantumPhysicist, line 211: I would call this RpcAuthData. I would use "Type" if this would be a numeric enumeration, not containing auth data.

TheQuantumPhysicist (approved):
Besides the minor comments, looking good! Please fix them and feel free to merge.

muursh (approved):

PR #959, opened 2023-06-13
fix rng range in wallet test

OBorce:


wallet/src/wallet/tests.rs
```diff
@@ -527,7 +527,7 @@ fn locked_wallet_cant_sign_transaction(#[case] seed: Seed) {
     assert_eq!(coin_balance, Amount::ZERO);
 
     // Generate a new block which sends reward to the wallet
-    let block1_amount = Amount::from_atoms(rng.gen_range(NETWORK_FEE + 1..NETWORK_FEE + 10000));
+    let block1_amount = Amount::from_atoms(rng.gen_range(NETWORK_FEE + 2..NETWORK_FEE + 10000));
```
iljakuklic, line 530: Not sure why it's `+2`. Could you add a comment explaining it?

azarovh, line 530: I think `NETWORK_FEE` is redundant here and just complicates the test. I'd remove it and just generate arbitrary big value like `rng.gen_range(10_000..200_000)`. We do this kind of generations all over the place and it shows more clearly that you just want some big value, whether 2 indicates some specific intention.

azarovh (approved):

discosultan (approved):

muursh (approved):

PR #960, opened 2023-06-13
Resend own address randomly (p2p)

pavel-kokolemin:


p2p/src/peer_manager/mod.rs
```diff
@@ -77,6 +80,10 @@ const PEER_MGR_HEARTBEAT_INTERVAL_MIN: Duration = Duration::from_secs(5);
 /// Upper bound for how often [`PeerManager::heartbeat()`] is called
 const PEER_MGR_HEARTBEAT_INTERVAL_MAX: Duration = Duration::from_secs(30);
 
+/// How often resend own address to one random outbound peer on average
+const RESEND_OWN_ADDRESS_DELAY: Duration =
```
TheQuantumPhysicist, line 84: Can you please explain the logic behind this? What's the motivation?

TheQuantumPhysicist (approved):
Besides the comment, all good.

Generally speaking, please always include justifications why we're doing a specific thing, especially in p2p... information seems to be scattered all over the place and many things are being done on-demand. Other devs would struggle to understand why certain things are being done.

iljakuklic (approved):
+1

PR #963, opened 2023-06-14
Do not spawn but return a future from p2p networking service

discosultan:
First of a series of PRs to resolve https://github.com/mintlayer/mintlayer-core/issues/888.

Refactors the `NetworkingService` trait such that the `start` method no longer spawns a task but instead returns a future. The name of the method was renamed to `initialize` to avoid confusion. `start` might be implied to spawn a task. This allows for a more granular control of async coroutine orchestration at the consumer level (p2p subsystem, for example).

I'm even thinking if it makes sense to match the interface of `NetworkingService` to that of `BlockSyncManager` and `PeerManager` which have `new` and `run` methods. This means `DefaultNetworkingService` would become the owner of `Backend`. However, I need to first ensure this wouldn't introduce any circular dependency.

Note that similar idea was also proposed in https://github.com/mintlayer/mintlayer-core/pull/943.

### DIFF
--- a/dns_server/src/crawler_p2p/crawler_manager/tests/mock_manager.rs
+++ b/dns_server/src/crawler_p2p/crawler_manager/tests/mock_manager.rs
@@ -25,10 +25,8 @@ use std::{
 };
 
 use async_trait::async_trait;
-use tokio::{
-    sync::{mpsc, oneshot},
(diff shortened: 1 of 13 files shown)


discosultan:
Will most likely solve the original issue in a different manner.

PR #964, opened 2023-06-14
Remove dependency to void

discosultan:
Ideally, we would use the [never](https://doc.rust-lang.org/std/primitive.never.html) type from standard library to represent a function that never returns. However, that is still a nightly-only experimental API.

Meanwhile, we can either use `std::convert::Infallible` or `futures::never::Never`. `std::convert::Infallible` is typically used for error types. `futures::never::Never` is currently a type alias for `std::convert::Infallible` but will be changed to the standard library never type once that becomes stable. Since most of our crates already either explicitly or implicitly include the `futures` crate, using `futures::never::Never` wouldn't add any additional dependency. This allows us to get rid of the dependency to `void` crate altogether.

Note that this idea was also proposed in https://github.com/mintlayer/mintlayer-core/pull/943.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -506,7 +506,6 @@ dependencies = [
  "thiserror",
  "tokio",
  "utils",
- "void",
 ]
 
 [[package]]
@@ -1537,6 +1536,7 @@ dependencies = [
  "common",
  "crypto",
  "directories",
+ "futures",
  "logging",
  "p2p",
  "p2p-test-utils",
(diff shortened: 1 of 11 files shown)


pavel-kokolemin (approved):

TheQuantumPhysicist (approved):

iljakuklic (approved):
Cool. There may also be occurrences of `Infallible` where `Never` would have been more appropriate due to the name if nothing else. On the other hand, "futures" is a bit heavier dependency than plain stdlib so there's also that tradeoff.

muursh (approved):

PR #965, opened 2023-06-15
Minor wallet fixes

pavel-kokolemin:


wallet/src/wallet/mod.rs
```diff
@@ -227,7 +227,11 @@ impl<B: storage::Backend> Wallet<B> {
             .get_mut(&account_index)
             .ok_or(WalletError::NoAccountFoundWithIndex(account_index))?;
         let value = f(account, &mut db_tx)?;
-        db_tx.commit()?;
+        // The in-memory wallet state has already changed, so rolling back
+        // the DB transaction will make the wallet state inconsistent.
+        // This should not happen with the sqlite backend in normal situations,
+        // so let's abort the process instead.
+        db_tx.commit().expect("RW transaction commit failed unexpectedly");
```
discosultan, line 234: Out of curiosity, why not indicate such situation with a specific error variant and have the consumer of the result decide to exit the process?

TheQuantumPhysicist: IIUC, there's no good answer to this question because this is about deciding *where* to make it crash. Generally speaking, file errors lead to exiting the node because it means it's something serious and unfixable. For the case of the wallet, like mentioned in the comment, the state will become inconsistent, so there's no point in continuing anyway... so being this aggressive is justified.

TheQuantumPhysicist (approved):

muursh (approved):

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -500,7 +500,7 @@ pub async fn handle_wallet_command(
                 .ok_or(WalletCliError::NoWallet)?
                 .get_utxos(utxo_type.to_wallet_types())
                 .map_err(WalletCliError::Controller)?;
-            Ok(ConsoleCommand::Print(format!("{utxos:?}")))
+            Ok(ConsoleCommand::Print(format!("{utxos:#?}")))
```
discosultan, line 503: This is a very welcome change. I just stumbled upon that yesterday where I wished the output was pretty printed ^^

discosultan (approved):

PR #967, opened 2023-06-15
Perform in memory reorg before validating a block in a branch

### LINKED ISSUE #752: Working reorgs with "verify stakes unspent" (to be able to detect/tolerate double-spends from other branches)
Using utxo set inside the `ChainstateRef::check_block_header` to validate consensus is incorrect.

It's prone to the attacks described in this topic:
https://medium.com/@dsl_uiuc/fake-stake-attacks-on-chain-based-proof-of-stake-cryptocurrencies-b8b05723f806

A possible solution would be to store spent stakes in a separate collection. Then for every branch recreate a temporary state from common ancestor block where we check that a stake output is either in utxo set or wasn't spent in that ancestor or current branch.

azarovh:
To validate a block with PoS in a branch we need a utxo set and accounting data to be up to that particular point in history and not the mainchain tip. So suggested solution is to perform in memory reorg and use the resulting data to validate a block. It's slow but relatively simple.

Regarding the speed if measuring the time of the `syncing_test::get_headers_branching_chains` test in debug with reorg depths of 1000 blocks the time is ~200s vs ~10s for prev version.  

Benchmark shows the time of 16.706s for 1000 blocks

Fixes #752

### DIFF
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -39,12 +39,12 @@ use common::{
     Uint256,
 };
 use logging::log;
-use pos_accounting::{PoSAccountingDB, PoSAccountingView};
+use pos_accounting::{PoSAccountingDB, PoSAccountingDelta, PoSAccountingView};
 use tx_verifier::transaction_verifier::{
     config::TransactionVerifierConfig, TransactionVerifier, TransactionVerifierDelta,
 };
 use utils::{ensure, tap_error_log::LogError};
-use utxo::{UtxosDB, UtxosView};
+use utxo::{UtxosCache, UtxosDB, UtxosView};
 
 use crate::{BlockError, ChainstateConfig};
 
@@ -474,17 +474,26 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         self.enforce_checkpoints(header).log_err()?;
         self.check_block_height_vs_max_reorg_depth(header)?;
 
+        let best_block_id = self.get_best_block_id()?;
+        let TransactionVerifierDelta {
+            utxo_cache: consumed_utxos,
+            accounting_delta: consumed_deltas,
+            ..
+        } = self.reorganize_in_memory(header, best_block_id)?;
+
         // TODO(Gosha):
         // using utxo set like this is incorrect, because it represents the state of the mainchain, so it won't
         // work when checking blocks from branches. See mintlayer/mintlayer-core/issues/752 for details
         let utxos_db = UtxosDB::new(&self.db_tx);
+        let utxos_cache = UtxosCache::from_data(&utxos_db, consumed_utxos).unwrap();
         let pos_db = PoSAccountingDB::<_, TipStorageTag>::new(&self.db_tx);
+        let pos_delta = PoSAccountingDelta::from_data(&pos_db, consumed_deltas);
         consensus::validate_consensus(
             self.chain_config,
             header,
             self,
-            &utxos_db,
-            &pos_db,
+            &utxos_cache,
+            &pos_delta,
             strict_pos_consensus_check,
         )
         .map_err(CheckBlockError::ConsensusVerificationFailed)
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -19,9 +19,8 @@ use chainstate_storage::{
     BlockchainStorageRead, BlockchainStorageWrite, TipStorageTag, TransactionRw,
 };
 use chainstate_types::{
-    block_index_ancestor_getter, get_skip_height, BlockIndex, BlockIndexHandle,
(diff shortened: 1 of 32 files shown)


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -762,6 +778,131 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
             .collect();
         Ok(result)
     }
+
+    // Disconnect all blocks from the mainchain up until common ancestor of provided block
+    // and then connect all block in the new branch.
+    // All these operations are performed via `TransactionVerifier` without db modifications
+    // and the resulting delta is returned.
+    pub fn reorganize_in_memory(
```
TheQuantumPhysicist, line 786: Please move this to a submodule. In the future we'll try to abstract this in a better way.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -428,36 +427,54 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
 
-        // TODO(Gosha):
-        // using utxo set like this is incorrect, because it represents the state of the mainchain, so it won't
-        // work when checking blocks from branches. See mintlayer/mintlayer-core/issues/752 for details
         let utxos_db = UtxosDB::new(&self.db_tx);
         let pos_db = PoSAccountingDB::<_, TipStorageTag>::new(&self.db_tx);
+
+        let best_block_id = self.get_best_block_id()?;
+        let (utxos_cache, pos_delta, epoch_data_cache) = if best_block_id == *header.prev_block_id()
+        {
```
TheQuantumPhysicist, line 439: There's a question on whether this is necessary. Ideally speaking, the "reorg_in_memory" function would be a no-op is best_block == previous_block. Is it possible to achieve that?

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -428,36 +427,54 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
         self.enforce_checkpoints(header).log_err()?;
 
-        // TODO(Gosha):
-        // using utxo set like this is incorrect, because it represents the state of the mainchain, so it won't
-        // work when checking blocks from branches. See mintlayer/mintlayer-core/issues/752 for details
         let utxos_db = UtxosDB::new(&self.db_tx);
         let pos_db = PoSAccountingDB::<_, TipStorageTag>::new(&self.db_tx);
+
+        let best_block_id = self.get_best_block_id()?;
+        let (utxos_cache, pos_delta, epoch_data_cache) = if best_block_id == *header.prev_block_id()
```
TheQuantumPhysicist, line 438: Keep in mind that this reorging is only necessary for select types of consensus. Perhaps we should wrap this whole thing with a boolean derived from ConsensusData, where it's only true for PoS. We don't need to reorg for PoW.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -91,10 +91,10 @@ pub struct Subsidy(pub Amount);
 #[derive(Debug, Eq, PartialEq)]
 pub struct TransactionVerifierDelta {
     tx_index_cache: BTreeMap<OutPointSourceId, CachedInputsOperation>,
-    utxo_cache: ConsumedUtxoCache,
+    pub utxo_cache: ConsumedUtxoCache,
```
TheQuantumPhysicist, line 94: No pub, please.

chainstate/src/detail/chainstateref/mod.rs
azarovh: actually it's already working as you described I just thought that it was a good check to avoid unnecessary computation

TheQuantumPhysicist (approved):
I don't see big issues. Good work. The only problem I see is the duplication, and I'd like (in the future) to unify reorging both in memory and in a database transaction such that we can just choose to keep it or persist it, instead of doing it twice.

azarovh:
>  I'd like (in the future) to unify reorging both in memory and in a database transaction

I tried to do that at first but then it just got too complicated, so I took this approach which was easier and more isolated

TheQuantumPhysicist:
> >  I'd like (in the future) to unify reorging both in memory and in a database transaction
> 
> I tried to do that at first but then it just got too complicated, so I took this approach which was easier and more isolated

And hence "future" :)

Please create an issue for it

chainstate/src/detail/chainstateref/mod.rs
azarovh: done

chainstate/src/detail/chainstateref/mod.rs
azarovh: removed the check

chainstate/src/detail/chainstateref/mod.rs
azarovh: added check for pos

muursh (approved):
Looks fine overall to me

iljakuklic (approved):

PR #968, opened 2023-06-15
Add new account command to wallet CLI

OBorce:
Add create new account command to the wallet CLI

- new account index creation happens sequentially 
- by default when creating a wallet the default account is also created and selected
- all commands use the selected account
- a new command for selecting an account has been added


wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -422,6 +433,30 @@ pub async fn handle_wallet_command(
...
+        WalletCommand::SelectAccount { account_index } => {
+            controller_opt
+                .as_mut()
+                .ok_or(WalletCliError::NoWallet)?
+                .select_account(account_index)
+                .map_err(WalletCliError::Controller)?;
+            Ok(ConsoleCommand::SetStatus {
+                status: format!(" ({})", account_index),
+                print: format!("Success, the selected account index is: {}", account_index),
+            })
```
iljakuklic, line 457: Looks like the status line is updated with the account number when the `SelectAccount` command is issued. Is the account number displayed if there are multiple accounts, regardless whether the select account command was run?

OBorce: I think it is missing an if account_index == DEFAULT then set the status to "" 
I think that is what @TheQuantumPhysicist mentioned to only show account if it is not the default one, not sure?

iljakuklic: In general, it seems quite error prone to manually set the status string every time it's supposed to change. Is it possible to somehow generate the status string on demand when it's being displayed? It would need the access to the wallet state to check current account and the total number of active accounts.

OBorce: that was my initial plan, but the prompt and the wallet communicate through channels in separate threads.

TheQuantumPhysicist: As long as this variable is trapped in the CLI... I'm good with it. It shouldn't touch the controller though.

wallet/src/wallet/mod.rs
```diff
@@ -213,7 +229,16 @@ impl<B: storage::Backend> Wallet<B> {
 
         self.accounts.insert(account.account_index(), account);
 
-        Ok(())
+        Ok(next_account_index)
+    }
+
+    pub fn select_account(&mut self, account_index: U31) -> WalletResult<()> {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Yes, this shouldn't be here. It should be part of the controller as a parameter. Making the lowest level stateful can be messy.

Every function in the controller must have an account parameter. Then the UI will choose the account to use based on what's available. Let me know, guys, if you have any objections to that.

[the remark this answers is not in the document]
TheQuantumPhysicist: To be clear, like I said in another comment, the CLI can be stateful with the account information. But the wallet backend should be neutral and the choice of the account should be a UI detail, not a backend detail.

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: Moved it to the WalletPrompt, but not sure what to do with non interactive modes, do I create another enum with commands with account_index, or add optional account_index in the existing ones and use it or if empty the selected one?

wallet/wallet-controller/src/lib.rs
```diff
@@ -267,6 +253,14 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
         self.sync_once().await
     }
 
+    pub fn create_account(&mut self) -> Result<U31, ControllerError<T>> {
+        self.wallet.create_account().map_err(ControllerError::WalletError)
+    }
```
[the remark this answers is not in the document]
OBorce: fixed the TODOs, now when creating a new account it is put in the unsynced_account member, and on each sync, it is first brought to the same height as the other accounts then proceeds as before to sync all of the accounts with the new blocks.

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: discussed with @pavel-kokolemin, I will make the non interactive modes work the same as the interactive one with a state of the selected account index.

wallet/src/wallet/mod.rs
```diff
@@ -63,8 +64,18 @@ pub enum WalletError {
...
+    #[error("Account with index {0} is not synced yet")]
+    AccountNotSyncedWithIndex(U31),
     #[error("Account with index {0} already exists")]
     AccountAlreadyExists(U31),
+    #[error("Cannot create a new account when last account is still empty")]
+    EmptyLastAccount,
+    #[error("Cannot create a new account when last account is still not in sync")]
+    LastAccountNotInSync,
+    #[error("The maximum number of accounts has been exceeded")]
+    MaxNumAccountsExceeded,
```
TheQuantumPhysicist, line 76: This is the maximum number of accounts due to it being u31, so it's `1 << 32 - 1`, right? If so, please call it "AbsoluteMaxNumAccountsExceeded", and change the message to "The absolute maximum number of accounts has been exceeded {}", and in it put that value.

wallet/src/wallet/mod.rs
```diff
@@ -187,19 +223,32 @@ impl<B: storage::Backend> Wallet<B> {
...
+
+        let next_account_index = self.accounts.last_key_value().map_or(
+            Ok(U31::ZERO),
+            |(last_account_index, last_account)| {
+                if last_account.has_transactions() {
+                    last_account_index.plus_one().map_err(|_| WalletError::MaxNumAccountsExceeded)
+                } else {
+                    // Cannot create a new account if the latest created one has no transactions
+                    // associated with it
+                    Err(WalletError::EmptyLastAccount)
```
TheQuantumPhysicist, line 243: Why?

wallet/src/wallet/mod.rs
```diff
@@ -97,6 +108,7 @@ pub struct Wallet<B: storage::Backend> {
     db: Store<B>,
     key_chain: MasterKeyChain,
     accounts: BTreeMap<U31, Account>,
+    unsynced_account: Option<(U31, Account)>,
```
TheQuantumPhysicist, line 111: Please add documentation explaining the logic behind this

TheQuantumPhysicist: Looking more into the code, why is this an `Option<(U31, Account)>`, and not a map so that multiple accounts can be synced and created? Is there a reason beyond that this is easier to do, in a way?

wallet/src/wallet/tests.rs
```diff
@@ -508,11 +487,32 @@ fn locked_wallet_accounts_creation_fail(#[case] seed: Seed) {
 
     let db = create_wallet_in_memory().unwrap();
     let mut wallet = Wallet::new_wallet(Arc::clone(&chain_config), db, MNEMONIC, None).unwrap();
+
+    // Need at least one address used from the previous account in otder to create a new account
```
TheQuantumPhysicist, line 491: otder

wallet/src/wallet/mod.rs
```diff
@@ -163,11 +180,30 @@ impl<B: storage::Backend> Wallet<B> {
...
+            None
+        };
+
+        // sanity check all other accounts have the same best block
+        ensure!(
+            accounts
+                .values()
+                .tuple_windows()
+                .all(|(acc1, acc2)| acc1.best_block() == acc2.best_block()),
+            WalletError::WalletNotInitialized
```
TheQuantumPhysicist, line 198: Can't we instead fix this problem? What's the benefit of being so harsh and fail loading the wallet?

TheQuantumPhysicist:
I think for this problem, having multiple accounts, we should first determine how to fix having multiple unsynced account in general (which is not that difficult, assuming I'm not missing something), and then we can generalize adding accounts. I don't particularly like this special treatment of "one account", and assuming everything else is synced. Too many assumptions and restrictions with little benefit. If there's a problem I'm not seeing, let's talk about it.

wallet/src/wallet/mod.rs
OBorce: It is discouraged by bip44, if you create multiple consecutive empty accounts, the default algorithm for recovering a wallet's accounts will stop on the first empty account and not discover the rest.

wallet/src/wallet/mod.rs
OBorce: it just fallows the logic that only 1 account can be empty, so you can't create multiple unsynced accounts.

wallet/src/wallet/mod.rs
TheQuantumPhysicist: I see

wallet/src/wallet/mod.rs
```diff
@@ -163,11 +180,30 @@ impl<B: storage::Backend> Wallet<B> {
...
         db_tx.close();
 
+        let first = accounts.values().next().ok_or(WalletError::WalletNotInitialized)?;
+        let last = accounts.values().last().ok_or(WalletError::WalletNotInitialized)?;
+
+        let unsynced_account = if first.best_block() != last.best_block() {
+            accounts.pop_last()
+        } else {
+            None
+        };
```
iljakuklic, line 190: Minor: I think this could be:
```rust
let unsynced_account = (first.best_block() != last.best_block()).then(|| accounts.pop_last());
```

wallet/src/wallet/mod.rs
```diff
@@ -187,19 +223,34 @@ impl<B: storage::Backend> Wallet<B> {
...
-            WalletError::AccountAlreadyExists(account_index)
-        );
+    pub fn number_of_accounts(&self) -> usize {
+        self.accounts.len() + self.unsynced_account.is_some() as usize
+    }
+
+    pub fn create_account(&mut self) -> WalletResult<U31> {
+        if self.unsynced_account.is_some() {
+            return Err(WalletError::LastAccountNotInSync);
+        }
```
iljakuklic, line 233: `ensure!`?

wallet/src/wallet/mod.rs
```diff
@@ -209,11 +260,16 @@ impl<B: storage::Backend> Wallet<B> {
...
-
-        self.accounts.insert(account.account_index(), account);
+        // if it is the only one insert it directly to accounts
+        if self.accounts.is_empty() {
+            self.accounts.insert(account.account_index(), account);
+        }
+        // else it is unsynced as the other accounts have already processed blocks
+        else {
+            self.unsynced_account = Some((account.account_index(), account));
+        }
```
iljakuklic, line 270: These comments break the usual formatting of the `if..else` construct.How about putting them inside the curly braces?

wallet/src/wallet/mod.rs
```diff
@@ -366,6 +427,41 @@ impl<B: storage::Backend> Wallet<B> {
...
+
+        if account.best_block().1
+            == self
+                .accounts
+                .values()
+                .next()
+                .ok_or(WalletError::WalletNotInitialized)?
+                .best_block()
+                .1
+        {
```
iljakuklic, line 456: Not  a big fan of massive multi-line if conditions. Could the comparison RHS be taken out into a separate variable defined before the `if`?

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -82,50 +102,91 @@ pub async fn sync_once<T: NodeInterface>(
...
+    wallet_block_id: Id<GenBlock>,
+    wallet_block_height: BlockHeight,
+    chain_config: &ChainConfig,
+    rpc_client: &T,
+    wallet_sync: &mut impl FnMut(BlockHeight, Block) -> Result<(), ControllerError<T>>,
+) -> Result<(), ControllerError<T>> {
+    utils::ensure!(
+        chain_info.best_block_height >= wallet_block_height,
+        ControllerError::NotEnoughBlockHeight(wallet_block_height, chain_info.best_block_height,)
+    );
```
iljakuklic, line 172: I am not convinced by this check. What if there is a better chain than the current one with fewer block? The best chain rule is not necessarily based on just the number of blocks in the chain. Am I missing something?

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -231,382 +246,483 @@ fn print_coin_amount(chain_config: &ChainConfig, value: Amount) -> String {
...
-                    "New wallet created successfully\nYour mnemonic: {}\nPlease write it somewhere safe to be able to restore your wallet."
-                , mnemonic)
-            } else {
-                "New wallet created successfully".to_owned()
-            };
-            Ok(ConsoleCommand::Print(msg))
-        }
+pub struct CommandHandler {
+    state: Option<(U31, usize)>,
+}
```
iljakuklic, line 251: Would appreciate some comment describing what the tuple components are and what it means for the `Option` to be `None`.

iljakuklic (commented):
A couple of comments below. Mostly minor coding suggestions.

wallet/wallet-controller/src/sync/mod.rs
pavel-kokolemin: This check was added here to detect situations where you connect to a node that is still downloading blocks (this can happen after removing the chainstate DB). Without this, the wallet will initiate a full rescan (because the top block is not yet known), which is unnecessary in most cases. I understand that it's possible to have a better chain with a lower height, but I don't think it's a big deal. Once the same block height is reached (which should not take much time), the wallet would resume synchronization.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -82,50 +102,91 @@ pub async fn sync_once<T: NodeInterface>(
         let chain_info =
             rpc_client.chainstate_info().await.map_err(ControllerError::NodeCallError)?;
 
-        let (wallet_block_id, wallet_block_height) =
-            wallet.best_block().map_err(ControllerError::WalletError)?;
+        // first sync the unsynced account until it matches the other accounts
+        if let Some((wallet_block_id, wallet_block_height)) = wallet.best_block_unsynced_acc() {
```
pavel-kokolemin, line 106: I think there is a bug. What would happen if there is a reorg while scanning the new account? This would always be `false`:
```
        if account.best_block().1
            == self
                .accounts
                .values()
                .next()
                .ok_or(WalletError::WalletNotInitialized)?
                .best_block()
                .1
        {
            self.unsynced_account
                .take()
                .map(|(index, acc)| self.accounts.insert(index, acc));
        }
```
Personally, I would prefer to keep things simple and sync each account independently.

OBorce: I think it should be fine, the unsynced account starts with genesis block, and the other accounts have already some blocks up to some block height X, this syncs the unsynced account up to that height X not comparing the block ID. If a reorg has happened the unsynced will now be on the new chain but on the same height, and after that, the rest of the accounts will also sync with the node.

wallet/wallet-controller/src/sync/mod.rs
pavel-kokolemin: No, because `is_initial_block_download` is set to false as soon as a block received that is less than 24 hours old.

wallet/src/wallet/mod.rs
TheQuantumPhysicist: You seem to have forgotten to write the maximum value in the error message.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -231,382 +246,501 @@ fn print_coin_amount(chain_config: &ChainConfig, value: Amount) -> String {
...
 
-            Ok(ConsoleCommand::Print("Success".to_owned()))
+    fn repl_status(&mut self) -> String {
+        match self.state {
+            Some(CLIWalletState {
+                selected_account,
+                total_accounts,
+            }) if total_accounts > 1 => {
+                format!("({}/{})", selected_account, total_accounts)
+            }
```
iljakuklic, line 304: I suspect `selected_account` start indexing at 0. It may be a confusing UI when using the `(X/N)` notation, since `X`  does not really reach `N`, i.e. users may expect the last account to be called `N` rather than `N-1`. Not that I have a good idea how to resolve it though.

wallet/src/wallet/mod.rs
iljakuklic: Should there also be a check that the transaction in the previous account has some sufficient number of confirmations? I can imagine a user who has a fresh transaction in account 0 creating account 1 immediately. The transaction in account 0 then gets reorged out. When re-syncing the wallet, whatever transactions may have been submitted to account 1 will not be discovered. May be left to a future discussion.

iljakuklic (approved):

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: yes exactly, it shows 0/2 and 1/2, but not sure what is the best solution to this.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -231,382 +246,501 @@ fn print_coin_amount(chain_config: &ChainConfig, value: Amount) -> String {
...
-                    controller.encrypt_wallet(&None).map_err(WalletCliError::Controller)?;
+            WalletCommand::CloseWallet => {
+                utils::ensure!(controller_opt.is_some(), WalletCliError::NoWallet);
+
+                *controller_opt = None;
+
+                self.state = None;
+                Ok(ConsoleCommand::SetStatus {
+                    status: self.repl_status(),
+                    print_message: "Success".to_owned(),
```
alfiedotwtf, line 398: Should this and the following "Success" messages output more context like the ones above e.g. "Wallet was successfully closed"? Otherwise, multiple lines of "Success" can get a bit confusing.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -231,382 +246,501 @@ fn print_coin_amount(chain_config: &ChainConfig, value: Amount) -> String {
...
+                let height =
+                    rpc_client.get_best_block_height().await.map_err(WalletCliError::RpcError)?;
+                Ok(ConsoleCommand::Print(height.to_string()))
+            }
 
-        WalletCommand::ChainstateInfo => {
-            let info = rpc_client.chainstate_info().await.map_err(WalletCliError::RpcError)?;
-            Ok(ConsoleCommand::Print(format!("{info:#?}")))
-        }
+            WalletCommand::BlockHash { height } => {
```
alfiedotwtf, line 472: I know this was the same as before your change, but should this command be changed to "BlockId" rather than "BlockHash"?

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -231,382 +246,501 @@ fn print_coin_amount(chain_config: &ChainConfig, value: Amount) -> String {
...
-                .generate_block(transactions_opt)
-                .await
-                .map_err(WalletCliError::Controller)?;
-            rpc_client.submit_block(block).await.map_err(WalletCliError::RpcError)?;
-            Ok(ConsoleCommand::Print("Success".to_owned()))
-        }
+            WalletCommand::SelectAccount { account_index } => {
+                self.set_selected_account(account_index).map(|_| ConsoleCommand::SetStatus {
+                    status: self.repl_status(),
+                    print_message: "Sucess".into(),
```
alfiedotwtf, line 543: Typo: Success.

wallet/wallet-cli-lib/src/lib.rs
```diff
@@ -100,30 +102,30 @@ pub async fn run(
...
-            commands::WalletCommand::OpenWallet { wallet_path },
-        )
-        .await?;
+        let (res_tx, _res_rx) = tokio::sync::oneshot::channel();
+        event_tx
+            .send(Event::HandleCommand {
+                command: WalletCommand::OpenWallet { wallet_path },
+                res_tx,
+            })
+            .expect("should not fail");
```
alfiedotwtf, line 116: Should there be more context here e.g. `.expect("Opening a wallet should not fail")`

wallet/wallet-cli-lib/src/lib.rs
```diff
@@ -100,30 +102,30 @@ pub async fn run(
...
-            commands::WalletCommand::StartStaking,
-        )
-        .await?;
+        let (res_tx, _res_rx) = tokio::sync::oneshot::channel();
+        event_tx
+            .send(Event::HandleCommand {
+                command: WalletCommand::StartStaking,
+                res_tx,
+            })
+            .expect("should not fail");
```
alfiedotwtf, line 126: Same as above.

wallet/wallet-controller/src/lib.rs
```diff
@@ -296,7 +311,7 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
             }
 
             if self.staking_started {
-                let generate_res = self.generate_block(None).await;
+                let generate_res = self.generate_block(DEFAULT_ACCOUNT_INDEX, None).await;
```
alfiedotwtf, line 314: Is staking on the `DEFAULT` account always the case? Should this be a parameter instead?

alfiedotwtf (approved):

wallet/src/wallet/mod.rs
TheQuantumPhysicist: Checking whether there are sufficient confirmations should also be centrally controlled somehow. The wallet should have a setting to view only confirmed transactions after some confirmations. But I view this as a nicity more than a necessity.

wallet/src/wallet/mod.rs
```diff
@@ -187,19 +214,37 @@ impl<B: storage::Backend> Wallet<B> {
...
+            WalletError::LastAccountNotInSync
         );
 
+        let next_account_index = self.accounts.last_key_value().map_or(
+            Ok(U31::ZERO),
+            |(last_account_index, last_account)| {
+                if last_account.has_transactions() {
+                    last_account_index.plus_one().map_err(|_| {
+                        WalletError::AbsoluteMaxNumAccountsExceeded(
+                            last_account_index.into_u32() + 1,
```
TheQuantumPhysicist, line 233: I think this is wrong. You want to show the maximum, not the next-to-maximum, right? This may or may not overflow based on whether the derivation is hard. Just use the last available value, without the +1, which is the absolute maximum.

wallet/wallet-controller/src/sync/mod.rs
TheQuantumPhysicist: Well, can't we just use the chain-trust instead?

TheQuantumPhysicist: We can do that in another PR, btw, but it should be done, I believe. There's no reason to depend on height when height isn't really the criterion to determine whether we need a scan.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -231,382 +246,501 @@ fn print_coin_amount(chain_config: &ChainConfig, value: Amount) -> String {
...
+    }
 
-            Ok(ConsoleCommand::Print("Success".to_owned()))
+    fn repl_status(&mut self) -> String {
+        match self.state {
+            Some(CLIWalletState {
+                selected_account,
+                total_accounts,
+            }) if total_accounts > 1 => {
+                format!("(Account No. {})", selected_account)
```
TheQuantumPhysicist, line 303: Didn't we initially agree to make it possible to give an optional name to accounts?

wallet/wallet-cli-lib/src/commands/mod.rs
TheQuantumPhysicist: Is there a reason why BlockHash instead of Id is used?

wallet/src/wallet/mod.rs
```diff
@@ -97,6 +108,7 @@ pub struct Wallet<B: storage::Backend> {
     db: Store<B>,
     key_chain: MasterKeyChain,
     accounts: BTreeMap<U31, Account>,
+    unsynced_accounts: BTreeMap<U31, Account>,
```
TheQuantumPhysicist, line 111: Apologies for the confusion. But this was an Option and the reasoning was that it can only be one "unsynced account". Has that changed now?

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -231,382 +246,501 @@ fn print_coin_amount(chain_config: &ChainConfig, value: Amount) -> String {
...
-                Some(block) => Ok(ConsoleCommand::Print(block.hex_encode())),
-                None => Ok(ConsoleCommand::Print("Not found".to_owned())),
+            WalletCommand::CreateNewAccount => {
+                let new_account_index = controller_opt
+                    .as_mut()
+                    .ok_or(WalletCliError::NoWallet)?
+                    .create_account()
+                    .map_err(WalletCliError::Controller)?;
+
+                self.set_total_accounts(new_account_index.into_u32() as usize);
```
pavel-kokolemin, line 530: I think total should be `new_account_index + 1`.

wallet/wallet-cli-lib/src/lib.rs
pavel-kokolemin, line 116: `_res_rx` should be used here, to make sure that the `wallet_file` command line argument handling errors are correctly detected.

wallet/wallet-cli-lib/src/lib.rs
```diff
@@ -100,30 +102,30 @@ pub async fn run(
...
 
     if start_staking {
-        commands::handle_wallet_command(
-            &chain_config,
-            &rpc_client,
-            &mut controller_opt,
-            commands::WalletCommand::StartStaking,
-        )
-        .await?;
+        let (res_tx, _res_rx) = tokio::sync::oneshot::channel();
```
pavel-kokolemin, line 120: Same here (`_res_rx` should be used).

pavel-kokolemin (changes requested):

wallet/wallet-cli-lib/src/commands/mod.rs
pavel-kokolemin: No particular reason, I just used the name from Bitcoin Core. @OBorce please update it.

wallet/wallet-cli-lib/src/commands/mod.rs
pavel-kokolemin: I think it's OK because you can see the command that produced the output. It should be clear enough.

wallet/src/wallet/mod.rs
OBorce: It was to fix the case when the wallet process is killed in the middle of syncing the accounts with new blocks, if some accounts are synced and the rest not, loading up the wallet can have multiple unsynced accounts.

wallet/wallet-cli-lib/src/lib.rs
OBorce: The except is actually for sending the event through the channel.

wallet/src/wallet/mod.rs
OBorce: removed the +1

wallet/wallet-cli-lib/src/lib.rs
OBorce: I put them in a vector and passed them to the `repl_handle` to process/print/error the responses

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: Added optional names, errors out if name is empty string. Can add account renaming in another PR?

wallet/wallet-cli-lib/src/lib.rs
pavel-kokolemin: Sorry, but I don't think that's enough, the app should exit early with an error code (like it worked before).

OBorce: Fixed the early exit

pavel-kokolemin (approved):

wallet/src/account/mod.rs
```diff
@@ -664,6 +666,14 @@ impl Account {
             self.account_info.best_block_height(),
         )
     }
+
+    pub fn has_transactions(&self) -> bool {
+        !self.output_cache.txs().is_empty()
+    }
+
+    pub fn name(&self) -> Option<String> {
+        self.account_info.name().clone()
```
TheQuantumPhysicist, line 675: Why not return a reference, and make the caller choose to clone?

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -103,7 +105,21 @@ pub enum WalletCommand {
...
+    CreateNewAccount {
+        name: Option<String>,
+    },
+
+    /// Select a wallet account for usage
+    SelectAccount {
+        account_index: U31,
+    },
+
+    /// Start staing using the specified account
```
TheQuantumPhysicist, line 119: staking?

TheQuantumPhysicist (approved):
Besides the minor things I mentioned, I think this is now in good shape.

wallet/src/account/mod.rs
OBorce: fixed

wallet/wallet-cli-lib/src/config.rs
```diff
@@ -38,7 +39,7 @@ pub struct WalletCliArgs {
 
     /// Start staking after the start
     #[clap(long)]
-    pub start_staking: bool,
+    pub start_staking: Option<U31>,
```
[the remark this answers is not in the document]
OBorce: reverted back to bool, and use the selected account index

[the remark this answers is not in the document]
TheQuantumPhysicist: While this is OK, keep in mind that staking automatically can be problematic. But I guess, one problem at a time.

PR #969, opened 2023-06-16
Establish p2p task shutdown order and remove shared shutdown flag

discosultan:


discosultan:
Will be broken down into smaller pieces.

PR #970, opened 2023-06-16
Block status field was added to block index

ImplOfAnImpl:


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -762,6 +753,55 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+        self.get_gen_block_index(&block.prev_block_id())
+            .map_err(BlockError::PropertyQueryError)?
+            .ok_or(BlockError::PrevBlockNotFound)
+    }
+
+    pub fn new_block_index(
+        &self,
+        block: &WithId<Block>,
+        block_status: BlockStatus,
+    ) -> Result<BlockIndex, BlockError> {
```
ImplOfAnImpl, line 774: Note: this is basically the old "add_to_block_index" function, but without checking whether the block index already exists.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -918,6 +958,8 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
     ) -> Result<GenBlockIndex, BlockError> {
         let best_block_id = self
             .get_best_block_id()
+            // FIXME: here we actually "expect" on a db error, not on the absence of the best block
```
TheQuantumPhysicist, line 961: Disconnection should never fail. There's absolutely no way to recover from it.

Think about it: We're basically unable to switch to another chain due to I/O error. What are we supposed to do, exactly?!

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -945,75 +988,37 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
...
+            .get_gen_block_index(best_block_id)
             .map_err(BlockError::BestBlockLoadError)
             .log_err()?
             .expect("Inconsistent DB");
 
         if new_block_index.chain_trust() > current_best_block_index.chain_trust() {
-            self.reorganize(&best_block_id, &new_block_index).log_err()?;
-            return Ok(Some(new_block_index));
+            self.reorganize(best_block_id, new_block_index).log_err()?;
+            return Ok(true);
```
TheQuantumPhysicist, line 1006: Why did you change this to return a boolean instead of the new tip's block index?

chainstate/src/detail/error.rs
```diff
@@ -46,10 +46,12 @@ pub enum BlockError {
     CheckBlockFailed(#[from] CheckBlockError),
     #[error("Failed to update the internal blockchain state: {0}")]
     StateUpdateFailed(#[from] ConnectTransactionError),
+    // FIXME: remove this in favor of PropertyQueryError?
     #[error("Failed to load best block")]
     BestBlockLoadError(PropertyQueryError),
-    #[error("Failed to load block")]
-    BlockLoadError(PropertyQueryError),
+    #[error("Generic property query error")]
+    // FIXME: is #[from] a good idea, provided that we have several errors that can contain PropertyQueryError?
```
TheQuantumPhysicist, line 53: If I understand correctly, you mean that we have different scenarios that return the same error, making their treatment different, where #[from] will make that seamless and cause ignoring it?

In that case, I'm OK with removeing that #[from], but please add a comment above this error explaining why we're not using #[from] to avoid falling into that mistake again.

chainstate/src/detail/error.rs
```diff
@@ -164,6 +171,8 @@ pub enum BlockSizeError {
 pub enum InitializationError {
     #[error("Block storage error: `{0}`")]
     StorageError(#[from] chainstate_storage::Error),
+    // FIXME: rename to PropertyQueryError? (otherwise it sounds like just querying a property
```
TheQuantumPhysicist, line 174: I don't mind.

utils/src/status.rs
```diff
@@ -0,0 +1,455 @@
...
+///     statuses.set(MyStatusField::A, Status::Good);
+///     statuses.set(MyStatusField::B, Status::Bad);
+///     # assert_eq!(statuses.all_good(), Status::Bad);
+/// }
+/// # fn main() {
+/// #   do_stuff();
+/// # }
+/// ```
+#[derive(Debug, Clone, Copy, Encode, Decode)]
+pub struct Statuses<const NUM_BYTES: usize, IndexType>([u8; NUM_BYTES], PhantomData<IndexType>)
```
TheQuantumPhysicist, line 123: I'm confused... why is this not just an enum?

Also this seems to be belong in chainstate types... not in utils.

chainstate/test-suite/src/tests/reorgs_tests.rs
```diff
@@ -391,6 +391,7 @@ fn subscribe_to_events(tf: &mut TestFramework, events: &EventList) {
     tf.chainstate.subscribe_to_events(subscribe_func);
 }
 
+// FIXME: rename this because "block_status" now means a more specific thing.
```
TheQuantumPhysicist, line 394: Agreed. Maybe something like `check_block_reorg_state`?

chainstate/test-suite/src/tests/pos_processing_tests.rs
```diff
@@ -523,12 +523,29 @@ fn pos_block_signature(#[case] seed: Seed) {
     );
 
     // bad block signature
+    // Note: this tx is not strictly needed for the test, but without it the block id here
```
TheQuantumPhysicist, line 526: Ususally we solve this problem with a different timestamp.

chainstate/src/detail/chainstateref/mod.rs
TheQuantumPhysicist: It's moved. Got it. All good.

chainstate/src/detail/error.rs
ImplOfAnImpl: [discussed] Actually, it was vice versa - I added the PropertyQueryError enum variant here and was wondering whether having "#[from]" for it was a good idea. But having PropertyQueryError is itself not a good idea, so I reverted this change.

utils/src/status.rs
ImplOfAnImpl: [discussed] The idea was to be able to have an independent three-state status (Good, Bad, Unknown) for each of the checks performed on a block. But the independent statuses are not needed because checks are always performed in the same order. So, this generic implementation will be removed and a simple enum will be used instead.

P.S. we also discussed at some point that the status enum has to be a bitfield, but it seems that there is no need for it to be one, it can be just a simple sequential enum, e.g.:
```
pub enum BlockValidationStage {
    Initial,
    ParentOk,
    CheckBlockOk,
    FullyChecked,
}

pub struct BlockStatus {
    last_successfull_validation_stage: BlockValidationStage,
}
```
Btw, in bitcoin-core, BlockStatus is not a real bitfield either. Rather, it has a sequential part in its lower bits that represents the block check status (BLOCK_VALID_RESERVED = 1, BLOCK_VALID_TREE = 2, BLOCK_VALID_TRANSACTIONS = 3, BLOCK_VALID_CHAIN = 4, BLOCK_VALID_SCRIPTS = 5) and then it also has a bunch of extra flags in its higher bits. For now, we don't need those extra flags, but if/when we do, we'd rather define a separate bitfield enum for them.

chainstate/src/detail/error.rs
```diff
@@ -165,7 +172,7 @@ pub enum InitializationError {
     #[error("Block storage error: `{0}`")]
     StorageError(#[from] chainstate_storage::Error),
     #[error("{0}")]
-    PropertyQuery(#[from] PropertyQueryError),
+    PropertyQueryError(#[from] PropertyQueryError),
```
ImplOfAnImpl, line 175: This variant is actually never used. Should I delete it instead?

TheQuantumPhysicist: Yep

discosultan:
Very nice job with code comments. Makes it much easier to follow 👍

chainstate/src/detail/mod.rs
```diff
@@ -271,52 +274,179 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+        // Update block index status.
+        let status_update_result = self.with_rw_tx_for_block_id(
+            |chainstate_ref| chainstate_ref.set_block_status(&block_index),
+            &block_id,
+            "block index update",
+        );
+
+        // If both block validation and block index update failed, we want to return the first
+        // error, so we check it first.
+        let result = if check_result? {
```
azarovh, line 442: you could use `bool::then()` to convert to Option

muursh (approved):
This is fine for me

chainstate/types/src/block_status.rs
```diff
@@ -0,0 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use enum_iterator::Sequence;
+use serialization::{Decode, Encode};
+
+/// Block validation steps are always performed in the same order, which is represented by this enum.
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Sequence, Decode, Encode)]
+pub enum BlockValidationStage {
+    Initial,
```
muursh, line 22: I'm somewhat hesitant over this but I have nothing better to offer

chainstate/src/detail/mod.rs
```diff
@@ -271,52 +274,179 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
         }
+
+        let block_index = {
+            let mut block_index = block_index;
+            block_index.set_status(block_status);
+            block_index
+        };
+
+        // Update block index status.
+        let status_update_result = self.with_rw_tx_for_block_id(
```
azarovh, line 434: wouldn't it be simpler to set block status inside `perform_block_checks` to make it one transaction?

chainstate/test-suite/src/tests/pos_processing_tests.rs
azarovh: I think this should never happen because the time is advanced after the first block. Previous block was invalid so unless we saved invalid block index the hash cannot be equal.

@ImplOfAnImpl do you have the seed of the test failure? I can take a look

ImplOfAnImpl: > I think this should never happen because the time is advanced after the first block.

Not sure I understand you correctly. In the test, both blocks re-use the same block_timestamp.

> Previous block was invalid so unless we saved invalid block index the hash cannot be equal.

We now do save indices of invalid blocks, that's exactly the reason for the test failure.

> @ImplOfAnImpl do you have the seed of the test failure? I can take a look

It's reproducible with any seed in my branch.

I'd say that the problem here is that several tests are packed into one test function and because of that they began interfering with each other. Perhaps I should split the test into several smaller tests?

chainstate/src/detail/mod.rs
azarovh: aah right, missed that

chainstate/test-suite/src/tests/pos_processing_tests.rs
azarovh: ok I see that we save these blocks, nevermind then

azarovh (approved):

chainstate/types/src/block_status.rs
```diff
@@ -0,0 +1,60 @@
...
+    ParentOk,
+    CheckBlockOk,
+    FullyChecked,
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Decode, Encode)]
+pub struct BlockStatus {
+    last_successfull_validation_stage: BlockValidationStage,
```
TheQuantumPhysicist, line 30: successful

chainstate/types/src/block_index.rs
```diff
@@ -83,6 +86,14 @@ impl BlockIndex {
         &self.some_ancestor
     }
 
+    pub fn status(&self) -> BlockStatus {
+        self.status
+    }
+
+    pub fn set_status(&mut self, status: BlockStatus) {
```
TheQuantumPhysicist, line 93: Perhaps instead of freely setting the status here, you can implement the stepping here too, somehow, and do a read-modify-write operation such that the status cannot be freely changed to a completely different object.

chainstate/src/detail/mod.rs
```diff
@@ -271,52 +274,179 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+            comment,
+        );
+
+        (state, result)
+    }
 
-            return Ok(result);
+    // Perform all the regular block checks, updating `block_status` after each successful check.
+    // The returned bool indicates whether a reorg has occurred.
+    fn perform_block_checks(
```
TheQuantumPhysicist, line 350: I wouldn't call this just "checks"... it's more. I don't know whether we should just call this "put_in_blocktree" or something similar to signal that this "mutates" the state, unlike just checking.

chainstate/src/detail/error.rs
```diff
@@ -60,8 +60,15 @@ pub enum BlockError {
...
-    DatabaseCommitError(Id<Block>, usize, chainstate_storage::Error),
+    #[error("Block {0} has already been processed")]
+    BlockAlreadyProcessed(Id<Block>),
+    #[error("Block {0} has already been processed and marked as invalid")]
+    InvalidBlockAlreadyProcessed(Id<Block>),
+    #[error("Block {0} has invalid parent block")]
+    InvalidParent(Id<Block>),
+    // Note: the last String here serves as a "comment" that helps distinguish separate commit attempts.
+    #[error("Failed to commit block state update to database for block: {0} after {1} attempts with error {2} ({3})")]
+    DatabaseCommitError(Id<Block>, usize, chainstate_storage::Error, String),
```
TheQuantumPhysicist, line 71: We don't prefer to have strings in errors because it makes comparing them difficult. It's always better to have a proper enum to represent all possible states, which can be compared with Eq.

TheQuantumPhysicist (approved):
Besides the minor comments I made, looking good.

chainstate/src/detail/mod.rs
```diff
@@ -271,52 +274,179 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+            // where there is a BlockIndex in the DB with a "fully checked" status, but the
+            // block itself is missing. So we bail out in this case.
+            return Err(err);
         }
+
+        let block_index = {
+            let mut block_index = block_index;
+            block_index.set_status(block_status);
+            block_index
+        };
```
iljakuklic, line 431: Should there be `BlockIndex::with_status` to do this?

chainstate/types/src/block_status.rs
iljakuklic: `NotValidated`/`UnValidated`/`NotChecked`? I don't know some seem too close to `Invalid` which may be confusing. Agreed this should be something more explicit about the fact that the block has not been validated, `Initial` does not quite cut it.

iljakuklic (approved):
Looks good. A couple of minor style comments below:

chainstate/src/detail/error.rs
ImplOfAnImpl: Oops, I lied, it is used, just not explicitly.

chainstate/types/src/block_status.rs
ImplOfAnImpl: I've changed it to Unchecked, because "checked" is already used in FullyChecked (and IMO Unchecked sounds better than NotChecked). Let me know if this is OK.

iljakuklic: > I've changed it to Unchecked, because "checked" is already used in FullyChecked (and IMO Unchecked sounds better than NotChecked). Let me know if this is OK.

Sounds good to me

chainstate/src/detail/mod.rs
```diff
@@ -271,52 +274,179 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+        comment: &str,
+    ) -> (State, Result<Res, BlockError>)
+    where
+        Func: FnMut(
+            &mut chainstateref::ChainstateRef<TxRw<'_, S>, V>,
+            &mut State,
+        ) -> Result<Res, BlockError>,
+        State: Clone,
+    {
+        let mut state = initial_state.clone();
```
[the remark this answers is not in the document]
ImplOfAnImpl: Oops, that's a bug, `state` is supposed to be reset to `initial_state` on each commit attempt (i.e. inside the closure that is passed to `with_rw_tx_for_block_id`). It was supposed to be like this:
```
        let mut state = initial_state.clone();

        let result = self.with_rw_tx_for_block_id(
            |chainstate_ref| {
                state = initial_state.clone();
                func(chainstate_ref, &mut state)
            },
            block_id,
            comment,
        );

        (state, result)
```
So, I can't move from initial_state right away. I wonder though, whether I should mess with optionality in order to get rid of this extra clone at the beginning (the actual "state" is cheap to clone, but the function is generic, so I'm not sure).

chainstate/src/detail/mod.rs
```diff
@@ -271,52 +274,179 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+        initial_state: &State,
+        block_id: &Id<Block>,
+        comment: &str,
+    ) -> (State, Result<Res, BlockError>)
+    where
+        Func: FnMut(
+            &mut chainstateref::ChainstateRef<TxRw<'_, S>, V>,
+            &mut State,
+        ) -> Result<Res, BlockError>,
+        State: Clone,
```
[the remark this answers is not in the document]
ImplOfAnImpl: > I am wondering if there is much value in a function like this. Since func is a FnMut, you can just mutably capture any state that needs to be modified from the outer scope (as the implementation here does). No extra argument is necessary.

Well, yeah, with the missing clone on each iteration is doesn't make much sense, but that was a bug, as I've already mentioned.

> If you want to make the state manipulation more functional in style, I'd avoid using &mut references to it. Just pass it in/out by value everywhere.

In general, the &mut parameter is needed by `perform_block_checks`, where I want to update the status without the need to return it explicitly. IMO without it, the logic of `perform_block_checks` would be more complicated.

chainstate/src/detail/mod.rs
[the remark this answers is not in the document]
iljakuklic: Okay, understandable to have a `Clone` constraint if multiple copies are needed.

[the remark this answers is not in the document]
iljakuklic: > So, I can't move from initial_state right away. I wonder though, whether I should mess with optionality in order to get rid of this extra clone at the beginning (the actual "state" is cheap to clone, but the function is generic, so I'm not sure).

How about making the inner closure return the final state? Rough sketch:
```rust
self.with_rw_tx_for_block_id(
    |chainstate_ref| {
        let state = initial_state.clone();
        let result = func(chainstate_ref, &mut state)?;
        Ok((result, state))
    },
    // other params
)
```

chainstate/src/detail/mod.rs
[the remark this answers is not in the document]
iljakuklic: > > If you want to make the state manipulation more functional in style, I'd avoid using &mut references to it. Just pass it in/out by value everywhere.
> 
> In general, the &mut parameter is needed by `perform_block_checks`, where I want to update the status without the need to return it explicitly. IMO without it, the logic of `perform_block_checks` would be more complicated.

Fair point. Though my original suggestion was supposed to only apply to the generic helper function(s). Ultimately it's a matter of taste to a degree so I don't mind either way. The whole discussion was prompted by the initial confusion around `Clone` anyway.

chainstate/src/detail/mod.rs
ImplOfAnImpl: I've renamed it to `integrate_block`, let me know, if this is ok.

TheQuantumPhysicist: OK for now. Thank you.

chainstate/types/src/block_index.rs
ImplOfAnImpl: IMO it will overcomplicate things. First of all, in `perform_block_checks` (AKA `integrate_block`) I'll have to have a `&mut BlockIndex` instead of just `&mut BlockStatus`, so more space for errors. Secondly, `perform_block_checks` needs to persist a `BlockIndex` before calling `activate_best_chain`; right now it persists one with the "Initial" status, but after the proposed change it will have to persist `BlockIndex` with some intermediate status, which doesn't feel nice.
So unless you insist, I'd leave it as is. (Or maybe I don't understand your idea fully)

chainstate/src/detail/mod.rs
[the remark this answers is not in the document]
ImplOfAnImpl: Well, the updated state must be returned on error too. I decided to use Option after all.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -674,6 +665,29 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+    }
+
+    // Return Ok(()) if the specified block has a valid parent and an error otherwise.
+    pub fn check_block_parent(&self, block: &WithId<Block>) -> Result<(), BlockError> {
+        let parent_block_index = self.get_previous_block_index_or_block_error(block)?;
+        if !parent_block_index.status().is_valid() {
+            return Err(BlockError::InvalidParent(block.get_id()));
+        }
+
+        Ok(())
```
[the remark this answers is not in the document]
ImplOfAnImpl: I must mention that `cargo fmt` turned this into 4 lines:
```
ensure!(
    parent_block_index.status().is_valid(),
    BlockError::InvalidParent(block.get_id())
);
```

chainstate/src/detail/mod.rs
```diff
@@ -557,5 +688,125 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
     }
 }
 
+// FIXME: should I move the tests to a separate file?
 #[cfg(test)]
-mod test;
+mod test {
```
iljakuklic, line 693: I would be in favour of this. Let's keep the `mod.rs` file just with implementation, it's already large enough.

Also, the file seems to be removed from the module tree in favour of an inline module but `test.rs` is not deleted as far as I can see.

iljakuklic (approved):

chainstate/src/detail/mod.rs
ImplOfAnImpl: Damn it, I was sure I was replacing an empty module `mod test {}` with my module.

I reverted this and moved my tests to test.rs

chainstate/types/src/block_index.rs
ImplOfAnImpl: [discussed] I'll skip this for now.

discosultan (approved):

PR #973, opened 2023-06-16
Use bounded channels for sync manager messages (p2p)

pavel-kokolemin:
After this PR, the peer's socket will be blocked if there are too many unprocessed sync messages. This should be enough to solve most problems with unbounded channels in p2p.
 I have removed all `nonexistent_peer` tests because they become irrelevant (the sync manager no longer dispatches received messages). The `block_announcement_3_peers` test was removed because it has been disabled for some time and fixing it would take too much time.


p2p/src/net/default_backend/backend.rs
```diff
@@ -543,95 +549,16 @@ where
         }
     }
 
-    fn handle_message(&mut self, peer: PeerId, message: Message) -> crate::Result<()> {
+    fn handle_message(&mut self, peer: PeerId, message: PeerManagerMessage) -> crate::Result<()> {
         // Do not process remaining messages if the peer has been forcibly disconnected (for example, after being banned).
         // Without this check, the backend might send messages to the sync and peer managers after sending the disconnect notification.
         if !self.peers.contains_key(&peer) {
             log::debug!("ignore received messaged from a disconnected peer");
```
TheQuantumPhysicist, line 556: Please change this to info

TheQuantumPhysicist: Also add the peer info here. This helps in debugging over time without debug mode.

p2p/src/net/default_backend/backend.rs
```diff
@@ -57,6 +57,10 @@ use crate::{
 
 use super::{peer::PeerRole, transport::TransportAddress, types::HandshakeNonce};
 
+/// Buffer size of the channel to the SyncManager peer task.
+/// How many unprocessed messages can be sent before the peer's event loop is blocked.
+const SYNC_CHAN_BUF_SIZE: usize = 20;
```
TheQuantumPhysicist, line 62: This sounds too small... are you sure this is an appropriate value? Not like 100-500?

TheQuantumPhysicist (approved):
Like we discussed. The block_announcement tests should either be replaced, or proven to be unnecessary. Otherwise, check my comments and feel free to merge afterwards.

pavel-kokolemin:
> Like we discussed. The block_announcement tests should either be replaced, or proven to be unnecessary. Otherwise, check my comments and feel free to merge afterwards.

I took a closer look at what the test is supposed to do. It was written specifically for the libp2p backend to check the validation of the lib2p2 gossip protocol (see commit `d437ab2d067954d6a30cdcf6bec90ca8ef15b73c`):
```
test libp2p gossipsub with multiple peers and verify that as our libp2p requires message validation, peers don't automatically forward the messages
```
It was never adapted for our "mock" backend (now the "default" backend) and has been constantly disabled after removing libp2p. And I think it's safe to remove it because our approach is very different (we send a new header notification when the tip changes, and only when chainstate validates new blocks).

TheQuantumPhysicist:
> > Like we discussed. The block_announcement tests should either be replaced, or proven to be unnecessary. Otherwise, check my comments and feel free to merge afterwards.
> 
> I took a closer look at what the test is supposed to do. It was written specifically for the libp2p backend to check the validation of the lib2p2 gossip protocol (see commit `d437ab2d067954d6a30cdcf6bec90ca8ef15b73c`):
> ```
> test libp2p gossipsub with multiple peers and verify that as our libp2p requires message validation, peers don't automatically forward the messages
> ```
> It was never adapted for our "mock" backend (now the "default" backend) and has been constantly disabled after removing libp2p. And I think it's safe to remove it because our approach is very different (we send a new header notification when the tip changes, and only when chainstate validates new blocks).

Cool. Go for it. Thanks.

p2p/src/net/default_backend/backend.rs
pavel-kokolemin: It's per connected peer, so 20 should be enough.
I guess it should mostly affect how much time is spent on initial blocks download, so I tried to test how much time it takes to synchronize a new node. The result is surprising:
```
Unbounded channels (current master): 4:36.10
Bounded channels (buffer is 20): 2:46.00
Bounded channels (buffer is 500): 4:18.96
```
I noticed that a lot of time is spent in mempool, which is probably pointless if the node is in initial block download state. I will create a ticket for this. We could test again once the issue with mempool is resolved (I added TODO).

pavel-kokolemin: Small update. I think I now understand why the times above are so different. I was stopping the timer when the node appeared to be in sync by stopping printing to the console, and a lot of time was actually spent printing from mempool. I got these times by checking the chainstate RPC instead:
```
Unbounded (current master): 1:29.46
Bounded (20): 1:31.60
Bounded (100): 1:28.44
Bounded (500) 1:30.68
```
So the buffer size of 20 should be to be sufficient.

discosultan (approved):

PR #977, opened 2023-06-20
More blockprod tests

alfiedotwtf:
Unit tests for blockprod.

As I was writing these tests, I've discovered a few minor problems, which I'll create GitHub Issues for.

### DIFF
--- a/blockprod/src/detail/tests.rs
+++ b/blockprod/src/detail/tests.rs
@@ -521,9 +521,7 @@ mod produce_block {
 
                 match result {
                     Err(BlockProductionError::FailedConsensusInitialization(
-                        ConsensusCreationError::PropertyQueryError(
-                            PropertyQueryError::BestBlockIndexNotFound,
-                        ),
+                        ConsensusCreationError::BestBlockIndexNotFound,
                     )) => {}
                     _ => panic!("Unexpected return value"),
                 }
--- a/blockprod/src/detail/tests.rs
+++ b/blockprod/src/detail/tests.rs
@@ -86,8 +86,7 @@ mod collect_transactions {
 
                 let accumulator = block_production.collect_transactions().await;
 
-                let collected_transactions =
-                    mock_mempool.collect_txs_called.load(Ordering::Relaxed);
+                let collected_transactions = mock_mempool.collect_txs_called.load();
                 assert!(collected_transactions, "Expected collect_tx() to be called");
 
                 match accumulator {
@@ -135,7 +134,10 @@ mod collect_transactions {
             let accumulator = block_production.collect_transactions().await;
 
             let collected_transactions = mock_mempool.collect_txs_called.load();
-            assert!(collected_transactions, "Expected collect_tx() to be called");
+            assert!(
+                !collected_transactions,
+                "Expected collect_tx() not to be called"
+            );
 
             match accumulator {
                 Err(BlockProductionError::SubsystemCallError(_)) => {}
--- a/blockprod/src/detail/tests.rs
+++ b/blockprod/src/detail/tests.rs
@@ -636,7 +636,7 @@ mod produce_block {
                     UpgradeVersion::ConsensusUpgrade(consensus_types[next_consensus_type].clone()),
                 ));
 
-                next_height_consensus_change = next_height_consensus_change + rng.gen_range(1..50);
+                next_height_consensus_change += rng.gen_range(1..50);
             }
 
             let net_upgrades =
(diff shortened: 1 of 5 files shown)


common/src/primitives/compact.rs
```diff
@@ -20,6 +20,17 @@ use std::ops::Shl;
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy, Encode, Decode)]
 pub struct Compact(pub u32);
 
+impl Compact {
+    pub fn highest_value() -> Self {
+        Uint256([0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF])
+            .into()
```
iljakuklic, line 26: Would `Uint256::MAX` work here?

common/src/primitives/compact.rs
```diff
@@ -20,6 +20,17 @@ use std::ops::Shl;
...
+    }
+
+    pub fn lowest_value() -> Self {
+        Uint256([0, 0, 0, 0]).into()
```
iljakuklic, line 30: `Uint256::ZERO`

iljakuklic: Also, if the conversion from `Uint256` to `Compact` was made a `const fn` (which should be possible), the `{lowest,highest}_value` could be constants.

blockprod/src/detail/tests.rs
```diff
@@ -400,291 +1411,374 @@ mod produce_block {
...
+            Err(BlockProductionError::JobManagerError(_)) => {}
+            _ => panic!("Unexpected return value"),
+        }
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn ok(#[case] seed: Seed) {
```
iljakuklic, line 1512: A more descriptive test name would be appreciated. Something that suggests what is expected to be ok.

iljakuklic (commented):
A couple of initial comments below:

common/src/primitives/compact.rs
alfiedotwtf: I originally used `MAX`, however since it's not a constant but rather a function, I thought an all-caps method felt wrong and so went with a lowercase function call. But then I didn't want to use `max()` since that might get confused with `std::cmp::max()`.

common/src/primitives/compact.rs
alfiedotwtf: Ah nice.

consensus/src/lib.rs
```diff
@@ -71,6 +71,8 @@ pub enum ConsensusCreationError {
     MiningStopped,
     #[error("Mining failed")]
     MiningFailed,
+    #[error("Property query error: `{0}`")]
+    PropertyQueryError(#[from] PropertyQueryError),
```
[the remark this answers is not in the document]
alfiedotwtf: Yeah, I only did this change because I noticed it during testing.

Did you want me to revert this bit, or leave it as is?

blockprod/src/detail/tests.rs
alfiedotwtf: Yep cool. I thought it would be ok since it's in a separate `mod` per function, but no problem.

common/src/primitives/compact.rs
iljakuklic: > I originally used `MAX`, however since it's not a constant but rather a function, I thought an all-caps method felt wrong and so went with a lowercase function call. But then I didn't want to use `max()` since that might get confused with `std::cmp::max()`.

I meant use `Uint256::MAX`, which is already defined, in place of the rather unpleasing array constant.

blockprod/src/detail/tests.rs
iljakuklic: If it's in a separate mod, that's fine, I'm ok with it either way. It's not obvious when just scrolling through the code though.

iljakuklic (approved):

common/src/primitives/compact.rs
alfiedotwtf: Thanks for that, I must have missed it. Fixed.

common/src/primitives/compact.rs
alfiedotwtf: Same here.

consensus/src/lib.rs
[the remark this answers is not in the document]
alfiedotwtf: No problem. Updated to make it an individual error.

PR #979, opened 2023-06-20
Turn peer activity into an enum and add a test for a disconnect bug

discosultan:
Resolves two TODOs in P2P block sync manager:
1. Turn peer activity from `Option<Duration>` to a new `PeerActivity` enum that better conveys the intent. It also allows to more easily expand the existing peer timeout logic.
2. Add a test for a previously fixed bug where a node unexpectedly disconnected a peer after it had sent a header list with all known blocks.

### DIFF
--- a/p2p/src/sync/peer.rs
+++ b/p2p/src/sync/peer.rs
@@ -56,7 +56,7 @@ use crate::{
         types::services::{Service, Services},
         NetworkingService,
     },
-    types::peer_id::PeerId,
+    types::{peer_activity::PeerActivity, peer_id::PeerId},
     utils::oneshot_nofail,
     MessagingService, PeerManagerEvent, Result,
 };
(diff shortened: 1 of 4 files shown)


p2p/src/sync/peer.rs
```diff
@@ -160,7 +159,6 @@ where
 
         if self.common_services.has_service(Service::Blocks) {
             self.request_headers().await?;
-            self.last_activity = Some(self.time_getter.get_time());
```
discosultan, line 163: I changed when we set `last_activity`. It is now set only _after_ we have resolved a peer action. For example, when a peer sends us a block response, we no longer update the activity immediately, but only after we have processed the response.

p2p/src/sync/peer.rs
```diff
@@ -308,7 +314,6 @@ where
 
     async fn handle_header_list(&mut self, headers: Vec<SignedBlockHeader>) -> Result<()> {
         log::debug!("Headers list from peer {}", self.id());
-        self.last_activity = Some(self.time_getter.get_time());
```
discosultan, line 311: I changed when we set `last_activity`. It is now set only _after_ we have resolved a peer action. For example, when a peer sends us a block response, we no longer update the activity immediately, but only after we have processed the response.

p2p/src/types/peer_activity.rs
```diff
@@ -0,0 +1,33 @@
...
+
+/// Activity with a peer.
+#[derive(Debug)]
+pub enum PeerActivity {
+    /// Node is pending for further actions with a peer.
+    Pending,
+    /// Node has sent a header list request to a peer and is expecting a header list response.
+    ExpectingHeaderList {
+        /// A time when the header list request was sent.
+        time: Duration,
```
discosultan, line 26: Thinking if these should be called `last_updated` instead.

TheQuantumPhysicist (approved):
I don't see a big issue. Though I'd appreciate waiting for others to review too.

pavel-kokolemin (approved):

PR #981, opened 2023-06-20
Fix the random token issuance amount in 'issue_and_transfer_tokens' test

discosultan:
The CI caught [a failure](https://github.com/mintlayer/mintlayer-core/actions/runs/5322405108/jobs/9638806808#step:7:51334) in the following test:
```rust
---- wallet::tests::issue_and_transfer_tokens::case_1 stdout ----
------------ TEST ARGUMENTS ------------
seed = Seed(8346885526136640763)
-------------- TEST START --------------
thread 'wallet::tests::issue_and_transfer_tokens::case_1' panicked at 'cannot sample empty range', /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/rng.rs:134:9
```

Looks like we need to issue a minimum of 2 tokens (was min 1 before) since the test expects to transfer some and ensure there's still an existing balance (min 1) left.

### DIFF
--- a/wallet/src/wallet/tests.rs
+++ b/wallet/src/wallet/tests.rs
@@ -733,7 +733,7 @@ fn issue_and_transfer_tokens(#[case] seed: Seed) {
     assert_eq!(*token_amount, token_amount_to_issue);
 
     let tokens_to_transfer =
-        Amount::from_atoms(rng.gen_range(1..token_amount_to_issue.into_atoms()));
+        Amount::from_atoms(rng.gen_range(1..=token_amount_to_issue.into_atoms()));
 
     let some_other_address = PublicKeyHash::from_low_u64_be(1);
     let new_output = TxOutput::Transfer(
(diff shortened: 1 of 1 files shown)


muursh (approved):

wallet/src/wallet/tests.rs
```diff
@@ -681,7 +681,7 @@ fn issue_and_transfer_tokens(#[case] seed: Seed) {
 
     let amount_fraction = (block1_amount.into_atoms() - NETWORK_FEE) / 10;
 
-    let token_amount_to_issue = Amount::from_atoms(rng.gen_range(1..amount_fraction));
+    let token_amount_to_issue = Amount::from_atoms(rng.gen_range(2..amount_fraction));
```
iljakuklic, line 684: > Looks like we need to issue a minimum of 2 tokens (was min 1 before) since the test expects to transfer some and ensure there's still an existing balance (min 1) left.

Not critical but would be nice if this was mentioned in a comment here.

iljakuklic (approved):

wallet/src/wallet/tests.rs
OBorce, line 684: 🤦 same bug as before, I thought I fixed it, apparently not. 
the actual bug is on line 736,
`Amount::from_atoms(rng.gen_range(1..token_amount_to_issue.into_atoms()));`
you can change it to 
`Amount::from_atoms(rng.gen_range(1..=token_amount_to_issue.into_atoms()));`
but then you will also need to handle the case when all of the tokens are transferred so line 778 will be empty, as the balance will be 0.

OBorce: I guess assert for <= 1 or remove the assert altogether?, and then get first() or Amount::ZERO

TheQuantumPhysicist:
Guys, please fix this ASAP so that CT isn't obstructed for a long time. I already got this error in all CT systems.

TheQuantumPhysicist (approved):

wallet/src/wallet/tests.rs
discosultan: Went with these changes.

PR #982, opened 2023-06-21
SeqCst atomics were added; other atomics were renamed.

ImplOfAnImpl:
Here I add the SeqCstAtomic. The main reason is that currently many atomics in the codebase use this ordering and, when replacing them with "simple atomics", I only wanted to ensure consistency, while not changing the existing semantics (at least for now).
The second reason is that it makes sense to have this variant just for completeness.

Because of this addition, I also decided to rename the old ones. The reason is that in the presence of SeqCst, the Syn prefix became kind of confusing, because it suggested that SeqCst is not a "synchronizing" atomic.

PR #984, opened 2023-06-21
time_value in TestFramework is now a simplified atomic

ImplOfAnImpl:
time_value in TestFrameword, which used SeqCst ordering, was replaced with `SeqCstAtomicU64`.

This PR depends on PR #982.

### DIFF
--- a/chainstate/src/detail/median_time.rs
+++ b/chainstate/src/detail/median_time.rs
@@ -60,7 +60,8 @@ mod test {
         },
         primitives::Idable,
     };
-    use std::sync::{atomic::Ordering, Arc};
+    use std::sync::Arc;
+    use utils::atomics::SeqCstAtomicU64;
 
     fn make_block(prev_block: Id<GenBlock>, time: BlockTimestampInternalType) -> Block {
         Block::new(
@@ -166,7 +167,7 @@ mod test {
         utils::concurrency::model(|| {
             let chain_config = Arc::new(create_unit_test_config());
 
-            let current_time = Arc::new(std::sync::atomic::AtomicU64::new(
+            let current_time = Arc::new(SeqCstAtomicU64::new(
                 chain_config.genesis_block().timestamp().as_int_seconds(),
             ));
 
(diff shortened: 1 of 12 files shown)


muursh (approved):

iljakuklic (approved):

alfiedotwtf (approved):

chainstate/src/detail/median_time.rs
```diff
@@ -185,11 +186,11 @@ mod test {
...
-            let block1_time = current_time.load(Ordering::SeqCst) + 1;
-            let block2_time = current_time.load(Ordering::SeqCst) + 10;
-            let block3_time = current_time.load(Ordering::SeqCst) + 17;
-            let block4_time = current_time.load(Ordering::SeqCst) + 18;
-            let block5_time = current_time.load(Ordering::SeqCst) + 20;
+            let block1_time = current_time.load() + 1;
+            let block2_time = current_time.load() + 10;
+            let block3_time = current_time.load() + 17;
+            let block4_time = current_time.load() + 18;
+            let block5_time = current_time.load() + 20;
```
TheQuantumPhysicist, line 193: There's something something simple I wanted to fix these but never had the time. Would be great if you could. There's a story there.

When I developed the BlockIndexHistoryIterator, this test was meant to have non-sequential time. But then when PoS was developed, we added a restriction to check_block to enforce time-ordering of blocks, which then was reverted and this time-ordering was enforced in PoS consensus validity check. So this current ordering doesn't cover the case where consecutive blocks have unordered timestamps.

Someone should revert the order of these times and ensure that the old block times are used, where no order is enforced. Please check the history to see when this change was made a couple of months ago.

TheQuantumPhysicist (approved):

chainstate/src/detail/median_time.rs
ImplOfAnImpl: Done in PR #998 (if I've understood you correctly)

TheQuantumPhysicist: Yep. Thanks.

PR #985, opened 2023-06-21
MempoolInterfaceMock now uses simple atomics

ImplOfAnImpl:
A bunch of atomics in MempoolInterfaceMock, which used the Relaxed ordering, were replaced with RelaxedAtomicBool.

This PR depends on PR #982.

### DIFF
--- a/mocks/src/mempool.rs
+++ b/mocks/src/mempool.rs
@@ -15,10 +15,7 @@
 
 #![allow(clippy::unwrap_used)]
 
-use std::sync::{
-    atomic::{AtomicBool, Ordering::Relaxed},
-    Arc,
-};
+use std::sync::Arc;
 
 use common::{
     chain::{GenBlock, SignedTransaction, Transaction},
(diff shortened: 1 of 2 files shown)


iljakuklic (approved):
I am fine with this change.

It would, however, be preferable to use something like `mockall` for the whole thing. What was the reason for not using it in the first place and is it still relevant? @alfiedotwtf

muursh (approved):

alfiedotwtf:
I can't remember the original reason, but IIRC it either had something to do with not implementing `Subsystem` within a `mockall` or another case of circular dependancies. However, a few weeks ago I gave it another go and got it working. So to convert this to `mockall`, I just need to cleanup my WIP repo, update the tests that currently rely on `MempoolInterfaceMock`, and then merge - I'll do this once my current run of tests is complete.

I'll create an issue for this so it's not forgotten about just in case.

alfiedotwtf:
As for this change, I think it's fine to include it along with the all the other atomic simplification PR.

alfiedotwtf (approved):

mocks/src/mempool.rs
```diff
@@ -30,21 +27,22 @@ use mempool::{
...
-    pub get_all_should_error: Arc<AtomicBool>,
-    pub contains_transaction_called: Arc<AtomicBool>,
-    pub contains_transaction_should_error: Arc<AtomicBool>,
-    pub collect_txs_called: Arc<AtomicBool>,
-    pub collect_txs_should_error: Arc<AtomicBool>,
-    pub subscribe_to_events_called: Arc<AtomicBool>,
-    pub subscribe_to_events_should_error: Arc<AtomicBool>,
-    pub run_called: Arc<AtomicBool>,
-    pub run_should_error: Arc<AtomicBool>,
+    pub add_transaction_called: Arc<RelaxedAtomicBool>,
```
TheQuantumPhysicist, line 34: Can you please replace all these (in all mocks) with SeqCst or AcqRel?

PR #986, opened 2023-06-21
Miscellaneous atomics-related changes

ImplOfAnImpl:
1) blockuntilzero and counttracker in utils now use custom (loom-aware) atomic traits; because of this, the "atomic-traits" crate is no longer needed, so I removed it from the project.
2) in some other places std atomics were replaced with utils::sync atomics.
(Plus I also replaced one more atomic with a simplified one).

As I've mentioned during the daily, I wasn't able to replace all existing atomics with the simplified ones or at least with loom versions of the standard ones because:
1) Loom atomics' "new" function is not const, so static atomics won't compile. The solution might be to use lazy_static but that kind of defeats the purpose of the atomic, because lazy_static itself uses synchronization primitives under the hood (that are probably heavier than a single atomic).
2) Sometimes code that uses atomics is called from tests that are not "loom-ready" (i.e. they don't call concurrency::model). Probably, we should disable those tests under cfg(loom).
The atomic in test-utils/src/test_dir.rs is one of such atomics.
3) In one case tests failed with the error "Model exceeded maximum number of branches. This is often caused by an algorithm requiring the processor to make progress, e.g. spin locks.", this happened after replacing the atomic in storage/lmdb/src/lib.rs. Googling suggests that this might be fixable by increasing the "max_branches" setting of the loom model.

For now just I added comments about the issues. I don't plan to fix them in the nearby future.

There are also a couple of FIXMEs, that I would like to discuss.

### DIFF
--- a/p2p/src/sync/mod.rs
+++ b/p2p/src/sync/mod.rs
@@ -31,10 +31,8 @@ use common::{
 };
 use logging::log;
 use mempool::MempoolHandle;
-use utils::sync::{
-    atomic::{AtomicBool, Ordering},
-    Arc,
-};
+use utils::atomics::AcqRelAtomicBool;
+use utils::sync::Arc;
 use utils::tap_error_log::LogError;
 
 use crate::{
@@ -69,7 +67,7 @@ pub struct BlockSyncManager<T: NetworkingService> {
     mempool_handle: MempoolHandle,
 
     /// A cached result of the `ChainstateInterface::is_initial_block_download` call.
-    is_initial_block_download: Arc<AtomicBool>,
+    is_initial_block_download: Arc<AcqRelAtomicBool>,
 
     /// The list of connected peers
     peers: HashSet<PeerId>,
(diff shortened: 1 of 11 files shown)


p2p/src/sync/mod.rs
```diff
@@ -188,6 +186,12 @@ where
 
     /// Announces the header of a new block to peers.
     async fn handle_new_tip(&mut self, block_id: Id<Block>) -> Result<()> {
+        // FIXME: I'd like to make is_initial_block_download a simple atomic too,
```
TheQuantumPhysicist, line 189: You can use AcqRel for the whole thing.

storage/lmdb/src/lib.rs
```diff
@@ -142,6 +145,10 @@ impl LmdbImpl {
         })
     }
 
+    // FIXME: I'd like to replace the SeqCst ordering here with Release, because it looks
```
TheQuantumPhysicist, line 148: I don't mind. Please check with Lukas.

TheQuantumPhysicist, line 148: It's designed to be an overkill. I don't mind though. Feel free to change it to release.

TheQuantumPhysicist (approved):

muursh (approved):

PR #988, opened 2023-06-21
Derive Hash for Id

discosultan:
Small change but to a primitive type. This PR serves as a place to discuss whether it is viable to derive `Hash` for `Id<T>`.

It is often useful to track entities in a hashmap where the key is the entity ID (i.e `HashMap<Id<_>, _>`. Currently, we'd need to represent it as `HashMap<H256, _>` and then use `Id<_>::get(&self)` before accessing the hashmap.

### DIFF
--- a/common/src/primitives/id/mod.rs
+++ b/common/src/primitives/id/mod.rs
@@ -89,7 +89,7 @@ impl<'de> serde::Deserialize<'de> for H256 {
     }
 }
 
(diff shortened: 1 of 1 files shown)


TheQuantumPhysicist:
Please, no. Use BTreeMap instead, which most of the time is more than enough. We avoid HashMap and HashSet as much as possible because it can easily cause confusion because we have hashing for data to H256.

discosultan:
Thanks for the input @TheQuantumPhysicist, makes sense!

PR #989, opened 2023-06-21
Keep track of unconfirmed transactions in the Wallet

OBorce:
- Added mempool rpc client and new get_all_transactions
- after syncing all of the blocks, scan all of the mempool transactions as well
- now able to use utxos from unconfirmed transactions

Priority will need to be added later to favor confirmed utxos or unconfirmed but sent from this wallet vs unconfirmed that are being sent from another wallet.
Some better way to scan the mempool instead of pulling all of the transactions at once.


wallet/src/wallet/mod.rs
```diff
@@ -283,7 +283,7 @@ impl<B: storage::Backend> Wallet<B> {
             .accounts
             .get(&account_index)
             .ok_or(WalletError::NoAccountFoundWithIndex(account_index))?;
-        let utxos = account.get_utxos(utxo_types, self.latest_median_time);
+        let utxos = account.get_utxos(utxo_types, self.latest_median_time, false);
```
OBorce, line 286: Should we make this an enum instead of bool, with support for Unconfirmed and/or Locked Utxos?

wallet/src/account/output_cache/mod.rs
```diff
@@ -60,35 +59,50 @@ impl OutputCache {
...
     pub fn remove_tx(&mut self, tx_id: &AccountWalletTxId) {
-        let tx_opt = self.txs.remove(tx_id);
+        let tx_opt = self.txs.remove(tx_id.item_id());
         if let Some(tx) = tx_opt {
             for input in tx.inputs() {
                 match input {
                     TxInput::Utxo(outpoint) => {
                         self.consumed.remove(outpoint);
+                        self.unconfirmed_consumed.remove(outpoint);
                     }
```
alfiedotwtf, line 106: Maybe overkill, but instead of calling remove() on both, should this also do `match tx.state()` here and then call the right one? May not matter now, but just thinking in the future if more states would ever be added?

wallet/src/wallet/mod.rs
alfiedotwtf: I was thinking the bool could be combined with `utxo_types`, but looking at it I think it's not really a UTXO type but a state, so I would keep it separate. Maybe a UTXO State would be useful, then you could add `Confirmed`, `Unconfirmed`, `Valid`, `Invalid`, `Orphaned` etc.

alfiedotwtf (approved):

wallet/src/account/output_cache/mod.rs
OBorce: One thing I am thinking now is to maybe change it from 2 x Set<UtxoOutPoint> to a Map<UtxoOutPoint, TxState>, where the value is the TxState of the transaction that consumed it.

wallet/src/account/mod.rs
```diff
@@ -658,6 +671,42 @@ impl Account {
...
+
+        // check them again after adding all we could
+        for signed_tx in not_added {
+            let wallet_tx = WalletTx::Tx(TxData::new(
+                signed_tx.transaction().clone().into(),
+                TxState::InMempool,
+            ));
+
+            self.add_wallet_tx_if_relevant(db_tx, wallet_tx)?;
+        }
```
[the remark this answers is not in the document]
OBorce: From my understanding no, as no matter the chain length in every transaction of the chain up until the last one you will always have an output that belongs to this wallet which will be identified in the first for loop. The last one that doesn't have an output but only consumes an input will be caught in the second one.
If a long chain has transactions that don't have neither an output nor an input belonging to the wallet, then you can think of that chain as multiple chains from the parts that we care about.

wallet/src/wallet/mod.rs
OBorce: We have a TxState enum that is used, but this is a different thing, it is supposed to be an enum from wallet commands, e.g. for get_balance/ show utxos to show only confirmed balance, or unconfirmed, or confirmed including still locked utxos...

wallet/src/account/mod.rs
```diff
@@ -146,8 +146,12 @@ impl Account {
         let network_fee = Amount::from_atoms(10000);
 
         let utxos_by_currency = group_utxos_for_input(
-            self.get_utxos(UtxoType::Transfer | UtxoType::LockThenTransfer, median_time)
-                .into_iter(),
+            self.get_utxos(
+                UtxoType::Transfer | UtxoType::LockThenTransfer,
+                median_time,
+                true,
```
TheQuantumPhysicist, line 152: This should be controllable by the user at the wallet controller level.

wallet/src/account/mod.rs
```diff
@@ -529,7 +535,7 @@ impl Account {
         median_time: BlockTimestamp,
     ) -> WalletResult<BTreeMap<Currency, Amount>> {
         let amounts_by_currency = group_utxos_for_input(
-            self.get_utxos(utxo_types, median_time).into_iter(),
+            self.get_utxos(utxo_types, median_time, false).into_iter(),
```
TheQuantumPhysicist, line 538: It should be possible to get the balance, confirmed and not confirmed. The GUI may need to see which is which (the CLI as well). Knowing whether the balance is confirmed or not is important.

wallet/wallet-controller/src/sync/mod.rs
```diff
@@ -92,6 +98,19 @@ pub async fn sync_once<T: NodeInterface>(
             wallet.best_block().map_err(ControllerError::WalletError)?;
 
         if chain_info.best_block_id == wallet_block_id {
+            // if everything is synced up scan the mempool
+            let transactions: Vec<_> = rpc_client
+                .get_all_mempool_transactions()
```
TheQuantumPhysicist, line 103: This isn't really a good idea. The mempool can potentially have hundreds of thousands of transactions. At best this will take forever, and at worst this will timeout your RPC layer and will make your whole sync function fail.

This needs some smart design where we use some kind of pub/sub to get this information. Now the problem is that our RPC doesn't inherently support pub/sub since it's http based (which Pavel got around in the wallet for blocks by polling), so the only solution (I have in mind) is to use some kind of (bloom, or otherwise) filter to tell the mempool that we're interested in some specific transactions, and then keep polling for them. This is more complex than it looks and has more complex design requirements. Pavel has implemented a bloom filter for p2p already, btw. But bloom isn't the only way to do this.

Another solution is to make the wallet launch a p2p node. This way it can receive transactions like nodes and filter them based on whether it's interested in their outcome. This is what light wallets do, like Electrum.

This is one of the reasons why I was asking to postpone this whole problem. It's unclear how we're gonna solve it and it's absolutely not a priority. The only thing that gives this a priority is making the wallet nicer, where users feel that the node is snappy (in receiving transactions), despite it being deceiving since transactions will show up in the wallet while not confirmed yet.

What I recommend now is putting this PR aside, and working on whatever is prior on the queue. Once we decide to work on this problem, we should come up with a reasonable design that takes everything into consideration.

wallet/src/account/mod.rs
OBorce: Added utxo_states as argument for get_balance and get_utxos

wallet/src/account/mod.rs
OBorce: added an argument for controlling this

wallet/wallet-controller/src/sync/mod.rs
OBorce: removed the sync from mempool

wallet/wallet-node-client/src/node_traits.rs
```diff
@@ -57,4 +58,8 @@ pub trait NodeInterface {
     async fn p2p_get_connected_peers(&self) -> Result<Vec<ConnectedPeer>, Self::Error>;
     async fn p2p_add_reserved_node(&self, address: String) -> Result<(), Self::Error>;
     async fn p2p_remove_reserved_node(&self, address: String) -> Result<(), Self::Error>;
+
+    async fn get_all_mempool_transactions(
```
TheQuantumPhysicist, line 62: This has to go, unfortunately. It encourages the bad practice of pulling all transactions in from the mempool. I see it isn't used anywhere.

wallet/types/src/utxo_types.rs
```diff
@@ -29,6 +31,16 @@ pub enum UtxoType {
     DelegateStaking = 1 << 6,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u16)]
+pub enum UtxoState {
```
TheQuantumPhysicist, line 36: Are these supposed to be mutually exclusive? If they are, then having them encoded as bit flags is a bad idea. Besides that, I think this structure is too forgiving and lacks checks. I don't have anything specific in mind, but can't we code some sanity verification within it, so that if something goes wrong we can detect it?

wallet/types/src/wallet_tx.rs
```diff
@@ -35,6 +37,22 @@ pub enum TxState {
...
+    /// abandoned by the user
+    #[codec(index = 4)]
+    Abandoned,
+}
+
+impl Display for TxState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            TxState::Confirmed(_, _) => f.write_str("Confirmed"),
+            TxState::Conflicted(_) => f.write_str("Conflicted"),
```
TheQuantumPhysicist, line 50: For both Confirmed and Conflicted, can't we add the information of the arms? Like can't we instead of "Confirmed", have it as "Confirmed(1,2)"? Any reason we're not doing that?

wallet/wallet-cli-lib/src/commands/helper_types.rs
```diff
@@ -42,3 +44,47 @@ impl CliUtxoTypes {
...
+    Conflicted,
+    InMempool,
+    Inactive,
+    Abandoned,
+}
+
+impl Display for CliUtxoState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CliUtxoState::Confirmed => f.write_str("confirmed"),
```
TheQuantumPhysicist, line 60: I know this is dumb... but please capitalize the strings since they're independent states and not part of something else, linguistically speaking.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -145,18 +145,31 @@ pub enum WalletCommand {
         transaction: HexEncoded<SignedTransaction>,
     },
 
+    /// Abandon an unconfirmed transaction, and make the consumed inputs available to be used again
+    AbandonTransaction {
```
TheQuantumPhysicist, line 149: Shouldn't there also be a command to list all the transactions (or their ids) that can be abandoned, I.e., pending? Something like `ListPendingTransactions` is probably needed.

wallet/wallet-controller/src/sync/tests/mod.rs
```diff
@@ -112,6 +113,10 @@ impl SyncingWallet for MockWallet {
     ) -> WalletResult<()> {
         Err(wallet::WalletError::NoUnsyncedAccount)
     }
+
+    fn scan_mempool(&mut self, _transactions: &[SignedTransaction]) -> WalletResult<()> {
```
TheQuantumPhysicist, line 117: This should go, right?

wallet/src/account/output_cache/mod.rs
```diff
@@ -151,4 +187,84 @@ impl OutputCache {
 
         utxos
     }
+
+    pub fn get_abandonable_transactions(&self) -> Vec<&WithId<Transaction>> {
```
TheQuantumPhysicist, line 191: Should we call these `pending_transactions`?

wallet/src/account/output_cache/mod.rs
```diff
@@ -151,4 +187,84 @@ impl OutputCache {
...
+                    WalletTx::Tx(tx) => match tx.state() {
+                        TxState::Inactive => {
+                            tx.set_state(TxState::Abandoned);
+                            for input in tx.get_transaction().inputs() {
+                                match input {
+                                    TxInput::Utxo(outpoint) => {
+                                        self.consumed.insert(outpoint.clone(), *tx.state());
+                                    }
+                                    TxInput::Account(_) => {
+                                        unimplemented!()
```
TheQuantumPhysicist, line 228: What's the story here? Please add a comment explaining what's required and what's stopping us from implementing it.

wallet/src/account/output_cache/mod.rs
```diff
@@ -151,4 +187,84 @@ impl OutputCache {
...
+                },
+                Entry::Vacant(_) => Err(WalletError::CannotFindTransactionWithId(tx_id)),
+            }?;
+        }
+
+        Ok(())
+    }
+}
+
+// TODO: similar code from tx verifier
```
TheQuantumPhysicist, line 245: You mean use the code from tx verifier? If yes, what's stopping you?

wallet/src/account/output_cache/mod.rs
```diff
@@ -151,4 +187,84 @@ impl OutputCache {
...
+            })
+            .collect()
+    }
+
+    pub fn abandon_transaction(&mut self, tx_id: Id<Transaction>) -> WalletResult<()> {
+        let mut to_abandon = BTreeSet::new();
+        to_abandon.insert(OutPointSourceId::from(tx_id));
+
+        while let Some(outpoint_source_id) = to_abandon.pop_first() {
+            if let Some(descendants) = self.unconfirmed_descendants.remove(&outpoint_source_id) {
```
TheQuantumPhysicist, line 212: Do we have tests for abandoning transactions with descendants?

TheQuantumPhysicist (approved):
Looking good. Let's discuss these comments.

wallet/src/account/output_cache/mod.rs
OBorce: I haven't looked at it closely, the TxInput::Account is unimplemented in all of the places in the wallet, but there is a point in the wallet issues for it.

wallet/src/account/output_cache/mod.rs
OBorce: It is not exactly the same and I was not sure if I should introduce the dependency.

wallet/src/account/output_cache/mod.rs
OBorce: yes

wallet/types/src/utxo_types.rs
OBorce: not mutually exclusive, it is the same bit flag approach as the above one for UtxoType

wallet/wallet-cli-lib/src/commands/helper_types.rs
TheQuantumPhysicist: Alright. No problem. Please add a comment explaining this.

wallet/src/account/output_cache/mod.rs
TheQuantumPhysicist: Alrighty. Please add this somewhere so that we don't forget this spot. This is actually quite important to be supported.

wallet/src/account/output_cache/mod.rs
TheQuantumPhysicist: What's the difference? Can't one be a superset of the other?

wallet/wallet-node-client/src/node_traits.rs
OBorce: removed

wallet/types/src/wallet_tx.rs
OBorce: added more information

wallet/wallet-cli-lib/src/commands/helper_types.rs
OBorce: commented

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: Renamed ListAbandonableTransactions to ListPendingTransactions

wallet/src/account/mod.rs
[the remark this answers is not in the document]
pavel-kokolemin: I think there should be a loop. It is possible that the wallet does not recognize some transactions as its own because the address pool is not big enough (for example, if you use two wallets with the same mnemonic and the remote wallet tries to scan mempool).

pavel-kokolemin (approved):

wallet/src/wallet/mod.rs
```diff
@@ -494,8 +536,18 @@ impl<B: storage::Backend> Wallet<B> {
     }
 
     /// Rescan mempool for unconfirmed transactions and UTXOs
-    pub fn scan_mempool(&mut self, _transactions: Vec<SignedTransaction>) -> WalletResult<()> {
-        Err(WalletError::NotImplemented("scan_mempool"))
+    pub fn scan_mempool(&mut self, transactions: &[SignedTransaction]) -> WalletResult<()> {
```
pavel-kokolemin, line 539: `scan_mempool` is not called from the CLI wallet.

wallet/src/account/mod.rs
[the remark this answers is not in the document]
OBorce: fixed

PR #993, opened 2023-06-22
Separate GUI backend

pavel-kokolemin:
I think it would be good to split the GUI node into two functional parts: backend and UI, and only allow message passing between them.
Pros:
- Simplify testing (the backend can be started standalone and in testing is as simple as sending plain messages and expecting responses).
- Easier debugging (all sent/received messages between UI/backend can be logged and shared).
- Faster development (UI/backend work can be done independently).
- Easier to rewrite/replace the UI (for example if we are not happy with Iced), and thus less risky overall.

Iced is already message oriented, so this way should work well.


TheQuantumPhysicist:
Thank you, Pavel. This is a necessary thing. One thing also to add is that the node-lib uses command line arguments, which should be separated, so that unit tests work.

One problem at a time though.

wallet/wallet-controller/src/lib.rs
```diff
@@ -347,36 +363,108 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+                Some(pool_id)
+            }
+            TxOutput::Transfer(_, _)
+            | TxOutput::LockThenTransfer(_, _, _)
+            | TxOutput::Burn(_)
+            | TxOutput::CreateDelegationId(_, _)
+            | TxOutput::DelegateStaking(_, _) => None,
+        });
+        let mut balances = BTreeMap::new();
+        for pool_id in pool_ids {
```
TheQuantumPhysicist, line 417: It seems to me this can be a filter_map followed by collect to that BTreeMap.

TheQuantumPhysicist: Oh, never mind. This is async.

wallet/wallet-controller/src/lib.rs
```diff
@@ -347,36 +363,108 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
 
             if let Err(e) = sync_res {
                 log::error!("Wallet sync error: {e}");
                 tokio::time::sleep(ERROR_DELAY).await;
                 continue;
             }
 
-            if let Some(account_index) = self.staking_started {
-                let generate_res = self.generate_block(account_index, None).await;
+            for account_index in self.staking_started.clone().iter() {
```
TheQuantumPhysicist, line 466: Is this clone() necessary?

wallet/src/wallet_events.rs
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use wallet_types::{AccountWalletTxId, WalletTx};
+
+pub trait WalletEvents {
```
TheQuantumPhysicist, line 18: Please document the purpose of this... What problem it solves.

I'm trying to understand why everything here is `&mut self`, too.

wallet/src/wallet_events.rs
```diff
@@ -0,0 +1,30 @@
...
+    fn new_block(&mut self);
+    fn set_transaction(&mut self, id: &AccountWalletTxId, tx: &WalletTx);
+    fn del_transaction(&mut self, id: &AccountWalletTxId);
+}
+
+pub struct WalletEventsNop;
```
TheQuantumPhysicist, line 24: NoOp?

node-gui/src/backend/wallet_events/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+        self.updated
+    }
+
+    pub fn reset(&mut self) {
+        self.updated = false;
+    }
+}
+
+impl WalletEvents for GuiWalletEvents {
+    fn new_block(&mut self) {
```
TheQuantumPhysicist, line 50: From the looks of it, these functions shouldn't be mutable. It only complicates things. It seems to me that the right approach is to use some kind of channels to isolate this mutability.

node-gui/src/backend/mod.rs
```diff
@@ -0,0 +1,755 @@
...
+    wallet_events: GuiWalletEvents,
+    best_block: (Id<GenBlock>, BlockHeight),
+    accounts: BTreeMap<AccountId, AccountData>,
+}
+
+struct AccountData {
+    transaction_list_skip: usize,
+
+    /// If set, pool balances should be updated in the UI.
+    /// This is necessary because the pool balances load requres RPC call and may fail.
```
TheQuantumPhysicist, line 82: requres

node-gui/src/backend/mod.rs
TheQuantumPhysicist, line 1: Please try to break this file down to multiple files, and also please document the different structs and explain their purpose...

Too much code in one file the structure isn't really clear.

node-gui/src/main_window/main_widget/tabs/wallet.rs
```diff
@@ -0,0 +1,555 @@
...
+
+    // `next_height` is used to prevent flickering when a new block is found
+    let show_scan_progress = wallet_info.best_block.1.next_height() < node_state.best_block_height;
+    let scan_progress_widget = if show_scan_progress {
+        // TODO: Fix scan progress when the node is in the initial block download state
+        let scan_progress = wallet_scan_progress(
+            node_state.best_block_height.into_int(),
+            wallet_info.best_block.1.into_int(),
+        );
+        let scan_progress_str = format!("{:.0}%", scan_progress);
```
TheQuantumPhysicist, line 387: Please also add (1234/5555 blocks) to that text. It should be intuitive for people to understand what this is.

node-gui/src/main_window/main_widget/tabs/wallet.rs
```diff
@@ -0,0 +1,555 @@
...
+}
+
+fn view_top_panel(
+    chain_config: &ChainConfig,
+    wallet_info: &WalletInfo,
+    account: &AccountInfo,
+) -> Element<'static, WalletMessage> {
+    let balance = account.balance.get(&Currency::Coin).cloned().unwrap_or(Amount::ZERO);
+    let balance = print_coin_amount(chain_config, balance);
+    let balance = format!("{} Mlt", balance);
```
TheQuantumPhysicist, line 287: Please use a variable from ChainConfig instead of hard-coded Mlt.

node-gui/src/main_window/main_widget/tabs/wallet.rs
TheQuantumPhysicist, line 1: There's a general problem with the wallet, which is that nothing works while the wallet is syncing. Clicking on anything just takes a long time. It seems that clicks/events get queued in the event queue and take a while to pop out.

node-gui/src/main_window/main_widget/tabs/mod.rs
```diff
@@ -13,104 +13,132 @@
...
     Summary(SummaryMessage),
+    Network(NetworkMessage),
     Settings(SettingsMessage),
+    WalletMessage(WalletId, WalletMessage),
 }
 
 pub struct TabsWidget {
     active_tab: usize,
     summary_tab: SummaryTab,
+    network_tab: NetworkTab,
```
TheQuantumPhysicist, line 50: networking

We use the word "networking" because "network" can be confused with the network being mainnet, testnet, etc.

node-gui/src/main_window/main_widget/tabs/network.rs
```diff
@@ -0,0 +1,80 @@
...
+use iced_aw::{tab_bar::TabLabel, Grid};
+
+use crate::main_window::NodeState;
+
+use super::{Tab, TabsMessage};
+
+#[derive(Debug, Clone)]
+pub enum NetworkMessage {}
+
+pub struct NetworkTab {}
```
TheQuantumPhysicist, line 31: NetworkingTab

node-gui/src/main_window/main_widget/tabs/wallet.rs
```diff
@@ -0,0 +1,555 @@
...
+    name: String,
+}
+
+impl std::fmt::Display for AccountPickItem {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(&self.name, f)
+    }
+}
+
+fn view_left_panel(
```
TheQuantumPhysicist, line 334: Is it possible to make the left panel resizable with the mouse?

node-gui/src/main_window/main_widget/tabs/wallet.rs
```diff
@@ -0,0 +1,555 @@
...
+            panel_button("Send", SelectedPanel::Send, selected_panel),
+            panel_button("Stake", SelectedPanel::Stake, selected_panel),
+        ]
+        .height(Length::Fill),
+        scan_progress_widget,
+    ]
+    .into()
+}
+
+fn view_addresses(account: &AccountInfo) -> Element<'static, WalletMessage> {
```
TheQuantumPhysicist, line 424: Please separate the different views into submodules so that they can be expanded.

TheQuantumPhysicist:
Looking good. I've made many suggestions, but I'll take another look after the fixes are done. Cheers!

wallet/wallet-controller/src/lib.rs
pavel-kokolemin: Yes, because `generate_block` in the loop below is mutable.

node-gui/src/main_window/mod.rs
```diff
@@ -13,75 +13,605 @@
...
+    value.into_fixedpoint_str(chain_config.coin_decimals())
+}
+
+fn print_timestamp(timestamp: Duration) -> Option<String> {
+    let timestamp: i64 = timestamp.as_secs().try_into().ok()?;
+    let timestamp = chrono::NaiveDateTime::from_timestamp_opt(timestamp, 0)?;
+    Some(timestamp.format("%Y-%m-%d %H:%M:%S").to_string())
+}
+
+fn print_block_timestamp(timestamp: BlockTimestamp) -> Option<String> {
```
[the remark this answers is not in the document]
pavel-kokolemin: Done (printed in the local time zone in the format `2023-07-23 15:59:59`).

node-gui/src/main_window/main_widget/tabs/wallet.rs
pavel-kokolemin: I did not find a better way and implemented it using the `pane_grid` widget.

node-gui/src/main_window/mod.rs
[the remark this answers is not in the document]
pavel-kokolemin: Timestamps are now printed in UTC.

node-gui/src/main_window/main_widget/tabs/wallet.rs
pavel-kokolemin: Thanks for the report, I did not notice such a problem and will test more. But I'm going to change the way we download blocks from the node (to speed up the wallet scan), which may improve the situation, so let's leave it as is for now.

node-gui/src/backend/wallet_events/mod.rs
pavel-kokolemin: I tried to use channels but I did not like the result. The main problem with channels is that we only need to update the UI once, but with channels we can get many duplicate events.
I don't see the problem with mutability.

wallet/src/wallet_events.rs
pavel-kokolemin: Comments added. `&mut self` is not a problem here, so I don't see why we should change it.

pavel-kokolemin:
@TheQuantumPhysicist I noticed that some mainnet/testnet address prefixes were the same on different networks. I think this was a bug, so I cleaned them up a bit. I hope that's OK.

TheQuantumPhysicist:
> @TheQuantumPhysicist I noticed that some mainnet/testnet address prefixes were the same on different networks. I think this was a bug, so I cleaned them up a bit. I hope that's OK. 

Thank you! It seems to have been a typo.

common/src/chain/config/mod.rs
```diff
@@ -155,6 +155,7 @@ pub struct ChainConfig {
     version: SemVer,
     target_block_spacing: Duration,
     coin_decimals: u8,
+    coin_name: &'static str,
```
TheQuantumPhysicist, line 158: coin_ticker

common/src/chain/config/mod.rs
```diff
@@ -256,6 +257,12 @@ impl ChainConfig {
         self.coin_decimals
     }
 
+    /// The coin name
+    #[must_use]
+    pub fn coin_name(&self) -> &'static str {
```
TheQuantumPhysicist, line 262: coin_ticker

common/src/chain/mlt.rs
```diff
@@ -30,6 +30,8 @@ impl Mlt {
     pub const ZERO: Self = Self(Amount::from_atoms(0));
     /// Maximum representable amount of MLTs
     pub const MAX: Self = Self(Amount::MAX);
+    /// The coin name
+    pub const NAME: &'static str = "Mlt";
```
TheQuantumPhysicist, line 34: `TICKER: &'static str = "ML"`

We decided to move away from MLT. I guess it has to be changed in many places.

mempool/src/error/mod.rs
```diff
@@ -61,7 +61,7 @@ pub enum MempoolPolicyError {
     TransactionFeeLowerThanConflictsWithDescendants,
     #[error("Underflow in computing transaction's additional fees.")]
     AdditionalFeesUnderflow,
-    #[error("Transaction does not pay sufficient fees to be relayed.")]
+    #[error("Transaction does not pay sufficient fees to be relayed (tx_fee: {tx_fee:?}, relay_fee: {relay_fee:?}).")]
```
TheQuantumPhysicist, line 64: This problem seems to apply to many others... for some reason the whole team has been neglecting including the error information in the errors. Everyone, please let's be more strict about this. It helps in debugging.

node-gui/src/main_window/main_widget/tabs/wallet.rs
TheQuantumPhysicist: Looking fine. Thanks.

common/src/chain/mlt.rs
pavel-kokolemin: > I guess it has to be changed in many places.

I fixed it here, but let's please do other renames in another PR.

node-gui/src/backend/mod.rs
pavel-kokolemin: I moved it to a new file (`backend_impl.rs`), but it's not much smaller.

node-gui/src/main_window/main_widget/tabs/wallet.rs
pavel-kokolemin: OK, I noticed this too. If the UI is not able to process all the messages in time, they fill up the channel and increase the latency. I fixed this by using a bounded channel (the backend event loop is be paused once it's full).

wallet/wallet-controller/src/lib.rs
```diff
@@ -225,8 +241,10 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
             .map_err(ControllerError::WalletError)
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub async fn issue_new_token(
         &mut self,
+        wallet_events: &mut impl WalletEvents,
```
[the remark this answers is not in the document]
pavel-kokolemin: Seems to work well, thanks for the hint.

common/src/chain/mlt.rs
TheQuantumPhysicist: Yes that's OK, but you still are calling this "name" in the places you defined.

node-gui/src/main_window/main_widget/tabs/wallet.rs
TheQuantumPhysicist: Cool. Thanks!

node-gui/src/backend/wallet_events/mod.rs
pavel-kokolemin: I switched to channels and made the `WalletEvents` trait non-mutable.

common/src/chain/mlt.rs
pavel-kokolemin: Should be fixed now, thanks.

TheQuantumPhysicist: Cheers!

TheQuantumPhysicist (approved):

node-gui/src/main_window/main_widget/tabs/networking.rs
```diff
@@ -0,0 +1,80 @@
...
+        let field = |text: String| container(Text::new(text)).padding(5);
+        let mut peers = Grid::with_columns(5)
+            .push(header("id"))
+            .push(header("Socket"))
+            .push(header("Inbound"))
+            .push(header("User agent"))
+            .push(header("Version"));
+        for (peer_id, peer) in node_state.connected_peers.iter() {
+            let inbound_str = if peer.inbound { "Inbound" } else { "Outbound" };
+            peers = peers
```
alfiedotwtf, line 65: `peers = peers.push().push()...` doesn't look right?

node-gui/src/main_window/main_widget/tabs/wallet/addresses.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use crate::backend::messages::AccountInfo;
+
+use super::WalletMessage;
+
+pub fn view_addresses(account: &AccountInfo) -> Element<'static, WalletMessage> {
+    let field = |text: String| container(Text::new(text)).padding(5);
+    let mut addresses = Grid::with_columns(3);
+    for (index, address) in account.addresses.iter() {
+        addresses = addresses
```
alfiedotwtf, line 30: Same here `addresses = addresses.push()...`

node-gui/src/main_window/main_widget/tabs/wallet/stake.rs
```diff
@@ -0,0 +1,71 @@
...
+    account: &AccountInfo,
+    stake_amount: &str,
+) -> Element<'static, WalletMessage> {
+    let field = |text: String| container(Text::new(text)).padding(5);
+
+    let mut staking_balance_grid = Grid::with_columns(2)
+        .push(field("PoolId".to_owned()))
+        .push(field("Balance".to_owned()));
+    for (pool_id, balance) in account.staking_balance.iter() {
+        staking_balance_grid = staking_balance_grid
```
alfiedotwtf, line 39: Here and below too.

node-gui/src/main_window/main_widget/tabs/wallet/transactions.rs
```diff
@@ -0,0 +1,95 @@
...
+        let amount_str = tx
+            .tx_type
+            .amount()
+            .map(|amount| print_coin_amount(chain_config, amount))
+            .unwrap_or_default();
+        let timestamp = tx.timestamp.as_ref().map_or_else(
+            || "-".to_owned(),
+            |timestamp| print_block_timestamp(*timestamp),
+        );
+        transaction_list = transaction_list
```
alfiedotwtf, line 55: And here.

wallet/src/account/transaction_list/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+        .outputs()
+        .iter()
+        .filter(|output| own_output(key_chain, output))
+        .collect::<Vec<_>>();
+
+    let non_own_outputs = tx_data
+        .get_transaction()
+        .outputs()
+        .iter()
+        .filter(|output| !own_output(key_chain, output));
```
alfiedotwtf, line 167: I know it's an optimization, but should `own_outputs` and `non_own_outputs` be combined in a simple for loop so we're not looping over outputs twice (if it could be a large number of transactions)

alfiedotwtf (approved):

node-gui/src/main_window/main_widget/tabs/networking.rs
pavel-kokolemin: But this is the only API that Iced `grid` provides.
The official documentation/samples do as well:
```
let grid = Grid::<Message>::with_columns(2)
    .push(Text::new("First row, first column"))
    .push(Text::new("First row, second column"))
    .push(Text::new("Second row, first column"))
    .push(Text::new("Second row, second column"));
```

wallet/src/account/transaction_list/mod.rs
pavel-kokolemin: Right, thanks. I did not find out how to do it last time. I found it now with the help of ChatGPT.

PR #994, opened 2023-06-22
Skip first immediate peer stalling interval tick

discosultan:
The main loop of the block sync manager peer task periodically checks if a peer has stalled and should be disconnected. There is no value in immediately checking for that, therefore, the initial check can be skipped.

This also fixes an intermittent failure in `sync::tests::peer_events::do_not_disconnect_peer_after_receiving_known_header_list` that didn't expect for the stalling interval to trigger.


iljakuklic (approved):

muursh (approved):

pavel-kokolemin (approved):

TheQuantumPhysicist (approved):
I don't like special cases to solve instability problems because that tends to hide problems. Usually the right solution is to detect the state and react, instead or assuming the state.

However, I don't have the capacity to review this now, so I'm approving at least on temporary basis.

PR #995, opened 2023-06-22
Change VRF key derivation

### LINKED ISSUE #929: Fix derivation of VRF private keys
As a simplification, the VRF private keys are derived using normal private keys and hashing. We should fix this and use BIP-44 secrets to derive the VRF key.

OBorce:
- remove the temporary solution that was creating VRF keys from a secp256k1 key hash
- generate and store a root VRF key from the seed phrase
- derive the account VRF keys from the root VRF key and the derivation path as in BIP44

closes #929

### DIFF
--- a/crypto/src/util/mod.rs
+++ b/crypto/src/util/mod.rs
@@ -13,4 +13,66 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use generic_array::{sequence::Split, typenum::U32, GenericArray};
+use hmac::{Hmac, Mac};
+use sha2::Sha512;
+use zeroize::Zeroize;
+
+use crate::key::hdkd::{
+    chain_code::{ChainCode, CHAINCODE_LENGTH},
+    derivable::DerivationError,
+};
+
 pub mod eq;
+
+pub fn new_hmac_sha_512(key: &[u8]) -> Hmac<Sha512> {
+    Hmac::<Sha512>::new_from_slice(key).expect("HMAC can take key of any size")
+}
+
+pub fn to_key_and_chain_code<SecretKey>(
+    mac: Hmac<Sha512>,
+    to_key: impl FnOnce(&[u8]) -> Result<SecretKey, DerivationError>,
+) -> Result<(SecretKey, ChainCode), DerivationError> {
+    // Finalize the hmac
+    let mut result = mac.finalize().into_bytes();
+
+    // Split in to two 32 byte arrays
+    let (mut secret_key_bytes, mut chain_code_bytes): (
+        GenericArray<u8, U32>,
+        GenericArray<u8, U32>,
+    ) = result.split();
+    result.zeroize();
+
+    // Create the secret key key
+    let secret_key = to_key(secret_key_bytes.as_slice())?;
+    secret_key_bytes.zeroize();
+
+    // Chain code
+    let chain_code: [u8; CHAINCODE_LENGTH] = chain_code_bytes.into();
+    let chain_code = ChainCode::from(chain_code);
+    chain_code_bytes.zeroize();
+
+    Ok((secret_key, chain_code))
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::random::Rng;
+    use rstest::rstest;
+    use test_utils::random::{make_seedable_rng, Seed};
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    /// Test to prove that hmac-sha512 can be used with keys of any size, since there's an expect in it
+    fn mac_key_of_any_size(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+
+        let key_size = rng.gen_range(0..=1000);
+
+        let key = (0..key_size).map(|_| rng.gen::<u8>()).collect::<Vec<_>>();
+
+        let _hmac = new_hmac_sha_512(&key);
+    }
+}
--- a/crypto/src/vrf/mod.rs
+++ b/crypto/src/vrf/mod.rs
@@ -13,10 +13,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use generic_array::{sequence::Split, typenum::U32, GenericArray};
+use hmac::{Hmac, Mac};
 use merlin::Transcript;
 use serialization::{Decode, Encode};
+use sha2::Sha512;
+use zeroize::Zeroize;
 
-use crate::random::{make_true_rng, CryptoRng, Rng};
+use crate::{
+    key::hdkd::{
+        chain_code::{ChainCode, CHAINCODE_LENGTH},
+        child_number::ChildNumber,
+        derivable::{Derivable, DerivationError},
+        derivation_path::DerivationPath,
+    },
+    random::{make_true_rng, CryptoRng, Rng},
+};
 
 pub use self::primitives::VRFReturn;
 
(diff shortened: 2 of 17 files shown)


wallet/src/account/mod.rs
```diff
@@ -254,10 +252,15 @@ impl Account {
...
     ) -> WalletResult<(VRFPrivateKey, VRFPublicKey)> {
-        let vrf_key_path = make_path_to_vrf_key(&self.chain_config, self.account_index());
-        let private_key =
-            self.key_chain.get_private_key_for_path(&vrf_key_path, db_tx)?.private_key();
-        Ok(vrf_from_private_key(&private_key))
+        let (vrf_key_path, child_number) =
+            make_path_to_vrf_key(&self.chain_config, self.account_index());
+        let private_key = self.key_chain.get_private_key_for_path(&vrf_key_path, db_tx)?;
+        let (secret_key_bytes, _chain_code_bytes) =
+            private_key.to_hmac512_left_right_bytes(child_number);
```
TheQuantumPhysicist, line 259: This function is still using secp256k1 key as intermediary. That has to go.

OBorce: it is the parent private key

TheQuantumPhysicist: Still. There cannot be sec256k1 in the process. There seems to be a misunderstanding.

BIP-44 defines how we create a secret with length 512-bits. The left 256-bit should directly go into a `VRFPrivateKey::new_using_random_bytes()`. No secp256k1 should be involved. The secp256k1 is a temporary solution we made because we didn't have time to do the proper derivation.

muursh (approved):
Looks fine to me

crypto/src/vrf/mod.rs
```diff
@@ -151,6 +163,101 @@ impl VRFPublicKey {
...
+    pub fn new_master(
+        seed: &[u8],
+        keykind: VRFKeyKind,
+    ) -> Result<ExtendedVRFPrivateKey, DerivationError> {
+        // Create a new mac with the appropriate BIP39 constant
+        let mut mac = new_hmac_sha_512(b"Bitcoin seed");
+
+        mac.update(seed);
+
+        let (private_key, chain_code) = to_key_and_chain_code(mac, keykind)?;
```
iljakuklic, line 189: I know the point was to separate this from the normal key derivation. However, as a result there is some duplication, like this bit, the `new_hmac_sha_512` helper, etc. Is it possible/sensible to factor out the common parts into a shared module?

iljakuklic (approved):
Looks fine to me. Cannot make very strong comments about the crypto part though.

crypto/src/vrf/mod.rs
OBorce: moved the 2 functions , `new_hmac_sha_512` and `to_key_and_chain_code` to the util module

alfiedotwtf (approved):

crypto/src/util/mod.rs
```diff
@@ -13,4 +13,66 @@
...
+    derivable::DerivationError,
+};
+
 pub mod eq;
+
+pub fn new_hmac_sha_512(key: &[u8]) -> Hmac<Sha512> {
+    Hmac::<Sha512>::new_from_slice(key).expect("HMAC can take key of any size")
+}
+
+pub fn to_key_and_chain_code<SecretKey>(
```
TheQuantumPhysicist, line 32: If this is secp256k1 specific, why is it in the utils, instead of in a module there?

wallet/types/src/keys.rs
```diff
@@ -118,40 +119,13 @@ impl KeychainUsageState {
     }
 }
 
-/// The key id is described by it's public key
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
-pub struct RootKeyId(ExtendedPublicKey);
+#[derive(PartialEq, Eq, PartialOrd, Ord, Encode, Decode)]
+pub struct RootKeyConstant;
```
TheQuantumPhysicist, line 123: Please add documentation explaining why this exists and how it works

wallet/storage/src/internal/store_tx.rs
```diff
@@ -156,7 +156,7 @@ macro_rules! impl_read_ops {
                     .map(Iterator::collect)
             }
 
-            fn exactly_one_root_key(&self) -> crate::Result<bool> {
+            fn root_keys_exist(&self) -> crate::Result<bool> {
```
TheQuantumPhysicist, line 159: I would call this `check_root_keys_sanity`, and return a result. Also add documentation explaining what the condition of sanity is + error logs printing this error if it happens (at the caller).

crypto/src/util/mod.rs
OBorce: The `SecretKey` can be either secp256k1 or VRF key

TheQuantumPhysicist (approved):
This looks OK. Though please ask Pavel to take a look. Keep in mind that this will break all currently running nodes as derivation paths will change for VRF keys. So this will require relaunching the current testnet.

crypto/src/util/mod.rs
TheQuantumPhysicist: Alrighty.

wallet/types/src/keys.rs
OBorce: added a comment

crypto/src/vrf/mod.rs
```diff
@@ -151,6 +162,80 @@ impl VRFPublicKey {
...
+    private_key: VRFPrivateKey,
+}
+
+impl ExtendedVRFPrivateKey {
+    pub fn new_master(
+        seed: &[u8],
+        keykind: VRFKeyKind,
+    ) -> Result<ExtendedVRFPrivateKey, DerivationError> {
+        // Create a new mac with the appropriate BIP39 constant
+        let mut mac = new_hmac_sha_512(b"Bitcoin seed");
```
pavel-kokolemin, line 184: Maybe it would be better to separate the derivation starting from the root (using different arguments to `new_hmac_sha_512`)?
That way we can be a bit more confident that the keys will not interfere with each other. 

```
impl Secp256k1ExtendedPrivateKey {
    pub fn new_master(seed: &[u8]) -> Result<Secp256k1ExtendedPrivateKey, DerivationError> {
        // Create a new mac with the appropriate BIP39 constant
        let mut mac = new_hmac_sha_512(b"Bitcoin seed");

impl ExtendedVRFPrivateKey {
    pub fn new_master(
        seed: &[u8],
        keykind: VRFKeyKind,
    ) -> Result<ExtendedVRFPrivateKey, DerivationError> {
        // Create a new mac with the appropriate BIP39 constant
        let mut mac = new_hmac_sha_512(b"Bitcoin seed");
```

Probably no big deal, since VRFs use a different path at any case.

OBorce: Nice suggestion, this makes sense to me as with this even if the Secp256k1 master key is leaked the VRF master key and children will still be safe.
@TheQuantumPhysicist @muursh

TheQuantumPhysicist: I don't mind, sounds like a good idea. Though keep in mind that the secp256k1 derivation shouldn't change because it's compatible with hardware wallets, and I'm not sure how hard it'll be to change that. 

Let @muursh also provide his opinion.

wallet/storage/src/internal/store_tx.rs
```diff
@@ -158,13 +161,16 @@ macro_rules! impl_read_ops {
...
-            fn exactly_one_root_key(&self) -> crate::Result<bool> {
+            fn check_root_keys_sanity(&self) -> crate::Result<()> {
                 self.storage
                     .get::<db::DBRootKeys, _>()
                     .prefix_iter_decoded(&())
                     .map_err(crate::Error::from)
                     .map(Iterator::count)
-                    .map(|count| count == 1)
+                    .and_then(|count| {
+                        ensure!(count == 1, crate::Error::WalletWithoutARootKey);
```
TheQuantumPhysicist, line 171: Please change the error name to `WalletSanityErrorInvalidRootKeyCount(usize, usize)`, and in the text specify expected vs current.

Think of this from the user's perspective. If this error (ever) happens, it's of interest to understand what the problem is to be able to track it. In the future, we may add more sanity checks and they should have different errors/error messages.

TheQuantumPhysicist (approved):
Looking good. Please mind the comments before merging and decide on the master key derivation part.

wallet/storage/src/internal/store_tx.rs
OBorce: The count should not bring much to the error message as it should only be 0 or 1.
I will replace the count with a get in this case to make it clearer, that we are just checking if the root keys exist or not, if not then we have a wallet file without root keys which should not happen.

crypto/src/vrf/mod.rs
muursh: Yeah I don't see an issue with that

OBorce: secp256k1 remains the same as before, I only changed the constant for the VRF master key derivation to "Bitcoin VRF seed"

wallet/storage/src/internal/store_tx.rs
TheQuantumPhysicist: The current count at least helps because it can tell us whether there's a corruption in the wallet.

OBorce: The construction of the map itself restricts that, as the Key is an empty Struct which is unique and can't construct a map with more than 1 element.

TheQuantumPhysicist: That's why I said "corruption". If the error says there are 2710223923 keys, then we know the db is broken. The assumption that everything is bound to our code is very optimistic. So in database-related things, we try to be a little bit pessimistic to account for corruption.

PR #996, opened 2023-06-22
mempool: Transaction eviction

iljakuklic:
* Make mempool transaction dependencies more precise
* Evict transactions from mempool (and disconnect from its tx verifier) when mempool is full
* Transactions are removed in depth-first post-order starting from the evicted transaction to make sure they are disconnected in the correct order
* There's a (very inefficient) fallback mechanism in case transaction verifier and mempool still disagree about transaction dependencies. Ideally there would be single point of truth for this but does not seem to be easily achievable at the moment so left for future work.

### DIFF
--- a/mempool/src/config.rs
+++ b/mempool/src/config.rs
@@ -17,6 +17,32 @@ use std::time::Duration;
 
 pub type Time = Duration;
 
+/// Mempool size configuration
+#[derive(Debug, Eq, PartialEq, PartialOrd, Ord, Clone, Copy)]
+pub struct MempoolMaxSize(usize);
+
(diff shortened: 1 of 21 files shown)


azarovh (approved):

TheQuantumPhysicist (approved):
Looking good. Though can you please add functional tests for this? Trasnsaction eviction is relatively a complex operation.

muursh (approved):
This looks good to me.

iljakuklic:
Added:
* Interfaces to query and change mempool memory limit via API/RPC
* A functional test checking various scenarios, including some non-trivial ones

PR #997, opened 2023-06-23
Use a dedicated error when initializing net upgrades

discosultan:
Removes dependency to `anyhow` in the common library crate.


iljakuklic (approved):
Looks good and definitely an improvement over what we have now.

In the long run, in my view, these invariants should be enforced by construction. If, for example, the upgrades are represented by a function form block height to some secription of which consensus rules are in force at that height (instead of a vector of changes), we don't need to worry about missing entries.

muursh (approved):

azarovh (approved):

Issue #829, opened 2023-04-13
Optimize transaction requests


stanislav-tkach:
The transaction announcement can be received from multiple peers. Currently we send requests to each peer and it is somewhat wasteful. 

The `TxRequestTracker` class in the Bitcoin codebase can be used as an inspiration.

PR #1000, opened 2023-06-26
Allow only a single in-flight peer request per transaction

discosultan:
Closes https://github.com/mintlayer/mintlayer-core/issues/829

## Problem

Currently, when we receive the same transaction announcement from multiple peers, we request the transaction body from all of them. In most cases, this is unnecessary.

## Solution

Use block sync manager to synchronize transaction requests. Block sync manager acts a bit like a semaphore with a single permit. Competing peers are permitted in a FIFO order. Every transaction is synchronized separately.

If a peer successfully completes the transaction request, other waiting peers are cancelled. If a peer fails the transaction request (due to a timeout in receiving a response, for example) or gets disconnected, next waiting peer is permitted. If there are no more available peers for the transaction, the transaction body request is dropped.

## Questions / Considerations

- Should we allow requesting transaction body from more than a single peer at a time? For example, we could make this configurable. This would increase the chance of successfully receiving a transaction body early.

-----

Below are three different scenarios that have been tested. In all the scenarios two peers want to request for the same transaction body.

- Peer 1 succeeds, peer 2 is cancelled

image

- Peer 1 fails due to a timeout, peer 2 succeeds

image

- Peer 1 gets disconnected, peer 2 succeeds

image

### DIFF
--- a/p2p/src/sync/mod.rs
+++ b/p2p/src/sync/mod.rs
@@ -17,6 +17,7 @@
 //! to block announcement from peers and the announcement of blocks produced by this node).
 
 mod peer;
+mod request_tracker;
 mod types;
 
 use std::collections::HashMap;
@@ -47,7 +48,11 @@ use crate::{
         types::{services::Services, SyncingEvent},
         MessagingService, NetworkingService, SyncingEventReceiver,
     },
-    sync::peer::Peer,
+    sync::{
+        peer::Peer,
+        request_tracker::RequestTracker,
+        types::{PeerEvent, RequestTrackerEvent},
+    },
     types::peer_id::PeerId,
     PeerManagerEvent, Result,
 };
@@ -66,6 +71,11 @@ pub struct BlockSyncManager<T: NetworkingService> {
 
     /// A sender for the peer manager events.
     peer_manager_sender: UnboundedSender<PeerManagerEvent<T>>,
+
+    // Request tracker handle and sender channel.
+    _request_tracker_handle: JoinHandle<Never>,
+    request_tracker_sender: mpsc::UnboundedSender<RequestTrackerEvent>,
+    request_tracker_peer_event_sender: mpsc::UnboundedSender<PeerEvent>,
 
     chainstate_handle: subsystem::Handle<Box<dyn ChainstateInterface>>,
     mempool_handle: MempoolHandle,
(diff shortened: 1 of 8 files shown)


pavel-kokolemin:
Sorry, but why is `PeerManager` used to manage transaction loading? The job of `PeerManager` is to initiate/accept/drop peer connections. And its file is already too big without new functionality (@TheQuantumPhysicist said we should try to keep files under 500 lines).

p2p/src/peer_manager/mod.rs
```diff
@@ -144,6 +144,9 @@ where
     subscribed_to_peer_addresses: BTreeSet<PeerId>,
 
     peer_eviction_random_state: peers_eviction::RandomState,
+
+    // Map of transactions for which we need to request a body.
+    pending_transaction_requests: Mutex<BTreeMap<Id<Transaction>, OneshotQueue<PeerId>>>,
```
pavel-kokolemin, line 149: Why is `Mutex` used here? `PeerManager` is single-threaded.

discosultan: Good catch! Removed. I originally played around with a different pattern that warranted a mutex where we would spawn transaction trackers. However, I decided to drop the idea due to complexity + it probably scales worse.

discosultan:
> > Is PeerManager the right place to do such synchronization?
> 
> Sorry, I think it is not. The job of `PeerManager` is to initiate/accept/drop peer connections. And its file is already too big without new functionality (@TheQuantumPhysicist said we should try to keep files under 500 lines).

I was under the same impression that this shouldn't be the responsibility of a peer manager. However, it was the easiest to implement in because the control events happen there. What do you recon would make more sense? Move this to `BlockSyncManager` or have a dedicated `RequestTracker` that the `BlockSyncManager` could pass into spawned `Peer`s?

pavel-kokolemin:
> Move this to BlockSyncManager or have a new dedicated RequestTracker that the BlockSyncManager could pass into spawned Peers?

I think both are valid options.

> However, it was the easiest to implement in because the control events happen there.

I think we only interested in disconnected peer events? `BlockSyncManager` receive that too.

p2p/src/net/types/mod.rs
```diff
@@ -156,3 +156,17 @@ pub enum SyncingEvent {
     /// Peer disconnected
     Disconnected { peer_id: PeerId },
 }
+
+impl SyncingEvent {
+    pub fn to_p2p_event(&self) -> P2pEvent {
```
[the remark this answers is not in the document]
discosultan: I didn't like this approach and changed it.

p2p/src/sync/mod.rs
```diff
@@ -201,15 +227,19 @@ where
...
+    fn handle_peer_event(&mut self, event: SyncingEvent) -> Result<()> {
         match event {
             SyncingEvent::Connected {
                 peer_id,
                 services,
                 sync_rx,
             } => self.register_peer(peer_id, services, sync_rx),
-            SyncingEvent::Disconnected { peer_id } => self.unregister_peer(peer_id),
+            SyncingEvent::Disconnected { peer_id } => {
+                self.request_tracker_peer_event_sender.send(PeerEvent::Disconnected(peer_id))?;
```
discosultan, line 238: Ideally, I didn't want the block sync manager to have to care about forwarding peer lifecycle events to the events tracker. An alternative would be if during request tracker creation it can subscribe to the peer events sent by backend.

TheQuantumPhysicist:
We'll tackle this problem later.

PR #1006, opened 2023-06-28
Removing logging for a common occurrence

alfiedotwtf:
When there is a new tip within chainstate, `JobManager` will stop all jobs that will be out of date within `handle_chainstate_event()`. Any currently running thread that's still in `process_block()` will notice this event and exit. However, to make sure that `JobManager` always cleans up regardless, there is a `OnceDestructor` to stop the job within `JobManager` which will duplicate the call.

I don't believe by removing this logging event will mask an issue.

### DIFF
--- a/blockprod/src/detail/job_manager/jobs_container.rs
+++ b/blockprod/src/detail/job_manager/jobs_container.rs
@@ -76,10 +76,7 @@ impl JobsContainer {
     /// Returns true if the job was removed, false if it was not found.
     fn remove_job(&mut self, job_key: JobKey, and_stop: bool) -> bool {
         match self.jobs.entry(job_key) {
-            Entry::Vacant(j) => {
-                log::error!("Attempted to stop non-existent job: {j:?}");
-                false
-            }
+            Entry::Vacant(_) => false,
             Entry::Occupied(entry) => {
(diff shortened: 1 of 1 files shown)


blockprod/src/detail/job_manager/jobs_container.rs
```diff
@@ -76,10 +76,7 @@ impl JobsContainer {
     /// Returns true if the job was removed, false if it was not found.
     fn remove_job(&mut self, job_key: JobKey, and_stop: bool) -> bool {
         match self.jobs.entry(job_key) {
-            Entry::Vacant(j) => {
-                log::error!("Attempted to stop non-existent job: {j:?}");
-                false
-            }
+            Entry::Vacant(_) => false,
```
iljakuklic, line 79: Could you please elaborate on the rationale behind this change? Will it not just mask some sort of job management issue?

alfiedotwtf: No problem. I've updated the description.

TheQuantumPhysicist (approved):
I don't fully understand the context, but I'm fine with this.

iljakuklic (approved):

PR #1008, opened 2023-06-28
mempool: Detect some conflicting orphan transactions

iljakuklic:
Detect if an orphan transaction conflicts with a transaction already in mempool and reject it if it does.

### DIFF
--- a/mempool/src/pool/mod.rs
+++ b/mempool/src/pool/mod.rs
@@ -56,7 +56,7 @@ use self::{
 };
 use crate::{
     config,
-    error::{Error, MempoolPolicyError, OrphanPoolError, TxValidationError},
+    error::{Error, MempoolConflictError, MempoolPolicyError, OrphanPoolError, TxValidationError},
     tx_accumulator::TransactionAccumulator,
     MempoolEvent, TxStatus,
 };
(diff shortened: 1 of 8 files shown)


discosultan (approved):
Looks good to me, but I wouldn't put too much weight on my approval since I don't have much experience with mempool.

mempool/src/pool/mod.rs
```diff
@@ -531,19 +533,33 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
...
     }
 
+    fn orphan_rbf_checks(&self, tx: &TxEntry) -> Result<(), OrphanPoolError> {
+        let mut conflicts = self.conflicting_tx_ids(tx).peekable();
+        if conflicts.peek().is_none() {
+            // Early exit if there are no conflicts
+            return Ok(());
+        }
+
+        if ENABLE_RBF {
```
TheQuantumPhysicist, line 543: I would create an issue and reference this or at least throw in a TODO about this.

Also, while I completely understand the motivation and how this is done, it's possible that the first call in this block will panic and we'll never know if it's valid code after one commit in the future (theoretically speaking). Hence, I'd emphasize in a comment that the code below is virtually pseudo-code that shouldn't be accepted without complete revision once RBF is enabled.

TheQuantumPhysicist (approved):

mempool/src/pool/mod.rs
iljakuklic: I added the #1012 issue about re-enabling RBF. It mentions that bits of code guarded by `ENABLE_RBF` are stale and untested.

muursh (approved):

PR #1010, opened 2023-06-28
Fix for ChainstateRef::get_block_height_in_main_chain

ImplOfAnImpl:
`ChainstateRef::get_block_height_in_main_chain` was returning `Some(height)` even for blocks not on the mainchain; because of this, `ChainstateInterface::is_block_in_main_chain` worked incorrectly, returning true for all blocks.

What's interesting, `ChainstateInterface::is_block_in_main_chain` and `ChainstateRef::is_block_in_main_chain` have separate implementations; the latter also used (and still does) `get_block_height_in_main_chain` but it did it in a strange way, always rechecking its result by calling `get_block_id_by_height`. I simplified it.

One piece of code did rely in the wrong behaviour of `get_block_height_in_main_chain` - it's `check_height_order` in `chainstate/test-suite/src/tests/bootstrap.rs`. I fixed it.

Finally, `chainstate/test-suite/src/tests/reorgs_tests.rs` had its own implementation of `is_block_in_main_chain`, I removed it.

### DIFF
--- a/chainstate/src/interface/chainstate_interface_impl.rs
+++ b/chainstate/src/interface/chainstate_interface_impl.rs
@@ -91,14 +91,11 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> ChainstateInterfa
     }
 
     fn is_block_in_main_chain(&self, block_id: &Id<Block>) -> Result<bool, ChainstateError> {
-        // FIXME: should it just delegate to ChainstateRef::is_block_in_main_chain?
-        // (but it'll have to be added to ChainstateQuery too).
         self.chainstate
             .query()
             .map_err(ChainstateError::from)?
-            .get_block_height_in_main_chain(&(*block_id).into())
+            .is_block_in_main_chain(&(*block_id).into())
             .map_err(ChainstateError::FailedToReadProperty)
-            .map(|ht| ht.is_some())
     }
 
     fn get_block_height_in_main_chain(
--- a/chainstate/src/interface/chainstate_interface_impl.rs
+++ b/chainstate/src/interface/chainstate_interface_impl.rs
@@ -91,6 +91,8 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> ChainstateInterfa
     }
 
     fn is_block_in_main_chain(&self, block_id: &Id<Block>) -> Result<bool, ChainstateError> {
+        // FIXME: should it just delegate to ChainstateRef::is_block_in_main_chain?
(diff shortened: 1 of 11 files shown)


chainstate/src/interface/chainstate_interface_impl.rs
```diff
@@ -91,6 +91,8 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> ChainstateInterfa
     }
 
     fn is_block_in_main_chain(&self, block_id: &Id<Block>) -> Result<bool, ChainstateError> {
+        // FIXME: should it just delegate to ChainstateRef::is_block_in_main_chain?
```
TheQuantumPhysicist, line 94: I don't see why not. It's always preferable to centralize these calls for consistency.

TheQuantumPhysicist (approved):

chainstate/src/interface/chainstate_interface_impl.rs
```diff
@@ -94,9 +94,8 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> ChainstateInterfa
         self.chainstate
             .query()
             .map_err(ChainstateError::from)?
-            .get_block_height_in_main_chain(&(*block_id).into())
+            .is_block_in_main_chain(&(*block_id).into())
             .map_err(ChainstateError::FailedToReadProperty)
```
ImplOfAnImpl, line 98: Note that `ChainstateRef::is_block_in_main_chain` takes `&Id<GenBlock>`, but this one takes `&Id<Block>`, which is a bit strange. I wasn't sure what to use in `ChainstateQuery::is_block_in_main_chain` but decided to go with `Id<GenBlock>` in the end.
It would be nice to unify their interfaces though. Should I do it? Which kind of block id would be more logical in this case?

TheQuantumPhysicist: Sure... though Genesis is always part of main chain, but it's convenient to just be able to submit a id of genblock.

iljakuklic (approved):

chainstate/src/interface/chainstate_interface_impl.rs
ImplOfAnImpl: Ok, I've changed `ChainstateInterface::is_block_in_main_chain`, so that it uses `&Id<GenBlock>` too.

iljakuklic (approved):

alfiedotwtf (approved):

PR #1011, opened 2023-06-28
Add support for lock then transfer in the Wallet

OBorce:
Fix Wallet output cache to properly filter out locked outputs.

### DIFF
--- a/wallet/src/account/output_cache/mod.rs
+++ b/wallet/src/account/output_cache/mod.rs
@@ -15,11 +15,18 @@
 
 use std::collections::{BTreeMap, BTreeSet};
 
-use common::chain::{
-    tokens::{token_id, TokenId},
-    TxInput, TxOutput, UtxoOutPoint,
+use common::{
+    chain::{
+        timelock::OutputTimeLock,
+        tokens::{token_id, TokenId},
+        TxInput, TxOutput, UtxoOutPoint,
+    },
+    primitives::BlockDistance,
 };
-use wallet_types::{AccountWalletTxId, WalletTx};
+use wallet_types::{
+    wallet_tx::{BlockInfo, TxState},
+    AccountWalletTxId, WalletTx,
+};
 
 /// A helper structure for the UTXO search.
 ///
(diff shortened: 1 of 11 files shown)


wallet/src/wallet/mod.rs
```diff
@@ -129,11 +131,13 @@ impl<B: storage::Backend> Wallet<B> {
 
         db_tx.commit()?;
 
+        let latest_median_time = chain_config.genesis_block().timestamp();
```
alfiedotwtf, line 134: Maybe I'm misunderstanding, but should this be getting the newest rather than the oldest timestamp?

alfiedotwtf (approved):

wallet/src/wallet/mod.rs
OBorce: In this case I think it is all good, as it is a new wallet, only knows the genesis block, and it will get updated when it syncs from the node.

But I guess I need to update the one below when it is loading an existing wallet, in that case it can show a wrong balance in the short time between loading the wallet and the first sync with a node.

TheQuantumPhysicist (approved):
This looks fine. Though there should be a way to check locked balance.

Also, would be great if others can review the PR.

wallet/src/account/output_cache/mod.rs
```diff
@@ -91,17 +98,44 @@ impl OutputCache {
...
-    fn valid_utxo(&self, outpoint: &UtxoOutPoint) -> bool {
+    fn valid_utxo(
+        &self,
+        outpoint: &UtxoOutPoint,
+        output: &TxOutput,
+        transaction_block_info: &BlockInfo,
+        current_block_info: &BlockInfo,
+    ) -> bool {
         !self.consumed.contains(outpoint)
+            && output.timelock().map_or(true, |timelock| match timelock {
```
azarovh, line 109: Have we already considered reusing checks from tx_verifier? Not for this PR, but in general we might want to open an issue for that to look later.

azarovh (approved):

muursh (approved):

Issue #1004, opened 2023-06-28
Add PoS blockprod functional tests


alfiedotwtf:
Currently, all of the Python functional tests in `test/functional` are PoW-based.

This fix will add PoS tests to the Python functional tests.

- [x] https://github.com/mintlayer/mintlayer-core/pull/1016
- [x] https://github.com/mintlayer/mintlayer-core/pull/1021
- [x] https://github.com/mintlayer/mintlayer-core/pull/1039
- [x] https://github.com/mintlayer/mintlayer-core/pull/1040
- [x] #1092 

This is part of #1013

alfiedotwtf:
I originally had a task to create command line parameters to set genesis staking keys, VRF keys, and pool Id, but after looking at the config code again, I think it will be better to add getter API calls to `test-rpc-functions/` RPC.

PR #1016, opened 2023-06-30
Update RegTest Genesis to include PoS pools using fixed staking and VRF keys

alfiedotwtf:
These RegTest genesis values will be needed for the Python functional tests for PoS.

This is part of #1004

### DIFF
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -570,7 +570,7 @@ pub fn regtest_genesis_values() -> (
     )
 }
 
-pub fn create_regtest_pos_genesis() -> Genesis {
+pub fn create_regtest_pos_genesis(premine_destination: Destination) -> Genesis {
     let (
         genesis_pool_id,
         genesis_stake_pool_data,
@@ -583,10 +583,15 @@ pub fn create_regtest_pos_genesis() -> Genesis {
     let create_genesis_pool_txoutput =
         TxOutput::CreateStakePool(genesis_pool_id, genesis_stake_pool_data);
 
+    let premine_output = TxOutput::Transfer(
+        OutputValue::Coin(Amount::from_atoms(100_000_000_000_000_000_000)),
+        premine_destination,
+    );
+
     Genesis::new(
         String::new(),
         BlockTimestamp::from_int_seconds(1639975460),
-        vec![create_genesis_pool_txoutput],
+        vec![premine_output, create_genesis_pool_txoutput],
     )
 }
 
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -29,13 +29,14 @@ use crate::chain::tokens::OutputValue;
 use crate::chain::transaction::Destination;
 use crate::chain::upgrades::NetUpgrades;
 use crate::chain::TxOutput;
-use crate::chain::{GenBlock, Genesis};
+use crate::chain::{GenBlock, Genesis, PoolId};
 use crate::chain::{PoWChainConfig, UpgradeVersion};
 use crate::primitives::id::{Id, Idable, WithId};
 use crate::primitives::per_thousand::PerThousand;
 use crate::primitives::semver::SemVer;
 use crate::primitives::{Amount, BlockDistance, BlockHeight, H256};
 use crypto::key::hdkd::{child_number::ChildNumber, u31::U31};
+use crypto::{key::PrivateKey, vrf::VRFPrivateKey};
 use std::num::NonZeroU64;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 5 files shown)


chainstate/test-suite/src/tests/chainstate_accounting_storage_tests.rs
```diff
@@ -40,6 +42,27 @@ use rstest::rstest;
...
+    let (
+        genesis_pool_id,
+        genesis_stake_pool_data,
+        _genesis_stake_private_key,
+        genesis_stake_public_key,
+        _genesis_vrf_private_key,
+        genesis_vrf_public_key,
+    ) = get_regtest_genesis_value();
+
+    let genesis_pool_data = PoolData::new(
```
azarovh, line 55: `From` is implemented for this type so you can do `let genesis_pool_data = genesis_stake_pool_data.into()`

chainstate/test-suite/src/tests/chainstate_accounting_storage_tests.rs
```diff
@@ -154,9 +177,17 @@ fn store_pool_data_and_balance(#[case] seed: Seed) {
             1
         );
 
+        let (genesis_pool_id, genesis_pool_data) = get_genesis_pool_data();
+
         let expected_tip_storage_data = pos_accounting::PoSAccountingData {
```
azarovh, line 182: I'd create the same data for sealed storage to check that only genesis info is there. Right now the check is weak. In other tests as well

azarovh (approved):

azarovh:
After another look I'm not sure that adding a pool to regtest genesis is a good idea. It complicates simple tests that doesn't care about PoS and also requires hardcoded keys and so on.

Maybe better way is to use TestFramework to set up genesis if required. There are some helper functions that could be used from `chainstate/test-framework/src/utils.rs` like `create_chain_config_with_staking_pool`. I think it would be more flexible.

TheQuantumPhysicist:
> After another look I'm not sure that adding a pool to regtest genesis is a good idea. It complicates simple tests that don't care about PoS and also requires hardcoded keys and so on.
> 
> Maybe better way is to use TestFramework to set up genesis if required. There are some helper functions that could be used from `chainstate/test-framework/src/utils.rs` like `create_chain_config_with_staking_pool`. I think it would be more flexible.

The issue is that we need to be able to stake in regression tests. How else can this be done?

azarovh:
> How else can this be done?

My point is to leave genesis as is by default. If necessary any genesis can be initialised via `TestFramework`. For PoS tests there is already `create_chain_config_with_staking_pool` function which can be used to implement another function with default values.

Right now the genesis info is hardcoded and incapsulated inside genesis initialisation. In order to get the staking key every test calls `get_genesis_pool_data` which recreates keys just to be able to access them. This is kind of hacky way.

 Btw I believe keys can be generated with rng every time instead of hardcoding them.

alfiedotwtf:
> There are some helper functions that could be used from chainstate/test-framework/src/utils.rs like create_chain_config_with_staking_pool. I think it would be more flexible.

Ok thanks, I didn't know that existed. I'll take a look at it.

> The issue is that we need to be able to stake in regression tests. How else can this be done?

Because I spent the time looking at how all the argument parsing works yesterday, I think @azarovh is right - this change probably shouldn't touch everything else just to get what I need for the PoS Python functional tests...

Instead, what I think I can do is move some of this code, or use `create_chain_config_with_stasking_pool()`, into `regtest_chain_config()` like #1021 - this way, the only RegTest tests that has set the command line flag actually need modifying.

alfiedotwtf:
Now that `RegTest` Genesis will be modified within `regtest_chain_config()`, it now makes sense to build this off #1021.

alfiedotwtf:
> After another look I'm not sure that adding a pool to regtest genesis is a good idea. It complicates simple tests that don't care about PoS and also requires hardcoded keys and so on.

So I've now gated adding the pool to RegTest genesis only when specified on the command line in `regtest_chain_config()`.

> Maybe better way is to use TestFramework

I looked at this, but if I used this, then `regtest_chain_config()` would miss the Regtest-specific command line arguments (i.e `RegtestOptions` rather than `RunOptions`).

> Btw I believe keys can be generated with rng every time instead of hardcoding them.

Specifically for `RegTest`, we need these to be hardcoded so that the Python functional tests know what values to use. In the future we can add API accessors to `test-rpc-functions/` to get the randomly generated values, or even set these from the command line. But we have other priorities now.

common/src/chain/config/mod.rs
```diff
@@ -525,6 +526,70 @@ fn create_testnet_genesis() -> Genesis {
...
+        _genesis_vrf_public_key,
+    ) = regtest_genesis_values();
+
+    let create_genesis_pool_txoutput =
+        TxOutput::CreateStakePool(genesis_pool_id, genesis_stake_pool_data);
+
+    Genesis::new(
+        String::new(),
+        BlockTimestamp::from_int_seconds(1639975460),
+        vec![create_genesis_pool_txoutput],
```
azarovh, line 589: I'd still create some coins because it might be useful for delegation

common/src/chain/pos.rs
```diff
@@ -129,6 +129,18 @@ pub fn create_unittest_pos_config() -> PoSChainConfig {
     }
 }
 
+pub fn create_regtest_pos_config() -> PoSChainConfig {
```
azarovh, line 132: Ok stupid questions time becuase now I'm confused. Why do we need both unittest and regtest configs? Do we suggest functional tests == regression tests? And how it then relates to `ChainType`?

azarovh (approved):

common/src/chain/config/mod.rs
alfiedotwtf: Ah, good point. Thanks will do.

common/src/chain/pos.rs
alfiedotwtf: I've been treating the Python tests as the "regression tests".
 
The reason for the added `create_regtest_pos_config()` function is because this would have been a good place to set `difficulty_change_limit` to zero from the timeout PR. And as it was needed anyway, I put it in this PR.

As for the difference in `ChainType`, as far as I can tell where it's needed the most, is in `node-lib/src/runner.rs` which parses the command line options differently than the other chain types. This lets us have different `ChainConfig` settings that can be set from Python but shouldn't be set from the other chain types.

TheQuantumPhysicist: > Do we suggest functional tests == regression tests? 

Yes, RegTest is meant to be functional tests. But there has been lots of abuse for all kinds of chain-types. It's not a big deal as it can be fixed later.

TheQuantumPhysicist (approved):

PR #1017, opened 2023-07-01
Clock sync fixes

pavel-kokolemin:
There are some problems if the host clock is wrong.
If the host clock is ahead and staking is started, then once new block is found by this node and it is propagated between peers, they will reject the block and ban your node.
If the host clock is behind, once a new block is found by the network this node will reject it and ban all peers that propagated the block.
Even 1 second difference can cause this all these problems.

@TheQuantumPhysicist  suggested setting the `BlockFromTheFuture` ban score to 0. I tried this, but it did not work. Because once the first block is rejected by the chainstate, the block synchronization process is broken and later blocks are rejected with the `PrevBlockNotFound` error.
I think we should require the host clock to be in sync.

I did a little research and it looks like some other blockchains require this as well:
From https://geth.ethereum.org/docs/fundamentals/peer-to-peer
```
Local time might be incorrect. An accurate clock is required to participate in the Ethereum network. The local clock can be resynchronized using commands such as sudo ntpdate -s time.nist.gov (this will vary depending on operating system).
```
And from https://iohk.zendesk.com/hc/en-us/articles/360010230873-Machine-clock-out-of-sync-with-Cardano-network
```
Daedalus is displaying a red screen message related to your machine time being out of sync. 
```
So I think this is acceptable, and I don't see how else we can make it work.

Requiring high clock accuracy (less than 1 second) is probably too strong, so we can allow some tolerance. I chose 10 seconds (the new `max_clock_diff` parameter in `P2pConfig`). Making it too large can result in a poor performance, and making it too small can make it too difficult for the users to join the network.
To check the clock, I added the new `current_time` field to p2p handshakes. If two peers disagree on the current time (the difference is more than `MaxClockDiff`), the handshake will fail.

As I mentioned above, even a 1 second difference can cause issues.
To prevent this, the `BlockSyncManager` peer tasks will delay processing new block messages until they can be accepted by the chainstate (but not more than `max_clock_diff`).

I may be missing something, so feedback is welcome.


node-lib/src/config_files/p2p.rs
```diff
@@ -69,6 +69,9 @@ pub struct P2pConfigFile {
     pub ban_threshold: Option<u32>,
     /// Duration of bans in seconds.
     pub ban_duration: Option<u64>,
+    /// Maximum acceptable time difference between this node and the remote peer.
+    /// If a large difference is detected, the peer will be disconnected.
+    pub max_clock_diff: Option<u64>,
```
discosultan, line 74: Since we are not using a `Duration` here, I would specify in the doc comment that this value is in seconds (similar to how it's documented for `ban_duration`).

node-lib/src/options.rs
```diff
@@ -123,6 +123,11 @@ pub struct RunOptions {
     #[clap(long)]
     pub p2p_sync_stalling_timeout: Option<NonZeroU64>,
 
+    /// Maximum acceptable time difference between this node and the remote peer.
+    /// If a large difference is detected, the peer will be disconnected.
+    #[clap(long)]
+    pub p2p_max_clock_diff: Option<u64>,
```
discosultan, line 129: I'd specify in doc comments that this value is in seconds.

p2p/src/sync/peer.rs
```diff
@@ -203,12 +209,36 @@ where
...
+    /// because one side may see the new block as invalid.
+    async fn sync_clock(&self, block_timestamp: BlockTimestamp) {
+        let block_timestamp = block_timestamp.as_duration_since_epoch();
+        let max_block_timestamp =
+            self.time_getter.get_time() + *self.chain_config.max_future_block_time_offset();
+        if block_timestamp <= max_block_timestamp {
+            return;
+        }
+        let clock_diff = max_block_timestamp - block_timestamp;
+        let sleep_time = std::cmp::max(clock_diff, *self.p2p_config.max_clock_diff);
```
discosultan, line 224: Should this be `std::cmp::min` instead? According to the doc comment:
> Delays the processing of a new block ... (but not more than `max_clock_diff`).
In which case if the `clock_diff` is larger than `max_clock_diff`, we should sleep `max_clock_diff` instead of `clock_diff`.

p2p/src/sync/peer.rs
```diff
@@ -203,12 +209,36 @@ where
...
+
     async fn handle_message(&mut self, message: SyncMessage) -> Result<()> {
         let res = match message {
             SyncMessage::HeaderListRequest(r) => self.handle_header_request(r.into_locator()).await,
             SyncMessage::BlockListRequest(r) => self.handle_block_request(r.into_block_ids()).await,
-            SyncMessage::HeaderList(l) => self.handle_header_list(l.into_headers()).await,
-            SyncMessage::BlockResponse(r) => self.handle_block_response(r.into_block()).await,
+            SyncMessage::HeaderList(l) => {
+                if let Some(header) = l.headers().last() {
+                    self.sync_clock(header.timestamp()).await;
```
discosultan, line 234: I think it would be better if the clock sync would be performed inside `handle_header_list`. There are various checks there where we could skip the processing altogether. It doesn't make sense to sleep in those scenarios.

p2p/src/config.rs
```diff
@@ -92,6 +93,9 @@ pub struct P2pConfig {
     pub ping_check_period: PingCheckPeriod,
     /// When a peer is detected as dead and disconnected
     pub ping_timeout: PingTimeout,
+    /// Maximum acceptable time difference between this node and the remote peer.
+    /// If a large difference is detected, the peer will be disconnected.
+    pub max_clock_diff: MaxClockDiff,
```
alfiedotwtf, line 98: Like the other PR comments, add "in seconds"

p2p/src/sync/mod.rs
```diff
@@ -162,16 +163,23 @@ where
...
+        let prev_task = self.peers.insert(peer_id, peer_task);
+        assert!(prev_task.is_none(), "Registered duplicated peer: {peer_id}");
     }
 
     /// Stops the task of the given peer by closing the corresponding channel.
-    fn unregister_peer(&mut self, peer: PeerId) {
-        log::debug!("Unregister peer {peer} from sync manager");
-        let removed = self.peers.remove(&peer);
-        assert!(removed, "Unregistering unknown peer: {peer}");
+    fn unregister_peer(&mut self, peer_id: PeerId) {
```
alfiedotwtf, line 175: Are these `register_peer()`/`unregister_peer()` changes supposed to be included in this PR?

p2p/src/sync/peer.rs
```diff
@@ -203,12 +209,36 @@ where
         Ok(())
     }
 
+    /// Delays the processing of a new block until it can be accepted by the chainstate (but not more than `max_clock_diff`).
+    /// This is needed to allow the local or remote node to have slightly inaccurate clocks.
+    /// Without it, even a 1 second difference can break block synchronization
+    /// because one side may see the new block as invalid.
+    async fn sync_clock(&self, block_timestamp: BlockTimestamp) {
```
alfiedotwtf, line 216: Should this be called something like `delay_new_block()` instead, because `sync_clock()` sounds like a time somewhere is getting set

alfiedotwtf (approved):

p2p/src/config.rs
pavel-kokolemin: I don't think it's needed here because `MaxClockDiff` is stored internally as `Duration`.

p2p/src/sync/mod.rs
pavel-kokolemin: It's needed to abort async tasks if they are sleeping.

p2p/src/sync/peer.rs
pavel-kokolemin: Nice catch, thanks!

p2p/src/sync/peer.rs
pavel-kokolemin: I moved the check to `handle_header_list` and removed the check from `handle_block_response` (because blocks are downloaded after the header list and the node should be already in sync).

> There are various checks there where we could skip the processing altogether. It doesn't make sense to sleep in those scenarios.

Why bother with such optimizations? The nodes are expected to have correct clocks. And if the remote peer doesn't, it's their fault.

p2p/src/sync/peer.rs
pavel-kokolemin: The `sync_clock` function was removed.

pavel-kokolemin:
There was a question about security implications in the presence of malicious nodes. This new approach should be fine as long as the clock is correct. But if the node's clock is wrong, it will be trivial to isolate the node by connecting malicious nodes with the same invalid clock (while honest nodes can't connect). After that, the node will be open to all kinds of attacks.

p2p/src/sync/peer.rs
discosultan: I was also thinking if certain checks fail, maybe we want to increase the peer's ban score or disconnect them. I'd expect we'd want to take such actions as quickly as possible instead of potentially sleeping first due to clock diff.

discosultan (approved):

p2p/src/sync/peer.rs
```diff
@@ -317,6 +320,26 @@ where
...
         log::debug!("Headers list from peer {}", self.id());
 
+        if let Some(last_header) = headers.last() {
+            // Delays the processing of a new block until it can be accepted by the chainstate (but not more than `max_clock_diff`).
+            // This is needed to allow the local or remote node to have slightly inaccurate clocks.
+            // Without it, even a 1 second difference can break block synchronization
+            // because one side may see the new block as invalid.
+            let block_timestamp = last_header.timestamp().as_duration_since_epoch();
+            let max_block_timestamp =
+                self.time_getter.get_time() + *self.chain_config.max_future_block_time_offset();
```
TheQuantumPhysicist, line 330: There's also the tolerance from the chainstate... blocks are allowed to be off by that much (which is 120 seconds now).

Please use checked_add when adding that into account instead of direct addition that can panic.

TheQuantumPhysicist:
What happens if the clock drifts over time? I've seen that computer clocks (that don't automatically synchronize) drift by around 10 seconds per week.

This topic seems shaky the more I think about it. We need a campaign to study this problem separately.

I'm not saying this to reject this PR, btw. Just random thoughts on the topic.

p2p/src/net/default_backend/peer.rs
```diff
@@ -114,22 +114,39 @@ where
         }
     }
 
-    async fn handshake(&mut self) -> crate::Result<()> {
+    fn validate_peer_time(
+        p2p_config: &P2pConfig,
+        local_time: BlockTimestamp,
```
TheQuantumPhysicist, line 119: BlockTimestamp shouldn't be used for this. It's somewhat an abuse of the type system. Perhaps use Duration?

p2p/src/net/default_backend/types.rs
```diff
@@ -86,6 +86,8 @@ pub enum HandshakeMessage {
         /// Socket address of the remote peer as seen by this node (addr_you in bitcoin)
         receiver_address: Option<PeerAddress>,
 
+        current_time: BlockTimestamp,
```
TheQuantumPhysicist, line 89: Please use another type.

p2p/src/net/default_backend/types.rs
```diff
@@ -98,6 +100,8 @@ pub enum HandshakeMessage {
 
         /// Socket address of the remote peer as seen by this node (addr_you in bitcoin)
         receiver_address: Option<PeerAddress>,
+
+        current_time: BlockTimestamp,
```
TheQuantumPhysicist, line 104: Please use another type.

TheQuantumPhysicist (approved):
While this is OK as a start, it's by no means sufficient. There are many considerations to be added, which include accounting for drifts. We need to study this issue more deeply in a separate campaign. I'll create an issue for it.

p2p/src/sync/peer.rs
```diff
@@ -317,6 +320,26 @@ where
     async fn handle_header_list(&mut self, headers: Vec<SignedBlockHeader>) -> Result<()> {
         log::debug!("Headers list from peer {}", self.id());
 
+        if let Some(last_header) = headers.last() {
```
TheQuantumPhysicist, line 323: Please move this block to a separate function, let's call it `wait_for_clock_diff`.

p2p/src/sync/peer.rs
pavel-kokolemin: > There's also the tolerance from the chainstate... blocks are allowed to be off by that much (which is 120 seconds now).

Yes, sure, but I already use it here (`max_future_block_time_offset` equals to 120 seconds).

> Please use checked_add when adding that into account instead of direct addition that can panic.

I think it's OK, both `time_getter` and `max_future_block_time_offset` can be trusted.

p2p/src/sync/peer.rs
pavel-kokolemin: Sure, but let's please do this as a next step.

PR #1018, opened 2023-07-01
Increase peer ban score when receive a header with invalid parent

discosultan:


p2p/src/sync/tests/ban_scores.rs
```diff
@@ -76,3 +83,59 @@ async fn peer_handle_result() {
...
+    let mut tf = TestFramework::builder(&mut rng).build();
+
+    // Create and process an invalid block. It is invalid because an included transaction does not
+    // have any inputs and outputs. We deliberately ignore the processing error.
+    let invalid_tx = TransactionBuilder::new().build();
+    let invalid_parent_block = tf.make_block_builder().with_transactions(vec![invalid_tx]).build();
+    let invalid_parent_block_id = invalid_parent_block.get_id();
+    tf.process_block(invalid_parent_block, BlockSource::Local).unwrap_err();
+
+    // Create a valid block linked to the invalid parent block. Do not process it. Its' header will
```
ImplOfAnImpl, line 102: `Its'` -> `Its`

ImplOfAnImpl (commented):
Do I understand it correctly that all the changes in `chainstate` are there only to be able to check that the p2p test works properly?
In any case, those changes have the same problems that my now abandoned PR #1001 had. They should be fixed in PR #1019, so probably you should wait until it lands.

Regarding the test itself - LGTM.

discosultan:
> Do I understand it correctly that all the changes in chainstate are there only to be able to check that the p2p test works properly?
> In any case, those changes have the same problems that my now abandoned PR https://github.com/mintlayer/mintlayer-core/pull/1001 had. They should be fixed in PR https://github.com/mintlayer/mintlayer-core/pull/1019, so probably you should wait until it lands.

Correct, I just made those changes quickly to get the test passing. Planned to discuss this topic with you on Monday.
I think once we merge your PR, it's sufficient for me just to leave the p2p test. Don't think any other changes are necessary.

p2p/src/sync/tests/ban_scores.rs
discosultan: Thanks, fixed.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -512,6 +512,18 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+
+        if let GenBlockId::Block(prev_block_id) = header.prev_block_id().classify(self.chain_config)
+        {
+            let prev_block_index = self.get_block_index(&prev_block_id)?;
+            if let Some(prev_block_index) = prev_block_index {
+                ensure!(
+                    prev_block_index.status().is_valid(),
+                    CheckBlockError::InvalidParent(header.block_id())
+                );
+            }
```
iljakuklic, line 524: This check passes if the parent index is not found. Is that intentional? I guess the block ends up among orphans in that case. Is the valid parent check only performed once block is out of the orphan block set? I am not quite clear on the check sequence logic and the invariants upheld at each stage.

discosultan: Any changes to chainstate made here are superseded by the changes in https://github.com/mintlayer/mintlayer-core/pull/1019. I've updated the PR to indicate that. There's only a test left in the p2p now. We may not even want that because essentially we're testing chainstate. However, there may be some value in testing p2p -> chainstate integration.

TheQuantumPhysicist (approved):

PR #1021, opened 2023-07-03
Include PoS NetUpgrades in RegTest when specified from the command line

alfiedotwtf:
This will be needed for #1004

### DIFF
--- a/node-lib/src/runner.rs
+++ b/node-lib/src/runner.rs
@@ -31,10 +31,9 @@ use chainstate::rpc::ChainstateRpcServer;
 use common::{
     chain::{
         config::{Builder as ChainConfigBuilder, ChainConfig, ChainType, EmissionScheduleTabular},
-        create_testnet_pos_config, initial_difficulty, ConsensusUpgrade, NetUpgrades,
-        UpgradeVersion,
+        NetUpgrades,
     },
-    primitives::{semver::SemVer, BlockHeight},
+    primitives::semver::SemVer,
 };
 use logging::log;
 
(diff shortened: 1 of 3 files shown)


node-lib/src/runner.rs
```diff
@@ -319,6 +322,27 @@ fn regtest_chain_config(options: &ChainConfigOptions) -> Result<ChainConfig> {
...
+        let pos_net_upgrades = vec![
+            (
+                BlockHeight::new(0),
+                UpgradeVersion::ConsensusUpgrade(ConsensusUpgrade::IgnoreConsensus),
+            ),
+            (
+                BlockHeight::new(1),
+                UpgradeVersion::ConsensusUpgrade(ConsensusUpgrade::PoS {
+                    initial_difficulty: initial_difficulty(ChainType::Regtest).into(),
+                    config: create_testnet_pos_config(),
```
azarovh, line 335: seems like `create_unittest_pos_config` should be used here

azarovh (approved):

node-lib/src/runner.rs
```diff
@@ -319,6 +322,27 @@ fn regtest_chain_config(options: &ChainConfigOptions) -> Result<ChainConfig> {
     update_builder!(max_block_size_with_standard_txs);
     update_builder!(max_block_size_with_smart_contracts);
 
+    if chain_pos_netupgrades.unwrap_or(false) {
+        let pos_net_upgrades = vec![
```
azarovh, line 326: btw there is `NetUpgrades::unit_test` function. Maybe it's better to create another one `unit_test_with_pos` because I see this code in lots of places

iljakuklic (approved):
I am fine with this in principle. But agree with @azarovh that some of the code duplication should be addressed and it should also be relatively easy to replicate the behaviour of the command line switch using a simple method call (on a builder or something like that) so we can get the same setting in unit tests too.

node-lib/src/runner.rs
alfiedotwtf: Nice.

alfiedotwtf:
> But agree with @azarovh that some of the code duplication should be addressed

Yep, agreed. Fixed.

> it should also be relatively easy to replicate the behaviour of the command line switch using a simple method call (on a builder or something like that) so we can get the same setting in unit tests too.

To make `pos-netupgrades` an accessor to `Builder` I would need to add it to `ChainConfigBuilder`, which would then be seen to all chain types - something that should probably be avoided given we only want this flag in `RegTest`. As for getting the same setting under unit tests, `ChainConfigOptions` has all fields as `pub` so if need be it can be accessed there.

TheQuantumPhysicist (approved):
Unify the impl blocks and rename to regtest. Otherwise this looks good.

PR #1022, opened 2023-07-03
Consume self when running a manager

discosultan:
Small refactoring that I wanted to do separately from feature PRs to reduce noise.

Consume `self` when running a manager object. It doesn't make sense to be able to run such objects multiple times. This also aligns with how a subsystem is run.

### DIFF
--- a/p2p/src/net/default_backend/mod.rs
+++ b/p2p/src/net/default_backend/mod.rs
@@ -129,22 +129,19 @@ impl<T: TransportSocket> NetworkingService for DefaultNetworkingService<T> {
         let socket = transport.bind(bind_addresses).await?;
         let local_addresses = socket.local_addresses().expect("to have bind address available");
 
-        let p2p_config_ = Arc::clone(&p2p_config);
-        let shutdown_ = Arc::clone(&shutdown);
+        let backend = backend::Backend::<T>::new(
+            transport,
+            socket,
+            chain_config,
+            Arc::clone(&p2p_config),
+            cmd_rx,
+            conn_tx,
+            sync_tx,
+            Arc::clone(&shutdown),
+            shutdown_receiver,
+            subscribers_receiver,
+        );
         let backend_task = tokio::spawn(async move {
-            let mut backend = backend::Backend::<T>::new(
-                transport,
-                socket,
-                chain_config,
-                p2p_config_,
-                cmd_rx,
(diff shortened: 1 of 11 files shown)


p2p/src/net/default_backend/mod.rs
```diff
@@ -129,22 +129,19 @@ impl<T: TransportSocket> NetworkingService for DefaultNetworkingService<T> {
...
-                cmd_rx,
-                conn_tx,
-                sync_tx,
-                shutdown_,
-                shutdown_receiver,
-                subscribers_receiver,
-            );
-
             match backend.run().await {
                 Ok(_) => unreachable!(),
```
iljakuklic, line 146: Not quite related to this PR but I just noticed. Since `run` returns `Result<Never>`, the `unreachable` can be avoided by using `Ok(never) => never.into()`. (Annotations may be needed to resolve type ambiguity, I did not try to actually compile it)

iljakuklic (approved):

p2p/src/net/default_backend/mod.rs
iljakuklic: Looks like there are two more cases of this also in `p2p/src/lib.rs`

iljakuklic: Hmm looking at it again, it does not seem the required `impl` is available in this case. You may need to explicitly match `Ok(never) => match never {}`

alfiedotwtf (approved):

p2p/src/net/default_backend/mod.rs
discosultan: What's the benefit of this instead of unreachable panic? Is it so that in case the return type changes from never, we wouldn't forget to update?

discosultan: https://github.com/mintlayer/mintlayer-core/pull/1022/commits/4aae0f85281f54873df1f31af9a53bde70150ce5

iljakuklic: > What's the benefit of this instead of unreachable panic? Is it so that in case the return type changes from never, we wouldn't forget to update?

Using `match never {}` formally shows that this is an unreachable branch, and the assertion is verified by the compiler. With `unreachable!`, the verification burden is on us.

pavel-kokolemin (approved):

PR #1023, opened 2023-07-04
Ignore timing out unit test

alfiedotwtf:
One of the unit tests from #977 sometimes fails after merging to master.

In the worst case, locally I'm getting 67s for 1000 PoS blocks. This is more than we want in CI, so I've reduced the test down to worse case 100 PoS blocks in a row - locally under 5s.

### DIFF
--- a/blockprod/src/detail/tests.rs
+++ b/blockprod/src/detail/tests.rs
@@ -1068,7 +1068,7 @@ mod produce_block {
                     shutdown_trigger.initiate();
                 });
 
-                let block_production = BlockProduction::new(
+                let mut block_production = BlockProduction::new(
                     chain_config.clone(),
                     chainstate.clone(),
                     mempool,
@@ -1111,7 +1111,10 @@ mod produce_block {
 
                             job_finished_receiver.await.expect("Job finished receiver closed");
 
-                            assert_job_count(&block_production, 0).await;
+                            // TODO: until duplicate job keys is fixed
+                            // (Issue #1003), manually stop all jobs
+                            _ = block_production.stop_all_jobs().await;
+
                             assert_process_block(&chainstate, new_block.clone()).await;
                         }
                         RequiredConsensus::PoS(_) => {
@@ -1134,9 +1137,8 @@ mod produce_block {
                             }
 
                             // TODO: until duplicate job keys is fixed
-                            // in produce_block(), manually wait until
-                            // the job manager has cleaned up
-                            assert_job_count(&block_production, 0).await;
+                            // (Issue #1003), manually stop all jobs
+                            _ = block_production.stop_all_jobs().await;
 
                             // Try PoW input data for PoS consensus
 
@@ -1154,9 +1156,8 @@ mod produce_block {
                             }
 
                             // TODO: until duplicate job keys is fixed
-                            // in produce_block(), manually wait until
-                            // the job manager has cleaned up
-                            assert_job_count(&block_production, 0).await;
+                            // (Issue #1003), manually stop all jobs
+                            _ = block_production.stop_all_jobs().await;
 
                             // Try PoS input data for PoS consensus
 
@@ -1167,7 +1168,10 @@ mod produce_block {
 
                             job_finished_receiver.await.expect("Job finished receiver closed");
 
-                            assert_job_count(&block_production, 0).await;
+                            // TODO: until duplicate job keys is fixed
+                            // (Issue #1003), manually stop all jobs
+                            _ = block_production.stop_all_jobs().await;
+
                             let result = assert_process_block(&chainstate, new_block).await;
 
                             // Update kernel input parameters for future PoS blocks
@@ -1204,9 +1208,8 @@ mod produce_block {
                             }
 
                             // TODO: until duplicate job keys is fixed
-                            // in produce_block(), manually wait until
-                            // the job manager has cleaned up
-                            assert_job_count(&block_production, 0).await;
+                            // (Issue #1003), manually stop all jobs
+                            _ = block_production.stop_all_jobs().await;
 
                             // Try PoS input data for PoW consensus
 
@@ -1224,9 +1227,8 @@ mod produce_block {
                             }
 
                             // TODO: until duplicate job keys is fixed
-                            // in produce_block(), manually wait until
-                            // the job manager has cleaned up
-                            assert_job_count(&block_production, 0).await;
+                            // (Issue #1003), manually stop all jobs
+                            _ = block_production.stop_all_jobs().await;
 
                             // Try PoW input data for PoW consensus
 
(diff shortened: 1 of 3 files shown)


discosultan (approved):

iljakuklic (approved):

blockprod/src/detail/tests.rs
```diff
@@ -1000,7 +1000,7 @@ mod produce_block {
             )
         };
 
-        let blocks_to_generate = rng.gen_range(100..=1000);
+        let blocks_to_generate = rng.gen_range(20..=100);
```
TheQuantumPhysicist, line 1003: If this is taking a long time due to difficulty issues, it's probably a good idea to limit the difficulty so that it doesn't take a long time, instead of setting these arbitrary numbers that have no inherent guarantees. In practice the generation should have a time bound (statistically speaking).

alfiedotwtf: I think we're already limiting difficulty by using `create_unittest_pos_config()` - it sets `target_limit` to `Uint256::MAX`.

Ah, I guess I can set a timeout for the test itself, but would we want to automatically treat that as a pass if it times out? I don't think that's a good idea because we may never see it fail then

TheQuantumPhysicist: > I think we're already limiting difficulty by using `create_unittest_pos_config()` - it sets `target_limit` to `Uint256::MAX`.

That's probably the upper limit for the target (hence limit on the easiest possible difficulty). You need the opposite limitation. Please double-check.

> Ah, I guess I can set a timeout for the test itself, but would we want to automatically treat that as a pass if it times out? I don't think that's a good idea because we may never see it fail then

No. That's not what I meant. I meant limiting the difficulty by never allowing it to be too hard.

alfiedotwtf: Yep, you were right - the docs say `target_limit` is the lowest difficulty bound meaning the hardest, so I was interpreting this incorrectly.

If this is the case, then I'm thinking the following line in `consensus/src/pos/target.rs` in `calculate_new_target()` should be changed to `max()`:

```
    let new_target = std::cmp::min(new_target, pos_config.target_limit());
```

... maybe @azarovh can have a look at `calculate_new_target()` to see if that's correct.

Either way, I've made a local change which sets `difficulty_change_limit` to 0. This essentially fixes `new_target` difficulty during each calculation to the easiest PoS difficulty. So for PoS on my machine, I'm getting around 20 blocks per second, which gets us close to the 1 minute timeout for a unit test when we try and generate 1000 blocks.

I know it's arbitrary, think if I keep `let blocks_to_generate = rng.gen_range(20..=100);` and also fix `difficulty_change_limit`, then this particular unit test should never hit the timeout even in the worst case while still exercising the various consensus `NetUpgrades`.

azarovh: > If this is the case, then I'm thinking the following line in consensus/src/pos/target.rs in calculate_new_target() should be changed to max():

Changing this to `max` doesn't seem right because it will always be true. 
Keeping `difficulty_change_limit` at 0 is more reasonable fix. It actually says that for tests we don't want the difficulty to change at all.

I'll take a look if I can see other options to fix that for tests.

azarovh: At this point I'm not sure that there is a particular problem with this test or retargeting algo. If running with the same Seed changing `difficulty_change_limit` doesnt effect result at all.
Also I tried disabling PoS consensus upgrade or PoW. And the test is just slow anyway.

I have a profiler result if you are interested in looking deeper. 
flamegraph

alfiedotwtf: Thanks for that. So for each of the consensus types, I try to generate the blocks using the other types before I create the real one. However at the moment the JobManager takes a custom Id, but it's it's `None`, it will just use that. But before I can call `produce_block()` again, I have to wait for the JobManager to clean up.

I'm thinking that at the moment it's not cleaning up as fast as I thought... completely removing `Ignore` and `PoW`, then as @TheQuantumPhysicist said, `PoS` flies - 1000 `PoS` blocks in 10 seconds.

... so the difficulty was another red herring I think.

alfiedotwtf:
In the large test, we're now stopping all jobs rather than wait for them to go away by themselves. It's now completing in around 10 seconds rather than the > 60 seconds as before.

azarovh (approved):

TheQuantumPhysicist (approved):

muursh (approved):

blockprod/src/detail/tests.rs
```diff
@@ -1019,14 +1018,25 @@ mod produce_block {
                 vec![kernel_input_utxo.clone()],
             );
 
+            let easy_pos_config = PoSChainConfig::new(
```
azarovh, line 1021: isn't it the same as `create_unittest_pos_config`? Difficulty is already at MAX by default. No sure that I understand why it's necessary

alfiedotwtf: The last value is different.

azarovh: oh right I confused it with another PerThousand value. But anyway does it make a difference? Could you give me a seed where you see it so I can check as well?

alfiedotwtf: I don't have any tests that I recorded, but I did have that test run over 60s while debugging

azarovh: still not sure that this change is necessary. I'd fix other problem and then look if retargeting code has an issues.

alfiedotwtf: Now that it's no longer failing with a Chainstate-less JobManager from `2f8f466`, I've fixed it to get the worst case:

```
modified   blockprod/src/detail/tests.rs                                                            
@@ -999,7 +999,7 @@ mod produce_block {                                                             
             )                                                                                      
         };                                                                                         
                                                                                                    
-        let blocks_to_generate = rng.gen_range(100..=1000);                                        
+        let blocks_to_generate = 1000;                                                             
                                                                                                    
         let override_chain_config = {                                                              
             let genesis_block = Genesis::new(                                                      
@@ -1048,14 +1048,14 @@ mod produce_block {                                                         
             let mut next_height_consensus_change = 1;                                              
                                                                                                    
             while next_height_consensus_change < blocks_to_generate {                              
-                let next_consensus_type = rng.gen_range(0..consensus_types.len());                 
+                let next_consensus_type = 2;                                                       
                                                                                                    
                 randomized_net_upgrades.push((                                                     
                     BlockHeight::new(next_height_consensus_change),                                
                     UpgradeVersion::ConsensusUpgrade(consensus_types[next_consensus_type].clone()),
                 ));                                                                                
                                                                                                    
-                next_height_consensus_change += rng.gen_range(1..50);                              
+                next_height_consensus_change += 1000;                                              
             }                                                                                      
``` 

- IgnoreConsensus: ~1s
- PoW: ~1s
- PoS: ~10s

However, if I change that last `PosChainConfig` value from `PerThousand::new(1)` back to `PerThousand::new(100)` like in `create_unittest_pos_config()`, I get the following:

- IgnoreConsensus: ~1s
- PoW: ~1s
- PoS: ~160s

We could revert it back to `PerThousand::new(100)`, but every now and again we might land on the worst case scenario and it will break CI.

Once last experiment - I just tried bumping it a bit higher to `PerThousand::new(10)`, and I'm getting:

- IgnoreConsensus: ~1s
- PoW: ~1s
- PoS: ~13s

azarovh: interesting, ok let's merge it and I'll look into it

alfiedotwtf: Yep cool. Just waiting on @TheQuantumPhysicist to discuss the Chainstate-less commit.

TheQuantumPhysicist (approved):

PR #1025, opened 2023-07-04
Type networking service join handle and add p2p event topic to it

discosultan:
This PR is an attempt to improve the solution for https://github.com/mintlayer/mintlayer-core/pull/1000.

### Problem

Various components within the p2p subsystem need to react to peer lifecycle events (i.e connected, disconnected). There are currently two event types for this:
1. A `P2pEvent` which is currently only used to allow components _outside_ p2p to subscribe to peer events.
2. A `SyncingEvent` which is used specifically between the network backend and block sync manager.

It would be nice if additional components within the p2p had easy access to either of these events.

### Solution

Since `SyncingEvent` is a fairly complicated type which cannot be trivially cloned (it contains an `mpsc::Receiver`), I decided to make the `P2pEvent` also available internally within p2p, albeit in a slightly different manner. For external use, the p2p events are broadcasted into an `EventsController` which allows external components to subscribe to the events by passing a callback. For internal use, we prefer using a `broadcast::Receiver` which can be used within `tokio::select!`. Therefore, an additional p2p events topic has been made available as part of the backend handle that allows subscribing to the events through a tokio broadcast channel.


pavel-kokolemin:
Are you sure about this? I don't like the way `broadcast` works. Messages can be lost if you subscribe after the node has started, or if the `broadcast` queue is full (`RecvError::Lagged` error). This can lead to bugs that are hard to reproduce.

discosultan:
> Are you sure about this? I don't like the way broadcast works. Messages can be lost if you subscribe after the node has started, or if the broadcast queue is full (RecvError::Lagged error). This can lead to bugs that are hard to reproduce.

Thanks for the feedback @pavel-kokolemin !

There would have a be certain pattern in place when subscribing to a broadcast channel to avoid losing messages. What I had in mind is the following: currently we first initialize a manager object with a `new` static function and then run the manager object in a Tokio task with the `run` method. We would initialize all manager objects before running any of them. Subscription to a broadcast channel would happen in the initialization phase. This way we can ensure that no messages will be sent before all subscriptions are in place.

The lagging with multiple subscribers is a potential issue but I don't know how relevant it is in this context.

Note that I already have a working code for request tracker which doesn't use the broadcast channel. We can also just go with that and skip the changes introduced in this PR. Here's what I was hoping to improve with this PR: https://github.com/mintlayer/mintlayer-core/pull/1000/files#r1252244388

p2p/src/net/mod.rs
```diff
@@ -88,12 +89,35 @@ pub trait NetworkingService {
...
+            Self::SyncingEventReceiver,
+        >,
+    >;
+}
+
+#[derive(Debug)]
+pub struct NetworkingServiceJoinHandle<ConnectivityHandle, MessagingHandle, SyncingEventReceiver>
+where
+    SyncingEventReceiver: Send,
+    MessagingHandle: Send + Sync + Clone,
```
TheQuantumPhysicist, line 105: I don't think Clone makes sense here. It's not needed.

p2p/src/net/default_backend/backend.rs
```diff
@@ -115,6 +115,9 @@ pub struct Backend<T: TransportSocket> {
     /// TX channel for sending syncing events
     sync_tx: mpsc::UnboundedSender<SyncingEvent>,
 
+    /// TX channel for broadcasting p2p events
+    p2p_tx: broadcast::Sender<P2pEvent>,
```
TheQuantumPhysicist, line 119: I'm not really a big fan of this. This is because P2pEvent is exclusively meant for external subsystems. P2pEvents have no order requirements for components within p2p, for example, and hence can be delayed. Despite this being just an enum, it's made for P2pEventHandler, not to be thrown around. This is a leak in the opposite direction (to the inside, instead of to the outside).

If you need this information for other p2p components (e.g., clients connected/disconnected), then perhaps you should create a separate abstraction for that, which would be relevant for individual components within p2p, not as a general P2pEvent abstraction for both inside and outside.

p2p/src/net/mod.rs
TheQuantumPhysicist: Btw, you didn't have to create another PR to remove this Clone. I only meant that in this struct Clone doesn't make sense. But anyway, I'm OK with that either way.

TheQuantumPhysicist:
I think this PR should be closed.

discosultan:
Closing as the `P2pEvent` should be exclusively used by external subsystems.

PR #1031, opened 2023-07-06
mempool: Propagate transactions once fully validated

iljakuklic:
Once an orphan transaction has been fully processed and moved outside of the orphan pool, we take action according to the outcome:
* If the transaction is valid, we pass it on to peers as appropriate.
* If the transaction is not valid, we punish the originator.

### DIFF
--- a/mempool/src/interface/mempool_interface.rs
+++ b/mempool/src/interface/mempool_interface.rs
@@ -14,7 +14,8 @@
 // limitations under the License.
 
 use crate::{
-    error::Error, tx_accumulator::TransactionAccumulator, MempoolEvent, MempoolMaxSize, TxStatus,
+    error::Error, event::MempoolEvent, tx_accumulator::TransactionAccumulator, MempoolMaxSize,
+    TxOrigin, TxStatus,
 };
 use common::{
     chain::{GenBlock, SignedTransaction, Transaction},
@@ -25,7 +26,11 @@ use subsystem::{CallRequest, ShutdownRequest};
 
 pub trait MempoolInterface: Send + Sync {
     /// Add a transaction to mempool
-    fn add_transaction(&mut self, tx: SignedTransaction) -> Result<TxStatus, Error>;
+    fn add_transaction(
+        &mut self,
+        tx: SignedTransaction,
+        origin: TxOrigin,
+    ) -> Result<TxStatus, Error>;
 
     /// Get all transactions from mempool
     fn get_all(&self) -> Vec<SignedTransaction>;
@@ -52,10 +57,7 @@ pub trait MempoolInterface: Send + Sync {
     ) -> Result<Box<dyn TransactionAccumulator>, Error>;
 
     /// Subscribe to events emitted by mempool
-    fn subscribe_to_events(
-        &mut self,
-        handler: Arc<dyn Fn(MempoolEvent) + Send + Sync>,
-    ) -> Result<(), Error>;
+    fn subscribe_to_events(&mut self, handler: Arc<dyn Fn(MempoolEvent) + Send + Sync>);
 
     /// Get current memory usage
     fn memory_usage(&self) -> usize;
--- a/mempool/src/tx_origin.rs
+++ b/mempool/src/tx_origin.rs
@@ -1,0 +1,52 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use p2p_types::peer_id::PeerId;
+
+/// Tracks where a transaction originates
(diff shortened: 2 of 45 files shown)


p2p/src/sync/tests/tx_announcement.rs
```diff
@@ -353,6 +356,97 @@ async fn valid_transaction(#[case] seed: Seed) {
...
+    for _ in 0..2 {
+        let (_peer, msg) = handle.message().await;
+        logging::log::error!("Msg new: {msg:?}");
+        let tx_id = match msg {
+            SyncMessage::NewTransaction(tx_id) => tx_id,
+            msg => panic!("Unexpected message {msg:?}"),
+        };
+
+        let _expected_tx = txs.remove(&tx_id).expect("An existing transaction");
+    }
```
iljakuklic, line 442: This part of the test does not work. It's supposed to check two transactions announcements are issued. However, it hangs on the second one. The functionality this is supposed to check works: it's also covered by the functional test. Therefore, I suspect it's the test that is not set up correctly but not quite sure what's wrong with it exactly. Any ideas?

mempool/src/interface/mempool_interface.rs
```diff
@@ -25,7 +26,11 @@ use subsystem::{CallRequest, ShutdownRequest};
 
 pub trait MempoolInterface: Send + Sync {
     /// Add a transaction to mempool
-    fn add_transaction(&mut self, tx: SignedTransaction) -> Result<TxStatus, Error>;
+    fn add_transaction(
+        &mut self,
+        tx: SignedTransaction,
+        origin: TxOrigin,
```
TheQuantumPhysicist, line 32: I would call it "Source", since we also have BlockSource. Just for consistency.

iljakuklic: There already is `TransactionSource` in the tx verifier, it's used for something else. It's probably better to avoid unifying the two although there are some similarities.

TheQuantumPhysicist: I see.

mempool/src/tx_origin.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Transaction was submitted via local node's RPC subsystem. It should be propagated if valid.
+    LocalP2p,
+
+    /// Transaction was in a block but moved into the mempool upon a reorg.
+    PastBlock,
+
+    /// Transaction was received from a peer.
+    ///
+    /// If it eventually turns out to be valid, it should be propagated further to other peers.
+    /// If it's not valid, the original peer should be penalized as appropriate.
```
TheQuantumPhysicist, line 33: I just want to say that there are interesting implications to wanting to use the peer id for extended periods of time. I think currently in the p2p system, if a peer connects and then disconnects and then connects again, they will get a new peer id. The peer id is not associated with an entity, it's just an atomic counter for connections. So this can be easily circumvented. 

I believe the peer id is the same in bitcoin core. It doesn't have any identity associated with it.

There are multiple solutions to this problem, including creating a special type in p2p that's associated with a peer, truly.

wallet/wallet-node-client/src/handles_client/mod.rs
```diff
@@ -147,7 +147,11 @@ impl NodeInterface for WalletHandlesClient {
     }
 
     async fn submit_transaction(&self, tx: SignedTransaction) -> Result<(), Self::Error> {
-        self.p2p.call_async_mut(move |this| this.submit_transaction(tx)).await??;
+        // TODO: Should this return mempool::TxStatus?
```
TheQuantumPhysicist, line 150: This should return the low-level result of the call, whether TxStatus or otherwise.

The wallet controller then should interpret that result and inform the user about it in the appropriate manner, which may be GUI or CLI.

TheQuantumPhysicist (approved):
This PR looks fine. My only contention, however, is that using peer id as a universal id won't work as intended. This just occurred to me now while studying the code. Please see the related comment in the code in the relevant enum. 

While we can merge this as the only problematic part is the type PeerId, which can be changed easily,we certainly can't forget that this problem exists. At least a TODO/issue should be created for this and should be studied. Once a solution exists, PeerId should be made local again to p2p and the new type should be used.

wallet/wallet-node-client/src/handles_client/mod.rs
iljakuklic: > This should return the low-level result of the call, whether TxStatus or otherwise.

That is currently not that completely trivial to achieve without pulling some heavy dependencies into the wallet interface. The most straightforwad way to achieve it is to split out the `TxStatus` type out of mempool (like introducing `mempool-types` or such). That looks like a job for a separate PR.

TheQuantumPhysicist: I see. Though `mempool::TxStatus` is quite a simple type. If we need a mempool-types, I don't see a problem with that. You can make it in another PR on top of this one. Keep in mind that without that, there's no way for the wallet to know what happened with the transactions.

mempool/src/tx_origin.rs
iljakuklic: > I believe the peer id is the same in bitcoin core. It doesn't have any identity associated with it.

I just checked and indeed it is.

discosultan (approved):

pavel-kokolemin (approved):

p2p/src/sync/mod.rs
```diff
@@ -134,6 +137,11 @@ where
                     self.handle_new_tip(block_id).await?;
                 },
 
+                orphan_proc = orphan_tx_processed_receiver.recv() => {
```
pavel-kokolemin, line 140: The name "orphan" is no longer accurate, is it?

PR #1032, opened 2023-07-06
Create a controller function that returns pool ids

OBorce:
Add new functionality to the wallet to list pool ids

### DIFF
--- a/wallet/src/account/output_cache/mod.rs
+++ b/wallet/src/account/output_cache/mod.rs
@@ -45,6 +45,7 @@ pub struct OutputCache {
     txs: BTreeMap<OutPointSourceId, WalletTx>,
     consumed: BTreeMap<UtxoOutPoint, TxState>,
     unconfirmed_descendants: BTreeMap<OutPointSourceId, BTreeSet<OutPointSourceId>>,
+    pools: BTreeMap<PoolId, (UtxoOutPoint, BlockInfo)>,
 }
 
 impl OutputCache {
@@ -53,6 +54,7 @@ impl OutputCache {
             txs: BTreeMap::new(),
             consumed: BTreeMap::new(),
             unconfirmed_descendants: BTreeMap::new(),
+            pools: BTreeMap::new(),
         }
     }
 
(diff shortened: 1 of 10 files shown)


wallet/src/account/output_cache/mod.rs
```diff
@@ -68,6 +68,22 @@ impl OutputCache {
         &self.consumed
     }
 
+    pub fn pool_ids(&self) -> BTreeSet<PoolId> {
+        self.txs
```
TheQuantumPhysicist, line 72: Perhaps this can be optimized later by making a specific key/value store for staking pools. A wallet growing to thousands of transactions can easily make this process unnecessarily complex and hard.

wallet/wallet-controller/src/lib.rs
```diff
@@ -200,6 +200,10 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
             .map_err(ControllerError::WalletError)
     }
 
+    pub fn get_pool_ids(&self, account_index: U31) -> Result<BTreeSet<PoolId>, ControllerError<T>> {
+        self.wallet.get_pool_ids(account_index).map_err(ControllerError::WalletError)
```
TheQuantumPhysicist, line 204: We probably should list more information about this, including when the pool was created, etc. Pool Ids alone are just meaningless gibberish.

TheQuantumPhysicist (approved):
Please mind the minor things I mentioned. Otherwise, looks good.

wallet/wallet-controller/src/lib.rs
OBorce: Added block height and timestamp of the pool creation and the current balance retrieved from the node.

wallet/src/account/output_cache/mod.rs
OBorce: added a map for the pool ids to creating block heigh and timestamp

wallet/src/account/output_cache/mod.rs
```diff
@@ -42,13 +42,15 @@ use wallet_types::{
 pub struct OutputCache {
     txs: BTreeMap<AccountWalletTxId, WalletTx>,
     consumed: BTreeSet<UtxoOutPoint>,
+    pools: BTreeMap<PoolId, (UtxoOutPoint, BlockInfo)>,
```
iljakuklic, line 45: Would it make sense to turn the tuple into a proper struct? It could define some convenience methods to access the components of `BlockInfo`.

iljakuklic (approved):
Looks good

wallet/src/account/output_cache/mod.rs
OBorce: As the only use is for printing, I will wrap it in a struct at the CLI level with a custom Display for a better message to the user.

azarovh (approved):

wallet/wallet-controller/src/lib.rs
```diff
@@ -200,6 +203,34 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static> Controller<T> {
...
+    async fn get_pool_data(
+        &self,
+        pool_id: PoolId,
+        block_info: BlockInfo,
+    ) -> Result<(PoolId, BlockInfo, Amount), ControllerError<T>> {
+        self.rpc_client
+            .get_stake_pool_balance(pool_id)
+            .await
+            .map_err(ControllerError::NodeCallError)
+            .and_then(|balance| balance.ok_or(ControllerError::SyncError("".into())))
```
OBorce, line 215: need to log the error as pool id found in wallet but not in node and skip it

PR #1035, opened 2023-07-07
Initial implementation of block status as a set of bit fields

ImplOfAnImpl:


TheQuantumPhysicist:
This is unnecessary since the other one seems fine.

PR #1036, opened 2023-07-07
Simplified implementation of block status as a set of bit fields

ImplOfAnImpl:


TheQuantumPhysicist (approved):
I think this looks fine. Simple enough and does the job. It can also be ensured to be backwards compatible by proper unit testing.

My only request is to add local tests for this, in the sense where individual effects of the struct in isolation are tested, in addition to tests that fail if someone attempts to change bytes randomly.

PR #1038, opened 2023-07-11
Formatting change example

ImplOfAnImpl:
This is an example of how different the code formatting would be if we forced rustfmt to use the same width limit of 100 chars for all lines.
There are 3 commits here, which can be examined independently:
1) [Formatting changes after setting use_small_heuristics = "Max"](https://github.com/mintlayer/mintlayer-core/commit/b940cfdd05eb84b6659b1240487d9e39ba0f9bce); this forces full line width for all lines, except for those that correspond to the settings `array_width`, `chain_width` and `single_line_if_else_max_width`, which are currently explicitly set in `rustfmt.toml`.
2) [Also, array_width = 80 and chain_width = 80 were removed](https://github.com/mintlayer/mintlayer-core/commit/9c7f98e937fc860410e38a05bf690abc549270f6); here `array_width` and `chain_width` are removed, so the corresponding lines are also full width.
3) [Also, single_line_if_else_max_width = 50 was removed](https://github.com/mintlayer/mintlayer-core/commit/776cabe45f725a5fbd2b306f60151fa5f9ee315a) - single_line_if_else_max_width was removed.

Some notes:
1) I don't suggest to merge these changes or even start a discussion about applying any formatting changes right now. It's just something to think about and hopefully discuss at some point in the future.
2) The reason why IMO we should consider this is to reduce vertical space waste, which should improve code readability.
3) Sam expressed a concern about such formatting changes in general - they spoil the output of "git blame", making it harder to navigate the history. One solution to this is to create a file ".git-blame-ignore-revs", which would contain a list of revisions to be ignored by "blame", as explained [in this article](https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/). GutHub is supposed to know about this file. The problem though is that everyone who will want to use this feature locally will have to add a configuration line to the local `.git/config` file, which is not very convenient.

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -98,10 +98,7 @@ impl BlockProduction {
     }
 
     pub async fn stop_all_jobs(&mut self) -> Result<usize, BlockProductionError> {
-        self.job_manager_handle
-            .stop_all_jobs()
-            .await
-            .map_err(BlockProductionError::JobManagerError)
+        self.job_manager_handle.stop_all_jobs().await.map_err(BlockProductionError::JobManagerError)
     }
 
     pub async fn stop_job(&mut self, job_key: JobKey) -> Result<bool, BlockProductionError> {
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -127,12 +127,7 @@ impl BlockProduction {
         input_data: GenerateBlockInputData,
         time_getter: TimeGetter,
     ) -> Result<
-        (
-            ConsensusData,
-            BlockReward,
-            GenBlockIndex,
-            FinalizeBlockInputData,
-        ),
+        (ConsensusData, BlockReward, GenBlockIndex, FinalizeBlockInputData),
         BlockProductionError,
     > {
         let consensus_data = self
@@ -198,12 +193,7 @@ impl BlockProduction {
                         input_data,
                     )?;
 
-                    Ok((
-                        consensus_data,
-                        block_reward,
-                        best_block_index,
-                        finalize_block_data,
-                    ))
+                    Ok((consensus_data, block_reward, best_block_index, finalize_block_data))
                 }
             })
             .await?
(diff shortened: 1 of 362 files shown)


chainstate/test-suite/src/tests/syncing_tests.rs
```diff
@@ -131,10 +128,7 @@ fn get_headers(#[case] seed: Seed) {
 
         // The locator is from this exact chain, so `get_headers` should return an empty sequence.
         let locator = tf.chainstate.get_locator().unwrap();
-        assert_eq!(
-            tf.chainstate.get_headers(locator.clone(), header_limit).unwrap(),
-            vec![],
-        );
+        assert_eq!(tf.chainstate.get_headers(locator.clone(), header_limit).unwrap(), vec![],);
```
iljakuklic, line 131: Quite annoying that the formatter does not handle trailing commas in macros. At least for some known ones where it's expected to be fine.

blockprod/src/detail/mod.rs
```diff
@@ -98,10 +98,7 @@ impl BlockProduction {
     }
 
     pub async fn stop_all_jobs(&mut self) -> Result<usize, BlockProductionError> {
-        self.job_manager_handle
-            .stop_all_jobs()
-            .await
-            .map_err(BlockProductionError::JobManagerError)
+        self.job_manager_handle.stop_all_jobs().await.map_err(BlockProductionError::JobManagerError)
```
alfiedotwtf, line 101: > The reason why IMO we should consider this is to reduce vertical space waste, which should improve code readability.

I know I'm probably in the minority, but I somewhat disagree that moving things to longer lines, and even to one-liners helps readability - and that's coming from a Perl one-line golfer 😆

There's a concept called Saliency [1] which suggests it's better to maximise information in one area rather than spread them to the periphery e.g in the above code, your eye can see the whole modification to `self` because each operation has an obvious "dot" at the start of each line vs the one-liner where your eye has to carefully visually scan (which is easier to miss things).

Anothe thing going for having multiple things over multiple lines is that it makes diffing code a lot easier to read e.g if you were to remove the `await` above, in a multi-line diff it would be obvious at a glance vs having to scan the single line looking for a change.

[1] https://jov.arvojournals.org/article.aspx?articleid=2193531

alfiedotwtf:
> The reason why IMO we should consider this is to reduce vertical space waste

Emacs in my terminal shows 92 lines with the default settings for my font (Victor Mono SemiBold). Though I've found this too short and so I instead use 0.755 for "Line Spacing" which gives me 120 lines. Not only does this give me more vertical space, it emphasises groupings of code making me better see "paragraphs".

Also, if more vertical code is what you're after, I've found that using horizontal split windows (i.e split full screen into 3) and then scrolling the code between the 3 columns, you got 360 vertical lines :)

PR #1039, opened 2023-07-12
Add Blockprod functional tests

alfiedotwtf:
This is part of #1004

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/error.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/error.rs
@@ -121,6 +121,8 @@ pub enum ConnectTransactionError {
         "Reward in block {0} for owner of the pool {1} which is {2:?} cannot be bigger than total reward {3:?}"
     )]
     PoolOwnerRewardCannotExceedTotalReward(Id<Block>, PoolId, Amount, Amount),
+    #[error("Unexpected pool id in kernel {0} doesn't match pool id {1}")]
+    UnexpectedPoolId(PoolId, PoolId),
     #[error("Failed to sum block {0} reward for pool {1} delegations")]
     DelegationsRewardSumFailed(Id<Block>, PoolId),
     #[error("Reward for delegation {0} overflowed: {1:?}*{2:?}/{3:?}")]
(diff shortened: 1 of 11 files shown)


azarovh (approved):

chainstate/tx-verifier/src/transaction_verifier/error.rs
```diff
@@ -280,7 +280,7 @@ pub enum SpendStakeError {
     InvalidBlockRewardOutputType,
     #[error("Stake pool data in kernel doesn't match data in block reward output")]
     StakePoolDataMismatch,
-    #[error("Pool id in kernel {0} doesn't match id in block reward output {1}")]
+    #[error("Pool id in kernel {0} doesn't match the expected pool id {1}")]
```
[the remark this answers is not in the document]
alfiedotwtf: Ah yeah, makes sense. I didn't know if it was better to go more general or making a new error. Cool, will fix.

test/functional/test_framework/__init__.py
```diff
@@ -34,18 +34,82 @@ def init_mintlayer_types():
...
             "TxOutput": {
                 "type": "enum",
                 "type_mapping": [
                     ["Transfer", "(OutputValue, Destination)"],
-                    # TODO other purpose types
+                    ["LockThenTransfer", "(OutputValue, Destination, OutputTimeLock)"],
+                    ["Burn", "(OutputValue)"],
+                    ["CreateStakePool", "(PoolId, Box<StakePoolData>)"],
+                    ["ProduceBlockFromStake", "(Destination, PoolId)"],
+                    ["CreateDeligationId", "(Destination, PoolId)"],
```
TheQuantumPhysicist, line 73: Delegation

test/functional/blockprod_generate_pos_blocks.py
```diff
@@ -0,0 +1,372 @@
...
+
+            kernel_input_utxo = {
+                "ProduceBlockFromStake": [
+                    {
+                        "PublicKey": new_stake_public_key,
+                    },
+                    new_pool_id,
+                ],
+            }
+
```
TheQuantumPhysicist, line 370: Could you also pull blocks by height (all the ones you expect to have been created), and verify that they're PoS in ConsensusData? Then verify it's constituting a chain (every previous block is the one before).

First you get the block id by height, then you pull the block by the id you got.

test/functional/blockprod_generate_pos_blocks.py
```diff
@@ -0,0 +1,372 @@
...
+            },
+            "signatures": [
+                {
+                    "NoSignature": None,
+                },
+            ],
+        }
+
+        (transfer_transaction_encoded, transfer_transaction_id) = self.pack_transaction(transfer_transaction)
+        self.generate_block(block_input_data, [transfer_transaction_encoded])
```
TheQuantumPhysicist, line 230: Please split this test into two tests. In one you only use the genesis pool, and in another, you use the one you created. Each prove different properties.

Keep in mind that the one with the pool may not work in the future if we enforce having stake balances come from sealed blocks.

TheQuantumPhysicist: To be clearer, the shortest path here is to add another functional test where you just stake from the genesis pool.

PR #1040, opened 2023-07-12
Add RegTest RPC methods for Genesis values

alfiedotwtf:
This is part of #1004

### DIFF
--- a/test-rpc-functions/src/rpc.rs
+++ b/test-rpc-functions/src/rpc.rs
@@ -29,19 +29,19 @@ use crate::RpcTestFunctionsError;
 #[rpc::rpc(server, namespace = "test_functions")]
 trait RpcTestFunctionsRpc {
     #[method(name = "genesis_pool_id")]
-    async fn genesis_pool_id(&self) -> rpc::Result<String>;
+    async fn genesis_pool_id(&self) -> rpc::Result<Option<String>>;
 
     #[method(name = "genesis_private_key")]
-    async fn genesis_private_key(&self) -> rpc::Result<String>;
+    async fn genesis_private_key(&self) -> rpc::Result<Option<String>>;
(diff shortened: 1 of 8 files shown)


test-rpc-functions/src/rpc.rs
```diff
@@ -77,6 +93,31 @@ trait RpcTestFunctionsRpc {
 
 #[async_trait::async_trait]
 impl RpcTestFunctionsRpcServer for super::RpcTestFunctionsHandle {
+    async fn genesis_pool_id(&self) -> rpc::Result<String> {
+        let (genesis_pool_id, _, _, _, _, _) = regtest_genesis_values();
+        Ok(genesis_pool_id.hex_encode())
+    }
+
+    async fn genesis_private_key(&self) -> rpc::Result<String> {
+        let (_, _, genesis_stake_private_key, _, _, _) = regtest_genesis_values();
```
TheQuantumPhysicist, line 102: Perhaps there should be asserts that double-check that the private key(s) are compatible with the values inside genesis. This whole thing is pegged on regtest_genesis_values() being consistent with genesis, which can change in the future.

TheQuantumPhysicist (approved):

test-rpc-functions/src/rpc.rs
```diff
@@ -27,6 +28,21 @@ use crate::RpcTestFunctionsError;
 
 #[rpc::rpc(server, namespace = "test_functions")]
 trait RpcTestFunctionsRpc {
+    #[method(name = "genesis_pool_id")]
+    async fn genesis_pool_id(&self) -> rpc::Result<String>;
```
azarovh, line 32: Should these functions return Option?

azarovh (approved):

PR #1041, opened 2023-07-12
Change parameters of difficulty adjustment alrgorithm

azarovh:
The goal was to improve current DAA performance and get less swings in difficulty and more consistent block time.

Simulation results using `difficulty_adjustment_simulation` can be found here:
[block time simulation.xlsx](https://github.com/mintlayer/mintlayer-core/files/12030350/block.time.simulation.xlsx)
Setting from master gives result represented on sheet sma_5
The accepted setting is located on sheet sma_100_1_per_1000

### DIFF
--- a/common/src/chain/pos.rs
+++ b/common/src/chain/pos.rs
@@ -106,14 +106,17 @@ impl PoSChainConfig {
 }
 
 pub fn create_testnet_pos_config() -> PoSChainConfig {
+    let target_block_time = NonZeroU64::new(2 * 60).expect("cannot be 0");
+    let target_limit = Uint256::MAX / Uint256::from_u64(target_block_time.get());
+
     PoSChainConfig {
-        target_limit: Uint256::MAX,
-        target_block_time: NonZeroU64::new(2 * 60).expect("cannot be 0"),
+        target_limit,
+        target_block_time,
         reward_maturity_distance: 2000.into(),
         decommission_maturity_distance: 2000.into(),
         spend_share_maturity_distance: 2000.into(),
-        block_count_to_average_for_blocktime: 5,
-        difficulty_change_limit: PerThousand::new(100).expect("must be valid"),
+        block_count_to_average_for_blocktime: 100,
+        difficulty_change_limit: PerThousand::new(1).expect("must be valid"),
     }
 }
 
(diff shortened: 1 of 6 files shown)


common/src/chain/pos.rs
```diff
@@ -106,14 +106,17 @@ impl PoSChainConfig {
...
-        target_block_time: NonZeroU64::new(2 * 60).expect("cannot be 0"),
+        target_limit,
+        target_block_time,
         reward_maturity_distance: 2000.into(),
         decommission_maturity_distance: 2000.into(),
         spend_share_maturity_distance: 2000.into(),
-        block_count_to_average_for_blocktime: 5,
-        difficulty_change_limit: PerThousand::new(100).expect("must be valid"),
+        block_count_to_average_for_blocktime: 100,
+        difficulty_change_limit: PerThousand::new(1).expect("must be valid"),
```
iljakuklic, line 119: 1/1000 change limit seems pretty tight. Is the adjustment done on every block?

iljakuklic:
> Simulation results using `difficulty_adjustment_simulation` can be found here: [block time simulation.xlsx](https://github.com/mintlayer/mintlayer-core/files/12030350/block.time.simulation.xlsx) Setting from master gives result represented on sheet sma_5 The accepted setting is located on sheet sma_100_1_per_1000

Would be interesting to see how these react to a "shock" increase (say doubling) in block production rate. How long does it take for the difficulty to get close to double?

common/src/chain/pos.rs
TheQuantumPhysicist: Yes

azarovh: right, that's tight because we adjust per block

common/src/chain/pos.rs
```diff
@@ -124,20 +127,23 @@ pub fn create_unittest_pos_config() -> PoSChainConfig {
         reward_maturity_distance: 2000.into(),
         decommission_maturity_distance: 2000.into(),
         spend_share_maturity_distance: 2000.into(),
-        block_count_to_average_for_blocktime: 5,
-        difficulty_change_limit: PerThousand::new(100).expect("must be valid"),
+        block_count_to_average_for_blocktime: 100,
```
TheQuantumPhysicist, line 130: If this is the same number everywhere, please use a `const` somewhere for it. Same for others.

TheQuantumPhysicist (approved):
Looking good. Please fix the const thing I mentioned.

azarovh:
> Would be interesting to see how these react to a "shock" increase (say doubling) in block production rate. How long does it take for the difficulty to get close to double?

tbh it's hard to analyse such things using a spreadsheet. I will try to learn how to work with Mathematica during the weekend and hope it will help. What I can say now is that it takes a while.

PR #1042, opened 2023-07-13
Only sealed pools can be used for staking

azarovh:
Use Sealed storage for consensus which should eliminate potential adversary behaviour.

P.S. @alfiedotwtf this will break your upcoming functional tests, I can help fixing that when you finish with your changes


TheQuantumPhysicist (approved):
Looking good.

@alfiedotwtf please don't forget to rebase once this is merged.

iljakuklic (approved):

PR #1044, opened 2023-07-13
Remove reward maturity setting for PoS

azarovh:
Reward maturity setting doesn't make sense for PoS because it cannot be spent.

Also added tests to ensure that decommission_maturity and spend_share_maturity follows `NetUpgrade` setting


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -555,8 +555,16 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
                     ConsensusData::None | ConsensusData::PoW(_) => match output {
```
TheQuantumPhysicist, line 555: Instead of branching then joining, please create a `required` for each branch individually.

TheQuantumPhysicist (approved):
Looking good. Though please fix the little issue I mentioned, then feel free to merge.

PR #1047, opened 2023-07-14
Calculate transaction fees based on fee rate

OBorce:
Take current fee rate from mempool
Calculate fee based on the current fee rate and expected transaction size.

The fee can get overestimated by couple of bytes per change output because of the non-fixed size encoded numbers in them.

### DIFF
--- a/mempool/src/interface/mempool_interface.rs
+++ b/mempool/src/interface/mempool_interface.rs
@@ -14,8 +14,8 @@
 // limitations under the License.
 
 use crate::{
-    error::Error, event::MempoolEvent, tx_accumulator::TransactionAccumulator, MempoolMaxSize,
-    TxOrigin, TxStatus,
+    error::Error, event::MempoolEvent, tx_accumulator::TransactionAccumulator, FeeRate,
+    MempoolMaxSize, TxOrigin, TxStatus,
 };
 use common::{
     chain::{GenBlock, SignedTransaction, Transaction},
@@ -70,7 +70,7 @@ pub trait MempoolInterface: Send + Sync {
     fn set_max_size(&mut self, max_size: MempoolMaxSize) -> Result<(), Error>;
 
     /// Get current fee rate
-    fn get_fee_rate(&self) -> u128;
+    fn get_fee_rate(&self, in_top_x_mb: usize) -> Result<FeeRate, Error>;
 }
 
 #[async_trait::async_trait]
--- a/mempool/src/interface/mempool_interface.rs
+++ b/mempool/src/interface/mempool_interface.rs
@@ -68,6 +68,9 @@ pub trait MempoolInterface: Send + Sync {
 
     /// Set the maximum mempool size
     fn set_max_size(&mut self, max_size: MempoolMaxSize) -> Result<(), Error>;
+
+    /// Get current fee rate
+    fn get_fee_rate(&self) -> u128;
 }
 
 #[async_trait::async_trait]
--- a/mempool/src/interface/mempool_interface.rs
+++ b/mempool/src/interface/mempool_interface.rs
@@ -69,7 +69,8 @@ pub trait MempoolInterface: Send + Sync {
     /// Set the maximum mempool size
     fn set_max_size(&mut self, max_size: MempoolMaxSize) -> Result<(), Error>;
 
-    /// Get current fee rate
+    /// Get the fee rate such that it would put the new transaction in the top X MB of the mempool
+    /// making it less likely to get rejected or trimmed in the case the mempool is full
     fn get_fee_rate(&self, in_top_x_mb: usize) -> Result<FeeRate, Error>;
 }
 
--- a/mempool/src/pool/mod.rs
+++ b/mempool/src/pool/mod.rs
@@ -1032,7 +1032,7 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
         self.events_controller.broadcast(event.into());
     }
 
-    pub fn current_fee_rate(&self, in_top_x_mb: usize) -> Result<FeeRate, MempoolPolicyError> {
+    pub fn get_fee_rate(&self, in_top_x_mb: usize) -> Result<FeeRate, MempoolPolicyError> {
         let mut total_size = 0;
         self.store
             .txs_by_descendant_score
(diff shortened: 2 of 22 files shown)


mempool/src/pool/mod.rs
```diff
@@ -975,6 +976,10 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
         let event = event::NewTip::new(block_id, block_height);
         self.events_controller.broadcast(event.into());
     }
+
+    pub fn current_fee_rate(&self) -> u128 {
+        self.rolling_fee_rate.read().rolling_minimum_fee_rate().atoms_per_kb()
+    }
```
iljakuklic, line 982: Please note that minimum fee rate is the minimum fee rate to get the transaction into mempool. It's probably not a good idea to use this for transaction fee rate if there should be any hope for the transaction to be included soon. If mempool is close to being full, chances of the transaction of this feerate being evicted are pretty high. A better number is something like the fee rate that places you N megabytes from the top, where N is a user-tunable parameter.

mempool/src/interface/mempool_interface.rs
```diff
@@ -67,6 +67,9 @@ pub trait MempoolInterface: Send + Sync {
 
     /// Set the maximum mempool size
     fn set_max_size(&mut self, max_size: MempoolMaxSize) -> Result<(), Error>;
+
+    /// Get current fee rate
```
TheQuantumPhysicist, line 71: I think this is the upper-bound of the feerate, right?

Besides that, when returning a raw value, u128, it's always helpful to add the units. I believe it's atoms/kB, right?

wallet/src/account/mod.rs
```diff
@@ -868,5 +939,66 @@ fn group_utxos_for_input<T, Grouped: Clone>(
...
+    let tx = SignedTransaction::new(
+        Transaction::new(1, vec![], outputs.into()).expect("should not fail"),
+        vec![],
+    )
+    .expect("should not fail");
+    serialization::Encode::encoded_size(&tx)
+}
+
+fn input_signature_size(destination: &Destination) -> usize {
+    let sig_hex = "003c002dd5ea8f05240394eb109b8e9b52716db2720cebbc5dd66394afdf761d2ef76dc8a292d3a44e4d9f1f8fd8e17dc404e317082f5a2ce8adafde01e766aaa6";
```
TheQuantumPhysicist, line 952: This isn't really what I meant. In the crypto crate, there's the `Signature` type, which can have different signatures. There should be something like a method there that returns the maximum encoded size of the signature.

wallet/src/account/mod.rs
```diff
@@ -868,5 +939,66 @@ fn group_utxos_for_input<T, Grouped: Clone>(
...
+    };
+
+    let input_sig = InputWitness::Standard(StandardInputSignature::new(
+        SigHashType::default(),
+        raw_signature,
+    ));
+    serialization::Encode::encoded_size(&input_sig)
+}
+
+fn coin_and_token_output_fees(feerate: mempool::FeeRate) -> WalletResult<(Amount, Amount)> {
```
TheQuantumPhysicist, line 977: I'm wondering... the FeeRate type is public, why are we not using that in RPC, and are using u128 there instead?

wallet/src/account/mod.rs
```diff
@@ -186,14 +216,29 @@ impl Account {
...
+                    Currency::Token(_) => token_change_fee,
+                };
+                let selection_result = select_coins(
+                    utxos,
+                    *output_amount,
+                    PayFee::DoNotPayFeeWithThisCurrency,
+                    cost_of_change,
+                )?;
 
                 total_fees_not_payed = (total_fees_not_payed + selection_result.get_total_fees())
```
TheQuantumPhysicist, line 233: paid

TheQuantumPhysicist (approved):
I don't have big complaints. Looking good. Please mind the comments I made and the MAX issue I mentioned in chat.

wallet/src/account/mod.rs
OBorce: fixed

wallet/src/account/mod.rs
OBorce: replaced with precalculated sizes

mempool/src/interface/mempool_interface.rs
OBorce: replaced with the lowest feerate to be included in the first X MB in the mempool, hardcoded to 5MB currently.

mempool/src/pool/mod.rs
OBorce: Changed the algorithm, if you can check it out if you think it is ok

muursh (approved):
No real concerns from me

iljakuklic (commented):
Looks good. Some minor comments below.

wallet/src/account/mod.rs
```diff
@@ -868,5 +934,50 @@ fn group_utxos_for_input<T, Grouped: Clone>(
...
+    serialization::Encode::encoded_size(&tx)
+}
+
+fn input_signature_size(destination: &Destination) -> usize {
+    // Sizes calculated upfront
+    match destination {
+        Destination::Address(_) => 103,
+        Destination::PublicKey(_) => 69,
+        Destination::AnyoneCanSpend => 2,
+        Destination::ScriptHash(_) | Destination::ClassicMultisig(_) => unimplemented!(),
```
TheQuantumPhysicist, line 952: This is a bad idea. Please return an error instead. We don't want transactions to cause crashes in the wallet.

mempool/src/interface/mempool_interface.rs
TheQuantumPhysicist: Please add some more documentation

wallet/src/wallet/mod.rs
```diff
@@ -420,11 +421,19 @@ impl<B: storage::Backend> Wallet<B> {
         &mut self,
         account_index: U31,
         outputs: impl IntoIterator<Item = TxOutput>,
+        current_fee_rate: FeeRate,
+        long_term_fee_rate: FeeRate,
```
TheQuantumPhysicist, line 425: Please add docs explaining the different types of fee rate and why

TheQuantumPhysicist (approved):

azarovh (approved):

wallet/src/wallet/tests.rs
```diff
@@ -57,10 +57,34 @@ fn gen_random_password(rng: &mut (impl Rng + CryptoRng)) -> String {
...
 }
 
+fn gen_random_transfer(rng: &mut (impl Rng + CryptoRng), amount: Amount) -> TxOutput {
+    let destination = Destination::PublicKey(
+        crypto::key::PrivateKey::new_from_rng(rng, crypto::key::KeyKind::Secp256k1Schnorr).1,
+    );
+    match rng.gen::<bool>() {
+        true => TxOutput::Transfer(OutputValue::Coin(amount), destination),
+        false => {
+            let lock_for_blocks = rng.gen::<u32>() as u64;
```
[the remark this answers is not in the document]
azarovh: sorry this will probably overflow BlockDistance. But I think other locals can be generated as u64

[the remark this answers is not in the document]
OBorce: replaced with gen_range, as even with u32 it can overflow if both seconds_between_blocks and lock_for blocks get u32::max and then adding the result to the current_timestamp

PR #1048, opened 2023-07-14
Adding tip-checking before returning a block

alfiedotwtf:
This change ensures block production never returns a "current" out-of-tip block


alfiedotwtf:
Bah... current commit is still the old pulled tip. I need to pull again before checking the value. Updating now

TheQuantumPhysicist:
Closing in favor of #1049

PR #1049, opened 2023-07-14
Mempool rejects transaction accumulators with wrong tips

TheQuantumPhysicist:
It's possible that there's a race between the mempool updating its tip and the block production. Hence, in this PR, we enforce having both the mempool and block production having the same tip as a condition for the mempool to provide transactions.

In addition, we establish the concept of "rejecting the transaction accumulator" as basis for any other disagreements between the accumulator and the mempool.

### DIFF
--- a/mempool/src/interface/mempool_interface_impl.rs
+++ b/mempool/src/interface/mempool_interface_impl.rs
@@ -139,7 +139,7 @@ impl MempoolInterface for Mempool {
     fn collect_txs(
         &self,
         tx_accumulator: Box<dyn TransactionAccumulator + Send>,
-    ) -> Result<Box<dyn TransactionAccumulator>, Error> {
+    ) -> Result<Option<Box<dyn TransactionAccumulator>>, Error> {
         Ok(self.collect_txs(tx_accumulator))
     }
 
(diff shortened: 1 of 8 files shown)


alfiedotwtf (approved):

mempool/src/interface/mempool_interface_impl.rs
```diff
@@ -139,7 +139,7 @@ impl MempoolInterface for Mempool {
     fn collect_txs(
         &self,
         tx_accumulator: Box<dyn TransactionAccumulator + Send>,
-    ) -> Result<Box<dyn TransactionAccumulator>, Error> {
+    ) -> Result<Option<Box<dyn TransactionAccumulator>>, Error> {
```
iljakuklic, line 142: Why `Option` rather than adding an arm to `Error`, or having `Result<Result<`? It seems that having some sort of descriptive "error" type is warranted in this case.

iljakuklic (approved):
Looks good

mempool/src/interface/mempool_interface_impl.rs
TheQuantumPhysicist: The problem with errors is the different handling. Returning an error can mean something different. The ideal solution is to have another enum instead of the Option, but I don't care about that right now. In the future we can improve it.

PR #1059, opened 2023-07-18
Block invalidation

ImplOfAnImpl:
- `invalidate_block` and `reset_block_failure_flags` functions were implemented;
- `min_height_with_allowed_reorg` is now stored in the DB;
- casting between `&Id<Block>` and `&Id<GenBlock>` was implemented using `ref_cast`;
- some minor improvements;


common/src/chain/gen_block.rs
```diff
@@ -93,3 +115,23 @@ impl GenBlockId {
         }
     }
 }
+
+/// Classified generalized block reference
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub enum GenBlockIdRef<'a> {
```
TheQuantumPhysicist, line 121: Why do we need this type? Seems like an overkill and complicates things...

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -185,6 +193,31 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
             .map_err(PropertyQueryError::from)
     }
 
+    pub fn get_existing_gen_block_index(
```
TheQuantumPhysicist, line 196: Why is this function necessary? Why not use the one above it?

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -185,6 +193,31 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+            .ok_or(PropertyQueryError::BlockIndexNotFound(*block_id))
+    }
+
+    pub fn get_best_block_index(&self) -> Result<GenBlockIndex, PropertyQueryError> {
+        self.get_gen_block_index(&self.get_best_block_id().log_err()?)
+            .log_err()?
+            .ok_or(PropertyQueryError::BestBlockIndexNotFound)
+    }
+
+    /// Read previous block from storage and return its BlockIndex.
```
TheQuantumPhysicist, line 210: I don't think you're reading a block

chainstate/src/detail/chainstateref/mod.rs
ImplOfAnImpl: By "the one above it" do you mean `get_gen_block_index`? The "get_existing" variant implies that the block index must exist and returns an error when it doesn't, while `get_gen_block_index` returns an Option. It is not necessary but it reduces the verbosity of the caller code.
If you mean `get_existing_block_index` (i.e. the one without "gen") - well sometimes it's more convenient to work with "gen" variants of Ids and Indices, sometimes it's not. So, 2 variants of each function is often desired.

Btw, `get_existing_gen_block_index` is currently unused, I only used it in an intermediate implementation. So, I can remove it or keep it "for completeness".

TheQuantumPhysicist: Yes, please remove it. I don't think such redundancy is a good idea.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -932,36 +998,41 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
...
             let block: WithId<Block> = self
                 .get_block_from_index(&block_index)
                 .map_err(BlockError::StorageError)
-                .and_then(|opt| {
-                    opt.ok_or(BlockError::InvariantErrorBlockNotFoundAfterConnect(
-                        (*block_index.block_id()).into(),
-                    ))
+                .log_err()?
+                .ok_or_else(|| {
+                    // Note: missing block data is an expected situation; it can happen if
```
TheQuantumPhysicist, line 1012: One solution that occurred to me is to just delete the block index altogether in that case. But we don't have to do that right now.

chainstate/src/detail/ban_score.rs
```diff
@@ -61,18 +60,19 @@ impl BanScore for BlockError {
             BlockError::TxIndexConstructionError(_) => 100,
             BlockError::PoSAccountingError(err) => err.ban_score(),
             BlockError::EpochSealError(err) => err.ban_score(),
-            BlockError::BlockHeightTooBig(_) => 0,
+            BlockError::BlockTooDeepToInvalidate(_) => 0,
+            BlockError::BestChainCandidatesError(_) => 0,
+            BlockError::BlockNotFound(_) => 0,
```
TheQuantumPhysicist, line 65: Please make the error more verbose by including the context in the name. This runs the danger of someone else using this error without noticing it has 0 ban score.

chainstate/src/detail/ban_score.rs
```diff
@@ -61,18 +60,19 @@ impl BanScore for BlockError {
...
 
             BlockError::BestBlockIdQueryError(_) => 0,
+            BlockError::BestBlockIndexQueryError(_) => 0,
             BlockError::IsBlockInMainChainQueryError(_, _) => 0,
-            BlockError::BlockIdTreeTopQueryError(_, _) => 0,
             BlockError::BlockIndexQueryError(_, _) => 0,
+            BlockError::BlockIndicesForBranchQueryError(_) => 0,
+            BlockError::MinHeightForReorgQueryError(_) => 0,
```
TheQuantumPhysicist, line 72: If all these errors have to do with invalidation, perhaps we need another error type and a submodule that handles invalidation.

chainstate/src/detail/best_chain_candidates.rs
```diff
@@ -0,0 +1,177 @@
...
+use crate::TransactionVerificationStrategy;
+use chainstate_storage::BlockchainStorageRead;
+use chainstate_types::{BlockIndex, GenBlockIndex, PropertyQueryError};
+use common::{chain::Block, primitives::Id, Uint256};
+use utils::tap_error_log::LogError;
+
+#[derive(Eq, Copy, Clone, Debug)]
+pub struct BestChainCandidatesItem {
+    pub chain_trust: Uint256,
+    pub block_id: Id<Block>,
```
TheQuantumPhysicist, line 30: Please use getters...

chainstate/src/detail/best_chain_candidates.rs
```diff
@@ -0,0 +1,177 @@
...
+        min_chain_trust: Uint256,
+    ) -> Result<BestChainCandidates, BestChainCandidatesError>
+    where
+        S: BlockchainStorageRead,
+        V: TransactionVerificationStrategy,
+    {
+        let best_block_index = chainstate_ref.get_best_block_index().log_err()?;
+        let min_height_with_allowed_reorg = chainstate_ref.get_min_height_with_allowed_reorg()?;
+
+        let block_ids_by_height = chainstate_ref
```
TheQuantumPhysicist, line 79: Please add a comment explaining (again at the caller) that this is linear.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -832,7 +856,17 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+        self.get_block_ids_by_height(0.into())
+    }
+
+    /// Return ids of all blocks with height bigger than the specified one,
+    /// sorted by height (lower first).
+    // TODO: this function iterates over all block indices in the DB, which is too expensive
+    // for places where it's currently used (such as block invalidation or best chain selection).
+    // We need either to optimize it or replace it with some other solution.
+    // See https://github.com/mintlayer/mintlayer-core/issues/1033, item #5.
+    pub fn get_block_ids_by_height(
```
TheQuantumPhysicist, line 869: get_block_ids_higher_than()

TheQuantumPhysicist (approved):
Looking good, please, as agreed, do the refactoring to submodules and look into the comments.

chainstate/src/detail/chainstateref/mod.rs
ImplOfAnImpl: I've mentioned it in the comment for now.

chainstate/src/detail/best_chain_candidates.rs
ImplOfAnImpl: I've added the comment.

Also, I've just realized that the call to `last_common_ancestor` below is more expensive than I thought - it iterates towards the common ancestor block by block instead of taking bigger jumps via `get_ancestor`. So, first of all, I replaced the call to `last_common_ancestor` with `last_common_ancestor_in_main_chain`. Currently, the two functions basically do the same. But the latter can actually be optimized so that it takes bigger jumps via `get_ancestor`, so I wrote a TODO about it.

chainstate/src/detail/chainstateref/mod.rs
ImplOfAnImpl: I've decided to use the name `get_higher_block_ids_sorted_by_height` to reflect that the returned blocks both have a height bigger than the specified one and that they are sorted by height.

common/src/chain/gen_block.rs
ImplOfAnImpl: [discussed] `GenBlockIdRef` and `Id<GenBlock>::classify_ref` have been removed.

chainstate/src/detail/chainstateref/mod.rs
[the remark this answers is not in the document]
ImplOfAnImpl: [discussed] I've separated block invalidation and block checking off from Chainstate similarly to how it was done in https://github.com/mintlayer/mintlayer-core/pull/1075

chainstate/src/lib.rs
```diff
@@ -99,3 +102,35 @@ pub fn make_chainstate<
...
+
+#[doc(hidden)]
+pub mod integration_tests_support {
+    use super::*;
+
+    pub use crate::interface::chainstate_interface::integration_tests_support::*;
+
+    // This is the same as the normal make_chainstate, but it returns ChainstateTestInterface
+    // instead.
+    pub fn make_chainstate<
```
TheQuantumPhysicist, line 114: If you're going this far, why not just keep the tests locally here, in chainstate, instead of moving them to the framework?

ImplOfAnImpl: Well, they depend on the framework and in general they are integration tests rather than unit tests.

iljakuklic (approved):
Looks good. There's a number of comments, all of them minor IIRC.

Cargo.toml
```diff
@@ -157,6 +157,7 @@ quote = "1.0"
 rand = "0.8"
 rand_chacha = "0.3"
 reedline = "0.19"
+ref-cast = "1.0.18"
```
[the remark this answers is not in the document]
ImplOfAnImpl: No specific reason, it's just what `cargo add` does. Ok, I'll remove the patch version here and also for `derive_more` and `enum-iterator` which I've added earlier.

common/src/primitives/id/mod.rs
```diff
@@ -143,6 +145,10 @@ impl<T> Id<T> {
         self.id
     }
 
+    pub const fn get_ref(&self) -> &H256 {
+        &self.id
+    }
```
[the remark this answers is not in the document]
ImplOfAnImpl: I don't think it's a conversion, it's rather an accessor for the lower-level representation of the id.

Also, above it there is the function
```
pub const fn get(&self) -> H256 {
    self.id
}
```
So, in `get_ref` the `get` part is more like a stem of the name rather than a prefix, and `ref` is the suffix.

IMO both names should be consistent with each other. So if the current variant feels off, I'd suggest one of the following:
1) Rename `get` to `inner` and `get_ref` to `inner_ref`.
2) Rename `get` to `block_hash` and `get_ref` to `block_hash_ref` (it's kind of lengthy, but better reflects the meaning of the value).
2) Only leave one function that returns a reference, name it either `get`, `inner` or `block_hash`.

chainstate/src/detail/mod.rs
```diff
@@ -739,6 +669,16 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
         &self.events_controller
     }
 
+    // Note: this is exposed for testing only.
+    pub fn get_best_chain_candidates(
```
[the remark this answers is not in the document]
ImplOfAnImpl: Not sure.

I wonder how it'll work in this case though. Will chainstate be compiled twice, i.e. once for test-framework and once for everything else?

Cargo.toml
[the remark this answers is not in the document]
ImplOfAnImpl: I fixed it here - https://github.com/mintlayer/mintlayer-core/pull/1080

common/src/primitives/id/mod.rs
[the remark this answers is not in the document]
iljakuklic: I think `fn to_hash(&self) -> H256` and `fn as_hash(&self) -> &H256` should work fairly well and agree with the API guidelines document linked above. Equivalent signatures named `to_inner`, `as_inner` also work. However, renaming the existing `get` method may involve updating a non-trivial number of call sites.

azarovh (approved):

chainstate/src/detail/best_chain_candidates.rs
```diff
@@ -0,0 +1,177 @@
...
+impl BestChainCandidatesItem {
+    fn from_block_index(block_index: &BlockIndex) -> Self {
+        BestChainCandidatesItem {
+            block_id: *block_index.block_id(),
+            chain_trust: block_index.chain_trust(),
+        }
+    }
+}
+
+impl Ord for BestChainCandidatesItem {
```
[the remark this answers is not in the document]
ImplOfAnImpl: Initially, the struct had an extra field that didn't participate in the ordering, so I had to have a manual implementation. But after removing the field I decided to keep the manual implementation, because without it the correctness of the comparison would depend on the order of the fields in the struct, which seemed a bit fragile.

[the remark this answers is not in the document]
azarovh: > the comparison would depend on the order of the fields

But do we care unless it's consistent. The problem is that it's easy to forget to update the method if a field is added. I'd go with derives for Ord and also for PartialOrd and PartialEq below.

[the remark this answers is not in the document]
ImplOfAnImpl: The items must be sorted by chain_trust first, that's the whole point of the set.
Also, it's a kind of struct where when adding a field you should first ask yourself whether it should participate in the ordering in the first place.
Anyway, I don't think that this struct will ever be changed (other than being removed completely), so I don't mind getting rid of the manual impl in favour of a comment that would state that chain_trust should always come first.

chainstate/src/detail/block_invalidation/best_chain_candidates.rs
```diff
@@ -0,0 +1,191 @@
...
+use thiserror::Error;
+
+use crate::{detail::chainstateref::ChainstateRef, TransactionVerificationStrategy};
+use chainstate_storage::BlockchainStorageRead;
+use chainstate_types::{BlockIndex, GenBlockIndex, PropertyQueryError};
+use common::{chain::Block, primitives::Id, Uint256};
+use utils::tap_error_log::LogError;
+
+#[derive(Eq, Copy, Clone, Debug)]
+pub struct BestChainCandidatesItem {
```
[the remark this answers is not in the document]
iljakuklic: Ok, not a big deal

common/src/primitives/id/mod.rs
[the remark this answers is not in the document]
ImplOfAnImpl: > However, renaming the existing get method may involve updating a non-trivial number of call sites.

"Rename symbol" in VSCode does all the updates automatically, so it is trivial.
Ok, I've renamed them to `to_hash` and `as_hash` respectively and also renamed the `id: H256` field of `Id` to `hash` for consistency.

chainstate/src/detail/best_chain_candidates.rs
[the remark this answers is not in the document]
ImplOfAnImpl: [discussed] I've written a comment near Ord::cmp mentioning that the order of fields is important and that chain_trust must come first. Otherwise the code stays the same.

chainstate/src/detail/mod.rs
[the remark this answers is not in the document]
ImplOfAnImpl: [discussed with Sam] This function was removed.

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -234,3 +237,22 @@ pub trait ChainstateInterface: Send {
         account: AccountType,
     ) -> Result<Option<AccountNonce>, ChainstateError>;
 }
+
+#[doc(hidden)]
+pub mod integration_tests_support {
```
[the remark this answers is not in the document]
ImplOfAnImpl: [discussed with Sam] This module was removed.

chainstate/src/lib.rs
ImplOfAnImpl: [discussed] I removed the `integration_tests_support` module and the checks of BestChainCandidates that were performed by the integration tests. Then I'll have to refactor BestChainCandidates to abstract away ChainstateRef, so that it can be unit-tested more easily. Since at this moment there are more important things to do, this PR will be postponed.

Issue #1003, opened 2023-06-28
Assign Job Key ID when none supplied in produce_block_with_custom_id()


alfiedotwtf:
Currently when no `custom_id` is supplied to `produce_block_with_custom_id()`, the job manager will create the job with no key. However if invoked again while the first job is still running, as the second `None` key already exists from the first job, it will error with `JobAlreadyExists`.

This fix will generate a random key if `None` was provided.

alfiedotwtf:
We also need to allow for the following scenarios:

- calls to `produce_block_with_custom_id()` need to be allowed to run in parallel while testing
- calls to `produce_block()` need to be allowed to run in parallel while testing
- need to prevent the same staking key to be used in parallel because there could be penalties

PR #1063, opened 2023-07-19
Assign a JobKey Id when None is supplied

alfiedotwtf:
See #1003 for more details.


TheQuantumPhysicist (approved):
While this fix is fine from the outside, I worry that this will make it possible to create infinitely many jobs that are doing the first thing. My hope is that clients requesting to build blocks will look at the job id and use it.

muursh (approved):

PR #1064, opened 2023-07-19
mempool: Make transaction accumulator respect dependencies

iljakuklic:
This should fix the issue where an invalid block is sometimes constructed due to inter-transaction dependencies. There will be a follow up that addresses the timelock constraints.


TheQuantumPhysicist (approved):
Looking good. Let's merge this ASAP.

PR #1066, opened 2023-07-21
Introduce ConstrainedValueAccumulator

### LINKED ISSUE #945: Implement the ConstrainedValueAccumulator struct for checking spending conditions in the TransactionVerifier
Currently the way inputs and outputs are summed to figure out whether money is printed is messy. This is because it only takes value (`Amount`) into account and ignores constraints on values.

A good example to this problem is decommissioning a pool. Imagine a transaction in which the creator puts two inputs, a transaction pool (for decommissioning) and a regular input for value. Currently the way this is validated is messy and makes this transaction invalid, where ad-hoc rules prevent doing anything else while decommissioning a pool. However, the correct way would be by looking at those two values as orthogonal values that cannot be added. Simply, as apples and oranges. 

Hence, to solve this problem, we create a struct that represents the "sum" of these orthogonal values. This struct contains values that have values and their constraints. 

```
struct ConstrainedValueAccumulator {
    unconstrained_value: OutputValue,
    constrained_with_timelock: BTreeMap<TimelockByValue, OutputValue>,
    // ... more orthogonal types
}
```

Then we have a method that add values to those:

```
impl ConstrainedValueAccumulator {
    pub fn add_value(&mut self, output: TxOutput) {
        match output {
            Transfer(value) => unconstrained_value += value,
            CreateStakePool(value) => self.constrained_with_timelock.entry(OUTPUT_LOCK_DISTANCE).value_or_default() += value,
            // ... more output types
        }
    }
}
```

This way, we add all the values available in the inputs and constrain them with whatever constraints. This is how we process inputs.

Now for processing outputs (to detect invalid spending), we do the same, but we subtract from these data structures. However, the subtraction is done only according to rules that treat some types as superset of others.

Examples:

- When subtracting `Transfer` output values, then can only "take" from `self.unconstrained_value`
- When subtracting `LockThenTransfer`, it can only subtract from values `self.constrained_with_timelock` with timelocks equal or higher than the timelock provided by the inputs when added. Then, if none is found, we can subtract from the `self.unconstrained_value`
- ... many similar rules can be added to ensure valid usage of inputs.

Trying to understand these rules, we're basically saying, if an input mandates a timelock, then it can only be spent with a timelock equal or higher than mandated. However, when an input mandates nothing (is unconstrained), it's OK to add constraints to it. 

Of course this has interesting implications on block rewards as they may require locking rules as well. 

Open for discussion. Unfortunately, I can't find free time to implement this myself.

azarovh:
The main goal is to simplify input/output policies by checking properties instead of a particular combinations (that are effectively infinite).

Closes #945 . See it for more details


mempool/src/error/ban_score.rs
```diff
@@ -357,3 +354,18 @@ impl MempoolBanScore for accounting::Error {
         }
     }
 }
+
+impl MempoolBanScore for IOPolicyError {
+    fn mempool_ban_score(&self) -> u32 {
+        match self {
+            IOPolicyError::InvalidInputTypeInReward => 100,
```
TheQuantumPhysicist, line 361: @iljakuklic Can you please double-check that these values make sense in the context of the mempool?

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    pub fn collect_and_verify(
+        &mut self,
+        tx: &Transaction,
+        chain_config: &ChainConfig,
+        block_height: BlockHeight,
+        pos_accounting_view: &impl PoSAccountingView,
+        utxo_view: &impl utxo::UtxosView,
```
TheQuantumPhysicist, line 55: Like we discussed, this should be simplified such that it doesn't have information about the utxo set or the account state. These should be pulled beforehand and put in the function parameters, so that we can have security proofs about the ConstrainedValueAccumulator.

mempool/src/error/ban_score.rs
iljakuklic: > @iljakuklic Can you please double-check that these values make sense in the context of the mempool?

Just quickly skimming over it:
* The `Invalid*Type*` arms look fine
* So does `ProduceBlockInTx` and `ConstrainedAmountOverflow` if I understand them correctly.
* I am not convinced by `MultiplePoolCreated`, `MultipleDelegationCreated` since these look like something that depends on the current chain state.
* I don't know about `TimelockRequirementNotSatisfied`, have not looked in the code too deeply to understand that. I suspect that one may be problematic too.

Please note this is based on a rather superficial reading of these errors and there may be aspects that I did not understand correctly.

In general, a rule of thumb in mempool is:
* If the error depends on the current state of the blockchain, ban score should be 0. This is because in a distributed system, it has to be tolerated for nodes to be temporarily out of sync.
* If the error signifies the transaction is invalid no matter which block it is included in, give it a ban score 100.

azarovh: `MultiplePoolCreated` and `MultipleDelegationCreated` do not depend on a state. They indicate that transaction has > 1 `CreateStakePool` or `CreateDelegationId` outputs which is invalid since the id won't be unique.

`TimelockRequirementNotSatisfied` also should be fine as it shows that not enough coins were locked in the outputs.

iljakuklic: Makes sense, It's all fine then

chainstate/tx-verifier/src/transaction_verifier/timelock_check.rs
```diff
@@ -167,88 +159,9 @@ where
...
-                outpoint_source_id,
-            )?;
-        }
-    }
-
     Ok(())
 }
 
-/// Outputs that decommission a stake pool or spend delegation share must be timelocked
-fn check_outputs_timelock(
```
TheQuantumPhysicist, line 216: I'm keeping this comment as a reference to see why this was removed

azarovh: This check was introduced to validate that if input decommissions the pool or spends share then all outputs must be locked according to the chain config. ConstraintsAccumulator has the same functionality making this function obsolete.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/mod.rs
```diff
@@ -13,334 +13,87 @@
...
+    #[error("Attempted to create multiple stake pools in a single tx")]
+    MultiplePoolCreated,
+    #[error("Attempted to create multiple delegations in a single tx")]
+    MultipleDelegationCreated,
+    #[error("Attempted to produce block in a tx")]
+    ProduceBlockInTx,
+    #[error("Amount overflow")]
+    AmountOverflow,
+    #[error("Attempt to print money or violate timelock constraints")]
+    AttemptToPrintMoneyOrViolateTimelockConstraints,
```
TheQuantumPhysicist, line 46: Please include more information in the errors...

chainstate/test-suite/src/tests/double_spend_tests.rs
```diff
@@ -224,9 +224,8 @@ fn overspend_single_output(#[case] seed: Seed) {
                 .build_and_process()
                 .unwrap_err(),
             ChainstateError::ProcessBlockError(BlockError::StateUpdateFailed(
-                ConnectTransactionError::AttemptToPrintMoney(
-                    Amount::from_atoms(tx1_output_value),
-                    Amount::from_atoms(tx2_output_value)
+                ConnectTransactionError::IOPolicyError(
+                    chainstate::IOPolicyError::AttemptToPrintMoneyOrViolateTimelockConstraints
```
TheQuantumPhysicist, line 228: This loss of information is very unfortuante... we should try to do something about it.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/purposes_check.rs
```diff
@@ -0,0 +1,187 @@
...
+    block::BlockRewardTransactable, signature::Signable, Transaction, TxInput, TxOutput,
+};
+use consensus::ConsensusPoSError;
+use utils::ensure;
+
+use crate::transaction_verifier::input_output_policy::IOPolicyError;
+
+use super::super::error::{ConnectTransactionError, SpendStakeError};
+
+// TODO: avoid collecting by using FallibleIterator
```
TheQuantumPhysicist, line 26: Please explain why

TheQuantumPhysicist: Well, I think you mean "use FallibleIterator to avoid manually collecting to a Result"

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -0,0 +1,490 @@
...
+            .sum::<Option<Amount>>()
+            .and_then(|v| v + self.unconstrained_value)
+            .ok_or(IOPolicyError::AmountOverflow)
+    }
+
+    pub fn process_input_utxo(
+        &mut self,
+        chain_config: &ChainConfig,
+        block_height: BlockHeight,
+        pos_accounting_view: &impl PoSAccountingView,
```
TheQuantumPhysicist, line 64: This has to go...

If you're really stuck with it, you can simply pass an `Fn()` to get a specific piece of information. But I'd really prefer it if all the processing was done at the caller, and this class only remains for calculations, so that we can create proofs for it.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -0,0 +1,490 @@
...
+    pub fn consume(self) -> Result<Amount, IOPolicyError> {
+        self.timelock_constrained
+            .values()
+            .copied()
+            .sum::<Option<Amount>>()
+            .and_then(|v| v + self.unconstrained_value)
+            .ok_or(IOPolicyError::AmountOverflow)
+    }
+
+    pub fn process_input_utxo(
```
TheQuantumPhysicist, line 60: Can't we make this `process_input` instead of this level of granularity?

azarovh: `process_input` would require UtxoView to get utxo in case an inputs points to it

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
azarovh: I can make it Fn but I don't see how to get rid it completely. The function is pretty simple as it is because it's a single match for output type.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -0,0 +1,490 @@
...
+        &mut self,
+        chain_config: &ChainConfig,
+        block_height: BlockHeight,
+        pos_accounting_view: &impl PoSAccountingView,
+        output: &TxOutput,
+    ) -> Result<(), ConnectTransactionError> {
+        match output {
+            TxOutput::Transfer(value, _)
+            | TxOutput::LockThenTransfer(value, _, _)
```
TheQuantumPhysicist, line 69: Once we want to use this struct for fees, keep in mind that we have to do arithmetic with timelock values to be able to spend anything.

azarovh: but we already have a separate mod `timelock_check` for that

TheQuantumPhysicist: But we won't be able to calculate the valid timelocks if we don't do that. When the time comes we'll discuss it.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
TheQuantumPhysicist: You can pull the utxo then pass it to the function.

chainstate/test-suite/src/tests/double_spend_tests.rs
azarovh: change the order of call so that we got prev error first

TheQuantumPhysicist (approved):
This is looking fine. Please allow others to review it too.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -0,0 +1,519 @@
...
+/// `ConstrainedValueAccumulator` helps avoiding messy inputs/outputs combinations analysis by
+/// providing a set of properties that should be satisfied. For example instead of checking that
+/// all outputs are timelocked when the pool is decommissioned `ConstrainedValueAccumulator` gives a way
+/// to check that an accumulated output value is locked for sufficient amount of time which allows
+/// using other valid inputs and outputs in the same tx.
+///
+/// TODO: this struct can be extended to collect tokens replacing `AmountsMap`
+pub struct ConstrainedValueAccumulator {
+    unconstrained_value: Amount,
+    timelock_constrained: BTreeMap<BlockDistance, Amount>,
```
TheQuantumPhysicist, line 35: We should probably create a NonZeroBlockDistance for this. But maybe later.

azarovh: I don't think that we should treat BlockDistance(0) as a separate case. It cannot satisfy constrained value so we just subtract the value from unconstrained. The same way we would subtract BlockDistance(1) or any other value less then required by maturity.

TheQuantumPhysicist: Later when we apply this to fees it'll become apparent why this is necessary.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -0,0 +1,519 @@
...
+                            match self.timelock_constrained.entry(block_distance) {
+                                Entry::Vacant(e) => {
+                                    e.insert(*spend_amount);
+                                }
+                                Entry::Occupied(mut e) => {
+                                    let new_balance = (*e.get() + *spend_amount)
+                                        .ok_or(IOPolicyError::AmountOverflow)?;
+                                    *e.get_mut() = new_balance;
+                                }
+                            };
```
iljakuklic, line 128: I think this could be simplified with something like:
```rust
    let mut balance = self.timelock_constrained.entry(distance).or_insert(0);
    *balance = (*balance + *amount).ok_or(todo!())?;
```

(one more occurrence of this pattern above)

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/tests/purpose_tests.rs
```diff
@@ -0,0 +1,423 @@
...
+        BTreeMap::from_iter([(
+            outpoint.clone(),
+            Utxo::new_for_mempool(input_utxo),
+        )]),
+    );
+
+    let block = make_block(vec![outpoint.into()], vec![output]);
+
+    assert_eq!(result.map_err(|e| ConnectTransactionError::IOPolicyError(e, block.get_id().into())),
+               check_reward_inputs_outputs_purposes(&block.block_reward_transactable(), &utxo_db, block.get_id()));
```
iljakuklic, line 281: Weird formatting due to `rustfmt::skip`

iljakuklic (approved):
Cool stuff. Good to go as is. Adding some minor comments for completeness:

iljakuklic (approved):

Issue #1071, opened 2023-07-23
Block production should not produce blocks if chainstate is in initial block download mode


TheQuantumPhysicist:
Initial block download means that the chainstate is far from being synced. Block production should call the function ChainstateInterface::info() to know whether the chainstate is ready to have new blocks produced.

TheQuantumPhysicist:
@alfiedotwtf I believe this is done, right?

alfiedotwtf:
Yep. The commits are from 3665b4a383d to ee3b7271e6

PR #1074, opened 2023-07-24
Don't produce blocks until Chainstate is up to date

alfiedotwtf:
See #1071 for details.

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -260,6 +260,16 @@ impl BlockProduction {
         transactions_source: TransactionsSource,
         custom_id: Option<Vec<u8>>,
     ) -> Result<(Block, oneshot::Receiver<usize>), BlockProductionError> {
+        let is_initial_block_download = self
+            .chainstate_handle
+            .call(|this| this.info().map(|info| info.is_initial_block_download))
+            .await?
+            .map_err(|_| BlockProductionError::ChainstateInfoRetrievalError)?;
+
+        if is_initial_block_download {
+          return Err(BlockProductionError::ChainstateWaitForSync);
+        }
+
         let current_peer_count = self
             .p2p_handle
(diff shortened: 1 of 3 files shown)


blockprod/src/detail/mod.rs
```diff
@@ -260,6 +260,16 @@ impl BlockProduction {
...
+        let is_initial_block_download = self
+            .chainstate_handle
+            .call(|this| this.info().map(|info| info.is_initial_block_download))
+            .await?
+            .map_err(|_| BlockProductionError::ChainstateInfoRetrievalError)?;
+
+        if is_initial_block_download {
+            return Err(BlockProductionError::ChainstateWaitForSync);
+        }
+
```
alfiedotwtf, line 272: I think this change now puts us in a Catch-22 - blockprod tests fail because we start out at Genesis, but we can't move forward because of line 269.

I've been looking at possible solutions, but they're all hacks which I don't want to do (plus they won't work for functional tests).

Any suggestions?

alfiedotwtf: Thinking the further, I guess we can gate that `if` above with a `BlockProdConfig` param, then bypass it within unit and functional tests?

iljakuklic: Having a similar issue in mempool, which is also supposed to be disabled during IBD. Tried to fiddle with the `MaxTipAge` setting in tests but if the chain is at genesis, it is considered in the IBD stage no matter what. Maybe we could make it respect max tip age even at genesis?

TheQuantumPhysicist:
Done through gitlab.

alfiedotwtf:
Sorry... reopening as this hasn't been finished. I'll add the comments from Gitlab and can add this back on my TODO list in time.

alfiedotwtf:
Yep. The commits are from 3665b4a to ee3b7271e6

PR #1078, opened 2023-07-24
Skip scanning the blockchain for new wallets

OBorce:
When a new wallet is created, we assume there are no used addresses and the scanning of the blockchain can be skipped.

This PR also fixes the issue of accounts with new transactions not being detected and added automatically. Meaning: While account 0 is scanning, if account 1 is found to have a relevant transaction, account 1 should be added and automatically scanned too, and account 2 should start scanning too. If account 2 has transactions, then we do the same and move on to account 3. This behavior wasn't previously implemented.

### DIFF
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -777,6 +777,17 @@ impl Account {
         Ok(())
     }
 
+    pub fn sync_best_block<B: storage::Backend>(
+        &mut self,
+        db_tx: &mut StoreTxRw<B>,
+        best_block_height: BlockHeight,
+        best_block_id: Id<GenBlock>,
+    ) -> WalletResult<()> {
+        self.account_info.update_best_block(best_block_height, best_block_id);
+        db_tx.set_account(&self.key_chain.get_account_id(), &self.account_info)?;
+        Ok(())
+    }
+
     pub fn scan_new_unconfirmed_transactions(
         &mut self,
         transactions: &[SignedTransaction],
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -738,7 +738,7 @@ impl Account {
         wallet_events: &mut impl WalletEvents,
         common_block_height: BlockHeight,
         blocks: &[Block],
-    ) -> WalletResult<()> {
+    ) -> WalletResult<bool> {
         assert!(!blocks.is_empty());
         assert!(
             common_block_height <= self.account_info.best_block_height(),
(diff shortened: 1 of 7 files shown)


TheQuantumPhysicist (approved):
Please make sure this works fine when restoring wallets. Meaning: Restored wallets will rescan the blockchain.

Usually the user is asked for the date, from which the scan should start, when a wallet is restored. While you don't have to implement that right now, please make sure it starts from 0 at least.

muursh (approved):

wallet/src/account/mod.rs
```diff
@@ -738,7 +738,7 @@ impl Account {
         wallet_events: &mut impl WalletEvents,
         common_block_height: BlockHeight,
         blocks: &[Block],
-    ) -> WalletResult<()> {
+    ) -> WalletResult<bool> {
```
TheQuantumPhysicist, line 741: Please explain in documentation what this does

wallet/src/account/mod.rs
```diff
@@ -751,19 +751,23 @@ impl Account {
             self.reset_to_height(db_tx, wallet_events, common_block_height)?;
         }
 
+        let mut new_tx_was_added = false;
```
TheQuantumPhysicist, line 754: This is really ugly... can we do some mapping to make this more functional?

wallet/src/account/mod.rs
```diff
@@ -774,6 +778,17 @@ impl Account {
         self.account_info.update_best_block(best_block_height, best_block_id);
         db_tx.set_account(&self.key_chain.get_account_id(), &self.account_info)?;
 
+        Ok(new_tx_was_added)
+    }
+
+    pub fn sync_best_block(
```
TheQuantumPhysicist, line 784: Please document this function.

I have a feeling this function name is not really indicative to what this function is doing... after you document it let's talk again about it.

wallet/src/wallet/mod.rs
```diff
@@ -222,6 +300,8 @@ impl<B: storage::Backend> Wallet<B> {
             accounts,
             latest_median_time,
             unsynced_accounts,
+            state: WalletState::Syncing,
+            next_unused_account,
```
TheQuantumPhysicist, line 304: Can you please explain why we need to track this? What has changed in this PR that necessitates it?

wallet/src/wallet/mod.rs
```diff
@@ -277,31 +377,35 @@ impl<B: storage::Backend> Wallet<B> {
...
             &mut db_tx,
-            account_key_chain,
             name.clone(),
         )?;
 
+        std::mem::swap(&mut self.next_unused_account, &mut next_unused_account);
+        let (next_account_index, next_account) = next_unused_account;
+
+        // no need to rescan the blockchain from the start for the next unused account as we have been
+        // scaning for addresses of the previous next unused account and it is not alowed to create a gap in
```
TheQuantumPhysicist, line 394: scanning
allowed

TheQuantumPhysicist: two typos

wallet/src/wallet/tests.rs
```diff
@@ -1474,6 +1488,28 @@ fn wallet_sync_new_account(#[case] seed: Seed) {
     assert_ne!(new_account_index, DEFAULT_ACCOUNT_INDEX);
     assert_eq!(wallet.number_of_accounts(), 2);
 
+    // as the new accout will be in sync by default we need to make it out of sync
```
TheQuantumPhysicist, line 1491: account

TheQuantumPhysicist (changes requested):
I have to say... all these changes just to avoid scanning from zero? Too much! We should have a discussion about this. Though this PR isn't a priority, to be clear.

TheQuantumPhysicist:
Done through gitlab.

Issue #1005, opened 2023-06-28
Don't stake for previously staked times


alfiedotwtf:
Currently, each call to `produce_block()` will stake from the "last block timestamp" to 'last block timestamp + some chain configuration" value...

In the `produce_block()` call, if there is ever a recoverable error, we re-attempt to stake until we hit that maximum time limit. If still no successful block has been solved, we return with a `TryAgainLater`. This is fine for a single call to `produce_block()` as it prevents staking over timeframes that we have already attempted, but this currently doesn't prevent rework on subsequent calls.

This fix will make every call to `produce_block()` restrict the search space to only timestamps we have not yet searched.

PR #1083, opened 2023-07-26
Don't stake for previously staked times

alfiedotwtf:
See #1005 for more details.

### DIFF
--- a/blockprod/src/detail/job_manager/jobs_container.rs
+++ b/blockprod/src/detail/job_manager/jobs_container.rs
@@ -71,9 +71,7 @@ impl JobsContainer {
             .collect();
 
         for job_key in jobs_to_stop {
-            if let Some(handle) = self.jobs.remove(&job_key) {
-                _ = handle.cancel_sender.send(())
-            }
+            self.remove_job(job_key, true);
         }
     }
 
@@ -101,13 +99,16 @@ impl JobsContainer {
         let taken_jobs = std::mem::take(&mut self.jobs);
         let count = taken_jobs.len();
         let stop_results = taken_jobs
-            .into_values()
-            .map(|job_handle| job_handle.cancel_sender.send(()))
+            .into_keys()
+            .map(|job_key| self.remove_job(job_key, true))
+            .filter(|stopped| *stopped)
             .collect::<Vec<_>>();
 
-        let send_fail_count = stop_results.into_iter().filter_map(|r| r.ok()).count();
-        if send_fail_count > 0 {
-            log::info!("Sending stop jobs for block production failed for {send_fail_count}");
+        if !stop_results.is_empty() {
+            log::info!(
+                "Sending stop jobs for block production failed for {}",
+                stop_results.len()
+            );
         }
 
         // We don't do `count - send_fail_count` because the failures
--- a/blockprod/src/detail/job_manager/jobs_container.rs
+++ b/blockprod/src/detail/job_manager/jobs_container.rs
@@ -36,18 +36,13 @@ impl JobsContainer {
 
     pub fn handle_add_job(&mut self, event: NewJobEvent) {
         let NewJobEvent {
-            mut custom_id,
+            custom_id,
             current_tip_id,
             cancel_sender,
             result_sender,
         } = event;
 
-        if custom_id.is_none() {
-            let mut rng = make_true_rng();
-            custom_id = Some(rng.gen::<[u8; JOBKEY_DEFAULT_LEN]>().into())
-        }
-
-        let job_key = JobKey::new(custom_id, current_tip_id);
+        let job_key = JobKey::new(custom_id.clone(), current_tip_id);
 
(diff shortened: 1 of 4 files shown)


TheQuantumPhysicist (approved):
It's not perfect because it doesn't take into account different consensus algorithms, but I can live with this now.

However, please postpone merging this to after the release. I want this properly tested before it goes to the testnet.

blockprod/src/detail/tests.rs
```diff
@@ -851,16 +924,23 @@ mod produce_block {
                 mock_job_manager.expect_add_job().times(1).returning(move |_, _| {
                     let (_, cancel_receiver) = unbounded_channel::<()>();
                     let mut rng = make_seedable_rng(seed);
-                    let job_key =
-                        JobKey::new(None, Id::new(H256::random_using(&mut rng)) as Id<GenBlock>);
-                    Ok((job_key, cancel_receiver))
+                    let job_key = JobKey::new(
+                        vec![],
+                        Id::new(H256::random_using(&mut rng)) as Id<GenBlock>,
```
iljakuklic, line 929: Interesting construct with the `as` conversion. If I understand this correctly, I think using `Id::<GenBlock>::new(...)` would be more idiomatic and also easier to understand. NO conversion is actually happening despite `as` suggesting so. Did I miss something? This is replicated a number of times.

blockprod/src/detail/tests.rs
```diff
@@ -474,6 +474,79 @@ mod produce_block {
...
+                        vec![create_genesis_pool_txoutput],
+                    )));
+
+                _ = block_production
+                    .job_manager_handle
+                    .update_last_used_block_timestamp(
+                        block_production.generate_custom_id(&input_data),
+                        BlockTimestamp::from_int_seconds(u64::MAX),
+                    )
+                    .await;
```
iljakuklic, line 529: What is `_ =`? Would `let _ =` also work? If so I'd consider it preferable.

iljakuklic (commented):
Looks good. Some minor coding suggestions below.

iljakuklic (approved):

blockprod/src/detail/tests.rs
alfiedotwtf: Sorry, that's a carryover from Golang. Fixed

blockprod/src/detail/tests.rs
alfiedotwtf: Yep you're right - there's no casting done here, I just prefer it out of the way... but yeah, fixing because it's not idiomatic.

muursh (approved):

blockprod/src/detail/job_manager/jobs_container.rs
```diff
@@ -101,13 +101,16 @@ impl JobsContainer {
         let taken_jobs = std::mem::take(&mut self.jobs);
         let count = taken_jobs.len();
         let stop_results = taken_jobs
-            .into_values()
-            .map(|job_handle| job_handle.cancel_sender.send(()))
+            .into_keys()
+            .map(|job_key| self.remove_job(job_key, true))
+            .filter(|stopped| *stopped)
             .collect::<Vec<_>>();
```
azarovh, line 107: you don't actually need to collect, count is enough

blockprod/src/detail/mod.rs
```diff
@@ -242,6 +262,18 @@ impl BlockProduction {
         Ok(best_block_index)
     }
 
+    pub fn generate_custom_id(&self, input_data: &GenerateBlockInputData) -> CustomId {
+        match input_data {
+            GenerateBlockInputData::PoS(pos_input_data) => {
+                pos_input_data.stake_public_key().encode()
```
azarovh, line 268: not sure how important that is but it's confusing that PoS returns the same value every time. Given that struct is called "generate" and for PoW it actually generates random value. Is it intended?

azarovh (approved):

blockprod/src/detail/job_manager/mod.rs
```diff
@@ -44,6 +48,10 @@ pub enum JobManagerError {
     FailedToStopJobs,
     #[error("Error sending stop job event")]
     FailedToSendStopJobEvent,
+    #[error("Error sending update last timestamp secconds used event")]
+    FailedToSendUpdateLastTimestampSecondsUsed,
+    #[error("Error updating last timestamp secconds used")]
```
TheQuantumPhysicist, line 53: secconds

TheQuantumPhysicist (approved):
I'm approving, but please wait for others to provide a second optinion after my changes.

blockprod/src/detail/mod.rs
alfiedotwtf: Yeah. So the function "generates" for PoW, but for PoS the value is fixed. The struct has a "Generate" prefix because it's the struct for the "generate_block()" RPC call.

iljakuklic (approved):

Issue #1084, opened 2023-07-26
Processing blocks in the wallet is very slow


TheQuantumPhysicist:
In the GUI wallet, we can see that we're processing around 5 blocks/second, which is very slow. This can definitely be improved. Perhaps this can be achieved by batching the retrieval of blocks (despite me remembering that there's some kind of batching in there, but this is to be studied when its time comes).

PR #1086, opened 2023-07-27
Speed up wallet scan

pavel-kokolemin:
The problem with the slow wallet scan is that we download one block at a time and it takes many minutes to scan a new wallet. I added a new chainstate RPC function that returns a vector of mainchain blocks. With this change, the full wallet scan takes about 5 seconds for me. This should fix https://github.com/mintlayer/mintlayer-core/issues/1084

After this change, I got a problem with unsyncing accounts where the wallet would not correctly move an unsynced account to regular accounts. I decided not to spend time fixing it and removed the unsyncing accounts. Now all accounts are treated the same, which simplified the code.


OBorce (approved):
The syncing of the accounts looks fine to me.

PR #1090, opened 2023-07-27
mempool: Time lock handling

iljakuklic:
* Allow time locked transactions in mempool (within certain limit).
* Modifications to tx verifier to support time locked instructions in mempool.
* Transaction accumulator to only allow transactions where time lock constraints are satisfied.

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -119,6 +119,7 @@ impl BlockProduction {
     pub async fn collect_transactions(
         &self,
         current_tip: Id<GenBlock>,
+        min_block_timestamp: BlockTimestamp,
     ) -> Result<Option<Box<dyn TransactionAccumulator>>, BlockProductionError> {
         let max_block_size = self.chain_config.max_block_size_from_std_scripts();
         let returned_accumulator = self
@@ -127,6 +128,7 @@ impl BlockProduction {
                 mempool.collect_txs(Box::new(DefaultTxAccumulator::new(
                     max_block_size,
                     current_tip,
+                    min_block_timestamp,
                 )))
             })
             .await?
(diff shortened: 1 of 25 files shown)


blockprod/src/detail/mod.rs
```diff
@@ -305,17 +307,14 @@ impl BlockProduction {
...
-
-            current_timestamp
-                .add_int_seconds(self.chain_config.max_future_block_time_offset().as_secs())
-                .ok_or(ConsensusCreationError::TimestampOverflow(
-                    current_timestamp,
-                    self.chain_config.max_future_block_time_offset().as_secs(),
-                ))?
-        };
+        let min_block_timestamp =
+            BlockTimestamp::from_duration_since_epoch(self.time_getter().get_time());
```
TheQuantumPhysicist, line 311: This is not really correct. The minimum timestamp value is the previous block's timestamp + 1.

TheQuantumPhysicist: Unless I'm misunderstanding the meaning of this...

mempool/src/pool/tests/accumulator.rs
```diff
@@ -154,14 +169,148 @@ async fn collect_transactions(#[case] seed: Seed) -> anyhow::Result<()> {
...
+#[case::for_21s(Seed::from_entropy(), OutputTimeLock::ForSeconds(21), 0b0010)]
+#[trace]
+#[case::for_500s(Seed::from_entropy(), OutputTimeLock::ForSeconds(500), 0b0000)]
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn timelocked(#[case] seed: Seed, #[case] timelock: OutputTimeLock, #[case] expected: u32) {
+    // Unpack expected results:
+    let in_mempool_at0 = (expected & 0b1000) != 0;
+    let in_accumulator_at0 = (expected & 0b0100) != 0;
+    let in_mempool_at1 = (expected & 0b0010) != 0;
+    let in_accumulator_at1 = (expected & 0b0001) != 0;
```
TheQuantumPhysicist, line 233: I can live with those being for tests... but I wish we used some proper struct for readability.

azarovh (approved):

TheQuantumPhysicist (approved):
The only issue I see is the timestamps in block production. I'm not sure that's the right way to do this. Let's have a discussion about it. Besides that, looks good!

mempool/src/pool/tests/accumulator.rs
iljakuklic: Yea, I did not want the `#[case]` things to become too verbose. And quite a few of these combinations are legitimate outcomes, not easy to compress into a couple of enum variants.

blockprod/src/detail/mod.rs
alfiedotwtf: I've just gone through the whole change now. I see... I think the name `min_block_timestamp` is wrong, and should be named something like `min_verification_block_timestamp` so it doesn't get confused with the "minimum possible block timestamp"

alfiedotwtf (approved):

blockprod/src/detail/mod.rs
iljakuklic: Yea it's a bit subtle. Basically it's the timestamp where the search for a block starts. Set it too low and node operator will miss out on fees because mempool will not include some locked transactions. Set it to high and it will not be correct. An alternative is to construct the transaction sequence from the scratch every time block production attempts to construct a block with a different timestamp. Will update the name and put a comment explaining it.

TheQuantumPhysicist: Well, the timestamps are decided before any transactions are added to a block. So there's no way really to optimize this, assuming that the block reward is going to be higher than any transaction fee. Please ignore the game-theoretic aspect of optimizing locked outputs for fees for timestamps because the timestamps' time slots are mostly empty during normal operation with maximum probability of 1/120 per total, assuming the user has 100% of the stake.

iljakuklic:
Added a functional test

blockprod/src/detail/mod.rs
iljakuklic: Added some explanation comments to this bit.

Issue #1053, opened 2023-07-17
Provide an error when decoding fails due to database schema changes, instead of crashing


TheQuantumPhysicist:
Currently, changing the database schema (both in the wallet and blockchain) lead to a crash if the database is used after versioning (assuming versioning wasn't done properly):

```
thread 'tokio-runtime-worker' panicked at 'assertion failed: T::decode_all(&mut this.bytes()).is_ok()', serialization/src/encoded.rs:44:9
```

This is because of this assert: https://github.com/mintlayer/mintlayer-core/blob/3d0fd11adaaacf770f7090076d8b413e8ae58005/serialization/src/encoded.rs#L44

Two things have to be done:

1. This function must be marked unsafe, because it shouldn't be used under normal circumstances, and the risk must be understood (and documented).
2. There should be a variant that returns an error for this function. This is because we need to be able to display an error when decoding fails.

Note: Despite this error having a Testnet milestone, it can be done a little later because we assume we'll handle versioning properly.

iljakuklic:
While a friendlier error message could be provided, I would consider this to be a part of the larger task of introducing a mechanism for database schema upgrades.

Issue #1013, opened 2023-06-29
Block Production improvement tasks


alfiedotwtf:
A list of improvements to be done for Block Production:

- [x] #1004
- [x] #1003 
- [x] #1060 
- [x] Prevent staking at the same time with the same key (this should be currently prevented by the job keys)
- [x] #1002 
- [x] #756
- [x] #1005 
- [x] #957
- [x] #992
- [x] #991
- [x] calculate coins from atoms instead of strings full of zeros

PR #1092, opened 2023-07-28
Create command line arguments to set genesis staking keys, VRF keys, and pool Id

alfiedotwtf:
This is the last item for #1004

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -159,7 +159,6 @@ quote = "1.0"
 rand = "0.8"
 rand_chacha = "0.3"
 reedline = "0.22"
-regex = "1.9"
 replace_with = "0.1"
 rfd = { version = "0.11", default-features = false }
 ripemd = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -159,6 +159,7 @@ quote = "1.0"
 rand = "0.8"
 rand_chacha = "0.3"
 reedline = "0.22"
+regex = "1.9"
 replace_with = "0.1"
 rfd = { version = "0.11", default-features = false }
 ripemd = "0.1"
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -599,7 +599,7 @@ fn create_testnet_genesis() -> Genesis {
 }
 
 #[derive(Clone, Debug)]
-pub struct GenesisStakingSettings (String);
+pub struct GenesisStakingSettings(String);
 
 impl GenesisStakingSettings {
     pub fn new(settings: String) -> Self {
@@ -607,7 +607,7 @@ impl GenesisStakingSettings {
     }
 
     pub fn get_settings(&self) -> Vec<&str> {
-        self.0.split(",").collect()
+        self.0.split(',').collect()
     }
 }
 
@@ -643,7 +643,7 @@ pub fn regtest_genesis_values(
         .or(Some(
             "123c4c600097c513e088b9be62069f0c74c7671c523c8e3469a1c3f14b7ea2c4",
         ))
-        .map(|s| decode_hex::<PoolId>(s))
+        .map(decode_hex::<PoolId>)
         .expect("Pool Id decoded");
 
     let genesis_stake_private_key = genesis_staking_settings_parts
@@ -652,7 +652,7 @@ pub fn regtest_genesis_values(
         .or(Some(
             "008717e6946febd3a33ccdc3f3a27629ec80c33461c33a0fc56b4836fcedd26638",
         ))
-        .map(|s| decode_hex::<PrivateKey>(s))
+        .map(decode_hex::<PrivateKey>)
         .expect("Private key decoded");
 
     let genesis_vrf_private_key = genesis_staking_settings_parts
@@ -661,7 +661,7 @@ pub fn regtest_genesis_values(
         .or(Some(
             "003fcf7b813bec2a293f574b842988895278b396dd72471de2583b242097a59f06e9f3cd7b78d45750afd17292031373fddb5e7a8090db51221038f5e05f29998e",
         ))
-        .map(|s| decode_hex::<VRFPrivateKey>(s))
+        .map(decode_hex::<VRFPrivateKey>)
         .expect("VRF private key decoded");
 
     let genesis_stake_public_key = PublicKey::from_private_key(&genesis_stake_private_key);
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -16,6 +16,7 @@
 mod builder;
 mod checkpoints;
 pub mod emission_schedule;
+pub mod regtest;
 pub use builder::Builder;
 use crypto::key::PublicKey;
 use crypto::vrf::VRFPublicKey;
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -36,8 +36,8 @@ use crate::primitives::semver::SemVer;
 use crate::primitives::{Amount, BlockDistance, BlockHeight, H256};
 use crypto::key::hdkd::{child_number::ChildNumber, u31::U31};
 use crypto::{key::PrivateKey, vrf::VRFPrivateKey};
-use regex::Regex;
 use std::num::NonZeroU64;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -598,8 +598,35 @@ fn create_testnet_genesis() -> Genesis {
     )
 }
 
+#[derive(Clone, Debug)]
+pub struct GenesisStakingSettings (String);
+
+impl GenesisStakingSettings {
+    pub fn new(settings: String) -> Self {
+        Self(settings)
+    }
+
+    pub fn get_settings(&self) -> Vec<&str> {
+        self.0.split(",").collect()
+    }
+}
+
+impl Default for GenesisStakingSettings {
+    fn default() -> Self {
+        Self::new("".to_string())
+    }
+}
+
+impl FromStr for GenesisStakingSettings {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(Self::new(s.to_string()))
+    }
+}
+
 pub fn regtest_genesis_values(
-    genesis_staking_settings: Option<String>,
+    genesis_staking_settings: GenesisStakingSettings,
 ) -> (
     PoolId,
     Box<StakePoolData>,
@@ -608,38 +635,34 @@ pub fn regtest_genesis_values(
     VRFPrivateKey,
     VRFPublicKey,
 ) {
-    let genesis_staking_settings = genesis_staking_settings.unwrap_or_default();
+    let genesis_staking_settings_parts = genesis_staking_settings.get_settings();
 
-    let genesis_pool_id = Regex::new("genesis_pool_id:([A-Fa-f0-9]+)")
-        .expect("Regex is ok")
-        .captures(&genesis_staking_settings)
-        .and_then(|c| c.get(1))
-        .map(|c| c.as_str())
+    let genesis_pool_id = genesis_staking_settings_parts
+        .iter()
+        .find_map(|s| s.strip_prefix("genesis_pool_id:"))
         .or(Some(
             "123c4c600097c513e088b9be62069f0c74c7671c523c8e3469a1c3f14b7ea2c4",
         ))
-        .map(decode_hex::<PoolId>)
-        .expect("Parameter decoded {}");
+        .map(|s| decode_hex::<PoolId>(s))
+        .expect("Pool Id decoded");
 
-    let genesis_stake_private_key = Regex::new("genesis_stake_private_key:([A-Fa-f0-9]+)")
-        .expect("Regex is ok")
-        .captures(&genesis_staking_settings)
-        .and_then(|c| c.get(1))
-        .map(|c| c.as_str())
+    let genesis_stake_private_key = genesis_staking_settings_parts
+        .iter()
+        .find_map(|s| s.strip_prefix("genesis_stake_private_key:"))
         .or(Some(
             "008717e6946febd3a33ccdc3f3a27629ec80c33461c33a0fc56b4836fcedd26638",
         ))
-        .map(decode_hex::<PrivateKey>)
-        .expect("Parameter decoded {}");
+        .map(|s| decode_hex::<PrivateKey>(s))
+        .expect("Private key decoded");
 
-    let genesis_vrf_private_key = Regex::new("genesis_vrf_private_key:([A-Fa-f0-9]+)")
-        .expect("Regex is ok")
-        .captures(&genesis_staking_settings)
-        .and_then(|c| c.get(1))
-        .map(|c| c.as_str())
-        .or(Some("003fcf7b813bec2a293f574b842988895278b396dd72471de2583b242097a59f06e9f3cd7b78d45750afd17292031373fddb5e7a8090db51221038f5e05f29998e"))
-        .map(decode_hex::<VRFPrivateKey>)
-        .expect("Parameter decoded {}");
+    let genesis_vrf_private_key = genesis_staking_settings_parts
+        .iter()
+        .find_map(|s| s.strip_prefix("genesis_vrf_private_key:"))
+        .or(Some(
+            "003fcf7b813bec2a293f574b842988895278b396dd72471de2583b242097a59f06e9f3cd7b78d45750afd17292031373fddb5e7a8090db51221038f5e05f29998e",
+        ))
+        .map(|s| decode_hex::<VRFPrivateKey>(s))
+        .expect("VRF private key decoded");
 
     let genesis_stake_public_key = PublicKey::from_private_key(&genesis_stake_private_key);
     let genesis_vrf_public_key = VRFPublicKey::from_private_key(&genesis_vrf_private_key);
@@ -664,7 +687,7 @@ pub fn regtest_genesis_values(
 }
 
 pub fn create_regtest_pos_genesis(
-    genesis_staking_settings: Option<String>,
+    genesis_staking_settings: GenesisStakingSettings,
     premine_destination: Destination,
 ) -> Genesis {
     let (
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -36,6 +36,7 @@ use crate::primitives::semver::SemVer;
 use crate::primitives::{Amount, BlockDistance, BlockHeight, H256};
 use crypto::key::hdkd::{child_number::ChildNumber, u31::U31};
 use crypto::{key::PrivateKey, vrf::VRFPrivateKey};
+use regex::Regex;
 use std::num::NonZeroU64;
 use std::sync::Arc;
 use std::time::Duration;
@@ -597,7 +598,9 @@ fn create_testnet_genesis() -> Genesis {
     )
 }
 
-pub fn regtest_genesis_values() -> (
+pub fn regtest_genesis_values(
+    genesis_staking_settings: Option<String>,
+) -> (
     PoolId,
     Box<StakePoolData>,
     PrivateKey,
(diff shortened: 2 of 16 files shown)


common/src/chain/config/mod.rs
```diff
@@ -567,30 +568,51 @@ fn create_testnet_genesis() -> Genesis {
...
+    let genesis_pool_id = Regex::new("genesis_pool_id:([A-Fa-f0-9]+)")
+        .expect("Regex is ok")
+        .captures(&genesis_staking_settings)
+        .and_then(|c| c.get(1))
+        .map(|c| c.as_str())
+        .or(Some(
+            "123c4c600097c513e088b9be62069f0c74c7671c523c8e3469a1c3f14b7ea2c4",
+        ))
+        .map(decode_hex::<PoolId>)
+        .expect("Parameter decoded {}");
```
iljakuklic, line 592: I find it a bit overkill to use a regex for this. How about just splitting the string on whitespace/comma and using `strip_prefix` to parse key-value entries, roughly like:
```rust
if let Some(pool_id_str) = chunk.strip_prefix("genesis_pool_id:") {
    pool_id = decode_hex(pool_id_str)?;
}
// Same for other settings
```

node-lib/src/regtest_options.rs
```diff
@@ -61,4 +61,8 @@ pub struct ChainConfigOptions {
     /// PoS NetUpgrade override after Genesis
     #[clap(long)]
     pub chain_pos_netupgrades: Option<bool>,
+
+    /// PoS Genesis staking settings
+    #[clap(long)]
+    pub chain_genesis_staking_settings: Option<String>,
```
iljakuklic, line 67: I believe that if you create a type for the settings, like `GenesisSTakingSettings` and use it here, it will work as long as you provide `FromStr`. May be wrong though.

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
```diff
@@ -0,0 +1,428 @@
...
+                        "PublicKey": new_stake_public_key,
+                    },
+                    new_pool_id,
+                ],
+            }
+
+if __name__ == '__main__':
+    # Use the first instance to bootstrap new staking keys
+    GenerateGenesisKeys().main(exit_on_success=False)
+    GeneratePoSBlocksTest().main()
```
iljakuklic, line 428: Could you please explain what's going on here? Looks like a hack :laughing:

common/src/chain/config/mod.rs
```diff
@@ -567,30 +568,51 @@ fn create_testnet_genesis() -> Genesis {
     )
 }
 
-pub fn regtest_genesis_values() -> (
+pub fn regtest_genesis_values(
+    genesis_staking_settings: Option<String>,
+) -> (
```
iljakuklic, line 573: Would be nicer for this to take some sort of setting type with `Default` impl to avoid the weird special casing for `None`

iljakuklic (commented):
A couple of initial observations:

common/src/chain/config/mod.rs
alfiedotwtf: I'm a regex fan, and it's only called once for the whole process, but no problem.

node-lib/src/regtest_options.rs
alfiedotwtf: Nice... I didn't know you could create your own types. Done

common/src/chain/config/mod.rs
alfiedotwtf: Done

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
alfiedotwtf: Hack, magic... it's all perspective 😂

common/src/chain/config/mod.rs
```diff
@@ -587,30 +588,74 @@ fn create_testnet_genesis() -> Genesis {
     )
 }
 
-pub fn regtest_genesis_values() -> (
+#[derive(Clone, Debug)]
+pub struct GenesisStakingSettings(String);
```
iljakuklic, line 592: What I meant is something that actually stores the settings in a struct, not just a mere wrapper over a string. The `FromStr` parses the settings string into the struct, including all the error checking and such. The command line parser will then be able to report format errors directly.

node-lib/src/regtest_options.rs
```diff
@@ -61,4 +62,8 @@ pub struct ChainConfigOptions {
     /// PoS NetUpgrade override after Genesis
     #[clap(long)]
     pub chain_pos_netupgrades: Option<bool>,
+
+    /// PoS Genesis staking settings
+    #[clap(long)]
+    pub chain_genesis_staking_settings: Option<GenesisStakingSettings>,
```
iljakuklic, line 68: Not quite sure about it but you may be able to set something like `default_value = ""` in the `clap` attribute and get rid of the `Option` wrapper.

common/src/chain/config/mod.rs
```diff
@@ -587,30 +588,74 @@ fn create_testnet_genesis() -> Genesis {
...
+    }
+}
+
+impl FromStr for GenesisStakingSettings {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(Self::new(s.to_string()))
+    }
+}
```
iljakuklic, line 616: My other suggestion notwithstanding, when you have a method that cannot fail, the `std::convert::Infallible` (or equivalent) type should be used as the error type. If errors are possible, try to find or introduce an `enum` that reflects the possible error space.

common/src/chain/config/mod.rs
```diff
@@ -587,30 +588,74 @@ fn create_testnet_genesis() -> Genesis {
...
 }
 
-pub fn regtest_genesis_values() -> (
+#[derive(Clone, Debug)]
+pub struct GenesisStakingSettings(String);
+
+impl GenesisStakingSettings {
+    pub fn new(settings: String) -> Self {
+        Self(settings)
+    }
```
iljakuklic, line 597: If you parse the string in the constructor, you could take a `&str` rather than `String` to initialize the struct.

iljakuklic (commented):
Some more comments about input parsing below

azarovh (commented):
Agree with Lukas that it makes sense to make `GenesisStakingSettings` a proper struct with fields and parsing and decoding implemented in the constructor.

common/src/chain/config/mod.rs
alfiedotwtf: Ah, ok. This is done, but I've forgotten to handle the different errors... I'll update this now.

node-lib/src/regtest_options.rs
alfiedotwtf: Yep, updated with `default_value_t`.

common/src/chain/config/mod.rs
```diff
@@ -587,30 +588,74 @@ fn create_testnet_genesis() -> Genesis {
...
+
+    pub fn get_settings(&self) -> Vec<&str> {
+        self.0.split(',').collect()
+    }
```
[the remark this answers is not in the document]
alfiedotwtf: No longer needed.

common/src/chain/config/mod.rs
alfiedotwtf: As above, I updated to Infallible, but will change to an enum.

common/src/chain/config/mod.rs
alfiedotwtf: I keep forgetting this. Fixed.

node-lib/src/runner.rs
```diff
@@ -370,9 +371,14 @@ fn regtest_chain_config(options: &ChainConfigOptions) -> Result<ChainConfig> {
...
 
     if chain_pos_netupgrades.unwrap_or(false) {
-        builder = builder
-            .net_upgrades(NetUpgrades::regtest_with_pos())
-            .genesis_custom(create_regtest_pos_genesis(Destination::AnyoneCanSpend));
+        builder = builder.net_upgrades(NetUpgrades::regtest_with_pos()).genesis_custom(
+            create_regtest_pos_genesis(
+                chain_genesis_staking_settings
+                    .clone()
+                    .unwrap_or(GenesisStakingSettings::default()),
```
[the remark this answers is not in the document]
alfiedotwtf: Yeah, I should have used that... but it's no longer needed now that it's no longer `Option`.

common/src/chain/config/regtest.rs
```diff
@@ -0,0 +1,180 @@
...
+            .expect("Private key decoded");
+
+        let vrf_private_key = settings_parts
+            .iter()
+            .find_map(|s| s.strip_prefix("genesis_vrf_private_key:"))
+            .or(Some(
+                "003fcf7b813bec2a293f574b842988895278b396dd72471de2583b242097a59f06e9f3cd7b78d45750afd17292031373fddb5e7a8090db51221038f5e05f29998e",
+            ))
+            .map(decode_hex::<VRFPrivateKey>)
+            .expect("VRF private key decoded");
```
iljakuklic, line 73: I am a bit conflicted about the `.expect`s here. I know this is supposed to be used for testing only. However, it is still being parsed from user input from the command line, so some sort of nicer error reporting is probably preferable.

A related question is what to do about unrecognised keys. Let's say the user makes a typo `genesis_vrf_pirvate_key:BLAH`, then vrf key is silently going to be set to the default value instead of `BLAH`. That may be quite annoying to diagnose.

common/src/chain/config/regtest.rs
```diff
@@ -0,0 +1,180 @@
...
+    vrf_private_key: VRFPrivateKey,
+}
+
+impl GenesisStakingSettings {
+    pub fn new(settings: &str) -> Self {
+        let settings_parts: Vec<&str> = settings.split(',').map(|s| s.trim()).collect();
+
+        let pool_id = settings_parts
+            .iter()
+            .find_map(|s| s.strip_prefix("genesis_pool_id:"))
```
iljakuklic, line 50: Probably a matter of taste to some degree, but I find the `genesis_` prefix in the option names quite redundant. The type already has `Genesis` in the name. When taken from the command line, the option is called `--chain-genesis-staking-settings`. It seems to me dropping the word "genesis" here would not cause any ambiguities assuming you don't plan to generalise this in the future to include non-genesis stuff too.

common/src/chain/config/regtest.rs
alfiedotwtf: > I am a bit conflicted about the .expects here

Yep agreed. I've already removed these locally.

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
```diff
@@ -0,0 +1,438 @@
...
+                                "index": 1,
+                            },
+                        },
+                ],
+                 "kernel_input_utxo": [
+                    {
+                        "CreateStakePool": [
+                            genesis_pool_id,
+                            {
+                                "value": "4000000000000000",
```
TheQuantumPhysicist, line 234: Perhaps we should find a proper way to calculate coins from atoms instead this chaotic zeros' situation.

common/src/chain/config/regtest.rs
```diff
@@ -0,0 +1,180 @@
...
+            .or(Some(
+                "123c4c600097c513e088b9be62069f0c74c7671c523c8e3469a1c3f14b7ea2c4",
```
TheQuantumPhysicist, line 52: If these are default values, please move them to constants.

test-rpc-functions/src/rpc.rs
```diff
@@ -97,7 +99,8 @@ trait RpcTestFunctionsRpc {
 #[async_trait::async_trait]
 impl RpcTestFunctionsRpcServer for super::RpcTestFunctionsHandle {
     async fn genesis_pool_id(&self) -> rpc::Result<Option<String>> {
-        let (genesis_pool_id, genesis_stake_pool_data, _, _, _, _) = regtest_genesis_values();
+        let (genesis_pool_id, genesis_stake_pool_data, _, _, _, _) =
+            genesis_values(GenesisStakingSettings::new(""));
```
TheQuantumPhysicist, line 103: If this is the default value, it's preferable to use the Default::default() trait function to indicate what this really is.

node-lib/src/regtest_options.rs
```diff
@@ -61,4 +62,8 @@ pub struct ChainConfigOptions {
     /// PoS NetUpgrade override after Genesis
     #[clap(long)]
     pub chain_pos_netupgrades: Option<bool>,
+
+    /// PoS Genesis staking settings
+    #[clap(long, default_value_t = GenesisStakingSettings::new(""))]
```
TheQuantumPhysicist, line 67: default()

common/src/chain/config/regtest.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    let premine_output =
+        TxOutput::Transfer(OutputValue::Coin(DEFAULT_INITIAL_MINT), premine_destination);
+
+    Genesis::new(
+        String::new(),
+        BlockTimestamp::from_int_seconds(1639975460),
+        vec![premine_output, create_genesis_pool_txoutput],
+    )
+}
```
TheQuantumPhysicist, line 180: Some basic unit tests that explain the parsing would really help.

You don't have to test every corner case (please don't spend days creating them), but usually unit tests can be used by beginners to learn how a struct works.

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
```diff
@@ -0,0 +1,438 @@
...
+                    },
+                    new_pool_id,
+                ],
+            }
+
+if __name__ == '__main__':
+    # We need to spin up a node in order to use the "test-functions/"
+    # RPC endpoint so we can generate new genesis staking
+    # settings. However, to spin up a node, the framework will create
+    # a chaintate along with its own Genesis, defeating the purpose of
```
TheQuantumPhysicist, line 429: chaintate

TheQuantumPhysicist:
This is looking fine. I'll review again after addressing the comments.

common/src/chain/config/regtest.rs
alfiedotwtf: This was a tricky one...

`Display` is required for `Clap` for custom types but it doesn't really elaborate why. I thought it was just using this for user `--help` etc so I just did a `format!("{:?}...)`, but then everything broke. It turned out that when not provided on the command line, it uses `Display` to stringify `default()`, and then calls from_str() later on to parse the value.

So when passed on the command line, `GenesisStakingSettings::from_str(settings: &str)` would have `settings` set to whatever was taken from the command line, however when not passed on the command line, `settings` contained something like `{pool_id: Id<Pool>{0x123c4c600097c513e088b9be62069f0c74c7671c523c8e3469a1c3f14b7ea2c4}, stake_private_key: PrivateKey { key: Secp256k1Schnorr(Secp256k1PrivateKey { data: SecretKey(#0fd5243c90a0082f) }) },...`, all because my `format!()` in the `impl Display` wasn't "round tripping" the value as stated in the `Clap` documentation.

All good, it's working now.

common/src/chain/config/regtest.rs
alfiedotwtf: Done.

common/src/chain/config/regtest.rs
alfiedotwtf: Cool, done.

test-rpc-functions/src/rpc.rs
alfiedotwtf: Fixed.

node-lib/src/regtest_options.rs
alfiedotwtf: Turns out `default_value_t` doesn't need an actual parameter when there's a `default()`.

common/src/chain/config/regtest.rs
alfiedotwtf: No problem. Added.

`mod.rs`s' `decode_hex()` was breaking these tests because it panicks when decoding fails. Because I didn't want to change the upstream function, I've copied it into `regtest.rs` and made that return `Result`. If you want me to change upstreams to return a `Result` too and remove the duplicated code, let me know.

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
alfiedotwtf: Fixed.

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
alfiedotwtf: I've added it as an item to #1013

common/src/chain/config/regtest.rs
```diff
@@ -0,0 +1,430 @@
...
+
+impl FromStr for GenesisStakingSettings {
+    type Err = GenesisStakingSettingsInputErrors;
+
+    fn from_str(settings: &str) -> Result<Self, Self::Err> {
+        let mut settings_parts = settings
+            .split(',')
+            .filter_map(|s| s.split_once(':'))
+            .map(|(k, v)| (k.trim(), v.trim()))
+            .collect::<HashMap<&str, &str>>();
```
iljakuklic, line 116: The `filter_map` call will silently discard entries without a comma. Would it be better to report an error instead?

Also collecting into HashMap will handle duplicate keys in the "last one wins" fashion. Probably not a big deal but maybe consider making that into an error too?

common/src/chain/config/regtest.rs
```diff
@@ -0,0 +1,430 @@
...
+    key::{PrivateKey, PublicKey},
+    vrf::{VRFPrivateKey, VRFPublicKey},
+};
+use hex::FromHex;
+use serialization::hex::HexEncode;
+
+const GENESIS_POOL_ID: &str = "123c4c600097c513e088b9be62069f0c74c7671c523c8e3469a1c3f14b7ea2c4";
+const GENESIS_STAKE_PRIVATE_KEY: &str =
+    "008717e6946febd3a33ccdc3f3a27629ec80c33461c33a0fc56b4836fcedd26638";
+const GENESIS_VRF_PRIVATE_KEY: &str = "003fcf7b813bec2a293f574b842988895278b396dd72471de2583b242097a59f06e9f3cd7b78d45750afd17292031373fddb5e7a8090db51221038f5e05f29998e";
```
iljakuklic, line 42: Could these be proper types, like `PoolId`, `PrivateKey`, `VRFPrivateKey` instead of strings? You can use the `hex!` macro from the `hex` crate to enter hex strings and get a `[u8; N]`.

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
```diff
@@ -0,0 +1,438 @@
...
+                            },
+                            "index": 0,
+                        },
+                    }
+                ],
+                "outputs": [
+                    {
+                        "Transfer": [
+                            {
+                                "Coin": "10000000000000000",
```
iljakuklic, line 269: Nitpick: I think you can enter a number here instead of a string, and use the underscore syntax for better readability. Like `10_000_000_000_000_000`. Not sure it will fit into Python's numeric type but I suspect it uses some sort of bignum by default?

iljakuklic (approved):
Looks good. A couple more minor suggestions here:

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
TheQuantumPhysicist: While I don't disagree there, the problem is that the number of decimals isn't a multiple of three, so this won't help to make the number more understandable. We should use proper COIN factor, being 10^11 or whatever correct decimal count, then we have 10_000*COIN.

node-lib/src/runner.rs
```diff
@@ -333,6 +333,7 @@ fn regtest_chain_config(options: &ChainConfigOptions) -> Result<ChainConfig> {
         chain_max_block_size_with_standard_txs,
         chain_max_block_size_with_smart_contracts,
         chain_pos_netupgrades,
+        chain_genesis_staking_settings,
```
TheQuantumPhysicist, line 336: We should add guards such that if someone tries to pass staking settings, net upgrades, emission schedule, coin decimals, target spacing, and any other feature that's made for testing, in anything other than regtest, then we crash the software. This is important to prevent scammers from trying to get victims to create fake nodes somehow with custom settings and make them feel they got rich or something.

common/src/chain/config/regtest.rs
TheQuantumPhysicist: Are you talking about [#1053](https://github.com/mintlayer/mintlayer-core/issues/1053)?

TheQuantumPhysicist (approved):
Besides the comments, I think this is looking shiny enough.

common/src/chain/config/regtest.rs
alfiedotwtf: No, this was different. For example, in the `invalid_vrf_private_key()` test, because the input wasn't long enough nor hex, `mod.rs`s `decode_hex()` would panic and so my test wouldn't pass.

common/src/chain/config/regtest.rs
alfiedotwtf: I think it's best to keep them as hex strings, because this is the same format as what is set on the command line, and also to be consistent with the other networks (defined in `mod.rs`).

node-lib/src/runner.rs
alfiedotwtf: It looks like `node-lib/src/regtest_options.rs`, where all of these "dangerous" parameters are defined to be read from config, are only used in RegTest (from `node-lib/src/runner.rs`). They can also be set from code (i.e not the command line) using the Builder, but to play it safe I'll create an issue to check the network type when creating a new `ChainConfigOptions` instances

common/src/chain/config/regtest.rs
alfiedotwtf: The `split(',')` will return an array with the empty string, so will proceed through the chain, but you're right about the `filter_map()` on settings lacking `':'`. This is now fixed.

test/functional/blockprod_generate_pos_blocks_rand_genesis_keys.py
alfiedotwtf: Using `COIN` a 10^11, this is now fixed.

common/src/chain/config/regtest.rs
```diff
@@ -0,0 +1,430 @@
...
+            vrf_private_key,
+        }
+    }
+
+    pub fn pool_id(&self) -> PoolId {
+        self.pool_id
+    }
+
+    pub fn stake_private_key(&self) -> PrivateKey {
+        self.stake_private_key.clone()
```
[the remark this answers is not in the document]
alfiedotwtf: Thanks, I always forget to do this and opt for `clone()`. Fixed.

Cargo.toml
```diff
@@ -159,6 +159,7 @@ quote = "1.0"
 rand = "0.8"
 rand_chacha = "0.3"
 reedline = "0.22"
+regex = "1.9"
```
iljakuklic, line 162: Is this still used?

iljakuklic (approved):

Cargo.toml
alfiedotwtf: I removed it and it broke. Now that I've rebased I'll try it again

alfiedotwtf: Good pickup. It snuck back in because a few rebases failed. I'm surprised Clippy didn't complain. Fixed.

TheQuantumPhysicist: I have a huge problem with regex in security applications. It's the source of all evil.

node-lib/src/runner.rs
alfiedotwtf: Because of the way how our command line parsing works, if any of the regtest command line arguments are passed along to other networks, the cli will fail e.g `error: unexpected argument '--chain-pos-netupgrades' found` so I think this should already prevent non-regtest use.

However, https://github.com/mintlayer/mintlayer-core/pull/1092/commits/51789930a744537499582b420bd0994d5436ea03 adds a gate to the regtest options parsing, so at least now in code you must explicitly be passing in the `Command::RegTest`.

test/functional/blockprod_generate_pos_blocks.py
```diff
@@ -25,6 +25,8 @@
 
 import random, time
 
+COIN = 100_000_000_000
```
TheQuantumPhysicist, line 28: This should be a global definition somewhere. I bet there's already a definition somewhere for bitcoin. We shouldn't redefine it for every test.

node-lib/src/runner.rs
TheQuantumPhysicist: Looking good!

PR #1095, opened 2023-07-28
Remove the disconnected by user flag (p2p)

pavel-kokolemin:
There is another reason why the node might get isolated. We use a fairly small timeout before disconnecting peers for ignoring header/block requests (5 seconds). Sometimes busy nodes get disconnected for this reason. This is not a big problem in itself, but there is another bug when such disconnected nodes are flagged as "disconnected by user". This flag prevents reconnects and may look like the node is banned. Let's remove this flag completely, as it duplicates the functionality provided by the ban API.

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -661,12 +661,7 @@ where
             let resp_ch = self.pending_disconnects.remove(&peer_id).flatten();
 
             if peer.role == Role::Outbound {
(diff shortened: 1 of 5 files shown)


iljakuklic (approved):

TheQuantumPhysicist (approved):
I remember there was a good reason for adding DisconnectedByUser, which IIRC was that we don't want p2p to reconnect to nodes we decide to disconnect from, for any reason. While I'm approving this, please check why this was added and ensure this is the right thing to do.

PR #1096, opened 2023-07-28
Fix crash on disconnecting tx with pool decommissioning

azarovh:
The problem was that while disconnecting the tx with pool decommissioning we try to detect if undo delta should be applied. So it happens that after we switched from dedicated DecommissionPool output type to just locked outputs I missed to change this check. So inputs for tx like ProduceBlock -> LockedOutput were skipped.

P.S. also I split panic condition to fire only for Chain. It looks like a proper thing to do even though current mempool should never trigger that.

### DIFF
--- a/chainstate/test-suite/src/tests/pos_accounting_reorg.rs
+++ b/chainstate/test-suite/src/tests/pos_accounting_reorg.rs
@@ -356,7 +356,8 @@ fn in_memory_reorg_disconnect_produce_pool(#[case] seed: Seed) {
         ))
         .build();
 
-    tf.make_pos_block_builder(&mut rng)
+    let block_b_index = tf
+        .make_pos_block_builder(&mut rng)
         .add_transaction(decommission_pool_tx)
         .with_stake_pool(pool_2_id)
         .with_kernel_input(UtxoOutPoint::new(stake_pool_2_tx_id.into(), 0))
@@ -364,7 +365,12 @@ fn in_memory_reorg_disconnect_produce_pool(#[case] seed: Seed) {
         .with_stake_spending_key(staking_sk_2.clone())
         .with_vrf_key(vrf_sk.clone())
         .build_and_process()
+        .unwrap()
         .unwrap();
+    assert_eq!(
+        Id::<GenBlock>::from(*block_b_index.block_id()),
+        tf.best_block_id()
+    );
 
     // produce block at height 2 that should trigger in memory reorg for block `b`
     tf.make_pos_block_builder(&mut rng)
@@ -376,6 +382,11 @@ fn in_memory_reorg_disconnect_produce_pool(#[case] seed: Seed) {
         .with_vrf_key(vrf_sk)
         .build_and_process()
         .unwrap();
+    // block_b is still the tip
+    assert_eq!(
+        Id::<GenBlock>::from(*block_b_index.block_id()),
+        tf.best_block_id()
+    );
 }
 
 // Produce `genesis -> a -> b` chain, where block `a` creates additional staking pool and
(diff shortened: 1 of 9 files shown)


chainstate/test-suite/src/tests/pos_accounting_reorg.rs
```diff
@@ -291,3 +291,171 @@ fn long_chain_reorg(#[case] seed: Seed) {
...
+    tf.make_pos_block_builder(&mut rng)
+        .with_parent(block_a_id)
+        .with_stake_pool(pool_2_id)
+        .with_kernel_input(UtxoOutPoint::new(stake_pool_2_tx_id.into(), 0))
+        .with_block_signing_key(staking_sk_2.clone())
+        .with_stake_spending_key(staking_sk_2)
+        .with_vrf_key(vrf_sk)
+        .build_and_process()
+        .unwrap();
+}
```
iljakuklic, line 379: Should there be an assertion regarding which block is now the current tip?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -609,29 +633,33 @@ where
...
-                    })
+                apply_accounting_delta_undo = true;
             }
             TxOutput::Transfer(_, _)
             | TxOutput::LockThenTransfer(_, _, _)
             | TxOutput::Burn(_)
-            | TxOutput::ProduceBlockFromStake(_, _) => Ok(()),
-        })
+            | TxOutput::ProduceBlockFromStake(_, _) => { /* do nothing */ }
+        });
```
iljakuklic, line 646: Would this be maybe nicer with `.any()` instead of `.for_each()` and mutating? For additional benefit, it short circuits on the first `true`.

iljakuklic (approved):
Looks good but can't say I 100% understand everything. Some minor comments here:

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -591,10 +591,34 @@ where
         tx_source: TransactionSource,
         tx: &Transaction,
     ) -> Result<(), ConnectTransactionError> {
-        // decrement nonce if disconnected input spent from account
+        // Note: it's not strictly required to check inputs/outputs to undo accounting operations
+        //       because `accounting_block_undo` is not empty in that case. But we do it anyway
+        //       to double check that the logic is consistent.
+        let mut apply_accounting_delta_undo = false;
```
TheQuantumPhysicist, line 597: This mutability is really a bad idea... can't we just do the undo anyway and if it's empty just "do nothing" because it's empty?

TheQuantumPhysicist (approved):
Besides that mutable thing, I think this is fine. We have to do our best to reduce the complexity of our code. Introducing mutable variables complicates the code. We're already facing this issue in p2p as apparently no matter how much I warn against it, people keep thinking that adding more states fixes problems... but in reality it only makes it harder to prove that programs behave correctly, giving the illusion of correctness.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -910,10 +938,10 @@ where
             TransactionSource::Mempool => { /* do nothing */ }
         };
 
-        self.disconnect_pos_accounting_outputs(*tx_source, tx.transaction())?;
-
         self.utxo_cache.disconnect_transaction(tx.transaction(), tx_undo)?;
 
+        self.disconnect_pos_accounting_outputs(*tx_source, tx.transaction())?;
```
TheQuantumPhysicist, line 943: Are you sure this order is correct?

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: Ok I will remove these checks and just apply undo deltas if they are available.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: will revert this as well

TheQuantumPhysicist (approved):

Issue #1034, opened 2023-07-07
Updating the version of the blockchain database should result in a clean re-index


TheQuantumPhysicist:
Backwards compatibility should be handled by version numbers in the blockchain storage. Currently, there's nothing done for that, but a version number exists that can be used. What should happen is: If the hard-coded version number of the database is not equal to the currently stored version in the database on disk, a full re-index of the database should happen. The full re-index can be seen in two different ways:

- We wipe everything in the database and sync from scratch using p2p. This can be an initial solution that prevents the node from crashing due to invalid data in the database.
- To avoid having to re-download everything, we have to define the "re-index" operation. What it does is, it attempts to get all the blocks from the database, and tries to submit them to chainstate. 

Technically speaking, reading the database with an incompatible version isn't always safe. Hence, we should also have a mechanism in place that can detect whether a re-index is possible from the current data in the database. The minimum set of data we need to do a full re-index is:

1. The blocks
2. The height vs blocks (so that they can be submitted in order)
3. Any configuration needed to reconfigure the node (such as, whether the transaction index is enabled, etc).

Any information that will be read from the database should only be read when we're sure it's valid. Checksums can be used (by, for example, re-hashing blocks and comparing them to the values in height vs blocks map) to ensure that the data is valid.

Notice that a similar process should be there for wallets, but for wallets it's much more complicated because we MUST salvage data from wallets, unlike for the blockchain storage, where we could just resync from peers.

PR #1098, opened 2023-07-29
Add chainstate storage version mismatch check

azarovh:
1. Handle chainstate storage version. In case the version mismatch clean up the chainstate db directory and try initialization again. Relates to #1034 
2. The same applies for `ChainConfig::magic_bytes` and `ChainType::name`. They are now stored in db and if mismatch  the db is clean up and synced from scratch.
3. Add `clean-data` command which cleans up data directory before starting the node. Except for the lock file.

### DIFF
--- a/chainstate/launcher/src/lib.rs
+++ b/chainstate/launcher/src/lib.rs
@@ -19,6 +19,10 @@ mod config;
 
 use std::sync::Arc;
 
+use chainstate::InitializationError;
+use chainstate_storage::{BlockchainStorageRead, BlockchainStorageWrite};
+use storage_lmdb::resize_callback::MapResizeCallback;
+
 // Some useful reexports
 pub use chainstate::{
     chainstate_interface::ChainstateInterface, ChainstateConfig, ChainstateError as Error,
@@ -26,18 +30,24 @@ pub use chainstate::{
 };
 pub use common::chain::ChainConfig;
 pub use config::{ChainstateLauncherConfig, StorageBackendConfig};
-use storage_lmdb::resize_callback::MapResizeCallback;
 
 /// Subdirectory under `datadir` where LMDB chainstate database is placed
 pub const SUBDIRECTORY_LMDB: &str = "chainstate-lmdb";
+
+pub const CHAINSTATE_STORAGE_VERSION_UNINITIALIZED: u32 = 0;
+pub const CHAINSTATE_STORAGE_VERSION_V1: u32 = 1;
+pub const CURRENT_CHAINSTATE_STORAGE_VERSION: u32 = CHAINSTATE_STORAGE_VERSION_V1;
 
 fn make_chainstate_and_storage_impl<B: 'static + storage::Backend>(
     storage_backend: B,
     chain_config: Arc<ChainConfig>,
     chainstate_config: ChainstateConfig,
 ) -> Result<Box<dyn ChainstateInterface>, Error> {
-    let storage = chainstate_storage::Store::new(storage_backend)
+    let mut storage = chainstate_storage::Store::new(storage_backend)
         .map_err(|e| Error::FailedToInitializeChainstate(e.into()))?;
+
+    check_storage_version(&mut storage)?;
+
     let chainstate = chainstate::make_chainstate(
         chain_config,
         chainstate_config,
@@ -47,6 +57,28 @@ fn make_chainstate_and_storage_impl<B: 'static + storage::Backend>(
         Default::default(),
     )?;
     Ok(chainstate)
+}
+
+fn check_storage_version<B: 'static + storage::Backend>(
+    storage: &mut chainstate_storage::Store<B>,
+) -> Result<(), Error> {
+    let storage_version =
+        storage.get_storage_version().map_err(InitializationError::StorageError)?;
+
+    if storage_version == CHAINSTATE_STORAGE_VERSION_UNINITIALIZED {
+        storage
+            .set_storage_version(CURRENT_CHAINSTATE_STORAGE_VERSION)
+            .map_err(InitializationError::StorageError)?;
+    } else {
+        utils::ensure!(
+            storage_version == CURRENT_CHAINSTATE_STORAGE_VERSION,
+            InitializationError::ChainstateStorageVersionMismatch(
+                storage_version,
+                CURRENT_CHAINSTATE_STORAGE_VERSION
+            )
+        );
+    }
+    Ok(())
 }
 
 /// Create chainstate together with its storage
--- a/chainstate/launcher/src/lib.rs
+++ b/chainstate/launcher/src/lib.rs
@@ -16,13 +16,12 @@
 //! Tools to set up chainstate together with its storage
 
 mod config;
+mod storage_compatibility;
 
 use std::sync::Arc;
 
 use chainstate::InitializationError;
-use chainstate_storage::{BlockchainStorageRead, BlockchainStorageWrite};
 use storage_lmdb::resize_callback::MapResizeCallback;
-use utils::ensure;
 
 // Some useful reexports
 pub use chainstate::{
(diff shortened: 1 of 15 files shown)


chainstate/launcher/src/lib.rs
```diff
@@ -19,25 +19,50 @@ mod config;
...
+            .map_err(InitializationError::StorageError)?;
+    } else {
+        utils::ensure!(
+            storage_version == CURRENT_CHAINSTATE_VERSION,
+            InitializationError::ChainstateStorageVersionMismatch(
+                storage_version,
+                CURRENT_CHAINSTATE_VERSION
+            )
+        );
+    }
```
TheQuantumPhysicist, line 64: Please move this logic to a function

chainstate/launcher/src/lib.rs
```diff
@@ -19,25 +19,35 @@ mod config;
...
 pub use config::{ChainstateLauncherConfig, StorageBackendConfig};
-use storage_lmdb::resize_callback::MapResizeCallback;
 
 /// Subdirectory under `datadir` where LMDB chainstate database is placed
 pub const SUBDIRECTORY_LMDB: &str = "chainstate-lmdb";
 
+pub const CHAINSTATE_STORAGE_VERSION_UNINITIALIZED: u32 = 0;
+pub const CHAINSTATE_STORAGE_VERSION_V1: u32 = 1;
+pub const CURRENT_CHAINSTATE_STORAGE_VERSION: u32 = CHAINSTATE_STORAGE_VERSION_V1;
+
```
TheQuantumPhysicist, line 40: Why do these have to be public?

node-lib/src/runner.rs
```diff
@@ -226,13 +229,28 @@ pub async fn setup(options: Options) -> Result<Node> {
...
 fn lock_data_dir(data_dir: &PathBuf) -> Result<std::fs::File> {
-    let lock = std::fs::File::create(data_dir.join(".lock"))
+    let lock = std::fs::File::create(data_dir.join(LOCK_FILE_NAME))
         .map_err(|e| anyhow!("Cannot create lock file in {data_dir:?}: {e}"))?;
     fs4::FileExt::try_lock_exclusive(&lock)
         .map_err(|e| anyhow!("Cannot lock directory {data_dir:?}: {e}"))?;
     Ok(lock)
 }
 
+fn clean_data_dir(data_dir: &PathBuf, exclude: &Path) -> Result<()> {
```
TheQuantumPhysicist, line 239: This is fine... it works... but shouldn't the exclude thing be a slice?

node-lib/src/runner.rs
```diff
@@ -253,8 +271,40 @@ async fn start(
...
+
     log::info!("Starting with the following config:\n {node_config:#?}");
-    let (manager, controller) = initialize(chain_config, data_dir, node_config).await?;
+    let (manager, controller) = match initialize(
+        chain_config.clone(),
+        data_dir.clone(),
+        node_config.clone(),
+    )
+    .await
+    {
```
TheQuantumPhysicist, line 285: Nitpick: Please add an empty line after await

node-lib/src/runner.rs
```diff
@@ -253,8 +271,40 @@ async fn start(
...
+                chainstate::InitializationError::ChainstateStorageVersionMismatch(_, _) => {
+                    log::info!("Failed to init chainstate: {e} \n Cleaning up current db and trying from scratch.");
+
+                    let storage_config: StorageBackendConfig =
+                        node_config.chainstate.clone().unwrap_or_default().storage_backend.into();
+
+                    // cleanup storage directory and retry initialization
+                    if let Some(storage_subdir_name) = storage_config.subdirectory_name() {
+                        let path = data_dir.join(storage_subdir_name);
+                        std::fs::remove_dir_all(path).expect("Removing directory must succeed");
```
TheQuantumPhysicist, line 298: This can fail if the directory doesn't exist. 

Also, please write an error message that's more friendly. This message directly faces users at a very high level if something fails.

node-lib/src/runner.rs
```diff
@@ -253,8 +271,40 @@ async fn start(
...
+        data_dir.clone(),
+        node_config.clone(),
+    )
+    .await
+    {
+        Ok((manager, controller)) => (manager, controller),
+        Err(error) => match error.downcast_ref::<ChainstateError>() {
+            Some(ChainstateError::FailedToInitializeChainstate(e)) => match e {
+                chainstate::InitializationError::ChainstateStorageVersionMismatch(_, _) => {
+                    log::info!("Failed to init chainstate: {e} \n Cleaning up current db and trying from scratch.");
```
TheQuantumPhysicist, line 290: warn

chainstate/launcher/src/lib.rs
```diff
@@ -19,25 +19,35 @@ mod config;
...
 pub use common::chain::ChainConfig;
 pub use config::{ChainstateLauncherConfig, StorageBackendConfig};
-use storage_lmdb::resize_callback::MapResizeCallback;
 
 /// Subdirectory under `datadir` where LMDB chainstate database is placed
 pub const SUBDIRECTORY_LMDB: &str = "chainstate-lmdb";
 
+pub const CHAINSTATE_STORAGE_VERSION_UNINITIALIZED: u32 = 0;
+pub const CHAINSTATE_STORAGE_VERSION_V1: u32 = 1;
+pub const CURRENT_CHAINSTATE_STORAGE_VERSION: u32 = CHAINSTATE_STORAGE_VERSION_V1;
```
iljakuklic, line 39: This could be a custom type wrapper over u32 rather than plain u32, or even an enum with repr u32. These constants could then be turned into associated constants like `StorageVersion::CURRENT` etc.

iljakuklic (approved):
Looks good enough for now. Just one minor suggestion but it's fine to ignore it.

node-lib/src/runner.rs
azarovh: cannot do that because it's automatic formatting of match expression

TheQuantumPhysicist: Never mind.

chainstate/launcher/src/lib.rs
azarovh: ignoring for now

chainstate/launcher/src/lib.rs
```diff
@@ -19,25 +19,38 @@ mod config;
...
     chain_config: Arc<ChainConfig>,
     chainstate_config: ChainstateConfig,
 ) -> Result<Box<dyn ChainstateInterface>, Error> {
-    let storage = chainstate_storage::Store::new(storage_backend)
+    let mut storage = chainstate_storage::Store::new(storage_backend)
         .map_err(|e| Error::FailedToInitializeChainstate(e.into()))?;
+
+    check_storage_version(&mut storage)?;
+    check_magic_bytes(&mut storage, chain_config.as_ref())?;
+    check_chain_type(&mut storage, chain_config.as_ref())?;
```
TheQuantumPhysicist, line 52: Perhaps move these into one call, `check_storage_compatibility()`, and move all these calls and impls into a submodules. Then we can test them in isolation.

chainstate/src/detail/error.rs
```diff
@@ -206,6 +206,16 @@ pub enum InitializationError {
...
+    #[error(
+        "Node cannot load chainstate database because the versions mismatch: db `{0}`, app `{1}`"
+    )]
+    ChainstateStorageVersionMismatch(u32, u32),
+    #[error(
+        "Chain's config magic bytes do not match the one from database : expected `{0:?}`, actual `{1:?}`"
+    )]
+    ChainConfigMagicBytesMismatch([u8; 4], [u8; 4]),
+    #[error("Node's chain type doesn't match the one in the database : db `{0}`, app `{1}`")]
+    ChainTypeMismatch(String, String),
```
TheQuantumPhysicist, line 218: These 3 errors can go into that module and become a single error and can be matched once

TheQuantumPhysicist (approved):
Besides some clean up, this is good to go. We'll add functional tests later.

iljakuklic (approved):

PR #1099, opened 2023-07-29
Add ban/unban API

pavel-kokolemin:
I think we should not ban peers after the `DifferentNetwork` error, because it makes deploying a new testnet network annoying. Once a node is upgraded, it will ban all old peers and be banned by them.

### DIFF
--- a/p2p/src/interface/p2p_interface.rs
+++ b/p2p/src/interface/p2p_interface.rs
@@ -25,6 +25,10 @@ pub trait P2pInterface: Send + Sync {
     async fn connect(&mut self, addr: String) -> crate::Result<()>;
     async fn disconnect(&mut self, peer_id: PeerId) -> crate::Result<()>;
 
+    async fn list_banned(&mut self) -> crate::Result<Vec<String>>;
+    async fn ban(&mut self, addr: String) -> crate::Result<()>;
+    async fn unban(&mut self, addr: String) -> crate::Result<()>;
+
     async fn get_peer_count(&self) -> crate::Result<usize>;
     async fn get_bind_addresses(&self) -> crate::Result<Vec<String>>;
     async fn get_connected_peers(&self) -> crate::Result<Vec<ConnectedPeer>>;
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -328,7 +328,7 @@ where
         );
 
         if peer.score >= *self.p2p_config.ban_threshold {
-            self.peerdb.ban_peer(&peer.address);
+            self.peerdb.ban(peer.address.as_bannable());
             self.disconnect(peer_id, None);
         }
     }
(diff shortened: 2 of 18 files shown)


TheQuantumPhysicist (approved):

p2p/src/interface/p2p_interface.rs
```diff
@@ -25,6 +25,10 @@ pub trait P2pInterface: Send + Sync {
     async fn connect(&mut self, addr: String) -> crate::Result<()>;
     async fn disconnect(&mut self, peer_id: PeerId) -> crate::Result<()>;
 
+    async fn list_banned(&mut self) -> crate::Result<Vec<String>>;
```
iljakuklic, line 28: Could we do better than `Vec<String>` here? String manipulation tends to be brittle and exposes whole bunch of operations that are nonsensical for ban address identifiers.

p2p/src/peer_manager/mod.rs
```diff
@@ -986,6 +986,17 @@ where
...
+                response.send(self.peerdb.list_banned().cloned().collect())
+            }
+            PeerManagerEvent::Ban(address, response) => {
+                self.peerdb.ban(address);
+                response.send(Ok(()));
+            }
+            PeerManagerEvent::Unban(address, response) => {
+                self.peerdb.unban(&address);
+                response.send(Ok(()));
+            }
```
iljakuklic, line 999: Would it be useful for the response to include whether ban/unban has been successful?

iljakuklic (approved):
Looks good. A couple of minor suggestions below.

p2p/src/interface/p2p_interface.rs
pavel-kokolemin: We could, but we would have to remove generic addresses from `NetworkingService` and instead use concrete types  (`SocketAddress`/`IpAddress`):
```
pub trait NetworkingService {
    type Address...
    type BannableAddress...
```
It's probably not a good time to do this, and I don't know how else to do it.

p2p/src/peer_manager/mod.rs
pavel-kokolemin: I think they both can't fail. Ban will always succeed because we just store the IP address in the DB. Unban will always succeed because the specified IP won't be banned after the operation.

BTW, it just occurred to me that we should also disconnect selected peers when we ban them. I will do that in the next PR.

iljakuklic: Yea what I meant is something like:

* When you ban an address that's already banned, report that
* When you unban an address that has not been banned.

However, usefulness of that is questionable, so I don't mind not having it.

p2p/src/interface/p2p_interface.rs
iljakuklic: Ok, don't worry about it now

PR #1104, opened 2023-07-30
Fix token decimal parsing

OBorce:
Fetch token info and use the correct number of decimals to parse the input amount string in from the CLI


TheQuantumPhysicist (approved):
Looking fine, assuming my molten brain isn't imagining things... though I have to say there's a lack of tests.

iljakuklic (approved):

PR #1106, opened 2023-07-30
Create the Addressable trait to generalize how we handle Address and use it for more than just Destination

TheQuantumPhysicist:
In preparation to have staking pools and delegation ids become more readable, this is the first step.

Those may be included in this PR

### DIFF
--- a/common/src/address/mod.rs
+++ b/common/src/address/mod.rs
@@ -73,7 +73,7 @@ impl<T: Addressable> Address<T> {
     pub fn decode_object(&self, cfg: &ChainConfig) -> Result<T, AddressError> {
         let data = encoding::decode(&self.address)?;
         let raw_dest = data.data();
-        let result = T::decode_from_bytes_from_address(&mut &raw_dest[..])
+        let result = T::decode_from_bytes_from_address(&raw_dest[..])
             .map_err(|e| AddressError::DecodingError(e.to_string()))?;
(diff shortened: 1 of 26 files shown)


pavel-kokolemin (approved):

common/src/address/mod.rs
```diff
@@ -55,80 +41,72 @@ impl From<Bech32Error> for AddressError {
...
 
-    pub fn destination(&self, cfg: &ChainConfig) -> Result<Destination, AddressError> {
+    pub fn decode_object(&self, cfg: &ChainConfig) -> Result<T, AddressError> {
         let data = encoding::decode(&self.address)?;
-        let raw_dest = data.data();
-        let destination = Destination::decode_all(&mut &raw_dest[..])
-            .map_err(|e| AddressError::DestinationDecodingError(e.to_string()))?;
-        if data.hrp() != cfg.address_prefix(&destination) {
+        let raw_data = data.data();
+        let result = T::decode_from_bytes_from_address(raw_data)
```
alfiedotwtf, line 76: Reads a bit weird. Maybe `decode_from_address_bytes()`?

alfiedotwtf (approved):

muursh (approved):

common/src/address/mod.rs
TheQuantumPhysicist: I want to avoid saying "address bytes" because it can be interpreted as serialization of the address itself.

PR #1108, opened 2023-07-30
Temporary disable a p2p test (sporadic failures)

pavel-kokolemin:


alfiedotwtf (approved):

TheQuantumPhysicist:
I'd rather keep the test failing in CI and rerunning when it fails. It should stay as a reminder that there's an issue until we have the time to fix it. We will have time soon enough.

pavel-kokolemin:
> I'd rather keep the test failing in CI and rerunning when it fails. It should stay as a reminder that there's an issue until we have the time to fix it. We will have time soon enough.

OK, sure, no problem.

PR #1109, opened 2023-07-31
Create a Dockerfile for node, wallet cli and gui

erubboli:
This PR add a `docker` folder with the `Dockerfile` for the 3 binaries and relative containers. 
It includes a guide to build the images and a README to run node and cli. 
The README doesn't cover the GUI yet.


TheQuantumPhysicist (approved):
Haven't tested it, but the format looks good.

muursh (approved):

ImplOfAnImpl (approved):
It's better to have a `.dockerignore` file at the root of the source tree and specify `target` in it, so that the `target` directory is excluded from the docker's build context when an image is built (this is mostly relevant for developer machines, where the `target` directory can be huge).

Also, some nitpicks:
1) The base images for the "runtime" stages should probably use the "slim" version of the Debian image. The difference is not that big though, debian:latest AKA debian:bookworm is 116MB, while debian:bookworm-slim is 75Mb.

2) IMO it'd be better if the "build" stage of all the images was shared, so that same things (e.g. 3rd-party crates) are not rebuilt multiple times. AFAIK Docker doesn't allow to reference one docker file from another, but they can reference local images. So we could have a shell script that would first build the "builder" image, which would build all the apps, and then each "leaf" dockerfile would reference the "builder" image and copy the relevant app binary from it.

3) Even if ignoring the previous suggestion, IMO it's better to have a shell script that would build all the images and tag them appropriately. Also, it can ensure that the current directory (and therefore the build context) is correct when `docker build` is invoked.

4) Not sure if other devs will agree with me on this, but anyway: instead of the `docker` directory at the root of the source tree I'd create a `build` directory in the root and then put the `docker` directory inside it. Then we'd have a single place where all build-related stuff could go in the future.
(One example of existing build-related stuff that could go into `build` is the `codecheck.py` script; currently it lives in `.github/scripts`, which is not super nice).

Finally, the docker commands on the user part are rather complex, so perhaps we could use `docker compose` for that, so that the user could just run `docker compose up` in a directory and get everything working. But I'm not sure about this, because the docker compose script has to be put somewhere and it probably shouldn't be the main repository.

PR #1111, opened 2023-07-31
Fix pool id encodings in wallet CLI commands

OBorce:


azarovh (approved):

TheQuantumPhysicist (approved):

common/src/chain/tokens/mod.rs
```diff
@@ -13,11 +13,34 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use serialization::{Decode, Encode};
+use serialization::{Decode, DecodeAll, Encode};
 
 pub type TokenId = H256;
```
TheQuantumPhysicist, line 18: This can be a problem. Please do like PoolId and DelegationId and create an ID type. Using a type alias can create an issue with addresses because it removes type safety completely.

common/src/chain/tokens/mod.rs
```diff
@@ -13,11 +13,49 @@
...
+    address::{traits::Addressable, AddressError},
+    primitives::{Amount, Id, H256},
+};
+
+impl TokenId {
+    pub fn random_using<R: Rng + CryptoRng>(rng: &mut R) -> Self {
+        Self::new(H256::random_using(rng))
+    }
+
+    pub fn zero() -> Self {
```
TheQuantumPhysicist, line 35: const fn... if you get the chance. Not so important.

TheQuantumPhysicist (approved):

common/src/chain/tokens/mod.rs
OBorce: fixed

PR #1115, opened 2023-07-31
Fix decommission pool to use the pledge amount.

OBorce:
- fix typos


chainstate/src/rpc.rs
```diff
@@ -77,6 +77,9 @@ trait ChainstateRpc {
     #[method(name = "stake_pool_balance")]
     async fn stake_pool_balance(&self, pool_id: PoolId) -> RpcResult<Option<Amount>>;
 
+    #[method(name = "stake_pool_pledge")]
+    async fn stake_pool_pledge(&self, pool_id: PoolId) -> RpcResult<Option<Amount>>;
```
azarovh, line 81: why not `stake_pool_data`?  it might be useful to display pool info later

wallet/wallet-controller/src/lib.rs
```diff
@@ -643,15 +643,15 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static, W: WalletEvents> Controll
 
         let pool_balance = self
```
azarovh, line 644: staker_balance

azarovh (approved):

azarovh:
You can ignore my comments for now because of urgency.

iljakuklic (approved):

wallet/wallet-controller/src/lib.rs
OBorce: renamed

PR #1117, opened 2023-08-01
Remove BannableAddress generic (p2p)

pavel-kokolemin:
Up for discussion. I think it's time to remove two generic network addresses from p2p: `NetworkingService::Address` and `NetworkingService::BannableAddress`. Currently we use `SocketAddr`/`IpAddr` as `Address`/`BannableAddress` everywhere, and I don't see that changing anytime soon. Without these two generics, the code should be a bit cleaner. This PR removes `BanableAddress` to show this. If you agree, the next step will be to remove `NetworkingService::Address` as well.


p2p/types/src/banable_address.rs
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::net::IpAddr;
+
+pub type BannableAddress = IpAddr;
```
TheQuantumPhysicist, line 18: I'm still not done reviewing the PR, but just looking at this, can't we use a proper strong type for this? Type aliases are a real mess when implementing traits for them.

p2p/types/src/lib.rs
```diff
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+pub mod banable_address;
```
TheQuantumPhysicist, line 16: bannable

TheQuantumPhysicist (approved):
Besides the comments, I'm OK with this.

muursh (approved):

TheQuantumPhysicist (approved):
Looking fine now.

PR #1118, opened 2023-08-01
Move chainstate storage init inside constructor

azarovh:
This is a fix for previous review #1098:

- Create dedicated type for storage version
- Avoid using mut storage in check functions

### DIFF
--- a/chainstate/src/detail/error.rs
+++ b/chainstate/src/detail/error.rs
@@ -21,6 +21,7 @@ use super::{
         storage::TransactionVerifierStorageError,
     },
 };
+use chainstate_storage::ChainstateStorageVersion;
 use chainstate_types::{GetAncestorError, PropertyQueryError};
 use common::{
     chain::{
@@ -214,10 +215,18 @@ pub enum InitializationError {
 pub enum StorageCompatibilityCheckError {
     #[error("Block storage error: `{0}`")]
     StorageError(#[from] chainstate_storage::Error),
+    #[error("Storage version is missing in the db")]
+    StorageVersionMissing,
+    #[error("Failed to convert storage error")]
+    StorageVersionConversionError,
+    #[error("Magic bytes are missing in the db")]
+    MagicBytesMissing,
+    #[error("Chain type is missing in the db")]
+    ChainTypeMissing,
     #[error(
-        "Node cannot load chainstate database because the versions mismatch: db `{0}`, app `{1}`"
+        "Node cannot load chainstate database because the versions mismatch: db `{0:?}`, app `{1:?}`"
     )]
-    ChainstateStorageVersionMismatch(u32, u32),
+    ChainstateStorageVersionMismatch(ChainstateStorageVersion, ChainstateStorageVersion),
     #[error(
         "Chain's config magic bytes do not match the one from database : expected `{0:?}`, actual `{1:?}`"
     )]
--- a/chainstate/src/detail/error.rs
+++ b/chainstate/src/detail/error.rs
@@ -217,8 +217,6 @@ pub enum StorageCompatibilityCheckError {
     StorageError(#[from] chainstate_storage::Error),
     #[error("Storage version is missing in the db")]
     StorageVersionMissing,
-    #[error("Failed to convert storage error")]
-    StorageVersionConversionError,
     #[error("Magic bytes are missing in the db")]
     MagicBytesMissing,
     #[error("Chain type is missing in the db")]
--- a/chainstate/storage/src/internal/mod.rs
+++ b/chainstate/storage/src/internal/mod.rs
@@ -56,7 +56,7 @@ impl<B: storage::Backend> Store<B> {
         // Set defaults if missing
 
         if storage.get_storage_version()?.is_none() {
-            storage.set_storage_version(ChainstateStorageVersion::CURRENT as u32)?;
+            storage.set_storage_version(ChainstateStorageVersion::CURRENT)?;
         }
 
         if storage.get_magic_bytes()?.is_none() {
@@ -224,7 +224,7 @@ macro_rules! delegate_to_transaction {
 
 impl<B: storage::Backend> BlockchainStorageRead for Store<B> {
     delegate_to_transaction! {
-        fn get_storage_version(&self) -> crate::Result<Option<u32>>;
+        fn get_storage_version(&self) -> crate::Result<Option<ChainstateStorageVersion>>;
         fn get_magic_bytes(&self) -> crate::Result<Option<[u8; 4]>>;
         fn get_chain_type(&self) -> crate::Result<Option<String>>;
         fn get_best_block_id(&self) -> crate::Result<Option<Id<GenBlock>>>;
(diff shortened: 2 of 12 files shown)


chainstate/storage/src/internal/mod.rs
```diff
@@ -42,12 +42,35 @@ use crate::{
...
+            storage.set_storage_version(ChainstateStorageVersion::CURRENT as u32)?;
+        }
+
+        if storage.get_magic_bytes()?.is_none() {
+            storage.set_magic_bytes(chain_config.magic_bytes())?;
+        }
+
+        if storage.get_chain_type()?.is_none() {
+            storage.set_chain_type(chain_config.chain_type().name())?;
+        }
```
iljakuklic, line 68: Should we require an all-or-nothing storage state here? I.e. enforce storage version, magic bytes, chain type to be all `None` or all `Some(_)` and report an error is some of them are initialised but others are not.

Also the three updates could be done in a transaction to get atomic initialisation.

chainstate/storage/src/internal/version.rs
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[repr(u32)]
+pub enum ChainstateStorageVersion {
+    // skipping 0 for historical reasons
+    V1 = 1,
+}
```
iljakuklic, line 21: Would it make sense to make this implement `Encode`, `Decode` and have the strongly typed version in the DB schema?

iljakuklic (approved):
Nice, definitely an improvement. Some suggestions below:

chainstate/storage/src/internal/mod.rs
TheQuantumPhysicist: Sounds like a good idea, but I wouldn't harp on it a lot, because if any of them isn't set, the whole thing will fail.

TheQuantumPhysicist (approved):
This looks fine. Though just a question out of carefulness: I remember we deleting only the database directory. Please confirm that's the case, and that we're not deleting the whole data directory. This is because a user setting a wrong data directory could cause a disaster if a wildcard is used.

azarovh:
>  Though just a question out of carefulness: I remember we deleting only the database directory.

yes we are removing only that. So in case chainstate is configured with LMDB `~/Library/Application Support/Mintlayer/testnet/chainstate-lmdb` will be removed.

chainstate/storage/src/internal/mod.rs
azarovh: used transaction

muursh (approved):
Looks good to me

chainstate/src/detail/error.rs
```diff
@@ -214,10 +215,16 @@ pub enum InitializationError {
...
     StorageError(#[from] chainstate_storage::Error),
+    #[error("Storage version is missing in the db")]
+    StorageVersionMissing,
+    #[error("Magic bytes are missing in the db")]
+    MagicBytesMissing,
+    #[error("Chain type is missing in the db")]
+    ChainTypeMissing,
     #[error(
-        "Node cannot load chainstate database because the versions mismatch: db `{0}`, app `{1}`"
+        "Node cannot load chainstate database because the versions mismatch: db `{0:?}`, app `{1:?}`"
```
iljakuklic, line 225: For a friendlier error message, it may be useful to implement `Display` for `ChainstateStorageVersion`, since the auto-derived `Debug` is likely to be quite verbose. Fine to keep it like it is too.

iljakuklic (approved):

chainstate/storage/src/internal/version.rs
azarovh: just realised that this breaks db compatibility with testnet. @TheQuantumPhysicist what's our approach in such cases? Just revert it for now?

chainstate/storage/src/internal/version.rs
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+#[derive(Debug, Encode, Decode, Clone, Copy, Eq, PartialEq)]
+pub enum ChainstateStorageVersion {
+    V0,
+    V1,
+}
```
ImplOfAnImpl, line 22: I wonder what's the purpose of using an enum here. Why not just use a "newtype" struct holding a u32?

ImplOfAnImpl (approved):

chainstate/storage/src/internal/version.rs
iljakuklic: > just realised that this breaks db compatibility with testnet. @TheQuantumPhysicist what's our approach in such cases? Just revert it for now?

Why? Does it not encode/decode identically to the existing `u32`?

iljakuklic: Also the approach @ImplOfAnImpl proposes with making this a newtype over `u32` should work in a backwards compatible way.

TheQuantumPhysicist: > just realised that this breaks db compatibility with testnet. @TheQuantumPhysicist what's our approach in such cases? Just revert it for now?

It doesn't matter because the worst thing that can happen is a resync... assuming not everyone is going to do it at the same instance, we're fine.

TheQuantumPhysicist: But again, it should decide to wipe the db and resync. It shouldn't crash.

chainstate/storage/src/internal/version.rs
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serialization::{Decode, Encode};
+
+#[derive(Debug, Encode, Decode, Clone, Copy, Eq, PartialEq)]
+pub struct ChainstateStorageVersion(u32);
```
TheQuantumPhysicist, line 19: I've seen many discussions here about what to do about this... and I just want to point out that if we want to make it perfect, we can put all the information (version, magic bytes and net type) in one struct. But I don't really care that much. I'm OK with the way things are now as long as it doesn't lead to a crash and the worst thing that can happen is a wipe and resync.

chainstate/storage/src/internal/version.rs
azarovh: In case of an enum it actually crashes trying to decode storage version.
```
     Running `target/debug/node-daemon testnet`
thread 'main' panicked at 'db values to be encoded correctly: Error { cause: None, desc: "Input buffer has still data left after decoding!" }', /Users/heorhii/dev/mintlayer-core/chainstate/storage/src/internal/store_tx.rs:395:1
```

Works fine if struct is used (as in my latest commit)

TheQuantumPhysicist: We shouldn't let it crash. If there's an incompatibility, we should wipe the DB and start over. This may be related to #1053

PR #1123, opened 2023-08-01
Add CLI argument to show locked balance and utxos

OBorce:
Add optional arguments to getbalance and listutxos to show locked outputs

### DIFF
--- a/wallet/src/account/output_cache/mod.rs
+++ b/wallet/src/account/output_cache/mod.rs
@@ -30,6 +30,7 @@ use utils::ensure;
 use wallet_types::{
     utxo_types::{get_utxo_state, UtxoStates},
     wallet_tx::TxState,
+    with_locked::WithLocked,
     AccountWalletTxId, BlockInfo, WalletTx,
 };
 
@@ -202,13 +203,8 @@ impl OutputCache {
             }
         }
 
-        let tx_block_info = match tx.state() {
-            TxState::Confirmed(height, timestamp) => Some(BlockInfo { height, timestamp }),
-            TxState::Inactive
-            | TxState::Conflicted(_)
-            | TxState::InMempool
-            | TxState::Abandoned => None,
-        };
+        let tx_block_info = get_block_info(&tx);
+
         if let Some(block_info) = tx_block_info {
             for (idx, output) in tx.outputs().iter().enumerate() {
                 match output {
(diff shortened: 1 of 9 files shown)


wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -181,13 +183,17 @@ pub enum WalletCommand {
     SyncWallet,
 
     GetBalance {
+        #[arg(value_enum, default_value_t = CliWithLocked::Unlocked)]
```
azarovh, line 186: Up for a discussion but I'd show all utxos and balances by default. For balance it could be a separate "locked balance" line.

azarovh (approved):

wallet/src/account/output_cache/mod.rs
```diff
@@ -437,6 +424,36 @@ impl OutputCache {
     }
 }
 
+// Check if the TxOutput's timelock is in a specified state Locked/Unlocked or both
```
TheQuantumPhysicist, line 427: ///

wallet/src/account/output_cache/mod.rs
```diff
@@ -457,3 +474,8 @@ fn valid_timelock(
         })
     })
 }
+
+/// Check Tx is in the selected state Confirmed/Inactve/Abandoned...
```
TheQuantumPhysicist, line 478: Inactive

wallet/src/account/output_cache/mod.rs
TheQuantumPhysicist: I'm trying to understand this function. Is it better to describe this function as "checks the output against the current block height and compares it with the `locked_state` parameter. If they match, the function return true, if they don't, it returns false. For example, if we would like to check that an output is locked, we pass `locked_state = WithLocked::Locked`, and pass the output in question. If the output is locked, the function returns true. Otherwise, it returns false".

If this better describes this function, please use it as documentation.

wallet/types/src/with_locked.rs
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Enum used to specify whether to include locked balance/utxos for wallet commands
+#[derive(Debug, Clone, Copy)]
+pub enum WithLocked {
+    All,
```
TheQuantumPhysicist, line 19: I would call this `Any`

wallet/src/account/output_cache/mod.rs
OBorce: fixed

wallet/types/src/with_locked.rs
OBorce: changed

pavel-kokolemin (approved):

wallet/wallet-cli-lib/src/commands/helper_types.rs
```diff
@@ -124,3 +125,20 @@ pub fn format_delegation_info(
         balance.into_fixedpoint_str(chain_config.coin_decimals()),
     )
 }
+
+#[derive(Debug, Clone, Copy, ValueEnum)]
+pub enum CliWithLocked {
+    All,
```
TheQuantumPhysicist, line 131: Any

wallet/wallet-cli-lib/src/commands/mod.rs
TheQuantumPhysicist: I prefer to show only spendable coins. We can add a note that this displays only spendable coins, and they have to check `help getbalance` for more information.

TheQuantumPhysicist (approved):

PR #1124, opened 2023-08-02
Latch `is_initial_block_download_finished` flag

pavel-kokolemin:


chainstate/src/detail/mod.rs
```diff
@@ -744,28 +748,38 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+
+    /// Update `is_initial_block_download_finished` when tip has changes (can only be set once)
+    fn update_initial_block_download_flag(&mut self) -> Result<(), PropertyQueryError> {
+        if self.is_initial_block_download_finished.test() {
+            return Ok(());
         }
 
         // TODO: Add a check for importing and reindex.
 
         // TODO: Add a check for the chain trust.
```
ImplOfAnImpl, line 769: I wonder whether these TODOs are still relevant (and if so, what they actually mean).

chainstate/src/detail/mod.rs
```diff
@@ -744,28 +748,38 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
+    }
+
+    /// Returns true if the given block timestamp is newer than `ChainstateConfig::max_tip_age`.
+    fn is_fresh_block(&self, time: &BlockTimestamp) -> bool {
+        let now = self.time_getter.get_time();
+        time.as_duration_since_epoch() + self.chainstate_config.max_tip_age.clone().into() > now
+    }
+
+    /// Update `is_initial_block_download_finished` when tip has changes (can only be set once)
+    fn update_initial_block_download_flag(&mut self) -> Result<(), PropertyQueryError> {
```
ImplOfAnImpl, line 762: "when tip has changes" sounds broken.
Perhaps, it should be something like "if we've got a fresh tip"?

chainstate/src/detail/mod.rs
```diff
@@ -744,28 +748,38 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
 
-    /// Returns true if the given block timestamp is newer than `ChainstateConfig::max_tip_age`.
-    fn is_fresh_block(&self, time: &BlockTimestamp) -> bool {
-        let now = self.time_getter.get_time();
-        time.as_duration_since_epoch() + self.chainstate_config.max_tip_age.clone().into() > now
+        if self.is_fresh_block(&tip_timestamp) {
+            self.is_initial_block_download_finished.set();
+        }
+
+        Ok(())
```
ImplOfAnImpl, line 782: I wonder whether we should write `is_initial_block_download_finished` to the DB once it becomes true, so that if the node is relaunched, it doesn't get reset to false again.

ImplOfAnImpl (approved):

chainstate/src/detail/mod.rs
TheQuantumPhysicist: The first one: It's about reindexing. Reindexing is the process of reparsing blocks such that the database can be rebuilt. This is useful in cases of database corruption or when settings like "tx index" are changed. If the reindex process is happening, then it's always true that we're in the "initial block download" state.

The second one: Assuming a monolithic network that we work on during the mainnet, the chain-trust always goes up. Given that, on every release, we can say that a chain-trust below a certain point is always an "initial block download", because we know that the monotonous mainnet must be longer than that point.

Yes, both points are still relevant. But they're more like optimizations than necessities.

chainstate/src/detail/mod.rs
TheQuantumPhysicist: It should be reset when the node is shutdown, because the node could be down for a while and hence the "redownload" process should be started again.

chainstate/src/detail/mod.rs
TheQuantumPhysicist: I think it's typo from "when tip has changed", or "when tip changes"

pavel-kokolemin: Corrected, thanks.

TheQuantumPhysicist (approved):

chainstate/src/detail/mod.rs
iljakuklic: > It should be reset when the node is shutdown, because the node could be down for a while and hence the "redownload" process should be started again.

Something similar can happen when the network is down for a while but the node keeps on running. Once we're reconnected, the node has to catch up. How should that be handled? Seems the current PR prevents the node from reverting into the IBD stage in that case but maybe it should be handled somehow too.

TheQuantumPhysicist: Under normal operation circumstances, the node should never fall back. Besides, the IBD state should be seen as an optimization, more than it being a global state.

iljakuklic (approved):

PR #1125, opened 2023-08-02
Fix transaction broadcasting in wallet commands to be consistent

OBorce:
Reuse the broadcasting functionality in all commands to be the same


TheQuantumPhysicist (approved):
I'm OK with this... but we should start writing tests to ensure these contracts are handled properly.

pavel-kokolemin (approved):

Issue #1024, opened 2023-07-04
Study time synchronization issues


TheQuantumPhysicist:
PR #1017 is an attempt at fixing time synchronization issues between nodes. However, this is not enough. Here are a few issues:

- Nodes can have clocks that drift over time
- A block far in the future shouldn't always result in banning the peer, but should instead try again later. Currently an invalid block (due to being a little in the future, more than the limit) will make the block invalid, but then will result in subsequent blocks to have no parent, and hence being orphan, resulting in banning the peer.
- There has to be an error with zero ban score for blocks not too far in the future (a few minutes) and a ban on blocks way too far in the future.
- The strategy implemented in #1017 is problematic because it ruins the usage of mocked time. Waiting isn't really the best wait to do this. The `handle_header_list` check that delays time should go eventually.

pavel-kokolemin:
Some info about how other blockchains handle clock differences:

1. Ethereum:
From `https://github.com/ethereum/go-ethereum/issues/20498`:
```
The discovery protocol relies on clock drift being smaller than 30 seconds. Currently you need to ensure that in order to run a node. We're working on a secondary discovery protocol based on DNS that should not have this requirement.
```

2. Bitcoin:
No peer disconnects, the node adjusts the local time by taking the median of all known clock differences.

3. Cardano
From `https://github.com/input-output-hk/cardano-node/issues/2529`:
```
As is well-known, clock skew causes the node to suspend its peers, because the blocks downloaded from them are "in the future" (see the error message in the next section).
Once all peers are suspended, the node chain degrades, and after a while it obtaining the ledger state.
```

And some interesting reading about Cardano's problems with blocks from the future - `https://github.com/input-output-hk/ouroboros-network/issues/4251`

TheQuantumPhysicist:
It's helpful to take a look at how bitcoin deals with time offsets: https://github.com/bitcoin/bitcoin/blob/5bbf735defac20f58133bea95226e13a5d8209bc/src/timedata.cpp#L30

TheQuantumPhysicist:
Due to prioritizing creating an RPC wallet, this will most likely not make it to mainnet. But it's still a priority. It's going to be removed from the mainnet milestone.

PR #1130, opened 2023-08-02
Block announcement rework

ImplOfAnImpl:
1) The node should now only send header lists which are connected to a block that the peer already has. Also, it should correctly update its pending_headers list (previosly called known_headers) whenever it receives a header list.

2) tests in p2p/src/sync/tests/network_sync.rs should now be more robust and not fail spuriously.

I also did some renaming in various places.


chainstate/src/detail/query.rs
```diff
@@ -177,52 +177,79 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         self.chainstate_ref.get_block_height_in_main_chain(id)
     }
 
-    pub fn get_headers(
+    fn get_headers_above(
```
TheQuantumPhysicist, line 180: I really don't understand what "above" means... do you mean headers of descedants?

chainstate/src/detail/query.rs
```diff
@@ -177,52 +177,79 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+                    best_height = block_index.block_height();
+                    break;
+                }
+            }
+        }
+
+        self.get_headers_above(best_height, header_count_limit)
+    }
+
+    pub fn get_headers_since_fork_point(
```
TheQuantumPhysicist, line 227: get_headers_not_in_mainchain()?

Generally, we try to avoid using new terms all the time. It can easily create confusion.

TheQuantumPhysicist: Also some documentation would be great. I do understand what this function does based on the name, but generally it's not really clear. Either that, or explain even more in the name.

p2p/backend-test-suite/src/block_announcement.rs
```diff
@@ -33,6 +33,7 @@ use p2p::{
 };
 use utils::atomics::SeqCstAtomicBool;
 
+// FIXME: do these tests actually make any sense?
```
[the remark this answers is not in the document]
ImplOfAnImpl: Well, these are "backend" tests, so they can't really check block announcement, which exists at a higher level. What they do, as I can see, is just send a `SyncMessage::HeaderList` message with some bogus data and check that the other side gets that data.

In the past though they had more sense. E.g. [here](https://github.com/mintlayer/mintlayer-core/pull/557/files#diff-a693256045965766d4941ee10903ec0c74a9f6778dec46460b81f2ac6f2f02b0) is a past version of the `block_announcement` test. At that time, the `make_announcement` function existed on this level, so probably the purpose of the tests was to check that very function.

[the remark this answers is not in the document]
ImplOfAnImpl: FYI: I've removed this FIXME.

chainstate/test-suite/src/tests/syncing_tests.rs
```diff
@@ -129,14 +132,15 @@ fn get_headers(#[case] seed: Seed) {
         let mut last_block_id = tf.genesis().get_id().into();
         last_block_id = tf.create_chain(&last_block_id, blocks_count, &mut rng).unwrap();
 
-        // The locator is from this exact chain, so `get_headers` should return an empty sequence.
+        // The locator is from this exact chain, so get_mainchain_headers_by_locator\
```
TheQuantumPhysicist, line 135: Nitpick: I think there's an extra `\` at the end of the comment?

wallet/wallet-node-client/tests/call_tests.rs
```diff
@@ -51,27 +51,7 @@ pub async fn start_subsystems(
...
-        allow_discover_private_ips: Default::default(),
-        msg_header_count_limit: Default::default(),
-        msg_max_locator_count: Default::default(),
-        max_request_blocks_count: Default::default(),
-        user_agent: common::primitives::user_agent::mintlayer_core_user_agent(),
-        max_message_size: Default::default(),
-        max_peer_tx_announcements: Default::default(),
-        max_unconnected_headers: Default::default(),
-        sync_stalling_timeout: Default::default(),
+        ..p2p::config::P2pConfig::default()
```
TheQuantumPhysicist, line 54: Please don't change these instances. We deliberately keep the list so that it has a side effect on every change and makes developers react to compiler error to make the decision each time.

wallet/wallet-cli-lib/tests/cli_test_framework.rs
```diff
@@ -174,27 +174,7 @@ fn create_chain_config(rng: &mut impl Rng) -> ChainConfig {
...
-        allow_discover_private_ips: Default::default(),
-        msg_header_count_limit: Default::default(),
-        msg_max_locator_count: Default::default(),
-        max_request_blocks_count: Default::default(),
-        user_agent: common::primitives::user_agent::mintlayer_core_user_agent(),
-        max_message_size: Default::default(),
-        max_peer_tx_announcements: Default::default(),
-        max_unconnected_headers: Default::default(),
-        sync_stalling_timeout: Default::default(),
+        ..p2p::config::P2pConfig::default()
```
TheQuantumPhysicist, line 177: Here too

p2p/backend-test-suite/src/block_announcement.rs
```diff
@@ -167,26 +167,10 @@ where
...
-        allow_discover_private_ips: Default::default(),
-        msg_header_count_limit: Default::default(),
-        msg_max_locator_count: Default::default(),
-        max_request_blocks_count: Default::default(),
-        user_agent: mintlayer_core_user_agent(),
-        max_message_size: Default::default(),
-        max_peer_tx_announcements: Default::default(),
-        max_unconnected_headers: Default::default(),
-        sync_stalling_timeout: Default::default(),
+        ..P2pConfig::default()
```
TheQuantumPhysicist, line 173: Here too

p2p/src/peer_manager/peerdb/tests.rs
```diff
@@ -37,28 +35,9 @@ fn unban_peer() {
...
-            allow_discover_private_ips: Default::default(),
-            msg_header_count_limit: Default::default(),
-            msg_max_locator_count: Default::default(),
-            max_request_blocks_count: Default::default(),
-            user_agent: mintlayer_core_user_agent(),
-            max_message_size: Default::default(),
-            max_peer_tx_announcements: Default::default(),
-            max_unconnected_headers: Default::default(),
-            sync_stalling_timeout: Default::default(),
+            ..P2pConfig::default()
```
TheQuantumPhysicist, line 40: Here too

p2p/src/peer_manager/tests/connections.rs
```diff
@@ -617,28 +617,7 @@ async fn connection_timeout_rpc_notified<T>(
...
-        allow_discover_private_ips: Default::default(),
-        msg_header_count_limit: Default::default(),
-        msg_max_locator_count: Default::default(),
-        max_request_blocks_count: Default::default(),
-        user_agent: mintlayer_core_user_agent(),
-        max_message_size: Default::default(),
-        max_peer_tx_announcements: Default::default(),
-        max_unconnected_headers: Default::default(),
-        sync_stalling_timeout: Default::default(),
+        ..P2pConfig::default()
```
TheQuantumPhysicist, line 620: All over the place... I'm gonna stop here

p2p/src/sync/types.rs
```diff
@@ -17,17 +17,40 @@ use std::time::Duration;
...
-        /// A time when the header list request was sent.
-        time: Duration,
-    },
-    /// Node has sent a block list request to a peer and is expecting block responses.
-    ExpectingBlocks {
-        /// A time when either the block list request was sent or last block response was received.
-        time: Duration,
-    },
+pub struct PeerActivity {
+    expecting_headers_since: Option<Duration>,
```
TheQuantumPhysicist, line 21: Nitpick: "since" should be used with a reference point in time, not with a distance in time. So we say "I've been doing that since 2020" or "I've been doing that for 3 years" or "I started doing that 3 years ago". Whichever is the implication. Now feel free to correct my English :smile:. I believe it's a common mistake.

TheQuantumPhysicist: Never mind. I realized this is a timestamp written as duration... and this is why types are important :D

p2p/src/sync/peer.rs
```diff
@@ -388,6 +578,8 @@ where
     /// This is needed to allow the local or remote node to have slightly inaccurate clocks.
     /// Without it, even a 1 second difference can break block synchronization
     /// because one side may see the new block as invalid.
+    // TODO: this must be removed; but at this moment this is not really possible, because
```
TheQuantumPhysicist, line 581: Please reference this issue: https://github.com/mintlayer/mintlayer-core/issues/1024

p2p/src/sync/peer.rs
```diff
@@ -254,22 +386,36 @@ where
...
+            log::warn!(
+                "[peer id = {}] Sending locator of the length {}, which exceeds the maximum length {:?}",
+                self.id(),
+                locator.len(),
+                self.p2p_config.msg_max_locator_count
+            );
+        }
 
-        log::trace!("Sending header list request to {} peer", self.id());
+        log::debug!("[peer id = {}] Sending header list request", self.id());
```
TheQuantumPhysicist, line 401: Btw, I wanna mention a very important point about peer ids, because there's some legacy that has been ruining logging for that. A peer id doesn't really mean anything. Logging with peer id = X is completely meaningless, because that number comes from a counter. The correct behavior is to put identifying information about the peer in the logs, like, for example, their ip address.

TheQuantumPhysicist:
Done through gitlab.

PR #1135, opened 2023-08-03
mempool: Disable during initial block download

iljakuklic:
This also changes slightly how IBD is detected, where block timestamp is taken into account even for genesis.

### DIFF
--- a/chainstate/src/config.rs
+++ b/chainstate/src/config.rs
@@ -33,13 +33,6 @@ make_config_setting!(
 make_config_setting!(TxIndexEnabled, bool, false);
 make_config_setting!(MaxTipAge, Duration, Duration::from_secs(60 * 60 * 24));
 
-impl MaxTipAge {
-    /// Max tip age of 100 years, useful to avoid the IBD state during testing
-    pub const CENTURY: Self = Self {
-        value: Duration::from_secs(100 * 365 * 24 * 60 * 60),
-    };
-}
-
 /// The chainstate subsystem configuration.
 #[derive(Debug, Clone, Default)]
 pub struct ChainstateConfig {
--- a/chainstate/src/config.rs
+++ b/chainstate/src/config.rs
@@ -33,6 +33,13 @@ make_config_setting!(
 make_config_setting!(TxIndexEnabled, bool, false);
 make_config_setting!(MaxTipAge, Duration, Duration::from_secs(60 * 60 * 24));
 
+impl MaxTipAge {
+    /// Max tip age of 100 years, useful to avoid the IBD state during testing
+    pub const CENTURY: Self = Self {
+        value: Duration::from_secs(100 * 365 * 24 * 60 * 60),
+    };
+}
+
 /// The chainstate subsystem configuration.
 #[derive(Debug, Clone, Default)]
 pub struct ChainstateConfig {
--- a/chainstate/src/detail/mod.rs
+++ b/chainstate/src/detail/mod.rs
@@ -755,7 +755,9 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V> 
     /// Returns true if the given block timestamp is newer than `ChainstateConfig::max_tip_age`.
     fn is_fresh_block(&self, time: &BlockTimestamp) -> bool {
         let now = self.time_getter.get_time();
-        time.as_duration_since_epoch() + self.chainstate_config.max_tip_age.clone().into() > now
+        time.as_duration_since_epoch()
+            .checked_add(self.chainstate_config().max_tip_age.clone().into())
+            .map_or(true, |max_tip_time| max_tip_time > now)
     }
 
     /// Update `is_initial_block_download_finished` when tip changes (can only be set once)
@@ -768,12 +770,11 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V> 
 
         // TODO: Add a check for the chain trust.
 
-        let tip_timestamp = match self.query()?.get_best_block_header() {
-            Ok(h) => h.timestamp(),
-            // There is only the genesis block, so the initial block download isn't finished yet.
-            Err(PropertyQueryError::GenesisHeaderRequested) => return Ok(()),
-            Err(e) => return Err(e),
-        };
+        let tip_timestamp = self
+            .query()?
+            .get_best_block_index()?
+            .expect("Best block index should always exist")
+            .block_timestamp();
 
         if self.is_fresh_block(&tip_timestamp) {
             self.is_initial_block_download_finished.set();
--- a/chainstate/test-framework/src/framework_builder.rs
+++ b/chainstate/test-framework/src/framework_builder.rs
@@ -49,6 +49,7 @@ pub struct TestFrameworkBuilder {
     custom_orphan_error_hook: Option<Arc<OrphanErrorHandler>>,
     time_getter: Option<TimeGetter>,
     tx_verification_strategy: TxVerificationStrategy,
+    initial_time_since_genesis: u64,
 }
 
 impl TestFrameworkBuilder {
@@ -68,6 +69,7 @@ impl TestFrameworkBuilder {
         let chainstate_storage = TestStore::new_empty().unwrap();
         let time_getter = None;
         let tx_verification_strategy = TxVerificationStrategy::Default;
+        let initial_time_since_genesis = 0;
 
         TestFrameworkBuilder {
             chain_config,
@@ -76,6 +78,7 @@ impl TestFrameworkBuilder {
             custom_orphan_error_hook: None,
             time_getter,
             tx_verification_strategy,
+            initial_time_since_genesis,
         }
     }
 
@@ -94,6 +97,11 @@ impl TestFrameworkBuilder {
         self
     }
 
+    pub fn with_max_tip_age(mut self, max_tip_age: chainstate::MaxTipAge) -> Self {
+        self.chainstate_config.max_tip_age = max_tip_age;
+        self
+    }
+
     pub fn with_orphan_error_hook(mut self, hook: Arc<OrphanErrorHandler>) -> Self {
         self.custom_orphan_error_hook = Some(hook);
         self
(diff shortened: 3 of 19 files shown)


chainstate/src/detail/mod.rs
```diff
@@ -768,12 +770,11 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
-            Ok(h) => h.timestamp(),
-            // There is only the genesis block, so the initial block download isn't finished yet.
-            Err(PropertyQueryError::GenesisHeaderRequested) => return Ok(()),
-            Err(e) => return Err(e),
-        };
+        let tip_timestamp = self
+            .query()?
+            .get_best_block_index()?
+            .expect("Best block index should always exist")
+            .block_timestamp();
```
iljakuklic, line 777: What do you think about this? It changes genesis handling, so that it takes the timestamp into account rather than making genesis tip to be considered to always imply the initial block download state. This way, genesis and normal blocks are treated uniformly. However, there may be considerations for the different genesis handling that I had not realised.

Without this change, a bunch of mempool tests stops working because transactions get rejected due to IBD state (since the blockchain is at genesis initially).

Any comments @pavel-kokolemin @TheQuantumPhysicist ?

pavel-kokolemin: I think this is a reasonable change.

mempool/src/pool/reorg.rs
```diff
@@ -130,6 +130,21 @@ fn fetch_disconnected_txs<M>(
 pub fn handle_new_tip<M: MemoryUsageEstimator>(mempool: &mut Mempool<M>, new_tip: Id<Block>) {
     mempool.rolling_fee_rate.get_mut().set_block_since_last_rolling_fee_bump(true);
 
+    let is_ibd = mempool
+        .blocking_chainstate_handle()
+        .call(|cs| cs.is_initial_block_download())
+        .expect("Cannot connect to chainstate");
```
ImplOfAnImpl, line 136: Why not propagate the error via '?' instead of expecting?

mempool/src/pool/tests/mod.rs
```diff
@@ -183,13 +184,18 @@ async fn tx_no_inputs() {
 
 // TODO this is copy-pasted from libp2p's test utils. This function should be extracted to an
 // external crate to avoid code duplication
-pub async fn start_chainstate_with_config(
+pub async fn start_chainstate_with_config_and_max_tip_age(
```
ImplOfAnImpl, line 187: Is this TODO still relevant?

mempool/src/pool/tests/mod.rs
```diff
@@ -199,6 +205,12 @@ pub async fn start_chainstate_with_config(
     start_chainstate(chainstate).await
 }
 
+pub async fn start_chainstate_with_config(
+    chain_config: Arc<ChainConfig>,
+) -> subsystem::Handle<Box<dyn ChainstateInterface>> {
+    start_chainstate_with_config_and_max_tip_age(chain_config, MaxTipAge::CENTURY).await
+}
```
ImplOfAnImpl, line 212: It's better to rename this function, because it modifies the passed config in a specific way.

But in general I'd prefer to leave the original `start_chainstate_with_config` as it was, and just pass different configs into it depending on the needs of each particular test.

mempool/src/error/mod.rs
```diff
@@ -85,6 +85,8 @@ pub enum MempoolPolicyError {
 pub enum TxValidationError {
     #[error("Chainstate error")]
     ChainstateError(#[from] ChainstateError),
+    #[error("Not accepting transactions during initial block download")]
+    AddedDuringIBD,
```
ImplOfAnImpl, line 89: A nitpick: this sounds like not accepting the transaction was the actual problem (like we were supposed to accept it but couldn't). Probably something like "Transaction added during initial block download" would be a bit better, at least it'd be more consistent with other error messages in this file.

ImplOfAnImpl (approved):

mempool/src/pool/reorg.rs
iljakuklic: This is an event handler, so it does not return a value.

However, it may be improved somewhat. I think logging the error will be useful.

I am not sure if the expect should stay or be changed to early exit. There is  chance the `expect` will trigger when the node is being shut down. On the other hand, not sure just ignoring it (except for logging) is a very good idea.

mempool/src/pool/tests/mod.rs
ImplOfAnImpl: Ah, I see.
But anyway, this function uses some non-default value for a config parameter that has a specific purpose (preventing the IBD state), so IMO it deserves having a specific name.

I don't insist on it though.

chainstate/src/config.rs
```diff
@@ -33,6 +33,13 @@ make_config_setting!(
 make_config_setting!(TxIndexEnabled, bool, false);
 make_config_setting!(MaxTipAge, Duration, Duration::from_secs(60 * 60 * 24));
 
+impl MaxTipAge {
+    /// Max tip age of 100 years, useful to avoid the IBD state during testing
+    pub const CENTURY: Self = Self {
+        value: Duration::from_secs(100 * 365 * 24 * 60 * 60),
+    };
```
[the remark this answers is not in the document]
iljakuklic: OK, will find a better location for it, to make it specific to tests.

chainstate/src/detail/mod.rs
```diff
@@ -755,7 +755,9 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
     /// Returns true if the given block timestamp is newer than `ChainstateConfig::max_tip_age`.
     fn is_fresh_block(&self, time: &BlockTimestamp) -> bool {
         let now = self.time_getter.get_time();
-        time.as_duration_since_epoch() + self.chainstate_config.max_tip_age.clone().into() > now
+        time.as_duration_since_epoch()
+            .checked_add(self.chainstate_config().max_tip_age.clone().into())
+            .map_or(true, |max_tip_time| max_tip_time > now)
```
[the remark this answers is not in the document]
iljakuklic: There's a test in blockprod which sets genesis timestamp to `u64::MAX` and does some operations on it. That triggers an overflow together with the change to calculate IBD based on timestamp even for genesis. Ultimately, the block timestamp comes from an untrusted source. It should have been checked by the time it gets to this point but I don't think there's a harm in handling the overflow possibility here as well.

chainstate/test-framework/src/framework_builder.rs
```diff
@@ -94,6 +97,11 @@ impl TestFrameworkBuilder {
         self
     }
 
+    pub fn with_max_tip_age(mut self, max_tip_age: chainstate::MaxTipAge) -> Self {
```
TheQuantumPhysicist, line 100: Documentation would be nice

chainstate/test-framework/src/framework_builder.rs
```diff
@@ -109,13 +117,19 @@ impl TestFrameworkBuilder {
         self
     }
 
+    pub fn with_initial_time_since_genesis(mut self, initial_time_since_genesis: u64) -> Self {
```
TheQuantumPhysicist, line 120: Here too

chainstate/src/detail/mod.rs
TheQuantumPhysicist: I'm OK with this.

TheQuantumPhysicist (approved):
No complaints from my side. Looks good.

TheQuantumPhysicist:
Can we add functional tests though? How hard would that be?

iljakuklic:
> Can we add functional tests though? How hard would that be?

Not sure, the existing tests for this functionality all rely on the ability to mock time.

TheQuantumPhysicist:
> Not sure, the existing tests for this functionality all rely on the ability to mock time.

You can mock time in functional tests too. There's a command line argument for the initial time, and there's an rpc function to set the mock time. You can give it a shot. I don't believe any of us used it before though.

mempool/src/pool/tests/mod.rs
iljakuklic: > Ah, I see. But anyway, this function uses some non-default value for a config parameter that has a specific purpose (preventing the IBD state), so IMO it deserves having a specific name.

I did not end up changing the name, basically for two reasons:
* I did not find a name that would be both sufficiently descriptive and concise. Instead, I added a comment pointing out it uses some non-default values, which I think is a reasonable compromise.
* Vast majority of mempool tests (all except ones that specifically test IBD handling) rely on IBD being disabled. So in context of mempool testing, the huge max tip age is de-facto default anyway.

Also the two versions of the `start_chainstate_with_*` method are a residue of the development that is no longer necessary, so there's now only one version.

alfiedotwtf (approved):

mempool/src/pool/reorg.rs
iljakuklic: > But I'd do it on the highest level possible, which is, as I understand, the `run` method of `impl MempoolSubsystemInterface for MempoolInit`. And all functions that are called by `run` directly or indirectly IMO should just return `Result`.

Good point. Done.

iljakuklic:
> > Not sure, the existing tests for this functionality all rely on the ability to mock time.
> 
> You can mock time in functional tests too. There's a command line argument for the initial time, and there's an rpc function to set the mock time. You can give it a shot. I don't believe any of us used it before though.

A basic functional test is now there

TheQuantumPhysicist (approved):

mempool/src/pool/reorg.rs
```diff
@@ -127,16 +127,33 @@ fn fetch_disconnected_txs<M>(
...
+    mempool: &mut Mempool<M>,
+    new_tip: Id<Block>,
+) -> Result<(), ReorgError> {
     mempool.rolling_fee_rate.get_mut().set_block_since_last_rolling_fee_bump(true);
 
+    let is_ibd = mempool.blocking_chainstate_handle().call(|cs| cs.is_initial_block_download())?;
+    if is_ibd {
+        log::debug!("Not updating mempool tx verifier during IBD");
+
+        // We still need to update the current tx_verfier tip
```
TheQuantumPhysicist, line 140: tx_verfier

Issue #1002, opened 2023-06-28
Genesis ongoing consensus status is incorrect


alfiedotwtf:
Genesis needs to have it's consensus set in it's `NetUpgrade` as `IgnoreConsensus` because it is calculated as "ongoing" otherwise, which is incorrect.

**Incorrect**: _The fix here should make Genesis set as `Threshold` rather than the current error of `GenesisCannotHaveOngoingDifficulty`_

The fix is to return the initial chain difficulty when the previous block is Genesis.

PR #1136, opened 2023-08-04
Return initial difficulty for when the previous block is Genesis

alfiedotwtf:
More info in #1002


alfiedotwtf:
The failed test in CI was `cancel_received()` in `blockprod/src/detail/test.rs`. This test creates a new Genesis with an initial difficulty of zero...

The 57ee633 commit was supposed to return the chain config's current PoW limit i.e zero, however after following the call stack, it was never using the overridden value, as seen in `get_proof_of_work_config()` from commit e2b9d75.

These last few commits creates a `PoWChainConfig` builder which allows the caller to override chain settings -as is done now in `get_proof_of_work_config()` to override the difficulty by looking at `NetUpgrades` rather than straight to the default value.

common/src/chain/config/mod.rs
```diff
@@ -454,10 +456,27 @@ impl ChainConfig {
         self.empty_consensus_reward_maturity_distance
     }
 
-    // TODO: this should be part of net-upgrades. There should be no canonical definition of PoW for any chain config
     #[must_use]
-    pub const fn get_proof_of_work_config(&self) -> PoWChainConfig {
-        PoWChainConfig::new(self.chain_type)
+    pub fn get_proof_of_work_config(&self) -> PoWChainConfig {
```
TheQuantumPhysicist, line 460: Two points:

1. Why does this fix apply to PoW, not PoS?
2. This function was made const because this is called directly repeatedly in ChainConfig, which makes it expensive to create.

alfiedotwtf: For PoS in `calculate_target_required()`, in both `Threshold` or `Ongoing` we have either an `initial_difficulty` or `PoSChainConfig` to get the target value. However for PoW in `calculate_work_required(), we only have `initial_difficulty` for `Threshold` and nothing to go by for `Ongoing` and so have to trace back through the block index... but for Genesis since there we had no information, before this PR we returned a `ConsensusPoWError::GenesisCannotHaveOngoingDifficulty` error.

So instead of returning the error for PoW at Genesis-Ongoing, I initially created a new `PoWChainConfig` to return its `limit`... yet that above unit test was failing. Only after digging in I found that `PoW::new()` that returns a `PoWChainConfig` wasn't actually looking at the passed in `chain_config` and instead was just using the default values whereas my failing test wants to override these values.

Hopefully that all makes sense.

On 2, yeah I guess we need to keep it const... I'll have to rethink this one.

TheQuantumPhysicist: I see.

Either find a way to make it const, or change the way it's used to avoid constantly reconstructing the configuration. I hope PoS doesn't have the same problem for its configuration that it just keeps getting reconstructed from a non-const function.

It's a bad design that was accepted at the very beginning of the project just to get things moving. I don't like depending on things that are ambiguous to performance and may or may not be optimized.

alfiedotwtf: For PoS we're not reconstructing it each time, but we are deconstructuring `pos_status`. But since we have the config for each arm, we're doing minimal work to get the target limit. Not `const`, but also too bad.

Also I don't think it's that much of an issue because these functions we're trying to optimise for here are only when the previous block is Genesis... 99% of the time it won't be, and we'll need to do heavy lifting by looking at the block index anyway, so I think we're focusing on optimising the wrong thing here, unless I'm wrong?

If performance is an issue, we can think about doing something like pre-calculating once at each epoch, or when blocks come in over the wire etc, and then using those values etc.

TheQuantumPhysicist (approved):

common/src/chain/config/mod.rs
```diff
@@ -454,10 +455,9 @@ impl ChainConfig {
         self.empty_consensus_reward_maturity_distance
     }
 
-    // TODO: this should be part of net-upgrades. There should be no canonical definition of PoW for any chain config
```
azarovh, line 457: This TODO is still valid: PoWChainConfig should not be a part of ChainConfig but rather net-upgrades so that we could change the parameters.

common/src/chain/config/builder.rs
```diff
@@ -125,6 +126,7 @@ pub struct Builder {
     max_block_size_with_smart_contracts: usize,
     max_no_signature_data_size: usize,
     max_depth_for_reorg: BlockDistance,
+    pow_chain_config_builder: PoWChainConfigBuilder,
```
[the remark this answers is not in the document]
alfiedotwtf: Ah, you're right. Thanks

PR #1138, opened 2023-08-05
Create a docker build script

erubboli:
This script build and pushes the docker images, it retrieve the version from the Cargo file and can be invoked by make as `build_docker_images`.


alfiedotwtf (approved):

muursh (approved):
I've half tested this. I have no issues with what I see.

build-tools/docker/build.py
```diff
@@ -0,0 +1,99 @@
...
+    command = f"docker build -t {image_name}:{version} -f {dockerfile_path} ."
+
+    # Run the command
+    process = subprocess.Popen(command, shell=True)
+    output, error = process.communicate()
+
+    if error:
+        print(f"Error occurred: {error}")
+    else:
+        print(f"Built {image_name}:{version} successfully.")
```
ImplOfAnImpl, line 33: Shouldn't we stop the script if `docker build` fails? (i.e. why not just use `check_call` here like in other places?)

build-tools/docker/Dockerfile.builder
```diff
@@ -0,0 +1,8 @@
+FROM rust AS builder
+
+WORKDIR /usr/src/
+
+COPY . .
+
+ARG NUM_JOBS=1
```
ImplOfAnImpl, line 7: At this moment it's impossible to specify this value when using `build.py`. I'd be nice if the latter had an additional parameter for this.

ImplOfAnImpl (approved):

Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+# python scripts are requiring `python3.11-venv` and `python3-pip` installed
+install-pip-dependencies:
+	python3 -m venv env && . env/bin/activate && python3 -m pip install --upgrade pip && python3 -m pip install toml
+
+build_docker_images: install-pip-dependencies
+	. env/bin/activate && python3 build-tools/docker/build.py
+
+push_docker_images: install-pip-dependencies
+	. env/bin/activate && python3 build-tools/docker/build.py --build=false --push --latest
```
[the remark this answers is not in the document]
erubboli: I'll put the dependencies in the readme

PR #1141, opened 2023-08-05
Options add host resolution - remove websocket options

erubboli:
The objective is to leverage `std::net::ToSocketAddrs` instead of `SocketAddr`. This approach enables the use of a hostname for the RPC, which is particularly beneficial for docker container networking. With this method, one container can connect to another without the necessity of knowing the assigned IP.

NOTE: This PR also remove the unnecessary `ws` option settings.

### DIFF
--- a/node-lib/src/options.rs
+++ b/node-lib/src/options.rs
@@ -20,6 +20,8 @@ use std::{ffi::OsString, net::SocketAddr, num::NonZeroU64, path::PathBuf};
 use clap::{Args, Parser, Subcommand};
 use common::chain::config::ChainType;
 use p2p::types::ip_or_socket_address::IpOrSocketAddress;
+use std::io::{Error, ErrorKind};
+use std::net::ToSocketAddrs;
 use utils::default_data_dir::default_data_dir_common;
 
 use crate::{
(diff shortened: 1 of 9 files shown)


node-lib/src/options.rs
```diff
@@ -176,6 +178,48 @@ pub struct RunOptions {
     pub rpc_cookie_file: Option<String>,
 }
 
+impl RunOptions {
```
TheQuantumPhysicist, line 181: Two things:

1. The resolution should be part of RPC, not part of RunOptions. So this should be done in the `rpc` crate.
2. It seems that the resolution process is exactly the same for both http and websocket. There should be a way to abstract both calls into one.

Though I want to add that I'm planning to remove websocket at some point because it's kind of pointless. We decided to go with polling instead of pub/sub for the sake of simplicity, which is the way the wallet currently works. So feel free to eliminate websocket as well to simplify this task too.

pavel-kokolemin:
I tried this PR, but it does not allow connecting by hostname:
```
$ cargo run --bin wallet-cli -- --rpc-address=localhost:3030                                    
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/wallet-cli '--rpc-address=localhost:3030'`
error: invalid value 'localhost:3030' for '--rpc-address <RPC_ADDRESS>': invalid socket address syntax
```

Instead, it allows to bind to a hostname, which is not very useful:
```
$ cargo run --bin node-daemon -- testnet --http-rpc-addr=localhost:3030 
```

PR #1142, opened 2023-08-06
Initial implementation of the mintlayer-core API server

TheQuantumPhysicist:
The API server plays the role of a server that translates the blockchain data into data retrievable using an http server. This can be used for both a block explorer and a light wallet.

Design details (feel free to critique this):
- There are two components, the scanner and the web-server
- The scanner communicates with a node instance through RPC (similar to how the wallet works) and updates a storage instance
- The web-server communicates with the same database to retrieve data requested by the user. The web-server only delivers restful API access. The frontend is not included and depends on the application.
- The aim is to use an SQL server for this process due to the obvious emphasis on scalability through reading, where a single storage node (the master) is needed to update many distributed database instances (slaves)
- In this eco-system, there should be exactly one scanner that updates a database (or a cluster of master/slave databases), and as many web-server instances as desirable, where all of them use the slaves (and possibly the master, depending on load-balancing details) to read the blockchain data, where the master gives write access only to the scanner that updates it.


alfiedotwtf:
Looks good.

The only thing I would say so far is that we should probably not worry to much about the database architecture for now apart from having two distinct configs - one for the scanner, and the other for the web servers. On a simple installation, they'll both point to the same database, but on a larger install (e.g. at an exchange), they would most likely have their own individual database network topology which we shouldn't care about... as long as each web server can be configured to point to **_a_** slave, and the scanner can be pointed to **_a/the_** master, then we should be able to work with all cases.

TheQuantumPhysicist:
> we should probably not worry to much about the database architecture for now apart from having two distinct configs

> as long as each web server can be configured to point to a slave, and the scanner can be pointed to a/the master

That's the plan.

TheQuantumPhysicist:
> > Design details (feel free to critique this)
> 
> How do lightweight wallets know about new transactions? Periodic polling is inefficient and slow. What happens if something goes wrong with the scanner process? The whole system would stop working properly.
> 
> If it were up to me, I would use a simpler design:
> 
> * A single process that scans the network using the node RPC and caches some information in memory.
> * WebSocket API for explorers and lightweight wallets.
> 
> Any number of such servers can be deployed (for redundancy and scalability).

I'm not sure I understand the full plan. Let's discuss it tomorrow in the daily.

TheQuantumPhysicist:
We discussed. Conclusion:

- So about websocket for transactions, we have the same problem with the wallet-cli, but we'll come to that problem later and maybe add pub/sub through websocket or otherwise

- Pavel's concern was that some transactions should be kept in memory, and hence it's reasonable to combine scanners and web-servers instead of separating them. But this means that transactions will remain only in memory, which is not the desired behavior for the api server. We want to (at least) store stale transactions for a while, which means they have to go to the database.

pavel-kokolemin:
> We discussed. Conclusion...

I would say that the described architecture is good for a web explorer, but not ideal for lightweight wallets where we want real-time updates. IMO the easiest way to get real-time updates is to implement the new API as a single web server/scanner process (and use WebSocket for transport instead of REST).

api-server/api-server-common/src/storage/impls/in_memory/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+        Ok(storage_version_handle.is_some())
+    }
+
+    fn get_block(&self, block_id: Id<Block>) -> Result<Option<Block>, ApiStorageError> {
+        let block_result = self.block_table.get(&block_id);
+        let block = match block_result {
+            Some(blk) => blk,
+            None => return Ok(None),
+        };
+        Ok(Some(block.clone()))
```
iljakuklic, line 61: Minor coding suggestion. This could be reduced to:
```rust
Ok(self.block_table.get(&block_id).cloned())
```

There are multiple occurrences of this pattern in this file.

api-server/api-server-common/src/storage/impls/in_memory/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+use std::collections::BTreeMap;
+
+use common::{
+    chain::{Block, ChainConfig, GenBlock, SignedTransaction, Transaction},
+    primitives::{BlockHeight, Id},
+};
+
+use crate::storage::storage_api::{block_aux_data::BlockAuxData, ApiStorageError};
+
+pub struct ApiInMemoryStorage {
```
iljakuklic, line 29: I find the `Api` prefix in the name a bit weird, since it actually implements the database. Something with `Api` in name I'd expect to represent merely an interface.

api-server/api-server-common/src/storage/impls/in_memory/transactional/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    pub fn tx_ro(&self) -> RwLockReadGuard<'_, ApiInMemoryStorage> {
+        self.storage.read().expect("Poisoned mutex")
+    }
+
+    pub fn tx_rw(&mut self) -> RwLockWriteGuard<'_, ApiInMemoryStorage> {
+        self.storage.write().expect("Poisoned mutex")
+    }
```
iljakuklic, line 86: These are `pub` but return lock guards, which look like an implementation detail. Is that intentional?

api-server/api-server-common/src/storage/impls/in_memory/transactional/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+
+pub mod read;
+pub mod write;
+
+pub struct ApiInMemoryStorageTransactionalRo<'t> {
+    transaction: RwLockReadGuard<'t, ApiInMemoryStorage>,
+}
+
+impl<'t> ApiInMemoryStorageTransactionalRo<'t> {
+    pub fn new(storage: &'t ThreadSafeApiInMemoryStorage) -> Self {
```
iljakuklic, line 34: This constructor (and similar for `Rw`) can be private. It's still accessible through the `Transactional` trait. Probably better to have a canonical way of creating these objects.

iljakuklic (commented):
Only went over the storage module so far. At first glance, the overall architecture looks sensible to me. Some minor implementation comments below.

api-server/api-server-common/src/storage/impls/in_memory/mod.rs
TheQuantumPhysicist: I understand... though this is at worst temporary because honestly I don't know what to call this and I need to progress... ApiServer is better?

api-server/api-server-common/src/storage/impls/in_memory/mod.rs
TheQuantumPhysicist: I do have the preference to have a clean value of the result at the end. Helps with debugging.

api-server/api-server-common/src/storage/impls/in_memory/transactional/mod.rs
TheQuantumPhysicist: Not really. Removed. Thanks.

api-server/api-server-common/src/storage/impls/in_memory/transactional/mod.rs
TheQuantumPhysicist: Done.

api-server/api-server-common/src/storage/impls/in_memory/mod.rs
iljakuklic: >  ApiServer is better?

Yes

TheQuantumPhysicist: Done

TheQuantumPhysicist:
Done through gitlab.

PR #1143, opened 2023-08-07
Set RUST_BACKTRACE by default

pavel-kokolemin:
I think it will be a little easier for everyone if we set `RUST_BACKTRACE` by default for our executables.
This can be useful for regular GUI users (but only if we also use file logging and log crashes).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3894,6 +3894,7 @@ dependencies = [
  "logging",
  "node-lib",
  "tokio",
+ "utils",
 ]
 
 [[package]]
@@ -6979,6 +6980,7 @@ version = "0.1.1"
 dependencies = [
  "clap",
  "tokio",
+ "utils",
  "wallet-cli-lib",
 ]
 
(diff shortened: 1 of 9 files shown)


ImplOfAnImpl (approved):
IMO the "rust_" part of the module name is redundant, I'd remove it.

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
> IMO the "rust_" part of the module name is redundant, I'd remove it. 

It's helpful to keep it to recognize that this is something that's language specific, and not application specific.

iljakuklic (approved):

ImplOfAnImpl:
> > IMO the "rust_" part of the module name is redundant, I'd remove it.
> 
> It's helpful to keep it to recognize that this is something that's language specific, and not application specific.

Well, by this logic we'll probably have to use this prefix for anything that is a wrapper for standard library stuff, which sounds like an overkill.

TheQuantumPhysicist:
> Well, by this logic we'll probably have to use this prefix for anything that is a wrapper for standard library stuff, which sounds like an overkill.

Not really. I understand why you think like that, but when looking at utils, it's then obvious that this is something different. It's not really a wrapper, it's a configuration of the compiler. 

This is my opinion, and quite frankly doesn't matter that much.

PR #1144, opened 2023-08-07
Switch to non-generic p2p addresses

pavel-kokolemin:
This is a continuation of the address refactoring PR (https://github.com/mintlayer/mintlayer-core/pull/1117).

I think the code should be a bit cleaner now, and allow for future improvements.

There is some overlap between `PeerAddress` and `SocketAddress` (they both currently represent a TCP socket address), but I think they should be kept separate. `PeerAddress` is something you can connect to to reach a peer, and `SocketAddress` is just a wrapper for OS socket address. Also, in the future, `PeerAddress` may contain other peer addresses, like Tor.


TheQuantumPhysicist (approved):
Looking good. No big objections here.

PR #1152, opened 2023-08-09
Mempool: remove disconnected orphans

iljakuklic:
Remove orphan transactions sent by a peer when the peer disconnects.

I am not satisfied with the way the events are wired up. Since p2p already listens for mempool events (orphans being processed) and now mempool listens to p2p events, there is a bit of circularity. Usually, the subsystem subscribes to events it wants to listen to during its setup. However in this case, p2p setup code subscribes mempool to p2p events. Another unfortunate side effect of this is that mempool handler that processes p2p events is now exposed via public API.

A follow up PR that addresses these issues is in the works. It involves separating p2p interface and mempool interface into their own crates. As a side effect, we get a better interface / implementation separation for these two subsystems at crate level.


mempool/src/interface/mempool_interface.rs
```diff
@@ -72,6 +72,12 @@ pub trait MempoolInterface: Send + Sync {
     /// Get the fee rate such that it would put the new transaction in the top X MB of the mempool
     /// making it less likely to get rejected or trimmed in the case the mempool is full
     fn get_fee_rate(&self, in_top_x_mb: usize) -> Result<FeeRate, Error>;
+
+    /// Handle a p2p event.
+    // TODO This is here because of awkward inter-crate dependencies. A nicer solution would be to
+    // factor out subsystem interface traits into separate crates and subscribe to p2p events
+    // together with the other mempool setup code
+    fn handle_p2p_event(&mut self, evt: p2p_types::p2p_event::P2pEvent);
```
iljakuklic, line 80: Not very happy about exposing this via a public interface. See the PR description for details.

iljakuklic:
> I am not satisfied with the way the events are wired up. Since p2p already listens for mempool events (orphans being processed) and now mempool listens to p2p events, there is a bit of circularity. Usually, the subsystem subscribes to events it wants to listen to during its setup. However in this case, p2p setup code subscribes mempool to p2p events. Another unfortunate side effect of this is that mempool handler that processes p2p events is now exposed via public API.

This has now been changed to a subsystem call rather than event subscribtion. The issue that the event handler for processing peer disconnect events by mempool is now a part of the public API still remains.

> A follow up PR that addresses these issues is in the works. It involves separating p2p interface and mempool interface into their own crates. As a side effect, we get a better interface / implementation separation for these two subsystems at crate level.

Abandoning the approach mentioned here for now.

pavel-kokolemin (approved):

test/functional/mempool_orphan_peer_disconnected.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017-2021 The Bitcoin Core developers
```
muursh, line 2: Is this license correct?

iljakuklic: Uh, not sure. I started by copying another functional test so we multiple of these

iljakuklic: Changed this to the usual license but as I said, there still are tests with this

muursh: For some tests where we've taken (and modified) the original Bitcoin test it probably makes sense to have the original Bitcoin license header and then "update" it to add a Mintlayer-y bit to cover the work we've added. Where it's not a test derived from Bitcoin (like this one? I couldn't see a comparable test in Bitcoin core) I guess our license alone is the way to go. It probably means some of the other files need licenses updated. I'll go through the functional tests this arvo and have a look.

muursh (approved):
I have no qualms with this

p2p/types/src/peer_id.rs
```diff
@@ -41,6 +41,10 @@ impl PeerId {
         let id = NEXT_PEER_ID.fetch_add(1, Ordering::Relaxed);
         Self(id)
     }
+
+    pub fn from_u64(n: u64) -> Self {
```
TheQuantumPhysicist, line 45: This is kind of nit-picky, but if you just need this for tests to initialize a random value, you can have a function like `random_using(...)`

mocks/Cargo.toml
```diff
@@ -16,6 +16,7 @@ pos_accounting = { path = '../pos_accounting/' }
 subsystem = { path = "../subsystem/" }
 utils = { path = "../utils/" }
 utxo = { path = "../utxo/" }
+p2p-types = { path = "../p2p/types" } # TODO(PR) remove
```
TheQuantumPhysicist, line 19: I don't think the python script triggers on Cargo files... maybe we should tweak the script to do that.

TheQuantumPhysicist (approved):

Though the main message of the PR may need to be updated. I believe you're satisfied with the current state :-)

p2p/types/src/peer_id.rs
iljakuklic: The issue is that I use this in a simulation test that performs all sorts of random operations on the orphan storage data structure. To ensure good coverage, `PeerId`s are generated from a carefully selected random distribution. That distribution may not be suitable for other use cases, testing or otherwise, so not a good idea to move the generation here IMO.

mempool/src/pool/orphans/mod.rs
```diff
@@ -263,6 +273,27 @@ impl TxOrphanPool {
...
+    }
+
+    /// Pick one orphan from given origin
+    fn pick_by_origin(&self, origin: TxOrigin) -> Option<InternalId> {
+        self.maps
+            .by_origin
+            .range((origin, InternalId::ZERO)..=(origin, InternalId::MAX))
+            .map(|(_origin, iid)| *iid)
+            .next()
+    }
```
alfiedotwtf, line 296: Not sure if it's worth it or not, but if we're likely to have many transactions from peers, maybe have another level here - so that on cleanup you can just delete the whole origin in a single call rather than iterating?

p2p/types/src/peer_id.rs
TheQuantumPhysicist: Alrighty. No worries.

alfiedotwtf (approved):

mempool/src/pool/orphans/mod.rs
iljakuklic: There are not going to be many transactions from one peer. The whole orphan pool is limited to 100 transactions by default so there's your upper bound. Will be even more restricted once some sort of rate limiting is implemented

iljakuklic: In addition to that, you still need to iterate to get it out of all the other indices

mocks/Cargo.toml
iljakuklic: Removed the TODO, the dependency is needed with this approach.

Issue #756, opened 2023-03-11
Allow generate_block() to accept Transaction IDs along with encoded signed transactions


alfiedotwtf:
With #755, generate_block() allows the caller to create blocks via RPC. However, they will need to supply encoded signed transactions. This issue is a note to replicate Bitcoin's functionality, allowing the caller to accept transactions **or** transaction IDs (which refer to transactions in the mempool).

See https://github.com/mintlayer/mintlayer-core/pull/755 for more detail.

TheQuantumPhysicist:
@alfiedotwtf Can you please check what's the status with this? You said there's something to be covered here, but I think this may be half-done.

alfiedotwtf:
I think we have a solution, so now it's just a prioritization issue now.

Copying from GItlab's issue:

@alfiedotwtf :

During yesterday's meeting, it was decided that to prevent the issue of double spend and other potential issues, the supplied transaction hex should be sent through a transaction verifier before calling collect_tx()...
But before I do this change, I want to confirm something - after going through the transaction verifier code, I'm not sure which TransactionSourceForConnect the supplied transaction hex should fall under for the connect_transaction() call - There's currently TransactionSourceForConnect::Chain and TransactionSourceForConnect::Mempool, neither of which feels right to sneak these transaction hex into. So in this case, should I add a TransactionSourceForConnect::Provided arm?

@iljakuklic :

I would say `Mempool` is more appropriate here out of the two. Maybe it could be renamed to something more descriptive that covers both cases. It basically says this is a candidate for inclusion in the next block (as opposed a transaction already included in a block).

iljakuklic:
Closing this, mostly implemented in #1270. Remaining issues described in #1273

PR #1158, opened 2023-08-10
Allow generate_block() to accept Transaction IDs along with encoded signed transactions

alfiedotwtf:
This adds two new parameters to `generate_block()`:

  - transaction_ids so that it can include specific transaction Ids in the block
  - include_mempool to specify whether or not to include transactions from the mempool

See #756 for more details.


blockprod/src/detail/mod.rs
```diff
@@ -180,19 +180,51 @@ impl BlockProduction {
...
+                self.chain_config.max_block_size_from_std_scripts(),
+                current_tip,
+                min_block_timestamp,
+            ));
+
+        for transaction in transactions.into_iter() {
+            let transaction_id = transaction.transaction().get_id();
+            // TODO: fees
+            accumulator
+                .add_tx(transaction, Amount::ZERO.into())
```
alfiedotwtf, line 198: @iljakuklic I wasn't too sure what the fees should be here?

blockprod/src/detail/mod.rs
```diff
@@ -180,19 +180,51 @@ impl BlockProduction {
...
-                    max_block_size,
-                    current_tip,
-                    min_block_timestamp,
-                )))
+            .call({
+                move |this| -> Result<_, BlockProductionError> {
+                    for transaction_id in transaction_ids.iter() {
+                        if let Some(transaction) = this.transaction(transaction_id) {
+                            // TODO: fees
+                            accumulator.add_tx(transaction, Amount::ZERO.into()).map_err(|err| {
```
alfiedotwtf, line 209: Here too

blockprod/src/detail/mod.rs
```diff
@@ -180,19 +180,51 @@ impl BlockProduction {
...
+                                BlockProductionError::FailedToAddTransaction(*transaction_id, err)
+                            })?
+                        } else {
+                            // TODO: shoud this error out rather than ignore missing transactions?
+                        }
```
alfiedotwtf, line 214: Wasn't too sure if this should just ignore missing transactions, or this actually is an error.

blockprod/src/detail/mod.rs
alfiedotwtf: Also to note, there's a difference between `add_tx()` and `collect_txs()` which I'm not sure about in terms of safety - `collect_txs()` sets up a transaction verifier but add_tx() doesn't. Since the transactions for `add_tx()` will be supplied by the user, I'm thinking it would be a good idea to do some checking here.

If so, should this be done manually before `collect_txs()`, or should it actually be done inside `collect_txs()`?

blockprod/src/detail/mod.rs
```diff
@@ -180,19 +180,51 @@ impl BlockProduction {
         &self,
         current_tip: Id<GenBlock>,
         min_block_timestamp: BlockTimestamp,
+        transactions: Vec<SignedTransaction>,
+        transaction_ids: Vec<Id<Transaction>>,
+        include_mempool: bool,
```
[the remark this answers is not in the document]
alfiedotwtf: Yep, I think this is the issue I'm facing right now in the functional tests. Ok, I'll rethink this.

blockprod/src/detail/mod.rs
alfiedotwtf: > Since the transaction sequence is already passed in, I don't see why the mempool stuff should be involved at all.

By having the option of supplied transactions, supplied mempool transaction Ids, and also fill from random mempool transactions, I think that should cover all scenarios that a user may want... but yes, this does make dependency resolution a bit messy.

Doing it outside of mempool is doable, but mempool's `collect_txs()` function does a few handy things that I would need to duplicate i.e. timelocks, parent/child resolution etc.

What are your thoughts about the way to go about the design? I was thinking I could split `collect_txs()` into three functions - one internal that does all the work, one to keep the same interface as we have now for the current callers, and another which blockprod uses (which passes in a `Vec<Transaction>`, creates a tx verifier, then calls the internal function), and so there is less duplication... unless I'm misunderstanding and there is a better way to go about this?

alfiedotwtf: Thinking about this from another angle, for the first two parameters, the caller is already signalling their desired transaction sequence (both supplied and mempool txid), and as this is explicit, we should expect they have already done the resolution. So I just need to use a tx verifier here and gather within an accumulator. Then I can pass this into `collect_txs()` which will then fill the rest of the space with random mempool transactions.

Does this makes better sense to you?

alfiedotwtf: > Be aware that the accumulator already assumes the transactions have been verified by the mempool. Fees are extracted from the verifier.

Looking at the transaction verifier, I'm not sure that this is going to be a problem unless I'm mistaken - the is only used to build the block which is then returned to the caller. They still need to submit the block, which will then have to go through the transaction verifier on the round trip. Thoughts?

iljakuklic: > > Since the transaction sequence is already passed in, I don't see why the mempool stuff should be involved at all.
> 
> By having the option of supplied transactions, supplied mempool transaction Ids, and also fill from random mempool transactions, I think that should cover all scenarios that a user may want... but yes, this does make dependency resolution a bit messy.
> 
> Doing it outside of mempool is doable, but mempool's `collect_txs()` function does a few handy things that I would need to duplicate i.e. timelocks, parent/child resolution etc.
> 
> What are your thoughts about the way to go about the design? I was thinking I could split `collect_txs()` into three functions - one internal that does all the work, one to keep the same interface as we have now for the current callers, and another which blockprod uses (which passes in a `Vec<Transaction>`, creates a tx verifier, then calls the internal function), and so there is less duplication... unless I'm misunderstanding and there is a better way to go about this?

Ok, I see your points here. Probably the most straightforward, albeit probably not the cleanest, would be to pass an additional argument to `collect_txs` specifying all the "extra" transactions the user wants. It would then use these transactions as a matter of priority. We can see how bad it is and refine it further. Maybe the transaction accumulator machinery could be split off and generalized to support multiple "sources", mempool being one, user-supplied transactions being another.

Also note that mempool already has some information regarding transactions given by ID, i.e. the dependency info and fee rates. When looking up transaction by ID, you only get a `SignedTransaction`, so that information is lost. It is, however, quite useful while constructing blocks.

blockprod/src/detail/mod.rs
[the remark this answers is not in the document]
iljakuklic: I think the problem of relative ordering of transactions given by hex vs given by ID is easily solved by introducing an enum that is either a transaction or an id, and having one vector with the enum as values.

How will these interleave with transactions from mempool is a different question. One possibility is as now: mempool transactions go at the end. If the user wants to include a transaction that depends on a mempool transaction, the IDs of its dependencies have to be explicitly listed. Would it be good enough, at least for now?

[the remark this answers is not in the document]
alfiedotwtf: >  If the user wants to include a transaction that depends on a mempool transaction

hmm... unless this was before I started changing things and I'm getting confused, including transaction hex followed by transaction Ids (so that they're pulled from the mempool) that had inputs from those transaction hexes didn't work. From the error message, I think this was because when I inserted the transactions into the mempool they were going into the orphan pool and so `generate_block()` was failing.

So is that above scenario supposed to work?

> the IDs of its dependencies have to be explicitly listed

Yep, that's what I was thinking given we can assume that the caller knows what they're doing because they're overriding the defaults.

blockprod/src/detail/mod.rs
alfiedotwtf: So after some digging in order to find how the fees are set, I've found that the verifier needs the store (either chain or mempool) - both of which aren't the source for transactions that are provided over the wire...

So I'm now thinking that if we take transactions in hex, we should probably add them to the mempool. That also gives the benefit of only needing to add a "priority_transaction_ids" `Vec` to `collect_txs()`.

blockprod/src/detail/mod.rs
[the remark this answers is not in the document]
iljakuklic: We also need to be careful that the transaction that's passed in by the user as ID or in hex is not included again from mempool. That would lead to an invalid block due to double spending

alfiedotwtf:
As discussed last week with @iljakuklic, we'll do the following:

- Include transaction hex within the `TxAccumulator`, which will not effect the mempool
- Include transaction Ids within the `collect_txs()` call to take advantage of existing code
- Fill in the remaining space with random mempool transactions if specified

blockprod/src/interface/blockprod_interface.rs
```diff
@@ -35,6 +38,8 @@ pub trait BlockProductionInterface: Send {
     async fn generate_block(
         &mut self,
         input_data: GenerateBlockInputData,
-        transactions: Option<Vec<SignedTransaction>>,
+        transactions: Vec<SignedTransaction>,
+        transaction_ids: Vec<Id<Transaction>>,
```
TheQuantumPhysicist, line 42: Please update the docs for this function

mempool/src/pool/mod.rs
```diff
@@ -968,9 +970,35 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
             .expect("best index to exist");
         let tx_source = TransactionSourceForConnect::for_mempool(&best_index);
 
+        let mut dedupe_txids = BTreeSet::new();
+        let mut ordered_txids = vec![];
+
+        for transaction in tx_accumulator.transactions().iter() {
+            // Use transactions already in the Accumulated for
+            // deduping only i.e don't send them through the verifier
```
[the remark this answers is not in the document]
alfiedotwtf: What I meant here is that we want transaction Ids to be unique within the final accumulator. To do that, we need to de-duplicate all the transaction Ids. What we don't want, is to put the supplied transaction hex into the mempool.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -658,14 +658,15 @@ impl CommandHandler {
...
+                let transactions = transactions.into_iter().map(HexEncoded::take).collect();
                 let block = controller_opt
                     .as_mut()
                     .ok_or(WalletCliError::NoWallet)?
                     .generate_block(
                         selected_account.ok_or(WalletCliError::NoSelectedAccount)?,
-                        transactions_opt,
+                        transactions,
+                        vec![],
+                        false,
```
TheQuantumPhysicist, line 669: Unless I'm misunderstanding something, this seems to be a big, big mistake... we're basically blocking the main staking mechanism from using the mempool.

alfiedotwtf: Ah, good pick up. Thanks you're right.

test/functional/blockprod_generate_blocks_all_sources.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+"""Mempool reorg test
+
+Check that:
+* Transactions are collected into blocks when a new block is issued.
+* Transactions are correctly put back into mempool when the block is reorged out.
+"""
```
TheQuantumPhysicist, line 7: Please update the text

mempool/src/pool/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Well, my complaint is about the word... not sure that's English XD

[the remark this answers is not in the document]
alfiedotwtf: lol ah. Ok, I'll go for a better one :)

test/functional/blockprod_generate_blocks_all_sources.py
```diff
@@ -0,0 +1,79 @@
...
+            [1000 for _ in range(12)],
+        )
+
+        block = node.blockprod_generate_block(block_input_data, [generate_utxos], [], False)
+        node.chainstate_submit_block(block)
+        utxos = [(utxo_id, i) for i in range(12)]
+
+        for include_transaction in ([True, False]):
+            for include_transaction_id in ([True, False]):
+                for include_mempool in ([True, False]):
```
TheQuantumPhysicist, line 44: I like covering all cases, but I see below only the preparation, not testing the outcome. I would expect the test to first prepare the case (as you're doing below, maybe more if necessary), but also then to test the final result and ensure it's compatible with the expectation. For example, if we're not supposed to take transactions from the mempool, there should be a transaction in the mempool and we should prove that the code didn't pull that transaction from the mempool. Then when the mempool flag is enabled, it would pull that transaction from the mempool and we should see it as part of the final block that was produced.

So before calling `node.chainstate_submit_block(block)`, we should check the contents of the block and ensure that the combination of `include_transaction`, `include_transaction_id` and `include_mempool` was obeyed. Even if that means creating random transactions on every iteration. 

Note that the only issue with this (perfect) plan is that there's no functionality now to evict the mempool to recreate a clean situation on every iteration. A trick around that is to restart the node; as it'll lose all transactions since there's no network. The functionality to restart the node is available in functional tests, but I'm not sure whether it's complete (it's available from bitcoin, but not sure how adapted it's to our code). Worse-case scenario, write a TODO on it + create an issue until the eviction from the mempool is available.

alfiedotwtf: Yep, makes sense. I'll see if I can find the restart and get it working here.

blockprod/src/detail/mod.rs
```diff
@@ -424,36 +446,36 @@ impl BlockProduction {
...
+                    )
+                    .await?;
+
+                match accumulator {
+                    Some(acc) => acc.transactions().clone(),
+                    None => {
+                        // If the mempool rejects the accumulator (due
+                        // to tip mismatch, or otherwise), we should
+                        // start over and try again
+                        log::info!("Mempool rejected the transaction accumulator. Trying again.");
```
TheQuantumPhysicist, line 472: I don't like this "trying again" compared to what was there before. It's not clear what the implication is. What happens is that the whole block production attempt is restarted, but this isn't clear here.

test/functional/blockprod_generate_blocks_all_sources.py
alfiedotwtf: `stop_node()` and `start_node()` did the trick to flush the mempool.

blockprod/src/detail/mod.rs
alfiedotwtf: This one is tricky because we're now combining both cases. I've made a change to throw away mempool transactions if the mempool comes back None. This way, we're kind of back to the original behavior of only using the supplied transactions.

mempool/src/pool/mod.rs
```diff
@@ -968,9 +970,36 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
...
 
+        let mut unique_txids = BTreeSet::new();
+        let mut ordered_txids = vec![];
+
+        for transaction in tx_accumulator.transactions().iter() {
+            // Use transactions already in the Accumulator to sort for
+            // uniqueness only i.e don't send them through the
+            // verifier
+            unique_txids.insert(transaction.transaction().get_id());
+        }
```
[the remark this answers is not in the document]
iljakuklic: Also I don't think the case when a transaction given by hex spends the same utxo as one in mempool is handled properly.

It would be nice to start with a test to cover this:
* Add a transaction that spends an outpoint to mempool
* Construct a different transaction that spends the same outpoint, prepopulate the accumulator with it
* Call `collect_txs` with the accumulator and `fill_from_mempool` set `true`

mempool/src/pool/mod.rs
```diff
@@ -968,9 +970,36 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
...
+
+        if fill_from_mempool {
+            let fill_txids = self.store.txs_by_ancestor_score.iter().map(|(_, id)| id).rev();
+
+            for transaction_id in fill_txids {
+                if unique_txids.insert(*transaction_id) {
+                    ordered_txids.push(*transaction_id)
+                }
+            }
```
iljakuklic, line 996: This is not good. The original algorithm carefully used iterators specifically to avoid iterating over all transactions in the mempool, of which there can be many. That applies especially to sophisticated node operators that may want to increase the mempool size limit so they lose fewer transactions, e.g. stakers. However here, the whole score map is iterated over first to fill a vec but only a tiny fraction of it is used if mempool is anywhere close to full.

I would say in general, try to avoid filling in temporary data structures if using an iterator directly will do just fine.

iljakuklic (commented):
More feedbackz here:

mempool/src/pool/mod.rs
alfiedotwtf: Ah, the `fuse()` below. Nice.

mempool/src/pool/mod.rs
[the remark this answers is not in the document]
alfiedotwtf: Yep, I found that this scenario failed, but like we talked about yesterday, if the caller is being specific by supplying transactions, thten the onus is on them to get dependencies correct.

mempool/src/pool/mod.rs
iljakuklic: I don't think this has been addressed (as of 1351109e697df9f). More things are implemented in terms of iterators. However ultimately, it still has to iterate the whole collection in order to extend the `ordered_txids` vector. Ideally this is avoided so the loop below takes elements from the iterator directly. That way, we will not visit all tx ids in mempool when an error or full accumulator interrupts the iteration.

TheQuantumPhysicist:
Done through gitlab.

alfiedotwtf:
Reopening as this wasn't finished in Gitlab. I'll add the comments from Gitlab, and add this back on my TODO list in time.

alfiedotwtf:
@alfiedotwtf :
```
During yesterday's meeting, it was decided that to prevent the issue of double spend and other potential issues, the supplied transaction hex should be sent through a transaction verifier before calling collect_tx()...
But before I do this change, I want to confirm something - after going through the transaction verifier code, I'm not sure which TransactionSourceForConnect the supplied transaction hex should fall under for the connect_transaction() call - There's currently TransactionSourceForConnect::Chain and TransactionSourceForConnect::Mempool, neither of which feels right to sneak these transaction hex into. So in this case, should I add a TransactionSourceForConnect::Provided arm?
```

@iljakuklic :
```
I would say `Mempool` is more appropriate here out of the two. Maybe it could be renamed to something more descriptive that covers both cases. It basically says this is a candidate for inclusion in the next block (as opposed a transaction already included in a block).
```

PR #1163, opened 2023-08-14
Add option to save seed phrase in the wallet DB

OBorce:
Now you can optionally choose to save the seed phrase when creating the wallet for later retrieval. 
The seed phrase is encrypted if a password is set, same as the master keys.

### DIFF
--- a/node-gui/src/backend/backend_impl.rs
+++ b/node-gui/src/backend/backend_impl.rs
@@ -127,7 +127,7 @@ impl Backend {
             file_path.clone(),
             mnemonic,
             None,
-            false,
+            true,
         )
         .map_err(|e| BackendError::WalletError(e.to_string()))?;
 
--- a/node-gui/src/backend/backend_impl.rs
+++ b/node-gui/src/backend/backend_impl.rs
@@ -121,11 +121,13 @@ impl Backend {
                 .map_err(|err| BackendError::WalletError(err.to_string()))?;
         }
 
+        // TODO: add support for passphrase and saving of seed phrase
         let wallet = GuiController::create_wallet(
             Arc::clone(&self.chain_config),
             file_path.clone(),
             mnemonic,
             None,
+            false,
         )
         .map_err(|e| BackendError::WalletError(e.to_string()))?;
 
--- a/wallet/src/key_chain/master_key_chain/mod.rs
+++ b/wallet/src/key_chain/master_key_chain/mod.rs
@@ -69,7 +69,7 @@ impl MasterKeyChain {
         )
     }
 
-    pub fn new_from_root_key<B: storage::Backend>(
+    fn new_from_root_key<B: storage::Backend>(
         chain_config: Arc<ChainConfig>,
         db_tx: &mut StoreTxRwUnlocked<B>,
         root_key: ExtendedPrivateKey,
--- a/wallet/src/key_chain/master_key_chain/mod.rs
+++ b/wallet/src/key_chain/master_key_chain/mod.rs
@@ -25,6 +25,7 @@ use wallet_storage::{
     StoreTxRwUnlocked, WalletStorageReadLocked, WalletStorageReadUnlocked,
     WalletStorageWriteUnlocked,
 };
+use wallet_types::keys::SeedPhrase;
 
 use super::DEFAULT_VRF_KEY_KIND;
 
@@ -37,14 +38,14 @@ impl MasterKeyChain {
     pub fn mnemonic_to_root_key(
         mnemonic_str: &str,
         passphrase: Option<&str>,
-    ) -> KeyChainResult<(ExtendedPrivateKey, ExtendedVRFPrivateKey)> {
+    ) -> KeyChainResult<(ExtendedPrivateKey, ExtendedVRFPrivateKey, SeedPhrase)> {
         let mnemonic = zeroize::Zeroizing::new(
             bip39::Mnemonic::parse(mnemonic_str).map_err(KeyChainError::Bip39)?,
         );
         let seed = zeroize::Zeroizing::new(mnemonic.to_seed(passphrase.unwrap_or("")));
         let root_key = ExtendedPrivateKey::new_master(seed.as_ref(), DEFAULT_KEY_KIND)?;
         let root_vrf_key = ExtendedVRFPrivateKey::new_master(seed.as_ref(), DEFAULT_VRF_KEY_KIND)?;
-        Ok((root_key, root_vrf_key))
+        Ok((root_key, root_vrf_key, SeedPhrase { mnemonic }))
     }
 
     pub fn new_from_mnemonic<B: storage::Backend>(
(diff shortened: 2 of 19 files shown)


wallet/src/key_chain/master_key_chain/mod.rs
```diff
@@ -37,34 +38,43 @@ impl MasterKeyChain {
...
+        Self::new_from_root_key(
+            chain_config,
+            db_tx,
+            root_key,
+            root_vrf_key,
+            save_seed_phrase.then_some(seed_phrase),
+        )
     }
 
     pub fn new_from_root_key<B: storage::Backend>(
```
TheQuantumPhysicist, line 72: Does this method have to be public?

wallet/src/key_chain/master_key_chain/mod.rs
```diff
@@ -37,34 +38,43 @@ impl MasterKeyChain {
...
         chain_config: Arc<ChainConfig>,
         db_tx: &mut StoreTxRwUnlocked<B>,
         root_key: ExtendedPrivateKey,
```
TheQuantumPhysicist, line 75: Isn't the root key derived from the seed phrase?

wallet/types/src/keys.rs
```diff
@@ -130,6 +132,60 @@ pub struct RootKeys {
...
+
+    fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
+        f(self.encode().as_slice())
+    }
+
+    fn encoded_size(&self) -> usize {
+        self.encode().len()
+    }
+
+    fn encode_to<T: serialization::Output + ?Sized>(&self, dest: &mut T) {
```
TheQuantumPhysicist, line 165: You don't really need to manually implement all these functions. Just implementing `encode_to` is more than enough. Most of these call each other. See the Encode trait and how it already has default impls for these functions.

wallet/types/src/keys.rs
```diff
@@ -130,6 +132,60 @@ pub struct RootKeys {
...
+        // Preallocate enough space for the longest possible word list
+        33
+    }
+
+    fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
+        f(self.encode().as_slice())
+    }
+
+    fn encoded_size(&self) -> usize {
+        self.encode().len()
```
TheQuantumPhysicist, line 162: If it's possible to provided a provably correct value as a size-hint, you can use it as encoded size. This isn't always true for all types, like `Vec<u8>`, but is true for simple types like `u32`. So, if you know for sure that `self` has a constant size, you can just throw in that value with some debug assert.

wallet/types/src/keys.rs
```diff
@@ -130,6 +132,60 @@ pub struct RootKeys {
...
+    }
+
+    fn encode_to<T: serialization::Output + ?Sized>(&self, dest: &mut T) {
+        let entropy = self.mnemonic.to_entropy();
```
TheQuantumPhysicist, line 166: Are you sure using `to_entropy` is the correct way to encode thise? Why not use the serialized version of the struct. Looking briefly in it made it seem that 64-bytes do contain everything.

My concern is that the entropy you're serializing is not necessarily backward/forward compatible with changes in the library and is not part of the standard. We should serialize things that are specifically part of the standard to guarantee that things will never break.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -1104,6 +1116,22 @@ impl CommandHandler {
...
+                let phrase = controller_opt
+                    .as_mut()
+                    .ok_or(WalletCliError::NoWallet)?
+                    .seed_phrase()
+                    .map_err(WalletCliError::Controller)?;
+
+                let msg = if let Some(phrase) = phrase {
+                    format!("The saved seed phrase is \"{}\"", phrase)
+                } else {
+                    "No saved seed phrase for this wallet.".into()
```
TheQuantumPhysicist, line 1129: No saved seed phrase for this wallet. This was your choice when you created the wallet as a security option. Make sure not to lose this wallet file.

wallet/wallet-cli-lib/src/commands/helper_types.rs
```diff
@@ -142,3 +142,9 @@ impl CliWithLocked {
         }
     }
 }
+
+#[derive(Debug, Clone, Copy, ValueEnum)]
+pub enum CliSaveSeedPhrase {
+    Save,
+    DoNotSave,
```
TheQuantumPhysicist, line 149: NoSave

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -47,6 +48,9 @@ pub enum WalletCommand {
         /// File path
         wallet_path: PathBuf,
 
+        /// If true save the seed phrase to the DB for later retrieval
```
TheQuantumPhysicist, line 51: If 'save', the seed-phrase will be saved in the wallet file. If 'do-no-save', the seed-phrase will only be printed on the screen. Not saving the seed-phrase can be seen as a security measure to ensure sufficient secrecy in case that seed-phrase is reused elsewhere if this wallet is compromised.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -256,6 +260,9 @@ pub enum WalletCommand {
         pool_id: String,
     },
 
+    /// Show the seed phrase for the loaded wallet if it has been saved
+    ShowSeedPhrase,
```
TheQuantumPhysicist, line 264: Add an option also to delete the seed-phrase. Let's call it "PurgeSeedPhrase".

node-gui/src/backend/backend_impl.rs
```diff
@@ -121,11 +121,13 @@ impl Backend {
                 .map_err(|err| BackendError::WalletError(err.to_string()))?;
         }
 
+        // TODO: add support for passphrase and saving of seed phrase
         let wallet = GuiController::create_wallet(
             Arc::clone(&self.chain_config),
             file_path.clone(),
             mnemonic,
             None,
+            false,
```
TheQuantumPhysicist, line 130: Given the delete functionality, I would make this true by default.

TheQuantumPhysicist:
This is looking good so far. I'll review again after others have reviewed and the issues in the comments are addressed.

wallet/src/key_chain/master_key_chain/mod.rs
OBorce: made private

wallet/src/key_chain/master_key_chain/mod.rs
OBorce: yes

wallet/src/wallet/tests.rs
```diff
@@ -224,13 +225,53 @@ fn wallet_creation_in_memory() {
...
+
+    let err = wallet.seed_phrase().unwrap_err();
+    assert_eq!(
+        err,
+        WalletError::DatabaseError(wallet_storage::Error::WalletLocked)
+    );
+
+    wallet.unlock_wallet(&password).unwrap();
+    let seed_phrase = wallet.seed_phrase().unwrap().unwrap();
+    assert_eq!(seed_phrase.mnemonic.to_string(), MNEMONIC);
```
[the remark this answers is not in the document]
OBorce: only check for trying to read it while encrypted which returns the error wallet locked, but I am not sure there is an easy way to read the encrypted value from the DB

wallet/types/src/keys.rs
OBorce: removed

wallet/types/src/keys.rs
OBorce: I am now encoding the seed phrase words.

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: changed

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: changed

node-gui/src/backend/backend_impl.rs
OBorce: changed

TheQuantumPhysicist:
Done through gitlab.

Issue #1114, opened 2023-07-31
Currently in the GUI, only one staking pool will stake


TheQuantumPhysicist:
If multiple pools are found in the GUI in one account, one of them will only be staking.

This is a minor limitation due to time constraints. We have to fix it at some point.

pavel-kokolemin:
It's not limited to GUI, same happens with the CLI wallet.

pavel-kokolemin:
It was already fixed some time ago.

PR #1164, opened 2023-08-14
Minor staking fixes

pavel-kokolemin:
Resolve https://github.com/mintlayer/mintlayer-core/issues/1114


wallet/wallet-cli-lib/tests/cli_test_framework.rs
```diff
@@ -320,27 +297,60 @@ impl CliTestFramework {
             vi_mode: false,
         };
 
-        let console = MockConsole::new(commands);
+        let (output_tx, output_rx) = crossbeam_channel::unbounded::<String>();
+        let (input_tx, input_rx) = crossbeam_channel::unbounded::<String>();
```
TheQuantumPhysicist, line 301: Please be aware that crossbeam channels are not similar to boost signal in C++ (where all receivers get the message), but are more like a demultiplexer. They deliver messages only to one destination, not all. Are you sure this is what we need?

pavel-kokolemin: That's fine. Anyway, we do not really need the `crossbeam-channels` here, so I replaced it with `std::sync::mpsc` (after a little refactoring).

wallet/src/wallet/mod.rs
```diff
@@ -613,12 +617,16 @@ impl<B: storage::Backend> Wallet<B> {
...
+        &self,
         account_index: U31,
+        pool_id: PoolId,
     ) -> WalletResult<PoSGenerateBlockInputData> {
         let db_tx = self.db.transaction_ro_unlocked()?;
-        self.get_account(account_index)?
-            .get_pos_gen_block_data(&db_tx, self.latest_median_time)
+        self.get_account(account_index)?.get_pos_gen_block_data(
+            &db_tx,
+            self.latest_median_time,
```
TheQuantumPhysicist, line 627: If multiple pools are staking within the same wallet, dosen't that mean that each can have a different latest median time?

wallet/wallet-controller/src/lib.rs
```diff
@@ -651,38 +663,55 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static, W: WalletEvents> Controll
...
-            .map_err(ControllerError::NodeCallError)?;
-        Ok(block)
+            .map_err(ControllerError::NodeCallError)
     }
 
     pub async fn generate_blocks(
         &mut self,
         account_index: U31,
+        pool_id: Option<PoolId>,
         count: u32,
```
TheQuantumPhysicist, line 702: Is this like the number of attempts we try to generate a block? Or number of blocks? Or is it the number of blocks assuming all attempts succeeded? 

It's very unclear what this should be called and calling it "count" makes it even worse. Please add documentation to this function explaining what this parameter does. At least, if it's the number of blocks to be generated, let's call it `block_count`.

wallet/wallet-cli-lib/tests/cli_test_framework.rs
TheQuantumPhysicist: That's much better. Thank you. We don't need crossbeam at all. Using them can easily lead to someone calling Clone on rx, thinking that it'll act like boost signals. Using mpsc makes it much clearer. Cheers!

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -657,14 +663,21 @@ impl CommandHandler {
                 }
             }
 
-            WalletCommand::GenerateBlock { transactions } => {
+            WalletCommand::GenerateBlock {
+                pool_id,
+                transactions,
```
TheQuantumPhysicist, line 668: Oh, my... this brings even another level of complexity to the equation. Maybe @alfiedotwtf and @iljakuklic can chime in after the long discussions that were had about adding transactions manually with generate_block in Alfie's PR. What happens now when multiple pools take the same transactions and try to put them in blocks? Will it always work without issues and without double-spends?

TheQuantumPhysicist:
So, I did the first round. Please allow others to review this and let's see what the mempool/block production dudes think about the custom transactions. I'll review this again later.

wallet/src/wallet/mod.rs
pavel-kokolemin: The median time is always stored from the node (without looking at the wallet state), so it's theoretically possible. But the staking doesn't start until the wallet is fully synchronized, so this shouldn't be a problem.

wallet/wallet-controller/src/lib.rs
pavel-kokolemin: It's the number of blocks. I renamed it to `block_count` and added documentation.

wallet/src/account/mod.rs
```diff
@@ -546,16 +546,29 @@ impl Account {
...
+                pool_id
+                    == match kernel_input_utxo {
+                        TxOutput::CreateStakePool(pool_id, _) => *pool_id,
+                        TxOutput::ProduceBlockFromStake(_, pool_id) => *pool_id,
+                        TxOutput::Transfer(_, _)
+                        | TxOutput::LockThenTransfer(_, _, _)
+                        | TxOutput::Burn(_)
+                        | TxOutput::CreateDelegationId(_, _)
+                        | TxOutput::DelegateStaking(_, _) => panic!("Unexpected UTXO"),
+                    }
```
iljakuklic, line 569: Minor style suggestion: the `match` in the comparison expression here looks a bit strange. How about introducing a variable for it?

wallet/wallet-cli-lib/src/commands/mod.rs
pavel-kokolemin: I don't see a problem here because only one pool can succeed after the `generate_block` call.

wallet/wallet-cli-lib/tests/cli_test_framework.rs
```diff
@@ -16,11 +16,14 @@
...
 use crypto::{key::PublicKey, random::Rng, vrf::VRFPublicKey};
 use hex::FromHex;
+use tokio::task::JoinHandle;
 
 use std::{
-    collections::VecDeque,
     net::SocketAddr,
-    sync::{Arc, Mutex},
+    sync::{
+        mpsc::{Receiver, Sender},
```
iljakuklic, line 24: Minor: Maybe import just `mpsc` and refer to the channel endpoint types as `mpsc::Sender` and `mpsc::Receiver` in the code. It's a bit clearer since the other kinds of channels also have a `Sender` and a `Receiver`.

pavel-kokolemin:
> So, I did the first round. Please allow others to review this and let's see what the mempool/block production dudes think about the custom transactions. I'll review this again later.

I have made some changes to the `generate_block` wallet controller function. It will try all pools (in unspecified order, currently sorted by `pool_id`) and return the last pool error from the `generate_block` RPC call (or `ControllerError::NoStakingPool` if there are no staking pools).

wallet/wallet-cli-lib/tests/cli_test_framework.rs
```diff
@@ -320,27 +305,59 @@ impl CliTestFramework {
...
+
+        let wallet_task = tokio::spawn(async move {
+            tokio::time::timeout(
+                Duration::from_secs(120),
+                wallet_cli_lib::run(input, output, wallet_options, Some(chain_config)),
+            )
+            .await
+            .unwrap()
+            .unwrap();
+        });
```
iljakuklic, line 323: Just curious: Why did you decide to move this into a separate task? Is it because of the timeout?

wallet/wallet-controller/src/lib.rs
```diff
@@ -813,7 +855,7 @@ impl<T: NodeInterface + Clone + Send + Sync + 'static, W: WalletEvents> Controll
                         tokio::time::sleep(ERROR_DELAY).await;
                     }
 
-                    continue;
+                    continue 'outer;
```
iljakuklic, line 858: Congratulations, I think this is the first usage of this construct in the code base.

iljakuklic (approved):
LGTM

wallet/wallet-cli-lib/tests/cli_test_framework.rs
pavel-kokolemin: The tests now work differently, `MockConsoleInput` now sends input and `MockConsoleOutput` is expected output. For this to work, the CLI wallet task should be running in the background.

OBorce (approved):
looks good

TheQuantumPhysicist (approved):

PR #1167, opened 2023-09-13
Fix/wallet functional tests

OBorce:
Sometimes the initial read only reads the first couple of lines of the output, leaving other available lines in the buffer.


OBorce:
included in https://github.com/mintlayer/mintlayer-core/pull/1168

PR #1186, opened 2023-09-14
Use bounded channels for peer events

### LINKED ISSUE #664: Unbounded channels and DoS attacks
We use unbounded channels in p2p. This makes the node more vulnerable to DoS attacks when the used memory grows without bounds (if peers overload the node with requests). It would be good to do something about this.

Maybe we can can apply back-pressure on the connected peers, when we read from the network socket?
Check how many pending requests there are, and stop reading from the socket until the number is below some threshold.
Something like this:

```
response = self.cmd_rx.recv() {
    // Request was processed, send response to the peer...
    request_count -= 1;
},
// Read from the socket if only there is some free capacity
request = self.socket.recv(), if request_count < 10 => {
    // New request received, send it to backend...
    request_count += 1;
}
```

ImplOfAnImpl:
This fixes #664 (note that as we've discussed earlier, the solution proposed in the issue itself is redundant; using an unbounded channel for peer events should be enough).

P.S. I've based this PR on top of #1183 to simplify merging.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -165,6 +165,7 @@ where
     ) -> Self {
         // TODO: both "expect"s in this function may fire when exiting the node-gui app;
         // get rid of them and return a proper Result.
+        // See https://github.com/mintlayer/mintlayer-core/issues/1221
         let best_block = storage
             .get_best_block_for_utxos()
             .expect("Database error while reading utxos best block");
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -163,6 +163,8 @@ where
         accounting: A,
         verifier_config: TransactionVerifierConfig,
     ) -> Self {
+        // TODO: both "expect"s in this function may fire when exiting the node-gui app;
+        // get rid of them and return a proper Result.
         let best_block = storage
             .get_best_block_for_utxos()
             .expect("Database error while reading utxos best block");
--- a/p2p/src/net/default_backend/backend.rs
+++ b/p2p/src/net/default_backend/backend.rs
@@ -57,10 +57,15 @@ use super::{
     types::{HandshakeNonce, Message, P2pTimestamp},
 };
 
-/// Buffer size of the channel to the SyncManager peer task.
-/// How many unprocessed messages can be sent before the peer's event loop is blocked.
-// TODO: Decide what the optimal value is (for example, by comparing the initial block download time)
-const SYNC_CHAN_BUF_SIZE: usize = 20;
+/// Buffer sizes for the channels used by Peer to send peer messages to other parts of p2p.
+///
+/// If the number of unprocessed messages exceeds this limit, the peer's event loop will be
+/// blocked; this is needed to prevent DoS attacks where a peer would overload the node with
+/// requests, which may lead to memory exhaustion.
+/// Note: the value was chosen pretty much arbitrarily, but judging by the initial block download
+/// time there is no difference between 20 and 10000, so the former seems reasonable enough.
+const SYNC_MSG_CHAN_BUF_SIZE: usize = 20;
+const PEER_EVENT_CHAN_BUF_SIZE: usize = 20;
 
 /// Active peer data
 struct PeerContext {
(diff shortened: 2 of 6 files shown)


p2p/src/net/default_backend/backend.rs
```diff
@@ -57,10 +57,15 @@ use super::{
...
-// TODO: Decide what the optimal value is (for example, by comparing the initial block download time)
-const SYNC_CHAN_BUF_SIZE: usize = 20;
+/// Buffer sizes for the channels used by Peer to send peer messages to other parts of p2p.
+///
+/// If the number of unprocessed messages exceeds this limit, the peer's event loop will be
+/// blocked; this is needed to prevent DoS attacks where a peer would overload the node with
+/// requests, which may lead to memory exhaustion.
+/// Note: the value was chosen pretty much arbitrarily, but judging by the initial block download
+/// time there is no difference between 20 and 10000, so the former seems reasonable enough.
+const SYNC_MSG_CHAN_BUF_SIZE: usize = 20;
```
TheQuantumPhysicist, line 67: I really think it should much, much bigger. I mean, we went from infinity to 20? I wouldn't mind going for 1000 or even more.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -163,6 +163,8 @@ where
         accounting: A,
         verifier_config: TransactionVerifierConfig,
     ) -> Self {
+        // TODO: both "expect"s in this function may fire when exiting the node-gui app;
```
TheQuantumPhysicist, line 166: Please create an issue for this.

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
Please fix the minor issues I mentioned then we'll go ahead and merge.

p2p/src/net/default_backend/backend.rs
```diff
@@ -125,10 +130,10 @@ pub struct Backend<T: TransportSocket> {
 
     /// Channel sender for sending events from Peers to Backend; this will be passed to each
     /// Peer upon its creation.
-    peer_event_tx: mpsc::UnboundedSender<(PeerId, PeerEvent)>,
+    peer_event_tx: mpsc::Sender<(PeerId, PeerEvent)>,
```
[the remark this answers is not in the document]
ImplOfAnImpl: Fixed. Now each `Peer` has a dedicated channel for sending `PeerEvent`s. The channels' receivers are converted to tokio's `ReceiverStream` at the `Backend` and then aggregated into `StreamMap`, so that they can be `select`ed at once.

p2p/src/net/default_backend/backend.rs
ImplOfAnImpl: The sync messages can indeed vary wildly, from a single `Id` to a whole `Block`, which can be up to 1Mb in size. We've discussed this with Sam on Friday and decided that 100 should be a reasonable limit for the sync messages channel.
Peer manager messages, on the other hand, are relatively small and a bigger limit can be chosen, so we've decided on 1000. (I personally think that it's not necessary, but it won't hurt because those messages are small).

TheQuantumPhysicist (approved):

muursh (approved):

iljakuklic (approved):

iljakuklic:
Just noticed that `StreamMap` stores the channels in a `Vec` and polls them in order defined by the order of elements in the vec. This may give some peers advantage over others. For example the first connection will be in the first position and, as far as I can see, will stay there until disconnected.

Do we want to address this? Is it possible to periodically shuffle the elements or something like that?

ImplOfAnImpl:
> Just noticed that `StreamMap` stores the channels in a `Vec` and polls them in order defined by the order of elements in the vec. This may give some peers advantage over others. For example the first connection will be in the first position and, as far as I can see, will stay there until disconnected.
> 
> Do we want to address this? Is it possible to periodically shuffle the elements or something like that?

As far as I can see, `StreamMap::poll_next_entry` [chooses starting index randomly](https://docs.rs/tokio-stream/latest/src/tokio_stream/stream_map.rs.html#523), so this should not be a problem.

iljakuklic:
> As far as I can see, `StreamMap::poll_next_entry` [chooses starting index randomly](https://docs.rs/tokio-stream/latest/src/tokio_stream/stream_map.rs.html#523), so this should not be a problem.

Ok, good. Thanks for finding out.

PR #1187, opened 2023-09-14
Use tracing/tracing_subscriber instead of env_logger

ImplOfAnImpl:
This supersedes https://git.mintlayer.org/mintlayer/mintlayer-core/-/merge_requests/1149

The usage of `env_logger` was replaced with `tracing`/`tracing_subscriber`.

Logging in p2p tests is now initialized globally via the `ctor` crate, instead of doing so explicitly for certain tests only (this selectivity didn't work properly anyway).

Calls of the `#[tracing::instrument]` macro were put on every test on p2p, so that the test's name is included in the corresponding log messages.
For this to work properly, some futures had to be "instrumented" via the call to the `instrument` method.

The type of the log (json/text) and its coloring can now be selected via a separate env var `ML_LOG_STYLE`.

The unused "file path" parameter of `init_logging` was removed.


p2p/src/net/default_backend/transport/impls/channel.rs
```diff
@@ -211,6 +211,7 @@ mod tests {
         net::default_backend::{transport::BufferedTranscoder, types::Message},
     };
 
+    #[tracing::instrument(skip(seed))]
     #[rstest::rstest]
     #[trace]
     #[case(Seed::from_entropy())]
```
iljakuklic, line 217: Would the `instrument` annotation cause the seed to be printed if the `skip(...)` was removed? Maybe we could get rid of the annoying `#[trace]` attributes before every `#[case]`

p2p/src/net/default_backend/peer.rs
```diff
@@ -473,10 +481,13 @@ mod tests {
...
-            peer.handshake(P2pTimestamp::from_int_seconds(123456)).await.unwrap();
-            peer
-        });
+        let handle = tokio::spawn(
+            async move {
+                peer.handshake(P2pTimestamp::from_int_seconds(123456)).await.unwrap();
+                peer
+            }
+            .instrument(tracing::Span::current()),
+        );
```
iljakuklic, line 490: The pattern of spawning a tokio task and instrumenting it with the current span is repeated many times. Would it make sense to factor it out?

iljakuklic (commented):
Looks good. While this adds a fair bit of complexity which makes me a bit uncomfortable, I can see how the span feature and similar can be very useful. On balance, I think this is an improvement.

p2p/src/sync/tests/network_sync.rs
```diff
@@ -32,6 +32,7 @@ use crate::{
     testing_utils::for_each_protocol_version,
 };
 
+#[tracing::instrument(skip(seed))]
```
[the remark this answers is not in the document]
ImplOfAnImpl: 1) `tracing::instrument`:
This "wraps" the function in a tracing "span", the name of the function will be the span's name.
Each log line that is printed while a span is active will include the span's information - its name and fields.

2) `skip`:
By default, all arguments of the function are included in the span as fields; they all will be printed in each log line that belongs to the span.

    E.g. a random log line when `skip(seed)` is present (`block_announcement_disconnected_headers` is the name of the test and the span):
`2023-09-15T13:25:11.160035Z DEBUG block_announcement_disconnected_headers: p2p::sync: Register peer 1 to sync manager`
and the same line without it:
`2023-09-15T13:27:43.670527Z DEBUG block_announcement_disconnected_headers{seed=Seed(11837458260014005181)}: p2p::sync: Register peer 1 to sync manager`

    So, it makes sense to skip all agruments that don't help distinguish one span from another.

p2p/src/net/default_backend/transport/impls/channel.rs
ImplOfAnImpl: In that case the seed will be included in every log line.
So no, it can't replace `#[trace]`.

p2p/src/net/default_backend/peer.rs
ImplOfAnImpl: Yeah, I was also thinking about it. I'm not sure though, what's the best way to do it. I could just wrap `tokio::spawn`, factoring out the entire call to "instrument", or add a custom trait with a function like "forward_current_tracing_span", so that only the "tracing::Span::current()" part gets encapsulated.

iljakuklic: Yea I'm not sure. I would probably go either for wrapping spawn (something like `spawn_with_current_span`) or just leave it as it is, the repetition is not too bad after all. I don't think factoring out only the `Span` part would not be much different from repeating it since you still get the `spawn(future.forward_span())` patter repeated.

ImplOfAnImpl: So, I've added the function and named it `spawn_in_current_span`. There are two places though where I still have to instrument the futures explicitly, they are in `add_subsystem_with_config` where `JoinSet::spawn` is called instead of `tokio::spawn`. But it turned out that there is a more concise way of doing it - the extension trait that contains the `instrument` method also has `in_current_span`, which calls `self.instrument(Span::current())`.

subsystem/src/manager.rs
```diff
@@ -226,13 +227,13 @@ impl Manager {
...
                                     call(&mut *subsys).await
-                                });
+                                }.in_current_span());
                             },
                             Action::Ref(call) => {
                                 worker_tasks.spawn(async move {
                                     let subsys = subsys.read().await;
                                     call(&*subsys).await
-                                });
+                                }.in_current_span());
```
iljakuklic, line 236: I can see a future where we gradually move away from `task::spawn` in favour of various task pools and scoped tasks and similar. That means either invoking `in_current_span` manually like here or coming up with a wrapper to make it convenient (similar to the top-level `spawn_in_current_span`).

Does it make sense to address it now or should we leave it and deal with it as the need arises?

iljakuklic (approved):

logging/Cargo.toml
```diff
@@ -8,5 +8,8 @@ rust-version.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-env_logger.workspace = true
 log.workspace = true
+thiserror.workspace = true
+tokio = { workspace = true, default-features = false, features = ["rt", "sync"] }
```
TheQuantumPhysicist, line 13: Tokio with logging... what could go wrong...

Just for the record, I hate this... but I think I can't provide a case against it to stop it. I hope I'm wrong.

TheQuantumPhysicist (approved):

subsystem/src/manager.rs
ImplOfAnImpl: I'm not sure how to address this now, because it's not clear to me what those `spawn` replacements will look like.
Do you have any particular suggestions?

muursh (approved):

subsystem/src/manager.rs
iljakuklic: I'm fine doing it as the need arises and some patterns emerge

PR #1190, opened 2023-09-15
Replace API Server's mock data responses with database calls

alfiedotwtf:
Most of the current API server endpoints are now pulling from a database handle.

### DIFF
--- a/api-server/web-server/src/api/v1.rs
+++ b/api-server/web-server/src/api/v1.rs
@@ -13,18 +13,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use axum::{extract::Path, response::IntoResponse, routing::get, Json, Router};
+use api_server_common::storage::storage_api::{ApiServerStorage, ApiServerStorageRead};
+use axum::{
+    extract::{Path, State},
+    response::IntoResponse,
+    routing::get,
+    Json, Router,
+};
 use common::{
-    chain::{Block, Genesis, Transaction},
-    primitives::{BlockHeight, Id, H256},
+    chain::{Block, Transaction, TxOutput},
+    primitives::{BlockHeight, Id, Idable, H256},
 };
 use crypto::random::{make_true_rng, Rng};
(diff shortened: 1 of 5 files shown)


api-server/web-server/src/main.rs
```diff
@@ -37,8 +40,13 @@ async fn main() {
     let args = ApiServerWebServerConfig::parse();
     log::info!("Command line options: {args:?}");
 
+    // TODO: generalise network configuration
```
TheQuantumPhysicist, line 43: "generalize". We use American English in the code.

TheQuantumPhysicist (approved):
Actually this is looking good as is. I'm happy to merge it.

api-server/web-server/src/main.rs
alfiedotwtf: Muscle memory... I'm sure there's an Emacs macro to fix my spelling 😄

api-server/web-server/src/api/v1.rs
```diff
@@ -68,115 +81,242 @@ pub fn routes() -> Router<APIServerWebServerState> {
...
+            .into();
+
+        state
+            .db
+            .transaction_ro()
+            .map_err(|_| {
+                APIServerWebServerError::ServerError(
+                    APIServerWebServerServerError::InternalServerError,
+                )
+            })?
```
iljakuklic, line 102: Could you define a constant or something for the error? I don't like how a mere conversion of error type takes up 5 lines while the actual "substantive" code is a fraction of that.

iljakuklic (approved):
Looks good. A tiny style comment below:

api-server/web-server/src/api/v1.rs
iljakuklic: This is replicated many times over.

alfiedotwtf: Yep. There's currently not that many error types, but once it's fully fleshed out, I'll replace the errors with real constants.

PR #1195, opened 2023-09-17
Make API server database 100% async

TheQuantumPhysicist:
The only problem left that I don't know how to fix is how to make transaction drops async too... but I guess this is a problem in rust not support AsyncDrop yet... but otherwise, I think this is a viable, working code.

### DIFF
--- a/api-server/api-server-common/src/storage/impls/postgres/queries.rs
+++ b/api-server/api-server-common/src/storage/impls/postgres/queries.rs
@@ -26,7 +26,7 @@ use crate::storage::{
 };
 
 pub struct QueryFromConnection<'a, 'b> {
-    tx: &'a mut postgres::Transaction<'b>,
+    tx: &'a tokio_postgres::Transaction<'b>,
 }
 
 impl<'a, 'b> QueryFromConnection<'a, 'b> {
@@ -43,7 +43,7 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
 }
 
 impl<'a, 'b> QueryFromConnection<'a, 'b> {
-    pub fn new(tx: &'a mut postgres::Transaction<'b>) -> Self {
+    pub fn new(tx: &'a tokio_postgres::Transaction<'b>) -> Self {
         Self { tx }
     }
 
@@ -55,11 +55,12 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
             .unwrap_or_else(|e| panic!("Invalid block height: {e}"))
     }
 
-    pub fn is_initialized(&mut self) -> Result<bool, ApiServerStorageError> {
+    pub async fn is_initialized(&mut self) -> Result<bool, ApiServerStorageError> {
         let query_str = Self::get_table_exists_query("ml_misc_data");
         let row_count = self
             .tx
             .query_one(&query_str, &[])
+            .await
             .map_err(|e| ApiServerStorageError::LowLevelStorageError(e.to_string()))?;
 
         let row_count: i64 = row_count.get(0);
@@ -68,7 +69,7 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
             return Ok(false);
         }
 
-        let version = self.get_storage_version()?;
+        let version = self.get_storage_version().await?;
 
         let version = match version {
             Some(v) => v,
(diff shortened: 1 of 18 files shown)


api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -142,83 +149,91 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
     }
 
-    pub fn create_tables(&mut self) -> Result<(), ApiServerStorageError> {
+    pub async fn create_tables(&mut self) -> Result<(), ApiServerStorageError> {
         logging::log::info!("Creating database tables");
 
         self.just_execute(
             "CREATE TABLE ml_misc_data (
             name TEXT PRIMARY KEY,
             value BLOB NOT NULL
```
alfiedotwtf, line 173: I might change these `BLOB`s to `bytea` or `TEXT`

api-server/api-server-common/src/storage/impls/postgres/transactional/mod.rs
```diff
@@ -30,171 +30,182 @@ use crate::storage::storage_api::{
...
+            futures::executor::block_on(self.connection.batch_execute("ROLLBACK")).unwrap_or_else(
+                |e| {
+                    logging::log::error!(
+                        "CRITICAL ERROR: failed to rollback failed postgres RW transaction: {e}"
+                    )
+                },
+            );
         }
-        .try_build()?;
+    }
```
iljakuklic, line 133: I am not sure about all the implications of `block_on` in the destructor, seems like it could lead to problems if the future needs something else to resolve as part of its execution. Also, if I understand this correctly, the executor from the `futures` crate is different from the tokio one. Is it fine to mix and match them?

Not sure what to do about it. One option is to enforce that the user uses explicitly one of the `commit` or `rollback` methods to end the transaction. In the destructor, we just assert `self.finished`. However, it may be hard to chase up all the places where a transaction object is destroyed, especially considering all the `?`s and other forms of early return.

api-server/storage-test-suite/tests/in_memory.rs
```diff
@@ -13,20 +13,20 @@
...
+// }
 
 fn main() {
-    let storage_maker = || make_in_memory_storage(&create_unit_test_config());
-    let result = api_server_backend_test_suite::run(storage_maker);
+    // let storage_maker = || make_in_memory_storage(&create_unit_test_config());
+    // let result = api_server_backend_test_suite::run(storage_maker);
 
-    result.exit()
+    // result.exit()
```
iljakuklic, line 31: Do you plan to restore the test or should it just be removed?

iljakuklic (commented):
This looks good but not sure about the destructor stuff.

api-server/api-server-common/src/storage/impls/postgres/transactional/mod.rs
iljakuklic: Is there a test that exercises this code path?

api-server/storage-test-suite/tests/in_memory.rs
TheQuantumPhysicist: Yep. That's an oversight.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
TheQuantumPhysicist: Feel free to use whatever works for you.

api-server/api-server-common/src/storage/impls/postgres/transactional/mod.rs
TheQuantumPhysicist: There's zero tests for postgres yet. Alfie is working on the containerization.

I've been thinking about this block_on problem yesterday all night (and panicking when forgetting occurred to me, but it's ugly)... I have an idea on how to get rid of it here... I'll discuss it in the daily.

TheQuantumPhysicist: All my attempts to move the processing to the db as a parent failed so far. There's always the requirement to make something here or there 'static. The closest I came to succeeding is to make channels that send tasks to a tokio task, but then because the connection has a lifetime, it's mandating having 'static lifetime for the receiver.

scoped tokio didn't help because it returns a reference that cannot be moved out of scope.

Same for std scoped threads, because the scope cannot be moved.

No idea how I'll proceed. Maybe I'll have the patience to do it later. Feel free to take a poke at it.

alfiedotwtf (approved):

PR #1225, opened 2023-09-20
Fix delegation support in the wallet

OBorce:
- previously all delegation staking was added as own in the cache
- fix delegation rollback when a reorg happens
- fix delegation nonce rollback when abandoning inactive transactions with descendants

### DIFF
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -53,6 +53,7 @@ use crypto::key::hdkd::u31::U31;
 use crypto::key::PublicKey;
 use crypto::vrf::{VRFPrivateKey, VRFPublicKey};
 use itertools::Itertools;
+use std::cmp::Reverse;
 use std::collections::BTreeMap;
 use std::ops::Add;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


TheQuantumPhysicist (approved):
Can we please add some functional tests? Besides this, this is OK.

wallet/src/account/mod.rs
```diff
@@ -765,12 +771,19 @@ impl Account {
     /// Return true if this transaction output is can be spent by this account or if it is being
     /// watched.
     fn is_mine_or_watched(&self, txo: &TxOutput) -> bool {
-        Self::get_tx_output_destination(txo).map_or(false, |d| match d {
+        Self::get_tx_output_destination(txo)
+            .map_or(false, |d| self.is_mine_or_watched_destination(d))
+    }
+
+    /// Return true if this transaction output is can be spent by this account or if it is being
```
muursh, line 778:
```suggestion
    /// Return true if this transaction output can be spent by this account or if it is being
```

muursh (approved):

Issue #1219, opened 2023-09-18
Mempool chainstate synchronization issues


TheQuantumPhysicist:
(Created on Aug 31, 2023 by @iljakuklic)

Mempool and chainstate operate concurrently. When a transaction is processed, mempool employs the transaction verifier. The verifier in turn performs a series of chainstate queries. However, the state of the blockchain may change between the individual queries so it may happen that a transaction is validated with respect to an inconsistent blockchain state.

Currently, this is mitigated by checking the tip before and after the verification. If it does not match the verification is re-attempted up to a certain small number of times. This should almost always work since blocks are not supposed to come in very often by computer time standards. It is, however, still somewhat suboptimal and improvements can be made. Besides having an inconsistent view of the state, performing many inter-subsystem queries incurs non-trivial synchronisation overhead.

A possible improvement is to collect all the pieces of information required to verify the transaction in one query and store it in a data structure. The data structure can be then used as a storage backend for transaction verifier. That way, the number of subsystem calls is minimised and we get consistent state to query against. In addition to that, the `BlockingHandle` machinery to call subsystems from non-`async` contexts would likely be no longer needed.

TheQuantumPhysicist:
(Comment by @TheQuantumPhysicist)

I feel like there's a mixup in this description between two problems, one of them never has a solution and the other does have a solution.
Problem 1: Synchronization between the mempool and the chainstate. In the most general terms, there's no solution to this problem. This is the classical CAP theorem problem. The system we created (the subsystems manager) is consistent and partition tolerant. It can never be always available, which seems to be the requirement in this issue as I understand it. Now the solution proposed here (preparing data), unless I misunderstand something, doesn't make sense. Because it creates a trade-off between availability and consistency in CAP. Sure, you can make the operation more "available" by pre-caching information, but then the data we use will never be consistent with chainstate, because chainstate can still receive a new block in concurrently, and that new block will invalidate the current state of the mempool. So preparing data doesn't really help in solving the synchronization issue.
Problem 2: Optimizing the communication between the mempool and chainstate. Now that is completely correct and should be done, but this isn't gonna fix the synchronization issue. It's purely a performance fix that I totally agree with.

TheQuantumPhysicist:
(Comment by @iljakuklic)

[...]

PR #1227, opened 2023-09-20
Reduce log level of certain logging lines from "debug" to "trace" to reduce test logs bloat

ImplOfAnImpl:
Initially I was going to reduce the number of blocks generated by tests to fix this, but now I think it would be a bad idea. So, in the end I've just demoted the most spamming logging lines from "debug" to "trace".
(Just in case: the main offenders here are the lines in mempool, the other ones are just accomplices).

I also described the issue in detail here - https://github.com/mintlayer/mintlayer-core/issues/1219#issuecomment-1728176441

### DIFF
--- a/mempool/src/pool/mod.rs
+++ b/mempool/src/pool/mod.rs
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use mempool_types::tx_origin::LocalTxOrigin;
 use parking_lot::RwLock;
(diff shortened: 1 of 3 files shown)


iljakuklic (approved):
I have no problem with this, and no particularly strong opinion. Log printouts can be adjusted using the env variable if needed for debugging.

TheQuantumPhysicist (approved):

muursh (approved):

PR #1229, opened 2023-09-21
Use podman to run postgres tests

TheQuantumPhysicist:
I created my own facility to run podman since testcontainers-rs is ugly with its unjustified lifetime and lacks direct support for podman. 

At the time of creating this PR, I'm able to launch containers and stop them automatically, but I still need to configure the container properly to get the tests to work.


TheQuantumPhysicist:
This is ready for review. The only thing left is to add a feature that will force enabling the containerized test, which I'll do tomorrow because it's too late.

alfiedotwtf (approved):

PR #1230, opened 2023-09-22
Add CLI option to control feerate argument from memppol

OBorce:
- add CLI option in-top-x-mb argument to control how much to prioritize the
      transaction feerate so the transaction can get in the top x MB of the
    mempool
- add functional test


node-gui/src/backend/backend_impl.rs
```diff
@@ -376,7 +376,8 @@ impl Backend {
             .get_mut(&wallet_id)
             .ok_or(BackendError::UnknownWalletIndex(wallet_id))?
             .controller
-            .synced_controller(account_index)
+            // TODO: add option to select from GUI
+            .synced_controller(account_index, ControllerConfig { in_top_x_mb: 5 })
```
TheQuantumPhysicist, line 380: Please, no magic numbers in the middle of the code.

test-rpc-functions/src/rpc.rs
```diff
@@ -269,6 +280,59 @@ impl RpcTestFunctionsRpcServer for super::RpcTestFunctionsHandle {
...
+        mut input_tx_id: Id<Transaction>,
+        num_transactions: u32,
+        amount_to_spend: u64,
+        fee_per_tx: u64,
+    ) -> rpc::Result<Vec<HexEncoded<SignedTransaction>>> {
+        let coin_decimals = self
+            .call(|this| this.get_chain_config().map(|chain| chain.coin_decimals()))
+            .await
+            .expect("Subsystem call ok")
+            .expect("Subsystem call ok");
```
TheQuantumPhysicist, line 295: It's a very bad idea to have two expects, right after each other, and with the exact same message... basically if this error happens, it may be difficult to know where it really happened.

test-rpc-functions/src/rpc.rs
```diff
@@ -269,6 +280,59 @@ impl RpcTestFunctionsRpcServer for super::RpcTestFunctionsHandle {
...
+
+            transactions.push(HexEncoded::new(transaction));
+        }
+
+        Ok(transactions)
+    }
+}
+
+fn pow10(coin_decimals: u8) -> u128 {
+    let mut coin_decimal_factor = 1;
```
TheQuantumPhysicist, line 331: Why not just `10u128.pow(coin_decimals)`?

test/functional/wallet_high_fee.py
```diff
@@ -0,0 +1,151 @@
...
+
+            best_block_id = await wallet.get_best_block()
+            assert best_block_id == block_id
+
+            balance = await wallet.get_balance()
+            assert "Coins amount: 10" in balance
+
+            transactions = node.test_functions_generate_transactions(tx_id, 25, total - 300, 300)
+            for idx, encoded_tx in enumerate(transactions):
+                self.log.info(f"submiting tx {idx}")
```
TheQuantumPhysicist, line 139: submitting

TheQuantumPhysicist (approved):

node-gui/src/backend/backend_impl.rs
OBorce: added the constant at the top with a comment

test-rpc-functions/src/rpc.rs
OBorce: changed the message

test-rpc-functions/src/rpc.rs
OBorce: fixed

PR #1232, opened 2023-09-22
Update iced gui library 0.9 -> 0.10

OBorce:
Update the iced library because of a security issue related to its dependency ouroboros.

New:
- Show an info message when trying to mutate the wallet while still syncing.
- Use different channels for wallet update events to avoid flooding the channel while scanning many blocks when multiple wallets opened, and cause lagging in processing the user commands.


node-gui/src/main_window/main_widget/tabs/mod.rs
```diff
@@ -70,30 +70,35 @@ impl TabsWidget {
...
             .push(
+                2,
                 self.settings_tab.tab_label(),
                 self.settings_tab.view(node_state),
             );
 
-        for wallet in self.wallets.iter() {
-            tabs = tabs.push(wallet.tab_label(), wallet.view(node_state))
+        for (idx, wallet) in self.wallets.iter().enumerate() {
+            tabs = tabs.push(idx + 3, wallet.tab_label(), wallet.view(node_state))
```
iljakuklic, line 92: Could you read the number of tabs before the loop instead of hardcoding `+3`?

iljakuklic (approved):

TheQuantumPhysicist:
There are a few things broken in the wallet. When creating a new wallet, the wallet just keeps getting created indefinitely. Also the progress bar for wallet syncing is not showing up.

Please run every functionality in the old wallet and the new wallet side by side and ensure that everything behaves correctly. In fact, it's an open question how we test these features.

TheQuantumPhysicist:
I made an attempt to fix the issue Lukas pointed to. It's not perfect given that the number of wallets is dynamic, but perhaps this is good for what it's. Feel free to improve it.

node-gui/src/main_window/main_widget/tabs/mod.rs
TheQuantumPhysicist: Done.

TheQuantumPhysicist (approved):

PR #1234, opened 2023-09-26
Switch to Postgres storage

alfiedotwtf:
This change switches out the storage API for the Scanner and Web Server from InMemory to Postgres

### DIFF
--- a/api-server/scanner-daemon/src/config.rs
+++ b/api-server/scanner-daemon/src/config.rs
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use api_server_common::PostgresConfig;
 use clap::Parser;
 use common::chain::config::ChainType;
 use std::net::SocketAddr;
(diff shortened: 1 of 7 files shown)


api-server/scanner-daemon/src/config.rs
```diff
@@ -46,6 +46,30 @@ pub struct ApiServerScannerArgs {
...
+    #[clap(long)]
+    pub postgres_password: Option<String>,
+
+    /// Postgres database
+    #[clap(long)]
+    pub postgres_database: Option<String>,
+
+    /// Postgres max connections
+    #[clap(long, default_value = "10")]
+    pub postgres_max_connections: u32,
```
TheQuantumPhysicist, line 72: Please make an attempt to group postgres-related configuration to in its own struct. I think clap supports composable configuration, not sure though. 

Please note that we DO NOT want use commands for this. I know commands can do it, but it's the wrong tool. It would be sufficient to just be able to put these in a struct then throw the struct in here.

IF that works, then it would be great if we could have tree-like validation, where postgres settings only work if the database type is postgres. But that comes later, no need to spend too much time on it. Let's start with the struct. 

If that struct works, then please put it in api-server-common, and use it for both the scanner-daemon and the web-server.

alfiedotwtf: The new changes have a separate struct for Postgres, and are parsed via the `default()`. If we want to be optional vs other backing storage types as discussed in the meeting, I'm not sure if Clap supports the Enum like you mentioned so you may have to detect manually after `parse()` the top-level?

api-server/web-server/src/config.rs
```diff
@@ -27,6 +27,43 @@ pub struct ApiServerWebServerConfig {
...
     #[clap(long)]
     pub address: Option<ListenAddress>,
+
+    /// Postgres config values
+    #[clap(flatten)]
+    pub postgres_config: PostgresConfig,
+}
+
+#[derive(Parser, Debug)]
+pub struct PostgresConfig {
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Please don't forget to put them in the common crate too.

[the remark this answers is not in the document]
alfiedotwtf: Cool, this one is done.

iljakuklic (approved):

api-server/scanner-daemon/src/config.rs
OBorce: If you want to use an enum you can actually use a subcommand like: https://github.com/mintlayer/mintlayer-core/blob/e3c81965ceed2aa774793809c7d2e7ecccad5923/node-lib/src/options.rs#L44-L52

Issue #1221, opened 2023-09-18
`TransactionVerifier` shouldn't unwrap `Result`s.


ImplOfAnImpl:
Currently, `TransactionVerifier` is too optimistic about error handling; in particular, it assumes that DB errors cannot happen and `UtxosCache` creation will always succeed, calling `expect` on the corresponding `Result`s. This may lead to crashes during node shutdown; in particular, both `expect` calls in `TransactionVerifier::new_generic` may fire when exiting the `node-gui` app.

All calls to `Result::expect` must be removed and errors should be propagated properly.

P.S. This may require some refactoring in mempool, which currently assumes that `TransactionVerifier` never fails.
Also, mempool itself can sometimes be overly optimistic, assuming for example that chainstate is always alive; this should better be fixed too (e.g. look for the line `.expect("chainstate to live")` in `mempool/src/pool/mod.rs`)

TheQuantumPhysicist:
I've been looking into the TransactionVerifier, and I think this warrants a discussion more than action. The question is: What is the right thing to do when an irrecoverable error happens?

One side of the extreme: Let's define what irrecoverable error means, and while I admit this is a tongue-in-cheek point, I'm making this point to clarify the red line. Let's say we open the database, and the database returns an error. What are we supposed to do in this case? The bad part here is that this can easily be caused by a hardware issue. There's virtually no way to detect this.

The other extreme side: What if we expect on error that can happen for justified reasons?

Hence, the rule of thumb is, if it's a logical, invariant error or hardware error, then we expect, and crash. That's better than behaving randomly.

So, the better question for this issue is: What is it that isn't an invariant error that we should change? I'm open for that discussion. We can arrange a meeting for that.

ImplOfAnImpl:
> What is the right thing to do when an irrecoverable error happens?

IMO it depends on the location where it happens. I'd say that in places like `TransactionVerifier` we should propagate all errors to the caller (I mean the actual ones, produced by user actions or the environment, as opposed to invariant violations).
And then the application could:
1) Report the error to the user and exit gracefully.
2) Ignore the error and just log it with some "mild" severity. This is reasonable to do when the app is being shut down, so some subsystems may no longer be available.

> Hence, the rule of thumb is, if it's a logical, invariant error or hardware error, then we expect, and crash. That's better than behaving randomly.

[...]

PR #1235, opened 2023-09-26
Create a separate type for absolute time + some related fixes and refactoring

TheQuantumPhysicist:
Please check the code relevant to your work and see if it makes sense. 

There's a test broken in block production because it used the wrong time source and modified the global mocked time. That's gonna be fixed in this PR before merging it.

### DIFF
--- a/common/src/chain/block/timestamp.rs
+++ b/common/src/chain/block/timestamp.rs
@@ -19,6 +19,8 @@ use serde::{Deserialize, Serialize};
 
 use serialization::{Decode, Encode};
 
+use crate::primitives::time::Time;
+
 pub type BlockTimestampInternalType = u64;
 
 #[derive(
@@ -52,6 +54,14 @@ impl BlockTimestamp {
 
     pub const fn as_int_seconds(&self) -> BlockTimestampInternalType {
         self.timestamp
+    }
+
+    pub const fn from_time(time: Time) -> Self {
+        Self::from_duration_since_epoch(time.as_duration_since_epoch())
+    }
+
+    pub const fn into_time(self) -> Time {
+        Time::from_duration_since_epoch(self.as_duration_since_epoch())
     }
 
     pub fn add_int_seconds(&self, seconds: BlockTimestampInternalType) -> Option<BlockTimestamp> {
--- a/common/src/primitives/time.rs
+++ b/common/src/primitives/time.rs
@@ -29,10 +29,10 @@ pub fn duration_from_int(v: u64) -> Duration {
 static TIME_SOURCE: AtomicU64 = AtomicU64::new(0);
 
 /// Return mocked time if set, otherwise return `None`
-fn get_mocked_time() -> Option<Duration> {
+fn get_mocked_time() -> Option<Time> {
     let value = TIME_SOURCE.load(Ordering::SeqCst);
     if value != 0 {
-        Some(duration_from_int(value))
+        Some(Time::from_duration_since_epoch(duration_from_int(value)))
     } else {
         None
     }
@@ -52,7 +52,7 @@ pub fn set(now: Duration) -> Result<(), std::num::TryFromIntError> {
 /// Either gets the current time or panics
 pub fn get_time() -> Time {
     match get_mocked_time() {
-        Some(mocked_time) => Time::from_duration_since_epoch(mocked_time),
+        Some(mocked_time) => mocked_time,
         None => Time::from_duration_since_epoch(
             SystemTime::now()
                 .duration_since(SystemTime::UNIX_EPOCH)
@@ -170,7 +170,10 @@ mod tests {
 
         set(Duration::from_secs(1337)).unwrap();
         assert_eq!(get_time().as_duration_since_epoch().as_secs(), 1337);
-        assert_eq!(get_mocked_time(), Some(Duration::from_secs(1337)));
+        assert_eq!(
+            get_mocked_time(),
+            Some(Time::from_duration_since_epoch(Duration::from_secs(1337)))
+        );
 
         reset();
         assert_eq!(get_mocked_time(), None);
--- a/common/src/primitives/time.rs
+++ b/common/src/primitives/time.rs
@@ -72,6 +72,10 @@ impl Time {
         self.time
     }
 
+    pub const fn as_secs_since_epoch(&self) -> u64 {
+        self.time.as_secs()
+    }
+
     pub const fn from_duration_since_epoch(duration: Duration) -> Self {
         Self { time: duration }
     }
@@ -129,38 +133,23 @@ mod tests {
         logging::init_logging();
         set(Duration::from_secs(1337)).unwrap();
 
-        log::info!(
-            "p2p time: {}",
-            get_time().as_duration_since_epoch().as_secs()
-        );
+        log::info!("p2p time: {}", get_time().as_secs_since_epoch());
         std::thread::sleep(Duration::from_secs(1));
 
-        log::info!(
-            "p2p time: {}",
-            get_time().as_duration_since_epoch().as_secs()
-        );
-        assert_eq!(get_time().as_duration_since_epoch().as_secs(), 1337);
+        log::info!("p2p time: {}", get_time().as_secs_since_epoch());
+        assert_eq!(get_time().as_secs_since_epoch(), 1337);
         std::thread::sleep(Duration::from_secs(1));
 
-        log::info!(
-            "rpc time: {}",
-            get_time().as_duration_since_epoch().as_secs()
-        );
+        log::info!("rpc time: {}", get_time().as_secs_since_epoch());
         std::thread::sleep(Duration::from_millis(500));
 
-        assert_eq!(get_time().as_duration_since_epoch().as_secs(), 1337);
-        log::info!(
-            "rpc time: {}",
-            get_time().as_duration_since_epoch().as_secs()
-        );
+        assert_eq!(get_time().as_secs_since_epoch(), 1337);
+        log::info!("rpc time: {}", get_time().as_secs_since_epoch());
         std::thread::sleep(Duration::from_millis(500));
 
         reset();
-        assert_ne!(get_time().as_duration_since_epoch().as_secs(), 1337);
-        log::info!(
-            "rpc time: {}",
-            get_time().as_duration_since_epoch().as_secs()
-        );
+        assert_ne!(get_time().as_secs_since_epoch(), 1337);
+        log::info!("rpc time: {}", get_time().as_secs_since_epoch());
     }
 
     #[test]
@@ -169,7 +158,7 @@ mod tests {
         assert_eq!(get_mocked_time(), None);
 
         set(Duration::from_secs(1337)).unwrap();
-        assert_eq!(get_time().as_duration_since_epoch().as_secs(), 1337);
+        assert_eq!(get_time().as_secs_since_epoch(), 1337);
         assert_eq!(
             get_mocked_time(),
             Some(Time::from_duration_since_epoch(Duration::from_secs(1337)))
--- a/common/src/primitives/time.rs
+++ b/common/src/primitives/time.rs
@@ -50,12 +50,70 @@ pub fn set(now: Duration) -> Result<(), std::num::TryFromIntError> {
 }
 
 /// Either gets the current time or panics
-pub fn get_time() -> Duration {
+pub fn get_time() -> Time {
     match get_mocked_time() {
-        Some(mocked_time) => mocked_time,
-        None => SystemTime::now()
-            .duration_since(SystemTime::UNIX_EPOCH)
-            .expect("Time went backwards"),
+        Some(mocked_time) => Time::from_duration_since_epoch(mocked_time),
+        None => Time::from_duration_since_epoch(
+            SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .expect("Time went backwards"),
+        ),
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Time {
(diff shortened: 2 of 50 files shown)


mempool/src/pool/orphans/test.rs
```diff
@@ -119,7 +125,7 @@ fn insert_and_delete(#[case] seed: Seed) {
 fn capacity_reached(#[case] seed: Seed) {
     let mut rng = make_seedable_rng(seed);
     let mut orphans = TxOrphanPool::new();
-    let time = Time::from_secs(0);
+    let time = Time::from_duration_since_epoch(Duration::from_secs(0));
```
ImplOfAnImpl, line 128: `from_secs_since_epoch` would be nice to have for cases like this (especially if you implement `as_secs_since_epoch` as I suggested in another comment)

p2p/src/net/default_backend/types.rs
```diff
@@ -73,6 +73,10 @@ impl P2pTimestamp {
     pub fn as_duration_since_epoch(&self) -> Duration {
         Duration::from_secs(self.0)
     }
+
+    pub fn from_time(time: Time) -> Self {
+        Self::from_duration_since_epoch(time.as_duration_since_epoch())
+    }
```
ImplOfAnImpl, line 79: Same as for `BlockTimestamp`, should we have `impl From<Time> for P2pTimestamp` instead of/in addition to this?

p2p/src/peer_manager/peerdb/address_data/mod.rs
```diff
@@ -162,20 +163,27 @@ impl AddressData {
...
+                Some(r) => return r,
+                None => {
+                    // The random offset is too large, try again
+                    logging::log::debug!(
+                        "next_connect_time: Random_offset too large (now: {now:?}, offset: {random_offset:?}), trying again"
+                    );
+                }
+            }
+        }
     }
```
ImplOfAnImpl, line 181: This loop looks a bit unnecessary from the one hand and scary from the other.
1) The addition won't fail if `next_connect_delay` and `exponential_rand` behave as expected, because the returned numbers are small enough.
2) If, on the other hand, we assume that those functions, in particular `next_connect_delay`, may return something unexpected, then the loop may become infinite.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -93,6 +92,11 @@ impl<S: PeerDbStorage> PeerDb<S> {
             anchor_addresses,
         } = LoadedStorage::load_storage(&storage)?;
 
+        let banned_addresses = banned_addresses
+            .into_iter()
+            .map(|(addr, dur)| (addr, Time::from_duration_since_epoch(dur)))
+            .collect::<BTreeMap<BannableAddress, Time>>();
+
```
ImplOfAnImpl, line 99: It'd be nicer if `LoadedStorage` used Time already,

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -327,10 +331,11 @@ impl<S: PeerDbStorage> PeerDb<S> {
 
     /// Changes the address state to banned
     pub fn ban(&mut self, address: BannableAddress) {
-        let ban_till = self.time_getter.get_time() + *self.p2p_config.ban_duration;
+        let ban_till = (self.time_getter.get_time() + *self.p2p_config.ban_duration)
+            .expect("Ban duration is expected to be valid");
 
         update_db(&self.storage, |tx| {
-            tx.add_banned_address(&address.to_string(), ban_till)
+            tx.add_banned_address(&address.to_string(), ban_till.as_duration_since_epoch())
```
ImplOfAnImpl, line 338: Again, `add_banned_address` could already accept `Time`.

p2p/src/sync/peer_v1.rs
```diff
@@ -555,10 +555,15 @@ where
...
+        let max_accepted_time = (self.time_getter.get_time()
+            + *self.chain_config.max_future_block_time_offset())
+        .expect("Both values come from this node's clock; so cannot fail");
+        let max_block_timestamp = BlockTimestamp::from_time(max_accepted_time);
         if block_timestamp > max_block_timestamp {
+            let block_timestamp = block_timestamp.as_duration_since_epoch();
+            let max_block_timestamp = max_block_timestamp.as_duration_since_epoch();
+
             let clock_diff =
                 block_timestamp.checked_sub(max_block_timestamp).unwrap_or_else(|| {
```
ImplOfAnImpl, line 568: I'd suggest to implement checked subtraction for `Time` (which I'd call `checked_diff`) and use it here directly instead of converting times to durations.

Same for the V2 peer.

p2p/src/peer_manager/mod.rs
```diff
@@ -1515,7 +1522,8 @@ where
                 // that can have `discovered_own_address`.
                 let delay = (RESEND_OWN_ADDRESS_TO_PEER_PERIOD / OUTBOUND_FULL_RELAY_COUNT as u32)
                     .mul_f64(utils::exp_rand::exponential_rand(&mut make_pseudo_rng()));
-                next_time_resend_own_address += delay;
+                next_time_resend_own_address = (next_time_resend_own_address + delay)
+                    .expect("Time derived from local clock; cannot fail");
```
ImplOfAnImpl, line 1526: Again, `AddAssign` would be nice to have.

p2p/src/peer_manager/peerdb/address_data/tests.rs
```diff
@@ -77,11 +77,11 @@ fn reachable_reconnects(#[case] seed: Seed) {
             address.transition_to(AddressStateTransitionTo::ConnectionFailed, now, &mut rng);
             connection_attempts += 1;
         }
-        now += Duration::from_secs(60);
+        now = (now + Duration::from_secs(60)).unwrap();
```
ImplOfAnImpl, line 80: Same comment about `AddAssign`

ImplOfAnImpl (approved):
Yay!

common/src/chain/block/timestamp.rs
```diff
@@ -54,6 +56,14 @@ impl BlockTimestamp {
...
     }
 
+    pub const fn from_time(time: Time) -> Self {
+        Self::from_duration_since_epoch(time.as_duration_since_epoch())
+    }
+
+    pub const fn into_time(self) -> Time {
+        Time::from_duration_since_epoch(self.as_duration_since_epoch())
+    }
+
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I thought about that, but the reason I didn't go for it was that Time and BlockTimestamp don't necessarily have 1:1 equivalence. In fact, I was thinking of calling the function in BlockTimestamp "approximate_from_time()".

common/src/primitives/time.rs
```diff
@@ -50,12 +50,70 @@ pub fn set(now: Duration) -> Result<(), std::num::TryFromIntError> {
...
+        Self {
+            time: self.time.saturating_add(duration),
+        }
+    }
+
+    pub const fn saturating_duration_sub(&self, t: Duration) -> Self {
+        Self {
+            time: self.time.saturating_sub(t),
+        }
+    }
```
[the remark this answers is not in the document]
TheQuantumPhysicist: This also occurred to me, and I'm not insistent about it. My only issue with it is that usually saturating_sub usually takes Self. Using something other than Self seems unnatural.

common/src/primitives/time.rs
```diff
@@ -50,12 +50,70 @@ pub fn set(now: Duration) -> Result<(), std::num::TryFromIntError> {
...
+}
+
+impl std::ops::Add<Duration> for Time {
+    type Output = Option<Self>;
+
+    fn add(self, other: Duration) -> Option<Self> {
+        self.time.checked_add(other).map(|time| Self { time })
+    }
+}
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Please don't take this the wrong way... but... the irony... comparing this to issue #1221... :smile:

Having said that: I do prefer enforcing checked math wherever possible... especially when it's global like this one.

I do understand there are many expects, but I'm OK with that as long as it's fully controlled. In fact, for uint, we're moving to further checked_add by default everywhere.

common/src/primitives/time.rs
```diff
@@ -50,12 +50,70 @@ pub fn set(now: Duration) -> Result<(), std::num::TryFromIntError> {
...
 
 /// Either gets the current time or panics
-pub fn get_time() -> Duration {
+pub fn get_time() -> Time {
     match get_mocked_time() {
-        Some(mocked_time) => mocked_time,
-        None => SystemTime::now()
-            .duration_since(SystemTime::UNIX_EPOCH)
-            .expect("Time went backwards"),
+        Some(mocked_time) => Time::from_duration_since_epoch(mocked_time),
```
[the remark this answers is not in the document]
TheQuantumPhysicist: You're right. Will fix.

common/src/primitives/time.rs
```diff
@@ -71,23 +129,38 @@ mod tests {
         logging::init_logging();
         set(Duration::from_secs(1337)).unwrap();
 
-        log::info!("p2p time: {}", get_time().as_secs());
+        log::info!(
+            "p2p time: {}",
+            get_time().as_duration_since_epoch().as_secs()
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I did it, and it turned out there's like 4 instances of that in the whole source code, besides this...

dns_server/src/crawler_p2p/crawler/mod.rs
```diff
@@ -121,7 +121,7 @@ impl Crawler {
         loaded_addresses: BTreeSet<SocketAddress>,
         reserved_addresses: BTreeSet<SocketAddress>,
     ) -> Self {
-        let now = Duration::ZERO;
+        let now = common::primitives::time::get_time();
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I removed the comment

mempool/src/pool/orphans/test.rs
TheQuantumPhysicist: Done

p2p/src/net/default_backend/types.rs
TheQuantumPhysicist: Same issue

p2p/src/peer_manager/mod.rs
TheQuantumPhysicist: Doesn't work with `Option<_>`

p2p/src/peer_manager/peerdb/address_data/mod.rs
TheQuantumPhysicist: The reason why I used a loop here is that the value of totally probabilistic. I can't prove this won't produce an infinitely large value. Statistically, it's possible to prove the standard deviation is bounded. But for singular values, the standard deviation is virtually infinite, so it can be anything.

common/src/primitives/time.rs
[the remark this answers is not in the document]
ImplOfAnImpl: As you say, but I'd argue that this comment and the mentioned issue have nothing in common.
Here we talking about whether to panic implicitly or explicitly. IMO for types like Duration implicit panics are OK and since with explicit ones we lose convenience (quite a lot of it IMO), I'd go with the former. But I can see your point too.
In the issue though, the question is whether we should panic in certain situations or not.

p2p/src/peer_manager/peerdb/mod.rs
TheQuantumPhysicist: Done

p2p/src/peer_manager/peerdb/mod.rs
TheQuantumPhysicist: Done

p2p/src/peer_manager/peerdb/address_data/mod.rs
ImplOfAnImpl: I mean, as I can see, the result of `exponential_rand` is either relatively small or is infinity (because it's minus logarithm of a number from [0,1), so its largest finite value is `-f64::MIN_POSITIVE.ln()` which is `708.3964185322641`). Also, if it's infinite, `mul_f64` will panic, so the loop won't have a chance to fix this problem.
And `next_connect_delay` returns at most `MAX_DELAY_REACHABLE`, which is 1 hour. So, under normal circumstances the loop won't do anything (unless I'm missing something of course).
And if we expect some "abnormal" circumstances, it still won't solve all the problems, e.g. if `exponential_rand` could return arbitrarily large finite values, `next_connect_time` could produce a value which is billions of years into the future.

I'd suggest that we `clamp` the result of `exponential_rand` here instead, so that it's guaranteed to be reasonable.
(It also makes sense to fix `exponential_rand` so that it always return finite values. But `next_connect_time` should probably still not rely on the values being small).

TheQuantumPhysicist: Nothing prevents the random number from being large enough to cause an overflow in addition, but small enough to not trigger the mul_f64 overflow. That's totally possible. On the other hand, statistically speaking, the loop will never iterate more than once, if ever...

Clamping sounds like a reasonable solution. The only issue with it is that it breaks encapsulation. I'm happy to add a TODO to fix this later. Also, if you know how to fix this in a few minutes + add a test to ensure the correct behavior, please feel free to do it, and push a commit here.

alfiedotwtf (approved):

iljakuklic (approved):

common/src/primitives/time.rs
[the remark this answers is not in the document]
iljakuklic: One of the problems you hit when coming up with a name for this is that both `Time - Duration` and `Time - Time` are sensible operations. Plain `saturating_sub` should probably be reserved for the latter, even if it's never implemented.

azarovh (approved):

PR #1238, opened 2023-09-27
RPC json elements now can return proper addresses instead of hex values

### LINKED ISSUE #1200: Make it possible to encode to addresses when encoding blocks or transactions to json
In #1155, we serialize using serde_json. However, while we can serialize destinations and ids to hex, some of these elements are better serialized to a proper readable address. The problem is that the serialization of these elements depends on chain config, and serde_json doesn’t seem to have a way to accept a state, such as chain config.
We would like to avoid having to use a global variable that stores chain config for this.

TheQuantumPhysicist:
So, given that `serde::Serialize` has no state, this is the only way I could do this. Let me know what you think.

Closes #1200

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1180,6 +1180,7 @@ dependencies = [
  "parity-scale-codec",
  "proptest",
  "ref-cast",
+ "regex",
  "rstest",
  "script",
  "serde",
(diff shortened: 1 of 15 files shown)


OBorce (approved):
Looks OK

muursh (approved):

PR #1241, opened 2023-09-28
Mlt -> CoinUnit

### LINKED ISSUE #1178: Rename Mlt to ML in all the sources
From @TheQuantumPhysicist 
> We decided to move away from MLT. I guess it has to be changed in many places.

TheQuantumPhysicist:
Two problems are solved in this PR:

1. The struct Mlt is renamed to CoinUnit, to keep the code coin-agnostic... this may be annoying in cases where we're dealing with testnet or similar. It should have a different ticker.

2. The ticker configuration is removed from that struct, because it should be different for different networks. It was being set as a constant everywhere. Now it comes from `ChainType`.

Closes #1178

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,7 +62,9 @@ jobs:
       - name: Build
         run: cargo build --release --locked
       - name: Run tests
-        run: cargo test --release --all
+        run: cargo test --release --workspace
+      - name: Run doc tests
+        run: cargo test --release --doc
       # This test is ignored, so it needs to run separately.
       - name: Run mixed_sighash_types test
         run: cargo test --release -p common mixed_sighash_types -- --ignored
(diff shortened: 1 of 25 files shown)


common/src/chain/config/emission_schedule.rs
```diff
@@ -76,11 +76,11 @@ impl std::fmt::Debug for EmissionSchedule {
 #[derive(Eq, PartialEq, Debug, Clone)]
 pub struct EmissionScheduleTabular {
     /// The initial supply, in MLTs
```
azarovh, line 78: MLTs?

common/src/chain/config/mod.rs
```diff
@@ -550,7 +559,7 @@ fn create_mainnet_genesis() -> Genesis {
 
 fn create_testnet_genesis() -> Genesis {
     // We add 3_600_000_000 MLT to the genesis mint account since it's just for testing. Nothing else changes.
```
azarovh, line 561: MLT?

azarovh (approved):

iljakuklic (approved):

.github/workflows/build.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - name: Build
         run: cargo build --release --locked
       - name: Run tests
-        run: cargo test --release --workspace
+        run: cargo test --release --all
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Doc tests weren't running... I was wondering this or another step...

iljakuklic (approved):

PR #1243, opened 2023-09-28
Use only checked math ops for Uint* types

### LINKED ISSUE #1218: Use checked functions for all uint operations
All types created from `common/src/uint/impls.rs` must use checked operations only, the same way as Amount does.

TheQuantumPhysicist:
Closes #1218

### DIFF
--- a/common/src/chain/block/consensus_data.rs
+++ b/common/src/chain/block/consensus_data.rs
@@ -145,9 +145,9 @@ impl PoWData {
         let target: Uint256 = self.bits.try_into().ok()?;
         let mut ret = !target;
         let mut ret1 = target;
-        ret1.increment();
+        ret1 = (ret1 + Uint256::ONE)?;
         ret = (ret / ret1)?;
-        ret.increment();
+        ret = (ret + Uint256::ONE).unwrap_or(Uint256::MAX);
         Some(ret)
     }
 }
--- a/common/src/chain/block/consensus_data.rs
+++ b/common/src/chain/block/consensus_data.rs
@@ -146,7 +146,7 @@ impl PoWData {
         let mut ret = !target;
         let mut ret1 = target;
         ret1.increment();
-        ret = ret / ret1;
+        ret = (ret / ret1)?;
         ret.increment();
         Some(ret)
     }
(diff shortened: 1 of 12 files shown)


azarovh (approved):

common/src/chain/block/consensus_data.rs
```diff
@@ -146,7 +146,7 @@ impl PoWData {
         let mut ret = !target;
         let mut ret1 = target;
         ret1.increment();
-        ret = ret / ret1;
+        ret = (ret / ret1)?;
```
iljakuklic, line 149: The `ìncrement` operation is still not checked.

consensus/src/pow/helpers.rs
```diff
@@ -62,21 +64,22 @@ where
...
     let old_target = Uint256::try_from(old_target)
         .map_err(|_| ConsensusPoWError::PreviousBitsDecodingFailed(old_target))?;
 
     // new target is computed by  multiplying the old target by ratio of the actual timespan / target timespan.
     // see Bitcoin's Protocol rules of Difficulty change: https://en.bitcoin.it/wiki/Protocol_rules
-    let mut new_target = old_target * actual_timespan;
-    new_target = new_target / target_timespan;
+    let mut new_target = (old_target * actual_timespan).unwrap_or(Uint256::MAX);
+    new_target =
+        (new_target / target_timespan).expect("target_timespan cannot be zero as per its type");
```
iljakuklic, line 82: Not for this PR: This expect could be avoided if we had a `NonZero` version of the numeric types and an exceptionless division taking a `NonZeroBlah`.

iljakuklic (approved):
+1

common/src/chain/block/consensus_data.rs
TheQuantumPhysicist: Maybe I should get rid of it altogether... Not sure how to fix this.

consensus/src/pow/helpers.rs
TheQuantumPhysicist: It occurred to me yesterday while I was dancing with PoW target time, but that's a huge undertaking and I don't think it's worth it.

common/src/chain/block/consensus_data.rs
TheQuantumPhysicist: I got rid of it and discovered a bug in PoW... thanks for pointing it out.

PR #1246, opened 2023-09-29
Feature/regtest options cli wallet

OBorce:
- Add an option to change the regtest chain config options for the CLI wallet same as the GUI wallet and the node deamon.
- Add a call to decommission stake pool in the functional tests and check for locked balance after that.
- Refactor asserts in the wallet functional tests for better error messages if they fail.


wallet/wallet-cli-lib/tests/cli_test_framework.rs
```diff
@@ -298,7 +298,22 @@ impl CliTestFramework {
         let manager_task = manager.main_in_task();
 
         let wallet_options = WalletCliArgs {
-            network: Network::Regtest,
+            network: Network::Regtest(Box::new(config::regtest_options::ChainConfigOptions {
+                chain_magic_bytes: None,
```
azarovh, line 302: looks like a good reason to add `ChainConfigOptions::new`

azarovh (approved):

wallet/wallet-cli-lib/tests/cli_test_framework.rs
TheQuantumPhysicist: I think this is fine... with this huge number of parameters, using positional arguments would be a disaster.

TheQuantumPhysicist (approved):

wallet/wallet-cli-lib/tests/cli_test_framework.rs
azarovh: I meant `new()` or default

azarovh: can be dropped as discussed

Issue #1239, opened 2023-09-28
next_connect_time() uses a relatively complicated way to ensure the sanity of the PDF used


TheQuantumPhysicist:
In the function [`next_connect_time()`](https://github.com/mintlayer/mintlayer-core/blob/7cc49186b0b6ed04f8c15c6a0d22b1696dc1d762/p2p/src/peer_manager/peerdb/address_data/mod.rs#L166C22-L166C22), a solution that depends on the statistical properties of a probability density function is used. Now the loop is supposedly sane, but it raises many questions due to its complexity as in [this discussion](https://github.com/mintlayer/mintlayer-core/pull/1235#discussion_r1337479279).

The code should be cleaned up and rearranged to minimize code-complexity risk.

PR #1249, opened 2023-09-29
Fix next_connect_time

ImplOfAnImpl:
1) `exponential_rand` can no longer produce infinity.
2) `next_connect_time` now clamps the generated factor to a reasonable value (because the result of `exponential_rand` can still be quite big for a factor)

This fixes https://github.com/mintlayer/mintlayer-core/issues/1239


utils/Cargo.toml
```diff
@@ -23,6 +23,7 @@ zeroize.workspace = true
 test-utils = { path = "../test-utils" }
 
 criterion.workspace = true
+rand.workspace = true
```
TheQuantumPhysicist, line 26: We do not use rand from anywhere except from crypto library. This makes audits much easier.

ImplOfAnImpl: > We do not use rand from anywhere except from crypto library. This makes audits much easier.

Does this apply to dev-dependencies too? As I can see, rand is sometimes used as a dev dependency currently.

TheQuantumPhysicist: Nowhere. All randomness should come from the crypto library.

TheQuantumPhysicist (approved):

PR #1250, opened 2023-09-30
Introduce `PoSChainConfigBuilder` and change maturity distances for Testnet

azarovh:
The main goal for this PR is to change maturity distances for `Testnet` from 2000 to 7200 for both decommissioning and spending a share.  
But the way `PoSChainConfig` had been initialising was inconvenient so I added `PoSChainConfigBuilder`.

### DIFF
--- a/blockprod/src/detail/tests.rs
+++ b/blockprod/src/detail/tests.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{num::NonZeroU64, sync::Arc, time::Duration};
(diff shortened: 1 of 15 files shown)


TheQuantumPhysicist (approved):

TheQuantumPhysicist (changes requested):
As agreed, the builder should go to unit tests only to simplify, and there should not be defaults that specify complex consensus rules.

TheQuantumPhysicist (approved):

muursh (approved):

iljakuklic (approved):

PR #1252, opened 2023-10-02
Add function to ensure that consensus will never be IgnoreConsensus in production

### LINKED ISSUE #792: Ensure `ConsensusUpgrade::IgnoreConsensus` is not used in production code
It should be impossible to use `ConsensusUpgrade::IgnoreConsensus` anywhere other than in unit tests.

TheQuantumPhysicist:
Closes #792 

Please check if I forgot any corner case.

### DIFF
--- a/common/src/chain/config/mod.rs
+++ b/common/src/chain/config/mod.rs
@@ -650,25 +650,27 @@ pub fn assert_no_ignore_consensus_in_chain_config(chain_config: &ChainConfig) {
 
     let all_upgrades = upgrades.all_upgrades();
 
-    if all_upgrades.is_empty() {
-        panic!("Invalid chain config. There are no net-upgrades defined, not even for genesis.")
-    }
+    assert!(
+        !all_upgrades.is_empty(),
+        "Invalid chain config. There are no net-upgrades defined, not even for genesis."
+    );
 
-    if all_upgrades.len() < 2 {
-        panic!("Invalid chain config. There must be at least 2 net-upgrades defined, one for genesis and one for the first block after genesis.")
-    }
+    assert!(all_upgrades.len() >= 2, "Invalid chain config. There must be at least 2 net-upgrades defined, one for genesis and one for the first block after genesis.");
 
-    if all_upgrades[0].0 != 0.into() {
-        panic!("Invalid chain config. The first net-upgrade must be at height 0");
-    }
+    assert!(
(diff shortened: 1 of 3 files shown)


azarovh (approved):

common/src/chain/config/mod.rs
```diff
@@ -637,6 +637,40 @@ pub fn create_unit_test_config() -> ChainConfig {
...
+    let all_upgrades = upgrades.all_upgrades();
+
+    for upgrade in all_upgrades {
+        let upgrade_height = &upgrade.0;
+        let upgrade_data = &upgrade.1;
+
+        // We skip genesis
+        if *upgrade_height == 0.into() && all_upgrades.len() > 1 {
+            continue;
+        }
```
iljakuklic, line 660: Is the first upgrade required to be at genesis? If so, we can just assert the first upgrade is at height 0 and only iterate over `all_upgrades.into_iter().skip(1)`.

common/src/chain/upgrades/netupgrade.rs
```diff
@@ -42,6 +42,19 @@ impl NetUpgrades<UpgradeVersion> {
         )])
     }
 
+    pub fn deliberate_ignore_consensus_twice() -> Self {
+        Self(vec![
```
iljakuklic, line 46: As far as I can see, this is only used in tests. Should it be guarded by `#[cfg(test)]`?

iljakuklic (approved):

common/src/chain/upgrades/netupgrade.rs
TheQuantumPhysicist: Done

common/src/chain/upgrades/netupgrade.rs
```diff
@@ -57,6 +70,10 @@ impl NetUpgrades<UpgradeVersion> {
             ),
         ])
     }
+
+    pub fn all_upgrades(&self) -> &Vec<(BlockHeight, UpgradeVersion)> {
+        &self.0
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Done

common/src/chain/config/mod.rs
TheQuantumPhysicist: Done... despite it being more verbose.

iljakuklic (approved):

muursh (approved):

Issue #889, opened 2023-05-11
Improve the p2p subsystem creation


stanislav-tkach:
Currently the intermediate `P2pInit` structure is returned from the `make_p2p` function. Also the subsystem creation requires a custom event loop:
```rust
let p2p: P2pInit = make_p2p(
        Arc::clone(&chain_config),
        p2p_config,
        chainstate.clone(),
        mempool.clone(),
        Default::default(),
        peerdb_storage,
    )
    .unwrap();
    let p2p = manager.add_subsystem_with_custom_eventloop("p2p", {
        move |call, shutdown| p2p.run(call, shutdown)
    });
``` 
This is required because of the two reasons:
1. The `P2p` structure is generic over the transport and the concrete transport is chosen based on runtime parameters, so it is impossible to return it from a function directly, only as `Box<dyn P2pInterface>`.
2. P2p requires a specific shutdown logic, but it cannot be placed in the `Subsystem::shutdown` function because the `Subsystem` trait is implemented for `Box<dyn P2pInterface>` and not for the `P2p` structure directly.

Ideally neither the custom event loop nor the `P2pInit` structure should be needed, and the shutdown logic should be moved to the implementation of the `Subsystem` trait for `P2p`.

TheQuantumPhysicist:
I'm closing this as unnecessary. We may change our minds later, but for now it's fine.

Issue #131, opened 2022-04-08
Take advantage of the separation between mutable and immutable subsystem calls


iljakuklic:
We have two methods to call functions on subsystems: `Subsystem::call` and `Subsystem::call_mut`. These expose the subsystem internal state as an immutable and mutable reference, respectively. Currently, the `call` method just delegates to the `call_mut` method. A separate implementation for `call` could be provided that takes the advantage of the immutability to execute multiple calls simultaneously.

Ivansete:
Afternoont @iljakuklic,

While checking this issue I don't fully understand why "call_mut" is actually needed. I have the impression that all the Actions, and their CallResult responses, could be done properly throughout the UnboundedChannel only using the immutable methods (System::call) but I might be missing some details.

May kindly share a couple of use cases where "call_mut" and "call" is needed and why please ? That will help me to understand it better.

Thanks in advance

iljakuklic:
> Good afternoon @iljakuklic,
> 
> While checking this issue I don't fully understand why "call_mut" is actually needed. I have the impression that all the Actions, and their CallResult responses, could be done properly throughout the UnboundedChannel only using the immutable methods (System::call) but I might be missing some details.

In `call_mut` you are given a `&mut` reference to the subsystem object whereas `call` only gives you access to `&` reference to the subsystem object. The former is necessary if you want to apply any modifying operation (e.g. call a method that is defined on `&mut self`) on the subsystem.

The idea behind this issue is that when using shared references (`&`), multiple calls could be potentially dispatched and processed simultaneously in multiple threads/tasks. The difficulty is that more synchronisation will be needed to ensure that a `call_mut` is not processed simultaneously with another `call_mut` or with a `call`. It's hard to tell whether this would be a win overall, given the need for extra synchronisation, the fact that new tasks would have to be spawned, etc.

> May kindly share a couple of use cases where "call_mut" and "call" is needed and why please ? That will help me to understand it better.

You can have a look in `subsystem/tests/passive.rs`. It contains a simple example. It's rather artificial but it illustrates the high level concepts reasonably well.

> Thanks in advance

Hope this helps.

Ivansete:
Morning @iljakuklic and thanks for the reply.

I understand it but I have the impression that we could never achieve a parallel execution as there is a tokio::unbounded mpsc channel, meaning that all messages are enqueued and only processed by one single receiver, i.e. the messages can only be processed on after the other, sequentially.

[...]

PR #1256, opened 2023-10-03
subsystem: Remove custom event loop support

iljakuklic:
* Generalized the built-in event loop with
  * Custom initialization
  * Separation between subsystem state object and interface
  * Ability to specify background work in the `Subsystem` trait
* Ported `p2p` to the built-in event loop
* Ported tests to the built-in event loop
* Removed support for custom event loop from subsystem
* Substantial `subsystem` crate code reorganization

Related:
* #131 
* #889


TheQuantumPhysicist (approved):
Looking good. Great work!

mempool/src/interface/mempool_interface_impl.rs
```diff
@@ -149,16 +116,6 @@ impl MempoolImpl {
         !self.work_queue.is_empty()
     }
```
TheQuantumPhysicist, line 118: There's a little typo "parametrizing" down there.

rpc/examples/simple_server.rs
```diff
@@ -75,9 +73,9 @@ impl SomeSubsystemRpcServer for SomeSubsystemHandle {
 async fn main() -> anyhow::Result<()> {
     logging::init_logging();
 
-    let mut app = subsystem::Manager::new("rpc-example");
-    app.install_signal_handlers();
-    let some_subsystem = app.add_subsystem("some_subsystem", SomeSubsystem(0));
+    let config = subsystem::ManagerConfig::new("rpc-exmaple").enable_signal_handlers();
```
TheQuantumPhysicist, line 76: exmaple

subsystem/src/calls/handle.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    /// Submit an async procedure to be performed by the subsystem.
+    pub fn submit_async(
+        &self,
+        func: impl FnOnce(&T) -> BoxFuture<()> + Send + 'static,
+    ) -> Result<(), SubmissionError> {
+        self.send_action(Action::Ref(Box::new(func)))
+    }
+
+    /// Submit a procedure to be preformed by the susbsystem (mutable).
```
TheQuantumPhysicist, line 72: susbsystem

subsystem/src/calls/handle.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn submit_mut(
+        &self,
+        func: impl FnOnce(&mut T) + Send + 'static,
+    ) -> Result<(), SubmissionError> {
+        self.send_action(Action::Mut(Box::new(move |subsys| {
+            Box::pin(async { func(subsys) })
+        })))
+    }
+
+    /// Submit a procedure to be preformed by the susbsystem.
```
TheQuantumPhysicist, line 82: susbsystem

subsystem/tests/basic.rs
```diff
@@ -25,15 +25,14 @@ fn basic_passive_shutdown() {
...
-            let _substr = app.add_subsystem("substr", Substringer::new("abc".into()));
-            let _counter = app.add_subsystem("counter", Counter::new());
+            let _substr = app.add_direct_subsystem("substr", Substringer::new("abc".into()));
+            let _counter = app.add_direct_subsystem("counter", Counter::new());
 
-            // Start a subsystem that immediately terminates, instructing the remaining subsystems
-            // to terminate too.
-            let _shut: subsystem::Handle<()> =
-                app.add_subsystem_with_custom_eventloop("terminator", |_, _| async {});
+            // Task that shuts down the subsysytem
```
TheQuantumPhysicist, line 31: subsysytem

mempool/src/interface/mempool_interface_impl.rs
iljakuklic: Can't see what's wrong there?

TheQuantumPhysicist: Never mind. My spell-checker went nuts.

subsystem/src/calls/handle.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// Call an async procedure to the subsystem (immutable).
+    pub fn call_async<R: Send + 'static>(
+        &self,
+        func: impl FnOnce(&T) -> BoxFuture<R> + Send + 'static,
+    ) -> CallResult<R> {
+        let (rtx, rrx) = tokio::sync::oneshot::channel::<R>();
+        let result = self.0.submit_async(move |subsys| {
+            Box::pin(async move {
+                if rtx.send(func(subsys).await).is_err() {
+                    log::trace!("Subsystem call (mut) result ignored");
```
alfiedotwtf, line 144: Should the "(mut)" be here?

muursh (approved):
I'm alright with this

subsystem/src/manager/manager_impl.rs
```diff
@@ -0,0 +1,301 @@
...
+        // Shut down the subsystems in the reverse order of creation.
+        for subsys in subsystems.into_iter().rev() {
+            subsys.shutdown(self.config.shutdown_timeout_per_subsystem).await;
+        }
+
+        #[cfg(not(loom))]
+        if let Some(hdl) = signal_handler_task {
+            hdl.abort();
+            let _ = hdl.await;
+        }
```
alfiedotwtf, line 181: Not sure if this is relevant, but could any of the subsystems interact with the signal handler? If so, this block should probably be put above the shutdown for loop above.

subsystem/src/calls/handle.rs
iljakuklic: It should not. Well spotted, thanks. Fixed.

subsystem/src/manager/manager_impl.rs
iljakuklic: The shutdown signal may be delivered multiple times but there is no issue with it. That is a supported scenario because it's easily avoided in a concurrent system, e.g. you can issue a node shutdown RPC and hit Ctrl-C and both will issue the shutdown signal.

I'p pretty sure the shutdown procedure can be improved. For example, I don't think the Ctrl-C handler needs to be in a separate task at all. That requires further research. However, I think what we have now is correct.

iljakuklic: So I changed this so all the shutdown signals are rolled into a Future rather than having a separate task that listens to `Ctrl-C` and such. It makes the code somewhat simpler. It's in the last commit. Please take a look if that makes sense to you @alfiedotwtf.

alfiedotwtf: Much cleaner. I like it.

TheQuantumPhysicist (approved):

Issue #1199, opened 2023-09-18
Things to do/consider doing in the p2p protocol v2


TheQuantumPhysicist:
(Created on Aug, 16 2023)

- [x] V2 should only send (and allow receiving) "connected" block header lists, i.e. the ones that are connected to the chainstate of the receiving side (i.e. the parent of the first header in the list should exist in the chainstate).
Note: the old V1 implementation used to perform block announcements by sending a single header, which was generally "unconnected", so it had to allow receiving single unconnected headers as well. Because of that, the newer V1 implementation has to allow single unconnected headers too, though it itself never sends those.

- [x] When sending new headers, V2 should take into account the best header that it has sent so far.
Note: the newer V1 implementation already tracks it in its incoming.best_sent_block_header field. However, it cannot use it to determine which headers to send without the risk of breaking the V1 protocol, because the old implementation is known to have issues with correctly updating its "peer's known headers" list. So, if the sent header list is connected to best_sent_block_header, an old peer may see it as disconnected in some cases. And since even old implementations only allow single unconnected headers, this may lead to a ban.

- [ ] ~(do we need this?) V2 should better not rely on the number of headers in the header list response to determine whether the peer has more headers. Instead, this fact should be specified in the response explicitly.
Note: the current implementation uses msg_header_count_limit as the header count limit; if the number of received headers equals the limit, the sending side is expected to have more headers. So, if the sending side has exactly msg_header_count_limit headers, it'll result in an unneeded extra headers request and an empty response.~
(This is being removed because it doesn't help, and it's normal to send an empty message to indicate that "there's nothing more to send)

- [x] In V2 we might want to allow peers to ask for headers even if the node is in initial block download state (IBD), e.g. based on some kind of system of permissions. At this moment it's not clear whether it's a good idea, so it makes sense to first check whether bitcoin does something like that.

- [x] In V2 we probably should not silently ignore header requests; instead, we should communicate our best block (id/height/header?) from the start, so that the peer just knows that we don't have better blocks and doesn't ask us in the first place. Or, perhaps, the fact that the node is in IBD can be communicated as well.
Note: In V1, header requests can be ignored if the node is in IBD; this makes the handling of stalling a bit more convoluted than necessary.

PR #1258, opened 2023-10-04
Respond with an empty headers list if in IBD state

azarovh:
Closes bullets 5 and 6 in #1199

v1 logic: ignore headers list request. If the node that requested headers receives blocks request in assumes that the peer is in Initial Block Download state and won't disconnect it as stalled.

new v2 logic: if in IBD respond with an empty list. It's simple and effective. Empty headers list is legal and no assumptions are made. Bitcoin does the same. Todos suggested some advanced logic with exchanging best block on handshake or some permission system. Idk it looks too complicated. Empty list does the job + serves as a pong.

P.S. also in v2 I moved call to the `wait_for_clock_diff` to allow more checks to pass before waiting.

### DIFF
--- a/p2p/src/sync/peer_v2.rs
+++ b/p2p/src/sync/peer_v2.rs
@@ -356,6 +356,7 @@ where
             // Note: msg_max_locator_count is not supposed to be configurable outside of tests,
             // so we should never get here in production code. Moreover, currently it's not
             // modified even in tests. TODO: make it a constant.
+            // See https://github.com/mintlayer/mintlayer-core/issues/1201
             log::warn!(
                 "[peer id = {}] Sending locator of the length {}, which exceeds the maximum length {:?}",
                 self.id(),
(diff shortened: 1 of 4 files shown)


p2p/src/sync/peer_v2.rs
```diff
@@ -405,16 +406,9 @@ where
...
-            // not clear whether it's a good idea, so it makes sense to first check whether bitcoin
-            // does something like that.
-            // TODO: in the protocol v2 we should not silently ignore header requests; instead,
-            // we should communicate our best block (id/height/header?) from the start, so that
-            // the peer just knows that we don't have better blocks and doesn't ask us in the
-            // first place.
-            // See the issue #1110.
             log::debug!("[peer id = {}] Ignoring headers request because the node is in initial block download", self.id());
+            // Respond with an empty list to avoid being marked as stalled
+            self.send_headers(HeaderList::new(Vec::new()))?;
```
TheQuantumPhysicist, line 411: There's an open question here whether this has a semi-predicate problem... do we need to distinguish between IBD and not? By using this kind of empty response, we're doing the same "special value semantics" that can only lead to problems. Isn't it more reasonable to wrap this in an enum stating that it's empty because "I'm in IBD"?

TheQuantumPhysicist: The open question is whether there's some benefit we're not noticing.

TheQuantumPhysicist (approved):

Issue #1189, opened 2023-09-15
API server design/creation


alfiedotwtf:
(Copied over from Gitlab)

This issue incorporates the API server tasks that are required for its initial release

- [ ] Use the storage API to read data from the database in the web server   
- [ ] Write tests for the web server based on mocks of the database and ensure the behavior and format is acceptable   
- [ ] Attempt to store data in postgres from the scanner, and write the config/command line arguments to be used to access the database (host, port, credentials, etc)   
- [ ] Write the config/command line arguments for configuring the web server to access postgres (host, port, credentials, etc)   
- [ ] See if we can have a standalone test to write and read data from a docker instance of postgres. This can go into CI   
- [ ] Figure out the additional required functions, besides the basic ones available now, like get_best_block, to go into the API, and add them one by one
  - [ ] Currently there's no way to see all available staking pools with their information to know what to delegate to   
- [ ] Create tests that attempt SQL injection attacks on the web API

alfiedotwtf:
I've updated the description with the TODOs to get the Address endpoint complete.

PR #1259, opened 2023-10-05
Initial whole-stack tests for the API Server 

alfiedotwtf:
Part of #1189


api-server/stack-test-suite/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
```
TheQuantumPhysicist, line 1: Why do we need a whole new crate for these tests? The whole point of having in-memory storage is to make it casual... just like we're doing for the chainstate test suite or block production. We can construct a whole chain in unit tests.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -30,12 +30,18 @@ pub enum BlockchainStateError {
 }
 
 pub struct BlockchainState<S: ApiServerStorage> {
-    storage: S,
+    storage: Option<S>,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: I fixed this by creating an async clone.

api-server/web-server/src/api/mod.rs
```diff
@@ -14,3 +14,45 @@
...
+    socket: TcpListener,
+    state: APIServerWebServerState<Arc<T>>,
+) -> Server<hyper::server::conn::AddrIncoming, IntoMakeService<Router>> {
+    let routes = Router::new()
+        .route("/", get(server_status))
+        .nest("/api/v1", api::v1::routes())
+        .fallback(bad_request)
+        .with_state(state);
+
+    axum::Server::from_tcp(socket).unwrap().serve(routes.into_make_service())
```
iljakuklic, line 57: Clippy complains about the `unwrap` here

api-server/stack-test-suite/tests/in_memory.rs
```diff
@@ -0,0 +1,1160 @@
...
+    #[tokio::test]
+    async fn invalid_block_id() {
+        let url = "/api/v1/block/invalid-block-id";
+
+        let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+        let socket = listener.local_addr().unwrap();
+
+        tokio::spawn({
+            async move {
+                let web_server_state = {
```
iljakuklic, line 105: In addition to the race @TheQuantumPhysicist pointed out, also note that the tokio task is not cleaned up. Not sure to what extent it matters though.

Also a tiny style comment: There seems to be a redundant set of braces around the `async` block. Is that inserted by cargo fmt?

iljakuklic (commented):
Just a quick first pass

api-server/stack-test-suite/Cargo.toml
alfiedotwtf: Since these tests combined the scanner, API storage, and the web server, it didn't make sense to me at the time to have these tests just as unit tests within `web-server`.

> just like we're doing for the chainstate test suite or block production

I copied how `storage-test-suite` and chainstate's `test-framework` was done which also have a separate `[package]`.

Given the reasoning above, I'm happy to move them into the web-server unit tests if you want.

api-server/stack-test-suite/tests/in_memory.rs
```diff
@@ -0,0 +1,1160 @@
...
+                    }
+                };
+
+                web_server(listener, web_server_state).await
+            }
+        });
+
+        let response = reqwest::get(format!("http://{}:{}{url}", socket.ip(), socket.port()))
+            .await
+            .unwrap();
```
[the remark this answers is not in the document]
alfiedotwtf: I'm not too sure if this is actually a race...

The listening port is bound before splitting tasks, and so even if the socket hasn't been passed to `web_sever()` yet and so the `reqwest::get()` is called, since the socket is already listening, the `get()` will just block until it receives data.

I verified this on Mac OS by putting a `sleep(5)` on the first line within the `async move` task, and tests passed. Would also assume this works the same on all platforms.

api-server/scanner-lib/src/blockchain_state/mod.rs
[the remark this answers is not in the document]
alfiedotwtf: > So in your tests, you could just do local_node.storage().clone()

Once the storage was passed into the state, I needed it to be taken out again but couldn't because of a RwLock... but your change looks like it fixes my hack around the issue. Nice!

alfiedotwtf:
> It's still in draft, so I'm not sure what's left.

If you look at `web-server/src/api/v1.rs`, you'll see all the current endpoints. Currently only the `chain/` and `block/' have been done (with a hand full of TODOs within them), but there's `transaction/`, `destination/` and `pool/` that I'll work on today (and think about how to actually do `destination/` given that the current values are placeholders rather than going to the database.

api-server/web-server/src/error.rs
```diff
@@ -35,6 +35,14 @@ pub enum APIServerWebServerError {
 pub enum APIServerWebServerClientError {
     #[error("Bad request")]
     BadRequest,
+    #[error("Block not found")]
+    BlockNotFound,
```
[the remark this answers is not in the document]
alfiedotwtf: It's a bit awkward, but personally when I think "Web" on it's own I think "Web Client", and when looking at the `api-server/` as a whole, that makes sense - "APIServer" -> "WebServer" -> "ClientError".

api-server/stack-test-suite/tests/in_memory.rs
alfiedotwtf: Sorry that was me - a left over from when I was localizing a binding but have deleted them without cleaning up

iljakuklic:
> > It's still in draft, so I'm not sure what's left.
> 
> If you look at `web-server/src/api/v1.rs`, you'll see all the current endpoints. Currently only the `chain/` and `block/` have been done (with a hand full of TODOs within them), but there's `transaction/`, `destination/` and `pool/` that I'll work on today (and think about how to actually do `destination/` given that the current values are placeholders rather than going to the database.

Given the tests file is already over 1000 lines, would it make sense to break it down by endpoint tested? I guess the categorisation is not going to be 100% accurate since I can imagine tests that check a combination of endpoints but maybe better than a monolithic file.

api-server/stack-test-suite/Cargo.toml
TheQuantumPhysicist: That's fine. I just think it's an overkill.

api-server/stack-test-suite/tests/in_memory.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I don't think a sleep proves the race doesn't exist, because that depends on what kind of sleep you're doing and what. What kind of sleep did you use? If you used a synchronous sleep, that will just make the main async thread block and won't give the chance to other threads to take over, meaning that you probably didn't trigger proper concurrency. More importantly, you're using `tokio::test`, which is single-threaded, so you don't have parallelism anyway. 

To prove this, you should increase the number of threads and/or use async sleep.

alfiedotwtf:
> Given the tests file is already over 1000 lines, would it make sense to break it down by endpoint tested? I guess the categorisation is not going to be 100% accurate since I can imagine tests that check a combination of endpoints but maybe better than a monolithic file.

I use code folding in Emac a lot, and because I've put each set of tests within a separately named module within the file, when the test file is collapsed, it logically makes sense. Not as clear as spreading this out to multiple files though, but code folding does let you see the shape of the file and lets you drill down when needed.

alfiedotwtf:
I've added a few transaction() and transaction_merkle_path() tests, but there a many TODOs left in the comments.

This PR is getting quite big as it is, so maybe the TODOs can be done in a separate PR along with the Destination and Pool endpoint changes / tests.

TheQuantumPhysicist (approved):

TheQuantumPhysicist (approved):
Besides the TODOs, this is looking fine. I'm assuming these TODOs will be fixed next.

PR #1264, opened 2023-10-08
Fix p2p test timeout on coverage

azarovh:
This is the fix for the failed job that Boris reported earlier:
https://github.com/mintlayer/mintlayer-core/actions/runs/6425530789/job/17448253640?pr=1262

I see no other reason for the failure except that the syncing of 1000 blocks took more than 60s that is the standard timeout for the p2p tests. I suggest decreasing the number of processing blocks to 500 which won't make a difference for the quality of the test but allows to keep timeout the same.

### DIFF
--- a/p2p/p2p-test-utils/src/lib.rs
+++ b/p2p/p2p-test-utils/src/lib.rs
@@ -115,7 +115,7 @@ impl P2pBasicTestTimeGetter {
 }
 
 /// A timeout for blocking calls.
-pub const LONG_TIMEOUT: Duration = Duration::from_secs(60);
+pub const LONG_TIMEOUT: Duration = Duration::from_secs(120);
 /// A short timeout for events that shouldn't occur.
 pub const SHORT_TIMEOUT: Duration = Duration::from_millis(500);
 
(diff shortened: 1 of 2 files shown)


TheQuantumPhysicist:
This error has nothing to do with the 60 seconds you're seeing first. That's just rust testing framework warning that a test took more than 60 seconds.

azarovh:
> The error can be seen here:

That's exactly the error that fires on timeout. Every test calls `expect_future_val` on `SyncingEventReceiver` with 60s timeout.

That panic happens exactly 60s from the start. Also you can see that only block at `BlockHeight(895)` was processed.

TheQuantumPhysicist:
> > The error can be seen here:
> 
> That's exactly the error that fires on timeout. Every test calls `expect_future_val` on `SyncingEventReceiver` with 60s timeout.
> 
> That panic happens exactly 60s from the start. Also you can see that only block at `BlockHeight(895)` was processed.

Alrighty. I don't like the idea of reducing the test coverage just to appease a timeout. Maybe we should increase the timeout to 120 seconds. I think that's better.

TheQuantumPhysicist (approved):

PR #1268, opened 2023-10-10
Change binary names for consistency

erubboli:
### Changes to Binary Names and Default Behavior

Please consider this PR. I am happy to re-add `node-gui` to the defaults if this creates any issues. The goal of this PR is to simplify the build process. Most users are not interested in building the GUI, and it comes with many unwanted dependencies, such as GTK. Additionally, we aim to ensure consistency in binary names, as we are beginning to develop scripts and documentation that reference them.

#### 1. Updated Binary Names

The binary names have been updated as follows:

- `node-gui` is now `mintlayer-gui`
- `node-daemon` is now `mintlayerd`
- `wallet-cli` is now `mintlayer-cli`
- `wallet-address-generator` is now `mintlayer-address-generator`
- `web-server` is now `mintlayer-api`

To run the binaries, use the new names as in the example below:

```sh
cargo run --bin mintlayerd
```

#### 2. Changes to `node-gui`

`node-gui` has been removed from the default build and test targets. To run it, specify the package and binary explicitly:

```sh
cargo run -p node-gui --bin mintlayer-gui
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,7 +73,6 @@ default-members = [
   "logging",
   "mempool",
   "node-daemon",
-  "node-gui",
   "node-lib",
   "p2p",
   "rpc",
@@ -224,3 +223,4 @@ overflow-checks = true
 
 [profile.test.package.script]
 opt-level = 2
+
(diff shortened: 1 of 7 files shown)


TheQuantumPhysicist:
As discussed, this isn't necessary now.

PR #1270, opened 2023-10-11
blockprod: Ability to specify custom transactions

iljakuklic:
This gives the user the option to add a custom list of transactions when asking the node to generate a new block.

Based on the original work of @alfiedotwtf.

Related:
* #756 
* #1158

### DIFF
--- a/blockprod/src/detail/mod.rs
+++ b/blockprod/src/detail/mod.rs
@@ -30,9 +30,9 @@ use common::{
             block_body::BlockBody, signed_block_header::SignedBlockHeader,
             timestamp::BlockTimestamp, BlockCreationError, BlockHeader, BlockReward, ConsensusData,
         },
-        Block, ChainConfig, GenBlock, SignedTransaction,
+        Block, ChainConfig, GenBlock, SignedTransaction, Transaction,
     },
-    primitives::{BlockHeight, Id},
+    primitives::{Amount, BlockHeight, Id, Idable},
     time_getter::TimeGetter,
 };
 use consensus::{
@@ -42,7 +42,7 @@ use consensus::{
 use crypto::random::{make_true_rng, Rng};
 use logging::log;
 use mempool::{
-    tx_accumulator::{DefaultTxAccumulator, TransactionAccumulator},
+    tx_accumulator::{DefaultTxAccumulator, PackingStrategy, TransactionAccumulator},
     MempoolHandle,
 };
 use p2p::P2pHandle;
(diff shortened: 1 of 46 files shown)


blockprod/src/detail/mod.rs
```diff
@@ -435,36 +458,27 @@ impl BlockProduction {
...
-                        }
-                    }
-                }
-                TransactionsSource::Provided(txs) => txs,
+            // We conservatively use the minimum timestamp here in order to figure out
+            // which transactions are valid for the block.
+            // TODO: Alternatively, we can construct the transaction sequence from
+            // scratch every time a different timestamp is attempted. That is more costly
+            // in terms of computational resources but will allow the node to include more
+            // transactions since the passing time may release some time locks.
```
alfiedotwtf, line 466: Going through the above code and then this change, I'm thinking that the minimum here should be `MIN(last_timestamp_seconds_used, min_constructed_block_timestamp)` given that transactions may be earlier than `min_constructed_block_timestamp`.

`min_contstructed_block_timestamp` should probably by locally scoped to the construction of  `max_constructed_block_timestamp` as it's only used to calculate that value.

mempool/src/pool/collet_txs.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+pub fn collect_txs<M>(
+    mempool: &Mempool<M>,
+    mut tx_accumulator: Box<dyn TransactionAccumulator>,
+    transaction_ids: Vec<Id<Transaction>>,
+    packing_strategy: PackingStrategy,
+) -> Result<Box<dyn TransactionAccumulator>, BlockConstructionError> {
+    let mempool_tip = mempool.best_block_id();
+    let block_timestamp = tx_accumulator.block_timestamp();
```
alfiedotwtf, line 76: This could probably be consolidated with the `verifier_time` below

mempool/src/pool/collet_txs.rs
```diff
@@ -0,0 +1,247 @@
...
+        .iter()
+        .map(|transaction| {
+            let _fee =
+                tx_verifier.connect_transaction(&tx_source, transaction, &verifier_time, None)?;
+            Ok(transaction.transaction().get_id())
+        })
+        .collect::<Result<Vec<_>, TxValidationError>>()?;
+
+    // Set of transactions already placed into the accumulator
+    let mut emitted: BTreeSet<_> = accum_ids.iter().collect();
```
alfiedotwtf, line 115: I know this was existing code, but since we're here...`tx_accumulater.transactions() is iterated on and then collected for `accum_ids`, and then `accum_ids` is iterated on to collect for `emitted`. Could `emitted` be build within the `accum_ids` `.map())` with a `emmited.push()`?

mempool/src/pool/collet_txs.rs
```diff
@@ -0,0 +1,247 @@
...
+    // Set of already processed transactions, for de-duplication
+    let mut processed = emitted.clone();
```
alfiedotwtf, line 117: I guess for here too.

alfiedotwtf (approved):

mempool/src/pool/collet_txs.rs
iljakuklic: Yea I guess it's possible to do all this in a single pass but the proposed approach with `push` feels somewhat less declarative. Not that it makes a lot of difference, I don't feel strongly either way.

alfiedotwtf: All good. This PR is good to go anyway, and we go back to this code then we can update in in a future PR.

iljakuklic:
Apparently, the mempool sync issue resurfaced also in functional tests, need to fix it up.

blockprod/src/detail/mod.rs
```diff
@@ -180,19 +180,31 @@ impl BlockProduction {
...
+            current_tip,
+            min_block_timestamp,
+        ));
+
+        for transaction in transactions.into_iter() {
+            let transaction_id = transaction.transaction().get_id();
+
+            accumulator
+                .add_tx(transaction, Amount::ZERO.into())
+                .map_err(|err| BlockProductionError::FailedToAddTransaction(transaction_id, err))?
```
TheQuantumPhysicist, line 198: How are we making sure that these raw transactions are valid against what's coming from the mempool? I do hope we're not depending on the temporary solution we deployed that we wanted to solve later, where we include a transaction verifier inside the accumulator.

mempool/src/pool/collet_txs.rs
TheQuantumPhysicist, line 1: The file name has a typo

TheQuantumPhysicist:
I took a round and it looks fine. I just don't understand the consistency guarantees we're having on transactions. Besides that and the comments, looking good.

blockprod/src/detail/mod.rs
iljakuklic: There is no guarantee currently. Even transaction verifier is not sufficient because mempool deps are used to figure out the parent/child relationships. So it only properly works for transactions that are "independent" of transactions in the mempool. Could be removed since the guarantees are much weaker than for transactions specified by IDs.

iljakuklic: > I thought we agreed to just remove it...

Ok will do.

iljakuklic: Hmm there are many existing functional tests that rely on this functionality, and some are not easy to adapt so the transactions can be pushed through mempool.

TheQuantumPhysicist: Let's just keep it, and write in the documentation that adding transactions there relies on the user ensuring the validity, and is done only in testing.

TheQuantumPhysicist (approved):

PR #1276, opened 2023-10-16
mempool: Fix tx accumulator block size calculation

iljakuklic:
The encoded sequence of transactions starts with the number of transactions. However, the transaction accumulator failed to take that into account, only considering the space taken up by raw transaction data without the transaction count annotation. This is a fix.

### DIFF
--- a/mempool/src/tx_accumulator.rs
+++ b/mempool/src/tx_accumulator.rs
@@ -17,7 +17,7 @@ use common::{
     chain::{block::timestamp::BlockTimestamp, GenBlock, SignedTransaction},
     primitives::{Amount, Id},
 };
-use serialization::Encode;
+use serialization::{Compact, Encode};
 
 use crate::pool::fee::Fee;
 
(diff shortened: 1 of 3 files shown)


mempool/src/tx_accumulator.rs
```diff
@@ -65,35 +65,55 @@ impl DefaultTxAccumulator {
...
+            // Sanity check that total_size_with() and total_size() agree
+            assert_eq!(total_size_with_tx, self.total_size());
         } else {
-            self.done = true
+            self.done = true;
         };
+
+        // Sanity check the size tracking is accurate
+        #[cfg(test)]
+        assert_eq!(self.total_size(), self.transactions().encoded_size());
```
TheQuantumPhysicist, line 107: I would change this to a log message starting with CRITICAL. The last thing we want is an off-by-one error crashing the node.

TheQuantumPhysicist (approved):
Besides the little comment, looking good.

mempool/src/tx_accumulator.rs
TheQuantumPhysicist: Never mind. I just noticed the cfg(test). Cheers!

alfiedotwtf (approved):

PR #1281, opened 2023-10-18
Rework NetUpgrades

azarovh:
So I changed the structure of netupgrade to `Vec<(BlockHeight, (NetUpgradeVersion, ConsensusUpgrade))>`

`NetUpgradeVersion` is an enum of consecutive forks that we introduced to our network. For every version there is `ConsensusUpgrade` which contains consensus parameters for the version.

The thing that bothers me with current design is that on one hand it's made to be restrictive (the versions must be consecutive and not repeated) but on the other it's hard to apply this to the tests where we want to have different kind of upgrades that don't represent production forks. Potentially it can be solved by providing separate initialisation function for the tests that has less strict rules. But anyway wanted a feedback on this part.


common/src/chain/upgrades/netupgrade.rs
```diff
@@ -15,43 +15,35 @@
...
+use crate::Uint256;
+use crate::{
+    chain::{config::ChainType, pow::limit},
+    primitives::BlockHeight,
+};
+
+use super::{ConsensusUpgrade, NetUpgradeVersion, PoSStatus, PoWStatus, RequiredConsensus};
 
 #[derive(Debug, Clone)]
 pub struct NetUpgrades<T>(Vec<(BlockHeight, T)>);
```
TheQuantumPhysicist, line 34: Everything seems to be OK, but the only problem left is that this generic T shouldn't be part of the upgrades.

The data to be used on the upgrades should be available in the module doing the checking. Another option is to also put it in ChainConfig, which is what bitcoin does, btw. The problem here is that there's no generic solution that can contain all possible upgrades with their parametrization. This parametrization doesn't scale. The only reason this works now is that we only have consensus upgrades. In the future, if we have more, it won't fit in here.

TheQuantumPhysicist:
We're closing this for now. There doesn't seem to be a viable soution that supports both having monotonic upgrades AND supports testing different modes of operation.

PR #1283, opened 2023-10-19
Only ascii alphanumeric strings in tokens after v1

azarovh:


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -740,7 +740,11 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         Ok(())
     }
 
-    fn check_tokens_txs(&self, block: &Block) -> Result<(), CheckBlockTransactionsError> {
+    fn check_tokens_txs(
+        &self,
+        block: &Block,
+        block_height: BlockHeight,
```
TheQuantumPhysicist, line 746: Please load the block height in here. I think this is a reasonable proposition because later, after the fork, it'll be easy to remove it.

chainstate/src/detail/tokens/check_utils.rs
```diff
@@ -13,11 +13,21 @@
...
+    match String::from_utf8(str.to_vec()) {
+        Ok(text) => text.chars().all(char::is_alphanumeric),
+        Err(_) => false,
+    }
+}
+
+fn check_is_text_ascii_alphanumeric(str: &[u8]) -> bool {
     match String::from_utf8(str.to_vec()) {
         Ok(text) => text.chars().all(|c| c.is_ascii_alphanumeric()),
         Err(_) => false,
```
iljakuklic, line 33: Two points here:
* Can use `Result::is_ok_and` but that has been stabilised fairly recently, not sure we want to require such a recent compiler version
* `is_ascii_alphanumeric` may be somewhat restrictive for description for exmaple. It would reject anything with a space or usual punctuation `.,;:!?` in it. I think ticker and name need a separate set of valid characters different from description.

iljakuklic: Btw there's `std::str::from_utf8` that lets you convert `&[u8]` to `&str` without allocating a `String` in a validating way.

test-utils/src/lib.rs
```diff
@@ -158,4 +169,14 @@ mod tests {
...
     }
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn random_string_test(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+
+        let result = random_string(&mut rng, 1..100);
+        assert!(result.chars().all(|c| c.is_ascii_alphanumeric()));
```
ImplOfAnImpl, line 180: It's better to rename `random_string` to suggest that it can only produce alphanum characters

ImplOfAnImpl (approved):

chainstate/src/detail/tokens/check_utils.rs
azarovh: > is_ascii_alphanumeric may be somewhat restrictive for description for exmaple. 

that's by design to avoid consequences that we don't understand at the moment. Can easily be changed later

> Btw there's std::str::from_utf8

nice thanks

chainstate/src/detail/chainstateref/mod.rs
azarovh: done

chainstate/src/detail/tokens/check_utils.rs
azarovh: also simplified `check_is_text_ascii_alphanumeric`

iljakuklic: > Can easily be changed later

Is it not a part of consensus rules?

TheQuantumPhysicist:
Please don't merge this until I slap another approval. I still have to review it again. I have some doubts.

TheQuantumPhysicist (changes requested):
Just to ensure it won't be merged by mistake.

chainstate/src/detail/tokens/check_utils.rs
azarovh: I mean with a fork

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -743,6 +743,15 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
     }
 
     fn check_tokens_txs(&self, block: &Block) -> Result<(), CheckBlockTransactionsError> {
+        let prev_block_id = block.prev_block_id();
+        let current_height = self
+            .get_gen_block_index(&prev_block_id)?
+            .ok_or(CheckBlockTransactionsError::PropertyQueryError(
+                PropertyQueryError::BlockIndexNotFound(prev_block_id),
```
TheQuantumPhysicist, line 750: Better use the error PrevBlockIndexNotFound. Also, this is not a non-chalant error... this is an invariant broken. 

Why not use the function that was made for this anyway? `get_previous_block_index_for_check_block()`

test-utils/src/lib.rs
```diff
@@ -39,7 +39,21 @@ pub fn decode_from_hex<D: serialization::DecodeAll>(to_decode: &str) -> D {
...
 
-pub fn random_string<R: SampleRange<usize>>(rng: &mut impl Rng, range_len: R) -> String {
+pub fn get_random_non_ascii_alphanumeric_byte(rng: &mut impl Rng) -> u8 {
+    for _ in 0..1000 {
+        let random_byte = rng.gen::<u8>();
+        if !random_byte.is_ascii_alphanumeric() {
+            return random_byte;
+        }
+    }
+    // it's approximately 0.5^1000 that this panics
```
TheQuantumPhysicist, line 49: It's actually much less... because half of u8 range is printable characters, not necessarily alphanumeric.

TheQuantumPhysicist (approved):
Besides that minor thing, I think this is fine.

chainstate/src/detail/chainstateref/mod.rs
azarovh: > Better use the error PrevBlockIndexNotFound. 

Agree, done.

> Why not use the function that was made for this anyway?

It returns incompatible Error type. It doesn't do much so I think it's ok to leave it like this.

muursh (approved):

PR #1288, opened 2023-10-20
End-to-end encryption for staking key passing to block production

### LINKED ISSUE #801: Wallet handing staking key to block production should be end-to-end encrypted
The initial plan for testnet is to hand the private key without encryption from the wallet to the block production (on every block). This isn't a huge risk on a single machine unless someone is sniffing connections with something like wireshark, though it's good practice to never hand keys with clear connection. Also, we have to be ready for dumb people who will do dumb things and pass the wallet connection directly over network with no ssh tunnel or encryption of any kind. Better be ready for that.

Consider using this library to create the facility of end-to-end encryption shared key with Diffie-Hellman protocol: https://crates.io/crates/x25519-dalek

Then use our Chacha20-Poly1305 encryption tools that we already wrote in the crypto crate.

TheQuantumPhysicist:
Closes #801

### DIFF
--- a/crypto/src/ephemeral_e2e/mod.rs
+++ b/crypto/src/ephemeral_e2e/mod.rs
@@ -167,7 +167,7 @@ mod tests {
     #[rstest]
     #[trace]
     #[case(Seed::from_entropy())]
-    fn basic(#[case] seed: Seed) {
+    fn shared_secret_then_encrypt_decrypt(#[case] seed: Seed) {
         let mut rng = make_seedable_rng(seed);
         let private_key1 = EndToEndPrivateKey::new_from_rng(&mut rng);
         let public_key1 = private_key1.public_key();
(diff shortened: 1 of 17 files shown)


OBorce (approved):

crypto/src/ephemeral_e2e/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+impl SharedSecret {
+    pub fn encode_then_encrypt<T: Encode, R: Rng + CryptoRng>(
+        &self,
+        obj: &T,
+        rng: &mut R,
+    ) -> Result<Vec<u8>, Error> {
+        let encoded = obj.encode();
+        self.encrypt(&encoded, rng)
```
iljakuklic, line 116: With `using_encoded`, you can, in certain cases, avoid the extra `Vec`.

iljakuklic (approved):
Looks good to me. Although I cannot say I've fully scrutinised all the subtleties of the crypto stuff.

muursh (approved):

PR #1289, opened 2023-10-20
p2p: Improve peer clock diff detection

iljakuklic:
The clock diff detection now takes into account not only when the handshake message was sent but also when a reply was received. That way, we get a more precise idea regarding peer's time and can act accordingly.

Not exactly sure how to unit-test this, it needs a way to simulate a delay in peer's reply. Should be doable from functional tests but sadly don't have the time to do it now.


p2p/src/config.rs
```diff
@@ -29,7 +29,7 @@ make_config_setting!(NodeTypeSetting, NodeType, NodeType::Full);
 make_config_setting!(AllowDiscoverPrivateIps, bool, false);
 make_config_setting!(PingCheckPeriod, Duration, Duration::from_secs(60));
 make_config_setting!(PingTimeout, Duration, Duration::from_secs(150));
-make_config_setting!(MaxClockDiff, Duration, Duration::from_secs(10));
+make_config_setting!(MaxClockDiff, Duration, Duration::from_secs(12));
```
iljakuklic, line 32: I bumped the value here a little since the new approach is slightly less forgiving. Now both endpoints of local time interval in which the request is being processed must adhere to the max limit. It only checked one end of the interval before.

TheQuantumPhysicist:
The code is fine, but the algorithm... I'm not sure this is what we would use to minimize the damage from latency.

As far as I understand, the issue previously was that we're taking the handshake timestamp starting point, call it A, and learning the difference in time (with the max diff tolerance) by reading the timestamp the peer sent to us, call it B. Now |A-B| does not account for network latency, and hence it contains the network latency as an error in the measurement. So there's no way to know the relation between A and the true point in time the peer recorded timestamp B.

My understanding of the improvement solution would be to register local time at two points, once when we send the handshake and when we receive the response. Say the two points are A1 and A2. Now assuming the bidirectional latency is exactly equal (which is not absolutely true in real life, but again, probably a good approximation), and that computation time is negligible, we can assume that sending the handshake and receiving a response takes the same time, then the best approximation of the time instant, at which the peer recorded the timestamp and responded, would be the average of A1 and A2, hence (A1+A2)/2. In other words, this approximation makes the timestamp at which the peer recorded B, be equivalent to (A1+A2)/2, as much as possible, by minimizing network latency effects.

However, what I'm seeing in the code currently doesn't attempt to minimize the error using that logic. It tries, on the other hand, to be more strict in ensuring that timestamp of the peer is not too far from the time recorded when receiving the message. I may be wrong, but this does not really contribute to minimizing the error caused by network latency.

Please let me know if I'm misunderstanding something. Perhaps there's another angle I'm not seeing.

TheQuantumPhysicist:
I pushed a commit with my solution based on my explanation. No pressure. It can also be considered just for preview. Let's discuss it.

iljakuklic:
> The code is fine, but the algorithm... I'm not sure this is what we would use to minimize the damage from latency.
> 
> As far as I understand, the issue previously was that we're taking the handshake timestamp starting point, call it A, and learning the difference in time (with the max diff tolerance) by reading the timestamp the peer sent to us, call it B. Now |A-B| does not account for network latency, and hence it contains the network latency as an error in the measurement. So there's no way to know the relation between A and the true point in time the peer recorded timestamp B.
> 
> My understanding of the improvement solution would be to register local time at two points, once when we send the handshake and when we receive the response. Say the two points are A1 and A2. Now assuming the bidirectional latency is exactly equal (which is not absolutely true in real life, but again, probably a good approximation), and that computation time is negligible, we can assume that sending the handshake and receiving a response takes the same time, then the best approximation of the time instant, at which the peer recorded the timestamp and responded, would be the average of A1 and A2, hence (A1+A2)/2. In other words, this approximation makes the timestamp at which the peer recorded B, be equivalent to (A1+A2)/2, as much as possible, by minimizing network latency effects.
> 
> However, what I'm seeing in the code currently doesn't attempt to minimize the error using that logic. It tries, on the other hand, to be more strict in ensuring that timestamp of the peer is not too far from the time recorded when receiving the message. I may be wrong, but this does not really contribute to minimizing the error caused by network latency.
> 
> Please let me know if I'm misunderstanding something. Perhaps there's another angle I'm not seeing.

The idea was to take the latency into account in a bit more robust way, not relying on the assumption that the peer sampled the time in the middle of the time interval between sending the handshake message and receiving the reply. That way, assuming the peer is not making up the timestamp, we know that the peer is max 10s off (given 10s time offset limit).

When treating the midpoint as the peer's sample time, we can get out of this range. E.g. if the peer takes 4s to reply, we remain connected to the peer even if it may be up to 12s off with the 10s max time offset setting. So if nothing else, it makes the semantics of the max time offset setting make more sense.

I'm not saying my approach is better, there may be facets of the problem I'm missing. However, I think the response time probably should be taken into account somehow rather than just assuming the turnaround times are symmetric.

iljakuklic:
So I managed to find a way to test the timestamp negotiation, including networking delays (response timeout) albeit not particularly elegant.

p2p/Cargo.toml
```diff
@@ -56,6 +56,7 @@ p2p-backend-test-suite = { path = "backend-test-suite" }
 p2p-test-utils = { path = "p2p-test-utils" }
 storage-inmemory = { path = "../storage/inmemory" }
 test-utils = { path = "../test-utils" }
+tokio = { workspace = true, default-features = false, features = ["io-util", "macros", "net", "rt", "rt-multi-thread", "sync", "time", "test-util"] }
```
TheQuantumPhysicist, line 59: Well, at this point... why not just "full"?

p2p/src/net/default_backend/peer.rs
```diff
@@ -121,28 +125,48 @@ where
...
-            std::cmp::max(local_time, remote_time) - std::cmp::min(local_time, remote_time);
+
+        // We cannot know the time of the remote peer at the moment of recording the timestamp,
+        // hence, we do the following: We assume that the remote peer's time is somewhere in the middle,
+        // between the time when the message was sent and the time when the message was received.
+        // This is a reasonable assumption that requires:
+        // 1. The latency in sending and receiving the message is the same
+        // 2. The time required to do computations on the message to create the response is negligible
+        // The formula for the local time at the moment the remote time was recorded is:
+        //     remote_time = local_time_start + (local_time_end - local_time_start) / 2
```
TheQuantumPhysicist, line 149: This is outdated, right?

iljakuklic: Correct, will fix

p2p/src/config.rs
TheQuantumPhysicist: Do we want to make this smaller now?

p2p/Cargo.toml
iljakuklic: I'm not sure.
* The `full` does not include `test-util` anyway, so it would still have to be listed explicitly.
* It enables some features we don't use: some extra IO stuff, `parking_lot`, support for spawning processes...

I do get the appeal of brevity but still slightly prefer the explicit list.

p2p/src/net/default_backend/peer.rs
```diff
@@ -139,14 +139,15 @@ where
...
 
-        // We cannot know the time of the remote peer at the moment of recording the timestamp,
-        // hence, we do the following: We assume that the remote peer's time is somewhere in the middle,
-        // between the time when the message was sent and the time when the message was received.
-        // This is a reasonable assumption that requires:
-        // 1. The latency in sending and receiving the message is the same
-        // 2. The time required to do computations on the message to create the response is negligible
-        // The formula for the local time at the moment the remote time was recorded is:
-        //     remote_time = local_time_start + (local_time_end - local_time_start) / 2
+        // We do not know at what point point exactly the peer recorded its timestamp. However, we
```
TheQuantumPhysicist, line 142: I was looking for a typo. Wasn't disappointed 😄

iljakuklic: > I was looking for a typo. Wasn't disappointed 😄

Heh thanks. Always happy to include a handful of them :laughing:

p2p/src/config.rs
iljakuklic: Hmm I thought I reverted it. Must be a merge issue

iljakuklic: Reverted to 10s and removed some stuff that was mistakenly introduced by a merge. Do we want to go below 10s?

p2p/src/net/default_backend/peer.rs
```diff
@@ -121,28 +125,49 @@ where
...
             node_protocol_version,
+            time_getter,
         }
     }
 
     fn validate_peer_time(
         p2p_config: &P2pConfig,
-        local_time: Duration,
-        remote_time: Duration,
+        local_time: std::ops::RangeInclusive<Time>,
```
ImplOfAnImpl, line 134: IMO `local_time` that is a range looks rather cryptic. I'd suggest to make it two separate parameters with more descriptive names.

p2p/src/config.rs
TheQuantumPhysicist: That's an excellent question. I would go to 5 seconds, but I'd rather we do some research before making it this tight, and try to understand whether connections around the world can take this long under normal circumstances.

p2p/src/net/default_backend/peer.rs
```diff
@@ -121,28 +125,49 @@ where
...
-            std::cmp::max(local_time, remote_time) - std::cmp::min(local_time, remote_time);
+
+        // We do not know at what point exactly the peer recorded its timestamp. However, we do
+        // know it was somewhere between the request was initiated and the response was received.
+        // We give the peer some leeway when it comes to network latency so the acceptable time is
+        // in the interval [`init_time - tolerance`, `recv_time + tolerance`].
+        //
+        // Since the distance between `init_time` and `recv_time` is bounded by the handshake
+        // timeout, the effective max peer time offset is:
+        // ```tolerance + (recv_time - init_time) <= tolerance + handshake_timeout```
```
ImplOfAnImpl, line 149: I'm a bit confused by this. Did you mean `2*tolerance + (recv_time - init_time)`? Because that's the length of the interval that you've mentioned.

ImplOfAnImpl (approved):

p2p/src/config.rs
iljakuklic: > That's an excellent question. I would go to 5 seconds, but I'd rather we do some research before making it this tight, and try to understand whether connections around the world can take this long under normal circumstances.

The tolerance for connection latency is included in the handshake timeout so this value is the tolerance for clock differences _on top of_ latency. Tuning how strict our requirements are with respect to connection quality should be done by adjusting the handshake timeout parameter rather than this one. The effective tolerance is `handshake_timeout + max_clock_diff`.

Now this could be changed so `max_clock_diff` specifies the diff _including_ network latency, i.e. the effective tolerance mentioned in the previous paragraph. The interval of allowable peer time would be adjusted by `new_max_clock_diff - handshake_timeout`. Basically just shift the value by handshake timeout. There are tradeoffs:
* Pro: the semantics of `max_clock_diff` becomes clearer: it specifies by how much can peer's clock differ from ours.
* Con: Values smaller than handshake timeout no longer make sense. Would have to be handled somehow.
* Pro: When adjusting handshake timeout without changing effective latency does not involve changing the other parameter
* Con: When adjusting handshake timeout in a way where we want the max clock diff on top of latency remain constant requires also adjusting this parameter accordingly.

The last two bullets are the other way around in the current implementation. Seems there is no ideal solution here.

p2p/src/net/default_backend/peer.rs
iljakuklic: To summarise, the effective tolerances are:
* `tolerance + handshake_timeout` between our node and any one given peer
* `2 * tolerance + hendshake_timeout` between any two peer the node is connected to

Added the note of this to the comment, hopefully it's now clearer.

p2p/src/config.rs
TheQuantumPhysicist: We can put both variables in a struct and manage them together in a constructor to limit the cases where they don't make sense.

iljakuklic: > We can put both variables in a struct and manage them together in a constructor to limit the cases where they don't make sense.

That would be a nice implementation approach but there is still the question of which one to present to the user as the config option.

p2p/src/lib.rs
```diff
@@ -315,6 +316,37 @@ pub fn make_p2p<S: PeerDbStorage + 'static>(
...
+            // These are protocol-sensitive parameters and changing them has to be done with care.
+            // If the change has been sufficiently scrutinized an is deemed desirable, the values
+            // should be updated here as well.
+            //
+            // What to consider when changing these:
+            // * The `max_clock_diff` and `handshake_timeout` settings are related. The effective
+            //   max allowed difference between our clock and peer's clock is the sum of the two.
+            //   If one is changed, it should be considered how the other should be affected.
+            // * Peer may be stalled for the duration of up to `effective_max_clock_diff`.
+            //   This may interact with other timeouts.
```
iljakuklic, line 330: @TheQuantumPhysicist Anything to be added to the comment?

p2p/src/lib.rs
```diff
@@ -315,6 +316,48 @@ pub fn make_p2p<S: PeerDbStorage + 'static>(
     time_getter: TimeGetter,
     peerdb_storage: S,
 ) -> Result<P2pInit<S>> {
+    match chain_config.chain_type() {
+        ChainType::Mainnet | ChainType::Testnet | ChainType::Signet => {
+            // These are protocol-sensitive parameters and changing them has to be done with care.
+            // If the change has been sufficiently scrutinized an is deemed desirable, the values
```
TheQuantumPhysicist, line 322: and

p2p/src/lib.rs
TheQuantumPhysicist: I would just add that the implementation of handshake tolerance is in what struct/function as a soft reference to understand how this works. Otherwise, this looks good.

TheQuantumPhysicist (approved):
I'm good with this. Good job!

PR #1292, opened 2023-10-24
Tests for new peer discovery when the tip is stale

ImplOfAnImpl:
Making the tests work required some changes in production code as well:
1) I added `last_tip_block_time` to `PeerManager`, which is now used to force a call to `reload_dns_seed`.
Also, I had to add `PeerManagerEvent::NewLocalTip`, so that the peer manager could learn about locally-produced tips too. This in turn required to add `BlockSource` to `ChainstateEvent::NewTip`.
2) The logic from `reload_dns_seed` was moved to `DefaultDnsSeed::obtain_addresses`.

### DIFF
--- a/p2p/src/peer_manager/dns_seed.rs
+++ b/p2p/src/peer_manager/dns_seed.rs
@@ -1,0 +1,100 @@
+// Copyright (c) 2021-2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use common::chain::{config::ChainType, ChainConfig};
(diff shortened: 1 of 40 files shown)


p2p/src/peer_manager/dns_seed.rs
```diff
@@ -0,0 +1,100 @@
...
+const DNS_SEEDS_MAINNET: [&str; 0] = [];
+const DNS_SEEDS_TESTNET: [&str; 1] = ["testnet-seed.mintlayer.org"];
+
+/// Maximum number of records accepted in a single DNS server response
+const MAX_DNS_RECORDS: usize = 10;
+
+#[async_trait]
+impl DnsSeed for DefaultDnsSeed {
+    async fn obtain_addresses(&self) -> Vec<SocketAddress> {
+        let dns_seed = match self.chain_config.chain_type() {
```
ImplOfAnImpl, line 56: FYI: the logic in this function is not new; it was moved here from `reload_dns_seed`

p2p/src/peer_manager/mod.rs
```diff
@@ -1588,6 +1587,27 @@ where
 pub trait Observer {
     fn on_peer_ban_score_adjustment(&mut self, address: SocketAddress, new_score: u32);
     fn on_peer_ban(&mut self, address: BannableAddress);
+    // This will be called at the end of "heartbeat" function.
+    fn on_heartbeat(&mut self);
+    // This will be called for both incoming and outgoing connections.
+    fn on_connection_acccepted(&mut self, address: SocketAddress);
```
ImplOfAnImpl, line 1593: FYI: I ended up not needing these functions in the final implementation of the tests; but I decided to keep them (and the corresponding `PeerManagerNotification` variants), because they may become useful in the future.

TheQuantumPhysicist (approved):

PR #1293, opened 2023-10-24
Add initial Address balance endpoint support to the API Server

alfiedotwtf:
Part of #1189


api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
```diff
@@ -30,6 +30,13 @@ impl<'t> ApiServerStorageRead for ApiServerInMemoryStorageTransactionalRo<'t> {
         self.transaction.is_initialized()
     }
 
+    async fn get_address_balance(
+        &self,
+        address: &str,
+    ) -> Result<Option<Amount>, ApiServerStorageError> {
```
TheQuantumPhysicist, line 36: Please understand that the balance isn't just in coins, but also in tokens. So this should return a map of coins and tokens, similar to AmountsMap in TransactionVerifier. You don't have to do this right now, but it can be added to the list.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -67,17 +77,44 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
             db_tx.set_main_chain_block_id(block_height, block.get_id()).await?;
             logging::log::info!("Connected block: ({}, {})", block_height, block.get_id());
 
+            update_balances_from_outputs(
+                Arc::clone(&self.chain_config),
+                &mut db_tx,
+                block_height,
+                block.block_reward().outputs(),
+            )
+            .await?;
```
TheQuantumPhysicist, line 96: There are also inputs in the block header.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -91,3 +128,162 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+            | TxOutput::ProduceBlockFromStake(_, _) => {}
+            TxOutput::Transfer(output_value, destination)
+            | TxOutput::LockThenTransfer(output_value, destination, _) => {
+                match destination {
+                    Destination::PublicKey(_) | Destination::Address(_) => {
+                        let address = Address::<Destination>::new(&chain_config, destination)
+                            .map_err(|_| {
+                                ApiServerStorageError::DeserializationError(
+                                    "Unable to encode destination".to_string(),
+                                )
```
TheQuantumPhysicist, line 253: I think an expect() here is in order.

TheQuantumPhysicist (approved):
Besides the minor comment, I think this is fine.

Please wait until we finish the token work we're doing to merge. I don't want to cause conflicts.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -91,3 +128,162 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+                                    .get_address_balance(address.get())
+                                    .await?
+                                    .unwrap_or(Amount::ZERO);
+
+                                let new_amount = current_balance
+                                    .add(*amount)
+                                    .expect("Balance should not overflow");
+
+                                db_tx
+                                    .set_address_balance_at_height(
```
TheQuantumPhysicist, line 271: I've been thinking about this, and I think this is wrong. If you have one transaction that has multiple outputs paying to the same destination, one will overwrite the other, leading to an invalid balance.

The correct way is to create an object that you keep updating with everything, and then write that in the database after iterating over all outputs.

Please write a test that proves this before fixing it; so something like TDD to ensure the problem is fixed.

api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
alfiedotwtf: Yep, there's already a placeholder in the API endpoint for tokens, just not implemented yet. To lesson the complexity though, I was going to add token balances to a new function.

api-server/scanner-lib/src/blockchain_state/mod.rs
alfiedotwtf: Oh, ok thanks for the heads up.

api-server/scanner-lib/src/blockchain_state/mod.rs
alfiedotwtf: Ah you're right. Yep, TDD for this is good to prove it was incorrect to begin with.

alfiedotwtf: Interesting. I wrote the test and it passes without any fixes. Confused, I checked the logic and it seems to be correct - it loops over each output within the transaction, and gets the previous balance before updating to the new balance (i.e per output, not per transaction).

api-server/scanner-lib/src/blockchain_state/mod.rs
alfiedotwtf: I've just checked, and I can't find it. Did you mean the block reward here? If so, it's being taken care of with:

```
           update_balances_from_outputs(
                Arc::clone(&self.chain_config),
                &mut db_tx,
                block_height,
                block.**block_reward()**.outputs(),
            )
            .await
            .expect("Unable to update balances from block reward outputs");
```

api-server/scanner-lib/src/blockchain_state/mod.rs
alfiedotwtf: I've converted all `map_err()` an `await?` in `mod.rs` to `.expect()`.

TheQuantumPhysicist: I don't want to state the obvious, but obviously a blanket expect on everything isn't really what I meant. As long as you understand the circumstances and you find it OK, I'm fine with it.

api-server/scanner-lib/src/blockchain_state/mod.rs
TheQuantumPhysicist: in ConsensusData, for proof of stake, there are inputs.

api-server/stack-test-suite/tests/v1/address.rs
```diff
@@ -0,0 +1,474 @@
...
+        // response is made (by the web server, which takes the listener
+        // over)
+        let response = reqwest::get(format!("http://{}:{}{url}", addr.ip(), addr.port()))
+            .await
+            .unwrap();
+
+        assert_eq!(
+            response.status(),
+            200,
+            "Failed getting adddress balance for {address}"
```
TheQuantumPhysicist, line 254: adddress

api-server/stack-test-suite/tests/v1/address.rs
```diff
@@ -0,0 +1,474 @@
...
```
TheQuantumPhysicist, line 462: adddress

api-server/scanner-lib/src/blockchain_state/mod.rs
alfiedotwtf: I've just had a look, and I think PoS handling should go in the next PR for pools... the reward for PoS is `CreateStakePool|ProduceBlockFromStake` and both don't contain the output value. I haven't gotten there yet in the pool code, but I'm guessing the PoS reward is calculated on decommissioning.

TheQuantumPhysicist: Alrighty.

PR #1294, opened 2023-10-24
Make connection count limits modifiable in tests

ImplOfAnImpl:
The constants related to the number of connections were placed into a struct `ConnectionCountLimits` which is a part of `P2pConfig`.
This is a follow-up for the PR #1292. It's not strictly necessary, but it allows to make the tests in that PR less "heavy" by reducing the number of connections that those tests have to make.

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -43,7 +43,10 @@ use common::{
 };
 use crypto::random::{make_pseudo_rng, seq::IteratorRandom, Rng};
 use logging::log;
-use utils::{bloom_filters::rolling_bloom_filter::RollingBloomFilter, ensure, set_flag::SetFlag};
+use utils::{
+    bloom_filters::rolling_bloom_filter::RollingBloomFilter, ensure, make_config_setting,
+    set_flag::SetFlag,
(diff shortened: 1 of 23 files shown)


p2p/src/peer_manager/mod.rs
```diff
@@ -74,19 +77,44 @@ use self::{
...
 
-/// Desired number of full relay outbound connections.
-/// This value is constant because users should not change this.
-pub const OUTBOUND_FULL_RELAY_COUNT: usize = 8;
+#[derive(Default, Debug)]
+pub struct ConnectionCountLimits {
+    pub preserved_inbound_count_address_group: PreservedInboundCountAddressGroup,
+    pub preserved_inbound_count_ping: PreservedInboundCountPing,
+    pub preserved_inbound_count_new_blocks: PreservedInboundCountNewBlocks,
+    pub preserved_inbound_count_new_transactions: PreservedInboundCountNewTransactions,
```
ImplOfAnImpl, line 91: FYI: I ended up not needing to modify the inbound connection limits in the tests. But I decided to keep them here anyway, near their outbound counterparts, for completeness and because they may become useful in the future.

PR #1298, opened 2023-10-25
Add DataDeposit output type

TheQuantumPhysicist:
This output type is meant to deposit data on the blockchain, just like OP_RETURN in bitcoin. However, this we add a constant fee for that. The fee is configurable in ChainConfig.


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -742,6 +744,61 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         Ok(())
     }
 
+    fn check_data_deposit_outputs(&self, block: &Block) -> Result<(), CheckBlockTransactionsError> {
+        let prev_block_id = block.prev_block_id();
+        let current_height = self
+            .get_gen_block_index(&prev_block_id)?
```
azarovh, line 750: there is `get_previous_block_index`

muursh (approved):

azarovh (approved):
The only comment I have is that you also need to go through tests in `chainstate/tx-verifier/src/transaction_verifier/input_output_policy/tests/purpose_tests.rs` and place deposit output among invalid cases because we don't have compile time guard there and it has to be done manually.

azarovh:
Btw should we support this in the wallet?

chainstate/src/detail/chainstateref/mod.rs
TheQuantumPhysicist: I don't have the current block index here. Besides, this code will just go later.

TheQuantumPhysicist:
> Btw should we support this in the wallet?

Maybe... at some point. I'm not in a hurry for that.

azarovh (approved):

PR #1299, opened 2023-10-26
docker compose file for testing p2p v2 protocol

ImplOfAnImpl:
The `docker compose` file allows to run 15 nodes on the same machine simultaneously. The purpose is to be able to test the V2 protocol, so `PREFERRED_PROTOCOL_VERSION` was bumped up to V2 as well. I also increased the software version to 0.2.1 to be able to distinguish the new nodes from the old ones.

For nodes to find each other I specify them as "boot nodes" to each other. I decided not to uses reserved nodes, because it'd give them advantage over older nodes.

In addition to the docker files, there is a bunch of shell scripts to make working with the nodes easier, see `README.md` for details.

Some additional changes in the production code were also made:
1. The node now always does one dns reload early, see the `had_dns_reload` flag in peer manager and the corresponding comment. I'm not sure whether we need to merge this change into master (it's only useful when a fresh node is passed some boot nodes at startup and most of them belong to the same address group).
2. `ConnectedPeersJson` command was added to wallet-cli, which prints peers' info as json. This is used by one of the helper scripts.
3. Minor logging improvements.

P.S. This branch is not supposed to be merged into master, it's just for local testing.

### DIFF
--- a/build-tools/p2p-v2-test/docker-compose.yml
+++ b/build-tools/p2p-v2-test/docker-compose.yml
@@ -1,0 +1,229 @@
+version: "3.8"
+
+services:
+  node01:
+    build:
+      context: ../..
+      dockerfile: build-tools/p2p-v2-test/Dockerfile
+    container_name: $BASE_NAME-node01
+
+    volumes:
+      - data01:$DATA_DIR_IN_CONTAINER
+    ports:
+      - '40001:$RPC_PORT'
+    command: bash -c "$NODE_DAEMON_CMD"
+    environment:
+      - RUST_LOG
+
+  node02:
+    build:
+      context: ../..
+      dockerfile: build-tools/p2p-v2-test/Dockerfile
+    container_name: $BASE_NAME-node02
+
+    volumes:
+      - data02:$DATA_DIR_IN_CONTAINER
+    ports:
+      - '40002:$RPC_PORT'
+    command: bash -c "$NODE_DAEMON_CMD"
+    environment:
+      - RUST_LOG
+
+  node03:
+    build:
+      context: ../..
+      dockerfile: build-tools/p2p-v2-test/Dockerfile
+    container_name: $BASE_NAME-node03
+
+    volumes:
+      - data03:$DATA_DIR_IN_CONTAINER
+    ports:
+      - '40003:$RPC_PORT'
+    command: bash -c "$NODE_DAEMON_CMD"
+    environment:
+      - RUST_LOG
+
+  node04:
+    build:
+      context: ../..
+      dockerfile: build-tools/p2p-v2-test/Dockerfile
+    container_name: $BASE_NAME-node04
+
+    volumes:
+      - data04:$DATA_DIR_IN_CONTAINER
+    ports:
+      - '40004:$RPC_PORT'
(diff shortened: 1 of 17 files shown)


build-tools/p2p-v2-test/print_connected_peers.sh
TheQuantumPhysicist, line 1: Could you add a README.md explaining how this whole directory is to be used?

p2p/src/net/default_backend/default_networking_service.rs
```diff
@@ -41,7 +41,7 @@ use super::{backend::Backend, ConnectivityHandle, MessagingHandle, SyncingEventR
 // The preferred protocol version.
 // Note that we intentionally keep this constant private, because most of the code should
 // not depend on its value.
-const PREFERRED_PROTOCOL_VERSION: SupportedProtocolVersion = SupportedProtocolVersion::V1;
+const PREFERRED_PROTOCOL_VERSION: SupportedProtocolVersion = SupportedProtocolVersion::V2;
```
TheQuantumPhysicist, line 44: I would merge this PR... but upgrading this to V2 without the fix can be an issue.

build-tools/p2p-v2-test/docker-compose.yml
```diff
@@ -0,0 +1,199 @@
...
+
+services:
+  node01:
+    build:
+      context: ../..
+      dockerfile: build-tools/p2p-v2-test/Dockerfile
+    container_name: $BASE_NAME-node01
+
+    volumes:
+      - data01:$DATA_DIR_IN_CONTAINER
```
TheQuantumPhysicist, line 11: Instead of using these volumes and then going through hell to find the data directory, isn't it easier to make data01 a local data directory in the same directory of docker-compose, so with something like this:

```
    volumes:
      - data/node-data01:$DATA_DIR_IN_CONTAINER
```

This way, the data is accessible locally and the issue you mentioned in the readme file about the cookie wouldn't be really there. One only has to specify the port and data directory.

TheQuantumPhysicist:
I pushed a little commit.

build-tools/p2p-v2-test/docker-compose.yml
ImplOfAnImpl: You mean, create the directory inside the source tree? Well, I personally wouldn't like that. IMO it should be a separate directory, outside of the source tree.

Also, the custom directory would still be owned by root, so it's not a huge improvement.
It's possible to make the current user own it, I tried to do that initially. But it's extra hassle and AFAIR you have to pre-create the mount points before running `docker compose`, so that they're owned by the current user, which means you have to wrap `docker compose` in a bash script or something.

Cargo.toml
```diff
@@ -5,7 +5,7 @@ homepage = "https://mintlayer.org"
 repository = "https://github.com/mintlayer/mintlayer-core"
 readme = "README.md"
 license = "MIT"
-version = "0.1.2"
+version = "0.2.1"
```
TheQuantumPhysicist, line 8: Please revert this and the other one. We take care of bumping versions elsewhere.

TheQuantumPhysicist (approved):

PR #1300, opened 2023-10-26
Fix Postgres queries

alfiedotwtf:
No need to merge, will be merged as part of #1293...pushing to a branch here so that I'm not blocking Sergey.


api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -203,6 +277,16 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
         )
         .await?;
 
+        self.just_execute(
+            "CREATE TABLE ml_account_balance (
```
anyxem, line 281: In other places table name is `ml_address_balance`, I changed this table name to `ml_address_balance`

PR #1301, opened 2023-10-26
Move tokens fee check to ConstrainedValueAccumulator

azarovh:
Previous code was incorrect because it performed 2 independent checks instead of a cumulative one. So one could've issued a token and minted some with just one fee payment.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
@@ -16,7 +16,10 @@
 use std::collections::BTreeMap;
 
 use common::{
-    chain::{timelock::OutputTimeLock, AccountOp, ChainConfig, PoolId, TxInput, TxOutput},
+    chain::{
+        output_value::OutputValue, timelock::OutputTimeLock, tokens::TokenData, AccountOp,
+        ChainConfig, PoolId, TxInput, TxOutput,
+    },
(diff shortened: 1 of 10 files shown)


chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -33,13 +36,15 @@ use super::IOPolicyError;
 pub struct ConstrainedValueAccumulator {
     unconstrained_value: Amount,
     timelock_constrained: BTreeMap<BlockDistance, Amount>,
+    token_fee_burn_constrained: Amount,
```
TheQuantumPhysicist, line 39: This is exactly what I was worried about when I said "this is hard". 

This is not correct.

If I want to put this in mathematical terms, the dot-product of this variable and "unconstrained_value" is not zero. Meaning: You can represent one of them with the other by multiplying by a scalar (-1).

The correct solution for this problem should not involve adding any members here. The local variable `total_burned` should accumulate all the burned values. Then, by the end of the function you should subtract `total_burned` from `unconstrained_value`. This way, before the function call and after the function call, the state is valid and only orthogonal values exist in the struct, and this struct will still form a group, guaranteeing the properties we need without too many invasive tests.

And just a reminder: `timelock_constrained` can be made perfect (mathematically speaking) by changing BlockDistance to NonZeroBlockDistance. Currently, the two variables in the struct are not unrelated. Currently we're manually upholding the invariant that `timelock_constrained[0] == 0`, which is why what we're doing so far is valid.

PR #1305, opened 2023-10-26
Remove the transaction index and all its traces

### LINKED ISSUE #1274: Check whether we need the transaction index at all in the node
There's an open question on whether we really need the transaction index. It does complicate the logic in the transaction verifier. In bitcoin, there's a transaction index, accessible with `-txindex` command line argument. Do we need it for us?

TheQuantumPhysicist:
Based on the research we done, all the features of the transaction index can be done through the API server. Transactions that are interesting to the user should be taken from the wallet. PR #1304 should take care of that. 

Closes #1274


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -68,7 +68,7 @@ pub use epoch_seal::EpochSealError;
 
 pub struct ChainstateRef<'a, S, V> {
     chain_config: &'a ChainConfig,
-    chainstate_config: &'a ChainstateConfig,
+    _chainstate_config: &'a ChainstateConfig,
```
azarovh, line 71: Do you keep it for the future?

chainstate/test-suite/src/tests/tx_verification_simulation.rs
```diff
@@ -19,20 +19,14 @@ use common::chain::{tokens::TokenIssuanceVersion, ChainstateUpgrade, Destination
...
+#[case(Seed::from_entropy(), 20, 50)]
 #[trace]
-#[case(Seed::from_entropy(), 20, 50, true)]
-fn simulation(
-    #[case] seed: Seed,
-    #[case] max_blocks: usize,
-    #[case] max_tx_per_block: usize,
-    #[case] tx_index_enabled: bool,
-) {
+#[case(Seed::from_entropy(), 20, 50)]
```
azarovh, line 24: now that the cases are identical one can be removed

azarovh (approved):
nothing better than removing 3k lines of code

chainstate/src/detail/chainstateref/mod.rs
TheQuantumPhysicist: Correct.

chainstate/test-suite/src/tests/tx_verification_simulation.rs
TheQuantumPhysicist: Done.

azarovh (approved):
The only comment I have is that you also need to go through tests in `chainstate/tx-verifier/src/transaction_verifier/input_output_policy/tests/purpose_tests.rs` and place deposit output among invalid cases because we don't have compile time guard there and it has to be done manually.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -221,39 +213,6 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
-        };
-        log::trace!("Loading transaction with id: {}", id);
-        let position = match tx_index.position() {
-            common::chain::SpendablePosition::Transaction(pos) => pos,
-            common::chain::SpendablePosition::BlockReward(_) => {
-                panic!("In get_transaction(), a tx id led to a block reward")
-            }
-        };
-        Ok(self.db_tx.get_mainchain_tx_by_position(position)?)
-    }
```
alfiedotwtf, line 255: I'm not sure if this is going to be an issue or not, but here `get_transaction_in_block()` returns `SignedTransaction` whereas the API Server currently returns a simplified transaction:

```
Ok(Json(json!({
    "block_id": block_id.map_or("".to_string(), |b| b.to_hash().encode_hex::<String>()),
    "version_byte": transaction.version_byte(),
    "is_replaceable": transaction.is_replaceable(),
    "flags": transaction.flags(),
    "inputs": transaction.inputs(),
    "outputs": transaction.outputs(),
    })))
```

alfiedotwtf (approved):

chainstate/src/detail/chainstateref/mod.rs
TheQuantumPhysicist: I'm not sure I understand how that can be a problem. The API server doesn't use anything internal from ChainstateRef. Also the API server analyzes individual blocks separately, not using the anything exposed from the node other than individual blocks.

alfiedotwtf: No problem, this was more an FYI because you mentioned today to flag anything that you think the API Server can't replace the transaction index with. All good.

TheQuantumPhysicist: You can still including the signatures in the API server's analysis. That's all up to how it's implemented.

PR #1306, opened 2023-10-27
Fix some p2p issues

ImplOfAnImpl:
The following was done for both v1 and v2:

1) Fix potential header flooding by a malicious peer.
The node no longer allows new block headers to connect to `pending_headers` only.
Conversely, when sending new headers, v2 node no longer takes `best_sent_block_header` into account (and v1 didn't do this anyway).

2) Nodes no longer send block announcements if `outgoing.blocks_queue` is non-empty (because the peer will send a header request anyway when it's done downloading the blocks).

3) The check for a duplicate block request was only working for already downloaded blocks, but not for those which are inside `outgoing.blocks_queue`. I've added an additional check.

4) I renamed the flag `send_tip_updates` to `have_sent_all_headers`, because it's exactly what it means.

P.S. The fix for the failing test is in the separate PR #1309

Edit: PRs #1309 and #1316 were also merged into this one

### DIFF
--- a/p2p/src/sync/peer_v1.rs
+++ b/p2p/src/sync/peer_v1.rs
@@ -508,6 +508,16 @@ where
             ))?;
 
         // Check that all the blocks are known and haven't been already requested.
+        // First check self.outgoing.blocks_queue
+        let already_requested_blocks: BTreeSet<_> = self.outgoing.blocks_queue.iter().collect();
+        let doubly_requested_id = block_ids.iter().find(|id| already_requested_blocks.get(id).is_some());
+        if let Some(id) = doubly_requested_id {
+            return Err(P2pError::ProtocolError(
+                ProtocolError::DuplicatedBlockRequest(*id),
+            ));
+        }
+
+        // Then check the chainstate
         let ids = block_ids.clone();
         let best_sent_block = self.outgoing.best_sent_block.clone();
         self.chainstate_handle
--- a/p2p/src/sync/peer_v1.rs
+++ b/p2p/src/sync/peer_v1.rs
@@ -92,9 +92,10 @@ pub struct Peer<T: NetworkingService> {
     announced_transactions: BTreeSet<Id<Transaction>>,
     /// Current activity with the peer.
     peer_activity: PeerActivity,
-    /// If set, send the new tip notification when the tip moves.
-    /// It's set when we know that the peer knows about all of our current mainchain headers.
-    send_tip_updates: bool,
+    /// If this is set, it means that we've sent a HeaderList to the peer with the number
+    /// of headers less than the maximum. This is the signal to the peer that we have no more
+    /// headers, so it may not ask us for more of them in the future.
+    have_sent_all_headers: bool,
 }
 
 struct IncomingDataState {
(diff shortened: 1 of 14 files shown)


p2p/src/sync/peer_v1.rs
```diff
@@ -616,48 +612,44 @@ where
...
-        // the chainstate, pending_headers or requested_blocks).
+        // the chainstate or requested_blocks).
         let first_header_prev_id = *headers
             .first()
             // This is OK because of the `headers.is_empty()` check above.
             .expect("Headers shouldn't be empty")
             .prev_block_id();
 
+        // Note: we require a peer to send headers starting from a block that we already have
+        // in our chainstate or from one that we've already requested from the peer.
```
TheQuantumPhysicist, line 623: This "or from one that we've already requested from the peer" concerns me. Chainstate is the only source of truth.

p2p/src/sync/peer_v1.rs
```diff
@@ -616,48 +612,44 @@ where
...
+        // I.e. peers shouldn't track what block headers they've sent us already and use
+        // the last header (best_sent_block_header) as a starting point for future HeaderList
+        // updates.
+        // This restriction is needed to prevent malicious peers from flooding the node with
+        // headers, potentially exhausting the node's memory.
+        // The downside of this is that the peer may have to send the same headers multiple times.
+        // So, to avoid extra traffic, an honest peer should't send header updates when the node
```
TheQuantumPhysicist, line 630: There are tons of other ways to flood the node with traffic. This is a standard we can never attain.

p2p/src/sync/peer_v1.rs
```diff
@@ -616,48 +612,44 @@ where
...
+        // is already downloading blocks. But still, the node shouldn't punish the peer for
+        // doing so, because it's possible for it to do so on accident, e.g. a "new tip" event
```
TheQuantumPhysicist, line 632: by accident

p2p/src/sync/peer_v1.rs
```diff
@@ -616,48 +612,44 @@ where
...
-
         let first_header_is_connected_to_requested_blocks = first_header_prev_id
             .classify(&self.chain_config)
             .chain_block_id()
             .and_then(|id| self.incoming.requested_blocks.get(&id))
             .is_some();
 
         if !(first_header_is_connected_to_chainstate
-            || first_header_is_connected_to_pending_headers
             || first_header_is_connected_to_requested_blocks)
```
TheQuantumPhysicist, line 649: Please remove this. The only source of truth is chainstate.

TheQuantumPhysicist: Same applies to v2, btw.

p2p/src/sync/peer_v1.rs
ImplOfAnImpl: I've reformulated the comment a bit

p2p/src/sync/peer_v1.rs
ImplOfAnImpl: I've made this change in V1 too. We've never actually used this "optimization opportunity" when sending headers ourselves, so it shouldn't break anything.

p2p/src/sync/peer_v2.rs
```diff
@@ -719,6 +709,10 @@ where
         // The code below will set it again if needed.
         self.peer_activity.set_expecting_blocks_since(None);
 
+        // TODO: here we allow the peer to send blocks out of order. This means that we may receive
+        // an orphan block first and then its parent. Because of this, the NewTipReceived event
+        // generated below may not contain the actual tip. It doesn't seem like it may lead
+        // to problems at this moment, but it's still better to be more strict about it.
```
ImplOfAnImpl, line 715: FYI: This TODO was applied in PR #1316

TheQuantumPhysicist (approved):

PR #1307, opened 2023-10-27
Allow single unmint operation per token per tx

azarovh:
It's the result of fixing the crash encountered by random tests:
```
running 1 test
------------ TEST ARGUMENTS ------------
seed = Seed(1649375745970971719)
max_blocks = 20
max_tx_per_block = 50
tx_index_enabled = false
-------------- TEST START --------------
test tests::tx_verification_simulation::simulation::case_1 ... FAILED

failures:

failures:
    tests::tx_verification_simulation::simulation::case_1

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 243 filtered out; finished in 0.07s


--- STDERR:              chainstate-test-suite tests::tx_verification_simulation::simulation::case_1 ---
thread 'tests::tx_verification_simulation::simulation::case_1' panicked at 'called `Result::unwrap()` on an `Err` value: ProcessBlockError(StateUpdateFailed(TokensAccountingError(NotEnoughCirculatingSupplyToUnmint(Amount { val: 67489008815325409771966485925663979417 }, Amount { val: 102333752779816228051664371231704277194 }, Id<Token>{0xa4170880fecf3084b9b00fbb59a301c3d1691abc726f58214b58bd0138900a28}))))', chainstate\test-suite\src\tests\tx_verification_simulation.rs:63:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

In case multiple Unmint inputs are present the amount is deducted from circulating supply multiple times, which is incorrect because actual amount that should be unminted equals to total burned outputs. I suggest just forbidding such transactions.

P.S. also it contains trivial fix for the `wallet::change_token_supply_fixed` test


TheQuantumPhysicist (approved):
I really wonder whether we should just make a generic rule about AccountOps to be unique. Spending from delegation perhaps is the exception to the rules.

Maybe a better way is to create a filter that will check that automatically and will enforce (at the enum level) whether one or more is allowed. This fine-tuned solution is scary and is prone to errors.

I will be merging this now, but this is still a problem we have to think about.

azarovh:
> I really wonder whether we should just make a generic rule about AccountOps to be unique. 

I also tried an approach where we allow multiple Unmint ops but only 1 takes an effect. But it may be confusing and also has side-effect with how nonce must be handled. So I decided to just forbid it.

> Spending from delegation perhaps is the exception to the rules.

Minting also doesn't have a problem. 

> Maybe a better way is to create a filter

But io policies are that filter. They are pretty straight-forward. Not sure how it can be handle on enum level when inputs are stored in a vec.

PR #1309, opened 2023-10-27
Attempt to fix spuriously failing test "dont_make_announcements_while_blocks_are_being_sent"

ImplOfAnImpl:
The test `dont_make_announcements_while_blocks_are_being_sent` was added in PR #1306; it was failing spuriously on my local machine from time to time (depending on the CPU load), which forced me to increase the number of initial blocks from 10 to 100.  But then it failed on CI anyway.

The behaviour that it checks is basically a small optimization, so I was seriously considering just removing the test. However I decided to try to make it more reliable. I tries several approaches; the final "solution" (which I'm not particularly fond of, to be honest) was to make `MessagingService::send_message` async, so that test implementations of it could use a bounded channel. The problematic test now uses the channel capacity of 1, and can use `sleep` to slow down the rate at which BlockResponse messages are produced (their rate is what was causing the problem). Still, it doesn't guarantee that the spurious failures won't happen again, it just makes it less likely (hopefully much less).

On the other hand, `MessagingService::send_message` being async will allow us to use a bounded channel for sync messages in production code too, which may be a good thing, see the Note/TODO in `p2p/src/net/default_backend/mod.rs`

### DIFF
--- a/p2p/src/net/default_backend/mod.rs
+++ b/p2p/src/net/default_backend/mod.rs
@@ -142,8 +142,16 @@ where
     }
 }
 
+#[async_trait]
 impl MessagingService for MessagingHandle {
-    fn send_message(&mut self, peer_id: PeerId, message: SyncMessage) -> crate::Result<()> {
+    async fn send_message(&mut self, peer_id: PeerId, message: SyncMessage) -> crate::Result<()> {
+        // Note: send_message was made async to be able to use a bounded channel in tests.
+        // TODO: it may make sense to use a bounded channel in production as well. One possible
+        // reason for this might be to limit the number of BlockResponse messages that can exist
+        // at the same time. E.g. currently sync::Peer will produce a BlockResponse for each
+        // requested block almost immediately. The maximum number of simultaneously requested
+        // blocks is 500 per peer, a block can be up to 1Mb in size, so several malicious peers
+        // can eat up several Gb of RAM on the node.
         Ok(self.command_sender.send(types::Command::SendMessage {
             peer_id,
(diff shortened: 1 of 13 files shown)


p2p/src/net/default_backend/mod.rs
```diff
@@ -142,8 +142,16 @@ where
...
 impl MessagingService for MessagingHandle {
-    fn send_message(&mut self, peer_id: PeerId, message: SyncMessage) -> crate::Result<()> {
+    async fn send_message(&mut self, peer_id: PeerId, message: SyncMessage) -> crate::Result<()> {
+        // Note: send_message was made async to be able to use a bounded channel in tests.
+        // TODO: it may make sense to use a bounded channel in production as well. One possible
+        // reason for this might be to limit the number of BlockResponse messages that can exist
+        // at the same time. E.g. currently sync::Peer will produce a BlockResponse for each
+        // requested block almost immediately. The maximum number of simultaneously requested
+        // blocks is 500 per peer, a block can be up to 1Mb in size, so several malicious peers
+        // can eat up several Gb of RAM on the node.
```
ImplOfAnImpl, line 154: I've created an issue for it - https://github.com/mintlayer/mintlayer-core/issues/1314
As I've mentioned in the issue, I no longer think that using a bounded channel here will help, so I'll remove this TODO.

TheQuantumPhysicist (approved):
I don't really have big feelings about this.

TheQuantumPhysicist:
Please merge this only when we're ready to merge its base to master.

PR #1311, opened 2023-10-27
Enforce NoSignature data to follow a rule in ChainConfig for mainnet

TheQuantumPhysicist:
Also removed many tests' dependency on Mainnet ChainConfig. Soon we will be recreating the mainnet config from scratch, so nothing should depend on it.

### DIFF
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -836,7 +836,9 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
             for signature in tx.signatures() {
                 match signature {
                     common::chain::signature::inputsig::InputWitness::NoSignature(data) => {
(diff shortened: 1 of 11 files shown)


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -836,12 +836,22 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
             for signature in tx.signatures() {
                 match signature {
                     common::chain::signature::inputsig::InputWitness::NoSignature(data) => {
+                        if !self.chain_config.data_in_no_signature_witness_allowed()
+                            && data.is_some()
```
azarovh, line 840: Is it intended that this condition check data? As a result in mainnet it will be allowed to use NoSignature(None) which contradicts with the chain_config method name

azarovh (approved):

PR #1312, opened 2023-10-28
Feature/wallet token freeze unfreeze

OBorce:
- add new commands for freeze and unfreeze tokens
- check send/mint/unmint of tokens if they are frozen before creating the transaction
- add tests

### DIFF
--- a/wallet/src/account/output_cache/mod.rs
+++ b/wallet/src/account/output_cache/mod.rs
@@ -228,6 +228,12 @@ pub enum TokenFreezableState {
 }
 
 impl TokenFreezableState {
+    pub fn check_can_be_used(&self) -> WalletResult<()> {
+        match self {
+            Self::Frozen(_, _) => Err(WalletError::CannotUseFrozenToken),
+            Self::NotFrozen(_) => Ok(()),
+        }
+    }
     fn freeze(&self, is_unfreezable: IsTokenUnfreezable) -> WalletResult<Self> {
         match self {
             Self::NotFrozen(IsTokenFreezable::Yes) => {
@@ -796,6 +802,13 @@ impl OutputCache {
             WalletTx::Tx(tx_data) => make_token_id(tx_data.get_transaction().inputs()),
             WalletTx::Block(_) => None,
         };
+
+        // check token is not frozen and can be used
+        token_id
+            .and_then(|token_id| self.token_data(&token_id))
+            .map_or(Ok(()), |token_data| {
+                token_data.frozen_state.check_can_be_used()
+            })?;
 
         Ok((output, token_id))
     }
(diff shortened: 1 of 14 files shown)


azarovh (approved):

wallet/src/account/output_cache/mod.rs
```diff
@@ -221,21 +222,92 @@ impl TokenCurrentSupplyState {
     }
 }
 
+pub enum TokenFreezableState {
```
TheQuantumPhysicist, line 225: If possible, please always use False=0, and True=1. In many instances I see the opposite, which is confusing for no good reason. For example the IsTokenUnfreezable enum.

wallet/src/wallet/mod.rs
```diff
@@ -153,6 +153,14 @@ pub enum WalletError {
     CannotMintFixedTokenSupply(Amount, Amount, Amount),
     #[error("Trying to unmint {0:?} but the current supply is {1:?}")]
     CannotUnmintTokenSupply(Amount, Amount),
+    #[error("Cannot freeze an unfreezable token")]
+    CannotFreezeUnfreezableToken,
```
TheQuantumPhysicist, line 157: I think the word Unfreezable is ambiguous. It can mean two things:

1. A token that on issuance, was set to not be freezable (I would call that NonFreezable)
2. A token that was frozen, but can be unfrozen.

I would recommend disambiguating those cases by using the word NonFreezable where appropriate.

wallet/wallet-controller/src/synced_controller.rs
```diff
@@ -457,4 +506,65 @@ impl<'a, T: NodeInterface, W: WalletEvents> SyncedController<'a, T, W> {
...
+            self.wallet,
+            self.account_index,
+        )
+        .map_err(ControllerError::WalletError)?;
+
+        self.broadcast_to_mempool(tx).await
+    }
+
+    /// Create and broadcast a transaction that uses token,
+    /// check if that token can be used i.e. not friezed
```
TheQuantumPhysicist, line 532: frozen

wallet/src/wallet/tests.rs
```diff
@@ -1913,6 +1913,223 @@ fn check_tokens_v0_are_ignored(#[case] seed: Seed) {
...
+            DEFAULT_ACCOUNT_INDEX,
+            issued_token_id,
+            IsTokenUnfreezable::No,
+            FeeRate::new(Amount::ZERO),
+            FeeRate::new(Amount::ZERO),
+        )
+        .unwrap_err();
+    assert_eq!(err, WalletError::CannotFreezeAlreadyFrozenToken);
+
+    // cannot unfeeze
```
TheQuantumPhysicist, line 2121: unfreeze

TheQuantumPhysicist (approved):

Please merge this only after all the consensus stuff is merged, followed by a rebase.

wallet/src/account/output_cache/mod.rs
OBorce: changed the order

PR #1313, opened 2023-10-29
Temp fix for issuance and supply change fee in the same tx

azarovh:
I suggest this as a temporary fix until I implement this logic via ConstraintsAccumulator.
Doing it the right way via ConstraintsAccumulator turns out a bit more complicated than I thought. Also given the complexity of that solution I'm afraid that by rushing it I introduce more bugs than I fix.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/transferred_amount_check.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/transferred_amount_check.rs
@@ -21,10 +21,13 @@ use common::{
         block::{BlockRewardTransactable, ConsensusData},
         output_value::OutputValue,
         signature::Signable,
-        tokens::{get_tokens_issuance_count, make_token_id, TokenData, TokenId},
-        AccountOp, Block, OutPointSourceId, Transaction, TxInput, TxOutput,
+        tokens::{
+            get_token_supply_change_count, get_tokens_issuance_count, make_token_id, TokenData,
+            TokenId, TokenIssuanceVersion,
+        },
+        AccountOp, Block, ChainConfig, OutPointSourceId, Transaction, TxInput, TxOutput,
     },
-    primitives::{Amount, Id, Idable},
+    primitives::{Amount, BlockHeight, Id, Idable},
 };
 use fallible_iterator::FallibleIterator;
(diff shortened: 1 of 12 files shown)


chainstate/tx-verifier/src/transaction_verifier/transferred_amount_check.rs
```diff
@@ -84,13 +89,50 @@ where
...
+
+    let latest_token_version = chain_config
+        .chainstate_upgrades()
+        .version_at_height(current_height)
+        .1
+        .token_issuance_version();
+
+    match latest_token_version {
+        TokenIssuanceVersion::V0 => {
+            check_issuance_fee_burn_v0(chain_config, tx)?;
```
TheQuantumPhysicist, line 108: I'm assuming that this will completely die after the fork.

chainstate/tx-verifier/src/transaction_verifier/transferred_amount_check.rs
```diff
@@ -390,6 +432,72 @@ pub fn calculate_tokens_burned_in_outputs(
         .ok_or(ConnectTransactionError::BurnAmountSumError(tx.get_id()))
 }
 
+fn calculate_required_fee_burn(
```
TheQuantumPhysicist, line 435: I don't mind these gymnastics, considering they're temporary... but wouldn't have been easier to just make a copy of the current input calculation method, and add these to it as subtraction to the total?

TheQuantumPhysicist (approved):
Looking good.

chainstate/tx-verifier/src/transaction_verifier/transferred_amount_check.rs
azarovh: We will have to keep it in the testnet so that fee for older blocks is calculated correctly. But for the mainnet I hope this whole module will go away.

chainstate/tx-verifier/src/transaction_verifier/transferred_amount_check.rs
azarovh: Not sure it's rather big, I'd keep it like this

PR #1316, opened 2023-10-31
Fixes in p2p related to blocks order

ImplOfAnImpl:
1. `BlockResponse`s are now required to come in the same order in which the blocks were requested originally. If a `BlockResponse` comes out of order, it's treated in the same way as unsolicited blocks are treated.
2. When sending blocks we now update `best_sent_block` more carefully, taking into account the fact that they may have been requested in some random order.

The same fixes were made both in v1 and v2.

### DIFF
--- a/p2p/src/sync/peer_v1.rs
+++ b/p2p/src/sync/peer_v1.rs
@@ -103,7 +103,7 @@ struct IncomingDataState {
     /// requested the blocks for.
     pending_headers: Vec<SignedBlockHeader>,
     /// A list of blocks that we requested from this peer.
-    requested_blocks: BTreeSet<Id<Block>>,
+    requested_blocks: VecDeque<Id<Block>>,
     /// The id of the best block header that we've received from the peer and that we also have.
     /// This includes headers received by any means, e.g. via HeaderList messages, as part
     /// of a locator during peer's header requests, via block responses.
@@ -158,7 +158,7 @@ where
             time_getter,
             incoming: IncomingDataState {
                 pending_headers: Vec::new(),
-                requested_blocks: BTreeSet::new(),
+                requested_blocks: VecDeque::new(),
                 peers_best_block_that_we_have: None,
                 singular_unconnected_headers_count: 0,
             },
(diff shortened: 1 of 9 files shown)


p2p/src/sync/peer_v2.rs
```diff
@@ -708,14 +707,28 @@ where
...
+                            .front()
+                            .expect("The deque is known to be non-empty"),
+                        actual_block_id: block.get_id(),
+                    },
+                ));
+            } else {
+                return Err(P2pError::ProtocolError(
+                    ProtocolError::UnsolicitedBlockReceived(block.get_id()),
+                ));
+            }
```
ImplOfAnImpl, line 731: I'm not sure whether we need this distinction between `BlocksReceivedInWrongOrder` and `UnsolicitedBlockReceived`, provided that it's "informational" only (initially I was planning to treat them differently, that's why I introduced this separate `BlocksReceivedInWrongOrder` error)

p2p/src/sync/peer_v1.rs
```diff
@@ -734,10 +733,28 @@ where
...
 
-        if self.incoming.requested_blocks.take(&block.get_id()).is_none() {
-            return Err(P2pError::ProtocolError(ProtocolError::UnexpectedMessage(
-                "block response".to_owned(),
-            )));
+        if self.incoming.requested_blocks.front().is_some_and(|id| id == &block.get_id()) {
+            self.incoming.requested_blocks.pop_front();
+        } else {
+            let idx = self.incoming.requested_blocks.iter().position(|id| id == &block.get_id());
+            // Note: we treat wrongly ordered blocks in the same way as unsolicited ones, i.e.
```
TheQuantumPhysicist, line 740: Shouldn't we enforce this specifically in v2? Are we sure this is gonna be OK and not cause disconnections?

p2p/src/sync/peer_v2.rs
TheQuantumPhysicist: I think it's fine to keep the distinction. It may help to detect issues if we miss something.

test-utils/src/random.rs
```diff
@@ -60,3 +60,51 @@ impl From<u64> for Seed {
...
+pub fn shuffle_until_different<T>(slice: &mut [T], rng: &mut impl Rng) {
+    const MAX_ATTEMPTS: usize = 1000;
+
+    for _ in 0..MAX_ATTEMPTS {
+        let mut swapped = false;
+
+        for i in (1..slice.len()).rev() {
+            let other_idx = rng.gen_range(0..(i + 1));
+            if other_idx != i {
+                slice.swap(i, other_idx);
```
TheQuantumPhysicist, line 75: There's a method shuffle() from the SliceRandom trait, I believe...

TheQuantumPhysicist (approved):
Besides the minor comments, all good.

test-utils/src/random.rs
ImplOfAnImpl: Yes, but it may leave the collection unchanged as well. This variant ensures that the result is always different from the original.

TheQuantumPhysicist: If all you wanted to prove is that you're sending one block that isn't the first one, isn't easier to just swap the first element with a random element of the rest?

I don't really care that much, but this looks like hitting a fly with a bazooka. You can keep it, I don't mind.

p2p/src/sync/peer_v1.rs
ImplOfAnImpl: Even in V1 we maintain the same order when sending blocks (`outgoing.blocks_queue` is already a `VecDeque`).
So no, there shouldn't be any disconnects because of this change.
It's just that this behaviour wasn't enforced, but now it is.

PR #1318, opened 2023-10-31
Allow changing authority for a token

azarovh:
Allow changing authority for a token

I won't merge it before freezing functionality


chainstate/test-suite/src/tests/fungible_tokens_v1.rs
```diff
@@ -4814,3 +4814,471 @@ fn check_signature_on_freeze_unfreeze(#[case] seed: Seed) {
...
+            ))
+        );
+
+        let inputs_utxos = vec![
+            None,
+            tf.chainstate.utxo(&utxo_with_change).unwrap().map(|utxo| utxo.output().clone()),
+        ];
+        let inputs_utxos_refs = inputs_utxos.iter().map(|utxo| utxo.as_ref()).collect::<Vec<_>>();
+
+        // Try to lock with wrong signature
```
TheQuantumPhysicist, line 4886: We should also try locking with the old key.

common/src/chain/transaction/account_outpoint.rs
```diff
@@ -69,6 +72,9 @@ pub enum AccountOp {
     // By unfreezing token all operations are available for the tokens again
     #[codec(index = 5)]
     UnfreezeToken(TokenId),
+    // Change the authority who can authorize operations for a token
+    #[codec(index = 6)]
+    ChangeAuthority(TokenId, Destination),
```
TheQuantumPhysicist, line 77: ChangeTokenAuthority

muursh (approved):

tokens-accounting/src/operations.rs
```diff
@@ -68,6 +77,7 @@ pub enum TokenAccountingUndo {
     LockSupply(LockSupplyUndo),
     FreezeToken(FreezeTokenUndo),
     UnfreezeToken(UnfreezeTokenUndo),
+    ChangeTokenAuthority(ChangeTokenAuthorityUndo),
```
iljakuklic, line 80: Does this end up in the storage? If so, do we need to bump the version? My guess is that adding an enum variant at the end is fine but it's good to double check.

iljakuklic (approved):

tokens-accounting/src/operations.rs
TheQuantumPhysicist: We already bumped in another PR.

PR #1319, opened 2023-10-31
Miscellaneous fixes for the fork

TheQuantumPhysicist:
1. The asymptote for chain trust now has more points before flattening
2. The function `choose_peers_best_block()` now uses the chain trust instead of height, as should be.

### DIFF
--- a/p2p/src/sync/peer_common/mod.rs
+++ b/p2p/src/sync/peer_common/mod.rs
@@ -113,7 +113,7 @@ pub async fn handle_message_processing_result(
 }
 
 /// This function is used to update peers_best_block_that_we_have.
-/// The "better" block is the one that is on the main chain and has bigger height.
+/// The "better" block is the one that is on the main chain and has a higher chain-trust.
 /// In the case of a tie, new_block_id is preferred.
 pub fn choose_peers_best_block(
     chainstate: &dyn ChainstateInterface,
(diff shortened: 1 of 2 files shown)


p2p/src/sync/peer_common/mod.rs
```diff
@@ -124,11 +124,17 @@ pub fn choose_peers_best_block(
...
+                .get_gen_block_index(&old_id)?
+                .map_or(Uint256::ZERO, |idx: chainstate::GenBlockIndex| {
+                    idx.chain_trust()
+                });
+            let new_block_chain_trust = chainstate
+                .get_gen_block_index(&new_id)?
+                .map_or(Uint256::ZERO, |idx: chainstate::GenBlockIndex| {
+                    idx.chain_trust()
+                });
+            if new_block_chain_trust >= old_block_chain_trust {
```
ImplOfAnImpl, line 137: 1. The explicit type on the closures' parameters is not needed.
2. The documentation for this function is outdated - it still mentions "bigger height" and the block being "on the main chain".

ImplOfAnImpl (approved):

muursh (approved):

p2p/src/sync/peer_common/mod.rs
TheQuantumPhysicist: Done

PR #1327, opened 2023-11-02
Add support to change token authority in the wallet

OBorce:
- remove keeping track of token circulating supply and frozen state
- only keep track of unconfirmed transactions
- on each token change operation, the wallet will pull the latest token state from chainstate and apply the unconfirmed transactions on top of it
- add new command for changing the token authority from the wallet CLI

### DIFF
--- a/common/src/chain/tokens/rpc.rs
+++ b/common/src/chain/tokens/rpc.rs
@@ -13,9 +13,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::{Metadata, TokenCreator, TokenId, TokenTotalSupply};
+use super::{
+    IsTokenFreezable, IsTokenFrozen, IsTokenUnfreezable, Metadata, TokenCreator, TokenId,
+    TokenTotalSupply,
+};
 use crate::{
-    chain::{Block, Transaction},
+    chain::{Block, Destination, Transaction},
     primitives::{Amount, Id},
 };
 use serialization::{Decode, Encode};
@@ -35,9 +38,23 @@ impl RPCTokenInfo {
     pub fn new_nonfungible(token_info: RPCNonFungibleTokenInfo) -> Self {
         Self::NonFungibleToken(Box::new(token_info))
     }
+
+    pub fn token_id(&self) -> TokenId {
+        match self {
+            Self::NonFungibleToken(info) => info.token_id,
+            Self::FungibleToken(info) => info.token_id,
+        }
+    }
+
+    pub fn token_number_of_decimals(&self) -> u8 {
+        match self {
+            Self::FungibleToken(info) => info.number_of_decimals,
+            Self::NonFungibleToken(_) => 0,
+        }
+    }
 }
 
-#[derive(Debug, Clone, Encode, Decode, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, Copy, Encode, Decode, serde::Serialize, serde::Deserialize)]
 pub enum RPCTokenTotalSupply {
     Fixed(Amount),
     Lockable,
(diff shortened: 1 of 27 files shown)


test/functional/wallet_tokens_change_authority.py
```diff
@@ -0,0 +1,183 @@
...
+            assert_in("The transaction was submitted successfully", await wallet.unfreeze_token(token_id))
+
+
+
+if __name__ == '__main__':
+    WalletTokens().main()
+
+
+
+
```
TheQuantumPhysicist, line 183: Not that you have to change this now, maybe when you push next... but why too many lines at the end? One is more than enough :smile:

test/functional/wallet_tokens_change_authority.py
```diff
@@ -0,0 +1,180 @@
...
+            self.generate_block()
+            assert_in("Success", await wallet.sync())
+
+            ## check the new balance
+            assert_in(f"{token_id} amount: 9989.99", await wallet.get_balance())
+
+
+            assert_in("The transaction was submitted successfully", await wallet.change_token_authority(token_id, new_acc_address))
+
+            # randomly put the tx in a block or keep in mempool unconfirmed
```
TheQuantumPhysicist, line 154: Copy/paste error?

test/functional/wallet_tokens_freeze.py
```diff
@@ -147,6 +148,11 @@ async def async_test(self):
 
             assert_in("The transaction was submitted successfully", await wallet.freeze_token(token_id, 'unfreezable'))
 
+            # randomly put the tx in a block or keep in mempool unconfirmed
+            if random.choice([True, False]):
```
TheQuantumPhysicist, line 152: As you see in the tests in the node, we use a fixed seed to reproduce the exact same error. While I don't mind in this case because the choices are True/False, usually I'd print that outcome in the logs so that I can see what happened instead of having to guess.

Same applies to the other random.choice() below.

Also... Does the comment make sense?

TheQuantumPhysicist (approved):
Besides the comment, looking OK.

test/functional/wallet_tokens_change_authority.py
TheQuantumPhysicist: I actually meant one `\n` at the end, not an extra empty line :D

common/src/chain/tokens/rpc.rs
```diff
@@ -63,18 +136,23 @@ pub struct RPCFungibleTokenInfo {
     pub metadata_uri: Vec<u8>,
     pub circulating_supply: Amount,
     pub total_supply: RPCTokenTotalSupply,
-    pub is_frozen: bool,
+    pub is_locked: bool,
+    pub frozen: RPCIsTokenFrozen,
+    pub authority: Option<Destination>,
```
azarovh, line 141: not sure why is it optional? Is it because we still support v0 fungible tokens somewhere in the wallet?

azarovh (approved):

common/src/chain/tokens/rpc.rs
OBorce: because of the old v0 tokens

TheQuantumPhysicist: I'm completely OK with removing support for v0.

azarovh: I thought we dropped them from the wallet

test/functional/wallet_tokens_change_authority.py
OBorce: fixed

test/functional/wallet_tokens_freeze.py
OBorce: changed the seed log to info so we can see when a test fails what was the random seed number.

The comment means if true we will produce a new block and the wallet will sync and have the previous transaction as confirmed, otherwise it will be unconfirmed.
There is different logic in the wallet as it is fetching the token state from the chainstate which only contains the confirmed status, and for unconfirmed txs we apply them on top of that state in the wallet.

common/src/chain/tokens/rpc.rs
OBorce: removed

PR #1329, opened 2023-11-02
Add CORS layer to API Server

alfiedotwtf:
I'll set this to ready for review once it's tested on the frontend.

### DIFF
--- a/api-server/web-server/src/api/mod.rs
+++ b/api-server/web-server/src/api/mod.rs
@@ -27,9 +27,10 @@ use axum::{
     routing::{get, IntoMakeService},
     Json, Router, Server,
 };
+use hyper::Method;
 use serde_json::json;
 use std::{net::TcpListener, sync::Arc};
-use tower_http::cors::CorsLayer;
+use tower_http::cors::{Any, CorsLayer};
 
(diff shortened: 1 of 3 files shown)


TheQuantumPhysicist (approved):

api-server/web-server/src/api/mod.rs
```diff
@@ -52,7 +53,8 @@ pub fn web_server<T: ApiServerStorage + Send + Sync + 'static>(
         .route("/", get(server_status))
         .nest("/api/v1", api::v1::routes())
         .fallback(bad_request)
-        .with_state(state);
+        .with_state(state)
+        .layer(CorsLayer::permissive());
```
TheQuantumPhysicist, line 57: I'm double checking whether this is the right thing to do

TheQuantumPhysicist (changes requested):
I think using permissive() is a bad idea.

muursh (approved):

api-server/web-server/src/api/mod.rs
TheQuantumPhysicist: Changed to less permissive.

TheQuantumPhysicist (approved):

PR #1335, opened 2023-11-02
Remove access to chainstate database without a transaction

TheQuantumPhysicist:
Less macros, more responsibility.

### DIFF
--- a/chainstate/storage/src/internal/expensive.rs
+++ b/chainstate/storage/src/internal/expensive.rs
@@ -20,7 +20,7 @@ use common::chain::UtxoOutPoint;
 use storage::MakeMapRef;
 use utxo::Utxo;
 
-impl<'a, B: storage::Backend> StoreTxRo<'a, B> {
+impl<B: storage::Backend> StoreTxRo<'_, B> {
     /// Dump raw database contents
     pub fn dump_raw(&self) -> crate::Result<storage::raw::StorageContents<Schema>> {
         self.0.dump_raw().map_err(crate::Error::from)
--- a/chainstate/storage/src/internal/expensive.rs
+++ b/chainstate/storage/src/internal/expensive.rs
@@ -16,9 +16,11 @@
 use super::*;
 
 use crate::schema::{self as db};
+use common::chain::UtxoOutPoint;
 use storage::MakeMapRef;
+use utxo::Utxo;
 
-impl<B: storage::Backend> Store<B> {
+impl<'a, B: storage::Backend> StoreTxRo<'a, B> {
     /// Dump raw database contents
     pub fn dump_raw(&self) -> crate::Result<storage::raw::StorageContents<Schema>> {
         self.0.dump_raw().map_err(crate::Error::from)
@@ -26,8 +28,8 @@ impl<B: storage::Backend> Store<B> {
 
     /// Collect and return all utxos from the storage
     pub fn read_utxo_set(&self) -> crate::Result<BTreeMap<UtxoOutPoint, Utxo>> {
-        let db = self.transaction_ro()?;
-        db.0.get::<db::DBUtxo, _>()
+        self.0
+            .get::<db::DBUtxo, _>()
             .prefix_iter_decoded(&())
             .map(Iterator::collect)
             .map_err(crate::Error::from)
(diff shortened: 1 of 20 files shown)


chainstate/storage/src/internal/test.rs
```diff
@@ -112,23 +107,27 @@ fn test_storage_manipulation() {
...
-    assert_eq!(store.get_best_block_id(), Ok(Some(block1.get_id().into())));
+    assert_eq!(db_tx.get_best_block_id(), Ok(None));
+    assert_eq!(db_tx.set_best_block_id(&block0.get_id().into()), Ok(()));
+    assert_eq!(db_tx.get_best_block_id(), Ok(Some(block0.get_id().into())));
+    assert_eq!(db_tx.set_best_block_id(&block1.get_id().into()), Ok(()));
+    assert_eq!(db_tx.get_best_block_id(), Ok(Some(block1.get_id().into())));
 }
 
 #[test]
 fn get_set_transactions() {
```
TheQuantumPhysicist, line 118: There's an issue that I don't understand yet... this test some times fails, some times succeeds... It needs investigation because I think there's a race. The bad part there is that it it succeeds even when it fails, because the unwrap() that fails is in the thread, and it's not caught.

wallet/storage/src/internal/mod.rs
```diff
@@ -161,7 +161,7 @@ impl<B: storage::Backend> Store<B> {
 
     /// Dump raw database contents
     pub fn dump_raw(&self) -> crate::Result<storage::raw::StorageContents<Schema>> {
-        self.storage.dump_raw().map_err(crate::Error::from)
+        self.storage.transaction_ro().unwrap().dump_raw().map_err(crate::Error::from)
```
iljakuklic, line 164: I think `?` could be used instead of unwrap here but may be wrong.

chainstate/storage/src/lib.rs
```diff
@@ -258,4 +258,4 @@ pub trait Transactional<'t> {
     fn transaction_rw<'s: 't>(&'s self, size: Option<usize>) -> Result<Self::TransactionRw>;
 }
 
-pub trait BlockchainStorage: BlockchainStorageWrite + for<'tx> Transactional<'tx> + Send {}
```
iljakuklic, line 261: There is the `IsTransaction` marker trait, defined further up in this file. I believe it's used to differentiate storage access from transaction and storage access directly. Since the latter no longer exists, I think it can be removed. Unless it's also used for something else...

chainstate/storage/src/internal/mod.rs
```diff
@@ -117,96 +106,6 @@ impl<'tx, B: storage::Backend + 'tx> Transactional<'tx> for Store<B> {
...
-            Ok(val)
-        }
-    };
-    (@SIZE) => { None };
-    (@SIZE $s:literal) => { Some($s) };
-}
-
-impl<B: storage::Backend> BlockchainStorageRead for Store<B> {
-    delegate_to_transaction! {
-        fn get_storage_version(&self) -> crate::Result<Option<ChainstateStorageVersion>>;
```
iljakuklic, line 144: Nice to see all this stuff gone.

chainstate/storage/src/internal/expensive.rs
```diff
@@ -16,51 +16,56 @@
 use super::*;
 
 use crate::schema::{self as db};
+use common::chain::UtxoOutPoint;
 use storage::MakeMapRef;
+use utxo::Utxo;
 
-impl<B: storage::Backend> Store<B> {
+impl<'a, B: storage::Backend> StoreTxRo<'a, B> {
```
iljakuklic, line 23: Minor: You might be able to elide the lifetime here:

```suggestion
impl<B: storage::Backend> StoreTxRo<'_, B> {
```

chainstate/storage/src/internal/test.rs
```diff
@@ -353,11 +388,12 @@ fn undo_test(#[case] seed: Seed) {
...
     let mut rng = make_seedable_rng(seed);
-    let mut store = crate::inmemory::Store::new_empty().expect("should create a store");
-    store
+    let store = crate::inmemory::Store::new_empty().expect("should create a store");
+    let mut db_tx = store.transaction_rw(None).expect("should create a transaction");
+    db_tx
         .set_best_block_for_utxos(&H256::random_using(&mut rng).into())
         .expect("Setting best block cannot fail");
-    let mut db_interface = utxo::UtxosDB::new(&mut store);
+    let mut db_interface = utxo::UtxosDB::new(&mut db_tx);
```
iljakuklic, line 396: Looks like this whole test now runs in a rw transaction which at the end goes out of scope so nothing is written to the storage. While it seems fine in this test, it's a change in behaviour compared to what was there before.

chainstate/storage/src/internal/test.rs
```diff
@@ -331,20 +353,33 @@ fn undo_test(#[case] seed: Seed) {
     // create id:
     let id1: Id<Block> = Id::new(H256::random_using(&mut rng));
 
-    assert_eq!(store.get_undo_data(id1), Ok(None));
-    assert_eq!(store.set_undo_data(id1, &block_undo1), Ok(()));
+    assert_eq!(store.transaction_ro().unwrap().get_undo_data(id1), Ok(None));
     assert_eq!(
-        store.get_undo_data(id0).unwrap().unwrap(),
+        store.transaction_rw(None).unwrap().set_undo_data(id1, &block_undo1),
```
iljakuklic, line 358: This rw transaction should be committed to make the test behave as before. (although that scenario is probably also tested elsewhere)

iljakuklic (commented):
Just a quick first pass

chainstate/storage/src/lib.rs
TheQuantumPhysicist: Done

wallet/storage/src/internal/mod.rs
TheQuantumPhysicist: Done

chainstate/storage/src/internal/test.rs
TheQuantumPhysicist: Fixed. Thanks for noticing.

chainstate/storage/src/internal/expensive.rs
TheQuantumPhysicist: Done

TheQuantumPhysicist:
> Just a quick first pass

Pushed all fixes to the issues pointed out.

iljakuklic (approved):

azarovh (approved):
Nice!

PR #1337, opened 2023-11-03
fix wallet unittest gen_range invalid range

OBorce:
fix the range in the unit tests to have a valid range


TheQuantumPhysicist (approved):

TheQuantumPhysicist:
I force-merged this in the interest of testing more in CT.

PR #1348, opened 2023-11-08
p2p: sync manager's peer v2 was split into block sync manager and transaction sync manager

ImplOfAnImpl:
SyncMessage and the corresponding channel in the sync manager had to be split into 2 parts as well.

Edit: I've replaced the "txn" contraction with "tx" when referring to transactions. Also, to avoid confusion, I replaced "tx/rx" with "sender/receiver" when they refer to channel endpoints (everywhere in p2p).

Edit2: I've replaced  "tx" with "transaction" in most of the newly added code. I'm still keeping the "tx/rx"->"sender/receiver" renames though, because they improve readability and for consistency (some other places already use the "sender/receiver" suffixes).


p2p/src/message.rs
```diff
@@ -26,11 +26,22 @@ use serialization::{Decode, Encode};
...
     HeaderList(HeaderList),
     BlockResponse(BlockResponse),
+
+    // A "sentinel" message for testing purposes that allows to ensure that all block sync messages
+    // that were sent into a channel have been processed by the receiver.
+    // When the sync manager receives it, it should simply send it "back" via MessagingService.
+    // TODO: it would be nice to refactor tests that depend on this, so that it can be removed.
+    #[cfg(test)]
+    TestSentinel(Id<()>),
+}
```
ImplOfAnImpl, line 41: FYI: I intend to keep this for now

TheQuantumPhysicist: I'm OK with this as long as it's not in production. Though another solution would be to create a tunnel for the sender/receiver with a counter. But this is a fair amount of work and I don't really care right now.

TheQuantumPhysicist (approved):
Besides renaming txn to tx, I think this is looking OK.

p2p/src/sync/mod.rs
```diff
@@ -54,20 +54,20 @@ use crate::{
...
+#[derive(Debug, Clone)]
 pub enum LocalEvent {
     ChainstateNewTip(Id<Block>),
     MempoolNewTx(Id<Transaction>),
 }
 
 pub struct PeerContext {
-    task: JoinHandle<()>,
-    local_event_tx: UnboundedSender<LocalEvent>,
+    tasks: Vec<JoinHandle<()>>,
```
iljakuklic, line 64: Could this be a `tokio::task::JoinSet`?

p2p/src/sync/peer_v2/txn_mgr.rs
```diff
@@ -0,0 +1,334 @@
...
+    _chain_config: Arc<ChainConfig>,
+    p2p_config: Arc<P2pConfig>,
+    common_services: Services,
+    chainstate_handle: ChainstateHandle,
+    mempool_handle: MempoolHandle,
+    peer_manager_sender: UnboundedSender<PeerManagerEvent>,
+    messaging_handle: T::MessagingHandle,
+    sync_msg_rx: Receiver<TxnSyncMessage>,
+    local_event_rx: UnboundedReceiver<LocalEvent>,
+    _time_getter: TimeGetter,
```
iljakuklic, line 62: Should the unused members (chain config, time getter) be just removed for now? Can be added again quite easily if needed.

iljakuklic (approved):
+1 good stuff

PR #1351, opened 2023-11-10
Use ConstrainedValueAccumulator to check transferred amounts

azarovh:
New version of `ConstrainedValueAccumulator` replaces `transferred_amount_check` module for money printing checks and tx fees calculation.

`transferred_amount_check` module is still there for for checking the block reward. Though I think it can be deprecated at some point too because the check are too general for it now.


chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,60 +13,94 @@
...
-    primitives::{Amount, BlockDistance, BlockHeight},
+    primitives::{Amount, BlockHeight, Id},
 };
 use utils::ensure;
 
+use crate::{transaction_verifier::amounts_map::CoinOrTokenId, Fee};
+
 use super::IOPolicyError;
 
+type BlockCount = u64;
```
TheQuantumPhysicist, line 34: Isn't it better to use NonZeroU64?

azarovh: right, you mentioned this before and I tried it out. Maturity distance can be 0 and it actually is during IgnoreConsensus or PoW. If I make it NonZero I have to handle this specific case separately and account such coins in `unconstrained_values` instead which imho make the code more complicated.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,60 +13,94 @@
...
+            .map_err(|_| IOPolicyError::NegativeMaturityDistance)?;
+
+        let timelocked_change = self
+            .timelock_constrained
+            .into_iter()
+            .filter_map(|(lock, amount)| (lock <= maturity_distance).then_some(amount))
             .sum::<Option<Amount>>()
-            .and_then(|v| v + self.unconstrained_value)
-            .ok_or(IOPolicyError::AmountOverflow)
+            .ok_or(IOPolicyError::CoinOrTokenOverflow(CoinOrTokenId::Coin))?;
```
iljakuklic, line 77: I'm wondering if including timelocked coins does not bypass the point of the timelock. Some coins the owner would have to normally wait to be able to spend could be spent immediately if used to pay fees. That does not sound right, unless there are some considerations I missed.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -78,640 +112,440 @@ impl ConstrainedValueAccumulator {
...
+            OutputTimeLock::ForBlockCount(block_count) => {
+                // find the range that can be satisfied with the current timelock
+                let mut constraint_range_iter = self
+                    .timelock_constrained
+                    .range_mut((
+                        std::ops::Bound::Unbounded,
+                        std::ops::Bound::Included(block_count),
+                    ))
+                    .rev()
+                    .peekable();
```
iljakuklic, line 475: Looks like you could "prepend" (using `.chain`) the unconstrained coins amount and remove the `None` case  in the loop below.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -78,640 +112,440 @@ impl ConstrainedValueAccumulator {
...
+            }
+        }
+    }
+
     pub fn process_outputs(
         &mut self,
         chain_config: &ChainConfig,
+        block_height: BlockHeight,
         outputs: &[TxOutput],
     ) -> Result<(), IOPolicyError> {
```
iljakuklic, line 317: I'm wondering if the following approach would help:

The `process_outputs` function returns a value accumulator that represents when are coins/tokens spent and at what is the amount. Then we have inputs accumulator and outputs accumulator and can compare the "spending schedule" of both. I think it would make the checking algorithm simpler.

Just iterate over the amounts in the time lock order and keep a running total of "released" - "outputted" amounts and check that at no point it drops below zero. That works because from the accumulator, it is easy to get the amounts in the order in which the time locks are released.

iljakuklic (commented):
Some initial thoughts below. There are no showstoppers but would like to review once again when it's closer to the final form.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
azarovh: The fee is not available to the staker because it is distributed and added to the pool balance which cannot be spent without timelock constraints.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,60 +13,94 @@
...
 /// all outputs are timelocked when the pool is decommissioned `ConstrainedValueAccumulator` gives a way
 /// to check that an accumulated output value is locked for sufficient amount of time which allows
 /// using other valid inputs and outputs in the same tx.
-///
-/// TODO: this struct can be extended to collect tokens replacing `AmountsMap`
 pub struct ConstrainedValueAccumulator {
-    unconstrained_value: Amount,
-    timelock_constrained: BTreeMap<BlockDistance, Amount>,
+    unconstrained_value: BTreeMap<CoinOrTokenId, Amount>,
+    timelock_constrained: BTreeMap<BlockCount, Amount>,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: It should be possible to have tokens also in locked state. That's something to add in the future.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
azarovh: I like the idea, but couldn't implement it nicely. The problem is that unconstrained coins are optional so I resulted having conflicting types for the iterator.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
azarovh: So the last commit `2770982` implements this approach. Looks good, what's your thoughts?

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,705 +13,584 @@
...
-            )
-            .unwrap();
-
-        let result = constraints_accumulator.process_outputs(&chain_config, &outputs).unwrap_err();
-        assert_eq!(
-            result,
-            IOPolicyError::AttemptToPrintMoneyOrViolateTimelockConstraints
-        );
+    /// Return accumulated coins that are left
+    pub fn consume(
```
TheQuantumPhysicist, line 515: I would call this "map_into_block_fees", and write the documentation "Map/project the accumulator into an object that represents the accumulated block fees. We call this projection because the result of this projection is a subgroup of the `ConstrainedValueAccumulator` group under the 'accumulation' operation, where we can accumulate fees as well in the subgroup."

I'm OK with finding a better way to phrase this.

Currently it's OK to keep using the Fee type alias. Though in general, it'd be more meaningful to use a dedicated struct for this and implement `From<AccumulatedBlockFees> for ConstrainedValueAccumulator`. Maybe later.

TheQuantumPhysicist (approved):
Besides my minor comment, I think this is good to go.

@iljakuklic can you please review this again? I think this design has better mathematical meaning.

TheQuantumPhysicist:
Please isolate the accumulator in another PR into a separate crate so that it can easily be used in the API server.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
iljakuklic: > I like the idea, but couldn't implement it nicely. The problem is that unconstrained coins are optional so I resulted having conflicting types for the iterator.

Not sure it helps but you can get an iterator from `Option` (it returns zero or one element).

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/tests/homomorphism.rs
```diff
@@ -0,0 +1,289 @@
...
+        Amount::ZERO,
+    )
+}
+
+// Create 2 transactions. The first decommissions the pool and transfer some amount.
+// The second one spends share from delegation and also transfers some amount.
+// Actual amounts and the number of inputs/outputs used is random.
+// Process both transactions using a single ConstrainedValueAccumulator.
+// Then process them again but using separate accumulators per tx.
+// The resulting objects must be equal.
```
TheQuantumPhysicist, line 53: So this test is OK, but I have a few comments:

1. Put more emphasis on Case A and Case B, and then their combination Case C = A + B. I would explicitly add in the comment here that this test proves that f(a+b)=f(a)+f(b).
2. This is a very confined case. I would randomize this test much more than whatever is provided here, but I don't know the difficulty of doing this and whether this is urgent. You can probably do this later once you're done with the important stuff. Add a note about it. If it's gonna take you a week, then forget about it.
3. The `fee1` and `fee2` variables better be called after the convention, `fee_a` and `fee_b`, and then show that `fee_a`+`fee_b`=`fee_c`.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,705 +13,584 @@
...
-            )
-            .unwrap();
-
-        constraints_accumulator.process_outputs(&chain_config, &outputs).unwrap();
-
-        assert_eq!(constraints_accumulator.consume().unwrap(), Amount::ZERO);
+fn merge_amount_maps<K: Ord>(
+    left: &mut BTreeMap<K, Amount>,
+    right: BTreeMap<K, Amount>,
+) -> Result<(), IOPolicyError> {
```
iljakuklic, line 583: What slightly confused me when reading the code that uses this is the naming. `merge_` sounds like a pure function but it actually modifies the first argument. Since it behaves somewhat like `std::ops::AddAssign`, I'd prefer this to be called `merge_assign_amount_maps`.

Alternatively, if you can afford to change this method to take ownership of the first argument, it may make sense to change the method to have the following signature (where `AmountMap` = `BTreeMap<K, Amount>`):
```rust
fn merge(left: AmountMap, right: AmountMap) -> Result<AmountMap, IOPolicyError>;
```

That is, move the map into the merge function and return the modified one as a result. It has some advantages:
* More mathematical/functional interface avoiding implicit mutation of one of the arguments
* No danger of using the `left` map after an error.
* Could be implemented as an impl of `std::ops::Add` to get the `+` operator.
* The updates could still be performed in place.

The drawback is that it now requires passing the map into the function rather than just a reference.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,705 +13,584 @@
...
+    key: K,
+    amount: Amount,
+) -> Result<(), IOPolicyError> {
+    match total_amounts.entry(key) {
+        Entry::Occupied(mut entry) => {
+            let value = entry.get_mut();
+            *value = (*value + amount).ok_or(IOPolicyError::AmountOverflow)?;
+        }
+        Entry::Vacant(ventry) => {
+            ventry.insert(amount);
```
iljakuklic, line 555: Minor: I think the `Entry` API has methods for this pattern, something like this should do:
```rust
let value = total_amounts.entry(key).or_insert(Amount::ZERO);
*value = (*value + amount).ok_or(...)
```

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,705 +13,584 @@
...
-
-        constraints_accumulator.process_outputs(&chain_config, &outputs).unwrap();
-
-        assert_eq!(
-            constraints_accumulator.consume().unwrap().into_atoms(),
-            fee_atoms
-        );
+    pub fn combine(&mut self, other: ConstrainedValueAccumulator) -> Result<(), IOPolicyError> {
+        merge_amount_maps(&mut self.unconstrained_value, other.unconstrained_value)?;
+        merge_amount_maps(&mut self.timelock_constrained, other.timelock_constrained)?;
```
iljakuklic, line 451: Similar comment as for `merge_amounts_map`. If you could take `self` by value and return it, it would make for a more functional interface.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,705 +13,584 @@
...
-                )
-                .unwrap();
-
-            let result =
-                constraints_accumulator.process_outputs(&chain_config, &outputs).unwrap_err();
-            assert_eq!(
-                result,
-                IOPolicyError::AttemptToPrintMoneyOrViolateTimelockConstraints
-            );
+    pub fn subtract(&mut self, other: ConstrainedValueAccumulator) -> Result<(), IOPolicyError> {
```
iljakuklic, line 456: This would, I think, also be nicer if done in a more functional way.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
```diff
@@ -13,705 +13,584 @@
...
+                    std::ops::Bound::Included(timelock),
+                ))
+                .rev()
+                .peekable();
+
+            // iterate over the range until current output coins are completely used
+            // or all suitable constraints are satisfied
+            let mut output_coins = locked_coins;
+            while output_coins > Amount::ZERO {
+                match constraint_range_iter.peek_mut() {
```
iljakuklic, line 480: I am not quite convinced by the algorithm to subtract time-locked coins. We should check (probably using a randomized test) that the usual properties we would expect from an additive group hold, for exmaple:
* (A - B) + (C - D) = (A + C) - (B + D)
* A - (B + C) = A - B - C
* A - B = A + (0 - B)

I suspect they do not. If that's indeed the case, we should either:
* make the subtraction return a different type (maybe the resulting fees) so it's not being used in subsequent subsequent operations, or
* find a representation and algorithm that follow some sort of sensible set of algebraic identities

iljakuklic (commented):
Some more comments:

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
azarovh: Agree, reworked all the methods

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
azarovh: Good point, thanks. I added separate type and renamed `subtract` method to `satisfy_with` to avoid confusion.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/constraints_accumulator.rs
azarovh: Gave it another try and it worked out, thanks

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/consumed_constraints_accumulator.rs
```diff
@@ -0,0 +1,103 @@
...
+    chain::ChainConfig,
+    primitives::{Amount, BlockHeight},
+};
+
+use crate::{transaction_verifier::CoinOrTokenId, Fee};
+
+use super::IOPolicyError;
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct ConsumedConstrainedValueAccumulator {
```
iljakuklic, line 28: I'm not massively thrilled by the `ConsumedConstrainedValueAccumulator` name but was not able to come up with anything better. Maybe something that reflects that the Consumed version is used to accumulate values coming from transactions as whole whereas the normal (not Consumed) version accumulates over inputs and outputs.

iljakuklic (approved):
Looks good

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/consumed_constraints_accumulator.rs
azarovh: I also thought about `ConstrainedValueAccumulatorData`

TheQuantumPhysicist: I would call this AccumulatedFee.

PR #1354, opened 2023-11-13
Unify decommission delegation share spanding maturity and pool decommissioning maturity distances

TheQuantumPhysicist:
The way timelocks processes timelocked outputs into block rewards depends on this distance, so it makes sense to unify them as it simplifies processing anyway.

### DIFF
--- a/consensus/src/pos/effective_pool_balance.rs
+++ b/consensus/src/pos/effective_pool_balance.rs
@@ -210,7 +210,7 @@ fn effective_pool_balance_impl(
              Saturation level:    {pool_saturation_level}\n\
              Pledge influence:    {pledge_influence}\n\
              Final supply:        {final_supply_atoms}\n\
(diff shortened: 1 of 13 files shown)


consensus/src/pos/effective_pool_balance.rs
```diff
@@ -202,7 +202,7 @@ fn effective_pool_balance_impl(
         let pool_balance_atoms = pool_balance.into_atoms();
         let final_supply_atoms = final_supply.into_atoms();
 
-        log::debug!(
+        log::trace!(
```
azarovh, line 205: could you also add missing \n for line 213?

azarovh (approved):

consensus/src/pos/effective_pool_balance.rs
TheQuantumPhysicist: Done

PR #1355, opened 2023-11-13
Fix dns crawler crash

ImplOfAnImpl:
The problem was in the fact that:
1) `ConnectivityEvent::HandshakeFailed` is always accompanied by `ConnectivityEvent::ConnectionError`.
2) both `ConnectivityEvent::ConnectionError` and `ConnectivityEvent::HandshakeFailed` were translated into `CrawlerEvent::ConnectionError` in the dns crawler.

So, if `HandshakeFailed` was produced, the peer would effectively be "disconnected" twice, which lead to assertion failure.

The solution is to introduce a separate `CrawlerEvent` variant that corresponds to `HandshakeFailed`, and handle them separately: `CrawlerEvent::ConnectionError` will only result in peer being disconnected (and the corresponding `AddressData` being  updated) and the "handshake failed" variant will only result in the ban score handling.

To avoid confusion about what `HandshakeFailed` actually means, I renamed it to `MisbehavedOnHandshake` everywhere. Also, now I produce `PeerEvent::MisbehavedOnHandshake` only if the corresponsing error's ban score is non-zero. Though this is not strictly needed, it makes it more clear that the calling code shouldn't treat this event as a generic indication of connection failure.

Also, I added some additional tests for the dns crawler that check that `ConnectivityEvent::ConnectionError` by itself doesn't result in a ban score increase even if the error has non-zero ban score. Though I'm not sure whether such tests are really needed.


TheQuantumPhysicist (approved):

iljakuklic (approved):
Looks good to me. I am, however, not very familiar with the DNS server stuff so take my approval with a grain of salt.

PR #1356, opened 2023-11-14
API Server change TxOuput JSON format

OBorce:
- change Address and Amount outputs
- add Addressable for VRFPublicKey

### DIFF
--- a/api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
+++ b/api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
@@ -48,6 +48,13 @@ impl<'t> ApiServerStorageRead for ApiServerInMemoryStorageTransactionalRo<'t> {
         self.transaction.get_block(block_id)
     }
 
+    async fn get_transaction_with_block(
+        &self,
+        transaction_id: Id<Transaction>,
+    ) -> Result<Option<(Option<BlockAuxData>, SignedTransaction)>, ApiServerStorageError> {
+        self.transaction.get_transaction_with_block(transaction_id)
+    }
+
     async fn get_transaction(
         &self,
         transaction_id: Id<Transaction>,
--- a/api-server/stack-test-suite/tests/v1/block_reward.rs
+++ b/api-server/stack-test-suite/tests/v1/block_reward.rs
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use api_web_server::api::json_helpers::txoutput_to_json;
+
 use super::*;
 
 #[tokio::test]
@@ -154,12 +156,13 @@ async fn has_reward(#[case] seed: Seed) {
 
                     let genesis_id = tf.genesis().get_id();
 
+                    let (_, pk) = PrivateKey::new_from_rng(&mut rng, KeyKind::Secp256k1Schnorr);
                     let block = tf
                         .make_block_builder()
                         .with_parent(genesis_id.into())
                         .with_reward(vec![TxOutput::LockThenTransfer(
                             OutputValue::Coin(Amount::from_atoms(100)),
-                            Destination::AnyoneCanSpend,
+                            Destination::PublicKey(pk),
                             OutputTimeLock::ForBlockCount(0),
                         )])
                         .build();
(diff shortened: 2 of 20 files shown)


common/src/address/hexified.rs
```diff
@@ -337,47 +338,6 @@ mod tests {
         assert_eq!(res, "0xabcd");
     }
 
-    #[test]
-    fn invalid_match_should_replace_with_hex_creating_case_address_creation_error() {
```
TheQuantumPhysicist, line 341: Like discussed in the daily, see if you can restore these two tests with other cases that test the same thing and arrive to the same result.

common/src/chain/config/mod.rs
```diff
@@ -139,14 +139,14 @@ fn address_prefix(chain_type: ChainType, destination: &Destination) -> &'static
             Destination::ClassicMultisig(_) => "mmtc",
         },
         ChainType::Testnet => match destination {
-            Destination::AnyoneCanSpend => "",
+            Destination::AnyoneCanSpend => "anyonecanspend_address",
```
TheQuantumPhysicist, line 142: Please change it for every chain type, and add a prefix signifying the net type.

Does it have to be so long, btw? Does something like `tanyonecanspend` not work?

common/src/chain/config/mod.rs
```diff
@@ -245,6 +245,16 @@ impl ChainConfig {
...
     }
 
+    #[must_use]
+    pub fn vrf_public_key_address_prefix(&self) -> &'static str {
+        match self.chain_type {
+            ChainType::Mainnet => "mvrfpk",
+            ChainType::Testnet => "tvrfpk",
+            ChainType::Regtest => "rvrfpk",
+            ChainType::Signet => "svrfpk",
+        }
```
TheQuantumPhysicist, line 255: Should the wallet use these now for displaying the VRF public key? I don't know if we do display it for any reason.

api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
```diff
@@ -48,6 +48,13 @@ impl<'t> ApiServerStorageRead for ApiServerInMemoryStorageTransactionalRo<'t> {
         self.transaction.get_block(block_id)
     }
 
+    async fn get_transaction_with_block(
```
TheQuantumPhysicist, line 51: Why do we need a function to return a transaction with a block?

TheQuantumPhysicist (approved):

api-server/stack-test-suite/tests/v1/block_reward.rs
```diff
@@ -169,7 +172,12 @@ async fn has_reward(#[case] seed: Seed) {
 
                     _ = tx.send((
                         block_index.block_id().to_hash().encode_hex::<String>(),
-                        json!(block.block_reward().outputs().iter().clone().collect::<Vec<_>>()),
+                        json!(block
+                            .block_reward()
+                            .outputs()
+                            .iter()
+                            .map(|out| txoutput_to_json(out, &chain_config))
+                            .collect::<Vec<_>>()),
```
iljakuklic, line 180: It's a test so no big deal by anyway: Is the `json!` macro needed here? I don't see any json-specific syntax, just a rust expression. Also `serde_json::Value` implements `from_iter` so you should be able  to `.collect::<serde_jsno::Value>()` if I'm not mistaken.

iljakuklic (approved):
Looks good to me

api-server/stack-test-suite/tests/v1/block_reward.rs
OBorce: fixed with the collect to serde_json::Value

api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
OBorce: It is with the block aux data, used to get the block timestamp and height for the transaction endpoint's confirmations and timestamp.

common/src/chain/config/mod.rs
OBorce: changed the VRF public key printing in the wallet commands

common/src/chain/config/mod.rs
OBorce: used the shorted form works fine as well

Issue #832, opened 2023-04-14
Remaining p2p issues


pavel-kokolemin:
The list of some remaining p2p issues:

- Possible to ban output Tor nodes (bitcoin uses discouraging for this reason), forcing the node to use malicious exit nodes.

- Using Tor socks proxy has some deanonymization risks because we use static noise keys.

- IP whitelisting is not implemented. It may be difficult to establish persistent connections between your own nodes.

- Peer eviction logic does not take into account which peers are sending novel blocks and transactions.

- Feeler connections (used to test new addresses) are missing.

- Block-relay-only and anchor peers are not implemented (to mitigate topology inference and Eclipse attacks).

- Hard-coded list of reachable nodes is not used in cases when all DNS servers are unavailable.

- Unlimited address storage in PeerDb (DoS).

- Minimum relay fee is low (DoS).

- Transactions are relayed without delay, allowing deanonymization attacks.

- Address list responses are not cached to prevent fingerprinting.

Extensive reference to most known p2p attacks in the bitcoin network: https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9877811

TheQuantumPhysicist:
@pavel-kokolemin What's the status on this?

TheQuantumPhysicist:
To be clear I'm asking because it seems that we're working on some of them.

TheQuantumPhysicist:
Maybe update the readme file and let's close this one.

PR #1358, opened 2023-11-15
Initiate extra outbound full relay connections when the tip is stale

ImplOfAnImpl:
This completes the 2nd item of #832.

### DIFF
--- a/p2p/p2p-test-utils/src/lib.rs
+++ b/p2p/p2p-test-utils/src/lib.rs
@@ -50,13 +50,19 @@ pub fn start_subsystems(
         time_getter.clone(),
     )
     .unwrap();
-    start_subsystems_with_chainstate(chainstate, chain_config, time_getter)
+    start_subsystems_generic(
+        chainstate,
+        chain_config,
+        time_getter,
+        tracing::Span::current(),
+    )
 }
 
-pub fn start_subsystems_with_chainstate(
+pub fn start_subsystems_generic(
     chainstate: ChainstateSubsystem,
     chain_config: Arc<ChainConfig>,
     time_getter: TimeGetter,
+    tracing_span: tracing::Span,
 ) -> (
     ChainstateHandle,
     MempoolHandle,
@@ -71,7 +77,7 @@ pub fn start_subsystems_with_chainstate(
     let mempool = mempool::make_mempool(chain_config, chainstate.clone(), time_getter);
     let mempool = manager.add_custom_subsystem("p2p-test-mempool", |handle| mempool.init(handle));
 
-    let manager_handle = manager.main_in_task();
+    let manager_handle = manager.main_in_task_in_span(tracing_span);
 
     (chainstate, mempool, shutdown_trigger, manager_handle)
 }
--- a/p2p/p2p-test-utils/src/lib.rs
+++ b/p2p/p2p-test-utils/src/lib.rs
@@ -115,6 +115,10 @@ impl P2pBasicTestTimeGetter {
     pub fn advance_time(&self, duration: Duration) {
         self.current_time_millis.fetch_add(duration.as_millis() as u64);
     }
+
+    pub fn is_same_instance(&self, other: &P2pBasicTestTimeGetter) -> bool {
+        Arc::as_ptr(&self.current_time_millis) == Arc::as_ptr(&other.current_time_millis)
+    }
 }
 
 /// A timeout for blocking calls.
--- a/p2p/p2p-test-utils/src/lib.rs
+++ b/p2p/p2p-test-utils/src/lib.rs
@@ -123,7 +123,7 @@ impl P2pBasicTestTimeGetter {
     }
 
     pub fn is_same_instance(&self, other: &P2pBasicTestTimeGetter) -> bool {
-        Arc::as_ptr(&self.current_time_millis) == Arc::as_ptr(&other.current_time_millis)
+        Arc::ptr_eq(&self.current_time_millis, &other.current_time_millis)
     }
 }
 
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -79,6 +79,7 @@ use self::{
     peer_context::{PeerContext, SentPing},
     peerdb::storage::PeerDbStorage,
     peers_eviction::{
+        OutboundBlockRelayConnectionMinAge, OutboundFullRelayConnectionMinAge,
         PreservedInboundCountAddressGroup, PreservedInboundCountNewBlocks,
         PreservedInboundCountNewTransactions, PreservedInboundCountPing,
     },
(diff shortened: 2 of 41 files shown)


p2p/p2p-test-utils/src/lib.rs
```diff
@@ -115,6 +115,10 @@ impl P2pBasicTestTimeGetter {
     pub fn advance_time(&self, duration: Duration) {
         self.current_time_millis.fetch_add(duration.as_millis() as u64);
     }
+
+    pub fn is_same_instance(&self, other: &P2pBasicTestTimeGetter) -> bool {
+        Arc::as_ptr(&self.current_time_millis) == Arc::as_ptr(&other.current_time_millis)
```
TheQuantumPhysicist, line 120: I would recommend using `Arc::ptr_eq()`. No need to go this low.

p2p/src/peer_manager/peers_eviction/mod.rs
```diff
@@ -194,16 +186,60 @@ pub fn select_for_eviction_inbound(
...
+const BLOCK_RELAY_CONNECTION_MIN_AGE: Duration = Duration::from_secs(120);
+
+/// Outbound full relay connections younger than this age will not be taken into account
+/// during eviction.
+/// Note that extra full relay connections are established if the current tip becomes stale.
+/// There is no point in making this interval large, it should just give the peer enough
+/// time to send us a block.
+const FULL_RELAY_CONNECTION_MIN_AGE: Duration = Duration::from_secs(30);
+
 pub fn select_for_eviction_block_relay(
```
TheQuantumPhysicist, line 203: I suppose removing the `#[must_use]` is an oversight?

p2p/src/peer_manager/peers_eviction/mod.rs
```diff
@@ -194,16 +186,60 @@ pub fn select_for_eviction_inbound(
     find_group_most_connections(candidates)
 }
 
-#[must_use]
+/// Outbound block relay connections younger than this age will not be taken into account
+/// during eviction.
+/// Note that extra block relay connections are established and evicted on a regular basis
+/// during normal operation. So, this interval basically determines how often those extra
+/// connections will come and go.
+const BLOCK_RELAY_CONNECTION_MIN_AGE: Duration = Duration::from_secs(120);
```
TheQuantumPhysicist, line 194: I thought we're trying to get out of the excessive use of constants, and put them in configuration structs. I was happy we're moving towards that.

p2p/src/peer_manager/mod.rs
```diff
@@ -1574,16 +1654,21 @@ where
             let last_heartbeat_max =
                 (last_heartbeat + PEER_MGR_HEARTBEAT_INTERVAL_MAX).expect("Cannot happen 2");
 
+            let time_since_last_chainstate_tip =
+                (now - self.last_chainstate_tip_block_time).unwrap_or(Duration::ZERO);
+            let tip_is_stale = time_since_last_chainstate_tip > stale_tip_time_diff;
+
+            if tip_is_stale {
+                heartbeat_call_needed = true;
+            }
```
TheQuantumPhysicist, line 1663: I'm heavily opposed to solving problems by just adding a bunch of lines. If a bunch of lines are isolated in logic, they should go to another "thing", be it a function, struct or whatever. Please isolate the detection of stale tip. There's no reason why the logic of detecting a tip should be intermixed with other things, like clock and heartbeat stuff.

p2p/src/peer_manager/mod.rs
```diff
@@ -1662,13 +1751,19 @@ where
     }
 }
 
+pub fn stale_tip_time_diff(chain_config: &ChainConfig) -> Duration {
+    // Note: bitcoin core also uses "3 * block_spacing" for stale tip detection, but their
+    // "block spacing" is 10 min instead of out 2. TODO: should we use bigger time diff?
```
TheQuantumPhysicist, line 1756: Yes. It should be like 30 minutes to 1 hour. Block height doesn't mean anything on its own for security. What matters for security is time (which translates to chain-trust).

p2p/src/peer_manager/peers_eviction/mod.rs
```diff
@@ -194,16 +186,62 @@ pub fn select_for_eviction_inbound(
...
+
+/// Outbound full relay connections younger than this age will not be taken into account
+/// during eviction.
+/// Note that extra full relay connections are established if the current tip becomes stale.
+/// There is no point in making this interval large, it should just give the peer enough
+/// time to send us a block.
+const FULL_RELAY_CONNECTION_MIN_AGE: Duration = Duration::from_secs(30);
```
TheQuantumPhysicist, line 201: Isn't 30 seconds too short for this?

TheQuantumPhysicist:
Done with the first round.

p2p/src/peer_manager/peers_eviction/mod.rs
ImplOfAnImpl: I'm not sure. In bitcoin they use 30 seconds too (MINIMUM_CONNECT_TIME), this made me think it should be enough. (But note that they use the same 30s for extra block relay connections as well)

p2p/src/peer_manager/peers_eviction/mod.rs
ImplOfAnImpl: I renamed `ConnectionCountLimits` to `PeerManagerConfig` and put the new fields there.

p2p/src/peer_manager/mod.rs
ImplOfAnImpl: I made it 30 min (it's now in `PeerManagerConfig::stale_tip_time_diff`)

p2p/src/peer_manager/peers_eviction/mod.rs
ImplOfAnImpl: I made it 120 sec, the same as for block relay connections.

TheQuantumPhysicist (approved):

iljakuklic (approved):
Looks good

p2p/src/peer_manager/mod.rs
```diff
@@ -1090,29 +1145,107 @@ where
...
+                address,
+                OutboundConnectType::Automatic {
+                    block_relay_only: false,
+                },
+            );
+        }
+
+        let needed_outbound_block_relay_conn_count =
+            (*self.p2p_config.peer_manager_config.outbound_block_relay_count
+                + *self.p2p_config.peer_manager_config.outbound_block_relay_extra_count)
```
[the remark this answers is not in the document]
ImplOfAnImpl: Unless you insist, I'd rather keep it as is, because I'm not sure what would be a good name for this thing, provided that we also have the corresponding "full_relay_count/full_relay_extra_count" fields, for which such a sum doesn't always make sense.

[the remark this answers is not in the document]
iljakuklic: no problem

PR #1362, opened 2023-11-18
Appease clippy for 1.74

TheQuantumPhysicist:


merkletree/src/internal.rs
```diff
@@ -14,7 +14,7 @@
 // limitations under the License.
 
 // TODO: consider removing this in the future when fixed-hash fixes this problem
-#![allow(clippy::incorrect_clone_impl_on_copy_type)]
+#![allow(clippy::non_canonical_clone_impl)]
```
iljakuklic, line 17: Is the TODO logged as an issue?

iljakuklic (approved):

merkletree/src/internal.rs
TheQuantumPhysicist: Not sure I understand what you mean. This is a minor issue that can be casually checked every now and then. Nothing that we care about.

PR #1364, opened 2023-11-20
Detect panics and cancel (certain) p2p tests early if a panic has been detected

ImplOfAnImpl:
If a panic occurs inside a tokio task or a separate thread, the current test won't be aborted until the corresponding handle has been joined. In the case of p2p tests, depending on where the panic occurs, this may lead to the test being run until the timeout expires (which is currently 10 min).

Here I propose a (rather crude) solution to the problem - send a notification from a panic handler when a panic occurs; then make the function that waits for tests to complete also check for the panic notification.
Unfortunately, this solution only works if each test is run in a separate process, which is true for `nextest`, but not for the ordinary `cargo test`. So, it's disabled by default and is only enabled if `nextest` is detected or if a custom env var is set.


p2p/p2p-test-utils/src/lib.rs
```diff
@@ -132,6 +137,41 @@ pub const LONG_TIMEOUT: Duration = Duration::from_secs(600);
...
+/// may be important when debugging sporadically failing tests that can only be debugged by
+/// examining the logs (and running a test for LONG_TIMEOUT time may produce an enormous amount
+/// of logs). Unfortunately, this mechanism will not work properly if multiple tests are failing
+/// simultaneously or if there is a "should_panic" test. So, we only enable it if running under
+/// 'nextest' (which runs each test in a separate process) or if forced via a custom env var.
+pub async fn run_with_timeout<F>(future: F)
+where
+    F: Future,
+{
+    let running_under_nextest = std::env::var("NEXTEST").is_ok();
```
TheQuantumPhysicist, line 153: I would add a comment + documentation explaining that this is defined by default in nextest.

TheQuantumPhysicist (approved):
Besides the comment, looking good.

iljakuklic (approved):

PR #1367, opened 2023-11-21
Fix/api server json encodings

OBorce:
- fix wrong transaction ID in history
- add make_change_address function in wasm crypto
- fix the transaction output encodings in the block and genesis responses


TheQuantumPhysicist (approved):
Besides that tar.gz file that has to go, this looks fine.

PR #1368, opened 2023-11-22
Make min tx relay fee configurable

ImplOfAnImpl:
1) The minimum transaction fee per byte is now a config option.
2) Its default value was changed from 1 to 10'000 atoms.
3) We no longer punish peers for sending transactions with fees below this limit, we just ignore the transactions.

P.S. initially I added `min_tx_relay_fee_per_byte` to `ChainConfig`, but this turned out to be rather ugly, especially in the command line option/config file parsing code. So in the end I've added a separate `MempoolConfig` for it.

### DIFF
--- a/mempool/src/config.rs
+++ b/mempool/src/config.rs
@@ -18,6 +18,8 @@ use std::time::Duration;
 use common::primitives::{Amount, BlockDistance};
 use utils::make_config_setting;
 
+use crate::FeeRate;
+
 /// Mempool size configuration
 #[derive(Debug, Eq, PartialEq, PartialOrd, Ord, Clone, Copy)]
 pub struct MempoolMaxSize(usize);
@@ -71,12 +73,16 @@ pub const FUTURE_TIMELOCK_TOLERANCE: Duration = Duration::from_secs(5 * 60);
 
 pub const FUTURE_TIMELOCK_TOLERANCE_BLOCKS: BlockDistance = BlockDistance::new(5);
 
-// 10^-7 of a coin
-make_config_setting!(MinTxRelayFeePerByte, Amount, Amount::from_atoms(10_000));
+// 10^-4 of a coin per 1000 bytes
+make_config_setting!(
+    MinTxRelayFeeRate,
+    FeeRate,
+    FeeRate::from_amount_per_kb(Amount::from_atoms(10_000_000))
+);
 
 #[derive(Debug, Clone, Default)]
 pub struct MempoolConfig {
-    pub min_tx_relay_fee_per_byte: MinTxRelayFeePerByte,
+    pub min_tx_relay_fee_rate: MinTxRelayFeeRate,
 }
 
 impl MempoolConfig {
--- a/mempool/src/config.rs
+++ b/mempool/src/config.rs
@@ -52,9 +52,6 @@ pub const ROLLING_FEE_BASE_HALFLIFE: Duration = Duration::new(60 * 60 * 12, 1);
 
 pub const MAX_BIP125_REPLACEMENT_CANDIDATES: usize = 100;
 
-// TODO this should really be taken from some global node settings
-pub const RELAY_FEE_PER_BYTE: usize = 1;
-
 pub const MAX_MEMPOOL_SIZE_BYTES: usize = 300_000_000;
 
 pub const DEFAULT_MEMPOOL_EXPIRY: Duration = Duration::new(336 * 60 * 60, 0);
--- a/mempool/src/config.rs
+++ b/mempool/src/config.rs
@@ -73,11 +73,11 @@ pub const FUTURE_TIMELOCK_TOLERANCE: Duration = Duration::from_secs(5 * 60);
 
 pub const FUTURE_TIMELOCK_TOLERANCE_BLOCKS: BlockDistance = BlockDistance::new(5);
 
-// 10^-4 of a coin per 1000 bytes
+// 10^-3 of a coin per 1000 bytes
 make_config_setting!(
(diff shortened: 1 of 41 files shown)


mempool/src/config.rs
```diff
@@ -72,3 +70,17 @@ pub const MAX_ORPHAN_ACCOUNT_GAP: u64 = 2;
 pub const FUTURE_TIMELOCK_TOLERANCE: Duration = Duration::from_secs(5 * 60);
 
 pub const FUTURE_TIMELOCK_TOLERANCE_BLOCKS: BlockDistance = BlockDistance::new(5);
+
+// 10^-7 of a coin
+make_config_setting!(MinTxRelayFeePerByte, Amount, Amount::from_atoms(10_000));
```
iljakuklic, line 75: Since this is per byte, could the `FeeRate` type be used here?

mempool/src/config.rs
```diff
@@ -72,3 +70,17 @@ pub const MAX_ORPHAN_ACCOUNT_GAP: u64 = 2;
...
+
+#[derive(Debug, Clone, Default)]
+pub struct MempoolConfig {
+    pub min_tx_relay_fee_per_byte: MinTxRelayFeePerByte,
```
iljakuklic, line 79: Would be more accurate to rename both the field name and the type to state `feerate` rather than `fee`.

iljakuklic (approved):
Looks good. Some suggestions below:

ImplOfAnImpl:
I've rebased it on top of the current master.

mempool/src/pool/feerate.rs
```diff
@@ -31,10 +31,15 @@ pub struct FeeRate {
 }
 
 impl FeeRate {
+    // TODO: remove this, use the more descriptive from_amount_per_kb instead.
```
TheQuantumPhysicist, line 34: Why not kill this off immediately? Seems to be a copy of the function you wrote.

mempool/src/pool/mod.rs
```diff
@@ -1020,7 +1024,8 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
                         })
                 },
             )
-            .map(|feerate| std::cmp::max(feerate, INCREMENTAL_RELAY_FEE_RATE))
+            // TODO: should INCREMENTAL_RELAY_FEE_RATE also be taken into account here?
```
TheQuantumPhysicist, line 1027: @iljakuklic Could you please double-check this? This seems correct to me, but I'm not sure whether I'm missing something.

mempool/src/config.rs
```diff
@@ -72,3 +72,21 @@ pub const MAX_ORPHAN_ACCOUNT_GAP: u64 = 2;
 pub const FUTURE_TIMELOCK_TOLERANCE: Duration = Duration::from_secs(5 * 60);
 
 pub const FUTURE_TIMELOCK_TOLERANCE_BLOCKS: BlockDistance = BlockDistance::new(5);
+
+// 10^-4 of a coin per 1000 bytes
+make_config_setting!(
+    MinTxRelayFeeRate,
+    FeeRate,
+    FeeRate::from_amount_per_kb(Amount::from_atoms(10_000_000))
```
TheQuantumPhysicist, line 80: Please make it 100_000_000

TheQuantumPhysicist (approved):
Besides the few comments, looking good.

mempool/src/pool/tests/mod.rs
```diff
@@ -122,6 +130,52 @@ async fn add_single_tx() -> anyhow::Result<()> {
...
+    let tx1 = make_tx(&mempool, tx1_relay_fee).await;
+
+    let err = mempool.add_transaction_test(tx1).unwrap_err();
+    assert!(matches!(
+        err,
+        Error::Policy(MempoolPolicyError::InsufficientFeesToRelay {
+            tx_fee: _,
+            min_relay_fee: _
+        })
+    ));
```
iljakuklic, line 171: You can also assert ban score of the error is 0 here.

iljakuklic (approved):

mempool/src/pool/mod.rs
iljakuklic: Looks fine to me

mempool/src/pool/tests/mod.rs
ImplOfAnImpl: Won't it be redundant though? The corresponding p2p test already checks that the peer won't be punishes in such a situation.
Also, should mempool tests care about ban scores at all? As I can see, the existing ones don't.

iljakuklic: Yea it probably is redundant. Was thinking that if something can be checked using simpler setup, it should be preferred. Checking the ban score here only uses mempool (which is sufficient, everything needed to check it is defined in the mempool crate) rather than pulling in mempool _and_ p2p. Just a small suggestion though, doing it in p2p is fine too.

ImplOfAnImpl: The corresponding test in p2p is not really optional, because it actually checks the outcome of the transaction being received from a peer, rather than a ban score of certain error, which may or may not be taken into account later.

Then, checking the ban score here IMO is not only redundant, but it's also a bit strange. I.e. I've already checked here that a certain error is produced, so checking its ban score is equivalent to writing a separate test that only checks the ban score of that specific error. So the question arises why we only check the ban score of that error and not other ones.

Also, I don't think that ban scores, being a p2p concept, should exist outside of p2p at all. I think that the corresponding trait and its implementations should be moved to p2p eventually.

So, since you don't insist, I'm leaving this as is.

PR #1370, opened 2023-11-23
p2p, mempool: Introduce transaction submission options

iljakuklic:
* Add an argument to transaction submission RPCs specifying options how should the transaction be processed which is extensible in a backwards compatible way.
* Currently, it only supports passing trusted mode flag to the RPC which is intended to disable various mempool policy checks.
* The flag itself is not currently properly implemented.
* Refactor existing code to take transaction processing options.

### DIFF
--- a/mempool/src/event.rs
+++ b/mempool/src/event.rs
@@ -20,6 +20,7 @@ use common::{
 
 use crate::{
     error::{Error, MempoolBanScore},
+    tx_options::TxRelayPolicy,
     tx_origin::TxOrigin,
 };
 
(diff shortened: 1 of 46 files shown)


TheQuantumPhysicist (approved):

TheQuantumPhysicist:
@ImplOfAnImpl please check this PR.

ImplOfAnImpl (approved):

iljakuklic:
I found a way to do this in a backwards compatible way, which I have implemented locally in a separate commit. It changes how are the transaction submission arguments represented in Json.

In the following, `options` stands for the options key-value map (json object) where keys may be omitted to use defaults. In extreme case, it may be just `{}` to use defaults for all options.

### On master

```json
["TRANSACTION_HEX"]
```

* Not able to pass in options

### Proposal 1 (this PR)

```json
["TRANSACTION_HEX", options]
```

* Requests that use the old format will not work because two arguments are expected
* Forces the user to pass options explicitly (even if just defaults are used with `{}`)

### Proposal 2

Both of the following are supported

```json
["TRANSACTION_HEX"]
```
```json
[{"tx": "TRANSACTION_HEX", "options": options}]
```

* Backwards compatible
* Arguably more ergonomic if no special options have to be passed in
* Uses just one argument. Detects if options are passed based on the type (Json string or json object) in a slightly awkward way.
* If default options are used, there are two ways to do it (not just one canonical way)

Thoughts?

TheQuantumPhysicist:
Like mentioned in the chat, I'm OK with breaking backwards compatibility now since we're in testnet.

PR #1371, opened 2023-11-23
Dedicated fee for NFT issuance. Remove block time from PoSChainConfig

azarovh:
PR contains 2 changes:
- Separate NFT and funginble tokens issuance fee
- Remove block time from PoSChainConfig and use the one in ChainConfig

It's draft because there will be more conflicts when I finish with the accumulator


common/src/chain/config/mod.rs
```diff
@@ -459,9 +460,14 @@ impl ChainConfig {
         self.data_deposit_min_fee
     }
 
-    /// The fee for issuing a token
-    pub fn token_min_issuance_fee(&self) -> Amount {
-        self.token_min_issuance_fee
+    /// The fee for issuing a fungible token
+    pub fn fungible_token_min_issuance_fee(&self) -> Amount {
```
TheQuantumPhysicist, line 464: Btw, I think it's not appropriate to call them "min_xxxxx" anymore, because burning isn't explicit. Previously when we burned, burning more was OK, and hence this is "min" fee. Now since we're subtracting, this name doesn't make sense any more.

Please go through all the relevant "min" things and remove "min" from them.

consensus/src/pos/target.rs
```diff
@@ -69,24 +69,25 @@ fn calculate_new_target(
     pos_config: &PoSChainConfig,
     prev_target: &Uint256,
     actual_block_time: u64,
+    target_block_time: u64,
```
TheQuantumPhysicist, line 72: Why not use NonZeroU64, since we're checking this anyway in here?

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #1380, opened 2023-11-27
Ignore tokens v0

azarovh:
Tokens V0 are just ignored now, meaning they aren't processed and the info won't get to the storage. 
For backward compatibility with testnet the rule that tokens V0 must burn the fee was kept.


TheQuantumPhysicist:
You can remove more of such checks:

```
                        if current_tokens_version == TokenIssuanceVersion::V0 {
                            return Err(CheckBlockTransactionsError::DataDepositNotActivated(
                                current_height,
                                tx.transaction().get_id(),
                                block.get_id(),
                            ));
                        }
```

iljakuklic (approved):
+1 provided it cleanly syncs with the current chain. Nice cleanup otherwise.

azarovh:
> You can remove more of such checks:

Why? Do you mean that it's now safe to ignore it since the fork happened more than reorg depth?

TheQuantumPhysicist:
> > You can remove more of such checks:
> 
> Why? Do you mean that it's now safe to ignore it since the fork happened more than reorg depth?

Yep. We don't need to check if we're within the fork anymore.

azarovh:
> Yep. We don't need to check if we're within the fork anymore.

Ok, removed. Also removed the same rule for v1 tokens before the fork

TheQuantumPhysicist (approved):
Besides the discussed issue of NFT tests, this is good to go. Please ensure that any tests we remove for old NFTs are also tested for the new ones. If not, they should be shifted to the new version.

PR #1384, opened 2023-11-30
p2p: limit the amount of storage and memory occupied by addresses in peerdb

ImplOfAnImpl:
To limit the number of addresses, two fixed-size tables are maintained (one for "new" addresses, and the other one for "tried" addresses, i.e. the ones that we've connected to at least once). The tables consist of buckets; the index of a bucket and the position inside it is determined based on the hash of the address. So, when an address is added to a table, another one may be removed from it at the same time.
This is similar to what bitcoin-core does in its AddrMan.

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -1003,7 +1003,7 @@ where
                 PeerRole::OutboundFullRelay
                 | PeerRole::OutboundBlockRelay
                 | PeerRole::OutboundManual => {
-                    self.peerdb.report_outbound_failure(address, accept_err);
+                    self.peerdb.report_outbound_failure(address);
                 }
             }
         }
@@ -1022,7 +1022,7 @@ where
     /// Inform the [`crate::peer_manager::peerdb::PeerDb`] about the address failure so it knows to
     /// update its own records.
     fn handle_outbound_error(&mut self, address: SocketAddress, error: P2pError) {
-        self.peerdb.report_outbound_failure(address, &error);
+        self.peerdb.report_outbound_failure(address);
 
         let PendingConnect {
             outbound_connect_type,
@@ -1105,8 +1105,8 @@ where
     async fn query_dns_seed(&mut self) {
         let addresses = self.dns_seed.obtain_addresses().await;
         log::debug!("Dns seed queried, addresses = {addresses:?}");
-        for addr in addresses {
-            self.peerdb.peer_discovered(addr);
+        for addr in &addresses {
+            self.peerdb.peer_discovered(*addr);
         }
 
         self.last_dns_query_time = Some(self.time_getter.get_time());
@@ -1341,9 +1341,9 @@ where
             ))
         );
 
-        for address in addresses {
+        for address in &addresses {
             if let Some(address) = SocketAddress::from_peer_address(
-                &address,
+                address,
                 *self.p2p_config.allow_discover_private_ips,
             ) {
                 self.peerdb.peer_discovered(address);
@@ -1637,6 +1637,10 @@ where
             // might establish only a few OutboundBlockRelay connections and then stop establishing
             // new connections until both it's out of IBD and its tip becomes stale (following
             // the second part of the condition above).
+            // TODO: dns seeds are considered bad sources of peer addresses, because they can be
+            // compromised. It's better to avoid connecting to them unless necessary.
+            // See the TODO section of https://github.com/mintlayer/mintlayer-core/issues/832
+            // for extra details.
             true
         }
     }
--- a/p2p/src/peer_manager/peerdb/address_tables/mod.rs
+++ b/p2p/src/peer_manager/peerdb/address_tables/mod.rs
@@ -1,0 +1,209 @@
+// Copyright (c) 2021-2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::hash::Hash;
+
+use crypto::random::{make_pseudo_rng, Rng};
+use p2p_types::socket_address::SocketAddress;
+use serialization::{Decode, Encode};
+
+use self::table::Table;
+
+use super::config::PeerDbConfig;
+
+pub mod table;
+
+/// The random key that is used to randomize bucket selection inside the tables.
+///
+/// It will be randomly generated once and stored in the peer db.
+// TODO: is "salt" a better name for this?
+#[derive(Hash, Encode, Decode, Copy, Clone)]
+pub struct RandomKey(u64);
+
+impl RandomKey {
+    pub fn new_random() -> Self {
+        Self::new_random_with_rng(&mut make_pseudo_rng())
+    }
+
+    pub fn new_random_with_rng<R: Rng>(rng: &mut R) -> Self {
+        RandomKey(rng.gen::<u64>())
+    }
+
+    pub fn mix_with<T: Hash>(&self, data: T) -> RandomKey {
+        RandomKey(table::calc_hash(&(self.0, data)))
+    }
+}
+
+/// `AddressTables` provides a way of limiting the number of addresses in the peer db. When an
+/// address is added to it, an older one may be discarded at the same time.
+///
+/// `AddressTables` doesn't manage the actual `AddressData` itself. Instead, it just informs
+/// the caller code that a certain address was discarded, at which point the caller is supposed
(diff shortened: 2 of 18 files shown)


p2p/src/peer_manager/mod.rs
```diff
@@ -1637,6 +1637,8 @@ where
             // might establish only a few OutboundBlockRelay connections and then stop establishing
             // new connections until both it's out of IBD and its tip becomes stale (following
             // the second part of the condition above).
+            // TODO: dns seeds are considered bad sources of peer addresses, because they can be
+            // compromised. It's better to avoid connecting to them unless necessary.
             true
```
TheQuantumPhysicist, line 1642: Please create an issue about this, explaining the problem in more detail.

p2p/src/peer_manager/peerdb/new_tried_addr_table/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+pub struct RandomKey(u64);
+
+impl RandomKey {
+    pub fn new_random() -> Self {
+        RandomKey(make_pseudo_rng().gen::<u64>())
+    }
+}
+
+fn mix_random_key<T: Hash>(key: RandomKey, data: T) -> RandomKey {
+    RandomKey(calc_hash!(key.0, data))
```
TheQuantumPhysicist, line 45: Why not use a function instead of a macro?

p2p/src/peer_manager/peerdb/new_tried_addr_table/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+pub struct NewTriedAddrTable {
```
TheQuantumPhysicist, line 48: Please write a big comment explain what this is and how it works.

TheQuantumPhysicist:
So I did the first round, but I'll take another round later.

p2p/src/peer_manager/mod.rs
ImplOfAnImpl: I've described it in an additional item in the TODO section of https://github.com/mintlayer/mintlayer-core/issues/832

p2p/src/peer_manager/peerdb/new_tried_addr_table/mod.rs
ImplOfAnImpl: Done. Let me know if it's enough.
Also, I renamed the struct to just `AddressTables`.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -250,10 +343,92 @@ impl<S: PeerDbStorage> PeerDb<S> {
...
+
+            if let Some(addr_moved_to_new) = addr_moved_to_new {
+                tx.add_known_address(&addr_moved_to_new, KnownAddressState::New)?;
+            }
+
+            crate::Result::Ok(())
+        })
+        .expect("DB failure when updating known addresses}");
+
+        self.remove_from_addresses_if_some_non_reserved(discarded_addr);
```
iljakuklic, line 375: Might have missed something but a check that the tables contents is consistent with the persistent storage would be nice.

p2p/src/peer_manager/peerdb/address_tables/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+            new_addr_table,
+            tried_addr_table,
+        }
+    }
+
+    /// Remove the specified address from "new" and put it into "tried".
+    /// If an older "tried" address exists at the destination, move it into "new".
+    /// The return value is `Option<(old_tried, Option<old_new>)>`, where old_tried is the address
+    /// that was moved from 'tried' to 'new' and old_new is the address that was removed from
+    /// 'new' due to its place being taken by old_tried.
```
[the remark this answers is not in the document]
ImplOfAnImpl: I decided to make it a struct instead:
```
/// Optional side effects of `move_to_tried` (the additional addresses that it had to move).
pub struct MoveToTriedSideEffects {
    /// A "tried" address that had to be moved to "new".
    pub addr_moved_to_new: SocketAddress,
    /// The address that was removed from "new" and is no longer in the tables.
    pub discarded_new_addr: Option<SocketAddress>,
}
```
And `move_to_tried` returns `Option<MoveToTriedSideEffects>`.

I also changed the result of `move_to_new`, which now returns `MoveToNewOutcome`:
```
pub enum MoveToNewOutcome {
    /// The address was successfully put into the "new" table.
    Succeeded {
        /// The previous address that existed at the corresponding slot.
        prev_addr: Option<SocketAddress>,
    },
    /// `move_to_new` didn't do anything, because the address that already exists
    /// at the corresponding slot is better. 
    Cancelled,
}
```

p2p/src/peer_manager/peerdb/address_tables/table.rs
```diff
@@ -0,0 +1,532 @@
...
+    ($($val:expr),+) => {
+        {
+            let mut hasher = std::collections::hash_map::DefaultHasher::new();
+            $(
+                $val.hash(&mut hasher);
+            )*
+            hasher.finish()
+        }
+    };
+}
```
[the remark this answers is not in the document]
ImplOfAnImpl: I didn't realize I could use tuples to simulate a variadic function :-).
So, I replaced the macro with a function:
```
/// Calculate hash of a 'single' value.
/// 
/// Note: if multiple values are required to produce a single hash, pass them as a tuple.
pub fn calc_hash<T: Hash>(val: &T) -> u64 {
    let mut hasher = std::collections::hash_map::DefaultHasher::new();
    val.hash(&mut hasher);
    hasher.finish()
}
```
Using `BuildHasher::hash_one` IMO is redundant, because it's trivial and just introduces another level of indirection.

p2p/src/peer_manager/peerdb/address_tables/table.rs
```diff
@@ -0,0 +1,532 @@
...
+
+        (final_hash % self.buckets.len() as u64) as usize
+    }
+
+    fn bucket_pos(&self, addr: &SocketAddress, bucket_idx: usize) -> usize {
+        let hash = calc_hash!(self.random_key, addr, bucket_idx);
+        (hash % self.buckets[bucket_idx].len() as u64) as usize
+    }
+
+    pub fn entry(&self, addr: &SocketAddress) -> Option<&SocketAddress> {
```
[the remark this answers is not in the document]
ImplOfAnImpl: IMO `get` wouldn't be a good name, because the function may return a different address. So it's not like `BTreeSet::get`. And `get_entry` seems redundant.
Also, I already use the word "entry" in other places to refer to an "item" of the table. So IMO this function must have "entry" in its name. I don't mind using some other word, to avoid confusion with `BTreeMap::entry`, but I'm not sure which one would be descriptive enough.

p2p/src/peer_manager/peerdb/address_tables/table.rs
```diff
@@ -0,0 +1,532 @@
...
+                for id in bucket {
+                    if *id < self.id_max {
+                        assert!(self.addresses.contains_key(id));
+                        entries_in_buckets += 1;
+                    } else {
+                        assert!(!self.addresses.contains_key(id));
+                    }
+                }
+            }
+
```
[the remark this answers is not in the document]
ImplOfAnImpl: Good idea. But I added the extra checks to the existing loop instead.

p2p/src/peer_manager/peerdb/mod.rs
ImplOfAnImpl: The newly added tests call the `assert_addr_consistency` function which checks exactly this. Or do you mean I should add some checks to the production code too?

iljakuklic: > The newly added tests call the `assert_addr_consistency` function which checks exactly this. Or do you mean I should add some checks to the production code too?

That's fine, thanks

p2p/src/peer_manager/peerdb/address_tables/table.rs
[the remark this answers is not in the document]
iljakuklic: There's a subtle difference. Going over `addresses` separately also detects potential "dangling" entries with no corresponding entry in the vec table.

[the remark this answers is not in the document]
ImplOfAnImpl: Yeah, but I already check this indirectly by calculating the number of non-empty entries in the vec of vecs (`entries_in_buckets`) and then asserting that it's equal to `addresses.len()`. So, `addresses` can't contain extra elements.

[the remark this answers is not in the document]
iljakuklic: Ah, good point. All good then

iljakuklic (approved):

p2p/src/peer_manager/peerdb/new_tried_addr_table/mod.rs
TheQuantumPhysicist: It's OK now. But why is it thrown to the top of the file instead of it being on top of AddressTables?

TheQuantumPhysicist (approved):

p2p/src/peer_manager/peerdb/new_tried_addr_table/mod.rs
ImplOfAnImpl: I've put the comment on top of AddressTables.

ImplOfAnImpl:
Squashed and rebased on top of the current master

PR #1385, opened 2023-12-01
Feature/api server address delegations

OBorce:
- added next nonce for delegations endpoint
- a new endpoint for fetching delegations for an address
- a new wasm function for encoding account inputs


api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -614,10 +616,84 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+                "Delegation {} deserialization failed: {}",
+                delegation_id, e
+            ))
+        })?;
+
+        let delegation = Delegation::new(spend_destination, pool_id, balance, next_nonce);
         Ok(Some(delegation))
     }
 
+    pub async fn get_address_delegations(
```
TheQuantumPhysicist, line 630: `get_delegation_address`?

TheQuantumPhysicist: Or is it more appropriate to call it `get_delegation_from_address`?

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
Please mind the renames I mentioned in the chat and comments. Otherwise, good to go.

api-server/api-server-common/src/storage/storage_api/mod.rs
```diff
@@ -82,19 +89,25 @@ impl Delegation {
         &self.balance
     }
 
+    pub fn next_nonce(&self) -> &AccountNonce {
+        &self.next_nonce
+    }
+
     pub fn add_pledge(&self, rewards: Amount) -> Self {
```
azarovh, line 96: Missed that in previous PR, but `pledge` is not a correct term for delegation. We use delegate staking / spend share in context of delegation. Or add_delegation_balance alternatively

azarovh, line 96: look like it should accept self by value not ref. Same for sub

PR #1386, opened 2023-12-01
Create new noise adapter per socket

azarovh:
Create noise adapter every time to avoid de-anonymization through static key


TheQuantumPhysicist (approved):
Please let the others review this PR too. Looks good to me.

PR #1387, opened 2023-12-05
Set default BanDuration to 30 mins

azarovh:


ImplOfAnImpl (approved):
1. Please add a comment describing the reason behind choosing such small ban duration.

2. Probably `CrawlerConfig` in `dns_server/src/crawler_p2p/crawler/mod.rs` should now have its own `BanDuration` instead of re-using the p2p one.

ImplOfAnImpl:
> > Probably CrawlerConfig in dns_server/src/crawler_p2p/crawler/mod.rs should now have its own BanDuration instead of re-using the p2p one.
> 
> Should I leave the default value of 24h?

I think yes

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #1389, opened 2023-12-06
Implement feeler connections

ImplOfAnImpl:


p2p/src/peer_manager/tests/utils.rs
```diff
@@ -32,3 +37,21 @@ pub fn cmd_to_peer_man_msg(cmd: Command) -> (PeerId, PeerManagerMessage) {
...
+pub async fn recv_command_advance_time(
+    cmd_receiver: &mut mpsc::UnboundedReceiver<Command>,
+    time_getter: &P2pBasicTestTimeGetter,
+    advance_duration: Duration,
+) -> Result<Command, mpsc::error::TryRecvError> {
+    loop {
+        match cmd_receiver.try_recv() {
+            Err(mpsc::error::TryRecvError::Empty) => {
+                time_getter.advance_time(advance_duration);
+                tokio::time::sleep(Duration::from_millis(100)).await;
```
iljakuklic, line 50: Would `tokio::time::advance` work here? Could save on testing time.

p2p/src/peer_manager/peerdb/address_tables/table.rs
```diff
@@ -304,6 +303,24 @@ pub mod test_utils {
         result
     }
 
+    pub fn remove_collisions(
+        table: &Table,
+        addresses: impl Iterator<Item = SocketAddress>,
+    ) -> impl Iterator<Item = SocketAddress> {
```
iljakuklic, line 309: Does this actually `remove` anything? All I see is a map being filled with stuff and returned. The `table` argument is a shared reference so presumably not being changed.

iljakuklic (commented):
Code looks good. Still trying to understand the high-level architecture.

p2p/src/peer_manager/peerdb/address_tables/table.rs
ImplOfAnImpl: Well, it produces a new iterator over addresses, where all addresses map to distinct "bucket coordinates". I couldn't think of a better name.

iljakuklic: Ok, I see now. Sounds like `filter` or `dedup` but not sure how to incorporate it into a name. Something like `deduplicate_by_bucket_coords` seems a bit wordy. Not a big deal anyway.

iljakuklic (approved):

p2p/src/peer_manager/tests/utils.rs
ImplOfAnImpl: I don't think so. As I understand, `tokio::time::advance` will only cause the currently running calls of `sleep` to complete. Here the purpose of `sleep` is to give tasks some real time to do the job and to prevent the loop from advancing the mocked time too often.

p2p/src/peer_manager/peerdb/address_tables/table.rs
ImplOfAnImpl: Would `filter_out_collisions` be a better name?

iljakuklic: > Would `filter_out_collisions` be a better name?

Yea sounds good

PR #1390, opened 2023-12-07
Whitelist addresses

azarovh:
Adds ability to provide whitelisted addresses.

Some details:
- whitelisted address cannot be automatically banned by increased score;
- whitelisting uses ip address (no ports but no subnet either)
- adding to whitelist doesn't automatically unban 
- whitelisted addresses are not persisted


p2p/src/peer_manager/mod.rs
```diff
@@ -1466,6 +1473,14 @@ where
                 self.peerdb.unban(&address);
                 response_sender.send(Ok(()));
             }
+            PeerManagerEvent::Whitelist(address, response_sender) => {
+                self.peerdb.add_whitelisted_node(address);
```
TheQuantumPhysicist, line 1477: Shouldn't whitelisting be something in-memory? I might be wrong, but to me it sounds like whitelisting is something that can change per run.

azarovh: They are stored in memory only. I thought that it's a good place to put new set so that all address "groups" are managed by a single object. The fact the not everything is persisted might be confusing indeed. I can move it out to outer PeerManager

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -401,6 +401,16 @@ pub enum WalletCommand {
...
 
+    /// Add whitelist peer
+    AddWhitelistPeer {
+        address: IpOrSocketAddress,
+    },
+
+    /// Remove whitelist peer
+    RemoveWhitelistPeer {
+        address: IpOrSocketAddress,
+    },
```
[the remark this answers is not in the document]
azarovh: My bad, having reserved address commands confused me. Will remove it and add command line arguments

TheQuantumPhysicist:
I really don't mind keeping this separation of whitelisted and reserved addresses.

node-lib/src/options.rs
```diff
@@ -153,6 +158,10 @@ pub struct RunOptions {
     #[clap(long)]
     pub p2p_max_clock_diff: Option<u64>,
 
+    /// Optional list of whitelisted addresses.
+    #[clap(long, value_name = "ADDR")]
+    pub p2p_whitelist_addr: Option<Vec<IpAddr>>,
+
```
ImplOfAnImpl, line 164: Let's put this one right after `p2p_reserved_node`, as it's done in `P2pConfig`.

ImplOfAnImpl (approved):

TheQuantumPhysicist (approved):

PR #1392, opened 2023-12-08
P2p: choose roughly the same amount of new and tried addresses for outbound connections

ImplOfAnImpl:
When selecting addresses for outbound connections, roughly the same amount of "new" and "tried" addresses will now be selected.

Edit: I removed `choose_multiple_weighted` and re-implemented `PeerDb::select_non_reserved_outbound_addresses` in a simpler way.


utils/src/rand/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    let mut random_f64 = rng.gen::<f64>();
+    // The generated number will be in the range [0, 1). Turn it into (0, 1) to avoid
+    // infinity when taking the logarithm.
+    if random_f64 == 0.0 {
+        random_f64 = f64::MIN_POSITIVE;
+    }
+
+    #[allow(clippy::float_arithmetic)]
+    -random_f64.ln()
+}
```
ImplOfAnImpl, line 29: FYI: this is old code; for some reason git couldn't detect that the file was moved.

utils/src/rand/tests.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[test]
+    fn expect_finite_values_in_degenerate_cases() {
+        let mut always_zero_rng = StepRng::new(0, 0);
+        let val = exponential_rand(&mut always_zero_rng);
+        assert!(val.is_finite());
+
+        let mut always_max_rng = StepRng::new(u64::MAX, 0);
+        let val = exponential_rand(&mut always_max_rng);
+        assert!(val.is_finite());
+    }
```
ImplOfAnImpl, line 46: This is old code too.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -196,38 +195,77 @@ impl<S: PeerDbStorage> PeerDb<S> {
...
-                } else {
-                    None
+            }
+            None => {
+                debug_assert!(false, "Address {addr} not found in self.addresses");
+                false
+            }
+        };
+
+        let mut selected_new =
```
azarovh, line 232: I think it's worth leaving a comment explaining why exactly the algorithm is like that.

azarovh (approved):

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #1393, opened 2023-12-08
add feerate endpoint to API Server

OBorce:


iljakuklic (approved):
Some minor comments below:

api-server/web-server/src/api/v1.rs
```diff
@@ -68,12 +68,16 @@ pub fn routes<
...
-        router.route(
-            "/transaction",
-            post(submit_transaction).layer(DefaultBodyLimit::max(TX_BODY_LIMIT)),
-        )
+        router
+            .route(
+                "/transaction",
+                post(submit_transaction).layer(DefaultBodyLimit::max(TX_BODY_LIMIT)),
+            )
+            .route("/feerate", get(feerate))
```
TheQuantumPhysicist, line 76: I don't think it's the right thing to do to gate this. If this is made to prevent some kind of DoS attack against the node, the simpler solution is to cache the result locally in the web server and reuse it.

PR #1394, opened 2023-12-10
Introduce a wallet RPC daemon

iljakuklic:
Still missing:

* [x] Many RPC calls
* [ ] Tests
* [x] Quite a few command line configuration options
* [ ] RPC error handling could be improved
* [x] RPC input data decoding (from json) could be improved
* [x] Json representation of `Amount`s needs to be fixed
* [x] Json representation of `Address`es needs to be fixed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,8 @@ members = [
   "wallet/wallet-node-client",          # Wallet-to-node communication tools.
   "wallet/wallet-address-generator",    # Wallet address generator binary.
   "wallet/wallet-address-generator-lib",# Wallet address generator lib.
+  "wallet/wallet-rpc-daemon",           # Wallet RPC daemon binary.
+  "wallet/wallet-rpc-lib",              # Wallet RPC definitions library.
   "wallet/wallet-test-node",            # Node for wallet testing as a library.
   "wasm-crypto",                        # WASM bindings for the crypto crate.
 ]
--- a/wallet/wallet-rpc-daemon/src/cmdline.rs
+++ b/wallet/wallet-rpc-daemon/src/cmdline.rs
@@ -77,21 +77,21 @@ pub struct Args {
     /// Custom file path for the RPC cookie file.
     /// If not set, the cookie file is created in the data dir.
     #[arg(long, value_name("PATH"), conflicts_with_all(["rpc_username", "rpc_password"]))]
-    pub rpc_cookie_file: Option<PathBuf>,
+    rpc_cookie_file: Option<PathBuf>,
 
     /// Username for RPC server basic authorization.
     /// If not set, the cookie file is created.
     #[arg(long, value_name("USER"), requires("rpc_password"))]
-    pub rpc_username: Option<String>,
+    rpc_username: Option<String>,
 
     /// Password for RPC server basic authorization.
     /// If not set, the RPC cookie file is created.
     #[arg(long, value_name("PASS"), requires("rpc_username"))]
-    pub rpc_password: Option<String>,
+    rpc_password: Option<String>,
 
     /// Enable running the wallet service without RPC authentication
     #[arg(long, conflicts_with_all(["rpc_password", "rpc_username", "rpc_cookie_file"]))]
-    pub rpc_no_authentication: bool,
+    rpc_no_authentication: bool,
 }
 
 impl Args {
--- a/wallet/wallet-rpc-daemon/src/cmdline.rs
+++ b/wallet/wallet-rpc-daemon/src/cmdline.rs
@@ -1,0 +1,176 @@
+// Copyright (c) 2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::path::PathBuf;
+
+use clap::Parser;
+
+use common::chain::config::ChainType;
+use rpc::{
+    rpc_creds::{RpcCreds, RpcCredsError},
+    RpcAuthData,
+};
+use utils::ensure;
+use wallet_rpc_lib::config::{WalletRpcConfig, WalletServiceConfig};
+
+/// Service providing an RPC interface to a wallet
+#[derive(Parser)]
+#[command(
+    version,
+    about,
+    group(
+        clap::ArgGroup::new("rpc_auth")
+            .args(["rpc_cookie_file", "rpc_username", "rpc_password", "rpc_no_authentication"])
+            .required(true)
+    ),
+)]
+pub struct Args {
+    /// The wallet file to operate on
+    #[arg()]
+    wallet_path: PathBuf,
+
+    /// Address to bind the RPC interface to
+    #[arg(long, value_name("ADDR"))]
+    wallet_rpc_address: Option<String>,
+
+    /// Chain type to use
+    #[arg(long, value_name("TYPE"), value_parser(parse_chain_type))]
+    chain_type: ChainType,
+
+    /// RPC address of the node to connect to
+    #[arg(long, value_name("ADDR"))]
+    node_rpc_address: Option<String>,
+
+    /// Node RPC authentication cookie file path
+    #[arg(
+        long,
+        value_name("PATH"),
+        conflicts_with_all(["node_username", "node_password"]),
+    )]
+    node_cookie_file: Option<PathBuf>,
+
+    /// Node login user name
+    #[arg(long, value_name("NAME"), requires("node_password"))]
+    node_username: Option<String>,
+
+    /// Node login password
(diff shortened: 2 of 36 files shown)


Cargo.toml
```diff
@@ -63,6 +63,8 @@ members = [
   "wallet/wallet-node-client",          # Wallet-to-node communication tools.
   "wallet/wallet-address-generator",    # Wallet address generator binary.
   "wallet/wallet-address-generator-lib",# Wallet address generator lib.
+  "wallet/wallet-rpc-lib",              # Wallet RPC definitions library.
+  "wallet/wallet-rpcd",                 # Wallet RPC daemon binary.
```
TheQuantumPhysicist, line 67: We don't use the suffix `d` usually. I consider it an outdated standard that was used a long time ago when file names were short by necessity. Please call it wallet-rpc-daemon, in symmetry with node-daemon.

wallet/wallet-rpcd/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    let wallet_service_config = WalletServiceConfig {
+        chain_type,
+        wallet_file,
+        node_rpc_address,
+        node_cookie_file_path,
+    };
+
+    let rpc_config = WalletRpcConfig {
+        bind_addr: wallet_rpc_address.as_deref().unwrap_or("127.0.0.1:3333").parse().unwrap(),
```
TheQuantumPhysicist, line 47: Please use the same way we choose the default node RPC port. Having a fixed port for everything can cause conflicts between different network types. The right way is that for mainnet it can be 3333, for testnet 13333, for regtest 23333, etc. Check ChainConfig.

wallet/wallet-rpc-lib/src/rpc/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    async fn start(
+        wallet_handle: WalletHandle,
+        node_rpc: NodeRpcClient,
+        config: WalletRpcConfig,
+        chain_config: Arc<ChainConfig>,
+    ) -> anyhow::Result<rpc::Rpc> {
+        // TODO(PR) customize
+        let creds = None;
```
TheQuantumPhysicist, line 50: I think the same system for credentials that's used for node RPC should be supported for wallet RPC.

wallet/wallet-rpc-lib/src/config.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub wallet_file: PathBuf,
+
+    /// RPC address of the node to connect to
+    pub node_rpc_address: Option<String>,
+
+    /// Node RPC authentication cookie file path
+    pub node_cookie_file_path: Option<PathBuf>,
+}
+
+impl WalletServiceConfig {
```
TheQuantumPhysicist, line 35: Builder?

wallet/wallet-rpcd/src/main.rs
iljakuklic: This is preliminary, should have put a TODO there. Thanks for hints of how to go about it.

wallet/wallet-rpc-lib/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+use utils::shallow_clone::ShallowClone;
+
+const SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(thiserror::Error, Debug)]
+pub enum StartupError {
+    #[error(transparent)]
+    WalletService(#[from] service::InitError),
+
+    #[error("Failed to start RPC server: {0}")]
```
TheQuantumPhysicist, line 36: start?

TheQuantumPhysicist:
I seem to have missed that this is a draft. Apologies for the hastiness.

iljakuklic:
> I seem to have missed that this is a draft. Apologies for the hastiness.

No problem, you've given some useful feedback. The idea was to give people the chance to have a look at it before it's ready to be merged.

In general it seems three states could be useful:
* Ready: Please review, the author considers this to be a good candidate for merging.
* Draft 1: This is being actively worked on and will change (possibly on the basis of previous feedback), review not wanted
* Draft 2: The author does not consider this to be ready to be merged but would still like some feedback, reviews welcome.

This is more of a Draft 2 thing.

wallet/wallet-rpc-lib/src/lib.rs
iljakuklic: Yea this error type is only used during startup, there is a different error type for reporting things that go wrong while the system is running.

wallet/wallet-rpc-lib/src/service/handle.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use futures::future::{BoxFuture, Future};
+
+use utils::shallow_clone::ShallowClone;
+
+use super::worker::{self, WalletCommand, WalletController};
+
+/// Wallet handle allows the user to control the wallet service, perform queries etc.
+#[derive(Clone)]
+pub struct WalletHandle(worker::CommandSender);
```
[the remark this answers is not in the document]
iljakuklic: Yea. It's annoyingly close to subsystems (the RPC already is one out of the box). However the wallet service deals with the event loop differently from the other subsystems. It could be made into one by re-introducing custom event loops which I am not too keen on.

Alternatively, I could have a look how the `Handle` part of subsystems could be reused without pulling in the rest but then we only reuse the handle/call interface. We would not get the other things the subsystem crate implements like automatic clean shutdown and read-only calls parallelism. Especially the latter seems incompatible the structure of the wallet controller event loop. Even then some changes to the `subsytem` crate would be necessary even for this.

wallet/wallet-rpc-lib/src/rpc/types.rs
```diff
@@ -0,0 +1,88 @@
...
+    account: u32,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum RpcError {
+    #[error("Account index out of supported range")]
+    AcctIndexOutOfRange,
+
+    #[error("Invalid coin amount")]
+    InvaludCoinAmount,
```
OBorce, line 43: typo

wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -0,0 +1,66 @@
...
+    #[method(name = "shutdown")]
+    async fn shutdown(&self) -> rpc::RpcResult<()>;
+
+    #[method(name = "best_block")]
+    async fn best_block(&self, options: EmptyArgs) -> rpc::RpcResult<BlockInfo>;
+
+    #[method(name = "create_account")]
+    async fn create_account(&self, options: EmptyArgs) -> rpc::RpcResult<NewAccountInfo>;
+
+    #[method(name = "get_issued_addresses")]
```
TheQuantumPhysicist, line 36: We need to establish a new format for commands where we group related commands with a prefix. I'm working on that for the CLI wallet. Once the format is established, we'll use it here too.

wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    #[method(name = "send_coins")]
+    async fn send_coins(
+        &self,
+        account_index: AccountIndexArg,
+        address: String,
+        amount: AmountString,
+        options: TransactionOptions,
+    ) -> rpc::RpcResult<()>;
+}
```
TheQuantumPhysicist, line 66: We need a command to list all available commands. Knowing parity, I guess that doesn't exist. If you can do it, please go for it.

wallet/wallet-rpc-lib/src/service/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+            let rpc_address = {
+                let default_addr = || format!("127.0.0.1:{}", chain_config.default_rpc_port());
+                config.node_rpc_address.unwrap_or_else(default_addr)
+            };
+
+            wallet_controller::make_rpc_client(rpc_address, config.node_credentials).await?
+        };
+
+        let wallet = {
+            // TODO: Allow user to set password (config file only)
```
TheQuantumPhysicist, line 63: Please create an issue for this. Also please add in the issue that this file must have 600 permissions or similar to ensure that the password cannot be read by other users, otherwise it's kind of pointless.

TheQuantumPhysicist (approved):

wallet/wallet-rpc-lib/src/rpc/server_impl.rs
TheQuantumPhysicist, line 1: One thing I'm not particularly happy about is that generic functions like "get_balance" still have an implementation outside the controller. I would expect to have such a function in the controller, and UIs would just call that with zero logic in the UI implementation. Not sure whether I'm missing something that prevents this.

OBorce (approved):

TheQuantumPhysicist (approved):

wallet/wallet-rpc-daemon/src/cmdline.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub rpc_username: Option<String>,
+
+    /// Password for RPC server basic authorization.
+    /// If not set, the RPC cookie file is created.
+    #[arg(long, value_name("PASS"), requires("rpc_username"))]
+    pub rpc_password: Option<String>,
+
+    /// Enable running the wallet service without RPC authentication
+    #[arg(long, conflicts_with_all(["rpc_password", "rpc_username", "rpc_cookie_file"]))]
+    pub rpc_no_authentication: bool,
```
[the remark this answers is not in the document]
iljakuklic: This is due to historical development, I think I can just remove the `pub`s now. No longer required for anything.

PR #1400, opened 2023-12-13
Transactions should be relayed with random delay

azarovh:
This prevents deanonymization

### DIFF
--- a/p2p/src/sync/peer_v2/pending_transactions.rs
+++ b/p2p/src/sync/peer_v2/pending_transactions.rs
@@ -1,0 +1,58 @@
+// Copyright (c) 2021-2023 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{cmp::Reverse, collections::BinaryHeap, time::Duration};
+
+use common::{
+    chain::Transaction,
+    primitives::{Id, H256},
+    time_getter::TimeGetter,
+};
+
+use crate::Result;
+
+pub struct PendingTransactions {
+    time_getter: TimeGetter,
+    txs: BinaryHeap<Reverse<(Duration, Id<Transaction>)>>,
+}
+
+impl PendingTransactions {
+    pub fn new(time_getter: TimeGetter) -> Self {
+        Self {
+            time_getter,
+            txs: Default::default(),
+        }
+    }
+
+    pub fn push(&mut self, tx: Id<Transaction>, due_time: Duration) {
+        self.txs.push(Reverse((due_time, tx)));
+    }
+
+    pub async fn due(&mut self) -> Result<Id<Transaction>> {
+        if self.txs.is_empty() {
+            std::future::pending::<()>().await;
+            Ok(Id::<Transaction>::new(H256::zero())) // unreachable
+        } else {
+            let now = self.time_getter.get_time().as_duration_since_epoch();
+            let (due, _) = self.txs.peek().expect("cannot be empty").0;
+            if now < due {
+                tokio::time::sleep(due - now).await;
+            }
+
+            let (_, tx) = self.txs.pop().expect("cannot be empty").0;
+            Ok(tx)
+        }
+    }
+}
--- a/p2p/src/sync/peer_v2/pending_transactions.rs
+++ b/p2p/src/sync/peer_v2/pending_transactions.rs
@@ -15,13 +15,7 @@
 
 use std::{cmp::Reverse, collections::BinaryHeap, time::Duration};
 
-use common::{
-    chain::Transaction,
-    primitives::{Id, H256},
-    time_getter::TimeGetter,
-};
-
-use crate::Result;
+use common::{chain::Transaction, primitives::Id, time_getter::TimeGetter};
 
 pub struct PendingTransactions {
     time_getter: TimeGetter,
@@ -40,19 +34,23 @@ impl PendingTransactions {
         self.txs.push(Reverse((due_time, tx)));
     }
 
-    pub async fn due(&mut self) -> Result<Id<Transaction>> {
-        if self.txs.is_empty() {
-            std::future::pending::<()>().await;
-            Ok(Id::<Transaction>::new(H256::zero())) // unreachable
-        } else {
-            let now = self.time_getter.get_time().as_duration_since_epoch();
-            let (due, _) = self.txs.peek().expect("cannot be empty").0;
-            if now < due {
-                tokio::time::sleep(due - now).await;
-            }
+    pub fn pop(&mut self) -> Option<Id<Transaction>> {
+        self.txs.pop().map(|item| {
+            let (_, tx) = item.0;
+            tx
+        })
+    }
 
-            let (_, tx) = self.txs.pop().expect("cannot be empty").0;
-            Ok(tx)
+    pub async fn due(&self) {
+        match self.txs.peek() {
+            Some(item) => {
+                let now = self.time_getter.get_time().as_duration_since_epoch();
+                let (due, _) = item.0;
+                if now < due {
+                    tokio::time::sleep(due - now).await;
+                }
+            }
+            None => std::future::pending().await,
         }
     }
 }
--- a/p2p/src/sync/peer_v2/pending_transactions.rs
+++ b/p2p/src/sync/peer_v2/pending_transactions.rs
@@ -43,10 +43,9 @@ impl PendingTransactions {
     pub async fn due(&self) {
         match self.txs.peek() {
             Some(item) => {
-                let now = Instant::now();
                 let (due, _) = item.0;
-                if now < due {
-                    tokio::time::sleep(due - now).await;
+                if Instant::now() < due {
+                    tokio::time::sleep_until(due).await;
                 }
             }
             None => std::future::pending().await,
--- a/p2p/src/sync/peer_v2/pending_transactions.rs
+++ b/p2p/src/sync/peer_v2/pending_transactions.rs
@@ -13,24 +13,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{cmp::Reverse, collections::BinaryHeap, time::Duration};
+use std::{cmp::Reverse, collections::BinaryHeap};
+use tokio::time::Instant;
 
-use common::{chain::Transaction, primitives::Id, time_getter::TimeGetter};
+use common::{chain::Transaction, primitives::Id};
 
 pub struct PendingTransactions {
-    time_getter: TimeGetter,
-    txs: BinaryHeap<Reverse<(Duration, Id<Transaction>)>>,
+    txs: BinaryHeap<Reverse<(Instant, Id<Transaction>)>>,
 }
 
 impl PendingTransactions {
-    pub fn new(time_getter: TimeGetter) -> Self {
+    pub fn new() -> Self {
         Self {
-            time_getter,
             txs: Default::default(),
         }
     }
 
-    pub fn push(&mut self, tx: Id<Transaction>, due_time: Duration) {
+    pub fn push(&mut self, tx: Id<Transaction>, due_time: Instant) {
         self.txs.push(Reverse((due_time, tx)));
     }
 
@@ -44,7 +43,7 @@ impl PendingTransactions {
     pub async fn due(&self) {
         match self.txs.peek() {
             Some(item) => {
-                let now = self.time_getter.get_time().as_duration_since_epoch();
+                let now = Instant::now();
                 let (due, _) = item.0;
                 if now < due {
                     tokio::time::sleep(due - now).await;
--- a/p2p/src/sync/peer_v2/pending_transactions.rs
+++ b/p2p/src/sync/peer_v2/pending_transactions.rs
@@ -44,11 +44,115 @@ impl PendingTransactions {
         match self.txs.peek() {
             Some(item) => {
                 let (due, _) = item.0;
-                if Instant::now() < due {
-                    tokio::time::sleep_until(due).await;
-                }
+                tokio::time::sleep_until(due).await;
             }
             None => std::future::pending().await,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use super::*;
+
+    use common::primitives::H256;
+    use rstest::rstest;
+    use test_utils::random::{make_seedable_rng, Rng, Seed};
+
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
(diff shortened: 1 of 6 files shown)


test/functional/p2p_submit_orphan.py
```diff
@@ -71,7 +71,7 @@ def run_test(self):
         # Check both transactions have propagated to the last peer
         for node in self.nodes:
             has_txs = lambda: all(node.mempool_contains_tx(tx_id) for tx_id in [tx1_id, tx2_id])
-            self.wait_until(has_txs, timeout = 5)
+            self.wait_until(has_txs, timeout = 10)
```
TheQuantumPhysicist, line 74: Please add a comment specifying that this number should be larger than so and so

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -158,6 +174,23 @@ where
...
+        if now >= self.next_transaction_relay_time {
+            let ids = std::mem::take(&mut self.transactions_to_relay);
+            ids.into_iter().try_for_each(|txid| {
+                self.send_message(TransactionSyncMessage::NewTransaction(txid))
+            })?;
+
+            let limit = now + TX_RELAY_DELAY_INTERVAL_LIMIT;
+            let delay = TX_RELAY_DELAY_INTERVAL
+                .mul_f64(utils::exp_rand::exponential_rand(&mut make_pseudo_rng()));
+            self.next_transaction_relay_time = num_traits::clamp_max(now + delay, limit);
```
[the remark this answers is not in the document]
iljakuklic: Be aware that limiting the delay will create a spike in the delay probability distribution at the limit delay. I suspect it could be exploited for deanonmization. So if some limit is desirable, the probability of hitting it should be negligible. I see the current parameters, the limit is hit at twice the mean which gives the probability of the delay being exactly the (predictable) max amount of `exp(-2) = approx 0.135`. I don't know if that's sufficiently low to get the level of privacy we want but would be more comfortable with a lower value than this.

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -150,6 +164,8 @@ where
                 _ = maintenance_interval.tick() => {}
             }
 
+            self.announce_transactions_if_needed()?;
```
TheQuantumPhysicist, line 167: Why not use proper futures with await to do the delays, instead of this ad-hoc way of calculating times?

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -49,13 +50,18 @@ use crate::{
 
 use super::requested_transactions::RequestedTransactions;
 
+// TODO: add smaller interval for outbound connections
+pub const TX_RELAY_DELAY_INTERVAL: Duration = Duration::from_secs(5);
```
TheQuantumPhysicist, line 54: Why should there be a minimum amount and not start from zero?

azarovh: it's mean value not minimum

p2p/src/sync/peer_v2/transaction_manager.rs
azarovh: I'm not sure what you mean exactly by proper solution. There is an event loop per peer that we utilise to handle events and execute code.

TheQuantumPhysicist: Maybe I'm missing something, but let's see. Please discuss this with Mykhailo and Lukas and see whether they know how to exactly do this since they dealt with async more than I did.

Basically your event loop suspends at `select!`, and pops a future that's ready out of the different arms. What you're doing is ignoring that, and only after an event happens, manually check whether "it's time to send a transaction".

I believe there's a better way.

First, you have an async tokio sleep function that returns a future that can be an arm there. That's the first component in this puzzle. So, assuming we have only one transaction that you want to send out, all you have to do is await on the sleep future in that `select!` for that transaction.

But now we need to figure this out for multiple transactions. You need some kind of async iterator that iterates over elements of a container that contains all the transactions to be sent out, where every time a future is ready, it gets popped out of the container. Worst case scenario, you can create your own type and implement Future for that. However, I'm sure there's a way that wouldn't require you to do that manually since this is a typical use case. There's a `join!` macro for futures, but you want the opposite. You want something that takes a set of futures, and you should be able to pop elements out of it when they're ready. Not sure what that is. I'm sure we used that somewhere in the code, perhaps in the peers event loop to receive messages.

Now that would be a proper use of an async event loop. Hopefully I'm not missing something.

iljakuklic: > But now we need to figure this out for multiple transactions. You need some kind of async iterator that iterates over elements of a container that contains all the transactions to be sent out, where every time a future is ready, it gets popped out of the container. Worst case scenario, you can create your own type and implement Future for that. However, I'm sure there's a way that wouldn't require you to do that manually since this is a typical use case. There's a `join!` macro for futures, but you want the opposite. You want something that takes a set of futures, and you should be able to pop elements out of it when they're ready. Not sure what that is. I'm sure we used that somewhere in the code, perhaps in the peers event loop to receive messages.

The way you can go about this is to have a data structure collecting delayed set of actions. It could be something like a priority queue that returns the elements in the order of the delay. Also introduce a member function (let's call it `due`) that just sleeps for the amount of time needed for the first element to become due, or suspends (`future::pending`) if there's nothing. No need for a mechanism to wait on multiple futures since by keeping the elements in the priority queue, you can easily access the element that is supposed to be processed first and wait for that one. It could even be generic, parametrized over data it contains (transactions in this case), and chuck it in `utils`, I suspect it's useful for other things too.

TheQuantumPhysicist: Thanks for the details; that's very close to how I pictured it. I'm just hoping someone already created this because, in my head, it seems like a common thing to need. We should look online and try to find it.

azarovh: looks like `FuturesUnordered` is the standard tool for that

TheQuantumPhysicist: Sounds about right. I do remember we used this already elsewhere in the code. Cheers!

iljakuklic: Yea `FuturesUnordered` can work but may be too generic. It will poll all the futures in the set. However in this case you know the order in which the futures will resolve so it's sufficient to pool only the most immediate one. With `FuturesUnordered` the cost model (and by implication the security model with regards to DoS attacks) will be a bit more complicated (maybe not too bad since all transactions still have to pass the min fee check). Seems to me turning `O(n)` polls into `O(1)` is worth it since the additional complexity is not massive compared to the off the shelf solution. Remember the polls will be performed on every iteration of the event loop.

TheQuantumPhysicist: Two things:

1. This might be possible if it's possible to poll the remaining time from the future, which is not really that simple unless we want to recreate the adhoc problem. I might be wrong there though.
2. I'm OK with FuturesUnordered now, at least as a first go. We can create an issue and fix it later to make it idea. Spending days to create the ideal solution isn't really a priority now. I don't see how any of this can lead to DoS attacks, as long as the size of that container is limited.

iljakuklic: > 1. This might be possible if it's possible to poll the remaining time from the future, which is not really that simple unless we want to recreate the adhoc problem. I might be wrong there though.

It's not necessary to query the remaining time, just create the future as needed. Rough sketch:
```rust
loop {
    select! {
        // Other event loop stuff HERE
        _ = sleep(pending_txs.time_left_till_first_one()) => process_tx(pending_txs.pop()),
    }
}
```

Of course it should be properly encapsulated. Possibly the future for the first item deadline may be cached.

azarovh: take a look guys

p2p/src/sync/peer_v2/pending_transactions.rs
```diff
@@ -0,0 +1,58 @@
...
+            time_getter,
+            txs: Default::default(),
+        }
+    }
+
+    pub fn push(&mut self, tx: Id<Transaction>, due_time: Duration) {
+        self.txs.push(Reverse((due_time, tx)));
+    }
+
+    pub async fn due(&mut self) -> Result<Id<Transaction>> {
```
iljakuklic, line 43: I think it's better if `due` did not mutate the container to make the future cancellation safe [1]. A separate method will be needed to take the top entry. That way it will behave correctly if the future object is dropped. It may happen for example in `select!` with pattern match. I think it's fine the way it is used now but preferably the user does not need to think about it, making the interface simpler to reason about.

The downside of having a separate method is that the future checks if the value is there and you immediately pull it out using the other method. That introduces the Rust anti-pattern of having checks and return value separate instead of returning something like `Option`. That could be solved by returning some sort of `TopEntry` struct that allows you to pull out the top element later but that's maybe too convoluted?

Alternatively, just think through all the cancellation scenarios and write a big fat comment explaining why it's fine and under which conditions can this method be used without losing data.

By the way this returns `Result` but I don't see any errors.

[1] https://docs.rs/tokio/1.35.0/tokio/macro.select.html#cancellation-safety

iljakuklic (approved):

p2p/src/sync/peer_v2/pending_transactions.rs
azarovh: Agree, though it was cancellation safe I split the method anyway

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -171,10 +189,14 @@ where
                     && self.common_services.has_service(Service::Transactions)
                 {
                     self.add_known_transaction(txid);
-                    self.send_message(TransactionSyncMessage::NewTransaction(txid))
-                } else {
-                    Ok(())
+
+                    // TODO: whitelisted peers can get txs without delay
```
ImplOfAnImpl, line 191: "can" -> "should"?

p2p/src/sync/peer_v2/pending_transactions.rs
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{cmp::Reverse, collections::BinaryHeap, time::Duration};
+
+use common::{chain::Transaction, primitives::Id, time_getter::TimeGetter};
+
+pub struct PendingTransactions {
+    time_getter: TimeGetter,
+    txs: BinaryHeap<Reverse<(Duration, Id<Transaction>)>>,
```
ImplOfAnImpl: Plz don't use Duration when referring to a point in time. 
(Same below)

p2p/src/sync/peer_v2/pending_transactions.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    pub fn pop(&mut self) -> Option<Id<Transaction>> {
+        self.txs.pop().map(|item| {
+            let (_, tx) = item.0;
+            tx
+        })
+    }
+
+    pub async fn due(&self) {
```
ImplOfAnImpl, line 43: I'd suggest to pass `now` into this function instead of storing `time_getter` in the struct.

ImplOfAnImpl (approved):

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -171,10 +189,14 @@ where
...
+                    let now = self.time_getter.get_time().as_duration_since_epoch();
+                    let delay = TX_RELAY_DELAY_INTERVAL
+                        .mul_f64(utils::exp_rand::exponential_rand(&mut make_pseudo_rng()));
```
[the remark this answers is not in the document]
azarovh: good point thanks

p2p/src/sync/peer_v2/pending_transactions.rs
azarovh: Not using TimeGetter anymore

p2p/src/sync/peer_v2/pending_transactions.rs
```diff
@@ -0,0 +1,54 @@
...
+        match self.txs.peek() {
+            Some(item) => {
+                let (due, _) = item.0;
+                if Instant::now() < due {
+                    tokio::time::sleep_until(due).await;
+                }
```
iljakuklic, line 49: Is the check needed? Or is it an optimisation?

iljakuklic (approved):

p2p/src/sync/peer_v2/pending_transactions.rs
azarovh: It's me not sure if it's valid to sleep until instance in the past. Could see anything in the docs or the code.

iljakuklic: Ok. I don't mind keeping it. However, it should be fine to drop the check, `saturating_sub` is used to calculate the waiting time.

https://docs.rs/tokio/latest/src/tokio/time/sleep.rs.html#267

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -68,6 +74,9 @@ pub struct PeerTransactionSyncManager<T: NetworkingService> {
     /// This tracks transactions that we've requested from this peer but for which we haven't
     /// received a response yet.
     requested_transactions: RequestedTransactions,
+    /// Txs aren't relayed immediately but rather put into a collection to be propagated later
+    /// with random delay to avoid tracking
```
TheQuantumPhysicist, line 78: to make tracing transactions origin harder

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -171,10 +187,14 @@ where
                     && self.common_services.has_service(Service::Transactions)
                 {
                     self.add_known_transaction(txid);
-                    self.send_message(TransactionSyncMessage::NewTransaction(txid))
-                } else {
-                    Ok(())
+
+                    // TODO: whitelisted peers should get txs without delay
```
TheQuantumPhysicist, line 191: Is this really necessary?

p2p/src/sync/peer_v2/pending_transactions.rs
```diff
@@ -0,0 +1,54 @@
...
+            }
+            None => std::future::pending().await,
+        }
+    }
+}
```
TheQuantumPhysicist, line 54: Please add tests for the functionality of this struct.

TheQuantumPhysicist (approved):
Besides the comments, this is now OK.

p2p/src/sync/peer_v2/transaction_manager.rs
```diff
@@ -147,6 +157,12 @@ where
                     self.handle_local_event(event)?;
                 }
 
+                _ = self.pending_transactions.due() => {
+                    if let Some(new_tx) = self.pending_transactions.pop(){
```
TheQuantumPhysicist, line 161: Btw, you can merge both pop and due together and make something like async_pop where the returned value is something you can use directly, instead of having to pop then check.

azarovh: I originally had it that way, but Lukas made a point that it's harder to prove cancelation safety so a split it.

TheQuantumPhysicist: Alrighty

p2p/src/sync/peer_v2/transaction_manager.rs
TheQuantumPhysicist: Please create an issue for it.

PR #1401, opened 2023-12-13
Add mempool RPC for fetching feerate points

OBorce:


mempool/src/pool/mod.rs
```diff
@@ -1104,5 +1154,58 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
...
+
+fn generate_points(first: usize, last: usize, n: usize) -> Vec<usize> {
+    if first == last {
+        vec![first];
+    }
+
+    let mut points = Vec::with_capacity(n);
+    points.push(first);
+
+    let step = (last - first) / (n - 1);
```
TheQuantumPhysicist, line 1186: Please ensure that `n > 1`

mempool/src/pool/mod.rs
```diff
@@ -1104,5 +1154,58 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
...
+
+    for i in 1..(n - 1) {
+        points.push(first + step * i);
+    }
```
TheQuantumPhysicist, line 1190: This method will maximize off-by-one errors when `mod(last-first, n-1) == n-2`, but honestly I don't really care. Whatever works is fine since this is for interpolation.

However, please write tests for these simple functions.

mempool/src/pool/mod.rs
```diff
@@ -1104,5 +1154,58 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
...
+                Some(v2)
+            }
+        }
+        (Some((_, &v1)), None) => Some(v1),
+        (None, Some((_, &v2))) => Some(v2),
+        (None, None) => None,
+    }
+}
+
+fn generate_points(first: usize, last: usize, n: usize) -> Vec<usize> {
```
TheQuantumPhysicist, line 1178: call it generate_equidistant_span, please.

And write in the docs of the function `// Generate equidistant points to evaluate the mathematical function at for interpolation`

mempool/src/pool/mod.rs
```diff
@@ -1104,5 +1154,58 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
     }
 }
 
+fn get_closest_value<V: Copy>(size_to_score: &BTreeMap<usize, V>, key: usize) -> Option<V> {
```
TheQuantumPhysicist, line 1157: Can we use linear interpolation instead of this "closest" value? The problem here is that the error is really unbounded, especially at the boundaries.

Now this is not that hard to do. Two things need to happen to make this work:

1. `key` has to be bounded such that it's within the min and max of that map.
2. The value can be calculated with a fixed formula that can be found here under the "Linear interpolation between two known points": https://en.wikipedia.org/wiki/Linear_interpolation

Just slap the formula and it'll be fine.

Please write tests for this as well that ensure that the values make sense. So getting the value at a known point should yield that point, and in-between two points should be the value between them.

Finally, the linear interpolation function can be thrown into utils... since it has to be used also in the API server.

mempool/src/pool/mod.rs
```diff
@@ -1053,20 +1063,60 @@ impl<M: MemoryUsageEstimator> Mempool<M> {
...
-                },
+                || Ok(min_feerate),
+                |(score, _txs)| score_to_feerate(score, min_feerate),
             )
-            .map(|feerate| std::cmp::max(feerate, *self.mempool_config.min_tx_relay_fee_rate))
+    }
+
+    pub fn get_fee_rate_points(
+        &self,
+        num_points: usize,
```
TheQuantumPhysicist, line 1073: NonZeroUsize?

iljakuklic (commented):
Added some comments

mempool/src/pool/mod.rs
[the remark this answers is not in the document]
iljakuklic: > I am not convinced the number of points is a good way to parameterise this. Tn that case, the resolution depends on the number of transactions in the mempool. E.g. if the user asks for 10 points, in a mempool with a 200B transaction, you get a sample every 20B (needless accuracy) whereas in a mempool with 100MB worth of transactions, you get a sample every 10MB (way too inaccurate).
> 
> Maybe a more interesting would be to specify the sampling interval and a max limit. E.g. sample every 200kB and max 50 samples to cover top 10 blocks. Probably more intuitive that way.
> 
> As a bit more out there idea (not for this PR), maybe the sampling should not be uniform but logarithmic. I can imagine the feerate distribution close to the top of the mempool is more interesting than 10s or 100s of MB down the line.

Please file an issue for this and we can leave the code as is for now.

mempool/src/pool/feerate_points.rs
```diff
@@ -0,0 +1,190 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+use std::collections::BTreeMap;
+
+use common::primitives::Amount;
+
+use super::fee::Fee;
+
+pub fn linear_interpolation(x0: usize, y0: u128, x1: usize, y1: u128, x: usize) -> Option<u128> {
```
TheQuantumPhysicist, line 21: It's much more meaningful to make this tuples (x0, y0), (x1, y1), and x alone. 

Also, it would be really nice to make this work with generic types X and Y so that it can be reused/moved if necessary. I think it would be easy to do. The trait bounds are simply some math ops that will be given to you by the compiler, quite similar to the ones I did with the timing resistant equality in the crypto library. If it's too much work, don't do it.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -99,6 +100,16 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
         Ok(best_block)
     }
 
+    async fn update_mempool_feerate_points(
+        &mut self,
+        feerate_points: Vec<(usize, FeeRate)>,
+    ) -> Result<(), Self::Error> {
+        let mut db_tx = self.storage.transaction_rw().await.expect("Unable to connect to database");
+        db_tx.set_feerate_points(feerate_points).await?;
```
TheQuantumPhysicist, line 108: Why would you need the feerate points to be stored in the database at all? You just need some simple `Mutex<(Time,Points)>` on a value in your state. If current time minus Time stored there is larger than X, then update the cached value.

TheQuantumPhysicist (approved):
Besides the unnecessarily elaborate caching, this is fine. Please fix the caching issue then see if someone else has other problems with it. Then, feel free to merge.

mempool/src/pool/feerate_points.rs
OBorce: fixed to use generics

api-server/scanner-lib/src/blockchain_state/mod.rs
OBorce: fixed to use in memory caching state

mempool/src/pool/feerate_points.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    let mut points = Vec::with_capacity(n);
+    points.push(first);
+
+    let step = (last - first) / (n - 1);
+
+    for i in 1..(n - 1) {
+        points.push(first + step * i);
+    }
```
iljakuklic, line 84: I think the rounding error in `step` calculation may be quite significant here in certain cases. Let's say last = 27, first = 10, n = 10. Then step = floor(17 / 9) = 1, so we will generate [10, 11, 12, 13, 14, 15, 16, 17, 18, 27]. There's a significant jump between the second last and the last element. Even more extreme case, if last = 18 (other inputs unchanged), we get step = floor(8 / 9) = 0, generating [10, 10, 10, 10, 10, 10, 10, 10, 10, 17].

Evaluating the whole thing at each step (`i * (last - first) / (n - 1)`) should avoid the issue but then you may need to deal with overflow or use a wider type.

Not sure how much this is an issue in practice for calculating feerate points but the function looks quite generic and if somebody uses it in another setting, it may be a problem.

mempool/src/pool/feerate_points.rs
```diff
@@ -0,0 +1,196 @@
...
+};
+
+use common::primitives::Amount;
+
+use super::fee::Fee;
+
+pub fn linear_interpolation<X, Y>(p0: (X, Y), p1: (X, Y), x: X) -> Option<Y>
+where
+    X: Add<Output = X> + Sub<Output = X> + Eq + Ord + Copy,
+    Y: Mul<Output = Y> + Add<Output = Y> + Div<Output = Y> + Copy + From<X>,
```
iljakuklic, line 28: Note that all these bounds signify unchecked arithmetic. If the function is supposed to check for overflows of these operations, I'd expect to see something like `X: Op<Output = Option<X>>`

iljakuklic (commented):
I have some remaining concerns. Could you please handle the cases described below or add comments explaining why the current implementation is fine.

api-server/web-server/src/api/v1.rs
```diff
@@ -315,8 +317,77 @@ async fn get_transaction(
...
+                .collect()
+        }
+    };
+
+    let (min_size, max_feerate) = feerate_points.first_key_value().expect("not empty");
+    let (max_size, min_feerate) = feerate_points.last_key_value().expect("not empty");
+    let feerate = match in_top_x_mb {
+        _ if in_top_x_mb <= *min_size => *max_feerate,
+        _ if in_top_x_mb >= *max_size => *min_feerate,
+        _ => mempool::find_interpolated_value(&feerate_points, in_top_x_mb).ok_or_else(|| {
```
iljakuklic, line 376: Looks `in_top_x_mb` is matched but not inspected in patterns. Seems to me `if..else` cascade would be more idiomatic here. If you still want to use match for some reason (it kind of is more compact with less syntactic noise), just match over the unit `()`.

iljakuklic (approved):

api-server/web-server/src/api/v1.rs
```diff
@@ -315,8 +317,77 @@ async fn get_transaction(
...
+        .get(IN_TOP_X_MB)
+        .map(|str| usize::from_str(str))
+        .transpose()
+        .map_err(|_| ApiServerWebServerClientError::InvalidInTopX)?
+        .unwrap_or(DEFAULT_IN_TOP_X_MB);
+
+    let feerate_points = &state.cached_values.feerate_points;
+
+    let feerate_points: BTreeMap<_, _> = {
+        let current_time = get_time();
```
TheQuantumPhysicist, line 337: Please use TimeGetter, and don't use direct time access in production code. Then to test this feature, you can simulate time. See how it's done in other crates, such as ChainState.

PR #1402, opened 2023-12-14
Add estimate_transaction_size function to wasm crate

OBorce:


common/src/size_estimation/mod.rs
TheQuantumPhysicist, line 1: There are already size estimation tools in the mempool

common/src/size_estimation/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+pub fn tx_size_with_outputs(outputs: &[TxOutput]) -> usize {
+    let tx = SignedTransaction::new(
+        Transaction::new(1, vec![], outputs.into()).expect("should not fail"),
+        vec![],
+    )
+    .expect("should not fail");
+    serialization::Encode::encoded_size(&tx)
+}
+
+pub fn get_tx_output_destination(txo: &TxOutput) -> Option<&Destination> {
```
TheQuantumPhysicist, line 49: This function is basically playing with fire. Please remove it.

I understand it was in the wallet, but maybe in the wallet for that use-case it's fine. Putting it in common basically endangers people using this for general purpose stuff. This is the same problem we had before in TxOutput where we had generic functions like `destination()`, and no good came out of that. It only lead to confusion similar to that of polymorphism because things look alike, but they aren't.

TheQuantumPhysicist: You can make this function private, btw. That makes it OK since it's used only here. For the wallet, write another one (even a copy is OK, given that these have different purposes).

OBorce: changed to private

common/src/size_estimation/mod.rs
OBorce: I can see a memory usage estimator but I think that is for a different usage

common/src/size_estimation/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+/// Return the encoded size of an input signature
+pub fn input_signature_size_from_destination(
+    destination: &Destination,
+) -> Result<usize, SizeEstimationError> {
+    // Sizes calculated upfront
+    match destination {
+        Destination::Address(_) => Ok(103),
+        Destination::PublicKey(_) => Ok(69),
+        Destination::AnyoneCanSpend => Ok(2),
```
iljakuklic, line 38: Could you calculate size from the actual types (using `encoded_size`) by creating a dummy value with some dummy data? Should be more reliable than hardcoding numbers.

wasm-crypto/src/lib.rs
```diff
@@ -475,6 +476,38 @@ pub fn encode_input_for_account_outpoint(
...
+    let mut tx_outputs = vec![];
+    while !outputs.is_empty() {
+        let output = TxOutput::decode(&mut outputs).map_err(|_| Error::InvalidOutput)?;
+        tx_outputs.push(output);
+    }
+
+    let size = tx_size_with_outputs(&tx_outputs);
+    let inputs_size = inputs.len();
+
+    let mut total_size = size + inputs_size;
```
iljakuklic, line 494: Does inputs include the size of the input vec (= the number of inputs) at the beginning?

wasm-crypto/src/lib.rs
```diff
@@ -475,6 +476,38 @@ pub fn encode_input_for_account_outpoint(
...
+
+    while !opt_utxos.is_empty() {
+        let utxo = Option::<TxOutput>::decode(&mut opt_utxos).map_err(|_| Error::InvalidInput)?;
```
iljakuklic, line 497: This also misses the encoded length of the witness vector.Not sure what sort of precision is needed here. However for example in case of fee estamation, it may not be good to underestimate size if blocks are close to empty and the user wants to pay the minimum relay feerate.

common/src/size_estimation/mod.rs
iljakuklic: The mempool size tracker measures in-memory usage, not encoded size.

iljakuklic (approved):

wasm-crypto/src/lib.rs
OBorce: had me worried for a second why is the test passing if the vector length is not included,
but it is included in the `tx_size_with_outputs` function, it creates an empty signed transaction so all of the vectors are included there, inputs and witnesses are just later added.
The nodejs test checks that the estimated size is the same as the real encoded signed transaction

common/src/size_estimation/mod.rs
OBorce: added helper functions to create dummy witnesses to calculate the encoded size

common/src/size_estimation/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+/// Return the encoded size of an input signature
+pub fn input_signature_size(txo: &TxOutput) -> Result<usize, SizeEstimationError> {
+    get_tx_output_destination(txo).map_or(Ok(0), input_signature_size_from_destination)
+}
+
+fn no_signature_size() -> usize {
+    InputWitness::NoSignature(None).encoded_size()
+}
+
+fn publick_key_signature_size() -> usize {
```
TheQuantumPhysicist, line 47: public

iljakuklic:
There seems to be a fair amount of manual size calculations which may be somewhat error prone. I know it's a bit less efficient but how about constructing the whole transaction by plugging in the dummy data (for signatures and such) and just asking about `encoded_size` of the whole thing?

PR #1404, opened 2023-12-14
Make local socket address available in the peer manager

ImplOfAnImpl:
This is needed for the other PR, related to AddrListResponse caching (https://github.com/mintlayer/mintlayer-core/pull/1403)

I also did some renaming and minor cleanup.


TheQuantumPhysicist (approved):
I do hope this is worth it... seems to complicate the transport layer a lot. I don't have particular complaints.

iljakuklic (approved):

PR #1411, opened 2023-12-20
Change pool and delegation IDs in the DB to addressable strings

OBorce:


iljakuklic:
What is the reason for this change? From certain angle, it looks like a step back, changing strong types for strings.

OBorce:
Mainly it started when I was debugging some pool info, and it was very hard to query the DB directly when everything was bytea. Also I was under a wrong assumption I am not sure why I thought that you couldn't create an index on a bytea column, but I was wrong it seems.

For the strongly typed functions I can change them to Address<PoolId> and Address<DelegationId> ? 
Or do you think we should just drop this altogether.

iljakuklic:
> Mainly it started when I was debugging some pool info, and it was very hard to query the DB directly when everything was bytea. Also I was under a wrong assumption I am not sure why I thought that you couldn't create an index on a bytea column, but I was wrong it seems.

Hmm yea I can imagine debugging with binary data could be a pain. Not convinced it's a good idea to change data representation for this reason alone. Keep in mind that an encoded string is also a bit bigger than binary. 60% extra for mapping chars to bytes plus some more for hrp and checksum. On the other hand it becomes more self-explanatory so it's not 100% clear cut.

> For the strongly typed functions I can change them to `Address<PoolId>` and `Address<DelegationId>` ?

I'd say so if the representation ends up being a string. Either way, convert the raw data (binary blob or string) into properly typed objects as soon as practically possible.

TheQuantumPhysicist (approved):

PR #1412, opened 2023-12-21
Add tx id and pos target to API Server block endpoint response

OBorce:
Add more info in the endpoint responses,
added transaction ID in any transaction endpoint and PoS target in the block header response

### DIFF
--- a/api-server/web-server/src/api/json_helpers.rs
+++ b/api-server/web-server/src/api/json_helpers.rs
@@ -15,8 +15,10 @@
 
 use common::{
     address::Address,
-    chain::{output_value::OutputValue, ChainConfig, Transaction, TxOutput},
-    primitives::Amount,
+    chain::{
+        block::ConsensusData, output_value::OutputValue, Block, ChainConfig, Transaction, TxOutput,
+    },
+    primitives::{Amount, Idable},
 };
 use serde_json::json;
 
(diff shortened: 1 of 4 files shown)


api-server/web-server/src/api/json_helpers.rs
```diff
@@ -148,3 +153,22 @@ pub fn tx_to_json(tx: &Transaction, chain_config: &ChainConfig) -> serde_json::V
...
 }
+
+pub fn block_header_to_json(block: &Block) -> Result<serde_json::Value, ApiServerWebServerError> {
+    let target = match block.header().header().consensus_data() {
+        ConsensusData::PoS(pos) => pos.compact_target(),
+        ConsensusData::PoW(_) | ConsensusData::None => {
+            return Err(ApiServerWebServerError::ServerError(
+                ApiServerWebServerServerError::InternalServerError,
+            ));
+        }
```
iljakuklic, line 164: This effectively disables the whole `block` call for `PoW` and no consensus. Would it be possible to return a dummy value (json `null`) or just omit the target field completely in that case?

TheQuantumPhysicist: Or write their json values... why not?

TheQuantumPhysicist (approved):
Besides that comment from Lukas, this is fine.

iljakuklic (approved):

PR #1417, opened 2023-12-28
Fix issues after upgrade to 1.75

TheQuantumPhysicist:


crypto/src/kdf/mod.rs
```diff
@@ -93,10 +93,9 @@ pub enum KdfChallenge {
     },
 }
 
-fn make_salt<R: Rng + CryptoRng>(rng: &mut R, len: NonZeroUsize) -> Result<Vec<u8>, KdfError> {
-    let len = len.try_into().map_err(|_| KdfError::InvalidSaltSize)?;
-    let salt: Vec<u8> = (0..len).map(|_| rng.gen::<u8>()).collect();
-    Ok(salt)
+fn make_salt<R: Rng + CryptoRng>(rng: &mut R, len: NonZeroUsize) -> Vec<u8> {
+    let len = len.get();
+    (0..len).map(|_| rng.gen::<u8>()).collect()
```
ImplOfAnImpl, line 98: It seems like KdfError::InvalidSaltSize is no longer used, so it can be removed.

ImplOfAnImpl (approved):

PR #1422, opened 2024-01-03
Include outbound manual connections in address group

azarovh:
> Currently when establishing a new automatic outbound connection we ignore addresses in address groups that already have an automatic outbound connection.
But should we also ignore those that have manual outbound connections?

### DIFF
--- a/p2p/src/peer_manager/tests/connections.rs
+++ b/p2p/src/peer_manager/tests/connections.rs
@@ -49,8 +49,9 @@ use crate::{
     },
     testing_utils::{
         connect_and_accept_services, connect_services, get_connectivity_event,
-        peerdb_inmemory_store, test_p2p_config, TestTransportChannel, TestTransportMaker,
-        TestTransportNoise, TestTransportTcp, TEST_PROTOCOL_VERSION,
+        make_transport_with_local_addr_in_group, peerdb_inmemory_store, test_p2p_config,
+        TestTransportChannel, TestTransportMaker, TestTransportNoise, TestTransportTcp,
+        TEST_PROTOCOL_VERSION,
     },
     types::peer_id::PeerId,
     utils::oneshot_nofail,
@@ -907,7 +908,8 @@ async fn connection_reserved_node_channel() {
 
 // Verify that peers announce own addresses and are discovered by other peers.
 // All listening addresses are discovered and multiple connections are made.
-async fn discovered_node<A, T>()
+// All peers are in the same address group
+async fn discovered_node_same_address_group<A, T>()
 where
     A: TestTransportMaker<Transport = T::Transport>,
     T: NetworkingService + 'static + std::fmt::Debug,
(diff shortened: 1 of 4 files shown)


p2p/src/peer_manager/tests/connections.rs
```diff
@@ -1065,19 +1068,326 @@ where
...
-    discovered_node::<TestTransportChannel, DefaultNetworkingService<MpscChannelTransport>>().await;
+    discovered_node_same_address_group::<
+        TestTransportChannel,
+        DefaultNetworkingService<MpscChannelTransport>,
+    >()
+    .await;
+}
+
+// Create 3 peers and make one of them a reserved node.
+// Put reserved node in a separate address groupe.
```
ImplOfAnImpl, line 1099: Typo - groupe

p2p/src/peer_manager/tests/connections.rs
```diff
@@ -1040,15 +1037,21 @@ where
...
-        //   (each peer can make only one outbound connection because all discovered addresses are in the same address group).
-        // - 3 inbound connections to the discovered addresses.
+        // - 1 outbound and 2 inbound connections to/from reserved peer.
+        // - 1 outbound and 1 inbound connections from one of the peers to reserved.
+        // - 1 outbound connections to the reserved peer.
+        // No connections are made to discovered peers because they are in the same group and are already
+        // covered by reserved connections which are treated as manual.
         // Since outbound connections are random, we don't know which peer will get 4 connections.
-        if counts == [3, 3, 4] || counts == [3, 4, 3] || counts == [4, 3, 3] {
+        if counts == [1, 2, 3] {
```
ImplOfAnImpl, line 1054: Outdated comment - you can't have 4 connections here.

p2p/src/peer_manager/tests/connections.rs
```diff
@@ -1065,19 +1068,326 @@ where
...
+            get_connected_peers(&peer_mgr_event_sender1),
+            get_connected_peers(&peer_mgr_event_sender2),
+            get_connected_peers(&peer_mgr_event_sender3)
+        );
+        let counts = [connected_peers.0.len(), connected_peers.1.len(), connected_peers.2.len()];
+
+        // There should be:
+        // - 2 outbound and 2 inbound connections to/from reserved peers.
+        // - 3 outbound connections to the discovered addresses
+        //   (each peer can make only one outbound connection because all discovered addresses are in the same address group).
```
ImplOfAnImpl, line 1232: This part didn't seem to make much sense originally and now it looks wrong.

In any case, I'd just remove this test, instead of trying to clean it up, because IMO it's confusing and rather useless. In fact, I think we only need the `discovered_node_separate_groups` one (which can be named just `node_discovery`).

ImplOfAnImpl (approved):

TheQuantumPhysicist (approved):

p2p/src/peer_manager/tests/connections.rs
azarovh: Isn't it usefull to have tests that show that the number of connection is different if some nodes belong to the same group?

ImplOfAnImpl: Well, in general it does seem useful, but it also looks redundant in these tests, whose purpose is to check address discovery. On the other hand, we don't seem to have dedicated tests that would check this address-group-related logic.
If you decide to keep it, plz rephrase the comment a little, so that it's consistent with the similar comments in other tests.

azarovh: I'll leave them and update the comment

PR #1426, opened 2024-01-04
P2p: don't query the dns seed eagerly at the startup

ImplOfAnImpl:
Currently, we always perform one dns seed query after starting up. This was done at some point mainly to improve the start up times of nodes created in `build-tools/p2p-v2-test`, which is used to test p2p connectivity after a protocol upgrade. The "test" is basically a docker-compose script that starts a bunch of nodes, passing their addresses to each other via the `--p2p-boot-node` option; they are supposed to connect to each other as well as to nodes on the public network. Without the early dns query, those nodes would linger for up to 30 mins on the first start without being able to connect to the network, until the "stale tip" interval would have passed and the "normal" dns seed query would be performed.

But in the general case querying dns seeds without real necessity is not a good idea, so here I get rid of this behaviour. Still, it's better for `build-tools/p2p-v2-test` to be able to start fast; for this, I added a separate command line switch `--p2p-boot-nodes-will-stall`, which basically causes the node to perform an early dns seed query if the only peer addresses that it knows about are those specified via `boot_node`.

I also did minor cleanup here and there and renamed `p2p-v2-test` to just `p2p-test`.


azarovh (approved):

node-lib/src/config_files/p2p.rs
```diff
@@ -66,6 +66,13 @@ pub struct P2pConfigFile {
     pub disable_noise: Option<bool>,
     /// Optional list of boot node addresses to connect.
     pub boot_nodes: Option<Vec<IpOrSocketAddress>>,
+    /// If true, the node will assume that the nodes specified in `boot_nodes` won't be able
+    /// to provide it with fresh blocks.
+    /// This is used by p2p-test, where `boot_nodes` are other test nodes
+    /// that have no knowledge about nodes on the public network. Setting this to
+    /// true will force the node to make a dns seed query early if its peerdb only contains
+    /// addresses specified via `boot_nodes`.
+    pub boot_nodes_will_stall: Option<bool>,
```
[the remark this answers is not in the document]
ImplOfAnImpl: Well, `force_dns_seed_query` doesn't reflect what will happen exactly, because the query is not unconditional - it's performed only if peerdb has no addresses other than those specified in `boot_nodes`.

iljakuklic (approved):
Did not look into the docker stuff since I don't understand it well enough

node-lib/src/options.rs
```diff
@@ -124,6 +124,13 @@ pub struct RunOptions {
     #[clap(long, value_name = "NODE")]
     pub p2p_boot_node: Option<Vec<IpOrSocketAddress>>,
 
+    /// If true, the node will assume that the nodes specified via `boot_node` won't be able
+    /// to provide it with fresh blocks; this will cause it to make a dns seed query earlier than
+    /// usual if the only nodes it knows about are those specified via `boot_node`.
+    #[clap(long)]
+    #[arg(hide = true)]
+    pub p2p_boot_nodes_will_stall: Option<bool>,
```
TheQuantumPhysicist, line 132: I remember making a comment about this (even though I saw Gosha's comment). This option should have a better name. `force_dns_seed_query_if_none_found` is much better than this. The current name chosen is very confusing and doesn't represent the result.

TheQuantumPhysicist (approved):
Please consider finding a better name for that parameter.

node-lib/src/options.rs
ImplOfAnImpl: I decided to also change the logic behind the option, so that it has nothing to do with boot nodes. The option is now named `p2p_force_dns_query_if_no_global_addresses_known` and the dns query is forced if the peerdb has no global addresses at startup (where "global" addresses are those for which `is_global_unicast_ip` returns true).

TheQuantumPhysicist (approved):

PR #1434, opened 2024-01-06
Introduce `DecimalAmount` and refactor wallet controller to use it

iljakuklic:
* Add `DecimalAmount` which is like `Amount` with formatting information (number of decimals)
  *  it can be printed and parsed to/from strings and supports serde
  * The printing/parsing of `Amount` was changed to use it under the hood
* Use that in the wallet controller for querying balances
* That will enable removing some duplication between the CLI frontend and RPC frontend

### DIFF
--- a/wallet/wallet-controller/src/read.rs
+++ b/wallet/wallet-controller/src/read.rs
@@ -23,7 +23,7 @@ use common::{
         ChainConfig, DelegationId, Destination, GenBlock, PoolId, Transaction, TxOutput,
         UtxoOutPoint,
     },
-    primitives::{id::WithId, Amount, Id},
+    primitives::{id::WithId, Amount, DecimalAmount, Id},
 };
 use crypto::key::hdkd::{child_number::ChildNumber, u31::U31};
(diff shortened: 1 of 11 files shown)


OBorce (approved):
looks good

wallet/wallet-controller/src/read.rs
```diff
@@ -83,6 +83,34 @@ impl<'a, T: NodeInterface> ReadOnlyController<'a, T> {
...
+                Currency::Token(token_id) => token_id,
+            };
+
+            let info = super::fetch_token_info(&self.rpc_client, token_id).await?;
+            let decimals = info.token_number_of_decimals();
+            tokens.insert(
+                token_id,
+                DecimalAmount::from_amount_minimal(amount, decimals),
+            );
+        }
```
OBorce, line 109: One other option I used in the read controller for fetching multiple pool infos from the node was to use FuturesUnordered and try_collect().await to await all of them at once instead of one by one.

TheQuantumPhysicist (approved):

PR #1435, opened 2024-01-09
No announcements for banned peers

azarovh:
Here is the comparison of our approach and bitcoin's.

bitcoin’s discouragement:
- applies until shutdown
- allow incoming connections if not full
- no outgoing connections
- preferred for eviction
- no gossiping

bitcoins banning:
- can be set only manuallly
- no incoming/outgoing connections
- no gossiping
- immediate disconnect

mintlayer:
- applies for 30mins
- no incoming/outgoing connections
- immediate disconnect

Obviously that out approach currently lacks "no gossiping", i.e. no announcements for banned peers and no such addresses in address list response.

Open question is whether we want to prefer banned peers for eviction. It will be effectively applied after it comes out of ban. But I'm not sure about it.

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -449,7 +449,9 @@ where
     /// `peer_id` must be from the connected peer.
     fn announce_address(&mut self, peer_id: PeerId, address: SocketAddress) {
         let peer = self.peers.get_mut(&peer_id).expect("peer must be known");
-        if !peer.announced_addresses.contains(&address) {
+        if !peer.announced_addresses.contains(&address)
+            && !self.peerdb.is_address_banned(&address.as_bannable())
+        {
             Self::send_peer_message(
                 &mut self.peer_connectivity_handle,
                 peer_id,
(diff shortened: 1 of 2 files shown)


p2p/src/peer_manager/mod.rs
```diff
@@ -1370,6 +1372,7 @@ where
             .get_or_create(peer, now, || {
                 self.peerdb
                     .known_addresses()
+                    .filter(|address| !self.peerdb.is_address_banned(&(*address).as_bannable()))
```
azarovh, line 1375: I wasn't sure if this filter should be applied inside `known_addresses` function.

TheQuantumPhysicist: I don't think this is implemented correctly anyway. The `is_address_banned` function doesn't take into account whether the banning has expired.

iljakuklic: A comment clarifying this in the code would be nice.

TheQuantumPhysicist: I agree. It's difficult to figure this out based on a state.

TheQuantumPhysicist:
Closed in favor of #1451

PR #1437, opened 2024-01-09
Change the standard, by which wallet commands are written by having them grouped by prefix

TheQuantumPhysicist:
Please check the logic and the function names

### DIFF
--- a/wallet/wallet-cli-lib/src/commands/mod.rs
+++ b/wallet/wallet-cli-lib/src/commands/mod.rs
@@ -63,6 +63,31 @@ pub enum WalletCommand {
     #[clap(name = "account-select")]
     SelectAccount { account_index: U31 },
 
+    /// Lists all the utxos owned by this account
+    #[clap(name = "account-utxos")]
+    ListUtxo {
+        /// The type of utxo to be listed. Default is "all".
+        #[arg(value_enum, default_value_t = CliUtxoTypes::All)]
+        utxo_type: CliUtxoTypes,
+        /// Whether to include locked outputs. Default is "unlocked"
+        #[arg(value_enum, default_value_t = CliWithLocked::Unlocked)]
+        with_locked: CliWithLocked,
+        /// The state of the utxos; e.g., confirmed, unconfirmed, etc.
+        #[arg(default_values_t = vec![CliUtxoState::Confirmed])]
+        utxo_states: Vec<CliUtxoState>,
+    },
+
+    /// Get the total balance in the selected account in this wallet. See available options to include more categories, like locked coins.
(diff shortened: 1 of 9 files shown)


erubboli (approved):
Tested ACK

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -51,368 +51,507 @@ use self::helper_types::{
...
+        amount: String,
     },
 
-    /// Rescan
-    Rescan,
-
-    SyncWallet,
-
+    /// Get the total balance in this wallet. See available options to include more categories, like locked coins.
+    #[clap(name = "address-balance")]
```
iljakuklic, line 174: I am not convinced by the `address-` prefix. If I am not mistaken, it adds up the balance of all addresses in the account. What's the idea here? Maybe `account-` works better? It could also be applied to the utxo listing command.

iljakuklic (approved):

wallet/wallet-cli-lib/src/commands/mod.rs
iljakuklic: How about something stupid: `subwallet-`? :laughing: Trying to come up with a name for the set of addresses generated by fixing a particular account index in the derivation path but too many things are already called `account`...

TheQuantumPhysicist: Consider that for usability "subwallet" doesn't really mean anything... I would rather make it just "balance" as an exception.

TheQuantumPhysicist: I will keep this for now and maybe we can change it later if we find something better. Let's see whether people can suggest something better.

iljakuklic: > Trying to come up with a name for the set of addresses generated by fixing a particular account index in the derivation path but too many things are already called `account`...

I see there already is `account-create` and `account-select` where `account-` is used in this way. From that point of view, `account-balance` and `account-utxos` sound pretty sensible. However, there is still an ambiguity with "account" used for staking. Or do we just use the word address (perhaps with some modifier/prefix) for that?

TheQuantumPhysicist: I can live with `account-balance`, and cast out staking. 

Besides the `wallet-` commands, almost all commands have to do with the "selected account". This is why it's difficult to include staking in `account-`. The account is selected during the whole session. So `account-` doesn't really mean that the command has to do with the current account, but it means something like "account control", where accounts can be switched.

PR #1440, opened 2024-01-10
Split `pos_accounting::PoolData::pledge_amount` into pledge and reward

azarovh:
Previously `pledge_amount` represented original pledge + accumulated reward. This PR splits it into 2 distinct fields to make it possible in the future to spend reward without decommissioning the pool.

Note that this PR contains breaking change for api-server database.


api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -546,40 +544,42 @@ async fn update_tables_from_consensus_data<T: ApiServerStorageWrite>(
...
-                .await;
+                db_tx
+                    .set_address_balance_at_height(
+                        address.get(),
+                        *updated_delegation.balance(),
+                        CoinOrTokenId::Coin,
+                        block_height,
+                    )
+                    .await
+                    .expect("Unable to update balance");
```
OBorce, line 565: I think this is a mistake, you can't set the address balance to only the delegation balance, in the case that the same address has been reused for other things in the past.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -546,40 +544,42 @@ async fn update_tables_from_consensus_data<T: ApiServerStorageWrite>(
...
-            .await;
+            db_tx
+                .set_address_balance_at_height(
+                    address.get(),
+                    pool_data.staker_balance().expect("no overflow"),
+                    CoinOrTokenId::Coin,
+                    block_height,
+                )
+                .await
+                .expect("Unable to update balance");
```
OBorce, line 582: same here

api-server/scanner-lib/src/blockchain_state/mod.rs
azarovh: oh right, so the address is not unique and user can reuse it for multiple delegation? Interesting, in that case I indeed misunderstood the logic.

OBorce: yep, they can reuse the address for anything, normally it shouldn't happen if they are using the wallet but can't be sure what the user will do

azarovh: reverted

api-server/scanner-lib/src/blockchain_state/adapter.rs
```diff
@@ -83,8 +77,8 @@ impl PoSAccountingView for PoSAdapter {
         Ok(self.pools.get(&pool_id).cloned())
     }
 
-    fn get_pool_balance(&self, pool_id: PoolId) -> Result<Option<Amount>, Self::Error> {
-        Ok(self.pools.get(&pool_id).map(|data| data.pledge_amount()))
+    fn get_pool_balance(&self, _pool_id: PoolId) -> Result<Option<Amount>, Self::Error> {
+        unimplemented!()
```
TheQuantumPhysicist, line 81: Why?

azarovh: The implementation was incorrect. Pool balance is pledge + all delegation which is no accounted here. But since this function is never used I just removed the implementation

OBorce (approved):

PR #1444, opened 2024-01-11
Bunch of small api-server/node fixes

azarovh:
- if api-server db version mismatch clean up the db
- api-server db should have default "mintlayer-{CHAIN_TYPE}" name
- if node-lib was provided with the `clean-data` argument it should exit after that
- software version should be printed on startup

### DIFF
--- a/api-server/scanner-daemon/src/main.rs
+++ b/api-server/scanner-daemon/src/main.rs
@@ -17,7 +17,7 @@ use std::sync::Arc;
 
 use api_blockchain_scanner_lib::blockchain_state::BlockchainState;
 use api_server_common::storage::{
-    impls::postgres::TransactionalApiServerPostgresStorage,
+    impls::{postgres::TransactionalApiServerPostgresStorage, CURRENT_STORAGE_VERSION},
     storage_api::{
         ApiServerStorage, ApiServerStorageError, ApiServerStorageRead, ApiServerStorageWrite,
         ApiServerTransactionRw,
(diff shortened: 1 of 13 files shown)


api-server/scanner-daemon/src/main.rs
```diff
@@ -88,6 +88,19 @@ pub async fn run<S: ApiServerStorage>(
...
+                .get_storage_version()
+                .await
+                .unwrap_or_else(|e| panic!("Storage version read failed {}", e))
+                .expect("cannot be empty");
+
+            if storage_version != CURRENT_STORAGE_VERSION {
+                db_tx
+                    .reinitialize_storage(chain_config)
+                    .await
+                    .unwrap_or_else(|e| panic!("Storage re-initialization failed {}", e));
```
iljakuklic, line 101: Can we do better than panic here? Preferably, panic would be used for internal invariant violations. But here it seems it can happen due to external factors (e.g. disk full).

iljakuklic (approved):

api-server/scanner-daemon/src/main.rs
azarovh: I see that we don't care panicing or expecting everywhere across api-server

TheQuantumPhysicist (approved):

api-server/scanner-daemon/src/main.rs
iljakuklic: Ok, I guess it may be fine here because storage migration is a sort of one-off event.

PR #1448, opened 2024-01-14
Decommission pool request

azarovh:
Adds functionality to partially sign transaction to be able to pass it to a cold wallet and sign there.

The process of decommissioning is as follows:
- create a pool from hot wallet with decommission key from cold wallet
- call `staking-decommission-pool-request` from hot wallet
- copy the output and paste it to `node-sign-raw-transaction` command in cold wallet
- copy the output and paste it to `node-submit-transaction` command in hot wallet to broadcast

TODOs:

- [x] generate qr for output
- [x] decommission key in create pool command should not be optional
- [x] make decommission key in create pool a destination with an address, instead of a public key

### DIFF
--- a/chainstate/test-suite/src/tests/tx_fee.rs
+++ b/chainstate/test-suite/src/tests/tx_fee.rs
@@ -428,7 +428,7 @@ fn fee_from_decommissioning_stake_pool(#[case] seed: Seed) {
             required_maturity_distance.to_int()..(required_maturity_distance.to_int() * 2),
         );
 
-        let decomission_outputs =
+        let decommission_outputs =
             test_utils::split_value(&mut rng, amount_to_stake.into_atoms() / 2)
                 .into_iter()
                 .map(|value| {
@@ -447,7 +447,7 @@ fn fee_from_decommissioning_stake_pool(#[case] seed: Seed) {
                 TxInput::from_utxo(stake_pool_tx_id.into(), 0),
                 empty_witness(&mut rng),
             )
-            .with_outputs(decomission_outputs)
+            .with_outputs(decommission_outputs)
             .build();
 
         let mut verifier = TransactionVerifier::new(&storage, &chain_config);
--- a/test/functional/test_framework/wallet_cli_controller.py
+++ b/test/functional/test_framework/wallet_cli_controller.py
@@ -238,8 +238,8 @@ class WalletCliController:
                                 amount: int,
                                 cost_per_block: int,
                                 margin_ratio_per_thousand: float,
-                                decommission_key: Optional[str] = '') -> str:
-        return await self._write_command(f"staking-create-pool {amount} {cost_per_block} {margin_ratio_per_thousand} {decommission_key}\n")
+                                decommission_addr: Optional[str] = '') -> str:
+        return await self._write_command(f"staking-create-pool {amount} {cost_per_block} {margin_ratio_per_thousand} {decommission_addr}\n")
 
     async def decommission_stake_pool(self, pool_id: str) -> str:
         return await self._write_command(f"staking-decommission-pool {pool_id}\n")
--- a/test/functional/test_framework/wallet_cli_controller.py
+++ b/test/functional/test_framework/wallet_cli_controller.py
@@ -240,6 +240,15 @@ class WalletCliController:
     async def decommission_stake_pool(self, pool_id: str) -> str:
         return await self._write_command(f"staking-decommission-pool {pool_id}\n")
 
+    async def decommission_stake_pool_request(self, pool_id: str) -> str:
+        return await self._write_command(f"staking-decommission-pool-request {pool_id}\n")
+
+    async def sign_raw_transaction(self, transaction: str) -> str:
+        return await self._write_command(f"node-sign-raw-transaction {transaction}\n")
+
+    async def submit_transaction(self, transaction: str) -> str:
+        return await self._write_command(f"node-submit-transaction {transaction}\n")
+
     async def list_pool_ids(self) -> List[PoolData]:
         output = await self._write_command("staking-list-pool-ids\n")
         pattern = r'Pool Id: ([a-zA-Z0-9]+), Balance: (\d+),'
--- a/test/functional/test_framework/wallet_cli_controller.py
+++ b/test/functional/test_framework/wallet_cli_controller.py
@@ -106,9 +106,13 @@ class WalletCliController:
         await self.process.stdin.drain()
         return await self._read_available_output()
 
-    async def create_wallet(self) -> str:
-        wallet_file = os.path.join(self.node.datadir, "wallet")
+    async def create_wallet(self, name: str = "wallet") -> str:
+        wallet_file = os.path.join(self.node.datadir, name)
         return await self._write_command(f"wallet-create {wallet_file} store-seed-phrase\n")
+
+    async def open_wallet(self, name: str) -> str:
+        wallet_file = os.path.join(self.node.datadir, name)
+        return await self._write_command(f"wallet-open {wallet_file}\n")
 
     async def recover_wallet(self, mnemonic: str) -> str:
         wallet_file = os.path.join(self.node.datadir, "recovered_wallet")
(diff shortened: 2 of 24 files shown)


chainstate/test-suite/src/tests/tx_fee.rs
```diff
@@ -437,7 +437,7 @@ fn fee_from_decommissioning_stake_pool(#[case] seed: Seed) {
             required_maturity_distance.to_int()..(required_maturity_distance.to_int() * 2),
         );
 
-        let decomission_outputs =
+        let decommission_outputs =
```
TheQuantumPhysicist, line 440: To be fair, grepping for "decomission" shows many typos

test/functional/test_framework/wallet_cli_controller.py
```diff
@@ -240,6 +240,15 @@ async def create_stake_pool(self,
     async def decommission_stake_pool(self, pool_id: str) -> str:
         return await self._write_command(f"staking-decommission-pool {pool_id}\n")
 
+    async def decommission_stake_pool_request(self, pool_id: str) -> str:
+        return await self._write_command(f"staking-decommission-pool-request {pool_id}\n")
+
+    async def sign_raw_transaction(self, transaction: str) -> str:
+        return await self._write_command(f"node-sign-raw-transaction {transaction}\n")
```
TheQuantumPhysicist, line 247: Signing a transaction isn't a function for the node. We can put it under "account"?

wallet/wallet-controller/src/synced_controller.rs
```diff
@@ -580,6 +581,32 @@ impl<'a, T: NodeInterface, W: WalletEvents> SyncedController<'a, T, W> {
         .await
     }
 
+    pub async fn decommission_stake_pool_request(
+        &mut self,
+        pool_id: PoolId,
+    ) -> Result<HexEncoded<PartiallySignedTransaction>, ControllerError<T>> {
```
TheQuantumPhysicist, line 587: The wallet controller should not return HexEncoded stuff. HexEncoding should happen at the UI level. Not every system that uses the controller needs hex encoding, this is a UI thing.

wallet/wallet-controller/src/synced_controller.rs
```diff
@@ -593,6 +620,16 @@ impl<'a, T: NodeInterface, W: WalletEvents> SyncedController<'a, T, W> {
         Ok(())
     }
 
+    pub fn sign_raw_transaction(
+        &mut self,
+        tx: PartiallySignedTransaction,
+    ) -> Result<HexEncoded<SignedTransaction>, ControllerError<T>> {
```
TheQuantumPhysicist, line 626: Same here.

TheQuantumPhysicist (commented):
Looking good so far besides the comment(s). 

You don't need to care for the GUI, btw.

test/functional/wallet_decommission_request.py
```diff
@@ -0,0 +1,323 @@
...
+            # create decommission request
+            decommission_req_output = await wallet.decommission_stake_pool_request(pools[0].pool_id)
+            decommission_req = decommission_req_output.split('\n')[1]
+
+            # switch accounts and sign decommission request
+            assert_in("Success", await wallet.select_account(1))
+            decommission_signed_tx_output = await wallet.sign_raw_transaction(decommission_req)
+            decommission_signed_tx = decommission_signed_tx_output.split('\n')[1]
+            assert_in("The transaction was submitted successfully", await wallet.submit_transaction(decommission_signed_tx))
+
```
TheQuantumPhysicist, line 314: We need a test that creates the decommission request, signs it in another wallet, and submits it in the first wallet.

OBorce: as the test is using 2 accounts it should be similar to using 2 wallets, as each account has access to different private keys.

Would be good to add an assert that trying to sign the decommission request transaction with the other account will fail.

azarovh: the problem is that I'm not sure if python test framework allows to run several wallets. And what problems should be solved to do that. So I decided to write tests using accounts which proves the point but is much faster to do.

TheQuantumPhysicist: I can live with one wallet for the reasons Boris explained... but more robust testing is needed to test the counter case too, and achieve the fully signed transaction gradually, and ensure it works.

OBorce: If you want to create 2 wallets you can do it with just adding
       ` async with WalletCliController(node, self.config, self.log) as wallet, WalletCliController(node, self.config, self.log) as wallet2:`
       
With just 1 change to the `create_wallet(self)` function needs to get a wallet_file as currently it always creates it with the same name, so that both wallets can have different files, and it will work fine

TheQuantumPhysicist:
I added a missing point in the description

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -325,9 +335,7 @@ pub enum WalletCommand {
         margin_ratio_per_thousand: String,
 
         /// The key that can decommission the pool. It's recommended to keep the decommission key in a cold storage.
-        /// If not provided, the selected account in this wallet will control both decommission and staking.
-        /// This is NOT RECOMMENDED.
-        decommission_key: Option<HexEncoded<PublicKey>>,
+        decommission_key: HexEncoded<Destination>,
```
TheQuantumPhysicist, line 338: This is wrong. This should not be a hex-encoded destination. This should be an address.

OBorce, line 338: Destination is usually passed as the Address<Destination> string representation

TheQuantumPhysicist: Remember that destinations are `Addressable`.

OBorce (approved):
Looks good.

test/functional/wallet_decommission_request.py
```diff
@@ -0,0 +1,334 @@
...
+    async def async_test(self):
+        node = self.nodes[0]
+        decommission_pub_key_encoded = ""
+
+        async with WalletCliController(node, self.config, self.log, chain_config_args=["--chain-pos-netupgrades", "true"]) as wallet:
+            # new cold wallet
+            await wallet.create_wallet("cold_wallet")
+
+            decommission_pub_key_bytes = await wallet.new_public_key()
```
TheQuantumPhysicist, line 245: Please use addresses. No need to use public keys anymore. This function will be soon removed, btw.

wallet/src/wallet/mod.rs
```diff
@@ -182,6 +183,16 @@ pub enum WalletError {
     ReducedLookaheadSize(u32, u32),
     #[error("Wallet file {0} error: {1}")]
     WalletFileError(PathBuf, String),
+    #[error("Missing inputs signatures in a tx")]
+    InputsSignatureAreMissing,
```
TheQuantumPhysicist, line 187: This seems to be not used anymore?

wallet/src/wallet/mod.rs
```diff
@@ -182,6 +183,16 @@ pub enum WalletError {
...
+    #[error("Cannot use partially signed transaction in a decommission command")]
```
TheQuantumPhysicist, line 188: We need to deliver clearer information to the (noob) user to understand what's going on here. Basically the problem here is that we're trying to decommission the pool, but the wallet doesn't have all the required keys to fully sign the transaction. We could simply opt out to changing this error message to something like "Failed to completely sign the decommission transaction. This wallet does not seem to have the decommission key. Consider using a decommission-request, and passing it to the wallet that has the decommission key".

test/functional/wallet_decommission_request.py
```diff
@@ -0,0 +1,334 @@
...
+            decommission_signed_tx_output = await wallet.sign_raw_transaction(decommission_req)
+            decommission_signed_tx = decommission_signed_tx_output.split('\n')[1]
+
+        async with WalletCliController(node, self.config, self.log, chain_config_args=["--chain-pos-netupgrades", "true"]) as wallet:
+            # open hot wallet
+            await wallet.open_wallet("hot_wallet")
+
+            assert_in("The transaction was submitted successfully", await wallet.submit_transaction(decommission_signed_tx))
+
+            self.gen_pos_block(node.mempool_transactions(), 3)
```
TheQuantumPhysicist, line 326: Can we add a check to see if that particular transaction is included in the mempool?

OBorce (approved):

TheQuantumPhysicist (approved):

PR #1449, opened 2024-01-14
Rework staker reward distribution

azarovh:
Staker should get reward proportional to its balance.

Now if the total reward is T the process is as follows:

1. Subtract cost-per-block from T, what’s left is U
2. Split U pro-rata among delegators and staker, give the staker his amount, and what’s for delegators goes to the next step, call it X
3. Multiple X by margin ratio, give the result to the staker, and the rest is distributed among delegators, pro-rata.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
@@ -19,6 +19,7 @@ use std::collections::BTreeMap;
 use crate::error::ConnectTransactionError;
 
 use common::{
+    amount_sum,
     chain::{Block, DelegationId, PoolId, RewardDistributionVersion},
     primitives::{
         amount::UnsignedIntType as AmountUIntType, per_thousand::PerThousand, Amount, Id,
@@ -125,8 +126,8 @@ fn calculate_staker_reward_v0(
     mpt: PerThousand,
 ) -> Option<Amount> {
     let staker_reward = match total_reward - cost_per_block {
-        Some(v) => (v * mpt.value().into())
-            .and_then(|v| v / 1000)
+        Some(to_distribute) => (to_distribute * mpt.value().into())
+            .and_then(|v| v / mpt.denominator().into())
             .and_then(|v| v + cost_per_block)?,
         // if cost per block > total reward then give the reward to staker
         None => total_reward,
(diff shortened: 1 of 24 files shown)


OBorce (approved):

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -120,6 +136,51 @@ fn calculate_staker_reward(
...
+            let numer =
+                (Uint256::from_amount(v) * staker_balance).expect("Source types are smaller");
+            let pro_rata_staker_reward = (numer / pool_balance)
+                .ok_or(ConnectTransactionError::PoolBalanceIsZero(pool_id))?;
+            let pro_rata_staker_reward: AmountUIntType = pro_rata_staker_reward
+                .try_into()
+                .expect("cannot be greater than total_reward type");
+            let pro_rata_staker_reward = Amount::from_atoms(pro_rata_staker_reward);
+
+            let delegations_reward =
```
TheQuantumPhysicist, line 162: delegators_reward, because this is owned by delegators.

TheQuantumPhysicist: I think also from a type safety point of view, this should be Uint256, because then you multiply it by the margin ratio mpt, which will expand over u128. I know this is unlikely, but we should do it right anyway.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -120,6 +136,51 @@ fn calculate_staker_reward(
...
+                (v - pro_rata_staker_reward).expect("cannot be greater than total reward");
+
+            (delegations_reward * mpt.value().into())
+                .and_then(|v| v / 1000)
```
TheQuantumPhysicist, line 166: Can we get that `1000` from `v.denominator()`? I feel this is the right thing to do.

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -299,15 +361,154 @@ mod tests {
...
+        .is_ok());
+        assert!(calculate_staker_reward_v1(
+            reward,
+            pool_balance,
+            Amount::ZERO,
+            Amount::ZERO,
+            mpt,
+            pool_id
+        )
+        .is_ok());
```
TheQuantumPhysicist, line 430: Please test the expected value, not just `is_ok()`

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
```diff
@@ -120,6 +136,51 @@ fn calculate_staker_reward(
...
+    pool_balance: Amount,
+    staker_balance: Amount,
+    cost_per_block: Amount,
+    mpt: PerThousand,
+    pool_id: PoolId,
+) -> Result<Amount, ConnectTransactionError> {
+    debug_assert!(staker_balance <= pool_balance);
+
+    let staker_reward = match total_reward - cost_per_block {
+        Some(v) => {
```
iljakuklic, line 150: Not sure it's much of an improvement but I think using `let v = (total_reward - cost_per_block).unwrap_or(0);` will work without having to go into two separate code branches.

Note that there is currently a subtle difference in how the two code paths behave in case `pool_balance` is zero. An error is returned if and only if `total_reward >= cost_per_block` which seems like an irrelevant condition. Unifying the two paths will make the behaviour more consistent.

Also, have you thought about a more descriptive name for `v`?

chainstate/tx-verifier/src/transaction_verifier/reward_distribution.rs
azarovh: The idea was to check that if 0 passed to different arguments the result would be Ok/Err. The balances are random so to check exact values I'd have to duplicate the calculation. Also there are checks with arbitrary numbers below to pin the calculation.

PR #1450, opened 2024-01-14
hide decommissioned pools from responses in the API Server

OBorce:
Don't return decommissioned pools from the API Server
fix a bug where latest pools was returning the earliest data of the pool instead of the latest data.

### DIFF
--- a/api-server/api-server-common/src/storage/impls/postgres/queries.rs
+++ b/api-server/api-server-common/src/storage/impls/postgres/queries.rs
@@ -929,13 +929,13 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
         self.tx
             .query(
                 r#"
-                SELECT pool_id, data
+                SELECT sub.pool_id, data
                 FROM (
-                    SELECT pool_id, data, block_height, ROW_NUMBER() OVER(PARTITION BY pool_id ORDER BY block_height) as oldest
+                    SELECT pool_id, data, staker_balance, block_height, ROW_NUMBER() OVER(PARTITION BY pool_id ORDER BY block_height DESC) as newest
                     FROM ml_pool_data
-                ) AS sub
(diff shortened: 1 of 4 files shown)


TheQuantumPhysicist (approved):

pos_accounting/src/pool/pool_data.rs
```diff
@@ -70,6 +70,10 @@ impl PoolData {
         self.pledge_amount = Amount::ZERO;
         self
     }
+
+    pub fn is_decommissioned(&self) -> bool {
+        self.pledge_amount == Amount::ZERO
```
azarovh, line 75: after rebasing on master this has to check `staker_balance` which is pledge + rewards

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -929,13 +929,13 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
         self.tx
             .query(
                 r#"
-                SELECT pool_id, data
+                SELECT sub.pool_id, data
                 FROM (
-                    SELECT pool_id, data, block_height, ROW_NUMBER() OVER(PARTITION BY pool_id ORDER BY block_height) as oldest
+                    SELECT pool_id, data, pledge_amount, block_height, ROW_NUMBER() OVER(PARTITION BY pool_id ORDER BY block_height DESC) as newest
```
azarovh, line 934: this would probably fail anyway but `pledge_amount` column was also changed to `staker_balance`

azarovh (approved):
looks fine besides using staker balance instead of pledge

api-server/api-server-common/src/storage/impls/postgres/queries.rs
OBorce: fixed

PR #1451, opened 2024-01-15
P2p: peer discouragement

ImplOfAnImpl:
- Discouraged peers are prioritized during eviction and outbound connections are never established to discouraged addresses.
- Peers are disconnected automatically when discouraged.
- Inbound connections from discouraged addresses are allowed if the connection number limit hasn't been reached yet.
- Discouraged and banned addresses are not gossiped via `AnnounceAddrRequest` and are not included into `AddrListResponse`.
- Discouraged addresses are stored in the peer db the same way as banned ones. This is different from what bitcoin core does, where discouraged addresses are kept in a non-persistent rolling bloom filter (which means that the discouragement time is determined by the node's uptime).
- We no longer ban automatically.
- When banning a peer manually, the user has to specify the duration explicitly. The duration is parsed via `humantime` crate, so it can be something like `1y` (1 year) or `3M` (3 months).
- Printing banned or discouraged addresses in the wallet now also prints the expiration time.
- Small improvements in printing of `Time` - instead of having a separate function `as_standard_printable_time` it now has custom implementations of `Debug` and `Display`. The implementations delegate printing to `chrono::DateTime<Utc>`, just like `as_standard_printable_time`, but also handle the case when `Time` is not representable via `chrono::DateTime`.
- I also added rpc & wallet commands to print reserved peer addresses.


p2p/src/ban_config.rs
```diff
@@ -0,0 +1,54 @@
...
+// valid, the network may be split between old and new nodes, because the old ones would ban peers
+// that exhibit such behavior. Also, 30 min doesn't make much sense for manual banning. So, we
+// should probably get rid of automatic banning and return the old duration for the manual banning.
+// But if we decide to keep the auto-ban functionality:
+// a) the config names should be changed to AutoBanDuration, AutoBanThreshold;
+// b) manual banning should have a separate ManualBanDuration; or event better, the ban duration
+// should be specified by the user via RPC and the config should become ManualBanDefaultDuration.
+// c) there should be the ability to turn auto-banning off, e.g. by setting the threshold to 0;
+// by default, auto banning should be turned off.
+make_config_setting!(BanDuration, Duration, Duration::from_secs(60 * 30));
```
ImplOfAnImpl, line 35: Let's decide what should be done with the automatic banning.

node-lib/src/options.rs
```diff
@@ -136,7 +136,9 @@ pub struct RunOptions {
     #[clap(long)]
     pub p2p_max_inbound_connections: Option<usize>,
 
-    /// The p2p score threshold after which a peer is baned.
+    // TODO: add all the options related to banning/discouragement thresholds and durations,
+    // for completeness.
+    /// The p2p score threshold after which a peer is banned.
     #[clap(long)]
     pub p2p_ban_threshold: Option<u32>,
```
ImplOfAnImpl, line 144: This will be addressed once we decide what kind of banning (auto vs manual) we should keep.

p2p/src/peer_manager/config.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use super::{
+    peerdb::config::PeerDbConfig,
+    peers_eviction::{
+        OutboundBlockRelayConnectionMinAge, OutboundFullRelayConnectionMinAge,
+        PreservedInboundCountAddressGroup, PreservedInboundCountNewBlocks,
+        PreservedInboundCountNewTransactions, PreservedInboundCountPing,
+    },
+};
+
```
ImplOfAnImpl, line 28: FYI: except for the TODO about naming, there is nothing new in this file, I've just moved `PeerManagerConfig` to a separate file.

p2p/src/peer_manager/peerdb/mod.rs
```diff
@@ -534,24 +553,48 @@ impl<S: PeerDbStorage> PeerDb<S> {
...
     }
 
+    /// Checks if the given address is discouraged
+    pub fn is_address_discouraged(&self, address: &BannableAddress) -> bool {
+        self.discouraged_addresses.contains_key(address)
+    }
+
+    /// Changes the address state to discouraged
+    pub fn discourage(&mut self, address: BannableAddress) {
+        // TODO: do we need to prolong the discouragement if the peer continues misbehaving?
```
ImplOfAnImpl, line 581: Let's decide whether to apply or drop this TODO.

p2p/src/peer_manager/tests/connections.rs
```diff
@@ -79,6 +82,136 @@ use crate::{
     PeerManagerEvent,
 };
 
+async fn validate_invalid_connection<A, S>()
```
ImplOfAnImpl, line 85: FYI: I just moved this test here from `ban.rs`

p2p/src/peer_manager/tests/connections.rs
```diff
@@ -79,6 +82,136 @@ use crate::{
...
+
+#[tracing::instrument]
+#[tokio::test]
+async fn validate_invalid_connection_noise() {
+    validate_invalid_connection::<TestTransportNoise, DefaultNetworkingService<NoiseTcpTransport>>(
+    )
+    .await;
+}
+
+async fn inbound_connection_invalid_magic<A, T>()
```
ImplOfAnImpl, line 145: FYI: again, this was moved from `ban.rs`.

dns_server/src/main.rs
```diff
@@ -58,8 +58,9 @@ async fn run(config: Arc<DnsServerConfig>) -> Result<Never, error::DnsServerErro
         boot_nodes: Vec::new(),
         reserved_nodes: Vec::new(),
         whitelisted_addresses: Default::default(),
-        ban_threshold: Default::default(),
-        ban_duration: Default::default(),
+        // Note: this ban config (as well as any other non-backend setting here) won't have
```
TheQuantumPhysicist, line 61: Please define "non-backed". If it's something that isn't used in the DNS server, specifying them one by one helps.

node-lib/src/options.rs
TheQuantumPhysicist: Like discussed, auto will discourage, and banning will be through RPC, with an optional timeout.

p2p/src/peer_manager/peerdb/mod.rs
TheQuantumPhysicist: I think a refresh of the timeout is in order.

Also please create an RPC function to list discouraged peers.

p2p/src/peer_manager/config.rs
```diff
@@ -0,0 +1,130 @@
...
+make_config_setting!(MainLoopTickInterval, Duration, Duration::from_secs(1));
+make_config_setting!(
+    FeelerConnectionsInterval,
+    Duration,
+    Duration::from_secs(2 * 60)
+);
+make_config_setting!(EnableFeelerConnections, bool, true);
+make_config_setting!(ForceDnsQueryIfNoGlobalAddressesKnown, bool, false);
+
+// TODO: this name is too generic, because not all peer manager settings are contained here.
```
TheQuantumPhysicist, line 44: Let's have a discussion about this at some point when you feel it's appropriate. Using words like "internal" can quickly become a mess.

p2p/src/peer_manager/mod.rs
```diff
@@ -539,27 +430,31 @@ where
...
         if let Some(o) = self.observer.as_mut() {
             o.on_peer_ban_score_adjustment(peer.peer_address, peer.score)
         }
 
-        if peer.score >= *self.p2p_config.ban_threshold {
-            let address = peer.peer_address.as_bannable();
-            self.ban(address);
+        if new_score >= *self.p2p_config.ban_config.ban_threshold {
+            self.ban(peer_address);
+        }
```
TheQuantumPhysicist, line 447: This has to go. Only discouragement has to stay.

TheQuantumPhysicist (commented):
First round looks good. I'll review again when you're done with the details we discussed.

TheQuantumPhysicist (approved):

azarovh (approved):
Looks good.
I just have one question. Because of the bloom filter in bitcoin you cannot retrieve all the elements. I'm not sure what is exact reason in case of discouragement but we are doing this intentionally, right?

ImplOfAnImpl:
> Looks good. I just have one question. Because of the bloom filter in bitcoin you cannot retrieve all the elements. I'm not sure what is exact reason in case of discouragement but we are doing this intentionally, right?

[discussed] yes, we do this intentionally

PR #1455, opened 2024-01-17
p2p: fix self-connect detection

ImplOfAnImpl:
This fixes self-connect detection by putting a one-shot `confirmation_sender` into `PeerEvent::PeerInfoReceived`. After sending a `PeerInfoReceived` event to `Backend` during an inbound connection handshake, `Peer` will wait on the corresponding receiver to make sure that the event is received by `Backend` before the "outbound" part of the connection has a chance to complete the handshake (which is relied upon by self-connect detection logic in `Backend`).

No new tests have been added for this, because the existing `self_connect` tests can reproduce the problem if they are made multi-threaded.

There is also some minor unrelated cleanup which I did while trying to reproduce the issue.

### DIFF
--- a/build-tools/p2p-test/README.md
+++ b/build-tools/p2p-test/README.md
@@ -30,7 +30,7 @@ to shut down the containers.
 ```
 run_wallet_cmd.sh node_idx cmd [cmd_params...]
 ```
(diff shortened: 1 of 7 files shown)


TheQuantumPhysicist (approved):
Based on our discussion this is looks good. I want to merge this by the end of the day, but I'd appreciate more people reviewing this to ensure the logic is valid.

azarovh (approved):

PR #1456, opened 2024-01-17
Fix log coloring on Windows

### LINKED ISSUE #1445: Logging doesn't work correctly on Windows
For some reason, the color codes are displayed.

ImplOfAnImpl:
Disable log colouring on Windows unless we're run inside MSYS console.

This fixes #1445.

I also added `#![windows_subsystem = "windows"]` to node-gui to prevent it from opening a console window.

### DIFF
--- a/logging/src/lib.rs
+++ b/logging/src/lib.rs
@@ -88,11 +88,28 @@ where
     });
 }
 
+fn can_use_coloring() -> bool {
+    if cfg!(windows) {
+        // Allow using colors if run in an MSYS console, which includes Git Bash
+        // (the value of the variable will differ depending on how the console was started,
+        // so we don't check it).
+        // Also note that though Cygwin technically supports ansi coloring, it only works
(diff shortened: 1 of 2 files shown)


TheQuantumPhysicist:
Two things:

1. We need a console on Windows. It helps in monitoring the node progress since we have no console.
2. Please take a look at git bash and whether the colors will work there. I'm not sure how hard it's to fix it there.

ImplOfAnImpl:
>     1. We need a console on Windows. It helps in monitoring the node progress since we have no console.

 I've reverted the `#![windows_subsystem = "windows"]` part.

>     2. Please take a look at git bash (on Windows) and whether the colors will work there. I'm not sure how hard it's to fix it there.

Git Bash is based on MSYS, it has the `MSYSTEM` env var set, so the log colouring will continue to work there.

TheQuantumPhysicist (approved):

PR #1466, opened 2024-01-18
Feature/wallet different vrf keys

OBorce:
- use different vrf keys when creating new stake pools
- when scanning new blocks mark decommission and vrf keys as used
- new migration v5
- change wallet get vrf key command to show all issued vrf keys with their used status
- keep the old legacy VRF public key that was derived from a hardened last child number and use soft derivation for every new staking pool
- add more information to the list pool ids command


crypto/src/vrf/mod.rs
```diff
@@ -233,6 +241,63 @@ impl Derivable for ExtendedVRFPrivateKey {
     }
 }
 
+/// Given a tree of keys that are derived from a master key using BIP32 rules, this struct represents
+/// the private key at one of the nodes of this tree.
```
azarovh, line 245: probably "the public key"

wallet/src/account/mod.rs
```diff
@@ -1230,6 +1207,30 @@ impl Account {
         Ok(false)
     }
 
+    /// check if a the output is a CreateStakePool and check if VRR key or decommission_key
```
azarovh, line 1210: "check if the output"
also "VRF key"

azarovh (approved):
looks good to me though I can't verify the internal keychain changes

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -294,8 +294,8 @@ pub enum WalletCommand {
     #[clap(name = "staking-list-created-block-ids")]
     ListCreatedBlocksIds,
 
-    /// Get the current staking VRF key for this account
-    #[clap(name = "staking-get-vrf-key")]
+    /// Show the issued staking VRF keys for this account
+    #[clap(name = "staking-show-vrf-key")]
```
TheQuantumPhysicist, line 298: staking-show-vrf-keys

TheQuantumPhysicist: Also, make it possible to show the old, legacy key. Call it "staking-show-legacy-vrf-key", and write on the documentation "Shows the legacy VRF key that uses an abandoned derivation mechanism. This will not be used for new pools and should be avoided".

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -294,8 +294,8 @@ pub enum WalletCommand {
     #[clap(name = "staking-list-created-block-ids")]
     ListCreatedBlocksIds,
 
-    /// Get the current staking VRF key for this account
-    #[clap(name = "staking-get-vrf-key")]
+    /// Show the issued staking VRF keys for this account
```
TheQuantumPhysicist, line 297: Show the issued staking VRF (Verifiable Random Function) keys for this account. These keys are generated when pools are created. VRF keys are used as a trustless mechanism to ensure the randomness of the staking process, where no one can control the possible outcomes, to ensure decentralization.

wallet/src/key_chain/vrf_key_chain/mod.rs
TheQuantumPhysicist, line 1: I would love for this to be unified with account key chain using generics

wallet/src/key_chain/vrf_key_chain/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+        parent_pubkey: ExtendedVRFPublicKey,
+        legacy_testnet_pubkey: ExtendedVRFPublicKey,
+    ) -> Self {
+        Self {
+            chain_config,
+            account_id,
+            parent_pubkey: parent_pubkey.into(),
+            derived_public_keys: BTreeMap::new(),
+            public_key_to_index: BTreeMap::new(),
+            legacy_testnet_pubkey: legacy_testnet_pubkey.into(),
```
TheQuantumPhysicist, line 69: It's not only for testnet. It can still (theoretically) be used for mainnet. The derivation mechanism is legacy though.

wallet/src/account/mod.rs
```diff
@@ -1230,6 +1207,30 @@ impl Account {
...
+    /// are tracked by this wallet and mark them as used
+    fn mark_create_stake_pool_as_seen(
```
TheQuantumPhysicist, line 1212: "created"?

wallet/wallet-controller/src/read.rs
```diff
@@ -49,6 +52,8 @@ pub struct ReadOnlyController<'a, T> {
     account_index: U31,
 }
 
+type MapAddressWithUsage<T> = BTreeMap<ChildNumber, (Address<T>, bool)>;
```
TheQuantumPhysicist, line 55: In such aliases, it would be nice to explain in a little comment above it what this really means. Like the bool at the end is ambiguous. I know what it's because I'm... me... but the reader won't generally understand.

wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -205,7 +205,7 @@ trait WalletRpc {
     async fn get_vrf_public_key(
```
TheQuantumPhysicist, line 205: keys, same for CLI

TheQuantumPhysicist (approved):

PR #1471, opened 2024-01-18
wallet-rpc: Expose wallet events via RPC

iljakuklic:


wallet/wallet-rpc-lib/tests/basic.rs
```diff
@@ -104,21 +117,36 @@ async fn send_coins_to_acct1(#[case] seed: Seed) {
...
     assert!(coins_after <= (coins_before / 2).unwrap());
     assert!(coins_after >= (coins_before / 3).unwrap());
 
     let balances: Balances = wallet_rpc.request("account_balance", [ACCOUNT1_ARG]).await.unwrap();
     log::info!("acct1 balances: {balances:?}");
 
+    let _result: JsonValue = wallet_rpc
+        .request("node_generate_block", (ACCOUNT0_ARG, [(); 0]))
+        .await
+        .unwrap();
```
iljakuklic, line 143: I wanted to check the state of the transaction tracked by the wallet changes when it's in a block. This call does generate a block but it does not hit the chainstate so no further wallet events are generated. Any suggestions how to achieve that?

wallet/wallet-rpc-lib/src/service/handle.rs
```diff
@@ -83,10 +84,19 @@ impl WalletHandle {
...
+        self.send_raw(WalletCommand::Subscribe(tx))?;
+        Ok(EventStream(rx))
+    }
+
+    /// Stop the wallet service
     pub fn stop(self) -> Result<(), SubmitError> {
         self.send_raw(WalletCommand::Stop)
     }
 
+    /// Check if the wallet service is surrently running
```
TheQuantumPhysicist, line 99: currently

wallet/wallet-rpc-lib/src/service/worker.rs
```diff
@@ -209,6 +248,10 @@ impl WalletWorker {
         Ok(result)
     }
 
+    fn broadcast(&mut self, event: &Event) {
+        self.subscribers.retain(|sub| sub.send(event.clone()).is_ok())
```
TheQuantumPhysicist, line 252: I think this is a little too aggressive, removing subscribers just because there's an error. Perhaps there's a better way to handle errors. Off the top of my head, we should allow unsubscribing and/or clearing subscribers, and when a failure occurs, we write a log message.

In practice, an absent subscriber is the subscriber's problem, not the service's problem. Those who care can always add more subscribers, who cares about dead subscribers? We're only limited by memory and a slight dent in performance.

TheQuantumPhysicist: One important point here is to make sending notifications async. Given that websocket uses TCP, if one subscriber doesn't respond, this transaction will take a very long time to complete, blocking other subscribers. This glosses on the issue of the above comment. It shouldn't make a difference if a subscriber is dead/absent or whatever prevents the notifications from arriving and it shouldn't have an effect on others, to the extent possible by the async library.

TheQuantumPhysicist: If async isn't an option due to time constraints, going as far as running a thread pool isn't wouldn't be super-crazy... at least until we have the time to do this right.

wallet/wallet-rpc-lib/src/service/events.rs
```diff
@@ -0,0 +1,167 @@
...
+        log::trace!("Emitting event {event:?}");
+        if let Err(err) = self.sender.send(event) {
+            log::warn!("Events channel closed unexpectedly: {err}");
+        }
+    }
+}
+
+impl wallet::wallet_events::WalletEvents for WalletServiceEvents {
+    fn new_block(&self) {
+        // Do nothing for now
```
TheQuantumPhysicist, line 120: Can we add a new `Event::NewBlock` and just emit it?

TheQuantumPhysicist (approved):
Besides the comments, looking good. The async part is a little concerning, but I'm OK with moving that to an issue if it's a big deal.

The `retain` part should be at least changed to error log in case a full solution isn't devised.

Besides that, great work. I'm glad we're already done with this!

wallet/wallet-rpc-lib/src/service/worker.rs
iljakuklic: > I think this is a little too aggressive, removing subscribers just because there's an error. Perhaps there's a better way to handle errors. Off the top of my head, we should allow unsubscribing and/or clearing subscribers, and when a failure occurs, we write a log message.

This only happens when the other channel endpoint is already dead. There is an unsubscribe method generated by `jsonrpsee` so subscribers can clean up as they wish.

OBorce (approved):
If the notifications are intended to be used by a user it would be better to change the account_id to account_index, as the id would not mean much to the user.

You can change the types in the wallet_events, as from what I can see those arguments are never used anyway.

Otherwise looks good.

TheQuantumPhysicist (approved):

TheQuantumPhysicist:
I converted this to a draft. Please mark it as ready for review once the readme stuff we agreed about is done.

PR #1472, opened 2024-01-18
Tokens fees testnet fork

azarovh:
New fork for testnet at height 138244 now includes the following changes:
- new fees for tokens operations
- max ticker length increased to 12

Other networks don't fork and just use new values

### DIFF
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -672,9 +672,13 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         Ok(())
     }
 
-    fn check_transactions(&self, block: &Block) -> Result<(), CheckBlockTransactionsError> {
+    fn check_transactions(
+        &self,
+        block: &Block,
+        block_height: BlockHeight,
+    ) -> Result<(), CheckBlockTransactionsError> {
         for tx in block.transactions() {
-            tx_verifier::check_transaction(self.chain_config, tx)?;
+            tx_verifier::check_transaction(self.chain_config, block_height, tx)?;
         }
 
         // Note: duplicate txs are detected through duplicate inputs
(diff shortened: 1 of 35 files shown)


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -722,7 +726,15 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
             );
         }
 
-        self.check_transactions(block)
+        let prev_block_height = self
```
azarovh, line 729: please confirm that it's ok to get previous block index in `check_block`. I mean I assume that the index of previous block always exist if we got here

common/src/chain/upgrades/chainstate_upgrade.rs
```diff
@@ -25,30 +31,69 @@ pub enum RewardDistributionVersion {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum TokensTickerMaxLengthVersion {
+    /// Initial tokens ticker max length values
+    V0,
+    /// Updated tokens ticker max length values
+    V1,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
 pub struct ChainstateUpgrade {
```
azarovh, line 51: tbh this is getting ridiculous with all these versions. I'd rework this struct into an enum of continuous forks, so that right now it would be an enum of 2. Thoughts?

chainstate/src/detail/chainstateref/mod.rs
TheQuantumPhysicist: I believe so. You can see another height querying in check_block_header to check the reorg depth.

common/src/chain/upgrades/chainstate_upgrade.rs
TheQuantumPhysicist: You can do that (in another PR). Though keep in mind that you cannot make the upgrades monotonic (like we tried to do before) because that will make it some cases mutually exclude others. Like for example, you cannot run PoW tests and have them upgraded.

common/src/chain/upgrades/chainstate_upgrade.rs
```diff
@@ -25,30 +31,69 @@ pub enum RewardDistributionVersion {
...
     ) -> Self {
         Self {
             token_issuance_version,
             reward_distribution_version,
+            tokens_fee_version,
+            tokens_ticker_length_version,
         }
     }
 
+    pub fn latest() -> Self {
```
TheQuantumPhysicist, line 73: Please. No "latest". This is a hazard that can cause unintended forks.

TheQuantumPhysicist (approved):

TheQuantumPhysicist (commented):
Looking good. Please be let me know before merging this.

PR #1478, opened 2024-01-19
Add block height along the block ID in list produced blocks command

OBorce:


wallet/wallet-controller/src/read.rs
```diff
@@ -264,7 +264,9 @@ impl<'a, T: NodeInterface> ReadOnlyController<'a, T> {
         tasks.try_collect().await
     }
 
-    pub fn get_created_blocks(&self) -> Result<Vec<Id<GenBlock>>, ControllerError<T>> {
+    pub fn get_created_blocks(
+        &self,
+    ) -> Result<Vec<(BlockHeight, Id<GenBlock>)>, ControllerError<T>> {
```
iljakuklic, line 269: Can we move `wallet_rpc_lib::types::BlockInfo` into the controller crate and use that instead of the tuple?

wallet/wallet-rpc-lib/src/rpc/types.rs
```diff
@@ -376,3 +376,9 @@ pub struct RpcTokenId {
 pub struct NodeVersion {
     pub version: String,
 }
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct CreatedBlockInfo {
+    pub block_id: Id<GenBlock>,
+    pub block_height: BlockHeight,
+}
```
iljakuklic, line 384: Any particular reason you did not just use `BlockInfo`?

iljakuklic (approved):
Looks good. Some suggestions below.

wallet/wallet-rpc-lib/src/rpc/types.rs
OBorce: fixed

wallet/wallet-controller/src/read.rs
OBorce: moved

TheQuantumPhysicist (approved):

PR #1480, opened 2024-01-20
Fix decommission request from TxOutput::ProduceBlockFromStake

azarovh:
Decommission request only worked if a pool is decommissioned from `TxOutput::StakePoolData` meaning it wasn't used for staking. If it was cold wallet couldn't sign it. The problem was that wallet needs to track pool utxo using both staker and decommission keys for both StakePoolData and ProduceBlockFromStake outputs.

UPD: now PartiallySignedTransaction provides all the information for a not-synced wallet to sign

### DIFF
--- a/test/functional/wallet_decommission_request.py
+++ b/test/functional/wallet_decommission_request.py
@@ -320,11 +320,13 @@ class WalletDecommissionRequest(BitcoinTestFramework):
             pools = await wallet.list_pool_ids()
             assert_equal(len(pools), 0)
 
-if __name__ == '__main__':
-    tf1 = WalletDecommissionRequest()
-    tf1.use_wallet_to_produce_block = False
-    tf1.main(exit_on_success=False)
+# `use_wallet_to_produce_block` indicates whether a test should use a pool created by a wallet to produce block
+# `exit_on_success` indicates if the process should exit or continue and run next test case
+def wallet_decommission_request_test_case(use_wallet_to_produce_block, exit_on_success):
+    tf = WalletDecommissionRequest()
+    tf.use_wallet_to_produce_block = use_wallet_to_produce_block
+    tf.main(exit_on_success)
 
-    tf2 = WalletDecommissionRequest()
-    tf2.use_wallet_to_produce_block = True
-    tf2.main()
+if __name__ == '__main__':
+    wallet_decommission_request_test_case(False, False)
+    wallet_decommission_request_test_case(True, True)
--- a/test/functional/wallet_decommission_request.py
+++ b/test/functional/wallet_decommission_request.py
@@ -278,11 +278,13 @@ class WalletDecommissionRequest(BitcoinTestFramework):
             pools = await wallet.list_pool_ids()
             assert_equal(len(pools), 1)
             assert_equal(pools[0].balance, '40000')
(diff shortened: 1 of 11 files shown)


OBorce:
From what I can see in the current implementation of `sign_raw_transaction` the problem is that it tries to find all of the UTXOs of the transaction's inputs from the cache. If I understood how the cold wallet will function it will not be connected to the internet and it will not be syncing with the blockchain, so this will always fail.

One suggestion is to change the PartiallySignedTransaction's witnesses instead of Option<InputWitness> to be an enum { SignedInputWithness, ToBeSigned{ Optional<TxOutput>, Destination} }, that way the cold wallet will know what it needs to sign.

azarovh:
> the problem is that it tries to find all of the UTXOs of the transaction's inputs from the cache

I'd say these are 2 distinct problems. I agree that `sign_raw_transaction` should be changed to allow cold wallet stay "cold" by providing more information. But at the same time a wallet must watch both destinations of a pool (staking and decommission) so my fix is still valid.

test/functional/wallet_decommission_request.py
```diff
@@ -329,6 +320,11 @@ async def async_test(self):
             pools = await wallet.list_pool_ids()
             assert_equal(len(pools), 0)
 
-
 if __name__ == '__main__':
-    WalletSubmitTransaction().main()
+    tf1 = WalletDecommissionRequest()
+    tf1.use_wallet_to_produce_block = False
```
TheQuantumPhysicist, line 325: Why not pass this in the constructor of that class? This looks very messy... Just add an `__init__()` function, and default that parameter there.

wallet/src/account/mod.rs
```diff
@@ -1157,12 +1188,35 @@ impl Account {
         self.key_chain.get_addresses_usage_state()
     }
 
-    /// Return true if this transaction output is can be spent by this account or if it is being
+    fn collect_output_destinations(&self, txo: &TxOutput) -> Vec<Destination> {
```
TheQuantumPhysicist, line 1191: This function puts us back at the "every output has one destination" and "every output has a some 'value' associated with it", which is what we've trying to avoid everywhere, and which is why TxOutput doesn't have such getters. While this function is called "collect_output_destinations", we're actively preferring decommission destinations instead of staking destinations. Now I understand why you're doing this, but this doesn't rhyme with the name of the function or how it works based on output types. Please take another look at this.

TheQuantumPhysicist (commented):
Besides the one major design issue, this looks good.

wallet/src/account/mod.rs
azarovh: I'm not sure I understand your concern. The statement of this function is "every output can have [0;n] destinations". And it is valid only for this specific case in the wallet, where we want to watch all destinations of an output. I'm not preferring any destination here and just want to collect everything related.

OBorce: if we are going to include decommission key to be tracked, we will need a filter on the list pool ids to exclude pools in the case the Staker is not ours when called from the staking code.

azarovh: > we will need a filter on the list pool ids

added the filter, thanks for noticing

TheQuantumPhysicist (approved):

PR #1481, opened 2024-01-20
Save tx fee in API server and fix response to include it

OBorce:
Tx fee is now saved in the DB and returned in the transaction response

### DIFF
--- a/api-server/api-server-common/src/storage/impls/postgres/transactional/mod.rs
+++ b/api-server/api-server-common/src/storage/impls/postgres/transactional/mod.rs
@@ -20,7 +20,7 @@ use std::sync::Arc;
 
 use bb8_postgres::{bb8::PooledConnection, PostgresConnectionManager};
 use common::{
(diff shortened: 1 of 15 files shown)


TheQuantumPhysicist (approved):

api-server/api-server-common/src/storage/impls/postgres/transactional/mod.rs
```diff
@@ -131,7 +131,7 @@ impl<'a> ApiServerPostgresTransactionalRo<'a> {
     pub async fn get_transaction(
         &mut self,
         transaction_id: Id<Transaction>,
-    ) -> Result<Option<(Option<Id<Block>>, SignedTransaction)>, ApiServerStorageError> {
+    ) -> Result<Option<(Option<Id<Block>>, TransactionInfo)>, ApiServerStorageError> {
```
TheQuantumPhysicist, line 134: Please don't forget to bump the storage version when changing the internal format. It'll automatically wipe the database and start over.

PR #1483, opened 2024-01-21
Move transaction checks from chainstate to tx_verifier and use them in mempool

azarovh:
I moved the implementations, so Chainstate calls the same checks as before.

Not sure if I should remove tokens checks from wallet because it looked like a workaround but I might be wrong

### DIFF
--- a/chainstate/src/detail/error.rs
+++ b/chainstate/src/detail/error.rs
@@ -21,8 +21,7 @@ use super::{
     chainstateref::EpochSealError,
     orphan_blocks::OrphanAddError,
     transaction_verifier::{
-        error::{ConnectTransactionError, TokensError},
-        storage::TransactionVerifierStorageError,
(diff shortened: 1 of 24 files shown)


iljakuklic (approved):

OBorce:
Maybe leave the checks in the wallet as the error reported to the user will be shorter than the full response from the mempool?

chainstate/src/detail/error.rs
```diff
@@ -150,25 +149,9 @@ pub enum CheckBlockError {
 #[derive(Error, Debug, PartialEq, Eq, Clone)]
 pub enum CheckBlockTransactionsError {
     #[error("Blockchain storage error: {0}")]
```
TheQuantumPhysicist, line 151: Wrong description

TheQuantumPhysicist (approved):

PR #1485, opened 2024-01-22
Disable http_rpc_enabled by default for gui

azarovh:


node-gui/src/backend/mod.rs
```diff
@@ -82,7 +82,18 @@ pub async fn node_initialize(_time_getter: TimeGetter) -> anyhow::Result<Backend
         std::env::set_var("RUST_LOG", "info");
     }
 
-    let opts = node_lib::Options::from_args(std::env::args_os());
+    let opts = {
+        let mut opts = node_lib::Options::from_args(std::env::args_os());
+        opts.command = opts.command.clone().or_else(|| {
+            let options = node_lib::RunOptions {
+                http_rpc_enabled: Some(false),
```
TheQuantumPhysicist, line 89: It doesn't hurt to add a comment explaining what we're doing. I'll do it and push.

PR #1487, opened 2024-01-22
P2p: prevent connecting to the same ip address multiple times

ImplOfAnImpl:
Previously, we would allow connecting to a peer unless we're already connected to the same socket address (i.e. both the ip address and port are the same). And because inbound connections have some random port number, a socket address of an inbound peer is always different from a socket address of an outbound one. So it was common to have "double" connections, where peer A would connect to peer B and then peer B would connect to peer A, despite the fact that it already has an inbound connection from peer A. 

Now we also check whether ip addresses are the same. The logic is:
1) Inbound connections are allowed no matter what.
2) An outbound connection is not allowed if we already have an inbound connection from the same ip address, unless the outbound connection is a manual one.

### DIFF
--- a/p2p/src/peer_manager/mod.rs
+++ b/p2p/src/peer_manager/mod.rs
@@ -88,7 +88,7 @@ use self::{
     },
 };
 
-#[derive(Default, Debug)]
+#[derive(Default, Debug, Clone)]
 pub struct PeerManagerConfig {
     /// Maximum allowed number of inbound connections.
     pub max_inbound_connections: MaxInboundConnections,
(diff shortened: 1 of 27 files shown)


p2p/src/peer_manager/mod.rs
```diff
@@ -622,16 +651,16 @@ where
...
+        peer_role: PeerRole,
     ) -> crate::Result<()> {
         ensure!(
             !self.pending_outbound_connects.contains_key(&address),
             P2pError::PeerError(PeerError::Pending(address.to_string())),
         );
         ensure!(
-            !self.is_address_connected(&address),
+            !self.should_reject_because_already_connected(&address, peer_role),
             P2pError::PeerError(PeerError::PeerAlreadyExists),
```
TheQuantumPhysicist, line 662: Please make sure that this will translate to a log entry.

TheQuantumPhysicist (approved):

azarovh (approved):

azarovh (approved):

PR #1488, opened 2024-01-22
Return new Tx ID on wallet commands

OBorce:
- return the new Tx Id from wallet commands
- check created transactions in the wallet


wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -702,6 +704,14 @@ impl CommandHandler {
         ConsoleCommand::Print(status_text.to_owned())
     }
 
+    pub fn new_tx_submitted_command(new_tx: NewTransaction) -> ConsoleCommand {
+        let status_text = format!(
+            "The transaction was submitted successfully with ID {}",
```
TheQuantumPhysicist, line 709: Can you make the id go into a new line, to simplify it for parsers to find it?

TheQuantumPhysicist (approved):
Looking great, besides the minor comments. I will be testing that soon.

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: fixed

PR #1493, opened 2024-01-23
Create tarball/hashfile for each binary on tag release

erubboli:
This PR utilizes `cargo-dist` to generate releases for tags following the specified pattern `**[0-9]+.[0-9]+.[0-9]+*`.

- A typical release page will now look like this: [Mintlayer Core v0.2.11](https://github.com/erubboli/mintlayer-core/releases/tag/v0.2.11). Please note that it will only include the tarball and hash file, omitting the Windows installer and installation script.

Unfortunately, the ARM integration is currently missing since GitHub does not natively support it yet. We have requested access to the private beta ( see: [GitHub Roadmap Issue #836](https://github.com/github/roadmap/issues/836) ).

PS: worked on personal fork to avoid messing around with main repo actions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,260 @@
+# Copyright 2022-2023, axodotdev
+# SPDX-License-Identifier: MIT or Apache-2.0
+#
+# CI that:
+#
+# * checks for a Git Tag that looks like a release
+# * builds artifacts with cargo-dist (archives, installers, hashes)
+# * uploads those artifacts to temporary workflow zip
+# * on success, uploads the artifacts to a Github Release
+#
+# Note that the Github Release will be created with a generated
+# title/body based on your changelogs.
+
+name: Release
+
+permissions:
+  contents: write
+
+# This task will run whenever you push a git tag that looks like a version
+# like "1.0.0", "v0.1.0-prerelease.1", "my-app/0.1.0", "releases/v1.0.0", etc.
+# Various formats will be parsed into a VERSION and an optional PACKAGE_NAME, where
+# PACKAGE_NAME must be the name of a Cargo package in your workspace, and VERSION
+# must be a Cargo-style SemVer Version (must have at least major.minor.patch).
+#
+# If PACKAGE_NAME is specified, then the announcement will be for that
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -231,3 +256,8 @@ overflow-checks = true
 
 [profile.test.package.script]
 opt-level = 2
+
+# The profile that 'cargo dist' will build with
+[profile.dist]
+inherits = "release"
+lto = "thin"
```
TheQuantumPhysicist, line 263: We have to be careful about LTO. I do want it, but if we want it, we should test with it. There was a case when LTO lead to ruining secure memory cleanups.

TheQuantumPhysicist (approved):
Besides the LTO issue, this look fine.

erubboli:
rebased and fixed git stuff

erubboli:
reopening the PR

.github/workflows/release.yml
```diff
@@ -0,0 +1,260 @@
...
+# spin up, creating an independent announcement for each one. However Github
+# will hard limit this to 3 tags per commit, as it will assume more tags is a
+# mistake.
+#
+# If there's a prerelease-style suffix to the version, then the release(s)
+# will be marked as a prerelease.
+on:
+  push:
+    tags:
+      - '**[0-9]+.[0-9]+.[0-9]+*'
```
TheQuantumPhysicist, line 43: If we want to restrict this to `vX.Y.Z*`, shouldn't this be `'v[0-9]+.[0-9]+.[0-9]+*'`? I'm not sure whether this is regex, because as regex this doesn't make sense:

https://regex101.com/r/fEfP9i/1

erubboli: We can change to `vX.Y.Z`, sure. This is the default regex used by `cargo-dist`.

erubboli: had to revert to the original regex since `cargo-dist` enforce the `release.yml` file integrity.

Cargo.toml
erubboli: set `lto = "off"`.

erubboli:
output of the full run of the current configuration accessible here: https://github.com/erubboli/mintlayer-core/releases/tag/v0.2.40

PR #1496, opened 2024-01-24
Add delegation tab to node-gui

azarovh:
Screenshot 2024-01-24 at 13 01 35

### DIFF
--- a/common/src/address/mod.rs
+++ b/common/src/address/mod.rs
@@ -107,6 +107,23 @@ impl<T: Addressable> Address<T> {
         let qrcode = qrcode_from_str(&self.address)?;
         Ok(qrcode)
     }
+
+    pub fn to_short_string(&self, cfg: &ChainConfig) -> Result<String, AddressError> {
+        let obj = self.decode_object(cfg)?;
+        let prefix_len = obj.address_prefix(cfg).len();
+
+        let result = if self.address.len() < prefix_len + 8 {
+            self.to_string()
+        } else {
+            // prefix + 4 first chars + ... + 4 last chars
+            format!(
+                "{}...{}",
+                self.address[0..prefix_len + 4].to_owned(),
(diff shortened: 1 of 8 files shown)


common/src/address/mod.rs
```diff
@@ -107,6 +107,23 @@ impl<T: Addressable> Address<T> {
...
+        let obj = self.decode_object(cfg)?;
+        let prefix_len = obj.address_prefix(cfg).len();
+
+        let result = if self.address.len() < prefix_len + 8 {
+            self.to_string()
+        } else {
+            // prefix + 4 first chars + ... + 4 last chars
+            format!(
+                "{}...{}",
+                self.address[0..prefix_len + 4].to_owned(),
```
TheQuantumPhysicist, line 121: I'm not sure how forgiving clippy will be about this... it's heavily frowned upon to take a slice from strings. The correct, type-safe way to do it is to ensure it's all ASCII, convert it to a `Vec<u8>` (or use `chars()`), crop it, then back to String. Now we know that addresses are strictly ASCII so what you did is safe, but static analyzers are not that smart.

TheQuantumPhysicist (approved):

common/src/address/mod.rs
azarovh: clippy is fine with it

azarovh: added a test

TheQuantumPhysicist: Thanks

azarovh: I had to add the solution with bytes in another PR because clippy starts complaining there

TheQuantumPhysicist: Told you...

I don't understand how this is happening... this is the second time this is happening... it works at first but then it complains. Very confusing.

PR #1497, opened 2024-01-24
Add support for cold wallet

OBorce:
- add command line argument --cold-wallet to start the wallet without a node connection, with support of a subset of the commands


wallet/wallet-node-client/src/rpc_client/dummy_client.rs
TheQuantumPhysicist, line 1: Could we call this `ColdWalletClient` instead of dummy?

wallet/wallet-node-client/src/rpc_client/dummy_client.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    async fn get_mainchain_blocks(
+        &self,
+        from: BlockHeight,
+        max_count: usize,
+    ) -> Result<Vec<Block>, Self::Error> {
+        if let Some(client) = self.opt_client.as_ref() {
+            client.get_mainchain_blocks(from, max_count).await
+        } else {
+            Err(NodeRpcError::ResponseError(
```
TheQuantumPhysicist, line 73: Also could we replace this with a more meaningful error? Remember that users will be using this and will be getting "not implemented". This isn't really helpful. Please create a special type of errors for the cold wallet, and explain in it what the error means. Something like "This command requires communication with the node, but this cannot happen in cold-wallet mode". Most errors will be that one, but if need be others, please explain it.

wallet/wallet-node-client/src/rpc_client/mod.rs
```diff
@@ -33,6 +37,21 @@ pub enum NodeRpcError {
     ResponseError(jsonrpsee::core::ClientError),
 }
 
+#[derive(Clone, Debug)]
+pub struct MaybeDummyNode {
+    opt_client: Option<NodeRpcClient>,
```
TheQuantumPhysicist, line 42: Why do we need a "MaybeDummyNode" and an Option? Why don't we create an RPC client that's only for cold wallet?

wallet/wallet-node-client/src/rpc_client/dummy_client.rs
OBorce: The unsupported commands are filtered out before reaching here, so this error will never trigger

TheQuantumPhysicist: Still, this is not 100% filtered in RPC. So let's have some good error support so that if something goes wrong, users would understand.

TheQuantumPhysicist (approved):

wallet/wallet-node-client/src/rpc_client/cold_wallet_client.rs
```diff
@@ -0,0 +1,209 @@
...
+            best_block_height: BlockHeight::zero(),
+            best_block_timestamp: genesis.timestamp(),
+            median_time: genesis.timestamp(),
+            is_initial_block_download: false,
+        })
+    }
+
+    async fn get_block(&self, _block_id: Id<Block>) -> Result<Option<Block>, Self::Error> {
+        Err(ColdWalletRpcError::NotAvailable)
+    }
```
iljakuklic, line 59: Ideally this would be a compilation time error, where `ColdWalletClient` does NOT implement `NodeInterface` at all.

In the `NodeController<N>`, only the methods that require communication with node would require the `N: NodeInterface` constraint. Methods that do not need to communicate with the node leave `N` unconstrained.

It is probably not a good time to do it now but please file an issue.

iljakuklic (approved):

PR #1500, opened 2024-01-25
Fix `ConnectTransactionError::MissingTransactionNonce`

azarovh:
When we spend from delegation we increment nonces: 1 -> 2 -> 3
When we undo such txs we decrement nonces.

But if we spend the delegation entirely and the pool is decommissioned we delete nonce from db: 1->2->3->None
Now if we want to undo that last transaction we need to read the nonce and decrement it, but it’s not in the db because we deleted it.

The simplest solution would be to not delete nonces from the db.

### DIFF
--- a/chainstate/tx-verifier/src/transaction_verifier/mod.rs
+++ b/chainstate/tx-verifier/src/transaction_verifier/mod.rs
@@ -439,12 +439,8 @@ where
                             ConnectTransactionError::DelegationDataNotFound(delegation_id),
                         )?;
                     if !accounting_view.pool_exists(*delegation_data.source_pool())? {
-                        // clear the nonce
-                        self.account_nonce.insert(
-                            AccountType::Delegation(delegation_id),
-                            CachedOperation::Erase,
-                        );
-
+                        // When deleting a delegation nonce value is preserved in the db
+                        // in case reorg happens so we don't have to restore it
                         Some(
                             self.pos_accounting_adapter
                                 .operations(tx_source)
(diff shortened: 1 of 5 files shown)


chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -439,12 +439,8 @@ where
...
                     if !accounting_view.pool_exists(*delegation_data.source_pool())? {
-                        // clear the nonce
-                        self.account_nonce.insert(
-                            AccountType::Delegation(delegation_id),
-                            CachedOperation::Erase,
-                        );
-
+                        // When deleting a delegation nonce value is preserved in the db
+                        // in case reorg happens so we don't have to restore it
                         Some(
```
TheQuantumPhysicist, line 444: Please make sure we're not doing anything else like this. We should only delete such information in a pruned database after the max reorg depth.

TheQuantumPhysicist:
> P.S. This is a draft so that if the solution is acceptable I will have to add fork for testnet.

I don't think you need a fork. This isn't a validation rule.

chainstate/tx-verifier/src/transaction_verifier/mod.rs
azarovh: > Please make sure we're not doing anything else like this.

I'd merge this and then I'll go thought the code and see what tests I can add.

> We should only delete such information in a pruned database after the max reorg depth.

we don't do this yet

TheQuantumPhysicist (approved):

PR #1501, opened 2024-01-25
node-rpc: Expose chainstate events

iljakuklic:


chainstate/src/detail/mod.rs
```diff
@@ -191,14 +197,17 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
         Self {
             chain_config,
             chainstate_config,
             chainstate_storage,
             tx_verification_strategy,
             orphan_blocks,
             custom_orphan_error_hook,
-            events_controller: EventsController::new(),
+            events_controller,
+            event_broadcast,
```
iljakuklic, line 210: Here is the bad stuff. There are now two mechanisms for delivering and subscribing to events. The `EventController` does not seem to have an easy way to clean up dead subscriptions, producing a risk of memory exhaustion even in expected non-adversial scenarios. The implementation is also somewhat weird with the thread pool in place just to deliver events. On the other hand, the newly introduced `Broadcaster` lacks some features like the ability to wait until all events are processed.

Having both is highly unsatifying and error prone. There's also the potential issue with changing the event mechanism last minute and whatever unintended consequences that might cause in case we unify them in any way. Any suggestions?

chainstate/src/detail/mod.rs
```diff
@@ -191,14 +197,17 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
         time_getter: TimeGetter,
     ) -> Self {
         let orphan_blocks = OrphansProxy::new(*chainstate_config.max_orphan_blocks);
+        let events_controller = EventsController::new();
+        let event_broadcast = broadcaster::Broadcaster::new();
```
TheQuantumPhysicist, line 201: The naming is confusing... we have now two event subscription systems... which I'm not opposed to. But maybe one should reference RPC?

iljakuklic: Renamed both to make the use of each explicit.

chainstate/src/detail/mod.rs
```diff
@@ -191,14 +197,17 @@ impl<S: BlockchainStorage, V: TransactionVerificationStrategy> Chainstate<S, V>
...
         Self {
```
TheQuantumPhysicist, line 202: You probably meant to rename the variables too, right?

chainstate/src/detail/mod.rs
TheQuantumPhysicist: As discussed, this is OK, because these serve different purposes.

TheQuantumPhysicist (approved):

PR #1507, opened 2024-01-26
Don't update mempool in handle_new_tip if the actual chainstate tip differs

ImplOfAnImpl:


mempool/src/pool/reorg.rs
```diff
@@ -135,14 +141,31 @@ pub fn handle_new_tip<M: MemoryUsageEstimator>(
...
+    if is_ibd || new_tip != actual_tip {
+        log::debug!("Not updating mempool: is_ibd = {is_ibd}, new_tip = {new_tip:?}, actual_tip = {actual_tip:?}");
 
         // We still need to update the current tx_verifier tip
         let mut old_transactions = mempool.reset();
         if old_transactions.next().is_some() {
-            log::warn!("Discarding mempool transactions during IBD");
+            log::warn!("Discarding mempool transactions during IBD or because the tip has changed");
         }
         return Ok(());
```
ImplOfAnImpl, line 170: Btw, is this needed when the tip is out of sync? Why was it needed in the ibd case?

TheQuantumPhysicist (approved):

mempool/src/pool/reorg.rs
iljakuklic: I think in the IBD case, if the IBD correctly behaves as a latch, it should not happen. I just found it a bit risky to go for a full assertion. I would say dropping transactions from mempool is not appropriate when mempool is temporarily out of sync. However, when mempool catches up, it should `reset()` and reverify to see which transactions still apply at the current tip.

ImplOfAnImpl: It turned out that some functional tests do depend on this "mempool.reset" during ibd, so I left this call under "if is_ibd" and wrote a comment.

iljakuklic (approved):

PR #1510, opened 2024-01-27
Rename send from delegation to withdraw

TheQuantumPhysicist:


azarovh (approved):
In this case we should rename things in gui as well for consistency

iljakuklic (approved):

TheQuantumPhysicist:
> In this case we should rename things in gui as well for consistency

I think I did that already. Did I miss somewhere?

azarovh:
>I think I did that already. Did I miss somewhere?

right, missed that somehow

PR #1511, opened 2024-01-27
rpc: Send error Json object in case of auth failure

iljakuklic:
The following json snippet will be returned in case of auth failure instead of just an empty body
```json
{"code":401,"message":"Unauthorized"}
```

### DIFF
--- a/rpc/src/rpc_auth/mod.rs
+++ b/rpc/src/rpc_auth/mod.rs
@@ -112,15 +112,28 @@ impl<B> ValidateRequest<B> for RpcAuth {
     type ResponseBody = Body;
 
     fn validate(&mut self, request: &mut Request<B>) -> Result<(), Response<Self::ResponseBody>> {
+        use jsonrpsee::types;
+
         let res = self.check_auth(request);
         match res {
             Ok(true) => Ok(()),
             Ok(false) => {
                 log::error!("Unauthorized RPC request {:?}", request.uri());
+                let status = http::StatusCode::UNAUTHORIZED;
+                let err_obj = types::ErrorObject::owned(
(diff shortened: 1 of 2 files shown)


rpc/src/rpc_auth/mod.rs
```diff
@@ -117,10 +118,19 @@ impl<B> ValidateRequest<B> for RpcAuth {
...
+
                 Err(Response::builder()
-                    .status(http::StatusCode::UNAUTHORIZED)
+                    .status(status)
                     .header(http::header::WWW_AUTHENTICATE, "Basic")
-                    .body(Body::empty())
+                    .header(http::header::CONTENT_TYPE, "application/json")
+                    .body(body.into())
                     .expect("must be valid"))
             }
```
TheQuantumPhysicist, line 135: This doesn't conform to the correct format. Not that I'm an expert in this, but I checked online. This is the correct format for the error object, but not the json response. Trying this out:

```
curl -H 'Content-Type: application/json'   -d '{"jsonrpc": "2.0", "id": 1, "method": "chainstate_info", "params": []}' http://127.0.0.1:3030
{"code":401,"message":"Unauthorized"}
```

But the part `{"code":401,"message":"Unauthorized"}` should be inside a json RPC response, so something like this:

```
{
  "jsonrpc": "2.0",
  "error": {
    "code": 401,
    "message": "Unauthorized",
  },
  "id": 1
}
```

iljakuklic: Changed. The `id` field is null though, since authentication happens before request ID is available but I think this is correct according to the spec.

TheQuantumPhysicist (approved):

PR #1521, opened 2024-01-30
Create the signature facilities for arbitrary messages

TheQuantumPhysicist:


wallet/wallet-rpc-lib/src/rpc/mod.rs
```diff
@@ -422,6 +425,47 @@ impl<N: NodeInterface + Clone + Send + Sync + 'static> WalletRpc<N> {
...
     }
 
+    pub async fn sign_challenge(
+        &self,
+        account_index: U31,
+        challenge: Vec<u8>,
+        address: String,
+    ) -> WRpcResult<SignedArbitraryMessage, N> {
+        let config = ControllerConfig { in_top_x_mb: 5 }; // irrelevant
+        let destination = Address::from_str(&self.chain_config, &address)
```
ImplOfAnImpl, line 435: This `Address::from_str` function is a bit strange - it calls `decode_object`, ignores its result and stores the original string inside `Address`, on which you later call `decode_object` again. 
I suggest to introduce a separate function `decode_object_from_str` in Address:
```
pub fn decode_object_from_str(address: &str, cfg: &ChainConfig) -> Result<T, AddressError> {
    let data = encoding::decode(address)?;
    // here goes the rest of the body of the current "decode_object" 
}

 pub fn decode_object(&self, cfg: &ChainConfig) -> Result<T, AddressError> {
    Self::decode_object_from_str(&self.address, cfg)
}
```
and use it here and in similar places below.

ImplOfAnImpl (approved):

wallet/wallet-rpc-lib/src/rpc/mod.rs
TheQuantumPhysicist: from_str needs to ensure that the encoding is valid. That's why we do it like that.

common/src/chain/transaction/signature/inputsig/arbitrary_message/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    AttemptedToProduceSignatureForAnyoneCanSpend,
+    #[error("Classical multisig signature attempted in uni-party function")]
+    AttemptedToProduceClassicalMultisigSignatureInUnipartySignatureCode,
+    #[error("Unsupported yet!")]
+    Unsupported,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct SignedArbitraryMessage {
+    raw_signature: Vec<u8>,
```
iljakuklic, line 54: Seems to me `ArbitraryMessageSignature` would be a more appropriate name. Seems the struct does not actually contain the message.

common/src/chain/transaction/signature/inputsig/arbitrary_message/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const MESSAGE_MAGIC_PREFIX: &str = "===MINTLAYER MESSAGE BEGIN===";
+const MESSAGE_MAGIC_SUFFIX: &str = "===MINTLAYER MESSAGE END===";
```
iljakuklic, line 17: Does it make sense to put some newlines around it, so the prefix/suffix are on separate lines when printed?

iljakuklic (approved):

common/src/chain/transaction/signature/inputsig/arbitrary_message/mod.rs
TheQuantumPhysicist: Yeah maybe, but it doesn't matter all that much. These are just added as protection from spoofing transaction signatures challenges.

common/src/chain/transaction/signature/inputsig/arbitrary_message/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    pub fn to_hex(self) -> String {
+        hex::encode(self.raw_signature)
+    }
+
+    pub fn verify_signature(
+        &self,
+        chain_config: &ChainConfig,
+        outpoint_destination: &Destination,
```
azarovh, line 82: should it be just `destination`? outpoint_ part seems misleading

common/src/chain/transaction/signature/inputsig/arbitrary_message/mod.rs
TheQuantumPhysicist: Done.

azarovh (approved):

common/src/chain/transaction/signature/inputsig/arbitrary_message/mod.rs
TheQuantumPhysicist: Done.

PR #1526, opened 2024-02-01
Add a summary of the signed transaction

OBorce:
Print out a summary of the singed transaction form the wallet CLI

### DIFF
--- a/common/src/chain/transaction/printout.rs
+++ b/common/src/chain/transaction/printout.rs
@@ -90,7 +90,7 @@ pub fn transaction_summary(tx: &Transaction, chain_config: &ChainConfig) -> Stri
             fmt_dest(p.staker()),
             fmt_vrf(p.vrf_public_key()),
             fmt_dest(p.decommission_key()),
-            p.margin_ratio_per_thousand().into_decimal_str(),
+            p.margin_ratio_per_thousand().into_percentage_str(),
             fmt_ml(&p.cost_per_block())
         )
     };
--- a/common/src/chain/transaction/printout.rs
+++ b/common/src/chain/transaction/printout.rs
@@ -1,0 +1,382 @@
+// Copyright (c) 2021-2024 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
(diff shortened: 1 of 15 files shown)


iljakuklic (approved):

TheQuantumPhysicist:
I will change a few things and make transactions generally printable. I converted the PR to draft.

TheQuantumPhysicist (commented):
Switched to ready to review. Please double-check that what I did makes sense.

common/src/chain/transaction/printout.rs
```diff
@@ -0,0 +1,382 @@
...
+        "=== END OF INPUTS ===\n=== BEGIN OF OUTPUTS ==="
+    )
+    .expect("Writing to a memory buffer should not fail");
+
+    for output in tx.outputs() {
+        let s = match output {
+            TxOutput::Transfer(val, dest) => {
+                let val_str = fmt_val(val);
+                format!("Transfer({}, {val_str})", fmt_dest(dest))
+            }
```
iljakuklic, line 160: The output printout could be taken out to a separate function in case someone wants to print outputs separately. Similarly with inputs.

iljakuklic (approved):
This looks good to me.

A nuclear option would be to generalise this with traits, roughly like this:
* Introduce a `Summarize` trait that has the same signature as `std::fmt::Display` but takes an extra `chain_config` parameter
* Introduce a wrapper type `Summary(T, &ChainConfig)` which implements `Display` if `T: Summarize`. It now works everywhere you can use `Display`.
* There could be even extension trait that makes the syntax even more convenient, changing from `Summary(my_tx, &cc)` to `my_tx.summary(&cc)`
* The summary trait could be implemented for outputs, inputs, transactions, or whatever else needs to be summarized

The above reuses all the formatting machinery already in Rust and would integrate rather nicely with existing tools, for exmaple:
* `log::debug!("Summary of transaction: {}", my_tx.summary(&cc))`
* same with `println!`, `format!`, `write!` and such
* `my_tx.summary(&cc).to_string()` would work too

TheQuantumPhysicist (commented):
I will do the separation in a separate PR.

PR #1527, opened 2024-02-01
add list transactions command to the wallet

OBorce:
new wallet command to list the confirmed transactions, with an optional filter based on address, and a limit

### DIFF
--- a/wallet/src/account/output_cache/mod.rs
+++ b/wallet/src/account/output_cache/mod.rs
@@ -14,12 +14,14 @@
 // limitations under the License.
 
 use std::{
+    cmp::Reverse,
     collections::{btree_map::Entry, BTreeMap, BTreeSet},
     ops::Add,
 };
 
 use common::{
     chain::{
+        block::timestamp::BlockTimestamp,
         output_value::OutputValue,
         stakelock::StakePoolData,
         tokens::{
@@ -30,7 +32,7 @@ use common::{
         AccountCommand, AccountNonce, AccountSpending, DelegationId, Destination, GenBlock,
         OutPointSourceId, PoolId, Transaction, TxInput, TxOutput, UtxoOutPoint,
     },
-    primitives::{id::WithId, per_thousand::PerThousand, Amount, BlockHeight, Id},
+    primitives::{id::WithId, per_thousand::PerThousand, Amount, BlockHeight, Id, Idable},
 };
 use crypto::vrf::VRFPublicKey;
 use itertools::Itertools;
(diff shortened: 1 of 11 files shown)


wallet/wallet-rpc-lib/src/rpc/mod.rs
```diff
@@ -351,6 +351,29 @@ impl<N: NodeInterface + Clone + Send + Sync + 'static> WalletRpc<N> {
             .await?
     }
 
+    pub async fn confirmed_transactions(
```
TheQuantumPhysicist, line 354: The word "confirmed" is very dangerous to use. We still haven't formalized what "confirmed" really means. In bitcoin, confirmation depends on the number of blocks stacked above the block where the transaction resides. You'll notice also that "confirmed" always comes accompanied with a parameter "depth" or "number of confirmation", as a filter, exactly for this reason.

For us, it's much more complicated and there's a fair amount of math involved to decide what a "confirmed" transaction really is.

I would feel more comfortable calling this `mainchain_transactions`, because that doesn't give any sense of "confirmed". We can go back to the problem of "confirmations" later.

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -650,6 +650,16 @@ pub enum WalletCommand {
     #[clap(name = "transaction-list-pending")]
     ListPendingTransactions,
 
+    /// List confirmed transactions with optional address filter
+    #[clap(name = "transaction-list-confirmed")]
```
TheQuantumPhysicist, line 654: transaction-list-by-address

wallet/src/account/output_cache/mod.rs
```diff
@@ -1173,6 +1173,57 @@ impl OutputCache {
             .collect()
     }
 
+    pub fn confirmed_transactions(
+        &self,
+        destination: Option<Destination>,
+        limit: usize,
+    ) -> Vec<WithId<&Transaction>> {
+        self.txs
```
TheQuantumPhysicist, line 1181: Is it possible to order these transactions in descending order, from newest to oldest?

wallet/src/wallet/tests.rs
```diff
@@ -1078,6 +1078,70 @@ fn wallet_get_transaction(#[case] seed: Seed) {
     assert_eq!(found_tx.get_transaction(), tx.transaction());
 }
 
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn wallet_list_confirmed_transactions(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 1084: name

TheQuantumPhysicist (approved):
Besides the minor comment, good to go.

azarovh (approved):

PR #1528, opened 2024-02-02
Add utxo inputs to transaction endpoint and block height

OBorce:
- add utxos to transaction inputs so users can see what is being spent
- add optional block height to the block endpoint


TheQuantumPhysicist (approved):
Besides the db version bump, looking good.

azarovh (approved):

PR #1530, opened 2024-02-02
Create TextSummary trait

TheQuantumPhysicist:


iljakuklic (approved):
Looks good. I think it would be better to use `std::fmt::Write` to avoid all the intermediate strings but that's not that important.

PR #1532, opened 2024-02-03
[draft] update docker images and provide examples

erubboli:
This PR adds more images for `api-blockchain-scanner-daemon`, `api-web-server`, `dns_server`, `wallet-rpc-daemon` and removes the image for `node-gui`.

TODO:

- [ ] Ensure all ports are correctly exposed (default to mainnet).
- [ ] Add/fix documentation.
- [ ] Add an example to run a whole block explorer in a docker-compose file.
- [ ] Fix binding host on api-web-server and, if necessary, the other tools (e.g., `--rpc-address node-daemon:3030` should work).
- [ ] In Docker, it is very common to pass configuration details using environment variables so that this can work:
```
api-web-server:
  image: mintlayer/api-web-server:0.3.0
  environment: 
    - RPC_ADDRESS=node-daemon:3030 
    - POSTGRES_HOST=postgres 
    - POSTGRES_PASSWORD=use-strong-password 
    - POSTGRES_DATABASE=mintlayer_chain
```


ImplOfAnImpl:
So, I've taken over this PR and pushed some commits. The PR is curently based on the "fix/locked-utxo-conflict" branch, because that branch contains some fixes required for api scanner daemon to work properly.

1) Block explorer example hasn't been and will not be added here. Though technically it's not hard, Sam is against it. The main reason is that the block explorer is not core devs' responsibility; we don't want to have to support it. The better way should be to put the corresponding example into the block explorer's own repository.
2) All services are put into the same docker compose file; users are supposed to copy the directory and comment out the services they don't need.
The exception is dns-server, which has its own separate compose project.
(This is a Sam's suggestion as well; initially I put unrelated services into diferent yml files and it worked, but it was kind of inconvenient to use).
3) Instead of using a named volume, I just map the home directories of containers to a directory on the host machine; the latter is always a subdirectory of the current directory (i.e. the one that contains the docker compose files). This way, the user will have a predictable location for their data (wallet files in particular). Again, this approach assumes that the user will copy the docker compose project to a separate location.
4) The host system's directory where containers' home directory is mounted must be writable by the containers. It'll be the responsibility of the user to ensure it is. To make it possible, dockerfiles create 'user' and 'group' with the explicit uid/gid of 1000. Chances are that host system's 'user' will have the ids of 1000 as well, so nothing will have to be done in this case. But even if not, at least the user will know which 'user' should be granted access to the mount point.
5) I also removed the EXPOSE lines from docker files, because they don't do anything except for serving as documentation and IMO they are more misleading than helpful.


Regarding the question whether we should have the ability to run wallet-cli via docker compose.
1) I don't see why not. It's useful to have.
2) Running interactive shells via docker compose is possible and people do it (in the context of crypto wallets too).
There are basically 2 ways to do it - via `docker compose run`, which starts a new container based on a service definition, and `docker compose exec`, which runs a command in an already running container.
The "run" approach was the first I tried. I had to have an additional yml file for the wallet-cli service in this case, which then would have to be run like this:
`docker compose -f docker-compose.wallet-cli.yml -f docker-compose.yml run --rm wallet-cli`
But in the end I decided to put the wallet-cli binary into the same image as node-daemon, so now it's possible to run it via "exec":
`docker compose exec node-daemon wallet-cli`
And the separate `docker-compose.wallet-cli.yml` is not needed anymore.
Also, because the separate docker image for wallet-cli has become redundant, I removed it.

build-tools/docker/Dockerfile.node-daemon
```diff
@@ -1,16 +1,26 @@
...
+RUN groupadd -g 1000 mintlayer && \
+    useradd -u 1000 -g mintlayer -d /home/mintlayer -s /bin/bash -c "Mintlayer User" mintlayer
+
+# Create and set the home directory for the new user
+RUN mkdir /home/mintlayer && \
+    chown mintlayer:mintlayer /home/mintlayer
 
-# Node daemon listens on ports 13030 and 13031
-EXPOSE 13030 13031
+COPY --from=builder /usr/src/target/release/node-daemon /usr/src/target/release/wallet-cli /usr/bin/
```
ImplOfAnImpl, line 15: Since the "node-daemon" image now contains wallet-cli as well, should we rename it, e.g. to just "node"?

build-tools/docker/example-mainnet/docker-compose.yml
```diff
@@ -0,0 +1,101 @@
...
+    image: $DOCKERHUB_USERNAME/wallet-rpc-daemon:$ML_SOFTWARE_VERSION
+    volumes:
+      - ./container_home:/home/mintlayer
+    command: wallet-rpc-daemon
+    depends_on:
+      - node-daemon
+    environment:
+      - ML_WALLET_RPC_DAEMON_NODE_RPC_ADDRESS=node-daemon:3030
+      # TODO: should wallet-rpc-daemon use the default cookie path automatically, like wallet-cli does?
+      - ML_WALLET_RPC_DAEMON_NODE_COOKIE_FILE=/home/mintlayer/.mintlayer/mainnet/.cookie
```
ImplOfAnImpl, line 38: So, `wallet-rpc-daemon` behaves differently from `wallet-cli` with respect to how the rpc-related parameters are handled. In `wallet-cli` if neither username/password nor cookie file is specified, a cookie file at the default location is assumed. But in `wallet-rpc-daemon` the absence of the parameters means "no authentication". Which is kind of strange, because `wallet-rpc-daemon` has a dedicated parameter for this (rpc_no_authentication). Should I go ahead and make `wallet-rpc-daemon` behave like  `wallet-cli` instead?

wallet/wallet-rpc-lib/src/cmdline.rs
```diff
@@ -36,6 +36,7 @@ use crate::config::{WalletRpcConfig, WalletServiceConfig};
     group(
         clap::ArgGroup::new("rpc_auth")
             .args(["rpc_cookie_file", "rpc_username", "rpc_password", "rpc_no_authentication"])
+            .multiple(true)
             .required(true)
     ),
 )]
```
ImplOfAnImpl, line 42: FYI: without `multiple(true)`, only one of the `rpc_cookie_file`, `rpc_username`, `rpc_password`, `rpc_no_authentication` options can be specified at the same time (e.g. it's impossible to specify both `rpc_username` and `rpc_password`)

build-tools/docker/example-mainnet/docker-compose.yml
TheQuantumPhysicist: I would prefer the default to be the user setting a default username and password. Put a comment above it saying something like "WARNING: This must be a strong password as it gives access to the wallet", or something like that.

build-tools/docker/example-mainnet/.env
```diff
@@ -0,0 +1,36 @@
...
+# run in a dedicated docker container using a dedicated docker volume, the names don't matter much.
+# Just note that the db will be created when the volume is first initialized, so changing
+# the name later won't work.
+API_SERVER_POSTGRES_USER=postgres
+API_SERVER_POSTGRES_DB=postgres
+# The password for the postgres dbms.
+API_SERVER_POSTGRES_PASSWORD=use-strong-password
+
+# The node's rpc and p2p ports will be mapped to these ports on the host machine.
+NODE_RPC_HOST_PORT=3030
```
TheQuantumPhysicist, line 22: It's a bad idea to provide that option. People will think they can change this. They shouldn't.

Now I understand that when a proxy is used, one should set this at the external network interface, but if we want to keep this parameter accessible, we should put a big warning above it that the external port at the routable ip address (in case it's desired to reach this server from the outside) must be 3030, otherwise it won't be reachable.

build-tools/docker/example-mainnet/.env
```diff
@@ -0,0 +1,36 @@
...
+NODE_P2P_HOST_PORT=3031
```
TheQuantumPhysicist, line 23: It's a bad idea to provide that option. People will think they can change this. They shouldn't.

Now I understand that when a proxy is used, one should set this at the external network interface, but if we want to keep this parameter accessible, we should put a big warning above it that the external port at the routable ip address (in case it's desired to reach this server from the outside) must be 3031, otherwise it won't be reachable.

build-tools/docker/example-mainnet/docker-compose.yml
```diff
@@ -0,0 +1,101 @@
+version: '3'
+
+services:
+  node-daemon:
+    image: $DOCKERHUB_USERNAME/node-daemon:$ML_SOFTWARE_VERSION
+    volumes:
+      - ./container_home:/home/mintlayer
```
TheQuantumPhysicist, line 7: I would call it `./mintlayer-data`. We should use names that minimize the need to change.

build-tools/docker/Dockerfile.node-daemon
TheQuantumPhysicist: I don't prefer renaming executables.

build-tools/docker/Dockerfile.wallet-rpc-daemon
```diff
@@ -0,0 +1,28 @@
+# Build Stage
+FROM mintlayer-builder:latest AS builder
+
+# Runtime Stage
+FROM debian:bookworm-slim
+
+# Create a new user and group
+RUN groupadd -g 1000 mintlayer && \
+    useradd -u 1000 -g mintlayer -d /home/mintlayer -s /bin/bash -c "Mintlayer User" mintlayer
+
```
TheQuantumPhysicist, line 10: Please do not use uid 1000. Most people have `sudo` allowed for this user. This provides an attack surface to exit the container in case RCE is possible.

Pick some random uid... maybe 6000 or something.

build-tools/docker/example-mainnet-dns-server/container_home/README.md
```diff
@@ -0,0 +1,2 @@
+Container's home directory will be mapped to this directory.
+Make sure it's writable by the user with uid 1000.
```
TheQuantumPhysicist, line 2: Maybe grep 1000 to not forget any instance.

TheQuantumPhysicist (approved):
Besides the comments, this looks fine. I want Enrico to review it though to see if this is what he expects.

build-tools/docker/Dockerfile.node-daemon
anyxem: I have concerns about combining node-daemon with wallet. There is a use case when wallet in not needed, eg for running api-server. Running two or more containers could be done via docker-compose

ImplOfAnImpl: I reverted this

build-tools/docker/example-mainnet/.env
ImplOfAnImpl: Actually, this should work if the listening port inside the container is the same as the one exposed on the host machine. I've made the corresponding changes in docker-compose.yml and updated the comment there.

ImplOfAnImpl:
I've revamped dockerfiles once more. Now they don't use the `USER` directive; instead, there is the `entrypoint.sh` script, which is run as root initially and which then creates the user, `chown`'s the mounted `mintlayer-data` directory and runs the passed executable as the user. 
Basically, the main reason for this script to exist is the abovementioned `chown` call (because there is no other place to do this except as a part of `CMD` or `ENTRYPOINT` directive). But it also adds the possibility to specify the user id in the environment instead of hard-coding it in the docker file.

build-tools/docker/example-mainnet/docker-compose.yml
ImplOfAnImpl: > I would prefer the default to be the user setting a default username and password. Put a comment above it saying something like "WARNING: This must be a strong password as it gives access to the wallet", or something like that.

I've changed the auth method for node rpc from the cookie to username/password. Regarding the comment - I'm not sure it's needed, because the default passwords ('use-strong-password', 'use-another-strong-password', 'use-yet-another-strong-password') already suggest that.

build-tools/docker/example-mainnet/.env
TheQuantumPhysicist: Like explained before, this depends on whether you're doing network=host, and if the container behind its own NAT (which is the case if network=host isn't mentioned), changing the port won't have an effect. There's more to this than simply changing the port.

build-tools/docker/example-mainnet/docker-compose.yml
```diff
@@ -0,0 +1,132 @@
...
+      ML_MAINNET_NODE_P2P_BIND_ADDRESSES: 0.0.0.0:$NODE_P2P_HOST_PORT
+      # Rpc username and password.
+      ML_MAINNET_NODE_RPC_USERNAME: $NODE_RPC_USERNAME
+      ML_MAINNET_NODE_RPC_PASSWORD: $NODE_RPC_PASSWORD
+    ports:
+      # This is only needed if you want to access the node's rpc interface from the host system.
+      - "$NODE_RPC_HOST_PORT:3030"
+      # This is only needed if you want the node to be able to accept incoming p2p connections
+      # from other nodes.
+      - "$NODE_P2P_HOST_PORT:$NODE_P2P_HOST_PORT"
```
TheQuantumPhysicist, line 39: This is some weird port binding... I would understand if 3031 is fixed then you choose the exposed port...

TheQuantumPhysicist (approved):
Besides the p2p port issue, this is looking OK at this point.

build-tools/docker/example-mainnet/.env
ImplOfAnImpl: [discussed] I've added a warning comment to `.env`  explaining that changing this port value may not be enough.

build-tools/docker/example-mainnet/docker-compose.yml
ImplOfAnImpl: [discussed] there is a comment above that explains why this is needed. I've moved it closer to this line to make things clearer.

TheQuantumPhysicist (approved):

build-tools/docker/build.py
```diff
@@ -73,23 +76,42 @@ def delete_docker_image(image_name, version):
...
-    delete_docker_image("mintlayer-builder", "latest")
-
-
-def push_instances(version, latest):
-    push_docker_image("mintlayer/node-daemon",version , latest)
-    push_docker_image("mintlayer/node-gui",version , latest)
-    push_docker_image("mintlayer/wallet-cli",version , latest)
+def build_instances(version, docker_hub_user, num_jobs):
+    if num_jobs is None:
+        num_jobs = len(os.sched_getaffinity(0))
```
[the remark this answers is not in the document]
TheQuantumPhysicist: multiprocessing has a good, easy way to get CPU count, if none of this works.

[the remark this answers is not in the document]
ImplOfAnImpl: > how about we use num_jobs = os.cpu_count() or 1 instead ?

Done

> multiprocessing has a good, easy way to get CPU count

Is seems like `multiprocessing.cpu_count` is just a wrapper for `os.cpu_count`, so I decided to use the latter.

PR #1534, opened 2024-02-05
Remove inputting the mnemonic in command line arguments

TheQuantumPhysicist:
This program should only be used for generating new mnemonics, not for inputting them. Besides, it's a security flaw because it encourages writing the mnemonics in command line arguments, which is a security risk due to them being saved in shell history.

If someone needs this functionality, they should use the cli wallet in cold mode.

### DIFF
--- a/wallet/wallet-address-generator-lib/src/lib.rs
+++ b/wallet/wallet-address-generator-lib/src/lib.rs
@@ -68,7 +68,7 @@ pub struct CliArgs {
     pub network: Network,
 
     /// Number of addresses to generate and display
-    #[clap(long, default_value_t = 1)]
+    #[clap(long, short = 'n', default_value_t = 1)]
     pub address_count: u8,
 }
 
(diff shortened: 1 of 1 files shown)


wallet/wallet-address-generator-lib/src/lib.rs
```diff
@@ -68,11 +68,12 @@ pub struct CliArgs {
     pub network: Network,
 
     /// Number of addresses to generate and display
-    #[clap(long, default_value_t = 1)]
+    #[clap(long, short = 'n', default_value_t = 1)]
     pub address_count: u8,
 
     /// Mnemonic phrase (12, 15, or 24 words as a single quoted argument). If not specified, a new mnemonic phrase is generated and printed.
     #[clap(long)]
+    #[arg(hide = true)]
```
azarovh, line 76: this will hide it from displaying in help. But it still can be used, right? Is it good enough? just asking

azarovh (approved):

wallet/wallet-address-generator-lib/src/lib.rs
TheQuantumPhysicist: In a previous commit I removed it completely, but then found that it's used in an important test to verify address generation. So I decided hiding is good enough. Let's see how that goes.

PR #1535, opened 2024-02-05
Tolerate not having git installed for wallet version

TheQuantumPhysicist:


wallet/build.rs
```diff
@@ -28,8 +28,12 @@ fn main() {
 
     if let Ok(git_hash) = git_head_hash {
         println!("cargo:rustc-env=GIT_HEAD_HASH={}", git_hash);
+    } else {
+        println!("cargo:rustc-env=GIT_HEAD_HASH=");
```
azarovh, line 32: Not that important but "if else" can be replaced with
`println!("cargo:rustc-env=GIT_HEAD_HASH={}", git_hash.unwrap_or(""));`

azarovh (approved):

wallet/build.rs
TheQuantumPhysicist: Done.

PR #1536, opened 2024-02-05
Remove 0x prefix in Display and Debug for Id

TheQuantumPhysicist:


common/src/primitives/id/mod.rs
```diff
@@ -111,7 +111,9 @@ pub struct Id<T> {
 
 impl<T: TypeName> Debug for Id<T> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Id<{}>{{{:?}}}", T::typename_str(), self.hash)
+        let hash_str = format!("{:?}", self.hash);
+        let s = hash_str.strip_prefix("0x").unwrap_or(&hash_str);
+        write!(f, "Id<{}>{{{:?}}}", T::typename_str(), s)
```
ImplOfAnImpl, line 116: `H256` implements `LowerHex`/`UpperHex`, so it'd be a bit better to use one of those instead of printing and then stripping the prefix.
I.e. `write!(f, "Id<{}>{{{:x}}}", T::typename_str(), self.hash)` should do the job.
Same for `Display` below.

ImplOfAnImpl (approved):

common/src/primitives/id/mod.rs
TheQuantumPhysicist: Unfortuantely, for Display with ... this won't work.

iljakuklic (approved):

PR #1540, opened 2024-02-07
Dns server improvements

ImplOfAnImpl:
1) When announcing addresses, we now give preference to nodes with the same software version as the dns server itself.
2) We now clear the dns server storage if its version differs from the current one (same as we do in p2p).
3) The crawler now also asks peers for addresses via `AddrListRequest` every 24 hours (previously we relied only on `AnnounceAddrRequest`s)


dns-server/src/dns_server/mod.rs
```diff
@@ -211,15 +246,69 @@ impl AuthorityImpl {
...
+        let mut other_version_addrs = addrs
+            .iter()
+            .filter(|(_, software_info)| **software_info != same_software_info)
+            .map(|(addr, _)| addr.clone())
+            .choose_multiple(rng, count);
+        other_version_addrs.shuffle(rng);
+
+        #[allow(clippy::float_arithmetic)]
+        let same_version_addrs_preferred_count =
+            (count as f64 * SAME_SOFTWARE_VERSION_PEERS_RATIO) as usize;
```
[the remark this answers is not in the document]
TheQuantumPhysicist: You can also use `PerThousand` to have more stable behavior.

dns-server/src/dns_server/mod.rs
```diff
@@ -70,27 +75,35 @@ const TTL_IP: u32 = 3600;
...
 
-// Maximum number of IPv6 addresses in result
+/// Maximum number of IPv6 addresses in result
 const MAX_IPV6_RECORDS: usize = 14;
 
+/// When publishing addresses, we give preference to nodes that have the same software version as
+/// the dns server itself.
+/// This constant is a number between 0 and 1 that determines how many addresses of same-version
+/// nodes will be returned compared to nodes of any other version.
+const SAME_SOFTWARE_VERSION_PEERS_RATIO: f64 = 0.8;
```
TheQuantumPhysicist, line 88: 0.95 is probably even better.

TheQuantumPhysicist: Can we set this with a command line argument? How hard would that be?

TheQuantumPhysicist: Btw, PerThousand can be taken from a String in both percentage and fraction. So like I said in the other command, PerThousand sounds like a better way to represent this.

dns-server/src/dns_server/mod.rs
```diff
@@ -211,15 +246,69 @@ impl AuthorityImpl {
...
+            .choose_multiple(rng, count);
+        other_version_addrs.shuffle(rng);
+
+        #[allow(clippy::float_arithmetic)]
+        let same_version_addrs_preferred_count =
+            (count as f64 * SAME_SOFTWARE_VERSION_PEERS_RATIO) as usize;
+
+        let mut result = Vec::with_capacity(count);
+
+        // First take the required number of same-version addresses.
```
TheQuantumPhysicist, line 296: I think there's one corner case here that has to be covered with much more care. Before partitioning the software, we have to check the available ratio of new/(old+new) versions. If it's already heigher than `SAME_SOFTWARE_VERSION_PEERS_RATIO`, then nothing needs to be done, and a random selection of the available set is enough. This is because we do not want to increase the probability of picking an old node, which may be a side effect of what we're doing here, but we want to cap it at `1 - SAME_SOFTWARE_VERSION_PEERS_RATIO`.

common/src/primitives/per_thousand.rs
```diff
@@ -78,6 +93,16 @@ impl Decode for PerThousand {
     }
 }
 
+impl Display for PerThousand {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}%",
+            Amount::from_atoms(self.0.into()).into_fixedpoint_str(1)
```
TheQuantumPhysicist, line 101: Why did you remove the into_percentage_str() function? Why not call it in Display?

TheQuantumPhysicist (approved):

common/src/primitives/per_thousand.rs
ImplOfAnImpl: I've added `into_percentage_str` back.
Also, I've decided not to call it in `Display`, but instead implement them both via a separate function
```
fn write_as_percentage_str(&self, dest: &mut impl std::fmt::Write) -> std::fmt::Result {
    write!(
        dest,
        "{}%",
        Amount::from_atoms(self.0.into()).into_fixedpoint_str(1)
    )
}
```
This may look kind of redundant, but it's also a bit nicer, because `Display` doesn't have to do an unneeded string allocation.

PR #1542, opened 2024-02-07
Add transaction count to block index + upgrade storage version

TheQuantumPhysicist:


chainstate/src/rpc/types/block.rs
```diff
@@ -15,22 +15,28 @@
...
 use common::{
-    chain::Block,
+    chain::{Block, GenBlock},
     primitives::{BlockHeight, Id, Idable},
 };
 
 #[derive(Debug, Clone, serde::Serialize)]
 pub struct RpcBlock {
     id: Id<Block>,
+    prev_block_id: Id<GenBlock>,
```
azarovh, line 25: minor but why store `prev_block_id` and `transaction_count_in_block` when they can be accessed from `block` at any moment with no cost?

azarovh (approved):

chainstate/src/rpc/types/block.rs
TheQuantumPhysicist: This is not storage. This is rpc response. This is what users will see.

PR #1545, opened 2024-02-08
Rename parameters to distinguish between binding addresses and connection addresses

TheQuantumPhysicist:


OBorce (approved):

node-lib/src/config_files/rpc.rs
```diff
@@ -53,30 +53,30 @@ impl RpcConfigFile {
...
         options: &RunOptions,
     ) -> RpcConfigFile {
-        let default_http_rpc_addr = Self::default_bind_address(chain_config);
+        let default_rpc_bind_address = Self::default_bind_address(chain_config);
 
         let RpcConfigFile {
-            http_bind_address,
-            http_enabled,
+            bind_address,
+            rpc_enabled: http_enabled,
```
ImplOfAnImpl, line 60: Makes sense to rename this `http_enabled` too (and the one below)

ImplOfAnImpl (approved):

node-lib/src/config_files/rpc.rs
TheQuantumPhysicist: Fixed in another PR. Sorry I missed this.

PR #1548, opened 2024-02-08
Allowing passing domain addresses as command line parameters - WIP

ImplOfAnImpl:


node-lib/src/config_files/p2p.rs
```diff
@@ -114,14 +114,17 @@ impl From<P2pConfigFile> for P2pConfig {
             sync_stalling_timeout,
             node_type,
             force_dns_query_if_no_global_addresses_known,
-        } = config_file;
+        } = self;
 
-        P2pConfig {
+        let boot_nodes = resolve_all(boot_nodes.unwrap_or_default().into_iter()).await?;
+        let reserved_nodes = resolve_all(reserved_nodes.unwrap_or_default().into_iter()).await?;
```
ImplOfAnImpl, line 120: I wonder if we should use `resolve_all` or `resolve_all_take_first` here; the first function returns all addresses for each name and the second one only returns one address for each name.

p2p/types/src/network_address.rs
```diff
@@ -0,0 +1,452 @@
...
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, DeserializeFromStr, SerializeDisplay)]
+pub struct NetworkAddressWithOptionalPort {
+    pub address: NetworkAddress,
+    pub port: Option<u16>,
+}
```
TheQuantumPhysicist, line 136: This is actually something I wanted to do yesterday, but as usual, I never get the time to do these nicities. Thank you.

The problem I wanted to solve is that when specifying a host without a port for binding, it should use the default port. For example, if the default bind address for RPC is 127.0.0.1:3030, and I specify in command line arguments 0.0.0.0, it should automatically become 0.0.0.0:3030.

TheQuantumPhysicist: Maybe this should be taken out to be used by all crates?

node-lib/src/config_files/p2p.rs
TheQuantumPhysicist: It's wrong to do the name resolution at the config spot. The nodes should be stored as hostnames and resolved when the connection is to be made. This is the whole point of domain names + TTL. We're basically ignoring TTL by doing this.

p2p/types/src/network_address.rs
```diff
@@ -0,0 +1,452 @@
...
+        }
+
+        let separator = ':';
+        assert!(separator.is_ascii());
+        let separator_len = 1;
+
+        // Use rfind instead of find to produce better errors; e.g. for something like
+        // '[1:2:3:4:5:6:7:8]:foo' we'd prefer the error 'BadPortNumber("foo")' instead of
+        // 'BadPortNumber("2:3:4:5:6:7:8]:foo")'
+        let separator_pos = s.rfind(separator);
```
TheQuantumPhysicist, line 183: What if there's no port number specified and it's only an IPv6 address?

node-lib/src/options.rs
```diff
@@ -121,12 +122,29 @@ pub struct RunOptions {
...
-    #[clap(long, value_name = "NODE")]
-    pub p2p_boot_node: Option<Vec<IpOrSocketAddress>>,
+    /// Can be specified multiple times and/or be a comma-separated list.
+    #[clap(
+        long,
+        value_name = "NODE",
+        env = "MINTLAYER_P2P_BOOT_NODE",
+        // Note: comma is the default separator, but we have to specify it explicitly,
+        // because the vec is wrapped in Option.
+        // FIXME: get rid of Option?
```
TheQuantumPhysicist, line 132: No the option means that not specifying it is OK. Everything here should be an option. Am I missing something?

node-lib/src/config_files/p2p.rs
ImplOfAnImpl: Yes, but I was assuming that it would require a large enough refactoring, which we wouldn't want to do. Or would we?

p2p/types/src/network_address.rs
ImplOfAnImpl: > What if there's no port number specified and it's only an IPv6 address?

Then it should be specified without brackets. There is a test for it.

node-lib/src/options.rs
ImplOfAnImpl: > No the option means that not specifying it is OK. Everything here should be an option. Am I missing something?

Well, it's vector and zero length would mean the absence of a value too.

node-lib/src/config_files/p2p.rs
TheQuantumPhysicist: I mean, we shouldn't do it incorrectly... that big refactoring? Isn't it just a change of a type?

node-lib/src/options.rs
TheQuantumPhysicist: Actually an empty vector to being equivalent to no value is a bug in clap, I'd say. A None should mean that the value isn't present in the arguments. An empty vector should mean that it was present and was selected to be empty. But if it's like that, then sure, you can remove the option. Though keeping the symmetry is a nice thing.

ImplOfAnImpl:
Most of the changes from this PR have been merged via other PRs - #1561, #1565, #1566. What was not merged is the ability to pass domain names for "boot nodes" and "reserved nodes":
a) For "boot nodes" it doesn't seem like a good idea at all, because those are just some initial values for the peer database and they must be socket addresses.
b) For "reserved nodes" it seems to require a noticeable amount of refactoring (if we resolve the names at connect time, that is), because currently PeerDb/PeerManager depend of reserved nodes' addresses being normal socket addresses; the same is true for the dns server/crawler. So it's been decided to postpone this.

PR #1549, opened 2024-02-09
Feature/cli wallet connect to rpc wallet

OBorce:
Make the wallet generic so that it can be both RPC or local wallet in the command handler.


wallet/wallet-rpc-client/src/rpc_client/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum WalletRpcError {
+    #[error("Initialization error: {0}")]
+    InitializationError(Box<WalletRpcError>),
+    #[error("Decoding error: {0}")]
+    DecodingError(#[from] serialization::hex::HexError),
+    #[error("Client creation error: {0}")]
+    ClientCreationError(jsonrpsee::core::ClientError),
+    #[error("Response error: {0}")]
+    ResponseError(jsonrpsee::core::ClientError),
```
iljakuklic, line 33: Rather than using jsonrpsee directly here (and adding codecheck exception), re-export the type from the `rpc` crate and use that.

test/functional/test_framework/wallet_cli_controller.py
```diff
@@ -207,12 +215,16 @@ async def new_address(self) -> str:
...
-    async def get_transaction(self, tx_id: str) -> str:
-        return await self._write_command(f"transaction-get {tx_id}\n")
+        return [UtxoOutpoint(id=match["outpoint"]["id"]["Transaction"], index=int(match["outpoint"]["index"])) for match in j]
+
+    async def get_transaction(self, tx_id: str):
+        out = await self._write_command(f"transaction-get {tx_id}\n")
+        try:
+            return json.loads(out)
+        except:
+            return out
```
TheQuantumPhysicist, line 227: Why is this necessary? Shouldn't it be either one or the other?

wallet/wallet-cli-lib/src/config.rs
```diff
@@ -135,6 +135,22 @@ pub struct CliArgs {
...
+    #[clap(long)]
+    pub connect_to_rpc_wallet_cookie_file: Option<String>,
+
+    /// RPC username (either provide a username and password, or use a cookie file. You cannot use both)
+    #[clap(long)]
+    pub connect_to_rpc_wallet_username: Option<String>,
+
+    /// RPC password (either provide a username and password, or use a cookie file. You cannot use both)
+    #[clap(long)]
+    pub connect_to_rpc_wallet_password: Option<String>,
```
TheQuantumPhysicist, line 153: Please change `connect_to` to `remote` in all these parameters.

wallet/wallet-cli-lib/src/config.rs
```diff
@@ -135,6 +135,22 @@ pub struct CliArgs {
     /// Enable running the wallet service without RPC authentication (assuming wallet RPC server is enabled)
     #[arg(long, conflicts_with_all(["wallet_rpc_password", "wallet_rpc_username", "wallet_rpc_cookie_file"]))]
     pub wallet_rpc_no_authentication: bool,
+
+    /// Optional RPC address
```
TheQuantumPhysicist, line 139: "Optionally, the wallet CLI can only be an interface to another remote wallet, accessible through RPC. So, you can start an RPC daemon, and make this CLI connect to it to control it. This is useful for servers, where the RPC wallet can be left staking, and the wallet CLI is used to control its state."

wallet/wallet-cli-lib/src/config.rs
```diff
@@ -135,6 +135,22 @@ pub struct CliArgs {
...
+    #[arg(long, conflicts_with_all(["wallet_rpc_password", "wallet_rpc_username", "wallet_rpc_cookie_file", "wallet_rpc_no_authentication", "wallet_rpc_bind_address", "node_rpc_address", "node_rpc_cookie_file", "node_rpc_username", "node_rpc_password"]))]
+    pub connect_to_rpc_wallet_address: Option<String>,
```
iljakuklic, line 141: You can define an "option group" in clap to specify a collection of options. I believe you can then use te group in the `conflicts_with*` calls and similar. See `wallet-rpc-lib/src/cmdline.rs`.

iljakuklic (approved):
Looks good. Into the future, it would be nice to review how many of the abstraction layers we have are actually necessary to avoid the phenomenon of "lasagna code". I am not quite sure about it but it may be the case some of them are in for historical reasons moslty.

TheQuantumPhysicist (commented):
> Looks good. Into the future, it would be nice to review how many of the abstraction layers we have are actually necessary to avoid the phenomenon of "lasagna code". I am not quite sure about it but it may be the case some of them are in for historical reasons moslty. 

I'm hoping this is the last abstraction we'll need for the wallet... we covered all the possible cases so far. 

Btw, Boris still has to fix more stuff. He has to add state updates between the RPC and CLI. This isn't over. There's an inconsistency issue between the states and some abstraction is required to fix that. Not sure whether that's over.

test/functional/test_framework/wallet_cli_controller.py
OBorce: tries to load it as a json, but error messages are not in json format, so it fails and we return the string as is for matching in the test

test/functional/wallet_submit_tx.py
```diff
@@ -131,12 +131,16 @@ async def async_test(self):
...
+            assert_equal(1, len(output["inputs"]))
+            assert_equal(genesis_block_id, output["inputs"][0]["Utxo"]["id"]["BlockReward"])
+            assert_equal(0, output["inputs"][0]["Utxo"]["index"])
+
+            assert_equal(1, len(output["outputs"]))
+            assert_equal(coins_to_send * ATOMS_PER_COIN, output["outputs"][0]["Transfer"][0]["Coin"]["val"])
+
             expected_tx_state = f"state: Confirmed(BlockHeight(1), BlockTimestamp {{ timestamp: {timestamp} }}, 0)"
-            assert_in(expected_tx_state, output)
+            # assert_in(expected_tx_state, output)
```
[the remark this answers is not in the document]
OBorce: fixed

wallet/wallet-cli-lib/src/config.rs
OBorce: renamed

wallet/wallet-cli-lib/src/config.rs
OBorce: replaced

TheQuantumPhysicist (approved):

PR #1555, opened 2024-02-12
Random verification strategy in tests

azarovh:
Selection of transaction verification strategy is not random.

This revealed a problem with how undo data is flushed through hierarchy of transaction verifiers.
Disposable strategy creates dedicated derived verifier per transaction, flushing the data to the base verifier and eventually db. There is no problem when connecting block in such manner because information for every transaction is just accumulated. But when transaction is disconnected information that something was removed somewhere in the hierarchy must be preserved. That's why now it's all covered with CachedOperation.
Moreover block undo is a complex structure so BlockUndo itself is CachedOperation but also every transaction inside is CachedOperation too.


chainstate/tx-verifier/src/transaction_verifier/utxos_undo_cache/cached_utxo_block_undo.rs
```diff
@@ -0,0 +1,238 @@
...
+            Err(UtxosBlockUndoError::TxUndoWithDependency(*tx_id))
+        }
+    }
+
+    pub fn set_block_reward_undo(&mut self, reward_undo: UtxosBlockRewardUndo) {
+        debug_assert!(self.reward_undo.is_none());
+        self.reward_undo = Some(CachedOperation::Write(reward_undo));
+    }
+
+    pub fn combine(&mut self, other: Self) -> Result<(), UtxosBlockUndoError> {
```
TheQuantumPhysicist, line 199: Can you please document the public functions, explaining what they do and their role in the process?

chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs
```diff
@@ -75,6 +77,16 @@ where
         self.storage.get_gen_block_index(block_id)
     }
 
+    fn get_undo_data(
+        &self,
+        tx_source: TransactionSource,
+    ) -> Result<Option<CachedUtxosBlockUndo>, <Self as TransactionVerifierStorageRef>::Error> {
+        match self.utxo_block_undo.data().get(&tx_source) {
+            Some(op) => Ok(op.get().cloned()),
```
TheQuantumPhysicist, line 85: Why not return a reference instead of cloning?

chainstate/tx-verifier/src/transaction_verifier/tokens_accounting_undo_cache/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+    ) -> Result<Option<TxUndo>, ConnectTransactionError>
+    where
+        F: Fn(TransactionSource) -> Result<Option<CachedTokensBlockUndo>, E>,
+        ConnectTransactionError: From<E>,
+    {
+        let block_undo = match self.data.entry(*tx_source) {
+            Entry::Vacant(_) => fetcher_func(*tx_source)?,
+            Entry::Occupied(entry) => match entry.get() {
+                CachedOperation::Write(undo) | CachedOperation::Read(undo) => Some(undo.clone()),
+                CachedOperation::Erase => panic!("already empty"),
```
TheQuantumPhysicist, line 97: I understand this is impossible because it's an undo, but please explain in more detail why this is impossible in expect. Short messages like this aren't a good idea in large contexts.

chainstate/tx-verifier/src/transaction_verifier/pos_accounting_undo_cache/mod.rs
TheQuantumPhysicist, line 1: This and the other structs like it (I counted 3) require their own unit testing... too much mutability going on in there. There should be tests verifying their behavior.

TheQuantumPhysicist (commented):
First round seems fine, besides the comments. I'll take another round at this next week. I'll also ask Lukas to review this.

chainstate/tx-verifier/src/transaction_verifier/hierarchy.rs
azarovh: We are mimicking storage interface here and return everything by value

chainstate/tx-verifier/src/transaction_verifier/pos_accounting_undo_cache/mod.rs
azarovh: Added tests into `chainstate/tx-verifier/src/transaction_verifier/tests` folder

TheQuantumPhysicist: The tests that test isolated behavior of one struct should be in the same module with the struct. This makes it easy to track what is tested and what is not.

If the test tests multiple things together, then we move it to a more general thing.

iljakuklic (commented):
Overall I'm fine with this. A couple of minor suggestions below.

Ultimately, I'd like to see the notion of undoable operations and the notion of cached values to be formalised using a proper abstraction and implemented in terms of traits and/or generics.

chainstate/tx-verifier/src/transaction_verifier/utxos_undo_cache/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    {
+        let block_undo = match self.data.get(tx_source) {
+            Some(op) => match op {
+                CachedOperation::Write(undo) | CachedOperation::Read(undo) => {
+                    Ok(Some(undo.clone()))
+                }
+                CachedOperation::Erase => Ok(None),
+            },
+            None => fetcher_func(*tx_source).map_err(ConnectTransactionError::from),
+        }?;
```
iljakuklic, line 83: Tiny nitpick: I'd probably place the question mark operator after `map_err(...)`. It's closer to the source of the error, so maybe a little easier to see what's going on and you don't need the extra `Ok` wrappers in the `Some` branch.

iljakuklic (approved):

utxo/src/undo.rs
```diff
@@ -86,6 +87,18 @@ impl UtxosTxUndoWithSources {
...
+        (self.utxos, self.sources)
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ConsumedUtxosBlockUndo {
+    pub reward_undo: Option<UtxosBlockRewardUndo>,
+    pub tx_undos: BTreeMap<Id<Transaction>, UtxosTxUndo>,
+    pub child_parent_dependencies: BTreeSet<(Id<Transaction>, Id<Transaction>)>,
+    pub parent_child_dependencies: BTreeSet<(Id<Transaction>, Id<Transaction>)>,
```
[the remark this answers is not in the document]
TheQuantumPhysicist: Well, I'm not sure what dependency information is needed in the mempool more than what's already available, so I'd need more info. This was designed to be used by the mempool as a black box. There's a function can_be_disconnected() (or close to that name) that can be used to know whether a transaction can be logically evicted before evicting a child that depends on it, in order to maintain a correct set of transactions in the mempool. In my head, that's everything the mempool can gain from this.

TheQuantumPhysicist (approved):

PR #1559, opened 2024-02-12
Add passphrase parameter when creating a wallet

OBorce:
Add optional passphrase parameter to the wallet commands

### DIFF
--- a/wallet/wallet-cli-lib/src/commands/command_handler/mod.rs
+++ b/wallet/wallet-cli-lib/src/commands/command_handler/mod.rs
@@ -152,9 +152,22 @@ where
                 self.wallet.update_wallet::<N>().await;
 
                 let msg = match newly_generated_mnemonic {
-                    CreatedWallet::NewlyGeneratedMnemonic(mnemonic) => format!(
-                    "New wallet created successfully\nYour mnemonic: {}\nPlease write it somewhere safe to be able to restore your wallet."
-                , mnemonic),
+                    CreatedWallet::NewlyGeneratedMnemonic(mnemonic, passphrase) => {
+                        let passphrase = if let Some(passphrase) = passphrase {
+                            format!("passphrase: {passphrase}\n")
+                        } else {
+                            String::new()
+                        };
+                        format!(
+                            "New wallet created successfully\nYour mnemonic: {}\n{passphrase}\
+                        Please write it somewhere safe to be able to restore your wallet. \
+                        It's recommended that you attempt to recover the wallet now as practice\
+                        to check that you arrive at the same addresses, \
+                        to ensure that you have done everything correctly.
+                        ",
+                            mnemonic
+                        )
+                    }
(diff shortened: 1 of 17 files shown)


wallet/wallet-rpc-client/src/handles_client/mod.rs
```diff
@@ -103,14 +105,15 @@ impl<N: NodeInterface + Clone + Send + Sync + 'static + Debug> WalletInterface
         path: PathBuf,
         store_seed_phrase: bool,
         mnemonic: Option<String>,
+        passphrase: Option<String>,
     ) -> Result<CreatedWallet, Self::Error> {
```
TheQuantumPhysicist, line 109: This doesn't model the way the wallet work correctly. The case with mnemonic being None and passphrase being Some is possible and this leads to having to manually check this or expect on it, unnecessarily. This should be enforced at the compiler level. Despite being ugly, something like `Option<(String, Option<String>)>` is way more appropriate (with docs explaining it). A struct for this is also OK.

TheQuantumPhysicist (approved):
Besides the parametrization, this looks OK.

Though I would recommend bumping the database version, even though no upgrade is required. Make an upgrade and make it empty. This is because otherwise users could try to open the wallet with an older version and will get an error.

wallet/wallet-rpc-client/src/handles_client/mod.rs
OBorce: I thought the user might want to have a random seed phrase, but have his own passphrase

TheQuantumPhysicist: Actually you're right. This is a valid case. Please document the meaning of these options, if you haven't yet.

wallet/wallet-cli-lib/src/commands/command_handler/mod.rs
```diff
@@ -234,8 +260,12 @@ where
             ColdWalletCommand::PurgeSeedPhrase => {
                 let phrase = self.non_empty_wallet().await?.purge_seed_phrase().await?;
 
-                let msg = if let Some(phrase) = phrase.seed_phrase {
-                    format!("The seed phrase has been deleted, you can store it if you haven't do so yet: \"{}\"", phrase.join(" "))
+                let msg = if let Some(phrase) = phrase {
+                    if let Some(passphrase) = phrase.passphrase {
+                        format!("The seed phrase has been deleted, you can store it if you haven't do so yet: \"{}\" with passphrase \"{}\"", phrase.seed_phrase.join(" "), passphrase)
```
TheQuantumPhysicist, line 265: "haven't done so yet"

TheQuantumPhysicist: Also, in separate lines, please.

OBorce:
will create a new PR with the review comments

PR #1560, opened 2024-02-13
Feature/api server extra info

OBorce:
- Add extra transaction info in blocks (transaction fees and utxos for inputs)
- Add new endpoint for listing latest transactions with offset


TheQuantumPhysicist (approved):
Besides adding the new index for transactions to avoid O(N) complexity, like we discussed, and bumping the version, looks fine.

PR #1561, opened 2024-02-13
Specify "env" for all clap arguments

ImplOfAnImpl:
1) I've added "env" to (almost) all fields of all option structs parsed by `clap`, so that the corresponding arguments can be specified via env vars. 

2) "Singular" names that referred to a list have been renamed to the plural form and allowed to contain a comma-separated list. E.g. before it was
```
/// Reserved node address to connect. Can be specified multiple times.
#[clap(long)]
pub reserved_node: Vec<IpOrSocketAddress>,
```
and now it is
```
/// Reserved node addresses to connect.
/// Can be specified multiple times and/or be a comma-separated list.
#[clap(long, value_delimiter(','), env = "MINTLAYER_RESERVED_NODES")]
pub reserved_nodes: Vec<IpOrSocketAddress>,
```
I did it because from the one hand the names of env vars and command line parameters must be consistent and, on the other hand, it would be ugly for the env var to use a singular name, provided that it can contain a comma-separated list.

I also decided to do a little cleanup of parameters for the node. One of the things was specifying `action = clap::ArgAction::SetTrue` for the flags  `blockprod_skip_ibd_check`, `p2p_disable_noise`, `p2p_force_dns_query_if_no_global_addresses_known`, so instead of writing `--blockprod-skip-ibd-check=true` we now should write `--blockprod-skip-ibd-check`. I did it because specifying flags this way seems more natural and also for consistency with the `clean_data` option, which also uses this style. But the corresponding env vars still has to contain either `true` or `false`. Also, the `rpc_enabled` flag still requires an explicit `true`/`false` (because this value is true by default and the option is used to turn rpc off). I personally don't see this inconsistency as something important, but I had to mention it.


TheQuantumPhysicist (approved):

wallet/wallet-cli-lib/src/config.rs
```diff
@@ -58,82 +58,86 @@ impl WalletCliArgs {
...
 
     /// RPC username (either provide a username and password, or use a cookie file. You cannot use both)
-    #[clap(long)]
+    #[clap(long, env = "MINTLAYER_NODE_RPC_USERNAME")]
     pub node_rpc_username: Option<String>,
 
     /// RPC password (either provide a username and password, or use a cookie file. You cannot use both)
-    #[clap(long)]
+    #[clap(long, env = "MINTLAYER_NODE_RPC_PASSWORD")]
     pub node_rpc_password: Option<String>,
```
iljakuklic, line 86: Would it make sense to make these `MINTLAYER_RPC_*`  rather than `MINTLAYER_NODE_RPC_*`? That way, we share the variable between node and wallet, so the wallet connects to the node out of the box if run under the same env variables.

common/src/chain/config/regtest_options.rs
```diff
@@ -39,55 +39,55 @@ use paste::paste;
...
+    #[clap(long, env = "MINTLAYER_CHAIN_COIN_DECIMALS")]
     pub chain_coin_decimals: Option<u8>,
 
     /// Emission schedule (`<initial_supply>+<initial_subsidy>[, <height>+<subsidy>]`).
     pub chain_emission_schedule: Option<String>,
 
     /// The maximum block header size in bytes.
-    #[clap(long)]
+    #[clap(long, env = "MINTLAYER_CHAIN_MAX_BLOCK_HEADER_SIZE")]
     pub chain_max_block_header_size: Option<usize>,
```
iljakuklic, line 66: I'd probably slightly prefer `MINTLAYER_REGTEST_*` here in place of `MINTLAYER_CHAIN_*`. Thoughts?

wallet/wallet-rpc-lib/src/cmdline.rs
```diff
@@ -47,51 +47,82 @@ pub struct Args {
...
     /// Password for RPC server basic authorization.
     /// If not set, the RPC cookie file is created.
-    #[arg(long, value_name("PASS"), requires("rpc_username"))]
+    #[arg(
+        long,
+        value_name("PASS"),
+        requires("rpc_username"),
+        env = "MINTLAYER_RPC_PASSWORD"
+    )]
     rpc_password: Option<String>,
```
iljakuklic, line 118: These should be double checked. I think these credentials are used to connect to the wallet RPC, not to node RPC. However these variables are already used to set up node RPC authentication.

iljakuklic (commented):
Looks good. We need to be careful about RPC credentials though. There is a potential inconsistency between RPC credentials for the node and RPC credentials for the wallet.

wallet/wallet-cli-lib/src/config.rs
TheQuantumPhysicist: I wouldn't care that much about that. One could, in a script or so, assign one to the value of the other. I'd care more about the consistency of naming.

common/src/chain/config/regtest_options.rs
TheQuantumPhysicist: If it's regtest-specific, then sure. But I'd rather change both names, not just the env var value.

wallet/wallet-rpc-lib/src/cmdline.rs
TheQuantumPhysicist: The node RPC has the variables with prefix `node_*`. I'm open to a better naming scheme. I already gave it a shot to improve the naming, but we still have a chance to improve them further before the next release, after which we'll try to keep them fixed for interface stability.

wallet/wallet-cli-lib/src/config.rs
iljakuklic: > I wouldn't care that much about that. One could, in a script or so, assign one to the value of the other. I'd care more about the consistency of naming.

My thinking here is that with command line parameters, you pass them to each command separately, whereas with env variables, you can assign them once and invoke multiple commands with the same values. Therefore, it is preferable that variables with the same name specify the same thing (e.g. node RPC credentials). As a consequence, the consistency of naming is now a bit more important to achieve across multiple binaries too.

common/src/chain/config/regtest_options.rs
iljakuklic: > If it's regtest-specific, then sure. But I'd rather change both names, not just the env var value.

Hmm, ok. Changing switch names here might be a bit of a hassle, I don't have a strong opinion about this.

wallet/wallet-cli-lib/src/config.rs
TheQuantumPhysicist: I understand, but I'm more afraid of misunderstandings and conflicts than convenience. I'd rather have people get something not working because they didn't configure it right than not working due to some side effect we haven't thought through.

ImplOfAnImpl:
Turns out, clap has this "mut_args" feature, which allows to modify arguments all at once, so individual "env" attributes are not needed. Also, it solves the problem of some option structs being re-used in multiple places via "flatten"; with this approach, the "flattened" options will get their own env vars.

TheQuantumPhysicist (approved):

node-lib/src/options.rs
```diff
@@ -25,14 +25,17 @@ use std::{
...
 
 const CONFIG_NAME: &str = "config.toml";
 
 /// Mintlayer node executable
+// Note: this struct is shared between different node executables, namely, node-daemon and node-gui,
+// so the env vars for both of them will use the same infix; this is intended.
 #[derive(Parser, Debug)]
+#[clap(mut_args(clap_utils::env_adder("NODE")))]
 #[clap(author, version, about)]
 pub struct Options {
```
ImplOfAnImpl, line 40: I turned out that this doesn't work for subcommands. Instead, I have to specify it for each subcommand individually as well:
```
#[derive(Subcommand, Clone, Debug)]
pub enum Command {
    /// Run the mainnet node.
    #[clap(mut_args(clap_utils::env_adder("NODE")))]
    Mainnet(RunOptions),
    /// Run the testnet node.
    #[clap(mut_args(clap_utils::env_adder("NODE")))]
    Testnet(RunOptions),
    /// Run the regtest node.
    #[clap(mut_args(clap_utils::env_adder("NODE")))]
    Regtest(Box<RegtestOptions>),
}
```
Which makes me wonder whether we should use different infixes for different networks then. E.g.
```
#[derive(Subcommand, Clone, Debug)]
pub enum Command {
    /// Run the mainnet node.
    #[clap(mut_args(clap_utils::env_adder("MAINNET_NODE")))]
    Mainnet(RunOptions),
    /// Run the testnet node.
    #[clap(mut_args(clap_utils::env_adder("TESTNET_NODE")))]
    Testnet(RunOptions),
    /// Run the regtest node.
    #[clap(mut_args(clap_utils::env_adder("REGTEST_NODE")))]
    Regtest(Box<RegtestOptions>),
}
```
The env var names will then look like `ML_MAINNET_NODE_P2P_BIND_ADDRESSES`, `ML_TESTNET_NODE_P2P_BIND_ADDRESSES` etc. @TheQuantumPhysicist what do you think?

In the wallet it's a bit different btw. There, the same options are present both on the top-most and subcommand level. So if we do the same for the corresponding subcommand struct in wallet_cli_lib::config:
```
#[derive(Subcommand, Clone, Debug)]
pub enum Network {
    #[clap(mut_args(clap_utils::env_adder("MAINNET_WALLET")))]
    Mainnet(CliArgs),
    #[clap(mut_args(clap_utils::env_adder("TESTNET_WALLET")))]
    Testnet(CliArgs),
    #[clap(mut_args(clap_utils::env_adder("REGTEST_WALLET")))]
    Regtest(Box<RegtestOptions>),
    #[clap(mut_args(clap_utils::env_adder("SIGNET_WALLET")))]
    Signet(CliArgs),
}
```
while still having `#[clap(mut_args(clap_utils::env_adder("WALLET")))]` on `WalletCliArgs`, we'll have both kinds of names. E.g. we'll have `ML_WALLET_NODE_RPC_ADDRESS` for the case when the wallet is run without a subcommand, `ML_TESTNET_WALLET_NODE_RPC_ADDRESS` when it's run with "testnet" etc. But it still makes sense.

P.S. I've just noticed that the infixes that we use sometimes result in some strange env var names. E.g. for the `--wallet-file` option, the env var is named `ML_WALLET_WALLET_FILE`. Perhaps we should use more "schematic" infixes, e.g. `WLT` instead of `WALLET`, so that the var name would become `ML_WLT_WALLET_FILE`.

PR #1565, opened 2024-02-16
Allow domain names instead of ip addresses for some command line parameters

ImplOfAnImpl:
`NetworkAddressWithPort` is now used for destination rpc_address'es instead of `SocketAddr`.
Also, `SocketAddr` is used instead of String for some bind addresses, for early address validation.
Plus minor cleanup.

P.S. NetworkAddressXXX stuff was mostly taken from the PR #1548 (with some modifications). `NetworkAddressWithOptionalPort` is not currently used, but I expect to make use of it in the next PR.


utils/networking/src/network_address.rs
```diff
@@ -0,0 +1,302 @@
...
+use serde_with::{DeserializeFromStr, SerializeDisplay};
+use thiserror::Error;
+
+use crate::{
+    ip_or_socket_address::IpOrSocketAddress,
+    resolvable_name::{NameResolutionError, ResolvableName},
+};
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DomainName(String);
```
TheQuantumPhysicist, line 32: I've been back and forth in my mind to whether I should say this or just let it go... but the correct name is probably Hostname, not DomainName... the reason is that a domain is usually a superset of what hostnames are or where they are. Maybe I'm missing something, I don't know. But I find the name Hostname much more appropriate for a simple name that's resolvable using a DNS server.

TheQuantumPhysicist: https://www.baeldung.com/wp-content/uploads/sites/4/2022/02/szfafsd.drawio.png

utils/networking/src/network_address.rs
```diff
@@ -0,0 +1,302 @@
...
+
+impl FromStr for NetworkAddress {
+    type Err = NetworkAddressParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Ok(ip) = s.parse::<IpAddr>() {
+            Ok(NetworkAddress::Ip(ip))
+        } else {
+            Ok(NetworkAddress::Domain(s.parse::<DomainName>()?))
+        }
```
TheQuantumPhysicist, line 115: I think we should do a little better than this... but if it's too much trouble, I'm OK with what we have. How hard is it to do a simple character check? We should only allow characters for FQDNs:

Letters (a-z, A-Z)
Digits (0-9)
Hyphens (-) - However, hyphens are not allowed as the first or last character of a label (part between the dots), nor can they appear consecutively.
Periods (.) - Periods are used to separate the labels in the domain name.

From the RFC: "The labels must follow the rules for ARPANET host names.  They must start with a letter, end with a letter or digit, and have as interior characters only letters, digits, and hyphen.  There are also some restrictions on the length.  Labels must be 63 characters or less.".

https://www.rfc-editor.org/rfc/rfc1035.txt

utils/networking/src/lib.rs
```diff
@@ -17,4 +17,9 @@
 
 pub mod broadcaster;
 pub mod ip_or_socket_address;
+pub mod network_address;
+pub mod resolvable_name;
+
 pub use ip_or_socket_address::*;
+pub use network_address::*;
+pub use resolvable_name::*;
```
TheQuantumPhysicist, line 25: Same problem with the other PR. We shouldn't use pub for both modules and re-exports. This is redundant and can cause confusion.

utils/networking/src/network_address.rs
ImplOfAnImpl: Here we delegate to `impl FromStr for DomainName`; there, `addr::parse_domain_name` is used to parse the name, which should check its correctness.
I have a couple of checks for incorrect names in `network_address_printing_parsing`:
```
// Some garbage.
let err = "!@#$%^".parse::<NetworkAddress>().unwrap_err();
assert_eq!(
    err,
    NetworkAddressParseError::NotIpAddressOrDomainName("!@#$%^".to_owned())
);

// "Ipv6 address in brackets" (without a port) is not a valid address or domain name.
let err = "[1:2:3:4:5:6:7:8]".parse::<NetworkAddress>().unwrap_err();
assert_eq!(
    err,
    NetworkAddressParseError::NotIpAddressOrDomainName("[1:2:3:4:5:6:7:8]".to_owned())
);
``` 
I can add more if you think we need them.

utils/networking/src/lib.rs
ImplOfAnImpl: I've fixed this in a follow-up RP.

utils/networking/src/network_address.rs
ImplOfAnImpl: [discussed] I've left it as is.

utils/networking/src/network_address.rs
ImplOfAnImpl: I've left it as is.

PR #1569, opened 2024-02-19
Fix fetching UTXOs from txs within same block

OBorce:


api-server/stack-test-suite/tests/v1/transaction.rs
```diff
@@ -46,6 +46,241 @@ async fn transaction_not_found() {
...
+                        Some(signed_tx1.outputs()[outpoint.output_index() as usize].clone())
+                    }
+                    TxInput::Account(_) | TxInput::AccountCommand(_, _) => None,
+                });
+
+                let transaction = signed_tx2.transaction();
+
+                let expected_transaction = json!({
+                "block_id": block_id.to_hash().encode_hex::<String>(),
+                "timestamp": block.timestamp().to_string(),
```
iljakuklic, line 187: The formatting here is a bit weird. Is this what rustfmt produces? I would expect the keys to be indented.

iljakuklic (approved):

api-server/stack-test-suite/tests/v1/transaction.rs
OBorce: it is the json!() macro that is killing the formatting

PR #1574, opened 2024-02-20
Add account rename command to the wallet

OBorce:
- Add account-rename command to rename an existing account
- Add wallet-info command that returns the wallet's account names
- fix RPC wallet functional tests to start with different port

### DIFF
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -1936,10 +1936,10 @@ impl Account {
         &mut self,
         name: Option<String>,
         db_tx: &mut impl WalletStorageWriteLocked,
-    ) -> WalletResult<()> {
+    ) -> WalletResult<(U31, &Option<String>)> {
         self.account_info.set_name(name);
         db_tx.set_account(&self.get_account_id(), &self.account_info)?;
-        Ok(())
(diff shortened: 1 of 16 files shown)


TheQuantumPhysicist (approved):

wallet/src/account/mod.rs
```diff
@@ -1936,10 +1936,10 @@ impl Account {
         &mut self,
         name: Option<String>,
         db_tx: &mut impl WalletStorageWriteLocked,
-    ) -> WalletResult<()> {
+    ) -> WalletResult<(U31, &Option<String>)> {
         self.account_info.set_name(name);
         db_tx.set_account(&self.get_account_id(), &self.account_info)?;
-        Ok(())
+        Ok((self.account_index(), self.name()))
```
iljakuklic, line 1942: What's the idea behind introducing the return values? Whoever is calling this already has access to the `self` object from which it could be obtained.

As a side point, I don't think returning an `&Option<String>` makes much sense in general as it does not provide much value over `Option<&str>`.

iljakuklic (approved):

wallet/src/account/mod.rs
OBorce: removed

PR #1579, opened 2024-02-22
Support PoS accounting operations in simulation test

azarovh:
Now `chainstate/test-suite/src/tests/tx_verification_simulation.rs` supports all the functionality and can generate txs with pretty much every possible state on the blockchain (there are some limitations though).

Next goal is to add random reorg to that test which I want to do separately.

### DIFF
--- a/chainstate/test-framework/src/random_tx_maker.rs
+++ b/chainstate/test-framework/src/random_tx_maker.rs
@@ -37,7 +37,9 @@ use crypto::{
 };
 use itertools::Itertools;
 use pos_accounting::{
-    make_delegation_id, make_pool_id, DelegationData, InMemoryPoSAccounting, PoolData,
+    make_pool_id, DelegationData, InMemoryPoSAccounting, PoSAccountingDB, PoSAccountingDelta,
+    PoSAccountingDeltaData, PoSAccountingOperations, PoSAccountingUndo, PoSAccountingView,
+    PoolData,
 };
 use test_utils::nft_utils::*;
 use tokens_accounting::{
@@ -80,8 +82,8 @@ pub struct RandomTxMaker<'a> {
     // Transaction is composed of multiple inputs and outputs
     // but pools, delegations and tokens can be created only using input0 so following flags are required
     token_can_be_issued: Option<TokenId>,
-    stake_pool_can_be_create: Option<PoolId>,
-    delegation_can_be_create: Option<DelegationId>,
+    stake_pool_can_be_created: Option<PoolId>,
+    delegation_can_be_created: Option<UtxoOutPoint>,
 
     account_command_used: bool,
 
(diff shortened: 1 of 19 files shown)


chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -83,21 +134,28 @@ impl<'a> RandomTxMaker<'a> {
...
         mut self,
         rng: &mut (impl Rng + CryptoRng),
-    ) -> (Transaction, TokensAccountingDeltaData) {
+    ) -> (
+        Transaction,
+        TokensAccountingDeltaData,
+        PoSAccountingDeltaData,
+    ) {
         // TODO: ideally all inputs/outputs should be shuffled but it would mess up with token issuance
         // because ids are build from input0
```
TheQuantumPhysicist, line 143: built

chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -83,21 +134,28 @@ impl<'a> RandomTxMaker<'a> {
     pub fn make(
         mut self,
         rng: &mut (impl Rng + CryptoRng),
-    ) -> (Transaction, TokensAccountingDeltaData) {
+    ) -> (
+        Transaction,
+        TokensAccountingDeltaData,
+        PoSAccountingDeltaData,
+    ) {
         // TODO: ideally all inputs/outputs should be shuffled but it would mess up with token issuance
```
TheQuantumPhysicist, line 142: Even though this looks like a small issue, it could have a large effect on randomization. Perhaps you should come back to that when you're done with reorg tests.

chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -141,16 +203,31 @@ impl<'a> RandomTxMaker<'a> {
             .collect()
     }
 
-    fn select_accounts(&self, rng: &mut impl Rng) -> Vec<TokenId> {
+    fn select_accounts(&self, rng: &mut impl Rng) -> Vec<AccountType> {
         // TODO: it take several items from the beginning of the collection assuming that outpoints
         // are ids thus the order changes with new insertions. But more sophisticated random selection can be implemented here
```
TheQuantumPhysicist, line 208: Why not randomize with a slice of references then pick?

TheQuantumPhysicist (approved):

iljakuklic (approved):

chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -21,35 +21,82 @@ use chainstate::chainstate_interface::ChainstateInterface;
...
+    tip_view: &impl PoSAccountingView,
+) -> Option<(&'a DelegationId, &'a DelegationData)> {
+    let all_delegation_data = storage.all_delegation_data();
+    (!all_delegation_data.is_empty())
+        .then(|| {
+            {
+                all_delegation_data
+                    .iter()
+                    .nth(rng.gen_range(0..all_delegation_data.len()))
+                    .unwrap()
```
iljakuklic, line 78: The `crypto::random` module re-exports the `IteratorRandom` trait that can be used to sample from a iterator, or `SliceRandom` which can be used to sample from a slice. Would it make this more readable? (I believe I've seen multiple occurrences of this pattern).

chainstate/test-framework/src/random_tx_maker.rs
azarovh: ok I'll think what I can do there later

PR #1580, opened 2024-02-22
Mempool small refactors

iljakuklic:
A couple of small refactors for mempool.

* Remove a rather pointless trait
* Do not treat duplicate transaction insertion as error, report it as status instead


mempool/types/src/tx_status.rs
```diff
@@ -21,16 +21,45 @@
...
     InMempool,
+
+    /// Transaction has already been in the mempool, duplicate insertion
+    InMempoolDuplicate,
+
     /// Transaction is in orphan pool
     InOrphanPool,
+
+    /// Transaction has already been in the orphan pool, duplicate insertion
+    InOrphanPoolDuplicate,
```
TheQuantumPhysicist, line 32: Just to be clear, this is much better than before, but this may warrant changes in p2p and the wallet to expect different behavior. I'll notify the guys about this.

TheQuantumPhysicist (approved):

mempool/types/src/tx_status.rs
iljakuklic: I made some changes to p2p, I think it's fine now. Not 100% sure about the wallet.

iljakuklic:
Hmm functional test `wallet_delegation` fails, mentioning http connection timeout and node_shutdown failure. I don't see how that's related to my changes. Annoyingly enough, I cannot reproduce it on my machine. Anyone else seeing this?

TheQuantumPhysicist:
> Hmm functional test `wallet_delegation` fails, mentioning http connection timeout and node_shutdown failure. I don't see how that's related to my changes. Annoyingly enough, I cannot reproduce it on my machine. Anyone else seeing this?

There's a spurious failure I found with CT in delegation test and I reported in the wallet channel. Check if it's the same one. Try to rerun CI too. I'm assuming you rebased today.

mempool/types/src/tx_status.rs
ImplOfAnImpl: I don't see any problems in p2p either.

ImplOfAnImpl (approved):

iljakuklic:
> There's a spurious failure I found with CT in delegation test and I reported in the wallet channel. Check if it's the same one. Try to rerun CI too. I'm assuming you rebased today.

Yea looks similar. This is indeed on top of a very recent tip.

OBorce (approved):

mempool/types/src/tx_status.rs
TheQuantumPhysicist: All good. I just want to make sure all related parties agree about the change in the interface behavior that only affects runtime without having unexpected side-effects. That's the scariest thing in software. Cheers!

PR #1589, opened 2024-02-25
Enforce all rules of clippy with errors

TheQuantumPhysicist:
Using only warnings is causing rules slipping

### DIFF
--- a/utils/networking/src/network_address.rs
+++ b/utils/networking/src/network_address.rs
@@ -210,23 +210,20 @@ impl FromStr for NetworkAddressWithOptionalPort {
             });
         }
 
-        let separator = ':';
-        assert!(separator.is_ascii());
-        let separator_len = 1;
-
-        // Use rfind instead of find to produce better errors; e.g. for something like
+        // Use rsplit_once instead of split_once to produce better errors; e.g. for something like
         // '[1:2:3:4:5:6:7:8]:foo' we'd prefer the error 'BadPortNumber("foo")' instead of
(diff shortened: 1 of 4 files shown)


utils/networking/src/network_address.rs
```diff
@@ -224,6 +224,8 @@ impl FromStr for NetworkAddressWithOptionalPort {
         // either 'NotIpAddressOrDomainName("[1")' or 'BadPortNumber("2:3:4:5:6:7:8]")'.
         // The former looks too cryptic, so we choose the latter.
         let port = if let Some(pos) = separator_pos {
+            // Note: the position "pos + separator_len" is guaranteed to be at a character boundary.
+            #[allow(clippy::string_slice)]
             let port_str = &s[pos + separator_len..];
```
iljakuklic, line 229: I think this could be done without slice indexing using `str::rsplit_once`

iljakuklic (approved):

iljakuklic (approved):

build-tools/codecheck/codecheck.py
```diff
@@ -330,6 +330,7 @@ def check_trailing_whitespaces():
     exempted_files = [
         'crypto/src/symkey/chacha20poly1305/XCHACHA20POLY1305_TEST_VECTORS.tv',
         'script/src/test/test_vectors_4opc.csv.gz',
+        'wasm-crypto/pkg/wasm_crypto_bg.wasm',
```
TheQuantumPhysicist, line 333: wasm crypto was recently renamed to wasm wrappers... this should never happen.

PR #1594, opened 2024-02-26
wallet-rpc: Make command line args consistent with other executables

iljakuklic:
Network type is a subcommand rather than a switch like in other tools like the node binary etc.

### DIFF
--- a/wallet/wallet-rpc-lib/src/cmdline.rs
+++ b/wallet/wallet-rpc-lib/src/cmdline.rs
@@ -15,8 +15,6 @@
 
 use std::path::PathBuf;
 
-use clap::Parser;
-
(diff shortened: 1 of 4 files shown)


iljakuklic:
If anyone can think of other places like docs that need updating, please let me know.

Also existing wallet RPC users should be notified about this change.

wallet/wallet-rpc-lib/src/cmdline.rs
```diff
@@ -106,12 +130,14 @@ pub struct WalletRpcDaemonArgs {
     chain_config: ChainConfigOptions,
```
TheQuantumPhysicist, line 130: This should go only into Regtest. Mainnet and Testnet don't need it.

TheQuantumPhysicist (approved):

PR #1597, opened 2024-02-27
Api server add address/all-utxos endpoint

OBorce:
- add new address/all-utxos endpoint
- add available-utxos with different path address/spendable-utxos

### DIFF
--- a/api-server/storage-test-suite/src/basic.rs
+++ b/api-server/storage-test-suite/src/basic.rs
@@ -386,7 +386,7 @@ where
         db_tx.commit().await.unwrap();
     }
 
-    // Test setting/getting address available utxos
+    // Test setting/getting address spendable utxos
     {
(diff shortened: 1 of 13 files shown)


api-server/storage-test-suite/src/basic.rs
```diff
@@ -483,6 +483,103 @@ where
             assert_eq!(bob_utxos, vec![(outpoint.clone(), output.clone())]);
         }
 
+        // get all utxos
```
TheQuantumPhysicist, line 486: Please update the comment above for spendable to "spendable" instead of available

TheQuantumPhysicist (approved):

api-server/storage-test-suite/src/basic.rs
OBorce: fixed

PR #1601, opened 2024-02-28
Include caller info in tap_error_log's messages

ImplOfAnImpl:
Here `tap_error_log::LogError`'s function use `#[track_caller]` to get the location of their call and print it to the log as well.

An example of the output (to get it, I simulated a `BlockError::BlockAlreadyExists` error for 0-id block):
```
2024-02-28T08:24:14.309088Z ERROR process_block{block_id=4cbc…7c6d}: LogError: Block 0000…0000 already exists (chainstate/src/detail/mod.rs:404:10)
```

Here the `process_block{block_id=4cbc…7c6d}` part comes from `tracing::instrument` (PR #1600) and `(chainstate/src/detail/mod.rs:404:10)` is the caller location info that this PR adds.

P.S. in order to shorten the paths (make them relative to the workspace dir), I had to introduce a build script in `utils`, where I obtain the workspace dir.

### DIFF
--- a/utils/src/workspace_path.rs
+++ b/utils/src/workspace_path.rs
@@ -1,0 +1,35 @@
+// Copyright (c) 2021-2024 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Return the path to the workspace, as it was recorded on the build machine. The path has
+/// a trailing path separator.
+///
+/// Note that the path is produced by the build script using PathBuf::to_string_lossy, so if it
+/// contained non-Unicode characters, this value will be an approximation rather than the exact
+/// path itself.
+pub fn workspace_path() -> &'static str {
+    env!("WORKSPACE_PATH")
+}
+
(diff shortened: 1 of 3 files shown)


utils/src/workspace_path.rs
```diff
@@ -0,0 +1,35 @@
...
+// limitations under the License.
+
+/// Return the path to the workspace, as it was recorded on the build machine. The path has
+/// a trailing path separator.
+///
+/// Note that the path is produced by the build script using PathBuf::to_string_lossy, so if it
+/// contained non-Unicode characters, this value will be an approximation rather than the exact
+/// path itself.
+pub fn workspace_path() -> &'static str {
+    env!("WORKSPACE_PATH")
```
iljakuklic, line 23: Hmm so `CARGO_MANIFEST_DIR` is not accessible here directly?

ImplOfAnImpl: > Hmm so CARGO_MANIFEST_DIR is not accessible here directly?

Well, it is accessible, it's just that it contains the "/utils" suffix, so I'd have to make a lazy_static variable and remove the suffix at runtime. When I was writing this it seemed cleaner to do this in a build script. Right now I'm not sure if it's an overkill though.

TheQuantumPhysicist (approved):

utils/src/workspace_path.rs
iljakuklic: Hmm, ok. I usually prefer to avoid build scripts but don't have a very strong opinion in this case. Maybe instead of `lazy_static`, a function that strips whatever suffix is necessary every time would work too.

iljakuklic (approved):

utils/src/workspace_path.rs
ImplOfAnImpl: > Maybe instead of lazy_static, a function that strips whatever suffix is necessary every time would work too.

I'm not opposing lazy_static per se, I just don't like the extra runtime work that is always the same. So, a function that does it every time is not better.

I gave this a thought once more and in the end I think that the build script approach is cleaner. So, since you don't insist on the alternative approach, I'm planning to leave it this way.
In the meantime, I've updated some comments and added a basic test.

PR #1604, opened 2024-02-29
Update readme files

TheQuantumPhysicist:


README.md
```diff
@@ -18,17 +18,30 @@ For non-security related bugs, please open an [issue](https://github.com/mintlay
...
+- `node-daemon`: The node software as a command line software. This is the software that manages the blocks, p2p communication, and all the other important functions of the blockchain network. The `node-daemon` is at the center of all other executables, except for the GUI, which has an instance of the node included in it for simplicity.
+- `node-gui`: The graphical user interface of the node for people who are not as tech-savvy and just want to run a node using the simplest means possible. The GUI is almost always behind in development compared to other tools, but it contains what is necessary to run a node, stake, delegate, and do other important tasks. The GUI also contains the machinery to run a wallet and create one. You can open multiple wallets with the GUI.
+- `node-rpc-daemon`: A wallet that can be controlled remotely using RPC. This can be run as a service.
```
iljakuklic, line 31: This should be `wallet-rpc-daemon`

iljakuklic (approved):

README.md
TheQuantumPhysicist: Thank you. Fixed.

api-server/README.md
```diff
@@ -0,0 +1,112 @@
+## Mintlayer API server
+
+### Introduction
+
+The Mintlayer API server is a tool that scans the blockchain and publishes its data in a database for easy access. Technically speaking, this is done to achieve the trade-off where the blockchain itself contains the minimal required amount of data, while the API server indexes all the data for easy reach. The API server is used by block explorers and web wallets. The API server is made to be usable by anyway, such as exchanges, people interested in writing tooling for the blockchain, or anything else.
```
OBorce, line 5: This seems a bit weird:
The API server is made to be usable by **anyway**
probably should be "by anyone" or "in many ways"?

OBorce (approved):

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+The blockchain scanner daemon is a tool that runs in the backend, scans the blockchain, and updates the database information.
+
+#### How the API web server and the blockchain scanner daemon work together
+
+The architecture of the API server is made to be distributed as much as desired. You can run the database on as many servers as you wish in master-slave mode. This is achieved by separating the "API web server" from the "blockchain scanner daemon". You can have a single "blockchain scanner daemon", communicating with the `node-daemon` of Mintlayer, collecting information about new blocks and writing it to the master database, while having as many instances of the API web server reading from the slave databases. This ensures virtually an infinitely scalable infrastructure.
+
+### Using other database infrastructures
+
+Currently, the API server uses PostgreSQL for storage, but the design is extremely flexible and any desired database can be added if need by implementing some interface (trait) in the rust code.
+
```
ImplOfAnImpl, line 26: > can be added if ~need~ needed

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+In addition to the PostgreSQL, an implementation of a full in-memory storage exists, which we use for testing and as a reference implementation. Hence, when adding a new database implementation, the in-memory implementation can be used as a reference implementation. Our tests ensure that both PostgreSQL and in-memory implementation, through the beautiful abstractions of rust, both arrive to the same result. Any additional implementation can be added to the same test suite.
+
```
ImplOfAnImpl, line 28: > the in-memory implementation can be used as a reference ~implementation~ one

To avoid tautology

> both PostgreSQL and in-memory implementation, through the beautiful abstractions of rust, ~both~ arrive to the same result

That extra `both` sounds off

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+## How to run
+
+In the following we present the minimal requirements to run the API server in action. In this example, we will be using the testnet. Replace every `testnet` with `mainnet` for the mainnet.
+
+### Make sure you have a node running
+
+If you don't have a node running, the API server won't have a source, from which it can read block data. To run the node, please consult the [the main readme file of the repo](/README.md).
+
```
ImplOfAnImpl, line 36: > the [the main readme file

Double "the"

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+Notice a few things:
+
+1. This command doesn't need root
+2. The database is running in trust mode, which means no username or password required
+3. The command line argument `--rm` is used, so that the container will be deleted on exit
+4. The database name is `mintlayer-testnet`, which is a standard used in the form of `mintlayer-`, followed by the network name (mainnet, testnet, etc)
+5. The port binding `127.0.0.1:5432:5432` doesn't allow external computers in the network to connect. Binding to `0.0.0.0` has security implications that are out of the scope of this documentation. Please make sure to have proper network security when running a database.
+6. There's no volume set. Meaning: Once the container is stopped and removed, all the data in the database will be lost. Please consult the documentation of the container to learn how to preserve the database data.
+
```
ImplOfAnImpl, line 53: > There's no volume set. Meaning: Once the container is stopped and removed, all the data in the database will be lost

Well, postgres Dockerfile [has the `VOLUME` instruction](https://github.com/docker-library/postgres/blob/44ef8b226a40f86cf9df3f9299067db6779a3aa3/16/bookworm/Dockerfile#L189), which will create an anonymous volume if it's not mounted anywhere else.
Not sure if it's worth mentioning here though.

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+podman run --detach --rm --name MintlayerAPIServerDB -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_DB=mintlayer-testnet -p 127.0.0.1:5432:5432 docker.io/library/postgres
+```
+
+Notice a few things:
+
+1. This command doesn't need root
+2. The database is running in trust mode, which means no username or password required
+3. The command line argument `--rm` is used, so that the container will be deleted on exit
+4. The database name is `mintlayer-testnet`, which is a standard used in the form of `mintlayer-`, followed by the network name (mainnet, testnet, etc)
```
ImplOfAnImpl, line 50: > which is a standard used in the form of `mintlayer-`

This part sounds off. Perhaps it's better to say something like:
"This is the default database name that api server will use for testnet. In general, the default database name is `mintlayer-`, followed by the network name (mainnet, testnet, etc)"

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+or if you want to run using the executable directly:
+
+```
+api-blockchain-scanner-daemon --network testnet
+```
+
+And this should immediately work, where the blockchain scanner daemon will communicate with the default network RPC network port of the node (13030 for testnet, 3030 for mainnet), and it will also communicate with the database and write the data it finds in the blockchain.
+
```
ImplOfAnImpl, line 71: > And this should immediately work, where the blockchain scanner daemon will communicate ...

Sounds a bit off. I suggest to put full stop after `work` and start a new sentence:
"And this should immediately work. The blockchain scanner daemon will communicate..."

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+If you need to configure extra options, such as postgres username and password, just add `--help` to the commands above, and the commands will be shown.
+
```
ImplOfAnImpl, line 73: > and the ~commands~ options will be shown.

api-server/README.md
TheQuantumPhysicist: You're right. Fixed. Thanks.

azarovh (approved):

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+### How to run the API web server
+
+After having filled the database with information, the API web server can use this information to respond to http requests, whether for requests from the public, or your internal infrastructure for other purposes.
+
+Assuming the database server is setup on the same machine and is reachable with `127.0.0.1:5432`, you can use the following command to run the API web server, compiled from the source code:
+
```
ImplOfAnImpl, line 79: > with `127.0.0.1:5432`

Probably "via" is better than "with" here.

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+cargo run --bin api-web-server --release -- --network testnet --bind-address 127.0.0.1:3000
+```
+
+Or, you can just run the executable if you have:
```
ImplOfAnImpl, line 84: > Or, you can just run the executable if you have:

The sentence sounds unfinished.

api-server/README.md
```diff
@@ -0,0 +1,112 @@
...
+curl http://127.0.0.1:3000/api/v1/chain/10
+```
+
+to get the id of the block at height 10.
+
+Make sure the scanner is fully synced to get correct information about the current state of the blockchain.
+
+### Logging
+
+The same logging rules [in the main readme file](/README.md) apply here as well. By default, all the programs we use have INFO level logging.
```
ImplOfAnImpl, line 112: > By default, all the programs we use have INFO level logging.

Probably it'd be better to say something like "By default, all our programs use INFO level logging."

ImplOfAnImpl (approved):

api-server/README.md
TheQuantumPhysicist: Will mention it.

api-server/README.md
TheQuantumPhysicist: A little longer, but OK. Done.

api-server/README.md
TheQuantumPhysicist: Thanks for all your valuable comments. I believe I fixed all the issues. Pushing soon.

TheQuantumPhysicist:
This is getting merged next.

PR #1609, opened 2024-02-29
Use PoS for simulation test

azarovh:
Also changed the order in which reward and transactions are disconnected in block. It causes an error if delegation is created and staked to the same pool that is used for staking (see the test in `delegation_tests.rs`).


TheQuantumPhysicist (approved):

iljakuklic (approved):
Nice!

PR #1619, opened 2024-03-01
Feature/wallet sweep addresses

OBorce:
Add sweep functions for addresses and delegations.

### DIFF
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -139,6 +139,7 @@ BASE_SCRIPTS = [
     'wallet_data_deposit.py',
     'wallet_submit_tx.py',
     'wallet_select_utxos.py',
+    'wallet_sweep.py',
     'wallet_recover_accounts.py',
     'wallet_get_address_usage.py',
     'wallet_tokens.py',
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -139,7 +139,7 @@ BASE_SCRIPTS = [
     'wallet_data_deposit.py',
     'wallet_submit_tx.py',
     'wallet_select_utxos.py',
-    'wallet_sweep.py',
+    'wallet_sweep_address.py',
     'wallet_sweep_delegation.py',
     'wallet_recover_accounts.py',
     'wallet_get_address_usage.py',
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -140,6 +140,7 @@ BASE_SCRIPTS = [
     'wallet_submit_tx.py',
     'wallet_select_utxos.py',
     'wallet_sweep.py',
+    'wallet_sweep_delegation.py',
     'wallet_recover_accounts.py',
     'wallet_get_address_usage.py',
     'wallet_tokens.py',
--- a/test/functional/wallet_sweep.py
+++ b/test/functional/wallet_sweep.py
@@ -1,0 +1,226 @@
+#!/usr/bin/env python3
+#  Copyright (c) 2023 RBB S.r.l
+#  Copyright (c) 2017-2021 The Bitcoin Core developers
+#  opensource@mintlayer.org
+#  SPDX-License-Identifier: MIT
+#  Licensed under the MIT License;
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Wallet sweep test
+
+Check that:
+* We can create a new wallet,
+* get an address
+* send multiple times coins to the wallet's addresses
+* sync the wallet with the node
+* get utxos
+* sweep all coins from the address
+"""
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.mintlayer import (ATOMS_PER_COIN, make_tx, reward_input, tx_input)
+from test_framework.util import assert_in, assert_equal, assert_not_in
+from test_framework.mintlayer import mintlayer_hash, block_input_data_obj
+from test_framework.wallet_cli_controller import UtxoOutpoint, WalletCliController
+
+import asyncio
+import sys
+import random
+
+
+class WalletSweepAddresses(BitcoinTestFramework):
+
+    def set_test_params(self):
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+        self.extra_args = [[
+            "--blockprod-min-peers-to-produce-blocks=0",
+        ]]
+
+    def setup_network(self):
+        self.setup_nodes()
+        self.sync_all(self.nodes[0:1])
+
+    def generate_block(self):
+        node = self.nodes[0]
+
+        block_input_data = { "PoW": { "reward_destination": "AnyoneCanSpend" } }
+        block_input_data = block_input_data_obj.encode(block_input_data).to_hex()[2:]
+
+        # create a new block, taking transactions from mempool
+        block = node.blockprod_generate_block(block_input_data, [], [], "FillSpaceFromMempool")
+        node.chainstate_submit_block(block)
+        block_id = node.chainstate_best_block_id()
+
+        # Wait for mempool to sync
+        self.wait_until(lambda: node.mempool_local_best_block_id() == block_id, timeout = 5)
+
+        return block_id
+
+    def run_test(self):
+        if 'win32' in sys.platform:
+            asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
+        asyncio.run(self.async_test())
+
+    async def async_test(self):
+        node = self.nodes[0]
+        async with WalletCliController(node, self.config, self.log) as wallet:
+            # new wallet
+            await wallet.create_wallet()
+
+            # check it is on genesis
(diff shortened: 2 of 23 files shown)


wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -154,6 +154,24 @@ trait WalletRpc {
...
+        from_addresses: Vec<String>,
+        options: TransactionOptions,
+    ) -> rpc::RpcResult<NewTransaction>;
+
+    #[method(name = "staking_sweep_delegation")]
+    async fn sweep_delegation(
+        &self,
+        account: AccountArg,
+        destination_address: String,
+        delegation_id: String,
```
azarovh, line 171: I'm not sure why we need separate function for that. Can we extend `withdraw_from_delegation` with something like enum {Max, Amount(Amount)}; ? 

Imho separate function would make sense if we withdraw from all pool's delegations or all account's delegations. 

Just a suggestion

test/functional/wallet_sweep.py
```diff
@@ -0,0 +1,226 @@
...
+            self.log.info(f"new token id: {frozen_token_id}")
+            block_id = self.generate_block()
+            assert_in("Success", await wallet.sync())
+            assert_in("The transaction was submitted successfully", await wallet.mint_tokens(frozen_token_id, frozen_tokens_address, 10000))
+            block_id = self.generate_block()
+            assert_in("Success", await wallet.sync())
+            assert_in("The transaction was submitted successfully", await wallet.freeze_token(frozen_token_id, 'unfreezable'))
+            addresses.append(tokens_address)
+
+            # get a coin address to inclode in the sweep to pay for the sweep tx fees
```
TheQuantumPhysicist, line 198: include

test/functional/wallet_sweep.py
```diff
@@ -0,0 +1,226 @@
...
+
+            # now sweep all tokens to acc0
+            assert_in("The transaction was submitted successfully", await wallet.sweep_addresses(acc0_address, addresses))
+
+            block_id = self.generate_block()
+            assert_in("Success", await wallet.sync())
+
+            # check the balance
+            balance = await wallet.get_balance()
+            # frozen tokens can't be transfered so they should be here
```
TheQuantumPhysicist, line 211: transferred

test/functional/wallet_sweep.py
```diff
@@ -0,0 +1,226 @@
...
+            assert_in(f"{frozen_token_id} amount: 10000", balance)
+            # the other token should not be in the acc1 balance any more
+            assert_not_in(f"{token_id}", balance)
+
+            await wallet.select_account(0)
+            balance = await wallet.get_balance()
+            # frozen tokens can't be transfered so they are not in acc0 balance
```
TheQuantumPhysicist, line 218: transferred

test/functional/wallet_sweep.py
```diff
@@ -0,0 +1,226 @@
...
+            assert_not_in(f"{frozen_token_id}", balance)
+            # the other token should be fully transfered in the acc0 balance
```
TheQuantumPhysicist, line 220: transferred

wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -415,6 +415,27 @@ pub enum WalletCommand {
         utxos: Vec<String>,
     },
 
+    /// Sweep all spendable coins or tokens from an address or addresses to a given address.
+    /// Sepndable coins are any coins that are not locked, and tokens that are not frozen or locked.
```
TheQuantumPhysicist, line 419: Spendable

test/functional/test_runner.py
```diff
@@ -139,6 +139,8 @@ class UnicodeOnWindowsError(ValueError):
     'wallet_data_deposit.py',
     'wallet_submit_tx.py',
     'wallet_select_utxos.py',
+    'wallet_sweep.py',
```
TheQuantumPhysicist, line 142: Please rename the file to `wallet_sweep_address.py` since this is only about sweeping addresses, to avoid ambiguity.

wallet/src/account/currency_grouper/mod.rs
```diff
@@ -128,6 +130,36 @@ pub fn group_outputs_with_issuance_fee<T, Grouped: Clone>(
     Ok(tokens_grouped)
 }
 
+pub fn output_currency_value(output: &TxOutput) -> Result<(Currency, Amount), UtxoSelectorError> {
```
TheQuantumPhysicist, line 133: These kinds of functions are just pure play with fire... and I have consistently complained about them. Please avoid doing generic logical groupings like this for enums. 

A while back we had a `value()` function for TxOutput, and it was removed for the same reason I call this playing with fire: People tend to casually just call the "value" function when they want the value, while what they need in reality is a separate `match` for every different case. There's nothing such as "value" for a transaction output. Every case means a different thing. There may be a case where you need spendable values. There may be another case where you need "spendable values at any time in the future". There may be a case where you need "values locked in a pool". These umbrella functions just lead to issues and oversight by removing the protection the compiler provides. I'm more than happy to "bloat" the code with as many match statements as necessary, and in return avoid mistakes.

The only case where such a function is a good idea, is when it's private for a specific case that's included in the name of the function.

wallet/src/account/mod.rs
```diff
@@ -562,6 +562,124 @@ impl Account {
...
+            .sum::<Option<Amount>>()
+            .ok_or(WalletError::OutputAmountOverflow)?;
+
+        let coin_input = grouped_inputs.remove(&Currency::Coin).ok_or(WalletError::NoUtxos)?;
+
+        let mut outputs = grouped_inputs
+            .into_iter()
+            .filter_map(|(currency, (amount, _))| {
+                let value = match currency {
+                    Currency::Coin => None?,
```
TheQuantumPhysicist, line 591: Why not `return None`? This looks kinda weird... the try operator usually assumes a chance for success.

wallet/wallet-rpc-lib/src/rpc/interface.rs
TheQuantumPhysicist: It might work... we'll need to change the functions' parametrization so that it can take different input data with different meanings, because "max" is something that can be entered and it's not a number. Don't forget that fee calculation has to be done on the fly. It's not such a bad idea, but I think having "sweep" functions is just easier to understand, since it's available in other wallets for other blockchains.

TheQuantumPhysicist (commented):
Besides the "value" functions and the other minor issues, this is looking OK. I'll take another look at the PR once everything is fixed.

wallet/src/account/currency_grouper/mod.rs
```diff
@@ -128,6 +130,36 @@ pub fn group_outputs_with_issuance_fee<T, Grouped: Clone>(
...
+    let output_value = match output {
+        TxOutput::Transfer(v, _) | TxOutput::LockThenTransfer(v, _, _) => v.clone(),
+        TxOutput::IssueNft(token_id, _, _) => {
+            OutputValue::TokenV1(*token_id, Amount::from_atoms(1))
+        }
+        TxOutput::CreateStakePool(_, data) => OutputValue::Coin(data.pledge()),
+        TxOutput::ProduceBlockFromStake(_, _)
```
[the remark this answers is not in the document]
OBorce: removed and changed the tests that were using this option.

wallet/src/account/currency_grouper/mod.rs
```diff
@@ -128,6 +130,36 @@ pub fn group_outputs_with_issuance_fee<T, Grouped: Clone>(
     Ok(tokens_grouped)
 }
 
+fn output_spendable_value(output: &TxOutput) -> Result<(Currency, Amount), UtxoSelectorError> {
+    let output_value = match output {
+        TxOutput::Transfer(v, _) | TxOutput::LockThenTransfer(v, _, _) => v.clone(),
+        TxOutput::IssueNft(token_id, _, _) => {
+            OutputValue::TokenV1(*token_id, Amount::from_atoms(1))
```
TheQuantumPhysicist, line 137: Is this a mistake?

OBorce: no, you can spend IssueNft as input, but it could be clearer if it just returns it directly as (Currencty::Token(token_id), amount(1)) instead of making an OutputValue and then extracting it from there again.

TheQuantumPhysicist (approved):

wallet/src/wallet/tests.rs
```diff
@@ -1540,6 +1523,24 @@ fn create_stake_pool_and_list_pool_ids(#[case] seed: Seed) {
...
+        )
+        .unwrap();
+    assert_eq!(create_stake_pool_utxos.len(), 1);
+    let (_, output, _) = create_stake_pool_utxos.pop().unwrap();
+    match output {
+        TxOutput::CreateStakePool(id, data) => {
+            assert_eq!(id, *pool_id);
+            assert_eq!(data.pledge(), pool_amount);
+        }
+        _ => panic!("wront TxOutput type"),
```
iljakuklic, line 1541: typo

wallet/src/account/output_cache/mod.rs
```diff
@@ -698,28 +698,29 @@ impl OutputCache {
...
+                        TxOutput::IssueNft(_, _, _)
+                        | TxOutput::DataDeposit(_)
+                        | TxOutput::CreateStakePool(_, _)
+                        | TxOutput::DelegateStaking(_, _)
+                        | TxOutput::CreateDelegationId(_, _)
+                        | TxOutput::IssueFungibleToken(_)
+                        | TxOutput::ProduceBlockFromStake(_, _) => false,
+                    }
+                } else {
+                    false
```
iljakuklic, line 723: Not that it would make much difference but you can use `Option::is_some_and` here

iljakuklic (approved):

PR #1620, opened 2024-03-01
rpc: Make RPC interfaces self-describing

iljakuklic:
* Add a set of data structures statically describing RPC interfaces
* The description structures can be rendered into markdown.
* Introduce the `rpc::describe` macro that generates the structures from the RPC interface trait (extracting doc comments for description)
* Apply that to all the mintlayer node interfaces
* Create an example executable in `node-lib` that collects the node RPC interfaces and prints out the description to the output
* Use it to generate RPC docs in [./node-daemon/RPC_INTERFACE.md](https://github.com/mintlayer/mintlayer-core/blob/rpc-description/node-daemon/RPC_INTERFACE.md)
* A bunch of other things along the way, some shortcuts taken


node-daemon/RPC_INTERFACE.md
```diff
@@ -0,0 +1,1170 @@
...
+{}
+```
+
+Returns:
+```
+string
+```
+
+
+### Method `node_set_mock_time`
```
TheQuantumPhysicist, line 42: Perhaps we should have a directive to hide some methods from documentation. But this isn't a priority. You can do this next/later, once we have something that fully works.

node-daemon/RPC_INTERFACE.md
```diff
@@ -0,0 +1,1170 @@
...
+Parameters:
+```
+{
+    "block_id": hex string,
+}
+```
+
+Returns:
+```
+number OR null
```
TheQuantumPhysicist, line 244: I think this is a good example of docs that can be improved. We should explain why this can be null. Tomorrow I'll drill through them all.

node-daemon/RPC_INTERFACE.md
TheQuantumPhysicist, line 1: From the first look this looks good. Do we have the ability to produce the same for the wallet?

node-daemon/RPC_INTERFACE.md
```diff
@@ -0,0 +1,1170 @@
...
+}
+```
+
+Returns:
+```
+string
+```
+
+
+### Method `test_functions_generate_transactions`
```
TheQuantumPhysicist, line 1138: Another example of things that have to be hidden

iljakuklic: The test functions are in a separate RPC module. These can be hidden easily by excluding the module from the output. The tool that generates the docs already has the `--dev` switch that makes it include the test functions module. I can easily re-run it without the switch and this module will be gone.

rpc/description-macro/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+                    if arg.path.is_ident("item") {
+                        item_ty = Some(arg.value()?.parse::<syn::Type>()?);
+                    }
+                    Ok(())
+                })?;
+                let item_ty = item_ty.ok_or(syn::Error::new(meta.span(), "Item missing"))?;
+                let unsub_name = match unsub_name.as_str() {
+                    "" => match name.strip_prefix("subscribe") {
+                        Some(name) => format!("unsubscribe{name}"),
+                        None => return Err(syn::Error::new(meta.span(), "No unsubscibe")),
```
TheQuantumPhysicist, line 117: unsubscribe

rpc/description-macro/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+                };
+                Ok::<_, syn::Error>((name, unsub_name, Box::new(item_ty)))
+            })
+            .transpose()?;
+
+        let (name, kind_meta) = match (method_info, sub_info) {
+            (None, None) => {
+                return Err(syn::Error::new(
+                    item.span(),
+                    "Neither method nor subscribtion",
```
TheQuantumPhysicist, line 129: subscription

TheQuantumPhysicist (approved):
This looks fine, besides the minor issues mentioned, and besides all that, as mentioned before, we need to add the wallet as well.

iljakuklic:
Added docs for wallet RPC. Unfortunately, it involves a fair amount of manual implementation of the `ValueHint` trait. I took the liberty to stub out some `ValueHint` implementations so some less important paramsetres and return values are described as being a "json object" rather than describing the internal structure in detail.

The `ValueHint` trait implementation is quite suitable to be generated by a macro but I did not do it now due to the time constraints. Even with the macro, it will not be as straightforward as slappping `derive(ValueHint)` on the data structures. Quite a few arguments are currently simply a `String` whereas the provided manual `ValueHint` implementations differentiate between a string representing a hex value or just some sort of name or string that is expected to be in the `bech32` format. It would be a good idea to make this distinction at the type level too so we get a reasonable `ValueHint` for free, in addition to being a good practice in general.

iljakuklic:
Changed the docs generating tool from "example"s to a binary in a separate crate.

PR #1626, opened 2024-03-05
Make set_size_limit in the mempool accept both string and integer

TheQuantumPhysicist:


mempool/src/config.rs
```diff
@@ -46,6 +47,61 @@ impl std::fmt::Display for MempoolMaxSize {
...
+
+                Ok(max_size)
+            }
+        }
+    }
+}
+
+impl HasValueHint for MempoolMaxSize {
+    const HINT: rpc_description::ValueHint =
+        rpc_description::ValueHint::Prim("Size with units, such as MB/KB/GB, or integer for bytes");
```
iljakuklic, line 102: This is probably too long to be a value hint. It is included in the params in the generated docs and is expected to be rather short as it can be a part of a more complicated type description. Have you checked how it looks in the generated doc? It may be better just have some sort of succinct name here and add details to a doc comment somewhere.

mempool/src/config.rs
```diff
@@ -46,6 +47,61 @@ impl std::fmt::Display for MempoolMaxSize {
...
+            StringOrUint::UInt(bytes) => {
+                let bytes: usize = bytes.try_into().map_err(|_| {
+                    D::Error::custom(format!("Integer {} is too large to fit", bytes))
+                })?;
+
+                let max_size = MempoolMaxSize::from_bytes(bytes);
+
+                Ok(max_size)
+            }
+        }
```
iljakuklic, line 96: The `byte_unit::Byte` apparently already provides serde impls that support numbers and strings. The `serde` feature has to be enabled for it. Could that be reused?

https://docs.rs/byte-unit/latest/src/byte_unit/byte/serde_traits.rs.html#29-80

mempool/src/config.rs
TheQuantumPhysicist: I could use that to simplify the implementation. I'll give it a shot. But this doesn't provide the features we need. We need the serialization to always lead to a number, while this doesn't. Also we want this to be without a key, so simple deriving doesn't work.

TheQuantumPhysicist: I seem to have commented on the wrong comment. This is about serde.

TheQuantumPhysicist: Just checked. This looks fine.

iljakuklic (approved):

PR #1627, opened 2024-03-05
P2p: inform peers about disconnection

ImplOfAnImpl:
Functions/events that perform disconnection now accept an optional `DisconnectionReason`; if it's set, a `WillDisconnect` message will be sent to the peer before disconnecting if the peer has the appropriate protocol version.
The current protocol version has been upped to 3.

Also, I wrapped the "magic bytes" array that represents the network into a new-type `MagicBytes`; this is not really needed in this PR, but I wanted to implement `Display` for it to make log lines look nicer.

### DIFF
--- a/p2p/src/disconnection_reason.rs
+++ b/p2p/src/disconnection_reason.rs
@@ -15,7 +15,7 @@
 
 use std::fmt::Display;
 
-use common::primitives::time::Time;
+use common::{chain::config::MagicBytes, primitives::time::Time};
 
 use p2p_types::services::Services;
 
@@ -36,14 +36,13 @@ pub enum DisconnectionReason {
     SyncRequestsIgnored,
     TooManyInboundPeersAndThisOneIsDiscouraged,
     TooManyInboundPeersAndCannotEvictAnyone,
-
     UnsupportedProtocol,
     TimeDiff {
         remote_time: Time,
         accepted_peer_time: std::ops::RangeInclusive<Time>,
     },
     DifferentNetwork {
-        our_network: [u8; 4],
+        our_network: MagicBytes,
     },
     NoCommonServices,
     InsufficientServices {
@@ -90,7 +89,7 @@ impl Display for DisconnectionReason {
                 remote_time, accepted_peer_time
             ),
             DisconnectionReason::DifferentNetwork { our_network } => {
-                write!(f, "Wrong network; out network is {our_network:?}")
+                write!(f, "Wrong network; out network is '{our_network}'")
             }
             DisconnectionReason::NoCommonServices => write!(f, "No common services"),
             DisconnectionReason::InsufficientServices { needed_services } => {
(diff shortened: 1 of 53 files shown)


p2p/src/disconnection_reason.rs
```diff
@@ -0,0 +1,164 @@
...
+        accepted_peer_time: std::ops::RangeInclusive<Time>,
+    },
+    DifferentNetwork {
+        our_network: MagicBytes,
+    },
+    NoCommonServices,
+    InsufficientServices {
+        needed_services: Services,
+    },
+}
```
ImplOfAnImpl, line 51: FYI: here, ConnectionFromSelf and DifferentNetwork are kind of useless, they exist mostly for completeness:
1) It makes no sense to send WillDisconnect for ConnectionFromSelf if the connection is indeed a self-connection. But technically, detection of self-connections may have false positives (if 2 randomly chosen 64-bit nonces happen to be the same), so I decided to do it.
2) WillDisconnect for DifferentNetwork will never be received by a peer in reality, because both sides perform the check "simultaneously" after the handshake and will disconnect each other before any other message will have a chance to be handled. 

UnsupportedProtocol is also useless at this moment; it'll only become useful after/if we drop the support for V3.

p2p/src/tests/bad_time_diff.rs
```diff
@@ -0,0 +1,263 @@
...
+        TestTransportChannel, TestTransportMaker, TestTransportNoise, TestTransportTcp,
+    },
+    tests::helpers::TestNode,
+};
+
+// Check that a handshake is rejected if the time difference between the peers is too big.
+// Also check that the WillDisconnect message is sent in this case.
+
+// V3+ is needed for the WillDisconnect message to be sent.
+const TEST_PROTOCOL_VERSION: SupportedProtocolVersion = SupportedProtocolVersion::V3;
```
TheQuantumPhysicist, line 41: Are there tests for how V3 will act when meeting V2?

p2p/src/disconnection_reason.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl Display for DisconnectionReason {
```
TheQuantumPhysicist, line 53: Is in it better to make DisconnectionReason an error using thiserror?

p2p/src/tests/bad_time_diff.rs
ImplOfAnImpl: There are "generic" tests that check what happens when different protocol versions meet each other (the lowest version is selected). I don't think we need additional tests for this.

There was one thing I missed though - to check that `WillDisconnect` won't be sent for V2. So I've pushed an additional commit where I check this (by running the `bad_time_diff` and `same_handshake_nonce` tests for V2 as well.)

azarovh (approved):

TheQuantumPhysicist (commented):
--

p2p/src/disconnection_reason.rs
TheQuantumPhysicist: I'm fine either way... I was just surprised that this is too verbose for what it's.

PR #1635, opened 2024-03-06
wallet-rpc: Re-generate docs after the cold wallet split-off

iljakuklic:


iljakuklic:
There is currently some duplication.

It would be better to break down the wallet RPC interface into multiple modules:
* Daemon control (i.e. shutdown)
* Common wallet functionality
* Cold wallet functions
* Hot wallet functions
* Node functions

An object can implement multiple of these and it would provide a way of documenting these without much duplication.

TheQuantumPhysicist (approved):

PR #1640, opened 2024-03-06
change wasm lib to use decimal amounts instead of atoms

OBorce:
- wasm lib now uses decimal amounts instead of atoms

### DIFF
--- a/wasm-wrappers/WASM-API.md
+++ b/wasm-wrappers/WASM-API.md
@@ -182,3 +182,10 @@ A utxo can either come from a transaction or a block reward. This enum signifies
 
 The part of the transaction that will be committed in the signature. Similar to bitcoin's sighash.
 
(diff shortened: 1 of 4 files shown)


TheQuantumPhysicist (approved):
Besides the minor comments, I think this is OK.

I recommend waiting until I merge #1643, after which you don't have to bother about html. There's another generator which is friendler.

wasm-wrappers/WASM-API.md
```diff
@@ -164,7 +164,7 @@ The effective balance is how the influence of a pool is calculated due to its ba
 
 ### Enum: `Network`
 
-The network, for which an operation to be done. Mainnet, testnet, etc.
+
```
TheQuantumPhysicist, line 167: Why removed?

PR #1645, opened 2024-03-08
Introduce Rpc Amount types and use them in wallet rpc

iljakuklic:
* `RpcAmountIn` used for RPC params and allows the user to pick whether to specify amount in decimal units or in atoms
* `RpcAmountOut` used for RPC return values gives the user the amount in both decimal units and in atoms
* Use the above types in the wallet RPC
* `HasValueHint` derive macro now supports `#[serde(flatten)]` on enums

TODO:
* [x] Some serialization / deserialization unit tests for the `RpcAmount*` types

### DIFF
--- a/common/src/primitives/amount/rpc.rs
+++ b/common/src/primitives/amount/rpc.rs
@@ -114,8 +114,56 @@ impl RpcAmountOut {
     pub fn decimal(&self) -> DecimalAmount {
         self.decimal
     }
+
+    /// Check this is the same number presented in the same way
+    pub fn is_same(&self, other: &Self) -> bool {
+        self.atoms == other.atoms && self.decimal.is_same(&other.decimal)
+    }
 }
 
 impl HasValueHint for RpcAmountOut {
     const HINT: VH = VH::Object(&[("atoms", &VH::NUMBER_STRING), ("decimal", &VH::DECIMAL_STRING)]);
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+    use serde_json::json;
+
+    #[rstest]
+    #[case::zero(0, 0)]
+    #[case(123456, 0)]
+    #[case(123450, 1)]
+    #[case(123450, 2)]
+    #[case(123450, 9)]
+    #[case(4000000000, 3)]
+    #[case(1u128 << 55, 17)]
+    #[case(u128::MAX, 0)]
+    #[case(u128::MAX, 1)]
+    #[case(u128::MAX, 25)]
+    fn rpc_amount_serde(#[case] atoms: u128, #[case] n_decimals: u8) {
+        let atoms = Amount::from_atoms(atoms);
+        let decimal = DecimalAmount::from_amount_minimal(atoms, n_decimals);
+
+        let decimal_str = decimal.to_string();
+        let atoms_str = atoms.into_atoms().to_string();
+
+        assert!(!atoms_str.contains('.'));
+
+        let decimal_in = RpcAmountIn::from_decimal(decimal);
+        let decimal_in_json = json!({ "decimal": decimal_str });
+        assert_eq!(serde_json::to_value(decimal_in).unwrap(), decimal_in_json);
+        assert!(decimal_in.is_same(&serde_json::from_value(decimal_in_json).unwrap()));
+
+        let atoms_in = RpcAmountIn::from_atoms(atoms);
+        let atoms_in_json = json!({ "atoms": atoms_str });
+        assert_eq!(serde_json::to_value(atoms_in).unwrap(), atoms_in_json);
+        assert!(atoms_in.is_same(&serde_json::from_value(atoms_in_json).unwrap()));
+
+        let rpc_out = RpcAmountOut::from_amount_minimal(atoms, n_decimals);
+        let rpc_out_json = json!({ "atoms": atoms_str, "decimal": decimal_str });
+        assert_eq!(serde_json::to_value(&rpc_out).unwrap(), rpc_out_json);
+        assert!(rpc_out.is_same(&serde_json::from_value(rpc_out_json).unwrap()));
+    }
 }
--- a/common/src/primitives/amount/rpc.rs
+++ b/common/src/primitives/amount/rpc.rs
@@ -106,18 +106,18 @@ impl RpcAmountOut {
     }
 
     pub fn from_amount(amount: Amount, decimals: u8) -> Self {
-        Self::from_amount_minimal(amount, decimals)
+        Self::from_amount_no_padding(amount, decimals)
     }
 
     /// Construct from amount, keeping all decimal digits
-    pub fn from_amount_full(amount: Amount, decimals: u8) -> Self {
-        let decimal = DecimalAmount::from_amount_full(amount, decimals);
+    pub fn from_amount_full_padding(amount: Amount, decimals: u8) -> Self {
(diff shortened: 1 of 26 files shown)


wallet/wallet-rpc-lib/src/rpc/types.rs
```diff
@@ -359,26 +360,19 @@ impl NftMetadata {
...
-            Self::Lockable => Ok(tokens::TokenTotalSupply::Lockable),
-            Self::Unlimited => Ok(tokens::TokenTotalSupply::Unlimited),
-            Self::Fixed(amount) => {
-                let decimals = chain_config.coin_decimals();
-                let amount = amount.to_amount(decimals).ok_or(RpcError::InvalidCoinAmount)?;
-                Ok(tokens::TokenTotalSupply::Fixed(amount))
-            }
+            Self::Lockable => Some(tokens::TokenTotalSupply::Lockable),
+            Self::Unlimited => Some(tokens::TokenTotalSupply::Unlimited),
+            Self::Fixed(amount) => amount.to_amount(decimals).map(tokens::TokenTotalSupply::Fixed),
```
iljakuklic, line 375: This had `chain_config` passed to it and used coin decimals t parse the string. To me it looks like a bug: the number of decimal places specified by the token should be used here, so I changed it. Seems no tests picked up on it.

@OBorce could you please check this is correct?

wallet/wallet-rpc-lib/src/rpc/types.rs
```diff
@@ -393,11 +387,10 @@ pub struct TokenMetadata {
...
 impl TokenMetadata {
-    pub fn token_supply<N: NodeInterface>(
-        &self,
-        chain_config: &ChainConfig,
-    ) -> Result<tokens::TokenTotalSupply, RpcError<N>> {
-        self.token_supply.into_token_supply(chain_config)
+    pub fn token_supply<N: NodeInterface>(&self) -> Result<tokens::TokenTotalSupply, RpcError<N>> {
+        self.token_supply
+            .to_token_supply(self.number_of_decimals)
+            .ok_or(RpcError::InvalidCoinAmount)
```
iljakuklic, line 393: This is where it's called (follow up to the previous comment). @OBorce

common/src/primitives/amount/rpc.rs
```diff
@@ -0,0 +1,169 @@
...
+/// Amount type suitable for getting user input supporting both decimal and atom formats in Json.
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize, HasValueHint)]
+pub struct RpcAmountIn(RpcAmountData);
+
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize, HasValueHint)]
+#[serde(untagged)]
+enum RpcAmountData {
+    Decimal(DecimalAmount),
+    Atoms(Amount),
+}
```
TheQuantumPhysicist, line 29: If both are provided, it's an error, right?

wallet/wallet-rpc-lib/src/rpc/types.rs
OBorce: yes this is correct, the token functional tests are not yet running on the rpc wallet I will try to migrate them

common/src/primitives/amount/rpc.rs
iljakuklic: Let me add some tests

common/src/primitives/amount/rpc.rs
```diff
@@ -0,0 +1,169 @@
...
+        Self::from_amount_minimal(amount, decimals)
+    }
+
+    /// Construct from amount, keeping all decimal digits
+    pub fn from_amount_full(amount: Amount, decimals: u8) -> Self {
+        let decimal = DecimalAmount::from_amount_full(amount, decimals);
+        Self::new_internal(amount, decimal)
+    }
+
+    /// Construct from amount, keeping the minimal number of decimal places
```
TheQuantumPhysicist, line 104: I'm really having difficulty understanding what "minimal" really means. Can you please provide examples?

common/src/primitives/amount/rpc.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn new_internal(atoms: Amount, decimal: DecimalAmount) -> Self {
+        Self { atoms, decimal }
+    }
+
+    pub fn from_amount(amount: Amount, decimals: u8) -> Self {
+        Self::from_amount_minimal(amount, decimals)
+    }
+
+    /// Construct from amount, keeping all decimal digits
+    pub fn from_amount_full(amount: Amount, decimals: u8) -> Self {
```
TheQuantumPhysicist, line 99: IIUC, this should be called "all" instead of "full", though I'm not sure how to name the function.

common/src/primitives/amount/rpc.rs
```diff
@@ -0,0 +1,169 @@
...
+        let decimal_in = RpcAmountIn::from_decimal(decimal);
+        let decimal_in_json = json!({ "decimal": decimal_str });
+        assert_eq!(serde_json::to_value(decimal_in).unwrap(), decimal_in_json);
+        assert!(decimal_in.is_same(&serde_json::from_value(decimal_in_json).unwrap()));
+
+        let atoms_in = RpcAmountIn::from_atoms(atoms);
+        let atoms_in_json = json!({ "atoms": atoms_str });
+        assert_eq!(serde_json::to_value(atoms_in).unwrap(), atoms_in_json);
+        assert!(atoms_in.is_same(&serde_json::from_value(atoms_in_json).unwrap()));
+
```
TheQuantumPhysicist, line 163: I just want to confirm to ensure I understand how this works, even though it's written here: When the user uses RpcAmountIn as a parameter to input amount information, THEY MUST include a json object with being either `{"atoms": "12345"}` or `{"decimal": "123.45"}`. AND, there's no assumptions about `decimal`, as it's stored as a string, and the handler has to decide the number of decimals based on the application (coin vs token). Am I getting this correctly?

TheQuantumPhysicist (approved):
Besides the comments, looking good. Let's discuss the comments and ensure this is in its best state.

OBorce (approved):

common/src/primitives/amount/rpc.rs
iljakuklic: > I just want to confirm to ensure I understand how this works, even though it's written here: When the user uses RpcAmountIn as a parameter to input amount information, THEY MUST include a json object with being either `{"atoms": "12345"}` or `{"decimal": "123.45"}`. AND, there's no assumptions about `decimal`, as it's stored as a string, and the handler has to decide the number of decimals based on the application (coin vs token). Am I getting this correctly?

Yes, correct. When decimal representation is used, it behaves as `DecimalAmount`. It is stored as a sort of floating point representation holding the integral amount and the number of decimal digits used in the original string representation. In order to convert it to `Amount`, the number of decimals of the target coin/token has to be provided and appropriate conversion is done.

common/src/primitives/amount/rpc.rs
iljakuklic: Let's say you have `Amount` of `12300` atoms and given token uses 3 decimal places. Minimal will create string representation of `"12.3"` whereas full will create a string representation of `"12.300"` when serialised.

TheQuantumPhysicist: OK that means I misunderstood how this works. In that case we should rename these functions `with_full_padding` and `no_padding`. These would hopefully be self-documenting.

PR #1646, opened 2024-03-08
Move documentation from CLI commands to RPC interface

OBorce:
- use Description to add the documentation from the RPC interface back to the clap commands

### DIFF
--- a/wallet/wallet-cli-lib/src/repl/mod.rs
+++ b/wallet/wallet-cli-lib/src/repl/mod.rs
@@ -17,8 +17,9 @@ pub mod interactive;
 pub mod non_interactive;
 
 use clap::{Command, FromArgMatches, Subcommand};
+use rpc::description::{Described, Module};
 use tokio::sync::mpsc;
-use wallet_rpc_lib::types::NodeInterface;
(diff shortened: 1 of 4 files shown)


wallet/wallet-cli-lib/src/repl/mod.rs
```diff
@@ -57,7 +61,20 @@ pub fn get_repl_command(cold_wallet: bool) -> Command {
...
+                    .split('_')
+                    .zip(subcommand.get_name().split('-'))
+                    .all(|(x, y)| x == y)
+                    .then_some(method.description)
+            })
+        {
+            *subcommand = subcommand.clone().help_template(COMMAND_HELP_TEMPLATE).about(desc);
+        } else {
+            *subcommand = subcommand.clone().help_template(COMMAND_HELP_TEMPLATE);
+        }
```
TheQuantumPhysicist, line 77: This is great! I think the only concern is that the parameters have the same name. This can be done later using `syn` by parsing both interfaces. But I guess this is good enough for now. Not sure whether we should create an issue for this.

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #1648, opened 2024-03-10
fix size estimation function in wasm to work for accounts

OBorce:
- pass destinations instead of utxos for the size estimation function as accounts don't have utxos

### DIFF
--- a/wasm-wrappers/src/lib.rs
+++ b/wasm-wrappers/src/lib.rs
@@ -587,11 +587,15 @@ pub fn encode_input_for_withdraw_from_delegation(
     Ok(input.encode())
 }
 
-/// Given inputs, each input's destination (from the UTXO or Account) and outputs, estimate the transaction size.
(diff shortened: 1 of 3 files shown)


wasm-wrappers/src/lib.rs
```diff
@@ -587,13 +587,15 @@ pub fn encode_input_for_withdraw_from_delegation(
     Ok(input.encode())
 }
 
-/// Given inputs, outputs and utxos (each encoded as `Option<TxOutput>`), estimate the transaction size.
+/// Given inputs, each input's destination (from the UTXO or Account) and outputs, estimate the transaction size.
```
TheQuantumPhysicist, line 590: This does not explain how the information should be passed. Please explain it with an example.

TheQuantumPhysicist (approved):

iljakuklic (approved):

PR #1649, opened 2024-03-11
P2p: allow to disable/re-enable networking

ImplOfAnImpl:


TheQuantumPhysicist (approved):
There's a test not working, as discussed. Besides that, looks OK.

PR #1652, opened 2024-03-12
rpc: Introduce RpcString types and apply them to token metadata

iljakuklic:
* These types represent strings that are supposed to be human-readable most of the time. However, because arbitrary data are allowed, sometimes we have to use the hex representation
* `RpcStringIn` is used for parameters and tolerates either a string or a hex string as an input.
* `RpcStringOut` is used for return values and writes both textual string (if possible, `null` otherwise) and hex string
* Use that for token metadata

### DIFF
--- a/rpc/types/src/string.rs
+++ b/rpc/types/src/string.rs
@@ -1,0 +1,217 @@
+// Copyright (c) 2024 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use rpc_description::{HasValueHint, ValueHint as VH};
+
+/// Binary data encoded as a hex string in Serde/RPC.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[serde(try_from = "HexStringSerde", into = "HexStringSerde")]
+pub struct RpcHexString(Vec<u8>);
+
+impl RpcHexString {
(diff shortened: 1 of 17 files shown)


rpc/types/src/string.rs
```diff
@@ -0,0 +1,217 @@
...
+            RpcStringInSerde::Text(s) | RpcStringInSerde::Bare(s) => s.into(),
+            RpcStringInSerde::Hex(hex) => hex.into(),
+        }
+    }
+}
+
+/// A binary string type suitable for use in RPC return values.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
+#[serde(into = "RpcStringOutSerde", try_from = "RpcStringOutSerde")]
+pub struct RpcStringOut(Vec<u8>);
```
TheQuantumPhysicist, line 159: While the naming is not an important issue now, it's bad for something that goes straight to the blockchain to put "RPC" in it... I would call it something like `WrappedUtf8String`.

TheQuantumPhysicist (approved):

rpc/types/src/string.rs
iljakuklic: > While the naming is not an important issue now, it's bad for something that goes straight to the blockchain to put "RPC" in it... I would call it something like `WrappedUtf8String`.

I am not sure what you mean by straight to the blockchain. This is specifically for RPC. It does not have `Encode`/`Decode`. If you want to put it into blockchain, you have to convert it to `Vec<u8>` or something similar.

TheQuantumPhysicist: Oh, my mistake. Never mind. I confused it with something else.

rpc/types/src/string.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+#[derive(serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum RpcStringInSerde {
+    Text(String),
+    Hex(RpcHexString),
+    #[serde(untagged)]
+    Bare(String),
+}
```
TheQuantumPhysicist, line 139: I think this is unnecessary...

TheQuantumPhysicist: We don't have backwards compatibility requirement now. No one is using tokens.

TheQuantumPhysicist: Perhaps it doesn't matter. Never mind.

OBorce (approved):

PR #1653, opened 2024-03-12
rpc: Use `expect-test` to ensure RPC docs are up to date

iljakuklic:
* Use `expect-test` to check whether RPC docs have been updated after interface changes.
* Part of the normal test suite so fails testing (and CI) if docs are not fully up-to-date.
* Run tests with `UPDATE_EXPECT=1` to refresh the docs. E.g.:
  * `UPDATE_EXPECT=1 cargo test -p node-daemon` for node docs
  * `UPDATE_EXPECT=1 cargo test -p wallet-rpc-daemon` for wallet docs
* Added `node-daemon/docs/RPC_DEV.md` which documents testing and development related RPC functionality.
* Less docs handling code, fewer ad hoc scripts.


TheQuantumPhysicist (approved):
I really don't understand why we should play all the gymnastics here instead of some `--check` flag in the generator... 

I can live with this though.

PR #1654, opened 2024-03-13
add force option to wallet open to convert between hot and cold

OBorce:
- add command line options to CLI and RPC wallet 
- add optional flag to the wallet-open command to force change between hot and cold or cold and hot wallet types

### DIFF
--- a/wallet/wallet-cli-lib/src/commands/mod.rs
+++ b/wallet/wallet-cli-lib/src/commands/mod.rs
@@ -65,6 +65,9 @@ pub enum ColdWalletCommand {
         wallet_path: PathBuf,
         /// The existing password, if the wallet is encrypted.
         encryption_password: Option<String>,
+        /// Force change the wallet type from hot to cold or from cold to hot
+        #[arg(long)]
+        force_change_wallet_type: bool,
     },
 
     #[clap(name = "wallet-close")]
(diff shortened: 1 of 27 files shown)


wallet/wallet-cli-lib/src/commands/mod.rs
```diff
@@ -65,6 +65,9 @@ pub enum ColdWalletCommand {
         wallet_path: PathBuf,
         /// The existing password, if the wallet is encrypted.
         encryption_password: Option<String>,
+        /// Force change the wallet type from hot to cold or from cold to hot
+        #[arg(long)]
+        force_change_wallet_type: bool,
     },
```
TheQuantumPhysicist, line 71: This parameter should be optional, and it should be false by default.

TheQuantumPhysicist (approved):
Besides the CLI optional thing, this looks OK. Please fix that and feel free to merge.

wallet/wallet-cli-lib/src/commands/mod.rs
OBorce: clap treats it as a flag so it is false by default and only set to true when you pass --force-change-wallet-type

But can't make it optional in the RPC interface, as even as optional it needs None/null to be passed

PR #1661, opened 2024-03-14
Add checkpoints

ImplOfAnImpl:
The `get_block_ids_as_checkpoints` function has been added to `ChainstateInterface` . The wallet got the corresponding `node-get-block-ids-as-checkpoints` command (hidden).

Checkpoint enforcement has been improved a bit - the function `preliminary_header_check` has been replaced with `preliminary_headers_check` in `ChainstateInterface`, which accepts additional headers that follow the "main" one and checks them for checkpoints violations. This way p2p may detect the violations earlier.

Some tests have been added that check that p2p rejects peers that send headers that violate checkpoints.

Checkpoints for mainnet and testnet have been hardcoded in `chain::config`.


chainstate/src/detail/query.rs
```diff
@@ -161,6 +162,41 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+        step: usize,
+    ) -> Result<Vec<(BlockHeight, Id<GenBlock>)>, PropertyQueryError> {
+        ensure!(
+            end_height >= start_height,
+            PropertyQueryError::InvalidBlockHeightRange {
+                start: start_height,
+                end: end_height
+            }
+        );
+        ensure!(step > 0, PropertyQueryError::InvalidStep(step));
```
TheQuantumPhysicist, line 178: Why not use NonZeroUSize for step?

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -49,7 +49,17 @@ pub trait ChainstateInterface: Send + Sync {
...
+
+    /// Check the headers. The first header's parent block must be known.
+    /// Each of the "other" headers must be connected to the previous one and the first of them
+    /// should be connected to `first_header`. These headers only get the most basic checks (e.g.
+    /// checkpoint enforcement).
+    fn preliminary_headers_check(
+        &self,
+        first_header: &SignedBlockHeader,
+        other_headers: &[SignedBlockHeader],
+    ) -> Result<(), ChainstateError>;
```
TheQuantumPhysicist, line 61: If first_header is connected to other_headers... why aren't they all in one list?

common/src/chain/config/builder.rs
```diff
@@ -320,6 +324,7 @@ impl Builder {
         Self {
             chain_type,
             bip44_coin_type: chain_type.default_bip44_coin_type(),
+            checkpoints: None,
```
TheQuantumPhysicist, line 327: I think a better way to represent this is to enforce having genesis in it by default instead of None. So instead of `Option<BTreeMap>`, it becomes `BTreeMap` with the invariant that the first block in it is genesis.

common/src/chain/config/builder.rs
```diff
@@ -526,6 +545,11 @@ impl Builder {
     builder_method!(epoch_length: NonZeroU64);
     builder_method!(sealed_epoch_distance_from_tip: usize);
 
+    pub fn checkpoints(mut self, checkpoints: BTreeMap<BlockHeight, Id<GenBlock>>) -> Self {
+        self.checkpoints = Some(checkpoints);
+        self
+    }
```
TheQuantumPhysicist, line 551: Then if genesis is included by default, you can use the default builder_method macro.

common/src/chain/config/builder.rs
```diff
@@ -426,10 +435,20 @@ impl Builder {
...
+                checkpoints
+            } else {
+                match chain_type {
+                    ChainType::Mainnet => make_mainnet_checkpoints(),
+                    ChainType::Testnet => make_testnet_checkpoints(),
+                    ChainType::Regtest | ChainType::Signet => BTreeMap::new(),
+                }
+            };
+
+            checkpoints.entry(0.into()).or_insert(genesis_block.get_id().into());
```
TheQuantumPhysicist, line 449: I guess we have genesis anyway. So that Option isn't necessary.

Cargo.toml
```diff
@@ -165,6 +165,7 @@ http = "0.2"
 hyper = "0.14"
 iced = "0.10"
 iced_aw = "0.7"
+indoc = "2"
```
TheQuantumPhysicist, line 168: I think using numbers like "2" instead of "2.0" is a bad idea, because then "2.1" could be incompatible and then it won't build.

I know SemVer dictates that 2.0 and 2.1 should have compatibility, but in the real world this isn't how this works. I constantly have to fix API problems when updating the minor version... see my PR for jsonrpsee as an example. Iced also is driving us crazy with this.

Perhaps this rule should go into the python script for checking code.

TheQuantumPhysicist (approved):
Besides the comments, looking good.

chainstate/src/interface/chainstate_interface.rs
ImplOfAnImpl: > If first_header is connected to other_headers... why aren't they all in one list?

Actually, initially I had them all in one list, but then decided to split them. The reasons are that, first of all, `first_header` is treated differently than the rest of the headers. Secondly, `first_header` is required, and the others are optional.
So, if you don't insist, I'd leave it as is.

common/src/chain/config/builder.rs
ImplOfAnImpl: > I think a better way to represent this is to enforce having genesis in it by default instead of None. So instead of Option<BTreeMap>, it becomes BTreeMap with the invariant that the first block in it is genesis.

We can't enforce the presence of the genesis here, because we don't yet know what genesis will actually be used (e.g. some tests use a custom genesis).

chainstate/src/interface/chainstate_interface.rs
TheQuantumPhysicist: I actually kind of insist... it feels to me that this is leaking implementation details.

Besides, it's acceptable to return no errors for an empty list. It kinds of maps to "checking no blocks is not an error", unless it's required for another reason.

common/src/chain/config/builder.rs
TheQuantumPhysicist: Alrighty.

PR #1663, opened 2024-03-14
rpc: Introduce `RpcAddress` and use it in wallet RPC

iljakuklic:
* New `RpcAddress<T>` type with conversions to/from `Address<T>`
* Use in wallet RPC for a more type safe interface
* Makes the docs more precise regarding the string format
* Removes a bunch of remaining manual `ValueHints`
* Also replace some uses of plain `String` with `RpcHexString`, `HexEncoded<T>`, `Id<T>` as appropriate
* The changes are backwards compatible

`WalletRpcClient` would benefit from similar changes. Unfortunately I do not have the willpower to do it so there is a bunch of conversions to/from strings there.

### DIFF
--- a/common/src/address/mod.rs
+++ b/common/src/address/mod.rs
@@ -16,6 +16,7 @@
 pub mod dehexify;
 pub mod hexified;
 pub mod pubkeyhash;
+pub mod rpc;
 pub mod traits;
 
 use crate::chain::ChainConfig;
@@ -24,6 +25,7 @@ use std::fmt::Display;
 use utils::qrcode::{qrcode_from_str, QrCode, QrCodeError};
 
 use self::traits::Addressable;
+pub use rpc::RpcAddress;
 
 #[derive(thiserror::Error, Debug, Eq, PartialEq, Clone, PartialOrd, Ord)]
 pub enum AddressError {
@@ -126,6 +128,12 @@ impl<T: Addressable> Address<T> {
             )
         };
         Ok(result)
+    }
+}
+
+impl<T> Address<T> {
+    pub fn into_string(self) -> String {
+        self.address
     }
 }
 
--- a/common/src/address/mod.rs
+++ b/common/src/address/mod.rs
@@ -22,7 +22,10 @@ pub mod traits;
 use crate::chain::ChainConfig;
 use crate::primitives::{encoding, Bech32Error};
 use std::fmt::Display;
-use utils::qrcode::{qrcode_from_str, QrCode, QrCodeError};
+use utils::{
+    ensure,
+    qrcode::{qrcode_from_str, QrCode, QrCodeError},
+};
 
 use self::traits::Addressable;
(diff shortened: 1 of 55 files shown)


OBorce (approved):

common/src/address/mod.rs
```diff
@@ -129,6 +131,12 @@ impl<T: Addressable> Address<T> {
     }
 }
 
+impl<T> Address<T> {
+    pub fn into_string(self) -> String {
+        self.address
+    }
```
TheQuantumPhysicist, line 137: This method is a problem. The problem is that we do not assume that the address in `Address<T>` is a valid address. Hence, you could get a completely bogus address out of this. Actually, the fact that the address can be invalid is part of why RpcAddress works, because the checking is done afterwards, with decode_object.

Now this isn't a problem I know how to solve. Perhaps we should remove the Display impl and only keep your into_string, and make it return a `Result<String>`... but I really don't know the best way to do this.

TheQuantumPhysicist (approved):
Besides the comment, all good.

iljakuklic:
> The problem is that we do not assume that the address in Address<T> is a valid address. Hence, you could get a completely bogus address out of this. Actually, the fact that the address can be invalid is part of why RpcAddress works, because the checking is done afterwards, with decode_object.

So I dug a bit deeper into this and noticed that the method that handles decoding without hrp validation does not actually need to construct an `Address`. So I added the following changes:
* Moved the decoder without hrp validation from `Address` to `HexifiedAddress`.
* Changed the interface of `Address` to properly enforce its abstraction boundary, making sure it always holds data of a valid object.
* As a consequence, the likes of  `into_string` are now sound and `decode_object` no longer errors out.
* Updated call sites to these methods, which no longer need error handling. The error "handling" was mostly done by means of `expect`s and `unwrap`s anyway, so these are now gone.

Let me know if I missed something.

iljakuklic:
> > The problem is that we do not assume that the address in Address is a valid address. Hence, you could get a completely bogus address out of this. Actually, the fact that the address can be invalid is part of why RpcAddress works, because the checking is done afterwards, with decode_object.
> 
> So I dug a bit deeper into this and noticed that the method that handles decoding without hrp validation does not actually need to construct an `Address`. So I added the following changes:
> 
> * Moved the decoder without hrp validation from `Address` to `HexifiedAddress`.
> * Changed the interface of `Address` to properly enforce its abstraction boundary, making sure it always holds data of a valid object.
> * As a consequence, the likes of  `into_string` are now sound and `decode_object` no longer errors out.
> * Updated call sites to these methods, which no longer need error handling. The error "handling" was mostly done by means of `expect`s and `unwrap`s anyway, so these are now gone.
> 
> Let me know if I missed something.

Updated the `Address` type to also hold the decoded object of type `T` to eliminate potential decoding issues.

OBorce (approved):

TheQuantumPhysicist (approved):

PR #1669, opened 2024-03-19
Remove Address::from_str as an optimization and as cleanup

TheQuantumPhysicist:


common/src/address/mod.rs
```diff
@@ -56,7 +56,11 @@ impl<T: Addressable> Address<T> {
         Ok(Self { address, object })
     }
 
-    pub fn from_string(cfg: &ChainConfig, address: String) -> Result<Self, AddressError> {
+    pub fn from_string(
+        cfg: &ChainConfig,
+        address: impl Into<String>,
+    ) -> Result<Self, AddressError> {
+        let address = address.into();
```
iljakuklic, line 63: I am not sure this is much of an optimization. As far as I can see, `String` implements `From<&str>` but not `From<String>`. I suspect that what happens if you pass in a `String` is that it gets auto-deref-ed into a `&str` and than the `.into()` just allocates a new string and copies into it. The `address` string that was originally passed in is then thrown away. I would not bank on it being optimised away by the compiler, allocating memory tends to be an optimisation barrier more often than not.

I tried to find and existing trait (there is, for example, `ToOwned`) that covers both owned and borrowed use cases but did not find one that did what I wanted, i.e. move when passed in a `String` and allocate a new one if given a `&str`.

TheQuantumPhysicist: Does this prove that this isn't what's happening?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=df47590fee72fa33fe7180554ffab1e6

iljakuklic: If there's a well-behaved `From<String> for String`, then it should be fine.

TheQuantumPhysicist: There's perhaps a misunderstanding.

> That fails to compile when I hit the run button.

as it should. Basically what you're saying in your first comment, IIUC, please correct me, is that when `x.into()` is called, what's really called is `(&x).into()` with the `Into<String>` trait (bound by the type of `y`, explicitly mentioned). But then if that's the case, then `x` shouldn't be moved to `y`. The example doesn't compile because `x` is really moved, which means that it's not using `&str`, but it really is moving `String`.

iljakuklic: Ok, found it. All good. Cheers. https://doc.rust-lang.org/std/convert/trait.From.html#impl-From%3CT%3E-for-T

iljakuklic (approved):

PR #1671, opened 2024-03-19
Enforce running as non-root

TheQuantumPhysicist:


node-lib/src/options.rs
```diff
@@ -221,6 +221,12 @@ pub struct RunOptions {
     /// Minimum transaction relay fee rate (in atoms per 1000 bytes).
     #[clap(long, value_name = "VAL")]
     pub min_tx_relay_fee_rate: Option<u64>,
+
+    /// Allow the program to be run as root, which is NOT RECOMMENDED and DANGEROUS
+    /// There is no need to run the mintlayer-software as root. The correct
+    /// security practice is to only provide root access to programs that need it.
+    #[clap(long, action = clap::ArgAction::SetTrue)]
+    pub force_run_as_root: Option<bool>,
```
ImplOfAnImpl, line 229: IMO `allow_run_as_root` would be better (or maybe `force_allow_run_as_root`).
The current name sounds like the node itself will call `sudo` or something.

Also, the full stop after "DANGEROUS" is missing.

ImplOfAnImpl (approved):

node-lib/src/options.rs
TheQuantumPhysicist: Sounds good. Done.

node-lib/src/options.rs
```diff
@@ -221,6 +221,12 @@ pub struct RunOptions {
     /// Minimum transaction relay fee rate (in atoms per 1000 bytes).
     #[clap(long, value_name = "VAL")]
     pub min_tx_relay_fee_rate: Option<u64>,
+
+    /// Allow the program to be run as root, which is NOT RECOMMENDED and DANGEROUS.
+    /// There is no need to run the mintlayer-software as root. The correct
+    /// security practice is to only provide root access to programs that need it.
+    #[clap(long, action = clap::ArgAction::SetTrue)]
+    pub force_allow_run_as_root: Option<bool>,
```
iljakuklic, line 229: Minor: I think this can  be just `bool`, without the extra `Option`. Will be false by default and true if the flag is passed on the command line.

iljakuklic (approved):

node-lib/src/options.rs
TheQuantumPhysicist: Done.

PR #1672, opened 2024-03-19
Add creation_block to delegations API endpoints

OBorce:
Add creation block to delegation endpoints

### DIFF
--- a/api-server/api-server-common/src/storage/impls/postgres/queries.rs
+++ b/api-server/api-server-common/src/storage/impls/postgres/queries.rs
@@ -589,6 +589,7 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
             "CREATE TABLE ml.delegations (
                     delegation_id TEXT NOT NULL,
                     block_height bigint NOT NULL,
+                    creation_block_height bigint NOT NULL,
                     pool_id TEXT NOT NULL,
                     balance TEXT NOT NULL,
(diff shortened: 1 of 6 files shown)


api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -836,7 +837,7 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
         let row = self
             .tx
             .query_opt(
-                r#"SELECT pool_id, balance, spend_destination, next_nonce
+                r#"SELECT pool_id, balance, spend_destination, next_nonce, creation_time
```
TheQuantumPhysicist, line 840: Perhaps creation block height makes more sense... because the time is derived from that?

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -589,6 +589,7 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
             "CREATE TABLE ml.delegations (
                     delegation_id TEXT NOT NULL,
                     block_height bigint NOT NULL,
+                    creation_block bigint NOT NULL,
```
TheQuantumPhysicist, line 592: As discussed, please call it creation_block_height, and same in all relevant places.

PR #1674, opened 2024-03-20
Clean up epoch delta on disconnect

azarovh:
When last block from the epoch is disconnected its delta should be removed from the db

### DIFF
--- a/chainstate/test-suite/src/tests/pos_accounting_reorg.rs
+++ b/chainstate/test-suite/src/tests/pos_accounting_reorg.rs
@@ -13,7 +13,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use itertools::Itertools;
 use std::num::NonZeroU64;
 
 use super::helpers::{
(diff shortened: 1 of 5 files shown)


chainstate/test-suite/src/tests/pos_accounting_reorg.rs
```diff
@@ -51,156 +50,155 @@ use test_utils::random::{make_seedable_rng, Seed};
...
-fn stake_pool_reorg(#[case] seed: Seed) {
+#[case(Seed::from_entropy(), NonZeroU64::new(1).unwrap(), 0)]
+#[case(Seed::from_entropy(), NonZeroU64::new(1).unwrap(), 1)]
+#[case(Seed::from_entropy(), NonZeroU64::new(1).unwrap(), 2)]
+#[case(Seed::from_entropy(), NonZeroU64::new(2).unwrap(), 0)]
+#[case(Seed::from_entropy(), NonZeroU64::new(2).unwrap(), 1)]
+#[case(Seed::from_entropy(), NonZeroU64::new(2).unwrap(), 2)]
+#[case(Seed::from_entropy(), NonZeroU64::new(3).unwrap(), 0)]
+#[case(Seed::from_entropy(), NonZeroU64::new(3).unwrap(), 1)]
+#[case(Seed::from_entropy(), NonZeroU64::new(3).unwrap(), 2)]
```
TheQuantumPhysicist, line 61: Instead of this, can't we just generate a random number in that range?

TheQuantumPhysicist (approved):
Besides the minor comment, this is good to go.

PR #1677, opened 2024-03-21
Unify all related stuff of checking running as root to be able to put it by just adding a clap arg

TheQuantumPhysicist:
So, everyone can use this now. Just add it to your run options and call that function.

### DIFF
--- a/utils/src/root_user.rs
+++ b/utils/src/root_user.rs
@@ -1,0 +1,59 @@
+// Copyright (c) 2024 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! The node command line options.
+
+use clap::Args;
+use logging::log;
+
+fn is_unix() -> bool {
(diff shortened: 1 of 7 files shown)


utils/src/root_user.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl ForceRunAsRootOptions {
+    pub fn ensure_not_running_as_root_user(&self) -> anyhow::Result<()> {
+        if is_unix() && !self.force_allow_run_as_root {
+            use std::os::unix::fs::MetadataExt;
+            let uid = std::fs::metadata("/proc/self").map(|m| m.uid());
+            match uid {
+                Ok(id) => {
+                    if id == 0 {
+                        return Err(anyhow::anyhow!("ERROR: It is a mistake to run the node as root (user with uid=0), as it gives the node power that it does not need and violates good security practices. Either run the program as non-root, or do the VERY NOT RECOMMENDED THING TO DO, and add the flag `{FORCE_ALLOW_ROOT_FLAG}`"));
```
iljakuklic, line 50: The error message here mentions "node" but it is now more general and can be applied to other executables too.

utils/src/root_user.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// security practice is to only provide root access to programs that need it.
+    #[clap(long, default_value_t = false)]
+    pub force_allow_run_as_root: bool,
+}
+
+impl ForceRunAsRootOptions {
+    pub fn ensure_not_running_as_root_user(&self) -> anyhow::Result<()> {
+        if is_unix() && !self.force_allow_run_as_root {
+            use std::os::unix::fs::MetadataExt;
+            let uid = std::fs::metadata("/proc/self").map(|m| m.uid());
```
iljakuklic, line 46: Not quite sure but I suspect this might fail to compile on Windows.

utils/src/root_user.rs
TheQuantumPhysicist: Fixed

iljakuklic (approved):

PR #1684, opened 2024-03-21
Chainstate errors cleanup

ImplOfAnImpl:
Minor error-related cleanup here and there:
1) In `blockprod/src/detail/mod.rs` some PropertyQueryError's were created after accessing chainstate, while the actual ChainstateError was ignored. I decided that it'd be better to include the original error in the resulting error as well; for this purpose I added an additional `ConsensusPoSError::ChainstateError` variant and `consensus::ChainstateError` type that wraps `chainstate::ChainstateError`. Unfortunately, putting  `chainstate::ChainstateError` there directly turned out to be challenging, because from the one hand `consensus` can only depend on `chainstate-types` and, on the other hand, moving all chainstate error types to `chainstate-types` turned out to be hard if not impossible, due to lots of dependencied on all kinds of stuff.
So I ended up adding a `DynError` trait in `utils` which is the standard `Error` for which `Eq` and `Clone` are also implemented, and used it in `consensus::ChainstateError`.
2) I got rid of the `get_previous_block_index_for_check_block` function in the chainstate, which was only ever re-used because it accepted a block header, unlike the "normal" `get_previous_block_index`, which accepted a `BlockIndex`. To unify this, I introduced a trait `IdableWithParent`, which is implemented by blocks, block headers and indices.
3) I couldn't resist and added the `PropertyQueryError(#[from] PropertyQueryError)` variant to `BlockError` again. I do remember that it was frowned upon in the past, but maybe we can discuss this again. (I'd rather remove the specific "PropertyQueryError" wrappers from `BlockError`, because IMO they are useless)
4) For the same reason, I added `PropertyQueryError(#[from] PropertyQueryError)` to `OrphanCheckError`, replacing the specific error `PrevBlockIndexNotFound`.
The latter was not only (IMO) redundant, but also used incorrectly - it was effectively wrapping a storage error instead of representing the "not found" condition, which is checked explicitly later. (I believe this might be a common problem for these specific PropertyQueryError wrappers, which is one of the reasons I'd prefer to get rid of them).


utils/src/dyn_error.rs
```diff
@@ -0,0 +1,111 @@
...
+/// implements `Clone`.
+pub trait DynError: std::error::Error + Send + Sync {
+    fn as_any(&self) -> &dyn Any;
+
+    fn dyn_eq(&self, other: &dyn Any) -> bool;
+
+    fn dyn_clone(&self) -> Box<dyn DynError>;
+}
+
+impl<T: 'static + std::error::Error + Eq + Clone + Send + Sync> DynError for T {
```
TheQuantumPhysicist, line 28: Why not `<'a, T: 'a, ...>` instead of `<T: 'static, ...>`?

consensus/src/pos/error.rs
```diff
@@ -102,3 +108,14 @@ pub enum ConsensusPoSError {
...
+// TODO: include the original chainstate::ChainstateError in each error below.
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
+pub enum ChainstateError {
+    #[error("Failed to obtain epoch for block height {0}")]
+    FailedToObtainEpochData(BlockHeight, Box<dyn DynError>),
+    #[error("Failed to read data of pool {0}")]
+    StakePoolDataReadError(PoolId, Box<dyn DynError>),
+    #[error("Failed to read balance of pool {0}")]
+    PoolBalanceReadError(PoolId, Box<dyn DynError>),
+}
```
TheQuantumPhysicist, line 121: Why do we need this, with dyn Errors in it?

chainstate/types/src/block_index.rs
```diff
@@ -39,6 +40,22 @@ pub struct BlockIndex {
     status: BlockStatus,
 }
 
+impl Idable for BlockIndex {
+    type Tag = Block;
+
+    fn get_id(&self) -> Id<Block> {
+        *self.block_id()
+    }
+}
```
TheQuantumPhysicist, line 49: BlockIndex is not idable. 

Why is this necessary?

utils/src/dyn_error.rs
iljakuklic: > Why not `<'a, T: 'a, ...>` instead of `<T: 'static, ...>`?

I don't think that would work without introducing a lifetime parametrisation to the trait or the type.

ImplOfAnImpl: Yeah, it was needed because `Any` needs `'static`. This was has been removed though.

consensus/src/pos/error.rs
ImplOfAnImpl: I've replaced `Box<dyn DynError>` with `String` as discussed.

chainstate/types/src/block_index.rs
ImplOfAnImpl: > BlockIndex is not idable

I've removed this.

TheQuantumPhysicist (approved):

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -1318,3 +1326,55 @@ pub enum ReorgError {
     #[error("Generic error during reorg: {0}")]
     OtherError(#[from] BlockError),
 }
+
+trait BlockInfo: Sized {
```
TheQuantumPhysicist, line 1330: Maybe a separate module for this... the file is already 1300 lines long...

PR #1685, opened 2024-03-21
Fixes related to block invalidation

ImplOfAnImpl:
1) Block error "classification" is introduced, which separates errors into 3 categories - "BadBlock", "TemporarilyBadBlock" and "General". This is used to decide whether the block should be marked as invalid in the block index when the corresponding error occurs.
2) ~`attempt_to_process_block` now can handle blocks that were origianlly rejected, but whose status was manually reset to "ok" (i.e. the function now allows to add a block whose index already exists in the db, if it's not "failed" or "fully valid").~ I've reverted this change. Instead, now "ok" block indices are not saved by `attempt_to_process_block` if the block itself is not going to be persisted. Also, resetting failure flags now removes block indices if the corresponding blocks are not in the db. 
3) An assertion failure in `BlockInvalidator::find_and_activate_best_chain` was fixed.

### DIFF
--- a/chainstate/src/detail/block_invalidation/best_chain_candidates.rs
+++ b/chainstate/src/detail/block_invalidation/best_chain_candidates.rs
@@ -127,26 +127,28 @@ impl BestChainCandidates {
     }
 
     // Remove the block and its descendants, which must be specified in descendants_indices,
-    // from the set and add the block's parent block to the set.
+    // from the set and add the block's parent block to the set if its chain trust is not less than
+    // the specified minimum.
     #[log_error]
-    pub fn on_block_invalidated<Chs: ChainstateAccessor>(
+    pub fn remove_tree_add_parent<Chs: ChainstateAccessor>(
         &mut self,
         chs: &Chs,
-        invalidated_block_info: &Chs::BlockInfo,
+        root_block_info: &Chs::BlockInfo,
         descendant_block_infos: &[Chs::BlockInfo],
+        min_chain_trust: Uint256,
     ) -> Result<(), BestChainCandidatesError> {
-        self.remove(invalidated_block_info);
+        self.remove(root_block_info);
 
         for descendant_info in descendant_block_infos {
             self.remove(descendant_info);
         }
 
         // Add the parent to the list
-        if let Some(parent_block_id) =
-            chs.gen_block_id_to_normal(&invalidated_block_info.parent_id())
-        {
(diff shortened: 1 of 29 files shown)


chainstate/src/detail/block_invalidation/best_chain_candidates.rs
```diff
@@ -127,26 +127,28 @@ impl BestChainCandidates {
...
+    // the specified minimum.
     #[log_error]
-    pub fn on_block_invalidated<Chs: ChainstateAccessor>(
+    pub fn remove_tree_add_parent<Chs: ChainstateAccessor>(
         &mut self,
         chs: &Chs,
-        invalidated_block_info: &Chs::BlockInfo,
+        root_block_info: &Chs::BlockInfo,
         descendant_block_infos: &[Chs::BlockInfo],
+        min_chain_trust: Uint256,
```
TheQuantumPhysicist, line 138: Why?

Changing these algorithms needs very careful security consideration. We cannot just change them nonchalant to achieve some technical goal. Let's discuss this.

Keep in mind that adding things like this make tests reproducibility very difficult.

ImplOfAnImpl: This is called as a part of `find_and_activate_best_chain` call during block invalidation. Though `invalidate_block` gets called in `attempt_to_process_block` too, in that case it's only called for a stale block for which a reorg attempt has already failed, so it won't trigger another reorg. I.e. this logic will kick in only on a manual block invalidation. I don't think it can affect security.
This particular change (passing `min_chain_trust` into `on_block_invalidated` and taking it into account) is a fix for the assertion failure that we had recently when trying to manually invalidate a block (the test for this is called `test_invalidation_with_reorg_attempt_to_chain_with_lower_chain_trust`)

chainstate/src/detail/error_classification.rs
```diff
@@ -0,0 +1,777 @@
...
+use utxo::UtxosBlockUndoError;
+
+use crate::{BlockError, CheckBlockError, CheckBlockTransactionsError, OrphanCheckError};
+
+use super::{
+    block_invalidation::BestChainCandidatesError, chainstateref::EpochSealError, BlockSizeError,
+};
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum BlockProcessingErrorClass {
```
TheQuantumPhysicist, line 44: Please add explanation to why this exists and what problem it solves

TheQuantumPhysicist (approved):

PR #1696, opened 2024-03-26
add output encoding function for NFT issuance

OBorce:
- add encode output for nft issuance
- add functions to get token related fees from chain config

### DIFF
--- a/wasm-wrappers/src/lib.rs
+++ b/wasm-wrappers/src/lib.rs
@@ -28,7 +28,10 @@ use common::{
         },
         stakelock::StakePoolData,
         timelock::OutputTimeLock,
-        tokens::{IsTokenFreezable, TokenIssuance, TokenIssuanceV1, TokenTotalSupply},
+        tokens::{
+            IsTokenFreezable, Metadata, NftIssuance, NftIssuanceV0, TokenCreator, TokenIssuance,
+            TokenIssuanceV1, TokenTotalSupply,
+        },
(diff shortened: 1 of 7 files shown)


wasm-wrappers/src/lib.rs
```diff
@@ -513,6 +516,50 @@ pub fn encode_output_create_stake_pool(
     Ok(output.encode())
 }
 
+/// Returns the fee that needs to be paid by a transaction for issuing a new fungible token
```
TheQuantumPhysicist, line 519: Please also document the required parameters. Remember that the documentation doesn't include that.

wasm-wrappers/src/lib.rs
```diff
@@ -547,6 +594,61 @@ pub fn encode_output_issue_fungible_token(
...
+    network: Network,
+) -> Result<Vec<u8>, Error> {
+    let chain_config = Builder::new(network.into()).build();
+    let token_id = parse_addressable(&chain_config, token_id)?;
+    let authority = parse_addressable(&chain_config, authority)?;
+    let name = name.into();
+    let ticker = ticker.into();
+    let media_uri = media_uri.into();
+    let icon_uri = icon_uri.into();
+    let media_hash = media_hash.into();
```
TheQuantumPhysicist, line 621: These and other fields need to be tested before approving the output. See in check_transaction how we do that. Feel free to factor out things that are needed to just test an output.

TheQuantumPhysicist (commented):
Besides the check_transaction issue and documentation, this is fine. Please do that and I'll review again.

PR #1698, opened 2024-03-27
P2p: apply some minor todos

ImplOfAnImpl:
Also, some questionable and/or outdated todos have been removed.


p2p/src/peer_manager/peer_context.rs
```diff
@@ -43,7 +43,7 @@ pub struct PeerContext {
     /// Bind address of this node's side of the connection.
     pub bind_address: SocketAddress,
 
-    pub peer_role: PeerRole,
+    pub conn_type: ConnectionType,
```
TheQuantumPhysicist, line 46: I do find Role more descriptive than "type"... type is too generic for this... I don't understand what's the problem with peer_role and PeerRole. These are the right words.

ImplOfAnImpl: > I do find Role more descriptive than "type"... type is too generic for this... I don't understand what's the problem with peer_role and PeerRole. These are the right words.

Well, IMO some items of the enum don't sound right with `PeerRole`. E.g. "feeler" is a kind of connection, not a peer. 
Anyway, I've reverted the renaming of `PeerRole` to `ConnectionType`. But note that I also renamed `enum Role {Inbound, Outbound}` to `ConnectionDirection` and this hasn't been reverted (though `ConnectionDirection` may not be the best name, IMO having both `PeerRole` and `Role` is too confusing).

TheQuantumPhysicist (approved):

PR #1699, opened 2024-03-28
rpc: Support internally tagged enums in the `HasValueHint` macro

iljakuklic:
Just implemented the macro support for internally tagged enum representation from serde. Not applied anywhere yet.


TheQuantumPhysicist (approved):

iljakuklic:
Turns out this is somewhat more difficult when dealing with enums where the data objects are factored out, like so:
```rust
struct NftData { ... }
struct TokenData { ... }
enum TokenOrNftData {
    Nft(NftData),
    Token(TokenData),
}
```

TheQuantumPhysicist (approved):

PR #1704, opened 2024-03-29
API server per thousand different representations

TheQuantumPhysicist:


common/src/primitives/per_thousand.rs
```diff
@@ -82,6 +82,14 @@ impl PerThousand {
         result
     }
 
+    pub fn into_decimal_str(&self) -> String {
+        Amount::from_atoms(self.0.into()).into_fixedpoint_str(3)
+    }
+
+    pub fn into_per_thousand_int(&self) -> u16 {
+        self.0
+    }
```
iljakuklic, line 91: Minor: I think `to_` is a more appropriate method name prefix here.

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

iljakuklic (approved):

common/src/primitives/per_thousand.rs
TheQuantumPhysicist: Based on the table, I went with `as_`.

The others should be `to_` as well...

OBorce (approved):

PR #1705, opened 2024-03-29
rpc: Make most token-related enums use internal tagging

iljakuklic:
Please let me know if I missed something or you disagree with some naming choices.

Also changed the representation of utxo outpoints to use internal tagging. Can be reverted if deemed too disruptive.

Also removed `Encode`/`Decode` from a bunch of RPC-specific types.


common/src/chain/transaction/utxo_outpoint.rs
```diff
@@ -17,19 +17,7 @@ use crate::chain::{transaction::Transaction, Block, GenBlock, Genesis};
...
-    Eq,
-    Encode,
-    Decode,
-    Ord,
-    PartialOrd,
-    serde::Serialize,
-    serde::Deserialize,
-    rpc_description::HasValueHint,
-)]
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, Ord, PartialOrd)]
```
TheQuantumPhysicist, line 20: Some times I think we should just make a static check that prevent Encode/Decode and serde::Serialize/Deserialize to coexist.

TheQuantumPhysicist (approved):

common/src/chain/tokens/rpc.rs
```diff
@@ -201,19 +161,9 @@ impl RPCNonFungibleTokenInfo {
...
-    fn from(creator: &TokenCreator) -> Self {
-        // None-RPC type mustn't have serde requirements
-        RPCTokenCreator(creator.encode())
-    }
-}
-
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize, HasValueHint)]
 pub struct RPCNonFungibleTokenMetadata {
-    pub creator: Option<RPCTokenCreator>,
+    pub creator: Option<RpcHexString>,
```
OBorce, line 166: Maybe this can be RpcAddress<Destination> as the creator is a public key?

wallet/wallet-rpc-lib/src/rpc/types.rs
```diff
@@ -428,11 +428,34 @@ impl StakingStatus {
         }
     }
 }
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, HasValueHint)]
+#[serde(tag = "source")]
```
OBorce, line 432: seems a bit weird that sometimes we use "source" sometimes "type" for the tag, should we agree on a single one?

OBorce (approved):

wallet/wallet-rpc-lib/src/rpc/types.rs
iljakuklic: > seems a bit weird that sometimes we use "source" sometimes "type" for the tag, should we agree on a single one?

The idea was to use a name that describes what the enum discriminates over, falling back on "type" normally. I am open to suggestions / critiques.

common/src/chain/tokens/rpc.rs
iljakuklic: > Maybe this can be RpcAddress as the creator is a public key?

I don't think `PublicKey` implements `Addressable`. The closest thing that does is `Destination`.

Maybe there could be `PublicKey` directly instead of `RpcHexString`, I think it serializes to a hex string too. Would also remove the need for some conversions.

wallet/wallet-rpc-lib/src/rpc/types.rs
azarovh, line 432: For other rpc types we also use `content` tag. Should we do it here (and in other places) as well for consistency?

azarovh (approved):

iljakuklic:
Pushed the following changes:
* Added support for adjacent enum representation: `{"type": "EnumArm", "content": arm_data}`
* Removed the support for the internl enum repr along with some semi-hacky machinery that was there to support it.
* Added a bunch of checks to the `HasValueHint` macro to narrow down the set of allowed enum representations:
  * Untagged representation is allowed. It is opt-in and used in some cases where it is quite useful (various `StringOrUInt` types).
  * External representation (serde default) is restricted to only the cases where enum arms do not contain any data. In that case, the enum is represented just by the arm name string.
  * Internal representation is now not supported by the macro and therefore not allowed.
  * Adjacent representation is required for enums where one or more arms contain data. In addition, it is enforced that the object key that specifies the enum arm is `"type"` and the object key that holds the data is `"content"`.
* Applied the above to the enums that constitute the external RPC interface.

common/src/chain/transaction/input.rs
```diff
@@ -31,6 +31,7 @@ use super::{AccountCommand, AccountOutPoint, AccountSpending, OutPointSourceId,
     serde::Serialize,
     serde::Deserialize,
 )]
+#[serde(tag = "type", content = "content")]
```
azarovh, line 34: do we use it somewhere? there is dedicated `RpcTxInput` type

common/src/chain/transaction/utxo_outpoint.rs
```diff
@@ -105,6 +93,38 @@ impl UtxoOutPoint {
...
 }
 
+impl rpc_description::HasValueHint for UtxoOutPoint {
+    const HINT_SER: rpc_description::ValueHint = RpcUtxoOutPoint::HINT_SER;
+    const HINT_DE: rpc_description::ValueHint = RpcUtxoOutPoint::HINT_DE;
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, rpc_description::HasValueHint)]
+#[serde(tag = "type", content = "content")]
+enum RpcUtxoOutPoint {
```
azarovh, line 103: `RpcUtxoOutpoint` exists in `chainstate/src/rpc/types/input.rs`

azarovh (commented):
some new types duplicate the `chainstate/src/rpc/types/`

common/src/chain/transaction/input.rs
iljakuklic: Yea, it is used. It is used in the wallet and also in the `get_block_json` RPC call. To get rid of this, we'd need to create the RPC equivalent of `Block` and all its constituent parts, effectively duplicating a fair chunk of `common::chain`. I'd consider that to be out of scope for this PR.

azarovh: there is `RpcBlock` already as well

common/src/chain/transaction/utxo_outpoint.rs
iljakuklic: These are slightly different. The chainstate one is being directly used by the wallet. The common one is a private implementation detail for serialize/deserailize of UtxoOutPoint (which exists for the same reason I explained in the other comment).

iljakuklic: Will try to at least unify the underlying implementation.

common/src/chain/transaction/input.rs
azarovh: So should we revert this? I thought we only care about content for public api and the wallet can just serialize/deserialize as is

iljakuklic: Not sure. Is there any harm in keeping it? I think, as discussed, whatever wallet needs internally should eventually be transmitted as hex strings so hopefully it will go one day.

iljakuklic: Do you have a reasonably simple way of finding out where it should be kept and where to remove it? One heuristic is the presence of `HasValueHint` but it's not completely reliable.

common/src/chain/transaction/utxo_outpoint.rs
iljakuklic: Hmm it requires me break RPC backwards compatibility in existing calls like `get_utxo`. The hope was this would only affect the token stuff.

iljakuklic: It's a more general problem. In order to make the enum representation consistent accross the board, we also need to change `OutPointSourceId`. That, however, means changing stuff that's likely already used like coin selection in wallet rpc or `get_utxo`. It does not apply only to the token-related stuff.

azarovh (approved):
looks good to me

OBorce (approved):

PR #1707, opened 2024-04-02
add simulation tests for the API Server

OBorce:
- add simulation tests for API Server
- compare UTXOs, Pools, Delegations and Tokens/NFTs between the scanner and the chainstate


api-server/scanner-lib/src/sync/tests.rs
```diff
@@ -1207,3 +1215,436 @@ async fn check_all_destinations_are_tracked(#[case] seed: Seed) {
...
+        let block_height = BlockHeight::new(height as u64);
+        let mut block_builder = tf.make_pos_block_builder().with_random_staking_pool(&mut rng);
+
+        for _ in 0..rng.gen_range(10..max_tx_per_block) {
+            block_builder = block_builder.add_test_transaction(&mut rng);
+        }
+
+        let block = block_builder.build();
+        for tx in block.transactions() {
+            let new_utxos = (0..tx.inputs().len()).map(|output_index| {
```
TheQuantumPhysicist, line 1304: Please make sure that all kinds of time locks are tested.

api-server/scanner-lib/src/sync/tests.rs
```diff
@@ -1207,3 +1216,531 @@ async fn check_all_destinations_are_tracked(#[case] seed: Seed) {
...
+        }
+
+        // delegations that have not been staked yet are stored in the scanner but not in the node
+        for (id, scanner_delegation) in scanner_delegations {
+            let share = node_delegations.get(&id).unwrap_or(&Amount::ZERO);
+            // check the shares are the same
+            assert_eq!(share, scanner_delegation.balance());
+        }
+    } else {
+        // the pool has been decommissioned
```
azarovh, line 1616: - should the delegations be checked as well? they exist in the node still
- staker_balance balance is also 0 here

api-server/scanner-lib/src/sync/tests.rs
```diff
@@ -1207,3 +1216,531 @@ async fn check_all_destinations_are_tracked(#[case] seed: Seed) {
...
+    pool_id: PoolId,
+) {
+    let tx = local_state.storage().transaction_ro().await.unwrap();
+    let scanner_data = tx.get_pool_data(pool_id).await.unwrap().unwrap();
+
+    if let Some(node_data) = tf.chainstate.get_stake_pool_data(pool_id).unwrap() {
+        // check staker total balances are the same
+        assert_eq!(node_data.staker_balance(), scanner_data.staker_balance());
+        // check the pledges are the same
+        assert_eq!(node_data.pledge_amount(), scanner_data.pledge_amount());
```
azarovh, line 1590: why check separate fields when whole struct is compared in the next line?

api-server/scanner-lib/src/sync/tests.rs
```diff
@@ -1207,3 +1216,531 @@ async fn check_all_destinations_are_tracked(#[case] seed: Seed) {
...
+) {
+    let tx = local_state.storage().transaction_ro().await.unwrap();
+    let scanner_data = tx.get_delegation(delegation_id).await.unwrap().unwrap();
+
+    if let Some(node_data) = tf.chainstate.get_stake_delegation_data(delegation_id).unwrap() {
+        assert_eq!(node_data.source_pool(), scanner_data.pool_id());
+        assert_eq!(
+            node_data.spend_destination(),
+            scanner_data.spend_destination()
+        );
```
azarovh, line 1644: it's safer to compare structs not fields

api-server/scanner-lib/src/sync/tests.rs
```diff
@@ -1207,3 +1216,531 @@ async fn check_all_destinations_are_tracked(#[case] seed: Seed) {
...
+    token_id: TokenId,
+) {
+    let tx = local_state.storage().transaction_ro().await.unwrap();
+    let node_data = tf.chainstate.get_token_info_for_rpc(token_id).unwrap().unwrap();
+
+    match node_data {
+        RPCTokenInfo::FungibleToken(node_data) => {
+            let scanner_data = tx.get_fungible_token_issuance(token_id).await.unwrap().unwrap();
+
+            assert_eq!(node_data.authority, scanner_data.authority);
```
azarovh, line 1681: let's add convertion function like `From<TokenInfo> for RPCTokenInfo` and compare structs. Otherwise it's easy to forget to update this test when new field is added.

same for NFT

chainstate/test-framework/src/utils.rs
```diff
@@ -304,3 +306,78 @@ pub fn pos_mine(
     }
     None
 }
+
+pub fn sign_witnesses(
+    key_manager: &KeyManager,
+    chainstate: &TestChainstate,
```
azarovh, line 312: It's confusing that data is retrieved from both chainstate and InMemory* stores. I think you can remove chainstate and get all data for PoS and tokens from InMemory*. (passing &ChainConfig directly)

chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -174,17 +177,15 @@ impl<'a> RandomTxMaker<'a> {
 
         // Spending from a token account requires paying fee. Find sufficient utxo per account input.
         let fee = self.chainstate.get_chain_config().token_supply_change_fee(BlockHeight::zero());
-        let fee_inputs = self
-            .utxo_set
+        let fee_inputs = inputs
```
azarovh, line 180: nice thanks

azarovh (approved):

chainstate/test-framework/src/utils.rs
OBorce: we need both because InMemory Store is updated while the Tx is being created, so if a token authority is changed it will change in the store, but if a previous Tx in the same block uses the same token it needs the previous authority to sign the input.
Which now I realize if 2 tx change the authority in the same block this will fail...

OBorce: pos store is used for delegations, for pools if the decommission happens from a CreateStakPool then I take the key from the utxo itself, and if it is a ProduceFromBlock I take it from chainstate as it means it comes from a previous block.
Also my bad for thinking the token can be wrong as the signing happens per Transaction so the store will have the right info at that time of singing.

azarovh: right. Let's add some comments there to avoid confusion in the future

chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -700,10 +781,56 @@ impl<'a> RandomTxMaker<'a> {
...
-                            OutputValue::Coin(new_value),
-                            Destination::AnyoneCanSpend,
-                        )
+                        let destination =
+                            key_manager.new_destination(self.chainstate.get_chain_config(), rng);
+                        match rng.gen_range(0..5) {
+                            0 => TxOutput::LockThenTransfer(
+                                OutputValue::Coin(new_value),
+                                destination,
+                                OutputTimeLock::ForBlockCount(rng.gen_range(0..5)),
```
iljakuklic, line 790: How did you pick the `0..5` interval? Should it be made configurable or a constant?

chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -700,10 +781,56 @@ impl<'a> RandomTxMaker<'a> {
...
+                            ),
+                            2 => TxOutput::LockThenTransfer(
+                                OutputValue::Coin(new_value),
+                                destination,
+                                OutputTimeLock::ForSeconds(
+                                    self.chainstate
+                                        .get_chain_config()
+                                        .target_block_spacing()
+                                        .as_secs()
+                                        * rng.gen_range(0..5),
```
iljakuklic, line 811: Seems it would be better to scale the interval by the block spacing and generate number of seconds with one second resolution rather than multiples of `target_block_spacing`. Same for `LockThenTransfer` below.

iljakuklic (approved):
Looks good. Though I am not very familiar with the API server component, so might have missed things.

chainstate/test-framework/src/random_tx_maker.rs
TheQuantumPhysicist: I would prefer to put this in a constant and explain above the constant where this number came from. Same for all other magic numbers.

chainstate/test-framework/src/random_tx_maker.rs
TheQuantumPhysicist: Agreed. This limitation is unnecessary. I can also live with keeping this like this if we add a random offset in the range [0, target_spacing]. This would ensure that we cover all possible values.

TheQuantumPhysicist (approved):

PR #1708, opened 2024-04-03
Enforce block index invariants in chainstate

ImplOfAnImpl:
* `is_persistent` flag was added to `BlockIndex`, which means that this `BlockIndex` will never be deleted (until we implement purging, that is). In practice, a `BlockIndex` is persistent iff the corresponding `Block` is in the chainstate db.
* In `ChainstateInterface`, `get_block_index/get_gen_block_index` were replaced with `get_persistent_block_index/get_any_block_index`  and `get_persistent_gen_block_index/get_any_gen_block_index`; the `get_persistent` variants only return a `BlockIndex` is it's marked as "persistent".
* Some block index related consistency checks were added to `TestFramework`.
* Chainstate consistency checks were added, which are performed once at startup and then after a block is processed or invalidated; they are enabled by default for regtest only (but can be enabled via a command-line argument for mainnet and testnet too).


chainstate/src/detail/chainstateref/consistency_checker.rs
```diff
@@ -0,0 +1,221 @@
...
+            min_height_with_allowed_reorg,
+        })
+    }
+
+    pub fn check(&self) {
+        log::debug!("Running chainstate consistency checks");
+
+        self.check_block_index_consistency();
+        self.check_block_height_map_consistency();
+        // TODO: add consistency checks for other maps in the chainstate db.
```
ImplOfAnImpl, line 77: Need to create an issue for this TODO.

chainstate/storage/src/lib.rs
```diff
@@ -135,6 +140,16 @@ pub trait BlockchainStorageRead:
...
+
+    /// Get all keys (block ids) from the block map. This is used in the chainstate's
+    /// "heavy" consistency checks.
+    fn get_block_map_keys(&self) -> crate::Result<BTreeSet<Id<Block>>>;
+    /// Get the entire block index map as BTreeMap. This is used in the chainstate's
+    /// "heavy" consistency checks.
+    fn get_block_index_map(&self) -> crate::Result<BTreeMap<Id<Block>, BlockIndex>>;
+    /// Get the entire block-by-height map as BTreeMap. This is used in the chainstate's
+    /// "heavy" consistency checks.
+    fn get_block_by_height_map(&self) -> crate::Result<BTreeMap<BlockHeight, Id<GenBlock>>>;
```
TheQuantumPhysicist, line 152: Please add in the comment that this only returns main-chain blocks. An argument can be made that this is the only option due to return type, but it cannot be excluded that there's an implementation bug that fills the returned map with first-come-first-serve.

chainstate/src/detail/chainstateref/consistency_checker.rs
TheQuantumPhysicist: Please add the issue number to the TODO

chainstate/src/detail/chainstateref/consistency_checker.rs
```diff
@@ -0,0 +1,221 @@
...
+use logging::log;
+
+use super::calc_min_height_with_allowed_reorg;
+
+// Certain tests check for this panic message.
+const PANIC_MSG: &str = "Inconsistent chainstate";
+
+pub struct ConsistencyChecker<'a, 'b, DbTx> {
+    _db_tx: &'a DbTx,
+    chain_config: &'b ChainConfig,
```
TheQuantumPhysicist, line 34: Just use an Arc instead of complicating this with another lifetime...

chainstate/src/detail/chainstateref/consistency_checker.rs
```diff
@@ -0,0 +1,221 @@
...
+        log::debug!("Running chainstate consistency checks");
+
+        self.check_block_index_consistency();
+        self.check_block_height_map_consistency();
+        // TODO: add consistency checks for other maps in the chainstate db.
+    }
+
+    /// Check the block map vs block index map consistency.
+    fn check_block_index_consistency(&self) {
+        for merged in self
```
TheQuantumPhysicist, line 82: Please add a comment, something in the lines of "Ensure that every block has a corresponding block index, by looping over all blocks and block indices in a merged iterator that indicates whether both are found or only one of either is found"

chainstate/src/detail/query.rs
```diff
@@ -124,14 +124,30 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         Ok(res)
     }
 
-    pub fn get_block_index(
+    pub fn get_persistent_block_index(
```
TheQuantumPhysicist, line 127: Maybe all these "persistent" should be renamed to "persisted"?

chainstate/src/interface/chainstate_interface.rs
```diff
@@ -127,11 +129,31 @@ pub trait ChainstateInterface: Send + Sync {
         headers: Vec<SignedBlockHeader>,
     ) -> Result<(Vec<SignedBlockHeader>, Vec<SignedBlockHeader>), ChainstateError>;
 
-    fn get_block_index(&self, id: &Id<Block>) -> Result<Option<BlockIndex>, ChainstateError>;
-    fn get_gen_block_index(
+    /// Return the block index given a block id.
+    /// This function will only return persistent block indices; if a non-persistent index
```
TheQuantumPhysicist, line 133: I think a good definition of "persistent" or "persisted" (in case they were renamed) is needed. Can we add a definition here in the lines of "a persisted block index is a block index whose block is stored in the database"

chainstate/src/detail/chainstateref/consistency_checker.rs
ImplOfAnImpl: > Just use an Arc instead of complicating this with another lifetime...

Actually, separate lifetimes are not needed here, so I've used `'a` for both of the references.

chainstate/types/src/block_index.rs
```diff
@@ -36,7 +36,12 @@ pub struct BlockIndex {
     #[codec(compact)]
     /// The total number of transactions up to this block
     chain_transaction_count: u128,
+    /// The block status.
     status: BlockStatus,
+    /// If true, this block index will stay in the db forever; otherwise, it may be deleted
```
TheQuantumPhysicist, line 41: It's not forever. It's until the reorg depth is reached (when pruning is implemented); then, if it's mainchain, it'll stay, otherwise it'll be deleted.

chainstate/types/src/gen_block_index.rs
```diff
@@ -81,16 +81,43 @@ impl GenBlockIndex {
...
+                let eq = Arc::ptr_eq(g1, g2);
+                debug_assert!(eq, "Attempt to compare different geneses");
+                eq
+            }
+            (Self::Block(_), Self::Genesis(_)) | (Self::Genesis(_), Self::Block(_)) => false,
+        }
+    }
+}
+
+impl Eq for GenBlockIndex {}
```
TheQuantumPhysicist, line 123: Please remove these. Equality is not a proper thing for a block index.

Wherever you need this, replace it by comparing ids if necessary.

TheQuantumPhysicist (commented):
First round done. I'll review again tomorrow.

chainstate/types/src/gen_block_index.rs
```diff
@@ -81,16 +81,43 @@ impl GenBlockIndex {
...
+    }
 }
 
 impl From<BlockIndex> for GenBlockIndex {
     fn from(bi: BlockIndex) -> Self {
         GenBlockIndex::Block(bi)
     }
 }
+
+impl PartialEq for GenBlockIndex {
```
TheQuantumPhysicist, line 109: As discussed about equality, it should go. Please use this to prevent this from happening again: https://docs.rs/static_assertions/latest/static_assertions/macro.assert_not_impl_any.html

chainstate/src/interface/chainstate_interface.rs
ImplOfAnImpl: > I think a good definition of "persistent" or "persisted" (in case they were renamed) is needed. Can we add a definition here in the lines of "a persisted block index is a block index whose block is stored in the database"

As we've discussed, now I use the word "persisted" but it refers to the block itself rather than the index object. So, the fact that the block data has to exist is kind of self-explanatory now:
```
/// Return the block index given a block id.
/// This function will only return block indices for persisted blocks; if the corresponding
/// block hasn't been persisted, None will be returned.
fn get_block_index_for_persisted_block(
```
Instead, I've added a sentence to the comments for the "any" variants, mentioning that the index object may not live long:
```
/// Return the block index given a block id. The corresponding block may or may not be persisted.
/// Note that a block index object for a non-persisted block may be deleted from the db at any moment.
fn get_gen_block_index_for_any_block(
```

azarovh (approved):
Looks good. Though I'm not a fan of new functions in `BlockchainStorageRead` and it feels like they belong to `chainstate/storage/src/internal/expensive.rs`. But I don't see the way to actually do that so let it be.

chainstate/src/detail/chainstateref/consistency_checker.rs
```diff
@@ -0,0 +1,238 @@
...
+use itertools::{EitherOrBoth, Itertools};
+use logging::log;
+
+use super::calc_min_height_with_allowed_reorg;
+
+// Certain tests check for this panic message.
+const PANIC_MSG: &str = "Inconsistent chainstate";
+
+pub struct ConsistencyChecker<'a, DbTx> {
+    _db_tx: &'a DbTx,
```
iljakuklic, line 33: What is this field for?

chainstate/src/detail/chainstateref/consistency_checker.rs
```diff
@@ -0,0 +1,238 @@
...
+            let block_id_in_block_index = block_index.block_id();
+            assert_eq!(
+                block_id,
+                block_id_in_block_index,
+                "{PANIC_MSG}: block id from BlockIndex {block_id_in_block_index} doesn't match {block_id}"
+            );
+
+            // Check the parent, if it's not genesis.
+            if let Some(parent_id) =
+                block_index.prev_block_id().classify(self.chain_config).chain_block_id()
```
iljakuklic, line 134: Also check if `prev_block_id` is `None`, the current gen-block is genesis?

iljakuklic: Maybe also check parent block id in the index matches parent block id in the block.

chainstate/src/detail/chainstateref/consistency_checker.rs
ImplOfAnImpl: IMO, as more checks gets implemented in the future, the access to the db tx from the checking functions will be needed. So I decided to add it as a field right away.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -1214,9 +1231,21 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
...
-        self.db_tx.del_block_index(*block_id).map_err(BlockError::from)
+    pub fn del_non_persistent_block_index(
+        &mut self,
+        block_id: &Id<Block>,
+    ) -> Result<(), BlockError> {
+        if let Some(existing_block_index) = self.db_tx.get_block_index(block_id)? {
+            assert!(
+                !existing_block_index.is_persistent(),
+                "Trying to delete a persistent block index"
+            );
```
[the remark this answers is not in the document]
ImplOfAnImpl: > Any way of avoiding the panic here with some clever (but not over-engineered) type trickery?

I can't think of any.

chainstate/src/detail/chainstateref/consistency_checker.rs
ImplOfAnImpl: > Also check if prev_block_id is None, the current gen-block is genesis?

`prev_block_id` can't be `None`, because we're dealing with the "non-gen" `BlockIndex` here. Or do you mean to check the case when `chain_block_id` returns `None` here? In that case IMO it'll be a redundant check (because `classify` returns `None` only when the block id equals the genesis id obtained from chain config).

chainstate/src/detail/chainstateref/mod.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I think this should be changed to a debug_assert + an error log entry prefixed with CRITICAL. It's too dangerous to make this an assert like this for everyone. The fear isn't that this won't work as it's, but the fear is someone misunderstanding how it works in the future and causing a crash for the whole network.

chainstate/src/detail/chainstateref/consistency_checker.rs
ImplOfAnImpl: > Maybe also check parent block id in the index matches parent block id in the block.

I decided to compare the whole header then.
But note that either way this check requires loading the whole block from the db; this makes some tests run noticeably longer in debug builds (like 60s becomes 160s). So I enable this check only in release builds.

utils/src/debug_panic.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// Panic in debug builds or log the message in release builds.
+/// Note that unlike the standard `panic` macro this one requires at least one argument.
+#[macro_export]
+macro_rules! debug_panic_or_log {
+    ($($arg:tt)+) => {
+        if cfg!(debug_assertions) {
+            panic!($($arg)*);
+        } else {
+            logging::log::error!("CRITICAL - debug build would panic: {}", format!($($arg)*));
```
TheQuantumPhysicist, line 24: This is really a stupid thing to comment... but please either make it `CRITICAL:` so that it matches others. I usually grep for `CRITICAL:` or `CRITICAL ERROR:`. I'd rather change everything to the former.

chainstate/src/detail/chainstateref/consistency_checker.rs
```diff
@@ -0,0 +1,268 @@
...
+                block_id_in_block_index,
+                "{PANIC_MSG}: block id from BlockIndex {block_id_in_block_index} doesn't match {block_id}"
+            );
+
+            // If the block is persisted, calculate its id and check that it matches the id
+            // that was used as the key. Also compare the block header stored in the index vs the one
+            // in the block itself.
+            // Note that this check is relatively heavy, because it has to load the whole block from the db,
+            // so we do it only in release builds.
+            #[cfg(not(debug_assertions))]
```
TheQuantumPhysicist, line 143: Is this really necessary?

Enabling debug_assertions doesn't necessarily mean it's a release build.

ImplOfAnImpl: > Is this really necessary?

With this check disabled in debug, `cargo nextest run --workspace` takes 1m 24s on my laptop, with it enabled - 3m 31s. So yeah, it's noticeable. If I disable all the checks in the slowest test (syncing_tests::get_mainchain_headers_by_locator), like I already do for another test, the whole run now takes 1m 53s.
Probably, this is a better approach, so I'll leave it like this. Though the extra 30s are annoying too.

Another thing is that this particular check is not that useful, so probably just removing it might be better at this moment. But in the future we'll implement more checks and some of them will require loading whole blocks from the db, which is what makes it slow. So we'll return to this problem.

iljakuklic (approved):

TheQuantumPhysicist (approved):

PR #1709, opened 2024-04-04
Atomic swaps support in chainstate

azarovh:
This PR contains types required for the implementation of the following atomic swaps algorithm. Most of the business logic will be handled by the wallet (which will be done in a separate PR).

1. A generates a secret `x`
2. A creates TX1: transfer N coins to B with `TxOutput::AtomicSwap`. It reveals `H(x)` in the output and can be spent if B knows `x` or 2-of-2 multisig is signed by A&B but only after 48 hours.
    * Nothing is published nothing to reverse
    * If A publishes TX1 now he can lose access to his funds without possibility to refund because B could just ignore the process and not sign his part of refund multisig.
    * If A reveals incorrect `H(x)` then 2 scenarios are possible
        * B ignores the process and A’s funds are stuck
        * B follows the process then he will send his coins to A in TX3 with output that cannot be spent by A. Both can refund. 
3. A creates TX2: `TxOutput::Transfer` N coins from TX1 back to A with 2-of-2 `Destination::ClassicMultisig`. Multisig is signed by A.
    * Same 
4. A sends TX2 to B. B signs TX2 and return it to A. Now multisig is fully signed but can be accepted only in 48 hours.
    * Nothing is published nothing to reverse
    * B can safely sign TX2 because even if submitted immediately the tx would be rejected because of timelock.
    * Now that B signed TX2 the funds from TX1 could be returned to A after 48 hours even if B quit.
    * _How long should A wait for B to sign? B could delay the process; need to decide on a safe gap_
5. A submits TX1.
    * If B ignores the process A can return his funds in 48 hours
6. B creates TX3: transfer M coins to A with `TxOutput::AtomicSwap`. It can be spent if A knows `x` (which he knows) or multisig is signed by A&B but only after 24 hours.
    * If B ignores the process A can return his funds in 48 hours
    * If B publishes now he can lose the funds. B’s funds are now in control of A (who knows x); A just  spends B’s coins right away then wait 48 hour and get his refund..
    * If B uses larger timelock again he is increasing the risk of losing coins.
7. B creates TX4: `TxOutput::Transfer` M coins from TX3 back to B. Multisig is signed by B.
    * Same
8. B sends TX4 to A. A signs TX4 and return it to B.
    * _How long should B wait for A to submit? A could delay the process; need to decide on a safe gap_
9. B submits TX3.
    * Both can refund
    * Now A has 24 hours to spend from TX3 and reveal the secret. 
10. A spends coins from TX3. Input signature contains both signature and `x` which reveals the secret to B.
    * A is forced to reveal `x`, otherwise spending wouldn’t be authorized
11. B spends coins from TX1. B monitors txs and when he see TX3 he took x from there and singes TX1.

This is an adaptation of the original algorithm from here: https://bitcointalk.org/index.php?topic=193281.msg2224949#msg2224949

TODOs:
- [x] add fork
- [x] support in simulation test
- [ ] support in API server
- [ ] support in wallet


chainstate/tx-verifier/src/transaction_verifier/signature_check.rs
```diff
@@ -53,18 +53,22 @@ where
...
-            &destination,
-            transactable,
-            &inputs_utxos,
-            input_idx,
-        )
-        .map_err(ConnectTransactionError::SignatureVerificationFailed)
-    })
+    inputs
+        .iter()
+        .zip(transactable.signatures().unwrap().iter())
```
TheQuantumPhysicist, line 58: It's an invariant that witness size is equal to input size. You can use `zip_eq` from itertools. But before using it, make an assertion before that asserts it.

chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
```diff
@@ -77,96 +87,97 @@ impl<'a> SignatureDestinationGetter<'a> {
         tokens_view: &'a T,
         accounting_view: &'a P,
         utxos_view: &'a U,
+        spending_time: &BlockTimestamp,
```
TheQuantumPhysicist, line 90: This is wrong. Deciding the destination has nothing to do with the timestamp. We are NOT supposed to automatically make that decision based on time. We are supposed to get the user to choose what they're signing, and then we just compare against what they decided.

chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
```diff
@@ -233,8 +245,44 @@ impl<'a> SignatureDestinationGetter<'a> {
...
+                Ok(_) => htlc.spend_key,
+                Err(_) => {
+                    match htlc.refund_timelock {
+                        OutputTimeLock::UntilHeight(_)
+                        | OutputTimeLock::ForBlockCount(_)
+                        | OutputTimeLock::ForSeconds(_) => {
+                            return Err(SignatureDestinationGetterError::HashedTimelockContractUnsupportedTimelock);
+                        }
+                        OutputTimeLock::UntilTime(t) => {
+                            ensure!(spending_time <= t, SignatureDestinationGetterError::HashedTimelockContractTimelockViolation(outpoint.clone()));
```
TheQuantumPhysicist, line 278: This check shouldn't be done when getting the destination. It should be done in the spend logic.

common/src/chain/transaction/output/mod.rs
```diff
@@ -39,6 +39,15 @@ pub mod output_value;
 pub mod stakelock;
 pub mod timelock;
 
+fixed_hash::construct_fixed_hash! {
+    #[derive(Encode, Decode, serde::Serialize, serde::Deserialize)]
```
TheQuantumPhysicist, line 43: It's unacceptable to have both Encode and Serialize together. Same for Decode and Deserialize. We have learned from recent events that this can only lead to problems serde_json representation.

For RPC, you can create special wrappers for this.

chainstate/tx-verifier/src/transaction_verifier/signature_destination_getter.rs
```diff
@@ -78,150 +82,157 @@ impl<'a> SignatureDestinationGetter<'a> {
...
+                            Err(SignatureDestinationGetterError::SigVerifyOfNotSpendableOutput)
                         }
-                    },
+                        TxOutput::AtomicTransfer(_, htlc) => match input_witness {
+                            InputWitness::NoSignature(_) => todo!(),
+                            InputWitness::Standard(sig) => {
+                                let destination = AuthorizedPublicKeyHashSecretSpend::from_data(
+                                    sig.raw_signature(),
+                                )
+                                .map_or(htlc.refund_key, |_| htlc.spend_key);
```
TheQuantumPhysicist, line 146: There probably should be some enum inside that struct that tells you which arm was chosen.

common/src/chain/transaction/signature/inputsig/standard_signature.rs
```diff
@@ -67,44 +69,14 @@ impl StandardInputSignature {
...
-    }
-
     pub fn produce_uniparty_signature_for_input<T: Signable>(
         private_key: &crypto::key::PrivateKey,
         sighash_type: SigHashType,
         outpoint_destination: Destination,
         tx: &T,
         inputs_utxos: &[Option<&TxOutput>],
         input_num: usize,
+        htlc_secret: Option<HtlcSecret>,
```
TheQuantumPhysicist, line 79: I suggest creating another function that does the htlc stuff and that calls this one. This Option stuff is a bad idea.

api-server/web-server/src/api/json_helpers.rs
```diff
@@ -185,6 +185,18 @@ pub fn txoutput_to_json(
...
             })
         }
+        TxOutput::AtomicSwap(value, htlc) => {
+            json!({
+                "type": "AtomicSwap",
+                "value": outputvalue_to_json(value, chain_config, token_decimals),
+                "htlc": {
+                    "secret_hash": to_json_string(htlc.secret_hash.as_bytes()),
+                    "spend_key": Address::new(chain_config, htlc.spend_key.clone()).expect("no error").as_str(),
+                    "redun_timelock": htlc.refund_timelock,
```
TheQuantumPhysicist, line 195: refund?

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+#[derive(Zeroize, Clone, Encode, Decode, PartialEq, Eq, Debug)]
+pub struct HtlcSecret {
+    pub secret: [u8; 20],
+}
+
+fixed_hash::construct_fixed_hash! {
+    #[derive(Encode, Decode)]
+    pub struct HtlcSecretHash(20);
```
TheQuantumPhysicist, line 44: Why 20, not 32?

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub spend_key: Destination,
+
+    // or by a multisig after timelock expires making it possible to refund
+    pub refund_timelock: Duration,
+    pub refund_key: Destination,
+}
+
+#[derive(Zeroize, Clone, Encode, Decode, PartialEq, Eq, Debug)]
+pub struct HtlcSecret {
+    pub secret: [u8; 20],
```
TheQuantumPhysicist, line 39: Here too, why 20, not 32?

common/src/chain/transaction/output/htlc.rs
azarovh: I believe because 20 is enough in this case. It's not me who come up with this number:
https://lightning.network/lightning-network-paper.pdf#page=30

azarovh: I mean the size of the secret should be 32 but the resulting hash is 20

TheQuantumPhysicist: I'm OK with that.

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,71 @@
...
+        impl<'de> serde::de::Visitor<'de> for HashVisitor {
+            type Value = HtlcSecretHash;
+            fn expecting(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                fmt.write_str("a hex-encoded hash")
+            }
+            fn visit_str<E: serde::de::Error>(self, s: &str) -> Result<Self::Value, E> {
+                s.parse().map_err(serde::de::Error::custom)
+            }
+        }
+        d.deserialize_str(HashVisitor)
```
TheQuantumPhysicist, line 69: Why not simply expect hex and deserialize from hex?

TheQuantumPhysicist: I mean, isn't this what you're doing in the serialize function?

azarovh: It actually does it via FromStr implementation for fixed hash. I took it from H256

common/src/chain/transaction/output/htlc.rs
azarovh: done

common/src/chain/transaction/signature/inputsig/standard_signature.rs
```diff
@@ -155,7 +125,28 @@ impl StandardInputSignature {
...
+            | TxOutput::IssueNft(_, _, _)
+            | TxOutput::DataDeposit(_) => false,
+            TxOutput::AtomicSwap(_, _) => true,
+        });
+
+        let serialized_sig = if is_htlc_spend {
+            AuthorizedHashedTimelockContractSpend::Multisig(authorization.encode()).encode()
+        } else {
+            authorization.encode()
+        };
```
TheQuantumPhysicist, line 149: This has to go.

Please create a separate function for htlc that will call this function.

common/src/chain/transaction/signature/inputsig/standard_signature.rs
```diff
@@ -168,6 +159,35 @@ impl StandardInputSignature {
...
+            return Err(DestinationSigError::AttemptedToVerifyStandardSignatureForAnyoneCanSpend);
+        }
+        Destination::ClassicMultisig(h) => {
+            let sig_components = AuthorizedClassicalMultisigSpend::from_data(raw_signature)?;
+            verify_classical_multisig_spending(chain_config, h, &sig_components, sighash)?
+        }
+    };
+
+    Ok(())
+}
```
TheQuantumPhysicist, line 189: I would prefer keeping this as a member.

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,90 @@
...
+    random::Rng,
+};
+use serialization::{Decode, Encode};
+use zeroize::Zeroize;
+
+use super::{timelock::OutputTimeLock, Destination};
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, serde::Serialize, serde::Deserialize)]
+pub struct HashedTimelockContract {
+    // can be spend either by a specific address that knows the secret
```
TheQuantumPhysicist, line 30: spent

common/src/chain/transaction/output/htlc.rs
TheQuantumPhysicist: Please add tests below that serialize and deserialize, both in the cases of success and error.

TheQuantumPhysicist: Also, please add fixed-cases, to ensure that upgrades won't mess this up.

common/src/chain/transaction/signature/mod.rs
```diff
@@ -192,8 +197,42 @@ fn verify_standard_input_signature<T: Transactable>(
...
+                match htlc_spend {
+                    AuthorizedHashedTimelockContractSpend::Secret(_, raw_signature) => {
+                        raw_signature
+                    }
+                    AuthorizedHashedTimelockContractSpend::Multisig(raw_signature) => raw_signature,
+                }
+            }
+        },
+        None => witness.raw_signature().to_owned(),
+    };
```
TheQuantumPhysicist, line 228: This is wrong. Script execution should not be part of signature verification.

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub secret_hash: HtlcSecretHash,
+    pub spend_key: Destination,
+
+    // or by a multisig after timelock expires making it possible to refund
+    pub refund_timelock: OutputTimeLock,
+    pub refund_key: Destination,
+}
+
+#[derive(Zeroize, Clone, Encode, Decode, PartialEq, Eq, Debug)]
```
TheQuantumPhysicist, line 39: You do not need zeroize for this. The script is short-lived anyway. Zeroize is used for long-term secrets, like private keys.

common/Cargo.toml
```diff
@@ -38,6 +38,7 @@ thiserror.workspace = true
 variant_count.workspace = true
 paste.workspace = true
 regex.workspace = true
+zeroize.workspace = true
```
TheQuantumPhysicist, line 41: This can go, too.

chainstate/tx-verifier/src/transaction_verifier/atomic_swap_check.rs
```diff
@@ -0,0 +1,163 @@
...
+                    TxOutput::AtomicSwap(_, _) => true,
+                })
+            });
+
+            ensure!(
+                !atomic_swap_output,
+                ConnectTransactionError::AtomicSwapsAreNotActivated
+            );
+        }
+    };
```
TheQuantumPhysicist, line 79: Please move this to its own function

common/src/chain/transaction/signature/mod.rs
azarovh: but this is just signature extraction not verification

TheQuantumPhysicist: It's "script execution". Shouldn't be done here.

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,90 @@
...
+pub struct HtlcSecret {
+    secret: [u8; 32],
```
TheQuantumPhysicist, line 41: I think it's a good idea to support multiple hashing algorithms. Basically the same ones supported by bitcoin script.

TheQuantumPhysicist: An enum with each arm being a different algo is good enough.

chainstate/test-suite/src/tests/atomic_swap.rs
```diff
@@ -0,0 +1,617 @@
...
+        let mut rng = test_utils::random::make_seedable_rng(seed);
+        let mut tf = TestFramework::builder(&mut rng).build();
+
+        let chain_config = tf.chainstate.get_chain_config().clone();
+
+        let (alice_sk, alice_pk) = PrivateKey::new_from_rng(&mut rng, KeyKind::Secp256k1Schnorr);
+        let (bob_sk, bob_pk) = PrivateKey::new_from_rng(&mut rng, KeyKind::Secp256k1Schnorr);
+
+        let secret = HtlcSecret::new_from_rng(&mut rng);
+
```
TheQuantumPhysicist, line 265: Can you make a function that acts as a fixture to produce the same settings and then spend it in both cases, refund and normal?

TheQuantumPhysicist (commented):
First round done

common/src/chain/transaction/output/mod.rs
```diff
@@ -122,6 +123,8 @@ pub enum TxOutput {
     IssueNft(TokenId, Box<NftIssuance>, Destination),
     #[codec(index = 9)]
     DataDeposit(Vec<u8>),
+    #[codec(index = 10)]
+    AtomicSwap(OutputValue, HashedTimelockContract),
```
azarovh, line 127: rename to htlc

common/src/chain/block/block_reward.rs
```diff
@@ -88,7 +88,11 @@ impl<'a> Signable for BlockRewardTransactable<'a> {
...
 impl<'a> Transactable for BlockRewardTransactable<'a> {
-    fn signatures(&self) -> Option<&[InputWitness]> {
-        self.witness
+    fn signature(&self, index: usize) -> Option<&InputWitness> {
+        self.witness.and_then(|w| w.get(index))
+    }
+
+    fn signatures(&self) -> Vec<Option<InputWitness>> {
+        self.witness.map_or(vec![], |w| w.iter().map(|s| Some(s.clone())).collect())
     }
```
TheQuantumPhysicist, line 97: Why is this necessary? You can manipulate the slice in any way you like outside of here. This should be minimal.

common/src/chain/transaction/signature/mod.rs
TheQuantumPhysicist: We're getting closer to this making sense.

I think the right way to go is to create something that matches the evaluation of the "script" in these outputs and their corresponding inputs.

We have half of that in the destination getter. We need something similar to "evalulate signature". Both together will yield something that may call the signature verification function in this module, which certainly doesn't need to know anything about htlcs. This logic has to be moved out of here. It's something that's used by the transaction verifier, but it doesn't have to be in the transaction verifier. It can be a submodule in there. We can put it in the transaction verifier as a first step.

common/src/chain/block/block_reward.rs
azarovh: I want `verify_signature` (which requires `Transactable` trait) to work with `PartiallySignedTransaction` so that it can be used in the wallet.
The problem is that I cannot implement single getter for `Transaction` which stores `Vec<Witness>` and `PartiallySignedTransaction` which stores `Vec<Option<Witness>>` without copying data. As a partial optimisation I added new method `signature(index)` to get a reference without collect call.

TheQuantumPhysicist: Copying data for PartiallySignedTransaction is fine. It's meant to be serialized anyway.

chainstate/storage/src/internal/version.rs
```diff
@@ -19,7 +19,7 @@ use serialization::{Decode, Encode};
 pub struct ChainstateStorageVersion(u32);
 
 impl ChainstateStorageVersion {
-    pub const CURRENT: Self = Self(9);
+    pub const CURRENT: Self = Self(10);
```
TheQuantumPhysicist, line 22: Why do we need this? We didn't change the storage scheme at all, right?

common/src/chain/block/block_reward.rs
TheQuantumPhysicist: Please revert this before merging.

common/src/chain/transaction/signature/mod.rs
TheQuantumPhysicist: I will resolve this and continue at the next spot.

common/src/chain/transaction/signature/mod.rs
```diff
@@ -171,29 +155,44 @@ pub fn verify_signature<T: Transactable>(
...
-            tx,
-            inputs_utxos,
-            input_num,
-        )?,
+        InputWitness::Standard(witness) => {
+            let input_utxo = inputs_utxos.get(input_num).ok_or(
+                DestinationSigError::InvalidInputUtxoIndex(input_num, inputs_utxos.len()),
+            )?;
+
+            let raw_signature = evaluate_witness(input_utxo, witness)?;
```
TheQuantumPhysicist, line 163: So, this is a major improvement. We're getting there. But I think we can do much better.

I'll come up with a better execution model and come back to you.

chainstate/storage/src/internal/version.rs
azarovh: I added indexes to OutputValue enum

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    #[rstest]
+    #[case("\"0000000000000000000000000000000000000000\"")]
+    #[case("\"0000000000000000000000000000000000000001\"")]
+    #[case("\"ac7b960a8d03705d1ace08b1a19da3fdcc99ddbd\"")]
+    #[case("\"e4732fe6f1ed1cddc2ed4b328fff5224276e3f6f\"")]
+    #[case("\"0103b9683e51e5aba83b8a34c9b98ce67d66136c\"")]
+    fn deserialize_valid(#[case] s: String) {
+        serde_json::from_str::<HtlcSecretHash>(&s).unwrap();
```
TheQuantumPhysicist, line 117: There should be assert_eq here and the test below too. This isn't enough.

common/src/chain/transaction/output/mod.rs
```diff
@@ -31,13 +34,23 @@ use script::Script;
...
 
+fixed_hash::construct_fixed_hash! {
+    #[derive(Encode, Decode, serde::Serialize, serde::Deserialize)]
+    pub struct SecretHash(20);
+}
+
+impl rpc_description::HasValueHint for SecretHash {
+    const HINT_SER: rpc_description::ValueHint = rpc_description::ValueHint::HEX_STRING;
+}
+
```
TheQuantumPhysicist, line 53: Why not put these in the htlc module?

chainstate/storage/src/internal/version.rs
TheQuantumPhysicist: Just adding indexes doesn't warrant this as long as they're not change. Your change just made them explicit. In fact, if we changed them, we'd have much bigger problems.

common/src/chain/transaction/signature/mod.rs
```diff
@@ -108,64 +109,22 @@ pub trait Signable {
 }
 
 pub trait Transactable: Signable {
-    fn signatures(&self) -> Option<&[InputWitness]>;
+    fn signatures(&self) -> Vec<Option<InputWitness>>;
```
TheQuantumPhysicist, line 112: This is still a change I don't like. Do you still need this after having introduced mintscript? Can you please revert all the changes in this file?

TheQuantumPhysicist (commented):
Another round done. Please address the comments.

common/src/chain/transaction/signature/mod.rs
azarovh: The reason I did this is because I want `PartiallySignedTransaction` to implement this trait and unify signature verification.

chainstate/test-framework/src/signature_destination_getter.rs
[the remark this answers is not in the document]
azarovh: Yes it's used only in tests. I'd do it as a separate task

mintscript/src/translate.rs
```diff
@@ -200,6 +257,21 @@ impl<C: InputInfoProvider> TranslateInput<C> for TimelockOnly {
                 TxOutput::LockThenTransfer(_val, _dest, timelock) => {
                     Ok(WitnessScript::timelock(*timelock))
                 }
+                TxOutput::Htlc(_, htlc) => match ctx.witness() {
+                    InputWitness::NoSignature(_) => Ok(WitnessScript::TRUE),
```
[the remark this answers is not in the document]
azarovh: This is TimelockOnly check. Signatures are irrelevant here, right?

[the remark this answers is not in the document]
TheQuantumPhysicist: Still. This is wrong at a conceptual level.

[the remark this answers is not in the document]
azarovh: alright

ImplOfAnImpl (approved):

TheQuantumPhysicist (approved):

chainstate/test-framework/src/random_tx_maker.rs
```diff
@@ -794,52 +812,32 @@ impl<'a> RandomTxMaker<'a> {
 
                         let destination =
                             key_manager.new_destination(self.chainstate.get_chain_config(), rng);
-                        let target_block_spacing_sec =
-                            self.chainstate.get_chain_config().target_block_spacing().as_secs();
+                        let timelock = get_random_timelock(rng, self.chainstate);
                         match rng.gen_range(0..5) {
```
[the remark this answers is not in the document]
azarovh: it's intentional, I'd like simple transfers to appear more often. I'll make it explicit

common/src/chain/transaction/output/htlc.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use randomness::Rng;
+use serialization::{Decode, Encode};
+
+use super::{timelock::OutputTimeLock, Destination};
+
+fixed_hash::construct_fixed_hash! {
+    #[derive(Encode, Decode, serde::Serialize, serde::Deserialize)]
+    pub struct SecretHash(20);
+}
```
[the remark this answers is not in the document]
azarovh: nice, actually it's a duplicate of HtlcSecretHash

PR #1711, opened 2024-04-05
chainstate: Add a storage layer that occasionally randomly fails to the test framework

iljakuklic:
The tests now fail, needs a fix too. Will put one up in a separate PR.

### DIFF
--- a/chainstate/test-framework/src/failing_storage/read.rs
+++ b/chainstate/test-framework/src/failing_storage/read.rs
@@ -1,0 +1,269 @@
+// Copyright (c) 2024 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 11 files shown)


iljakuklic:
Changed to a draft because this will make tests fail without an accompanying fix.

chainstate/test-framework/src/failing_storage/read.rs
```diff
@@ -0,0 +1,269 @@
...
+    },
+    primitives::{Amount, BlockHeight, Id},
+};
+use pos_accounting::PoSAccountingStorageRead;
+use tokens_accounting::{TokenData, TokensAccountingStorageRead};
+use utxo::UtxosStorageRead;
+
+use super::{FailingStorage, FailingStorageTxRw};
+
+impl<T: BlockchainStorageRead> BlockchainStorageRead for FailingStorageTxRw<'_, T> {
```
TheQuantumPhysicist, line 35: Maybe I'm missing something, but shouldn't BlockchainStorageRead be already implemented for all types of internal storage? Can't we use this somehow?

```
impl<'st, B: storage::Backend> BlockchainStorageRead for super::StoreTxRo<'st, B>
```

PR #1713, opened 2024-04-05
Feature/wasm transfer tokens

OBorce:
- new functions to transfer/burn tokens

### DIFF
--- a/wasm-wrappers/src/lib.rs
+++ b/wasm-wrappers/src/lib.rs
@@ -355,7 +355,7 @@ pub fn encode_output_transfer(
     Ok(output.encode())
 }
 
-/// Given a destination address, an amount, token ID and a network type (mainnet, testnet, etc), this function
+/// Given a destination address, an amount, token ID (in address form) and a network type (mainnet, testnet, etc), this function
 /// creates an output of type Transfer for tokens, and returns it as bytes.
 #[wasm_bindgen]
 pub fn encode_output_token_transfer(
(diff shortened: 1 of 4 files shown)


wasm-wrappers/src/lib.rs
```diff
@@ -355,6 +355,24 @@ pub fn encode_output_transfer(
     Ok(output.encode())
 }
 
+/// Given a destination address, an amount, token ID and a network type (mainnet, testnet, etc), this function
+/// creates an output of type Transfer for tokens, and returns it as bytes.
+#[wasm_bindgen]
+pub fn encode_output_token_transfer(
```
TheQuantumPhysicist, line 361: This doesn't seem to be tested... please make sure everything has tests.

wasm-wrappers/src/lib.rs
```diff
@@ -355,6 +355,24 @@ pub fn encode_output_transfer(
     Ok(output.encode())
 }
 
+/// Given a destination address, an amount, token ID and a network type (mainnet, testnet, etc), this function
```
TheQuantumPhysicist, line 358: Two things:

1. I just noticed that we can specify (in a more clearer manner) that amounts are in decimal. This applies to all functions. So just add after `an amoun` in parenthesis (in decimal as a string).
2. Please mentioned after token ID in parenthesis (in address form)

TheQuantumPhysicist: Same applies to all functions

OBorce: amounts are a separate struct Amount which is constructed from atoms.

TheQuantumPhysicist (approved):
Please add positive tests too

PR #1715, opened 2024-04-12
update axum version to 0.7

OBorce:


api-server/web-server/Cargo.toml
```diff
@@ -30,4 +30,4 @@ serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true }
-tower-http = { workspace = true, features = ["cors"] }
+tower-http = { version = "0.5", features = ["cors"] }
```
iljakuklic, line 33: Why not bump the version in the workspace?

TheQuantumPhysicist: I think it's also used by jsonrpsee, that's why.

OBorce: yes jsonrpsee is still using version 0.4

TheQuantumPhysicist: Can we have them both in the workspace with different aliases? Having all of them in the workspace helps in finding new versions easily, since some UI plugins show outdated crates.

TheQuantumPhysicist (approved):

PR #1716, opened 2024-04-14
reduce the likelihood of filling the buffer on macos

OBorce:
- split the request from 1MB to 5x200KB
- add additional asserts


iljakuklic (approved):
Looks okay to me, although it feels more like an interim solution, not quite solving the underlying issue in general.

TheQuantumPhysicist (approved):

PR #1717, opened 2024-04-14
Fix cachedir in functional tests to be inside tmp dir

OBorce:
For functional tests if cachedir is not specified set it inside the tmpdir


iljakuklic (approved):

TheQuantumPhysicist (approved):
Looking OK. Please don't forget to rename `tmp` to `functional-tests-output`

PR #1720, opened 2024-04-16
Fix blockprod tx selection; make the initial target for regtest smaller.

ImplOfAnImpl:
1) The initial target for regtest was made smaller.
Before this change, the initial target multiplied by the effective balance of a "typical" test pool (with the balance equal to the minimum pledge) would be bigger than Uint256::MAX, so staking on top of genesis would always produce a block on the first iteration.
I didn't notice any slowdown in tests because of this change.
2) Blockprod now collects transactions based on `median_time_past` instead of the current time.
3) Some minor cleanup in blockprod tests.


blockprod/src/detail/mod.rs
```diff
@@ -471,16 +488,10 @@ impl BlockProduction {
...
-            // which transactions are valid for the block.
-            // TODO: Alternatively, we can construct the transaction sequence from
-            // scratch every time a different timestamp is attempted. That is more costly
-            // in terms of computational resources but will allow the node to include more
-            // transactions since the passing time may release some time locks.
             let accumulator = self
                 .collect_transactions(
                     current_tip_index.block_id(),
-                    min_constructed_block_timestamp,
+                    median_time_past,
```
ImplOfAnImpl, line 494: I've thought about this a bit more and now I'm not sure that this is the correct fix.
I mean, technically it works, but since there is only one correct value for this parameter, given a specific tip, and since mempool already knows about the current tip, perhaps it shouldn't allow to specify this timestamp at all and calculate it instead?
@iljakuklic I'm wondering what you think about this.

iljakuklic: It knows the current tip but it does not necessarily know the information about the block it's constructing the transaction sequence for. Passing the information for the candidate block being constructed allows it to include more transactions since some time locked transactions that may have not been valid for inclusion into the current tip may be valid for inclusion to the next block.

blockprod/src/detail/mod.rs
```diff
@@ -452,8 +464,13 @@ impl BlockProduction {
                     .await?;
             }
 
-            let (consensus_data, block_reward, current_tip_index, finalize_block_data) =
-                self.pull_consensus_data(input_data.clone(), self.time_getter.clone()).await?;
+            let (
+                consensus_data,
+                block_reward,
+                current_tip_index,
+                median_time_past,
```
TheQuantumPhysicist, line 471: Can we call this `curr_best_median_time_past`?

We can add documentation to this stating that "this is a conservative estimate of the median time point, at which locked transactions can be unlocked and other related calculations can be done.

blockprod/src/detail/mod.rs
TheQuantumPhysicist: My only issue with this is that the assumptions are not clear. The value `median_time_past` is not a constant for the whole chain, but depends on the block. As long as it's clear what this value is and what it's based on, this is OK if it works.

ImplOfAnImpl: > Passing the information for the candidate block being constructed allows it to include more transactions since some time locked transactions that may have not been valid for inclusion into the current tip may be valid for inclusion to the next block.

But we can't really pass anything other than this "median time past" here and this value only depends on the parent of the candidate block.

Also, as I can see, mempool does assume that the next block will be on top of the current tip. And it uses this assumption when resolving height-based time locks. So, since timestamp-based locks also only depend on the new block's parent, it looks like they should be treated similarly.

blockprod/src/detail/mod.rs
ImplOfAnImpl: > Can we call this curr_best_median_time_past?
We can add documentation to this stating that "this is a conservative estimate of the median time point, at which locked transactions can be unlocked and other related calculations can be done.

I've named it `current_tip_median_time_past`. The comment is also formulated a bit differently, let me know if you have any objections.

iljakuklic (approved):

PR #1722, opened 2024-04-16
Blockprod pow/pos separation

ImplOfAnImpl:
`BlockProduction::produce_block_with_custom_id` now delegates to one of 3 separate functions, based on the passed input data and the required consensus - `produce_block_pos`/`produce_block_pow`/`produce_block_ignore_consensus`; each of the functions works with the specific type of consensus and doesn't have to `match` it anymore (in the current master we `match` the required consensus multiple times on different levels).

Some common code that is shared among the functions has been put to the `Helper` struct.

The produce_block_xxx functions are now implemented differently:
1) Only the "pow" one spawns a new job on the thread pool (because only in this case the task is heavy and needs cancellation if the tip has changed).
2) the "pos" one constructs the block only after it has found a free time slot (though currently I don't take any advantage of this).

Also, the internal loop that the original `produce_block_with_custom_id` used to have no longer exists because IMO it was redundant (iteration could only happen on a "recoverable" mempool error, i.e. the one that happens when the tip changes; in this case we may just exit and let the wallet call blockprod again).

This is still work-in-progress. There is still more cleanup to be done (e.g. I'd get rid of the `last_used_block_timestamps` in `JobsContainer` because it's PoS specific and IMO should be stored somewhere else). Also, we need to make use of the fact that in PoS the block can now be constructed after staking has suceeded.

### DIFF
--- a/blockprod/src/detail/helper.rs
+++ b/blockprod/src/detail/helper.rs
@@ -33,8 +33,8 @@ use crate::BlockProductionError;
 
 use super::{job_manager::JobManagerInterface, CustomId};
 
-pub struct BlockProductionJobWrapper {
-    job_id: CustomId,
+pub struct Helper {
+    job_custom_id: CustomId,
     tip_block_index: GenBlockIndex,
     /// The so-called "median time past" timestamp calculated from the current tip.
     /// Note: when validating a block, the lock-time constraints of its transactions
@@ -50,9 +50,9 @@ pub struct BlockProductionJobWrapper {
     job_finished_receiver: Option<oneshot::Receiver<usize>>,
 }
 
-impl BlockProductionJobWrapper {
+impl Helper {
     pub async fn new(
-        job_id: CustomId,
+        job_custom_id: CustomId,
         job_manager_handle: &dyn JobManagerInterface,
         chainstate_handle: &ChainstateHandle,
         chain_config: &ChainConfig,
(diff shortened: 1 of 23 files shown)


blockprod/src/detail/helper.rs
```diff
@@ -0,0 +1,253 @@
...
+    starting_timestamp: BlockTimestamp,
+    max_timestamp: BlockTimestamp,
+    cancel_receiver: UnboundedReceiver<()>,
+    job_finished_receiver: Option<oneshot::Receiver<usize>>,
+}
+
+impl Helper {
+    pub async fn new(
+        job_custom_id: CustomId,
+        job_manager_handle: &dyn JobManagerInterface,
```
TheQuantumPhysicist, line 56: This can be generic

blockprod/src/detail/mod.rs
```diff
@@ -392,302 +349,398 @@ impl BlockProduction {
...
+            "Searching for a valid block ({}..={}), pool_id: {}",
+            starting_timestamp,
+            max_timestamp,
+            Address::new(&self.chain_config, *consensus_data.stake_pool_id())
+                .expect("Pool id to address cannot fail")
+        );
 
-            Arc::new(AcqRelAtomicU64::new(tip_plus_one.as_int_seconds()))
-        };
+        let search_start_time = std::time::Instant::now();
```
TheQuantumPhysicist, line 472: Please be aware that using Instant circumvents the mocked time mechanics for regtest.

TheQuantumPhysicist (commented):
I don't really see this big benefit from this separation... especially that now every time we add a new consensus type, we have to add new functions for it instead of handling a few things during block production.

ImplOfAnImpl:
Minor improvements from this PR have already been merged via other PRs; the refactoring part is superseded by #1758

PR #1724, opened 2024-04-17
Isolate networking from p2p

TheQuantumPhysicist:
We need the same networking primitives to respect that same requirements as normal connections, such as socks proxy to preserve privacy, and hence we should separate these two.


TheQuantumPhysicist (commented):
This is looking fine.

azarovh (approved):

PR #1729, opened 2024-04-20
VRF transcript with random number generator

TheQuantumPhysicist:
It seems that the Transcript used for VRF must contain a custom random number generator. This PR and #1726 are steps towards reaching this.

Now, a Transcript can be converted to one with a custom RNG. This new Transcript will produce deterministic proofs. This is just to be used with tests, so that everything can be deterministically reproduced and debugged.

~~Unfortunately, the use of a RefCell was necessary due to internal limitations.~~


test-utils/src/random.rs
```diff
@@ -39,6 +39,12 @@ impl Seed {
         self.0
     }
 
+    /// In tests where multiple seeds are needed, this can be used to deterministically create a new, different seed
+    pub fn derive_seed(&self) -> Self {
+        let new_seed = ChaChaRng::seed_from_u64(self.as_u64()).next_u64();
+        Self(new_seed)
+    }
```
iljakuklic, line 46: This may be potentially confusing for it generates the same seed every time and so has to be used in a linear way. I'd prefer slapping the `Distribution<Seed>` on `rand::distributions::Standard`, so you can generate new seeds using existing RNGs by calling `rng.gen::<Seed>()`.

TheQuantumPhysicist: Done

crypto/src/vrf/transcript/with_rng.rs
```diff
@@ -0,0 +1,151 @@
...
+            .finalize(&mut ChaChaRng::from_seed([0u8; 32]));
+
+        for _ in 0..100 {
+            assert_eq!(g1.gen::<u64>(), g2.gen::<u64>());
+        }
+    }
+
+    #[rstest]
+    #[trace]
+    #[case(1.into())]
```
azarovh, line 130: random seed can be used

azarovh (approved):

crypto/src/vrf/transcript/with_rng.rs
TheQuantumPhysicist: Done

crypto/src/vrf/transcript/with_rng.rs
```diff
@@ -0,0 +1,151 @@
...
+use super::{no_rng::VRFTranscript, traits::SignableTranscript};
+
+pub trait RngCoreAndCrypto: RngCore + CryptoRng {}
+
+impl<R> RngCoreAndCrypto for R where R: RngCore + CryptoRng {}
+
+/// A transcript that produces deterministic vrf signatures based on the provided rng
+#[must_use]
+#[derive(Clone)]
+pub struct VRFTranscriptWithRng<'a>(merlin::Transcript, Arc<Mutex<dyn RngCoreAndCrypto + 'a>>);
```
iljakuklic, line 29: I am slightly confused by having both `Arc` and a lifetime here. What is the lifetime param for?

iljakuklic (commented):
Looks good other than the `Arc` + lifetime thing. Requires some explanation.

crypto/src/vrf/transcript/with_rng.rs
TheQuantumPhysicist: It's just for being as generic as possible. This is an esoteric case, in which the lifetime of the dyn object is shorter than the smart pointer holding it.

PR #1730, opened 2024-04-23
Make secp256k1 signatures deterministic

TheQuantumPhysicist:


chainstate/test-framework/src/block_builder.rs
```diff
@@ -323,14 +324,17 @@ impl<'f> BlockBuilder<'f> {
     }
 
     /// Builds a block without processing it.
-    pub fn build(self) -> Block {
-        self.build_impl().0
+    pub fn build(self, rng: &mut (impl Rng + CryptoRng)) -> Block {
```
azarovh, line 327: I'd consider moving this to the constructor and storing a ref inside the builder. This would allow chaining methods that require `rng`. Otherwise rust will complain about multiple mutable references.

TheQuantumPhysicist: Storing a mutable reference is a big pain. Not only will this start requiring life time generic parameter, but also every test then will require multiple rngs. Now only some do due to the use of closures with mutable rng references.

common/src/chain/transaction/signature/inputsig/classical_multisig/authorize_classical_multisig.rs
```diff
@@ -233,7 +233,7 @@ pub fn sign_classical_multisig_spending(
         return Err(ClassicalMultisigSigningError::SpendeePrivateChallengePublicKeyMismatch);
     }
     let signature = private_key
-        .sign_message(&msg)
+        .sign_message(&msg, randomness::make_true_rng())
```
azarovh, line 236: I had to change this and pass rng from outside

TheQuantumPhysicist: Sounds good.

common/src/chain/transaction/signature/inputsig/authorize_pubkeyhash_spend.rs
```diff
@@ -69,7 +69,7 @@ pub fn sign_address_spending(
     }
     let msg = sighash.encode();
     let signature = private_key
-        .sign_message(&msg)
+        .sign_message(&msg, randomness::make_true_rng())
```
azarovh, line 72: why not pass as an argument?

consensus/src/lib.rs
```diff
@@ -102,17 +103,20 @@ pub enum FinalizeBlockInputData {
...
+#[allow(clippy::too_many_arguments)]
+pub fn generate_consensus_data_and_reward<G, R>(
     chain_config: &ChainConfig,
     prev_block_index: &GenBlockIndex,
     sealed_epoch_randomness: PoSRandomness,
     input_data: GenerateBlockInputData,
     block_timestamp: BlockTimestamp,
     block_height: BlockHeight,
     get_ancestor: G,
+    rng: R,
```
azarovh, line 115: it looks like this function can create true rng inside, same as `finalize_consensus_data`

TheQuantumPhysicist: Maybe. I don't think it matters.

common/src/chain/transaction/signature/inputsig/authorize_pubkeyhash_spend.rs
TheQuantumPhysicist: I don't mind.

TheQuantumPhysicist: To be clear, in many of these functions I just made a judgment call. I would change my mind if a case is presented. Feel free to change them in your next PR.

consensus/src/lib.rs
azarovh: just consistency, otherwise look suspicious

azarovh (approved):

PR #1733, opened 2024-04-24
Ensure tx_verification_simulation test is deterministic

azarovh:


wallet/src/account/mod.rs
```diff
@@ -1402,6 +1402,7 @@ impl Account {
                         challenge,
                         &sighash,
                         current_signatures,
+                        &mut randomness::make_true_rng(),
```
TheQuantumPhysicist, line 1405: randomness is already imported in this file

TheQuantumPhysicist (approved):

PR #1735, opened 2024-04-24
Feature/gui refactor wallet

OBorce:
- add support for cold wallet in the gui
- replace the usage of the controller to the rpcwallet so in the next PR we can use the command handler for the handling of the console commands

### DIFF
--- a/node-gui/Cargo.toml
+++ b/node-gui/Cargo.toml
@@ -16,6 +16,7 @@ crypto = { path = "../crypto" }
 logging = { path = "../logging" }
 mempool = { path = "../mempool" }
 node-lib = { path = "../node-lib" }
+node-comm = { path = "../wallet/wallet-node-client" }
 p2p = { path = "../p2p" }
 serialization = { path = "../serialization" }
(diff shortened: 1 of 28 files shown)


node-gui/Cargo.toml
```diff
@@ -16,13 +16,17 @@ crypto = { path = "../crypto" }
...
 p2p = { path = "../p2p" }
 serialization = { path = "../serialization" }
 subsystem = { path = "../subsystem" }
 utils = { path = "../utils" }
 wallet = { path = "../wallet" }
 wallet-controller = { path = "../wallet/wallet-controller" }
 wallet-types = { path = "../wallet/types" }
+wallet-rpc-lib = { path = "../wallet/wallet-rpc-lib"}
+wallet-rpc-client = { path = "../wallet/wallet-rpc-client"}
+wallet-cli-lib = { path = "../wallet/wallet-cli-lib"}
```
TheQuantumPhysicist, line 29: This makes me think that we should separate the backend into its own crate.

TheQuantumPhysicist (approved):

PR #1737, opened 2024-04-26
add console tab for wallets

OBorce:
- add a console tab in the wallet gui where we can input wallet-cli commands to execute on the opened wallet
- separate wallet commands in their own module
- separate management(create/open/close/exit) commands from wallet commands


TheQuantumPhysicist (approved):
Besides the issues mentioned, looks OK.

PR #1738, opened 2024-04-26
Make generic BlockUndoCache for accounting

azarovh:
Just refactoring, logic hasn't changed.

Previously we had 2 identical undo caches for `pos_accounting` and for `tokens_accounting`. I am about to add support for the 3rd accounting type (`order-accounting`) so I decided to refactor the code in advance.


TheQuantumPhysicist (approved):

TheQuantumPhysicist (approved):
We have to generalize even more. All operations that modify the database should be wrapper around "Undoable" or "Reversible". This way we can prevent problems, like the epoch bug, from happening ever again.

All in good time.

PR #1739, opened 2024-04-27
MintScript as a solution to creating composable transaction verification steps

TheQuantumPhysicist:
There are many situations where verification logic is done in separate sections. For example, the Transfer output type is a set, while LockThenTransfer is a superset, with the same requirements, but also requires to verify the timelock condition on spending. Currently in the transaction verifier, we run these checks in parallel, where we call a function that checks timelocks and we filter those output types and check those that implement LockThenTransfer to fulfill the timelock condition. And then, in another separate step, signatures are verified, which is done the same way for the Transfer output type.

This becomes much messier when adding more, complex output types. For example, in #1709, in its current state, we had to add another function that does another level of filtering for the output type when spending to check for the HTLC secret. Even making this worse, that witness evaluation is also a subset of what Transfer and LockThenTransfer does. All this ignores the other output types and their spend-conditions implications.

As an attempt to solve this problem, we introduce MintScript. An AST-based, non-encodable, composable, extensible, simple scripting language that is meant to be seen as a middle-layer (translation layer, if you will) to convert outputs and the spending outputs to them and create a verification procedure that can be generalized, but secure enough through its simplicity.

If this goes through, we'll replace the verification logic in the transaction verifier with logic that creates a script and evaluates it to decide whether a transaction is valid.

Another advantage is that depending on the growth and the demand of the market, this scripting AST can be transformed into a scriptable programming language. This is beneficial because the attack surface would be minimized, while the development of MintScript can be done and only be translated into a limited set of possibilities.

This is a prototype that I designed in a couple of hours. Design patterns and different decisions are up for discussion.


mintscript/src/script/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::{
+    helpers::{BlockchainState, SourceBlockState},
+    timelock_check::check_timelock,
+};
+
+#[derive(Debug)]
+pub enum MintScript {
+    Bool(bool),
+    And(Box<MintScript>, Box<MintScript>),
+    Or(Box<MintScript>, Box<MintScript>),
```
iljakuklic, line 39: It think a threshold construct would be more flexible/useful here in place of `And`, `Or`.

Something like:
```suggestion
    Threshold(usize, Vec<MintScript>),
```

* `Threshold(k, conds)` means at least `k` of the specified conditions have to be satisfied
* N-ary conjunction (and) becomes N-of-N threshold
* N-ary disjunction (or) becomes 1-of-N threshold
* `And`, `or` may be given a constructor function for convenience.
* Also allows for convenient encoding of the likes of 3-of-5 multisig and such which are rather awkward to expsess using `and`s / `or`s.

TheQuantumPhysicist: Done. Thanks for the suggestion.

mintscript/src/script/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+            ),
+            TxOutput::ProduceBlockFromStake(_, pool_id) => {
+                let pos_data = accounting_view
+                    .get_pool_data(pool_id)
+                    .ok()?
+                    .ok_or(Error::PoolDataNotFound(pool_id))
+                    .ok()?;
+
+                script_from_transactable(
+                    pos_data.decommission_destination().clone(),
```
azarovh, line 124: this is incorrect for BlockReward

iljakuklic:
Close this?

TheQuantumPhysicist:
Replaced with #1762

PR #1740, opened 2024-04-30
Orders accounting

azarovh:
Introduces new `TxOutput::AnyoneCanTake` which essentially represents a buy/sell order. This output creates an account with respective balances which can be filled by anyone partially/fully or withdrawn by an authority specified in the original output. Filling an order means creating a transaction that "burns" a currency that is being bought by the order and producing corresponding amount of other currency from the order account.

Allows to exchange a token for other token, coin for any token or vice versa. Given that the exchange is arbitrary and the atoms amount not divisible price arithmetics has it's restrictions. Currently it's implemented with integers arithmetics in a way that 3 tokens sold for 100 coins one by one will cost [33; 33; 34] coins each.

Suggestions for better arithmetics as well as naming are welcome.

Api-server and wallet are stubbed in this PR.


common/src/chain/transaction/output/mod.rs
```diff
@@ -123,6 +124,8 @@ pub enum TxOutput {
     IssueNft(TokenId, Box<NftIssuance>, Destination),
     #[codec(index = 9)]
     DataDeposit(Vec<u8>),
+    #[codec(index = 10)]
+    CreateOrder(OrderData),
```
TheQuantumPhysicist, line 128: Please call the arm `AnyoneCanTake`

orders-accounting/src/cache.rs
```diff
@@ -0,0 +1,258 @@
...
+    fn create_order(&mut self, id: OrderId, data: OrderData) -> Result<OrdersAccountingUndo> {
+        if self.get_order_data(&id)?.is_some() {
+            return Err(Error::OrderAlreadyExists(id));
+        }
+
+        if self.get_ask_balance(&id)?.is_some() {
+            return Err(Error::OrderAlreadyExists(id));
+        }
+
+        let ask_value = data.ask().clone();
```
azarovh, line 160: do we want to forbid orders where ask and give currencies are the same?

common/src/chain/transaction/account_outpoint.rs
```diff
@@ -110,6 +115,10 @@ pub enum AccountCommand {
     // Change the authority who can authorize operations for a token
     #[codec(index = 5)]
     ChangeTokenAuthority(TokenId, Destination),
+    #[codec(index = 6)]
+    WithdrawOrder(OrderId),
+    #[codec(index = 7)]
+    FillOrder(OrderId, OutputValue, Destination),
```
azarovh, line 121: I'm not sure that Destination here make sense but an input is required to have something to sign. For orders it authorise nothing because anyone can fill an order.

common/src/chain/config/mod.rs
```diff
@@ -329,6 +329,16 @@ impl ChainConfig {
...
+    #[must_use]
+    pub fn order_id_address_prefix(&self) -> &'static str {
+        match self.chain_type {
+            ChainType::Mainnet => "mordr",
+            ChainType::Testnet => "tordr",
+            ChainType::Regtest => "rordr",
+            ChainType::Signet => "sordr",
+        }
+    }
+
```
azarovh, line 341: do we want a fork with activation for this functionality?

api-server/web-server/src/api/json_helpers.rs
```diff
@@ -185,6 +185,14 @@ pub fn txoutput_to_json(
...
         }
+        TxOutput::AnyoneCanTake(data) => {
+            json!({
+                "type": "AnyoneCanTake",
+                "withdraw_key": Address::new(chain_config, data.withdraw_key().clone()).expect("no error").as_str(),
+                "ask_value": outputvalue_to_json(data.ask(), chain_config, token_decimals),
+                "give_value": outputvalue_to_json(data.give(), chain_config, token_decimals),
+            })
+        }
     }
```
TheQuantumPhysicist, line 196: If this PR to go through, this should be handled. We cannot keep todo!() like this because it will crash the API server when used. The API server is used by exchanges.

api-server/web-server/src/api/json_helpers.rs
```diff
@@ -321,6 +333,22 @@ pub fn tx_input_to_json(inp: &TxInput, chain_config: &ChainConfig) -> serde_json
...
+            AccountCommand::FillOrder(order_id, fill, dest) => {
+                json!({
+                    "input_type": "AccountCommand",
+                    "command": "FillOrder",
+                    "order_id": Address::new(chain_config, *order_id).expect("addressable").to_string(),
+                    "fill_value": outputvalue_to_json(fill, chain_config, token_decimals),
+                    "destination": Address::new(chain_config, dest.clone()).expect("no error").as_str(),
+                })
+            }
         },
```
TheQuantumPhysicist, line 352: Handle before merge

api-server/web-server/src/api/json_helpers.rs
TheQuantumPhysicist: I was on a wrong tip. Resolving.

common/src/chain/transaction/account_outpoint.rs
```diff
@@ -110,6 +115,10 @@ pub enum AccountCommand {
     // Change the authority who can authorize operations for a token
     #[codec(index = 5)]
     ChangeTokenAuthority(TokenId, Destination),
+    #[codec(index = 6)]
+    WithdrawOrder(OrderId),
```
TheQuantumPhysicist, line 119: Please call it CancelOrder

common/src/chain/transaction/account_outpoint.rs
TheQuantumPhysicist: I think there should be a signature for it anyway to ensure that no one tampers with it.

wallet/src/account/mod.rs
```diff
@@ -1515,6 +1517,9 @@ impl Account {
                             .token_data(token_id)
                             .map(|data| (None, Some(data.authority.clone())))
                             .ok_or(WalletError::UnknownTokenId(*token_id)),
+                        // TODO: support orders
+                        AccountCommand::WithdrawOrder(_) => unimplemented!(),
+                        AccountCommand::FillOrder(_, _, _) => unimplemented!(),
                     }
```
TheQuantumPhysicist, line 1523: Must be implemented before merging

wallet/src/account/mod.rs
```diff
@@ -2098,6 +2104,9 @@ impl Account {
                     self.find_token(token_id).is_ok()
                         || self.is_destination_mine_or_watched(address)
                 }
+                // TODO: support orders
+                AccountCommand::WithdrawOrder(_) => unimplemented!(),
+                AccountCommand::FillOrder(_, _, _) => unimplemented!(),
             },
```
TheQuantumPhysicist, line 2110: Must be implemented before merging

wallet/src/account/mod.rs
```diff
@@ -2459,6 +2468,8 @@ fn group_preselected_inputs(
                             output.clone(),
                         )))
                     }
+                    // TODO: support orders
+                    TxOutput::AnyoneCanTake(_) => unimplemented!(),
```
TheQuantumPhysicist, line 2472: Must be implemented before merging

wallet/src/account/mod.rs
```diff
@@ -2502,6 +2513,9 @@ fn group_preselected_inputs(
                             .ok_or(WalletError::OutputAmountOverflow)?,
                     )?;
                 }
+                // TODO: support orders
+                AccountCommand::WithdrawOrder(_) => unimplemented!(),
+                AccountCommand::FillOrder(_, _, _) => unimplemented!(),
             },
```
TheQuantumPhysicist, line 2519: Must be implemented before merging

wallet/src/account/output_cache/mod.rs
```diff
@@ -716,6 +719,8 @@ impl OutputCache {
                     | TxOutput::CreateDelegationId(_, _)
                     | TxOutput::IssueFungibleToken(_)
                     | TxOutput::ProduceBlockFromStake(_, _) => false,
+                    // TODO: support orders
+                    TxOutput::AnyoneCanTake(_) => unimplemented!(),
```
TheQuantumPhysicist, line 723: Must be implemented before merging

wallet/src/account/output_cache/mod.rs
```diff
@@ -725,6 +730,9 @@ impl OutputCache {
                 | AccountCommand::UnfreezeToken(token_id)
                 | AccountCommand::ChangeTokenAuthority(token_id, _)
                 | AccountCommand::UnmintTokens(token_id) => frozen_token_id == token_id,
+                // TODO: support orders
+                AccountCommand::WithdrawOrder(_) => unimplemented!(),
+                AccountCommand::FillOrder(_, _, _) => unimplemented!(),
             },
```
TheQuantumPhysicist, line 736: Must be implemented before merging

TheQuantumPhysicist: Many other `unimplemented!()` instances

wallet/src/account/currency_grouper/mod.rs
```diff
@@ -57,6 +57,8 @@ pub(crate) fn group_outputs<T, Grouped: Clone>(
                     get_tx_output(&output).clone(),
                 )))
             }
+            // TODO: support orders
+            TxOutput::AnyoneCanTake(_) => unimplemented!(),
```
TheQuantumPhysicist, line 61: Must be implemented before merging

serialization/src/extras/non_empty_vec.rs
```diff
@@ -21,7 +21,7 @@ use serialization_core::{Decode, Encode};
 /// - If the Vec has data, it encodes to just the Vec, the Option is omitted
 /// - If the Vec has no data, it encodes to None
 /// - Some(vec![]) and None are equivalent when encoded, but when decoded result in None
-#[derive(Debug, PartialEq, Eq, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, PartialEq, Eq, Clone, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
```
TheQuantumPhysicist, line 24: I'm not sure this is a necessary addition. I tracked this to the source and these can be removed. Please don't add them unless necessary.

orders-accounting/src/storage/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+pub mod db;
+pub mod in_memory;
+
+pub trait OrdersAccountingStorageRead {
+    type Error: std::error::Error;
+
+    fn get_order_data(&self, id: &OrderId) -> Result<Option<OrderData>, Self::Error>;
+    fn get_ask_balance(&self, id: &OrderId) -> Result<Option<Amount>, Self::Error>;
+    fn get_give_balance(&self, id: &OrderId) -> Result<Option<Amount>, Self::Error>;
+}
```
TheQuantumPhysicist, line 31: Please provide docs on what each of these terms mean in the context of someone making an order.

common/src/chain/config/mod.rs
TheQuantumPhysicist: Yes

orders-accounting/src/cache.rs
TheQuantumPhysicist: Yes

orders-accounting/src/storage/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+
+pub mod db;
+pub mod in_memory;
+
+pub trait OrdersAccountingStorageRead {
+    type Error: std::error::Error;
+
+    fn get_order_data(&self, id: &OrderId) -> Result<Option<OrderData>, Self::Error>;
+    fn get_ask_balance(&self, id: &OrderId) -> Result<Option<Amount>, Self::Error>;
+    fn get_give_balance(&self, id: &OrderId) -> Result<Option<Amount>, Self::Error>;
```
TheQuantumPhysicist, line 30: The wording here is very confusing. Usually it's ask/bid, not ask/give. Not sure whether you have a reason for a different naming.

orders-accounting/src/cache.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    fn get_give_balance(&self, id: &OrderId) -> Result<Option<Amount>> {
+        let parent_supply = self.parent.get_give_balance(id).map_err(|_| Error::ViewFail)?;
+        let local_delta = self.data.give_balances.data().get(id).cloned();
+        combine_amount_delta(&parent_supply, &local_delta).map_err(Error::AccountingError)
+    }
+}
+
+impl<P: OrdersAccountingView> OrdersAccountingOperations for OrdersAccountingCache<P> {
+    fn create_order(&mut self, id: OrderId, data: OrderData) -> Result<OrdersAccountingUndo> {
```
TheQuantumPhysicist, line 151: Reminder: Order id calculation should be automated + it should be enforced to use the automated results. We don't want people to be able to use any id they want.

azarovh: it is calculated in the tx-verifier based on tx input.

orders-accounting/src/cache.rs
```diff
@@ -0,0 +1,258 @@
...
+        }))
+    }
+
+    fn fill_order(&mut self, id: OrderId, fill_value: OutputValue) -> Result<OrdersAccountingUndo> {
+        let filled_amount = calculate_fill_order(self, id, &fill_value)?;
+
+        let ask_balance = self.get_ask_balance(&id)?.ok_or(Error::OrderAskBalanceNotFound(id))?;
+        let give_balance =
+            self.get_give_balance(&id)?.ok_or(Error::OrderGiveBalanceNotFound(id))?;
+
```
TheQuantumPhysicist, line 211: No check that the token vs coin pair are correct?

azarovh: checks are inside `calculate_fill_order` function

common/src/chain/transaction/output/output_value.rs
```diff
@@ -34,6 +45,13 @@ impl OutputValue {
             OutputValue::TokenV0(_) | OutputValue::TokenV1(_, _) => None,
         }
     }
+
+    pub fn amount(&self) -> Amount {
+        match self {
+            OutputValue::Coin(v) | OutputValue::TokenV1(_, v) => *v,
+            OutputValue::TokenV0(_) => panic!("deprecated token version"),
+        }
+    }
```
TheQuantumPhysicist, line 54: This function feels like playing with fire to me. I think it should go. 

If you need something like this, make it local to where you're using it.

orders-accounting/src/operations.rs
```diff
@@ -0,0 +1,64 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, VariantCount)]
+pub enum OrdersAccountingUndo {
+    CreateOrder(CreateOrderUndo),
+    WithdrawOrder(WithdrawOrderUndo),
+    FillOrder(FillOrderUndo),
+}
+
+pub trait OrdersAccountingOperations {
+    fn create_order(&mut self, id: OrderId, data: OrderData) -> Result<OrdersAccountingUndo>;
+    fn withdraw_order(&mut self, id: OrderId) -> Result<OrdersAccountingUndo>;
```
TheQuantumPhysicist, line 60: cancel_order

orders-accounting/src/operations.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub(crate) undo_data: Option<DataDeltaUndo<OrderData>>,
+    pub(crate) ask_balance: Amount,
+    pub(crate) give_balance: Amount,
+}
+
+#[must_use]
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, VariantCount)]
+pub enum OrdersAccountingUndo {
+    CreateOrder(CreateOrderUndo),
+    WithdrawOrder(WithdrawOrderUndo),
```
TheQuantumPhysicist, line 54: CancelOrder

orders-accounting/src/price_calculation.rs
```diff
@@ -0,0 +1,197 @@
...
+        .ok_or(Error::OrderAskBalanceNotFound(order_id))?;
+    let give_balance = view
+        .get_give_balance(&order_id)
+        .map_err(|_| crate::Error::ViewFail)?
+        .ok_or(Error::OrderGiveBalanceNotFound(order_id))?;
+
+    {
+        let ask_balance = match order_data.ask() {
+            OutputValue::Coin(_) => OutputValue::Coin(ask_balance),
+            OutputValue::TokenV0(_) => unreachable!(),
```
TheQuantumPhysicist, line 45: Please don't write code that crashes for branches that are disabled by forks and settings. Just return an error. This is spread everywhere.

common/src/chain/transaction/signature/tests/sign_and_mutate.rs
```diff
@@ -1088,6 +1088,7 @@ fn mutate_output(_rng: &mut impl Rng, tx: &SignedTransactionWithUtxo) -> SignedT
         TxOutput::IssueFungibleToken(_) => unreachable!(), // TODO: come back to this later
         TxOutput::IssueNft(_, _, _) => unreachable!(),     // TODO: come back to this later
         TxOutput::DataDeposit(_) => unreachable!(),
+        TxOutput::AnyoneCanTake(_) => unreachable!(),
```
TheQuantumPhysicist, line 1091: Please check if we have tests that check mutating account spendings/commands

chainstate/tx-verifier/src/transaction_verifier/mod.rs
```diff
@@ -119,11 +127,14 @@ pub struct TransactionVerifier<C, S, U, A, T> {
     tokens_accounting_cache: TokensAccountingCache<T>,
     tokens_accounting_block_undo: AccountingBlockUndoCache<TokenAccountingUndo>,
 
+    orders_accounting_cache: OrdersAccountingCache<O>,
+    orders_accounting_block_undo: AccountingBlockUndoCache<OrdersAccountingUndo>,
```
TheQuantumPhysicist, line 131: All these should be somehow unified in the future.

serialization/src/extras/non_empty_vec.rs
azarovh: It's required because I added `AccountCommand::FillOrder` with `OutputValue`. Comparison is used to store `TxInput` in BTreeSet and check uniqueness of inputs.

orders-accounting/src/cache.rs
```diff
@@ -0,0 +1,258 @@
...
+        // in case the order is completely filled it can be removed
+        let undo_data = if fill_value.amount() == ask_balance {
+            ensure!(
+                filled_amount == give_balance,
```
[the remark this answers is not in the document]
azarovh: yeah, I was so concentrated on making it correct that I missed the point that this doesn't make much practical sense. So we can't clean up orders automatically. Can it be solved with `sweep_orders` function in the wallet for now?

common/src/chain/transaction/signature/tests/sign_and_mutate.rs
azarovh: yes we do

orders-accounting/src/price_calculation.rs
```diff
@@ -0,0 +1,197 @@
...
+
+use common::{
+    chain::{output_value::OutputValue, OrderId},
+    primitives::Amount,
+};
+use utils::ensure;
+
+use crate::{error::Result, Error, OrdersAccountingView};
+
+pub fn calculate_fill_order(
```
TheQuantumPhysicist, line 24: Please make sure you have tests that incrementally fill an order until it's empty. I still haven't checked the tests, so this may or may not be there.

orders-accounting/src/cache.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: I can imagine situations where it's impossible to fill an order. If you ask for a=10A for b=3B tokens, that means, the price is a/b=3, and you buy `(a/b)*3B=3*3B=9A`. The last 1A will never be filled no matter what you do. I don't know what valid ways there are to fill the order to the latter when there's such a difference, but we have to be careful with this as if this is done wrong, it can be abused.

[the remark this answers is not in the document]
azarovh: not exactly. I don't calculated price anywhere to avoid integer division. Instead the amounts exchanged are calculated which gives you (10A*3B)/3B so when buying 3b you would pay 10A. In case of filling it one buy one the actual "price" paid would be: 3A for the first and second token and the last token would cost you 4A to buy.

orders-accounting/src/price_calculation.rs
azarovh: here is the test:
https://github.com/mintlayer/mintlayer-core/blob/1feed1126346c9c4aa14a593da070adf4548a0bb/orders-accounting/src/tests/operations.rs#L523

chainstate/test-suite/src/tests/orders_tests.rs
```diff
@@ -0,0 +1,664 @@
...
+            Some(give_amount),
+            tf.chainstate.get_order_give_balance(&order_id).unwrap()
+        );
+    });
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn create_two_orders_same_tx(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 122: I think this can be broken into two tests. One with the same order twice in two different transactions, and one with two difference orders in one transaction.

chainstate/test-suite/src/tests/orders_tests.rs
```diff
@@ -0,0 +1,664 @@
...
+                InputWitness::NoSignature(None),
+            )
+            .add_output(TxOutput::Transfer(
+                OutputValue::TokenV1(token_id, filled_amount),
+                Destination::AnyoneCanSpend,
+            ))
+            .add_output(TxOutput::Transfer(
+                OutputValue::Coin(left_to_fill),
+                Destination::AnyoneCanSpend,
+            ))
```
TheQuantumPhysicist, line 262: This output is confusing... so I spent some amount to fill an order, and got the corresponding token in the first output. But then why does this output exist here, one that contains the amount left in the order?

chainstate/test-suite/src/tests/orders_tests.rs
```diff
@@ -0,0 +1,664 @@
...
+            Some(left_to_fill),
+            tf.chainstate.get_order_ask_balance(&order_id).unwrap()
+        );
+        assert_eq!(
+            Some((give_amount - filled_amount).unwrap()),
+            tf.chainstate.get_order_give_balance(&order_id).unwrap()
+        );
+
+        // Fill the rest of the order
+        let fill_value = OutputValue::Coin(left_to_fill);
```
TheQuantumPhysicist, line 281: There should be tests that attempt `fill_value + 1` and `fill_value - 1`

chainstate/test-suite/src/tests/orders_tests.rs
```diff
@@ -0,0 +1,664 @@
...
+            None,
+            tf.chainstate.get_order_give_balance(&order_id).unwrap()
+        );
+    });
+}
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn withdraw_order_check_signature(#[case] seed: Seed) {
```
TheQuantumPhysicist, line 322: We need also tests to cancel (not withdraw, please rename) orders that are partially filled

TheQuantumPhysicist (commented):
First round is done. I guess this needs many reviews. Please resolve all the comments, rebase, and I'll start over.

chainstate/test-suite/src/tests/orders_tests.rs
azarovh: there are tests for that, actual the reason why I chose 'withdraw' was to heighlight that this functionality gets from the order whatever is left

chainstate/test-suite/src/tests/orders_tests.rs
azarovh: this is utxo for taker to fill the order completely in a later tx

chainstate/test-suite/src/tests/orders_tests.rs
azarovh: these tests are in chainstate/constraints-value-accumulator/src/tests/orders_constraints.rs

orders-accounting/src/storage/mod.rs
azarovh: So my idea was to represent order as an exchange of OutputValue types: you ask for x amount of A and gives y amount of B in return. This looks both universal and precise (i.e. avoids storing price).
Imo ask/bid are more applicable for trading with buy/sell orders that both parties can submit.

chainstate/test-suite/src/tests/orders_tests.rs
TheQuantumPhysicist: This doesn't seem to be done

azarovh: there are two tests now: create_two_2_orders_same_tx and create_two_2_orders_same_block. Isn't it what you suggested?

chainstate/test-suite/src/tests/orders_tests.rs
TheQuantumPhysicist: We need a test for that here, in this file, at the blockchain level.

common/src/chain/transaction/account_outpoint.rs
TheQuantumPhysicist: Please make sure to grep the whole code and change every withdraw to cancel... I can imagine comments having it. Be aware that the word "withdraw" is used for delegations.

chainstate/constraints-value-accumulator/src/tests/orders_constraints.rs
```diff
@@ -0,0 +1,730 @@
...
+            block_height,
+            &orders_db,
+            &pos_db,
+            &inputs,
+            &input_utxos,
+        );
+
+        assert_eq!(
+            result.unwrap_err(),
+            Error::OrdersAccountingError(orders_accounting::Error::OrderOverflow(order_id))
```
TheQuantumPhysicist, line 310: I wouldn't call this error overflow... maybe OrderOverbid?

TheQuantumPhysicist: I'd like to reserve the word "overflow" for integer computation issues.

chainstate/test-suite/src/tests/orders_tests.rs
TheQuantumPhysicist: Please explain it in comments. Explain these steps.

TheQuantumPhysicist: I would prefer to use a different number/variable to make this behavior clear.

TheQuantumPhysicist: Well, thinking about it, why do we have to include this in this transaction? I'd rather not have it here to have the intent clear.

orders-accounting/src/cache.rs
TheQuantumPhysicist: Are you sure it's not possible to use a custom id? This is manually enforced for staking pools.

azarovh: It is possible to pass custom id to this function, e.g. in tests. But in transaction verifier it is called with id generated from utxo input.
For staking pool the situation is different because `TxInput::CreateStakePool(PoolId, _)` has pool id it it. So manual enforcement is required. But not for tokens and orders

chainstate/test-suite/src/tests/orders_tests.rs
azarovh: the idea was to fill partially and then fill the order completely. For that I carry the change across txs and made it precisely the same to hit that case.

chainstate/test-suite/src/tests/orders_tests.rs
azarovh: ok I see

orders-accounting/src/cache.rs
```diff
@@ -0,0 +1,268 @@
...
+        match self.data.order_data.get_data(id) {
+            accounting::GetDataResult::Present(d) => Ok(Some(d.clone())),
+            accounting::GetDataResult::Deleted => Ok(None),
+            accounting::GetDataResult::Missing => {
+                Ok(self.parent.get_order_data(id).map_err(|_| Error::ViewFail)?)
+            }
+        }
+    }
+
+    fn get_ask_balance(&self, id: &OrderId) -> Result<Option<Amount>> {
```
azarovh, line 138: just a reminder to get back and fix Some(0) problem

orders-accounting/src/tests/operations.rs
```diff
@@ -0,0 +1,558 @@
...
+
+    assert_eq!(None, cache.get_order_data(&order_id).unwrap().as_ref());
+    assert_eq!(
+        Some(Amount::ZERO),
+        cache.get_ask_balance(&order_id).unwrap()
+    );
+    assert_eq!(
+        Some(Amount::ZERO),
+        cache.get_give_balance(&order_id).unwrap()
+    );
```
TheQuantumPhysicist, line 231: This is probably a good example where canceling doesn't lead to None... all these accounting primitives should return zero. We should discuss it at some point.

orders-accounting/src/price_calculation.rs
```diff
@@ -0,0 +1,204 @@
...
+        ensure_currencies_and_amounts_match(order_id, &ask_balance, fill_value)?;
+    }
+
+    calculate_filled_amount_impl(ask_balance, give_balance, output_value_amount(fill_value)?)
+        .ok_or(Error::OrderOverflow(order_id))
+}
+
+fn calculate_filled_amount_impl(ask: Amount, give: Amount, fill: Amount) -> Option<Amount> {
+    (give * fill.into_atoms()).and_then(|v| v / ask.into_atoms())
+}
```
TheQuantumPhysicist, line 58: I think this function has to have tests on its own, separate from the database stuff, as a demonstration of how it works.

orders-accounting/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+};
+
+use common::chain::output_value::OutputValue;
+
+fn output_value_amount(value: &OutputValue) -> crate::error::Result<common::primitives::Amount> {
+    match value {
+        OutputValue::Coin(amount) | OutputValue::TokenV1(_, amount) => Ok(*amount),
+        OutputValue::TokenV0(_) => Err(Error::UnsupportedTokenVersion),
+    }
+}
```
TheQuantumPhysicist, line 44: This function never stops to scare the hell out of me... I can't stop imagining how many things can go wrong with this.

My preference: Isolate this to where you need it in a private module.

orders-accounting/src/cache.rs
TheQuantumPhysicist: Please ensure there's a test for this.

orders-accounting/src/cache.rs
TheQuantumPhysicist: Please ensure this is tested, because abusing it can cause issues.

orders-accounting/src/cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    fn create_order(&mut self, id: OrderId, data: OrderData) -> Result<OrdersAccountingUndo> {
+        ensure!(
+            self.get_order_data(&id)?.is_none(),
+            Error::OrderAlreadyExists(id)
+        );
+
+        ensure!(
+            self.get_ask_balance(&id)?.is_none(),
+            Error::OrderAlreadyExists(id)
+        );
```
TheQuantumPhysicist, line 161: This can fire due to the issue we know

orders-accounting/src/cache.rs
azarovh: Checked, it is tested in chainstate tests that an order is created with id produced from input0.

orders-accounting/src/price_calculation.rs
azarovh: Still it would be possible to create an order where result won't fit into u128. Like `give*u128::MAX/1`. Also returning an option handles zero devision.

orders-accounting/src/price_calculation.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        ensure_currencies_and_amounts_match(order_id, &ask_balance, fill_value)?;
+    }
+
+    calculate_filled_amount_impl(ask_balance, give_balance, output_value_amount(fill_value)?)
+        .ok_or(Error::OrderOverflow(order_id))
+}
+
+fn calculate_filled_amount_impl(ask: Amount, give: Amount, fill: Amount) -> Option<Amount> {
+    (give * fill.into_atoms()).and_then(|v| v / ask.into_atoms())
```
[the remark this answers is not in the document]
azarovh: Still it would be possible to create an order where result won't fit into u128. Like give*u128::MAX/1. Also returning an option handles zero devision.

orders-accounting/src/cache.rs
azarovh: added additional test

orders-accounting/src/tests/operations.rs
azarovh: I'm thinking about doing that in the next PR for all accounting types

orders-accounting/src/price_calculation.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: Returning an option is a given. You can do that with Uint256 as well.

I'm not saying you should returning a Uint256. I'm saying that multiplying u128 with u128 should return a Uint256, which then, after division, should be cast back to uint128.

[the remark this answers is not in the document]
TheQuantumPhysicist: Btw, multiplication for Uint256 also returns an Option. We changed it a while ago.

chainstate/constraints-value-accumulator/src/constraints_accumulator.rs
```diff
@@ -99,39 +96,38 @@ impl ConstrainedValueAccumulator {
                     )?;
                 }
                 TxInput::AccountCommand(_, command) => {
-                    let fee_to_deduct = accumulator.process_input_account_command(
+                    let (id, to_deduct) = accumulator.process_input_account_command(
                         chain_config,
                         block_height,
                         command,
+                        orders_accounting_view,
```
TheQuantumPhysicist, line 103: Why don't we use the tracker here? I'm not very comfortable with this half and half approach

azarovh: I believe tracker's purpose is to track overspends from accounts across multiple inputs. Here I need orders_view to calculate actual amounts being spend. Am I missing something?

azarovh: the rule that there can only be single Command per tx

chainstate/test-suite/src/tests/orders_tests.rs
```diff
@@ -0,0 +1,1092 @@
...
+        assert_eq!(
+            Some((give_amount - filled_amount).unwrap()),
+            tf.chainstate.get_order_give_balance(&order_id).unwrap()
+        );
+
+        // Create alternative chain and trigger the reorg
+        let new_best_block = tf.create_chain(&reorg_common_ancestor, 3, &mut rng).unwrap();
+        assert_eq!(tf.best_block_id(), new_best_block);
+
+        assert_eq!(None, tf.chainstate.get_order_data(&order_id).unwrap());
```
TheQuantumPhysicist, line 877: Maybe add a check showing that the order doesn't exist. I know it's implicit, but it helps to make it explicit.

orders-accounting/src/price_calculation.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+fn ensure_currencies_and_amounts_match(
+    order_id: OrderId,
+    left: &OutputValue,
+    right: &OutputValue,
```
ImplOfAnImpl, line 68: Rename "left" => "asked" and "right" => "filled"?
(or any other names that would suggest why one value should be bigger than the other)
Same for amount1/amount2 in the body.

orders-accounting/src/price_calculation.rs
```diff
@@ -0,0 +1,248 @@
...
+                Error::OrderOverbid(order_id, *amount1, *amount2)
+            );
+            ensure!(id1 == id2, Error::CurrencyMismatch);
+            Ok(())
+        }
+        (OutputValue::Coin(_), OutputValue::TokenV1(_, _))
+        | (OutputValue::TokenV1(_, _), OutputValue::Coin(_)) => Err(Error::CurrencyMismatch),
+        (OutputValue::TokenV0(_), _) | (_, OutputValue::TokenV0(_)) => {
+            unreachable!()
+        }
```
ImplOfAnImpl, line 90: As I can see, the caller first calls `ensure_currencies_and_amounts_match` and then checks that `fill_value` is not `TokenV0`, so this code is not unreachable.

orders-accounting/src/error.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use common::{chain::OrderId, primitives::Amount};
+
+#[derive(thiserror::Error, Debug, PartialEq, Eq, Clone)]
+pub enum Error {
+    #[error("Accounting storage error")]
+    StorageError(#[from] chainstate_types::storage_result::Error),
+    #[error("Base accounting error: {0}")]
+    AccountingError(#[from] accounting::Error),
+    #[error("Order already exist: `{0}`")]
```
ImplOfAnImpl, line 24: exists

orders-accounting/src/error.rs
```diff
@@ -0,0 +1,79 @@
...
+    OrderAlreadyExists(OrderId),
+    #[error("Data for order {0}` not found")]
```
ImplOfAnImpl, line 26: Here and below you have a unmatched backtick in the error messages.
Also, `AccountingError`'s message (above) doesn't use backtick at all.

mempool/src/error/ban_score.rs
```diff
@@ -480,6 +487,44 @@ impl MempoolBanScore for CheckTransactionError {
...
+            Error::InvariantOrderDataNotFoundForUndo(_) => 100,
+            Error::InvariantOrderAskBalanceNotFoundForUndo(_) => 100,
+            Error::InvariantOrderAskBalanceChangedForUndo(_) => 100,
+            Error::InvariantOrderGiveBalanceNotFoundForUndo(_) => 100,
+            Error::InvariantOrderGiveBalanceChangedForUndo(_) => 100,
+            Error::InvariantOrderDataExistForCancelUndo(_) => 100,
+            Error::InvariantOrderAskBalanceExistForCancelUndo(_) => 100,
+            Error::InvariantOrderGiveBalanceExistForCancelUndo(_) => 100,
+            Error::InvariantNonzeroAskBalanceForMissingOrder(_) => 100,
+            Error::InvariantNonzeroGiveBalanceForMissingOrder(_) => 100,
```
ImplOfAnImpl, line 517: Not that it's a big deal, but IMO it's better for "invariant"  errors to have zero ban score, because they can't be peer's fault.
Same for `chainstate/src/detail/ban_score.rs`

common/src/chain/tokens/issuance.rs
```diff
@@ -99,9 +99,9 @@ impl IsTokenUnfreezable {
     }
 }
 
-// Indicates whether a token is frozen at the moment or not. If it is then no operations wish this token can be performed.
-// Meaning transfers, burns, minting, unminting, supply locks etc. Frozen token can only be unfrozen
-// is such an option was provided while freezing.
+/// Indicates whether a token is frozen at the moment or not. If it is then no operations with this token can be performed.
+/// Meaning transfers, burns, minting, unminting, supply locks etc. Frozen token can only be unfrozen
+/// is such an option was provided while freezing.
```
ImplOfAnImpl, line 104: "is" -> "if"

common/src/chain/transaction/output/mod.rs
```diff
@@ -140,6 +141,8 @@ pub enum TxOutput {
     /// Transfer an output under Hashed TimeLock Contract.
     #[codec(index = 10)]
     Htlc(OutputValue, Box<HashedTimelockContract>),
+    #[codec(index = 11)]
+    AnyoneCanTake(Box<OrderData>),
 }
```
ImplOfAnImpl, line 146: Not that I insist on renaming this, but I wonder why this name was chosen. Shouldn't it be "CreateOrder" or something similar?

ImplOfAnImpl (approved):

chainstate/constraints-value-accumulator/src/constraints_accumulator.rs
azarovh: will be reworked in a separate pr

common/src/chain/transaction/output/mod.rs
azarovh: The idea is that the output names describe a use case rather then specific operation performed. Like `Transfer` which doesn't transfer anything per se.

mempool/src/error/ban_score.rs
azarovh: Maybe, not sure, it's just how it's done for every other invariant error

orders-accounting/src/price_calculation.rs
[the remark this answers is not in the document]
azarovh: ok done

TheQuantumPhysicist (approved):
Everything seems fine except for the accumulator. Stuff need to be discussed further.

PR #1742, opened 2024-05-02
Pool balance retrieval in chainstate

ImplOfAnImpl:
The function `get_stake_pool_balances_at_heights` has been added to chainstate, which obtains balances of the specified pools at each height in the specified range.

### DIFF
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -1062,10 +1062,8 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
 
                 pool_ids_that_had_balance.extend(balances.keys().copied());
 
-                if cur_height <= max_height {
-                    if !balances.is_empty() {
-                        height_map.insert(cur_height, balances);
-                    }
+                if cur_height <= max_height && !balances.is_empty() {
+                    height_map.insert(cur_height, balances);
                 }
 
                 if pool_ids.is_empty() {
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -52,9 +52,7 @@ use common::{
     Uint256,
 };
 use logging::log;
-use pos_accounting::{
-    PoSAccountingDB, PoSAccountingDelta, PoSAccountingDeltaRef, PoSAccountingView,
-};
+use pos_accounting::{PoSAccountingDB, PoSAccountingDelta, PoSAccountingView};
 use tx_verifier::transaction_verifier::TransactionVerifier;
 use utils::{debug_assert_or_log, ensure, log_error, tap_log::TapLog};
 use utxo::{UtxosCache, UtxosDB, UtxosStorageRead, UtxosView};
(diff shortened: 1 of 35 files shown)


chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -987,6 +994,107 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+            |disconnected_block_index, tx_verifier, _| -> Result<_, BlockError> {
+                let cur_height = disconnected_block_index
+                    .block_height()
+                    .prev_height()
+                    .expect("Genesis can't be disconnected");
+                assert!(cur_height >= min_height);
+
+                let pos_db = PoSAccountingDB::<_, TipStorageTag>::new(&self.db_tx);
+                let pos_delta =
+                    PoSAccountingDeltaRef::new(&pos_db, tx_verifier.accounting_delta_data());
```
azarovh, line 1053: I think you don't need this at all. `TransactionVerifier` already implements `PoSAccountingView`

ImplOfAnImpl: > I think you don't need this at all. TransactionVerifier already implements PoSAccountingView

Indeed, thanks. I've reverted adding that `PoSAccountingDeltaRef` stuff.

azarovh (approved):

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -987,6 +992,104 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         Ok(block_index)
     }
 
+    pub fn get_stake_pool_balances_at_heights(
+        &self,
+        pool_ids: &[PoolId],
+        min_height: BlockHeight,
+        max_height: BlockHeight,
+    ) -> Result<BTreeMap<BlockHeight, BTreeMap<PoolId, Amount>>, BlockError> {
```
TheQuantumPhysicist, line 1000: Maybe use a range?

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -987,6 +992,104 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+                    .block_height()
+                    .prev_height()
+                    .expect("Genesis can't be disconnected");
+                assert!(cur_height >= min_height);
+
+                let balances = Self::collect_pool_balances(pool_ids.iter(), &tx_verifier)?;
+
+                pool_ids.retain(|pool_id| {
+                    // We didn't see this pool having balance yet.
+                    pool_ids_that_had_balance.get(pool_id).is_none() ||
```
TheQuantumPhysicist, line 1053: `!pool_ids_that_had_balance.contains(pool_id)`

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -987,6 +992,104 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
...
+                    // We did see this pool having balance and it still does.
+                    balances.get(pool_id).is_some()
```
TheQuantumPhysicist, line 1055: balances.contains(pool_id)

TheQuantumPhysicist (approved):

chainstate/src/detail/chainstateref/mod.rs
ImplOfAnImpl: > Maybe use a range?

Do you insist on this?
Rust's ranges are focused around iteration but it turns out we can't iterate over a range of `BlockHeight`s (the iteration requires to implement `std::iter::Step` which is not stable yet). And outside of iteration, `Range` doesn't seem to enforce any relation between its boundaries, so it just gives some false sense of safety IMO. Also, the naming of fields/methods is confusing - both `Range` and `RangeInclusive` have `end` and you really have to look at the type to understand whether its inclusive or exclusive (I'd prefer to give specific names to inclusive range boundaries, like "min/max" or "first/last", but not "start/end").

azarovh (approved):

TheQuantumPhysicist (approved):

chainstate/src/detail/chainstateref/mod.rs
ImplOfAnImpl: > Maybe use a range?

[discussed] Leaving this as is.

iljakuklic (approved):

azarovh (approved):

PR #1746, opened 2024-05-07
Some minor blockprod-related improvements and cleanup

ImplOfAnImpl:
I ended up not needed some of these changes, however they still seem useful. Also, my later PR depends on them.

The logic hasn't changed here, just some code was moved, functions split etc.

Some "noisy" code (that calls chainstate and then just maps error types) was moved from `blockprod/src/detail/mod.rs` to a separate `utils.rs`, to make the former cleaner.

Also, I partially got rid of the `block_timestamp_seconds` atomic (at least, it's no longer propagated from `blockprod` to `consensus`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -192,6 +192,7 @@ qrcodegen = "1.8"
 quote = "1.0"
 rand = "0.8"
 rand_chacha = "0.3"
+rayon = "1.10"
 reedline = "0.31"
 ref-cast = "1.0"
 regex = "1.10"
@@ -261,6 +262,11 @@ panic = "abort" # prevent panic catching (mostly for the tokio runtime)
 [profile.release]
 panic = "abort" # prevent panic catching (mostly for the tokio runtime)
 overflow-checks = true
+
+# "Release" profile with debug info enabled.
+[profile.release-with-debug-info]
+inherits = "release"
+debug = true
 
 # "Release" profile with debug info and debug assertions enabled.
 [profile.release-with-debug]
--- a/blockprod/src/detail/job_manager/mod.rs
+++ b/blockprod/src/detail/job_manager/mod.rs
@@ -48,14 +48,10 @@ pub enum JobManagerError {
     FailedToStopJobs,
     #[error("Error sending stop job event")]
     FailedToSendStopJobEvent,
-    #[error("Error sending update last timestamp secconds used event")]
+    #[error("Error sending update last timestamp seconds used event")]
     FailedToSendUpdateLastTimestampSecondsUsed,
-    #[error("Error updating last timestamp secconds used")]
+    #[error("Error updating last timestamp seconds used")]
     FailedToUpdateLastTimestampSecondsUsed,
-}
-
-pub struct JobHandle {
-    cancel_sender: UnboundedSender<()>,
 }
 
 #[derive(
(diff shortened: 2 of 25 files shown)


Cargo.toml
```diff
@@ -192,6 +192,7 @@ qrcodegen = "1.8"
 quote = "1.0"
 rand = "0.8"
 rand_chacha = "0.3"
+rayon = "1.10"
```
ImplOfAnImpl, line 195: FYI: this wasn't supposed to end up in this PR, it's used in a later one.

blockprod/src/detail/job_manager/mod.rs
```diff
@@ -72,33 +68,21 @@ pub struct JobHandle {
...
+    /// staking twice with the same key.
     custom_id: CustomId,
-    /// The current tip, which will be the "previous block" for the
-    /// block that will be produced
-    current_tip_id: Id<GenBlock>,
-    // TODO: in proof of stake, we also add some identifier of the
-    // current key so that we don't stake twice from the same key.
-    // This is because in PoS, there could be penalties for creating
-    // multiple blocks by the same staker.
 }
```
ImplOfAnImpl, line 76: FYI: here I've basically moved this `current_tip_id` from the "key" part  of the jobs map to "value", named it `expected_tip_id` and made it optional (the job is not cancelled automatically when the tip changes if `expected_tip_id` is none).

(This is one of those changes that I ended up not making use of, but IMO job_manager looks a bit nicer this way).

blockprod/src/detail/utils.rs
```diff
@@ -0,0 +1,185 @@
...
+use crate::BlockProductionError;
+
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum PoSAccountingError {
+    #[error("Staker balance retrieval error: {0}")]
+    StakerBalanceRetrievalError(String),
+}
+
+pub fn get_pool_staker_balance(
+    chainstate: &dyn ChainstateInterface,
```
TheQuantumPhysicist, line 39: Is there a reason for using dyn all over this file, instead of a generic? Anything that derefs to dyn ChainstateInterface also implements ChainstateInterface, meaning that a reference or a box should be passed to a generic with trait bount ChainstateInterface with no issues.

TheQuantumPhysicist (approved):

blockprod/src/detail/utils.rs
```diff
@@ -0,0 +1,185 @@
...
+        })?;
+
+    let balances = balances
+        .iter()
+        .filter_map(|(height, balances)| {
+            balances.get(pool_id).map(|balance| (*height, balance.clone()))
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    Ok(balances)
```
iljakuklic, line 95: Would it make sense here to just return the iterator and let the client decide whether to collect it to a data structure, take just one element or what not. Since it produces elements lazily, you could even remove `max_height` and let the client decide when to stop iteration.

blockprod/src/detail/utils.rs
```diff
@@ -0,0 +1,185 @@
...
+        })?
+        .ok_or(BlockProductionError::PoolBalanceNotFound(*pool_id))?;
+
+    Ok(pool_balance)
+}
+
+pub fn get_pool_balances_at_heights(
+    chainstate: &dyn ChainstateInterface,
+    min_height: BlockHeight,
+    max_height: BlockHeight,
```
iljakuklic, line 76: I vaguely recall something similar was already discussed recently but it seems appropriate to me to use a range here. Conceptually, it would be somewhat similar to [BTreeMap::range](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range), used to select a range of keys from a map.

iljakuklic (approved):
Looks good

blockprod/src/detail/utils.rs
ImplOfAnImpl: > I vaguely recall something similar was already discussed recently but it seems appropriate to me to use a range here. Conceptually, it would be somewhat similar to BTreeMap::range, used to select a range of keys from a map.

Well, after I have made this function generic, I can use the `RangeBounds` trait here, just like `BTreeMap::range` does. But the corresponding function on the `ChainstateInterface` level can't be generic, and the "concrete" range types in Rust are not that great IMO, so I'll probably stick to the separate `min_height`/`max_height` there. And using a range just for this function, converting it to `min_height`/`max_height` on the fly, seems redundant to me. So, unless you insist, I'd leave it as is.

P.S. The discussion you are referring to is probably this one - https://github.com/mintlayer/mintlayer-core/pull/1742#discussion_r1589088425 - and I ended up not using a range in that case.

PR #1747, opened 2024-05-07
Implement timestamp_searcher in blockprod

ImplOfAnImpl:
A `timestamp_searcher` module has been added to `blockprod` that allows to obtain a list of timestamps where staking is/was possible given a range of block heights and a pool id.

 Also, the corresponding command was added to wallet-cli, so that this functionality can be invoked from the console.

P.S.
1) There are no tests. I plan to add some later.
2) If the block height range is large, the searching may take dozens of seconds (or maybe minutes) to complete, even though it's run in parallel via `rayon`. This will be a problem if the searching is started across an RPC boundary. In the current implementation, this is possible if the wallet-cli acts as a client to wallet-rpc-daemon. This is adresses in the following PR (#1748).


blockprod/src/detail/timestamp_searcher.rs
```diff
@@ -0,0 +1,487 @@
...
+/// of the next block.
+pub async fn collect_timestamp_search_data(
+    chainstate_handle: &ChainstateHandle,
+    chain_config: Arc<ChainConfig>,
+    secret_input_data: PoSTimestampSearchInputData,
+    min_height: BlockHeight,
+    max_height: Option<BlockHeight>,
+    seconds_to_check_for_height: u64,
+    check_all_timestamps_between_blocks: bool,
+) -> Result<TimestampSearchData, BlockProductionError> {
```
ImplOfAnImpl, line 233: 2discuss: the `check_all_timestamps_between_blocks` parameter seems redundant here, probably it should be removed and the code should behave as if it was always true?

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -680,6 +681,26 @@ pub enum WalletCommand {
...
+    /// of the next block.
+    #[clap(name = "node-find-timestamps-for-staking")]
+    #[clap(hide = true)]
+    FindTimestampsForStaking {
+        pool_id: String,
+        min_height: BlockHeight,
+        max_height: BlockHeight,
+        seconds_to_check_for_height: u64,
+        check_all_timestamps_between_blocks: YesNo,
+    },
```
ImplOfAnImpl, line 702: 2discuss: I'm also not sure about the cli interface; perhaps, `pool_id` shouldn't be here and the search should always be performed for all pools in all accounts? (or may be we need an additional command for that).

blockprod/src/detail/timestamp_searcher.rs
```diff
@@ -0,0 +1,487 @@
...
+impl TimestampSearchData {
+    pub fn new(
+        chainstate: &dyn ChainstateInterface,
+        chain_config: &ChainConfig,
+        secret_input_data: &PoSTimestampSearchInputData,
+        min_height: BlockHeight,
+        max_height: Option<BlockHeight>,
+        seconds_to_check_for_height: u64,
+        check_all_timestamps_between_blocks: bool,
+    ) -> Result<Self, BlockProductionError> {
```
TheQuantumPhysicist, line 81: Perhaps a separate error type for this is a good idea

blockprod/src/detail/timestamp_searcher.rs
TheQuantumPhysicist: This is an auxiliary functionality, so I'm OK with it not being perfect. It's only gonna be used by us, devs, mostly.

TheQuantumPhysicist (approved):
Looking OK. I asked Boris to review as well.

OBorce (approved):
looks good

blockprod/src/detail/timestamp_searcher.rs
ImplOfAnImpl: > Perhaps a separate error type for this is a good idea

Actually, initially I tried using a separate `TimestampSearcherError` type for this and I didn't like it.
The main issue I had with this approach is that it makes it harder to re-use utility functions that are common to the "normal" blockprod stuff and timestamp searcher:
1) The utility functions can't return `TimestampSearcherError`, because it's too specific.
2) If they return `BlockProductionError`, then `TimestampSearcherError` has to contain a `BlockProductionError` variant. On the other hand, `BlockProductionError` is an "umbrella" error-type for everything blockprod-related, so `BlockProductionInterface::collect_timestamp_search_data_e2e` should return it as well, so `BlockProductionError` has to contain a `TimestampSearcherError` variant.

Surely, this can be worked around, e.g. by introducing an intermediate error type, or by duplicating some utility functions (which are mostly trivial), but I'd personally leave it as is, unless you insist.

TheQuantumPhysicist: It's OK.

wallet/wallet-rpc-lib/src/rpc/mod.rs
```diff
@@ -221,6 +222,33 @@ impl<N: NodeInterface + Clone + Send + Sync + 'static> WalletRpc<N> {
             .await?
     }
 
+    pub async fn find_timestamps_for_staking(
```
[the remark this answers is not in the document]
ImplOfAnImpl: > This function also needs documentation

Functions in this `WalletRpc` struct don't seem to have much documentation. And since this function just calls `find_timestamps_for_staking` in wallet-controller, I decided to put the documentation there instead, "for consistency". Again, this is a copy-paste of the documentation for `WalletCommand::FindTimestampsForStaking`, so it's triplicated now.

azarovh (approved):

wallet/wallet-cli-commands/src/command_handler/mod.rs
```diff
@@ -709,6 +709,27 @@ where
                 Ok(ConsoleCommand::Print("Success".to_owned()))
             }
 
+            WalletCommand::FindTimestampsForStaking {
```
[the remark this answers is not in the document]
ImplOfAnImpl: > Why do we want this function in the wallet?

This function is a diagnostic tool that is supposed to be used by developers. The (hidden) command in the wallet is just for convenience.

TheQuantumPhysicist (approved):

PR #1748, opened 2024-05-07
Call blockprod::find_timestamps_for_staking directly from wallet cli

ImplOfAnImpl:
This addresses the problem from #1747 where the timestamp search RPC call from wallet-cli to wallet-rpc-daemon would timeout if the search takes a long time.

To solve the problem, the call to `blockprod::find_timestamps_for_staking` was moved from the wallet controller to wallet-cli itself. Since the call needs `PoSTimestampSearchInputData`, I had to add the `get_timestamp_search_input_data` function to wallet RPC that returns the data. And since the data contains the private vrf key, I had to use e2e encryption for it.

In the end, I'm not sure it's worth the added complexity, because the timeout problem only exists if wallet-cli connects to wallet-rpc-daemon. 
Given that `FindTimestampsForStaking` is a "niche" (and hidden) command, probably no one will ever use it in this scenario.

### DIFF
--- a/wallet/wallet-cli-commands/src/command_handler/mod.rs
+++ b/wallet/wallet-cli-commands/src/command_handler/mod.rs
@@ -15,7 +15,7 @@
 
 mod local_state;
 
-use std::{fmt::Write, str::FromStr};
+use std::{fmt::Write, str::FromStr, sync::Arc};
 
 use common::{
     address::Address,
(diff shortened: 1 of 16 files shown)


wallet/wallet-cli-commands/src/command_handler/mod.rs
```diff
@@ -716,17 +718,43 @@ where
...
+                let encrypted_input_data = wallet
+                    .get_timestamp_search_input_data(
+                        HexEncoded::new(our_public_key.clone()),
                         pool_id,
+                    )
+                    .await?;
+                let shared_secret = our_private_key.shared_secret(&callee_public_key);
+                let secret_input_data = shared_secret
+                    .decrypt_then_decode::<PoSTimestampSearchInputData>(&encrypted_input_data)?;
+
```
TheQuantumPhysicist, line 738: Can't we pack all this in one function or something, instead of doing it here inline?

TheQuantumPhysicist (commented):
I don't think this is a good idea. We should attempt to do it through tower first.

ImplOfAnImpl:
> Obviously, this isn't a priority. I'm OK with putting this in the backlog.

I've added an issue for this - #1749

ImplOfAnImpl:
Closing this for now. It can be re-visited after #1749 is implemented.

PR #1750, opened 2024-05-13
tx-verifier: Introduce Context types

iljakuklic:
These types hold information about the block and transaction being verified. The verification procedure interfaces with the context types rather than inspecting block / transaction directly.

Not quite happy with this yet, will evolve this further. Feedback welcome.


chainstate/tx-verifier/src/transaction_verifier/input_check.rs
```diff
@@ -0,0 +1,301 @@
...
+use crate::TransactionVerifierStorageRef;
+
+use super::{
+    error::ConnectTransactionError, signature_destination_getter::SignatureDestinationGetter,
+    TransactionSourceForConnect,
+};
+
+pub struct BlockVerificationContext<'a> {
+    chain_config: &'a ChainConfig,
+    destination_getter: SignatureDestinationGetter<'a>,
```
TheQuantumPhysicist, line 36: Note that after introducing scripting, this will probably be eliminated. This is because SignatureDestinationGetter assumes that there's exactly one signature to verify per input, while in reality, it can be more complex than just that.

TheQuantumPhysicist:
Done in #1762

PR #1751, opened 2024-05-15
Reenable tokens in simulation tests

azarovh:
- enable tokens issuance in `RandomTxMaker` used for simulation tests
- store fee utxo as a member of `RandomTxMaker` instead of passing it around (which is confusing and had a bug)
- do not collect Destination for input signing in `RandomTxMaker` but rather get them via `SignatureDestinationGetter` afterwards. This simplifies the code and also fixes a bug in it.
- fix `ApiServerInMemoryStorage` address balance tables. The problem with previous data type is that it didn't consider the fact that fetching latest height balance might contain not all currencies. Note that new type is also not perfect because the removal "above height" is now linear which should be ok for tests.

### DIFF
--- a/api-server/scanner-lib/src/blockchain_state/mod.rs
+++ b/api-server/scanner-lib/src/blockchain_state/mod.rs
@@ -252,7 +252,7 @@ async fn update_locked_amounts_for_current_block<T: ApiServerStorageWrite>(
                             db_tx,
                             address,
                             amount,
-                            CoinOrTokenId::Coin,
+                            CoinOrTokenId::TokenId(*token_id),
                             block_height,
                         )
                         .await;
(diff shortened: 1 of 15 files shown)


TheQuantumPhysicist (approved):

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -1075,12 +1075,25 @@ async fn update_tables_from_transaction_outputs<T: ApiServerStorageWrite>(
...
+                    CoinOrTokenId::TokenId(*token_id),
+                    block_height,
+                )
+                .await;
+
                 db_tx.set_nft_token_issuance(*token_id, block_height, *issuance.clone()).await?;
                 set_utxo(
                     outpoint,
                     output,
-                    Some(1),
```
OBorce, line 1083: I believe this needs to stay, otherwise when fetching the utxos it will not know how many decimals to use for formatting the amount.

OBorce: nope, seems it is all good

OBorce (approved):

PR #1753, opened 2024-05-16
port more changes from master into release 0.4.3

OBorce:


azarovh (approved):
Given that we already merged bunch of changes yesterday I'm ok with this PR. But imho in the future we should only port specific critical fixes to release branches. Otherwise it's  hard to understand the influence of the changes.

PR #1755, opened 2024-05-16
remove second syncing in gui wallet

OBorce:
- the second syncing was causing to send a message to the rpc wallet handler interrupting the loop that is already doing the syncing and staking


TheQuantumPhysicist (approved):
Let's confirm that all functionality is tested before merging

ImplOfAnImpl (approved):

azarovh (approved):

PR #1757, opened 2024-05-20
Blockprod cleanup

ImplOfAnImpl:
Some minor cleanup in `blockprod` that my next PR depends upon:
1) Some `blockprod` tests now use `PoS` instead of `IgnoreConsensus`.
2) The loop inside `produce_block_with_custom_id` was removed; now the function will exit if a "recoverable" `mempool` error occurs, relying on the fact that the caller (i.e. the wallet) calls `blockprod` repeatedly in a loop anyway. The `TipChanged` error is no longer needed because of this, so it and the corresponding test were removed.

Also, the functions `generate_consensus_data_and_reward` and `generate_finalize_block_data` were removed and their contents were put into `pull_consensus_data`, which now has a single "match" over the provided input data and the required consensus type.


TheQuantumPhysicist (approved):
Please don't merge before the release

PR #1758, opened 2024-05-20
Staking algorithm improvement

ImplOfAnImpl:
~Refactor blockprod, so that in PoS a block may be created after the timestamp for staking has already been found.
This is the basis for the next step, where we'll be determining the parent for the new block based on the found timestamp and not vice versa.~

~Unlike #1722, where PoS/PoW/IgnoreConsensus were being handled separately, this approach maintains roughly the same flow for all consensus types - the single function `produce_block_with_custom_id` always performs the 3 steps "pull data / solve block / finalize block" and each step has a `match` statememt over its input data.~

~Also, I've moved the `last_used_block_timestamps` map from `JobsContainer` to the `BlockProduction` struct itself and replaced the corresponding atomic with a channel.~

~P.S. I've moved some code from `consensus` to `blockprod`, but it was done somewhat inconsistently. E.g. the `stake` function is now in `blockprod`, but `StakeResult` is still in `consensus`. I plan to revisit this at the next step.~

I decided to push the "next step" work into the same branch, so now this PR is about staking improvement:

1) the new block can now be created over a non-tip parent, if the resulting chain trust is bigger than tip's chain trust;
2) new tip arrival doesn't reset the last timestamp used for staking; instead, we always start from the point where we stopped at the previous staking attempt;
3) we now don't look into the future; the max_timestamp for staking is always the current time.

Note:
1) Since we can't reorg the mempool in memory at this moment, the block will be created without transactions if its parent is not the current tip. This PR won't be merged until this is resolved.
2) Currently, the `wallet_decommission_request` functional test is failing intermittently; partially it's because it expects that new blocks are always created on top of the current tip, which now may not be true (so the test needs some refactoring). But even if it's refactored, it'll still fail because that last block won't have the required transactions (because of the note 1), so I'm postponing the fix for now.


blockprod/src/detail/block_solver.rs
```diff
@@ -0,0 +1,726 @@
...
+
+use super::{
+    utils::{
+        get_existing_gen_block_index, get_pool_balances_at_heights, get_sealed_epoch_randomness,
+        make_ancestor_getter, timestamp_add_secs,
+    },
+    BlockProduction, TxData,
+};
+
+impl BlockProduction {
```
TheQuantumPhysicist, line 67: I'm not a big fan of having a single struct's impls in multiple files... this indicates that some restructuring might be necessary with better encapsulation and isolated responsibilities.

PR #1759, opened 2024-05-21
Feature/api server token statistics

OBorce:
add endpoints for statistics for coins and tokens, returning
- ciruculating_amount the total available circulating amount
- preminted the total amount of coins that were preminted in the genesis block
- burned the total amount of coins/tokens that got burned or spent on issuance fees
- staked the total amount of coins currently being staked in pools or delegations

Added new endpoint GET token that returns a list of token ids with length and offset query parameters

### DIFF
--- a/api-server/api-server-common/src/storage/storage_api/mod.rs
+++ b/api-server/api-server-common/src/storage/storage_api/mod.rs
@@ -70,7 +70,7 @@ pub enum ApiServerStorageError {
 
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub enum CoinOrTokenStatistic {
-    TotalSupply,
+    CirculatingSupply,
     Staked,
     Burned,
     Preminted,
@@ -81,7 +81,7 @@ impl FromStr for CoinOrTokenStatistic {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let statistic = match s {
-            "TotalSupply" => Self::TotalSupply,
+            "CirculatingSupply" => Self::CirculatingSupply,
             "Staked" => Self::Staked,
             "Burned" => Self::Burned,
             "Preminted" => Self::Preminted,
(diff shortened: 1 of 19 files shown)


api-server/api-server-common/src/storage/storage_api/mod.rs
```diff
@@ -64,6 +68,47 @@ pub enum ApiServerStorageError {
     TimestampToHigh(BlockTimestamp),
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub enum CoinOrTokenStatistic {
+    TotalSupply,
+    Staked,
+    Burned,
+    Preminted,
```
TheQuantumPhysicist, line 76: You mean Unminted?

api-server/api-server-common/src/storage/storage_api/mod.rs
```diff
@@ -64,6 +68,47 @@ pub enum ApiServerStorageError {
     TimestampToHigh(BlockTimestamp),
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub enum CoinOrTokenStatistic {
+    TotalSupply,
```
azarovh, line 73: It seems like it should be `CirculatingSupply` to conform with chainstate terminology where TotalSupply is the limit for minting and Circulating is current amount minted/unminted. But I'm confused why this arm is used for coins in transfer outputs.

api-server/api-server-common/src/storage/storage_api/mod.rs
OBorce: Preminted are the coins from genesis

api-server/api-server-common/src/storage/storage_api/mod.rs
OBorce: the transfer output is probably the arm from block reward, those are from genesis.

PR #1762, opened 2024-05-22
tx-verifier: Introduce and integrate mintscript

iljakuklic:
Still a fair amount of stuff missing:

* [x] Block reward to script translation
* [x] Timelock-only checks (for mempool)
* [ ] Random cleanups
  * [ ] Simplify traits for script translation (e.g. timelock-only translation does not need to be under a trait)
  * [ ] Split script translation module
  * [ ] Split tx verifier input check module (and maybe rename)
* [ ] Improve error reporting
  * [ ] Create an input verification error, separate from ConnectTransactionError which provides a contextual information about which input failed to verify
  * [ ] Remove no longer needed arms from ConnectTransactionError
* [ ] Review which parts of code can be removed, some candidates:
  * [ ] TransactionSourceForConnect
  * [x] SignatureDestinationGetter
  * [x] `transaction_verifier::signature_check`
* [x] Testing (scripts are easier to test in isolation but it's not taken advantage of very much at the moment)
* [ ] Future work
  * [ ] Translate `Destination`s into script.
  * [ ] Review how utxo data are passed into the checker. Maybe use a UtxoView with cache instead of requiring Utxo to be included in `InputInfo`. Having utxo included in input info makes the implementation details of signature checking seep into the checker interfaces.


chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub fn new<UW: utxo::UtxosView>(
+        block_ctx: &'a BlockVerificationContext<'a, S, AW>,
+        utxo_view: &UW,
+        transactable: &'a T,
+    ) -> Result<Self, ConnectTransactionError> {
+        let inputs = transactable.inputs().unwrap_or_default();
+        let sigs = transactable.signatures().unwrap_or_default();
+
+        // TODO(PR): Should this be a proper check rather than an assertion? Is it correct?
+        assert_eq!(inputs.len(), sigs.len());
```
TheQuantumPhysicist, line 104: I don't mind keeping this here. There's a check upon decoding a SignedTransaction that these are equal.

Cargo.toml
```diff
@@ -75,7 +76,7 @@ members = [
   "wallet/wallet-rpc-lib",              # Wallet RPC definitions library.
   "wallet/wallet-test-node",            # Node for wallet testing as a library.
   "wasm-wrappers",                      # WASM wrappers for various components.
-  "wasm-wrappers/wasm-doc-gen",         # WASM wrappers documentation generator.
+  "wasm-wrappers/wasm-doc-gen", "mintscript",         # WASM wrappers documentation generator.
```
TheQuantumPhysicist, line 79: Should be removed

mintscript/src/script/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /// Number of conditions that need to be satisfied in order for the threshold to be valid.
+    pub fn required(&self) -> usize {
+        self.required
+    }
+
+    /// Number of conditions that need to be dissatisfied in order for the threshold to be valid.
+    pub fn required_dissat(&self) -> usize {
+        self.conditions().len() - self.required()
```
TheQuantumPhysicist, line 90: Shouldn't this be checked or saturated?

mintscript/src/script/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        Ok(satisfied)
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+enum WitnessScriptInner {
+    Threshold(Threshold),
+    Signature(Destination, InputWitness),
+    Timelock(OutputTimeLock),
+}
```
TheQuantumPhysicist, line 130: I wonder whether we should add htlc now so that we ensure no design issues will appear

mintscript/src/translate.rs
```diff
@@ -0,0 +1,161 @@
...
+        match self {
+            InputInfo::Utxo { outpoint: _, utxo } => Some(utxo.output()),
+            InputInfo::Account { .. } | InputInfo::AccountCommand { .. } => None,
+        }
+    }
+}
+
+/// Provides information necessary to translate an input to a script.
+pub trait TranslationContext {
+    type Accounting: PoSAccountingView<Error = pos_accounting::Error>;
```
TheQuantumPhysicist, line 77: I would call this PoSAccounting, not just Accounting

azarovh (commented):
looks good to me so far

common/src/chain/transaction/signature/inputsig/mod.rs
```diff
@@ -30,3 +30,12 @@ pub enum InputWitness {
...
 }
+
+impl InputWitness {
+    pub fn as_standard_signature(&self) -> Option<&StandardInputSignature> {
+        match self {
+            InputWitness::NoSignature(_) => None,
+            InputWitness::Standard(sig) => Some(sig),
+        }
+    }
+}
```
TheQuantumPhysicist, line 41: I'm not sure how beneficial this is... I did do this same thing and undid it. The problem here is that NoSignature is also something that needs some level of verification.

In my branch, I added some signature tests, please make sure to cherry pick these commits too.

mintscript/src/checker/timelock.rs
```diff
@@ -0,0 +1,93 @@
...
+use common::{
+    chain::{block::timestamp::BlockTimestamp, timelock::OutputTimeLock},
+    primitives::{BlockDistance, BlockHeight},
+};
+
+pub trait TimelockChecker<C> {
+    type Error: std::error::Error;
+
+    /// Check timelock
+    fn check_timelock(&mut self, ctx: &mut C, lock: &OutputTimeLock) -> Result<(), Self::Error>;
```
TheQuantumPhysicist, line 25: Why does the context need to be mutable?

mintscript/src/checker/signature.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Signature checker that verifies signatures on basis of data provided by context.
+pub struct StandardSignatureChecker;
+
+impl<C: SignatureContext> SignatureChecker<C> for StandardSignatureChecker {
+    type Error = DestinationSigError;
+
+    fn check_signature(
+        &mut self,
+        ctx: &mut C,
```
TheQuantumPhysicist, line 73: Same here. Why mutable?

mintscript/src/checker/timelock.rs
iljakuklic: Strictly speaking it does not. The verification can be (and currently is) implemented without mutability. However, I was thinking of the following possible use cases:

* You can accumulate timelocks / signatures in a vec inside the context instead of checking right away. Already done in tests for checking iteration order.
* There can be a context adaptor that caches some information which requires updating some internal data structures as we go.
* Signature batching: On encountering a signature, just update some internal accumulator. When we're done with traversing input / transaction / block / whatever, check the accumulator.

All of the above could be achieved with interior mutability, not sure if that's better.

common/src/chain/transaction/signature/inputsig/mod.rs
iljakuklic: > I'm not sure how beneficial this is... I did do this same thing and undid it.

Apparently it's not even used anymore. Will remove.

mintscript/src/checker/timelock.rs
TheQuantumPhysicist: If anything is only for tests, then we definitely can use a RefCell/Mutex for that. 

About the other ones, this is actually a scary trade-off, and I don't know which is better in an absolute fashion. Better performance or better security. I would prefer the latter, and then worry about performance later. 

An example where this was done well (I hope so, at least) is in the transaction verifier. We isolated mutability from chainstate by making the base reference of TransactionVerifierStorageRef immutable, but then we have TransactionVerifierStorageMut to flush the changes when we're done, with a hiararchy that that can test all possible changes from one level upon the one on top of it. I'm sure there's a way to make the transaction verifier more performant by combining both, but I'm also sure we're gonna lose a dozen of security guarantees by doing that.

In short: I would prefer keeping this immutable, even if it meant we can lose some performance. Premature optimization isn't something I wanna worry about now, since the cost seems high. Usually I'm OK with it if it doesn't brick our plans.

mintscript/src/translate.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    #[error("Delegation {0} does not exist")]
+    DelegationNotFound(DelegationId),
+
+    #[error("Token with id {0} does not exist")]
+    TokenNotFound(TokenId),
+}
+
+/// Contextual information about given input
+pub enum InputInfo<'a> {
```
TheQuantumPhysicist, line 56: I would call this `InputStorageSourceType` or similar... I'm trying to avoid generic words like "info"

mintscript/src/checker/timelock.rs
iljakuklic: The motivation here is not performance but communicating correctly the semantics. In case something like signature batching is implemented, it is better to correctly communicate that something may be updated in types. That way, we know you are supposed to have a unique reference to it. If this is made immutable but mutation happens under the hood, someone in the future may take advantage of the fact that you can have multiple active references to the object. That will not result just in performance degradation but also, and more importantly, potentially *incorrect code*.

You are right that some (currently all non-test) implementations do not mutate any state and it is not ideal to expose them under a mutable ref in this case. We could change that to `&self` for now, adjusting tests to use interior mutability. However I am against keeping it as a shared ref if a mutating implementation comes along. Also, the current implementation partially mitigates the fact it's passed along mutably by having most of its members under a shared ref (e.g. the transaction is stored by shared reference), so most of the data cannot be mutated anyway.

Even with signature batching (for example), a big chunk of the context would still be immutable. It makes sense to split the context into mutable and immutable parts. I see two feasible ways of going about it:

1. Take the context as a shared ref and introduce an associated type `MyCntext::State` for the mutable part of the state.  The state part is passed along under a mutable ref, the context itself immutably. Alternatively, the state could be passed in by value and returned, to make it fell more functional, but I don't see this making much difference. This introduces more complexity in traits.
2. Leave the context ref mutable and adapt the following pattern for implementing contexts:
   ```rust
   struct MyFancyContext {
     immutable_part: &MyFancyContextRef,  // under a shared reference, does not mutate
     mutable_part: MyFancyContextMut, // owned, can freely change this
   }
   ```

Having this as a shared ref with interior mutability in presence of signature batching (or something else that updates some state in an observable way) is imho a wrong way to go about it.

mintscript/src/translate.rs
iljakuklic: Not sure, it's contextual information about the input, i.e. input paired up with some extra things needed for verification such as utxo. I agree that the name is not great but I don't see you suggestion improving it, it's not a `*SourceType` in any way as far as I can see.

mintscript/src/checker/signature.rs
```diff
@@ -0,0 +1,93 @@
...
+        let chain_config = ctx.chain_config();
+
+        // Note: The verify_signature function below looks up the signature in the transaction
+        // itself. In the future, the logic to find signature may not be so easy, so the signature
+        // from the script should be taken and passed to the signature verification code. This
+        // assertion should then go away. This goes hand in hand with turning Destinations, not
+        // just outputs/input pairs into script.
+        assert_eq!(
+            tx.signatures().unwrap_or(&[]).get(input_num),
+            Some(signature),
```
azarovh, line 88: and_then?

azarovh (approved):
I'd appreciate an extended comment somewhere in the lib file with general idea for the crate and potential future vision. Something to cover that mintscript is internal at the moment and was introduced with the purpose of this and that, potentially it could be stored on chain but consider this and that. Does it make sense?

mintscript/src/translate.rs
TheQuantumPhysicist: Can anyone suggest a better name?

mintscript/src/translate.rs
```diff
@@ -0,0 +1,231 @@
...
+use crate::WitnessScript;
+
+/// An error that can happen during translation of an input to a script
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum TranslationError {
+    #[error("Attempt to spend an unspendable output")]
+    Unspendable,
+
+    #[error("Illegal account spend")]
+    IllegalAccountSpend,
```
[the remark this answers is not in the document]
iljakuklic: Good point about the account ID.

Regarding the source ID, my vision is to attach input index to the script errors at the tx verifier level.

Could be done using a type like this:

```
struct LocalizedScriptError {
    input_num: usize,
    error: ScriptError,
}
```

That way, every script error will carry information about which input it belongs to.

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+    storage: &S,
+    tip: Id<GenBlock>,
+    spending_height: BlockHeight,
+    spending_time: BlockTimestamp,
+) -> Result<(), ConnectTransactionError>
+where
+    T: Transactable,
+    mintscript::translate::TimelockOnly: for<'a> TranslateInput<PerInputData<'a>>,
+    S: TransactionVerifierStorageRef,
+    UW: utxo::UtxosView,
```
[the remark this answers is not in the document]
iljakuklic: Will change it to `UV`, `AV`, `TV` and such. The problems is with tokens view. Calling it just `T` conflicts with another parameter which is already called `T` so a different name is needed.

mintscript/src/translate.rs
```diff
@@ -0,0 +1,231 @@
...
+                TxOutput::CreateStakePool(_pool_id, pool_data) => {
+                    Ok(checksig(pool_data.decommission_key()))
+                }
+                TxOutput::ProduceBlockFromStake(_dest, pool_id) => {
+                    let pool_data = pos_accounting
+                        .get_pool_data(*pool_id)?
+                        .ok_or(TranslationError::PoolNotFound(*pool_id))?;
+                    Ok(checksig(pool_data.decommission_destination()))
+                }
+                TxOutput::DelegateStaking(_amount, delegation_id) => {
```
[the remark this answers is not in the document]
iljakuklic: > Also I'd place a TODO to remove SignatureDestinationGetter

So far, it's a checkbox at the top of this PR. Will be turned into a TODO/issue or resolved.

ImplOfAnImpl (approved):

mintscript/src/script/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Number of conditions that need to be satisfied in order for the threshold to be valid.
+    pub fn required(&self) -> usize {
+        self.required
+    }
+
+    /// Number of conditions that need to be dissatisfied in order for the threshold to be valid.
+    pub fn required_dissat(&self) -> usize {
+        let num_conds = self.conditions().len();
+        num_conds.checked_sub(self.required()).expect("checked during construction")
+    }
```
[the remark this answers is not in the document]
ImplOfAnImpl: > I mean, if no conditions are unsatisfied, the whole `Threshold` check will still succeed, right?

Though now I can see that `Threshold` requires the exact number of conditions to be satisfied. In that case `Threshold` itself is probably a misnomer.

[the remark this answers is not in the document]
iljakuklic: > Probably `required_dissat` is a misnomer, because it's the number of conditions that are allowed to fail. I mean, if no conditions are unsatisfied, the whole `Threshold` check will still succeed, right?

No, for malleability concerns, the user must pick the exactly N coditions, and satisfy them. Even if he can in principle satisfy more. This is in line with the usual multisig implementation in Taproot using the `OP_CHECKSIGADD` and `OP_EQUALSVERYFY` pattern.

> P.S. also, "unsatisfied" is IMO a better word here.

This is inspired by the terminology form miniscript and refers to the conscious choice of explicily marking a condition as not satisfied.

mintscript/src/checker/timelock.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    fn check_timelock(&mut self, ctx: &mut C, lock: &OutputTimeLock) -> Result<(), Self::Error> {
+        use TimelockError as E;
+
+        match lock {
+            OutputTimeLock::UntilHeight(required) => {
+                check_at_least(ctx.spending_height(), *required, E::HeightLocked)
+            }
+            OutputTimeLock::UntilTime(required) => {
+                check_at_least(ctx.spending_time(), *required, E::TimestampLocked)
```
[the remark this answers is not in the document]
iljakuklic: Just did not want to repeat the spending time/height and required time/height in the condition and the error value, so I abstracted it out. Observe that the error takes on these values.

do_checks.sh
```diff
@@ -8,7 +8,7 @@ PYTHON=$(which python || which python3)
 cargo fmt --check -- --config newline_style=Unix
 
 # Install cargo deny first with: cargo install cargo-deny
-cargo deny check --hide-inclusion-graph
+#cargo deny check --hide-inclusion-graph
```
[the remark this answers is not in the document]
iljakuklic: Good catch, thanks for pointing this out

mintscript/src/translate.rs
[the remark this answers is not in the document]
iljakuklic: > this is not spendable. I see it's taken from SignatureDestinationGetter but that's actually a bug there (thoughn the code is effectively unreachable because no utxo is available).

Ok. Are you reasonably sure this will not cause forks? This is consensus code so I am a bit paranoid here.

chainstate/tx-verifier/src/transaction_verifier/tx_source.rs
```diff
@@ -112,6 +112,18 @@ impl<'a> TransactionSourceForConnect<'a> {
...
+            TransactionSourceForConnect::Chain { new_block_index } => {
+                (*new_block_index.block_id()).into()
+            }
+            TransactionSourceForConnect::Mempool {
+                current_best,
+                effective_height: _,
+            } => current_best.block_id(),
+        }
+    }
+
```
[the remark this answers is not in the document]
iljakuklic: removed

mintscript/src/translate.rs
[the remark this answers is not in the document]
iljakuklic: > I was hoping we could ditch SignatureDestinationGetter before merging this PR...

Tried to do it a while back but it's still used in tests to generate signatures. Can move it into a test module. In principle, it could be gutted out and the implementation replaced with something that translates the input to a script and extracts destinations.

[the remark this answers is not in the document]
azarovh: I am because such outputs are not included into utxo set see: https://github.com/mintlayer/mintlayer-core/blob/870434bd92f3c94d82f304efdc06ce0602a77d60/utxo/src/cache.rs#L495 so they cannot be spent and thus verified.

I don't see the test for that so you can add it with the patch:
[spend_delegate_staking_output_test.patch](https://github.com/user-attachments/files/15742718/spend_delegate_staking_output_test.patch)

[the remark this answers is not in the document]
TheQuantumPhysicist: > Can move it into a test module

I'm all in for that.

mintscript/src/translate.rs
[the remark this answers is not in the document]
TheQuantumPhysicist: But you see, if the translation fails for any reason, we won't be able to track how this happened.

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+pub fn verify_timelocks<T, S, UV>(
+    transaction: &T,
+    chain_config: &ChainConfig,
+    utxos_view: &UV,
+    storage: &S,
+    tip: Id<GenBlock>,
+    spending_height: BlockHeight,
+    spending_time: BlockTimestamp,
+) -> Result<(), ConnectTransactionError>
+where
```
TheQuantumPhysicist, line 434: This doesn't feel to be in the correct place. The transaction verifier is supposed to be fully verifying transactions. Given that timelock constraints are only needed in the mempool, perhaps we should move this to the mempool. The reason why I'm thinking about this is that I don't want there to be any impression that the verification process is fractured.

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+                Err(ConnectTransactionError::TimelockedAccount)
+            }
+        }
+    }
+}
+
+struct VerifyContextFull<'a, T, S> {
+    // Here we need more information about the transaction
+    transaction: &'a T,
+    spent_utxos: Vec<Option<&'a TxOutput>>,
```
TheQuantumPhysicist, line 303: I would question whether we need these in here. After all, when we do the verification, we are including the utxo source (InputInfo) anyway. Maybe I'm missing something, I'm now sure.

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        storage,
+        tx_source,
+        spending_time,
+        &core_ctx,
+    );
+    let ctx = VerifyContextFull::new(transaction, &tl_ctx);
+
+    for (n, inp) in core_ctx.inputs_iter() {
+        let script =
+            TranslationContextFull::new(pos_accounting, tokens_accounting, inp).to_script::<T>()?;
```
TheQuantumPhysicist, line 417: With regards to my other comment about utxos, isn't better to use the input info here and retrieve the needed utxos here instead of part here for accounting and part there for utxos?

mintscript/src/translate.rs
```diff
@@ -0,0 +1,231 @@
...
+    fn witness(&self) -> &InputWitness;
+}
+
+/// Provides information necessary to translate an input to a script.
+pub trait SignatureInfoProvider: InputInfoProvider {
+    type PoSAccounting: PoSAccountingView<Error = pos_accounting::Error>;
+    type Tokens: TokensAccountingView<Error = tokens_accounting::Error>;
+
+    fn pos_accounting(&self) -> &Self::PoSAccounting;
+    fn tokens(&self) -> &Self::Tokens;
```
TheQuantumPhysicist, line 89: Can't we add UtxosView here?

mintscript/src/tests/translate/snap.translate.tlock.transfer_01.txt
```diff
@@ -0,0 +1,26 @@
...
+            },
+        ),
+        Standard(
+            StandardInputSignature {
+                sighash_type: SigHashType(
+                    1,
+                ),
+                raw_signature: [
+                    81,
+                    81,
```
TheQuantumPhysicist, line 21: Why does a single byte of 0x51 translate into two 81s, where 0x51=81?

TheQuantumPhysicist: Note that this applies to all signatures.

mintscript/src/tests/translate/snap.translate.tlock.transfer_00.txt
```diff
@@ -0,0 +1,18 @@
...
+                        ),
+                    },
+                ),
+            },
+        ),
+        NoSignature(
+            None,
+        ),
+    ),
+)
```
TheQuantumPhysicist, line 18: I don't know if this is intended, but all tlock tests exactly match the txn transactions... so timelocks are not translated to scripts... this is probably not intended, right?

TheQuantumPhysicist (commented):
OK. I'm done with this round. I'll take another round later. I'm currently syncing the node with this version.

TheQuantumPhysicist (commented):
Sync successful. Let's discuss these details on Monday. Please try to fix whatever you can of these comments.

mintscript/src/tests/translate/snap.translate.tlock.transfer_00.txt
iljakuklic: > I don't know if this is intended, but all tlock tests exactly match the txn transactions... so timelocks are not translated to scripts... this is probably not intended, right?

Tests were mistakenly using the transaction translation modes for timelock-only and reward translation in tests. Fixed now.

mintscript/src/tests/translate/snap.translate.tlock.transfer_01.txt
iljakuklic: The function that creates the fake witness data drops (rather arbitrarily) two identical bytes there:
```rust
InputWitness::Standard(StandardInputSignature::new(sht, vec![byte; 2]))
```

mintscript/src/tests/translate/snap.translate.tlock.transfer_03.txt
```diff
@@ -0,0 +1,8 @@
+WitnessScript(
+    Threshold(
+        Threshold {
+            required: 0,
+            conditions: [],
+        },
+    ),
+)
```
TheQuantumPhysicist, line 8: All tlock transfer seem to have no signature verification

iljakuklic: That is correct. The timelock-only mode is intended to only have timelocks.

mintscript/src/translate.rs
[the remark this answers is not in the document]
iljakuklic: > > Can move it into a test module
> 
> I'm all in for that.

This is done

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
iljakuklic: > I would question whether we need these in here. After all, when we do the verification, we are including the utxo source (InputInfo) anyway. Maybe I'm missing something, I'm not sure.

The signature checking function expects the list of inputs to be committed to and it takes it in this format.

mintscript/src/translate.rs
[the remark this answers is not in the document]
iljakuklic: > > Are you reasonably sure this will not cause forks?
> 
> I am because such outputs are not included into utxo set see:
> 
> https://github.com/mintlayer/mintlayer-core/blob/870434bd92f3c94d82f304efdc06ce0602a77d60/utxo/src/cache.rs#L495
> 
> so they cannot be spent and thus verified.
> I don't see the test for that so you can add it with the patch: [spend_delegate_staking_output_test.patch](https://github.com/user-attachments/files/15742718/spend_delegate_staking_output_test.patch)

Changed this and included the test. Thanks.

mintscript/src/translate.rs
TheQuantumPhysicist: So we had a discussion about this yesterday, and it seems that there's some abstraction leakage and some broken symmetry. But the problem is that UtxosView is needed earlier to be able to calculate signatures.

TheQuantumPhysicist (approved):
I guess this is looking closer to what we would merge. I'm happy to merge it and fix the remaining issues in a subsequent PR.

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
TheQuantumPhysicist: So, for simplicity, we agreed to add docs explaining this.

mintscript/src/tests/translate/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+    InputWitness::Standard(StandardInputSignature::new(sht, vec![byte; 2]))
+}
+
+fn deleg0() -> (DelegationId, DelegationData) {
+    let data = DelegationData::new(fake_id(0x57), dest_pk(101));
+    (fake_id(0x75), data)
+}
+
+fn pool0_decom() -> Destination {
+    dest_pk(1337)
```
azarovh, line 160: could you use hex everywhere? hard to track this way

azarovh (approved):

chainstate/test-suite/src/tests/signature_tests.rs
```diff
@@ -87,26 +88,61 @@ fn signed_tx(#[case] seed: Seed) {
...
+            // The second transaction has the signed input.
+            let tx_2 = {
+                SignedTransaction::new(tx.clone(), vec![InputWitness::NoSignature(None)])
+                    .expect("invalid witness count")
+            };
+
+            tf.make_block_builder()
+                .with_transactions(vec![tx_1.clone(), tx_2])
+                .build_and_process(&mut rng)
+                .unwrap_err();
```
azarovh, line 102: `assert_eq` instead to check errors explicitly. same in other places

azarovh: will do it in my branch with tests

PR #1763, opened 2024-05-28
Staking test

ImplOfAnImpl:
A `docker compose` project and an init script that allow to start several `regtest` nodes locally to observe how staking works.

### DIFF
--- a/utils/src/default_data_dir.rs
+++ b/utils/src/default_data_dir.rs
@@ -52,25 +52,32 @@ pub enum PrepareDataDirError {
 
 /// Prepare data directory for the node.
 /// Two possibilities:
-/// 1. If no data directory is specified, use the default data directory provided by default_data_dir_getter;
-///    it doesn't have to exist. It will be automatically created.
-/// 2. If a custom data directory is specified, it MUST exist. Otherwise, an error is returned.
+/// 1. If no data directory is specified, use the default data directory provided by default_data_dir_getter.
+///    By default, the directory doesn't have to exist; it will be automatically created if it doesn't.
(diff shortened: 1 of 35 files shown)


utils/src/default_data_dir.rs
```diff
@@ -52,25 +52,32 @@ pub enum PrepareDataDirError {
...
-/// 2. If a custom data directory is specified, it MUST exist. Otherwise, an error is returned.
+/// 1. If no data directory is specified, use the default data directory provided by default_data_dir_getter.
+///    By default, the directory doesn't have to exist; it will be automatically created if it doesn't.
+/// 2. If a custom data directory is specified, it's expected to exist by default. An error is returned if it doesn't.
+///
+/// Additionally, `create_data_dir_if_missing` allows to override the default behavior.
 pub fn prepare_data_dir<F: Fn() -> PathBuf>(
     default_data_dir_getter: F,
     datadir_path_opt: &Option<PathBuf>,
+    create_data_dir_if_missing: Option<bool>,
```
TheQuantumPhysicist, line 63: Please note that the default policy for creating the data dir shouldn't change. It should be:

- If the default data dir is used, then it's created automatically
- If a custom data dir is chosen, it must already exist.

I hope it was done correctly before.

TheQuantumPhysicist (approved):
Please don't merge before the release

utils/src/default_data_dir.rs
ImplOfAnImpl: > Please note that the default policy for creating the data dir shouldn't change.

The default policy hasn't changed.

PR #1768, opened 2024-05-31
Store leaf blocks in chainstate

ImplOfAnImpl:
Chainstate db now maintains a set of "leaf" block ids, which are blocks without children. This is used to implement the functions `get_block_tree_top_by_height`/`get_block_tree_top_by_timestamp` in a more efficient way.

Also, it turned out that `get_block_id_tree_as_list`, which is used for bootstrapping, could return ids of non-persisted blocks. So, if such a block was encountered, bootstrapping would fail. This is also fixed.

### DIFF
--- a/chainstate/src/detail/chainstateref/mod.rs
+++ b/chainstate/src/detail/chainstateref/mod.rs
@@ -187,8 +187,11 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
     }
 
     #[log_error]
-    pub fn get_leaf_block_ids(&self) -> Result<BTreeSet<Id<Block>>, PropertyQueryError> {
-        Ok(self.db_tx.get_leaf_block_ids()?)
+    pub fn get_leaf_block_ids(
+        &self,
+        min_height: BlockHeight,
+    ) -> Result<BTreeSet<Id<Block>>, PropertyQueryError> {
+        Ok(self.db_tx.get_leaf_block_ids(min_height)?)
     }
 
     #[log_error]
@@ -1011,8 +1014,10 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         min_height: BlockHeight,
         block_validity: BlockValidity,
     ) -> Result<InMemoryBlockTrees, PropertyQueryError> {
+        let leaf_block_ids = self.get_leaf_block_ids(min_height)?;
         let result = in_memory_block_tree::get_block_tree_top(
             self,
+            leaf_block_ids.iter(),
             |block_index| block_index.block_height() >= min_height,
             block_validity,
         )?;
@@ -1039,8 +1044,11 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
         min_timestamp: BlockTimestamp,
         block_validity: BlockValidity,
     ) -> Result<InMemoryBlockTrees, PropertyQueryError> {
+        let min_height_with_allowed_reorg = self.get_min_height_with_allowed_reorg()?;
+        let leaf_block_ids = self.get_leaf_block_ids(min_height_with_allowed_reorg)?;
         let result = in_memory_block_tree::get_block_tree_top(
             self,
+            leaf_block_ids.iter(),
             |block_index| block_index.block_timestamp() >= min_timestamp,
             block_validity,
         )?;
@@ -1451,12 +1459,15 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
         }
         self.db_tx.set_block_index(block_index)?;
 
-        self.db_tx.mark_as_leaf(block_index.block_id(), true)?;
+        self.db_tx.mark_as_leaf(block_index, true)?;
 
         if let Some(non_genesis_parent) =
             block_index.prev_block_id().classify(self.chain_config).chain_block_id()
         {
-            self.db_tx.mark_as_leaf(&non_genesis_parent, false)?;
+            let parent_index = self
+                .get_existing_block_index(&non_genesis_parent)
+                .map_err(BlockError::PropertyQueryError)?;
+            self.db_tx.mark_as_leaf(&parent_index, false)?;
         }
 
         Ok(())
@@ -1469,7 +1480,12 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
         &mut self,
         tree: InMemoryBlockTreeRef<'b>,
     ) -> Result<(), BlockError> {
-        let old_leaf_block_ids = self.db_tx.get_leaf_block_ids()?;
+        let old_leaf_block_ids = self.db_tx.get_leaf_block_ids(
+            tree.root_block_index()
+                .block_height()
+                .prev_height()
+                .expect("Non-genesis can't have zero height"),
+        )?;
         let mut remaining_leaf_block_ids = old_leaf_block_ids.clone();
 
         for block_index in tree.iter_all_block_indices() {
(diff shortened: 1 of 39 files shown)


chainstate/src/detail/in_memory_block_tree.rs
```diff
@@ -0,0 +1,446 @@
...
+            handler(InMemoryBlockTreeRef::new(self.arena, node_id))
+        })
+    }
+}
+
+/// A single `BlockIndex` tree.
+pub struct InMemoryBlockTree {
+    arena: Arena<BlockIndex>,
+    root_id: NodeId,
+}
```
TheQuantumPhysicist, line 179: What's the point of this abstraction if it doesn't provide any guarantees? Even the constructor doesn't ensure that the root is really, the root. That new() function should be called new_unchecked, or something similar.

chainstate/src/detail/in_memory_block_tree.rs
```diff
@@ -0,0 +1,446 @@
...
+    fn new(arena: Arena<BlockIndex>, root_id: NodeId) -> InMemoryBlockTree {
+        Self { arena, root_id }
+    }
+
+    #[allow(unused)]
+    pub fn root_id(&self) -> NodeId {
+        self.root_id
+    }
+
+    pub fn as_ref(&self) -> InMemoryBlockTreeRef<'_> {
```
TheQuantumPhysicist, line 191: Why not use the AsRef trait?

chainstate/src/detail/in_memory_block_tree.rs
```diff
@@ -0,0 +1,446 @@
...
+    primitives::{BlockHeight, Id},
+};
+use indextree::{Arena, Node, NodeId};
+use utils::{debug_panic_or_log, log_error};
+
+use crate::{detail::chainstateref::block_validity_matches, TransactionVerificationStrategy};
+
+use super::chainstateref::{BlockValidity, ChainstateRef};
+
+/// Zero or more `BlockIndex` trees sharing the same arena.
```
TheQuantumPhysicist, line 31: Please explain why this is needed and what problem it solves

TheQuantumPhysicist: To be clear, intuitively, multiple trees should go into a `BTreeMap<root,arena>`. Would be great to explain unintuitive things.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -1330,39 +1452,123 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
...
+            }
+        }
+
         Ok(())
     }
 
+    /// Check if the block specified by `potential_parent_id` is a (possibly indirect) parent of
+    /// the block specified by `child_index`.
+    #[log_error]
+    fn is_parent_of(
```
TheQuantumPhysicist, line 1544: Question, why not use the function `last_common_ancestor` for the implementation here? 

If not... Note that this looping can be optimized by using "some_ancestor" in BlockIndex. It was used previously.

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -1330,39 +1452,123 @@ impl<'a, S: BlockchainStorageWrite, V: TransactionVerificationStrategy> Chainsta
...
+        Ok(())
     }
 
-    /// Delete the block index for the specified block id.
-    /// Panic if the block is marked as persisted.
+    /// Delete the specified block indices.
+    /// Panic if one of the blocks is marked as persisted.
     #[log_error]
-    pub fn del_block_index_of_non_persisted_block(
+    pub fn del_block_indices_of_non_persisted_block_tree<'b>(
```
TheQuantumPhysicist, line 1479: The complexity in this function is terrifying. Do we really need to do these deletions now? Or do we want to wait until we implement that circular buffer solution that doesn't persist them anyway?

common/src/primitives/height.rs
```diff
@@ -18,7 +18,7 @@ use std::ops::{Add, Sub};
 
 use serialization::{Decode, Encode};
 
-type HeightIntType = u64;
+pub type BlockHeightIntType = u64;
```
TheQuantumPhysicist, line 21: I would prefer to keep this private. I think the better way to do this is to keep this private, and in the database, use u64, and static_assert on it that it's u64 if necessary, and not another type.

chainstate/storage/src/lib.rs
```diff
@@ -198,6 +201,9 @@ pub trait BlockchainStorageWrite:
         block_id: &Id<GenBlock>,
     ) -> Result<()>;
 
+    /// Mark the specified block as a leaf or non-leaf, depending on the `is_leaf` parameter.
+    fn mark_as_leaf(&mut self, block_index: &BlockIndex, is_leaf: bool) -> Result<()>;
+
```
TheQuantumPhysicist, line 206: You forgot to increment `ChainstateStorageVersion` to enforce a resync.

TheQuantumPhysicist (commented):
First round done. Please look into the comments.

chainstate/src/detail/in_memory_block_tree.rs
ImplOfAnImpl: > What's the point of this abstraction if it doesn't provide any guarantees? Even the constructor doesn't ensure that the root is really, the root. That new() function should be called new_unchecked, or something similar.

I've added some checks

chainstate/src/detail/in_memory_block_tree.rs
ImplOfAnImpl: > Why not use the AsRef trait?

AsRef would require returning a literal reference to some type. Here I need to return `InMemoryBlockTreeRef` by value.

chainstate/src/detail/chainstateref/mod.rs
ImplOfAnImpl: > The complexity in this function is terrifying.

Just in case, this function was only called from `reset_block_failure_flags`, which can only be manually called via rpc. And the number of blocks to be traversed was supposed to be small.
Anyway, I've reworked `reset_block_failure_flags`, so searching for new leaves doesn't have quadratic complexity anymore.

common/src/primitives/height.rs
ImplOfAnImpl: > I would prefer to keep this private.

Why?

Btw, the similar `BlockTimestampInternalType` is public.

TheQuantumPhysicist (approved):
Besides the checks on the abstraction, I think this is OK.

chainstate/src/detail/in_memory_block_tree.rs
```diff
@@ -0,0 +1,629 @@
...
+}
+
+impl<'a> InMemoryBlockTreeRef<'a> {
+    fn new(arena: &'a Arena, root_id: NodeId) -> Self {
+        // Check that the arena has a node with this id.
+        // Note that unlike `InMemoryBlockTree`, here root_id may not be an actual root node
+        // (i.e. it may have a parent).
+        debug_assert!(arena.get(root_id).is_some());
+
+        Self { arena, root_id }
```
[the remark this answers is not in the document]
ImplOfAnImpl: > Why not return `Option<Self>`, and check in the arena if the root is really what's given? Then the caller can control whether this is debug or not through unwrap.

1. This is a private function, all its callers are known.
2. The callers should never try to create `InMemoryBlockTreeRef` using unrelated arena and root_id, if they do, it's always a bug. IMO it doesn't make much sense to report to a buggy caller that it's buggy.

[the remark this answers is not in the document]
TheQuantumPhysicist: I don't understand this limitation... why? Why do we have to make such limitations on data structures, in such a way where we make it unsafe to use them generally in the future? This InMemoryBlockTree is a very general thing that should be used as a general, isolated struct... by sheer coincidence you have for one usecase, but then someone needs it in the future and they can't use it... and then by that time, all the nuance that went into creating this abstraction are forgotten. So the future dev is left with two choices: Either recreate this, or use it and risk having bugs... sorry but this is too much for me to understand. I would prefer to isolate this in a module, and make it usable for anyone who needs it. The price is low and the win is high.

PR #1769, opened 2024-06-07
Implement `greater_equal_iter` and `OrderPreservingValue` in storage

ImplOfAnImpl:
`greater_equal_iter` allows iterating over items with keys bigger than or equal to the specified one. `OrderPreservingValue` is a wrapper whose `Encode`d form has the same ordering as the wrapped value itself (for now it's implemented only for u16/u32/u64).

I also did some cleanup here and there. In particular, `storage` traits now return `impl Iterator` and the associated type `PrefixIter` has been removed.

Note:
`storage-backend-test-suite` now has "frontend" tests (it's where I test `greater_equal_iter` and `OrderPreservingValue` together for all backend types); because of this, the crate now depends on the `storage` crate itself.
On the other hand, I removed the dependency of `storage` on `storage-inmemory` (there was an old TODO to remove this dependency), so now `storage` doesn't depend on any of the db-specific storage crates.


common/src/uint/endian.rs
```diff
@@ -40,42 +40,49 @@ macro_rules! define_slice_to_le {
     };
 }
 macro_rules! define_be_to_array {
-    ($name: ident, $type: ty, $byte_len: expr) => {
+    (@PRV $name: ident, $type: ty, $byte_len: expr) => {
         #[inline]
         pub fn $name(val: $type) -> [u8; $byte_len] {
-            debug_assert_eq!(::core::mem::size_of::<$type>(), $byte_len); // size_of isn't a constfn in 1.22
```
TheQuantumPhysicist, line 46: Why remove the asserts?

ImplOfAnImpl: > Why remove the asserts?

'$byte_len' will now always be equal to `::core::mem::size_of::<$type>()` if the "public" variant of the macro is called (i.e. the non-`@PRV` one). To me it looked like the assert is somewhat redundant now, so I removed it. Do we want to protect ourselves against direct calls to the `@PRV` version of the macro?

storage/backend-test-suite/src/frontend.rs
```diff
@@ -0,0 +1,82 @@
...
+            let mut map = dbtx.get_mut::<TestMap, _>();
+            for val in &test_values {
+                map.put(val, ()).unwrap();
+            }
+            dbtx.commit().unwrap();
+
+            let sorted_test_values = test_values.sorted_by(|v1, v2| {
+                // Note: we explicitly sort by "inner".
+                v1.main_part.inner().cmp(&v2.main_part.inner())
+            });
```
TheQuantumPhysicist, line 67: Please also add tests for tuples that prefix with `OrderPreservingValue`. Something like `(OrderPreservingValue, Vec<u8>)`. This way, you ensure that the behavior of the database is not only for small keys, but all kinds of keys, short and long.

common/src/uint/endian.rs
TheQuantumPhysicist: I don't see the harm... especially that any optimization will remove them.

utils/src/sorted.rs
```diff
@@ -0,0 +1,44 @@
...
+    where
+        T: Ord;
+
+    fn sorted_by<F>(&self, compare: F) -> Self
+    where
+        F: FnMut(&T, &T) -> std::cmp::Ordering;
+}
+
+impl<T: Clone> Sorted<T> for Vec<T> {
+    fn sorted(&self) -> Self
```
TheQuantumPhysicist, line 27: Why not pass by value then sort `self`, which will give the caller the choice ot clone or not, instead of forcing the caller to clone?

utils/src/sorted.rs
```diff
@@ -0,0 +1,44 @@
...
+    where
+        T: Ord,
+    {
+        let mut this = (*self).clone();
+        this.sort();
+        this
+    }
+
+    fn sorted_by<F>(&self, compare: F) -> Self
```
TheQuantumPhysicist, line 36: Same here

TheQuantumPhysicist (approved):
Besides the comments, looks fine. I'll provide an approval so that it can be merged after addressing the comments.

storage/src/database/mod.rs
```diff
@@ -200,19 +200,19 @@ where
     }
 
     /// Iterator over entries with key starting with given prefix
-    pub fn prefix_iter<Pfx>(&self, prefix: &Pfx) -> crate::Result<impl '_ + EntryIterator<DbMap>>
+    pub fn prefix_iter<Pfx>(&self, prefix: &Pfx) -> crate::Result<impl 'tx + EntryIterator<DbMap>>
```
TheQuantumPhysicist, line 203: As a matter of style, we usually put the lifetime bound after the trait.

TheQuantumPhysicist: I know you didn't change it, but apparently that was an oversight.

TheQuantumPhysicist (approved):

storage/src/database/mod.rs
ImplOfAnImpl: > As a matter of style, we usually put the lifetime bound after the trait.

FYI: there were  quite a few places like that, I decided to fix them all - https://github.com/mintlayer/mintlayer-core/pull/1769/commits/799373465d992965aa150ff2b2b38cabe43b91e2
(or at least this is what I've found via a regex)

PR #1770, opened 2024-06-07
extract a signer and account key chain traits

OBorce:
- remove signing from the account module
- make a separate signer module with a trait

### DIFF
--- a/wallet/src/key_chain/account_key_chain/mod.rs
+++ b/wallet/src/key_chain/account_key_chain/mod.rs
@@ -45,7 +45,7 @@ use super::vrf_key_chain::VrfKeySoftChain;
 use super::{make_path_to_vrf_key, AccountKeyChains, MasterKeyChain, VRF_INDEX};
 
 /// This key chain contains a pool of pre-generated keys and addresses for the usage in a wallet
-pub struct AccountKeyChain {
+pub struct AccountKeyChainImpl {
     chain_config: Arc<ChainConfig>,
 
     account_index: U31,
@@ -75,7 +75,7 @@ pub struct AccountKeyChain {
     lookahead_size: ConstValue<u32>,
 }
 
-impl AccountKeyChain {
+impl AccountKeyChainImpl {
     pub fn new_from_root_key(
         chain_config: Arc<ChainConfig>,
         db_tx: &mut impl WalletStorageWriteLocked,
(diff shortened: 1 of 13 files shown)


wallet/src/key_chain/account_key_chain/mod.rs
```diff
@@ -694,5 +694,30 @@ impl AccountKeyChain {
     }
 }
 
+impl AccountKeyChains for AccountKeyChain {
```
TheQuantumPhysicist, line 697: I have to say the naming is weird... maybe `impl AccountKeyChain for AccountKeyChainImpl` is better?

In the future, when we split the implementation, then struct will have more specific names, like `VrfAccountKeyChain`

wallet/src/signer/software_signer/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+use crate::{
+    account::PartiallySignedTransaction,
+    key_chain::{make_account_path, AccountKeyChains, FoundPubKey, MasterKeyChain},
+};
+
+use super::{Signer, SignerError, SignerResult};
+
+pub struct SoftwareSigner<'a, T: WalletStorageReadUnlocked> {
+    db_tx: &'a T,
+    chain_config: Arc<ChainConfig>,
```
TheQuantumPhysicist, line 54: I get that this is for simplicity... but you already have a lifetime `'a`... why not use it for ChainConfig as well?

TheQuantumPhysicist: Never mind. I think this is fine.

wallet/src/signer/software_signer/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+use wallet_types::signature_status::SignatureStatus;
+
+use crate::{
+    account::PartiallySignedTransaction,
+    key_chain::{make_account_path, AccountKeyChains, FoundPubKey, MasterKeyChain},
+};
+
+use super::{Signer, SignerError, SignerResult};
+
+pub struct SoftwareSigner<'a, T: WalletStorageReadUnlocked> {
```
TheQuantumPhysicist, line 52: Unnecessary trait bound

wallet/src/signer/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+    DestinationNotFromThisWallet,
+    #[error("{0}")]
+    SignArbitraryMessageError(#[from] SignArbitraryMessageError),
+}
+
+type SignerResult<T> = Result<T, SignerError>;
+
+/// Signer trait responsible for signing transactions or challenges using a software or hardware
+/// wallet
+pub trait Signer {
```
TheQuantumPhysicist, line 56: I have a feeling that this will soon become an async trait... because when dealing with the hardware wallet, things will become async.

TheQuantumPhysicist (approved):

PR #1772, opened 2024-06-10
Support hashlock check in mintscript

azarovh:
Only checker was implemented. Translation will be supported in the PR with atomic swaps where new TxOutput type is introduced.

### DIFF
--- a/mintscript/src/checker/hashlock.rs
+++ b/mintscript/src/checker/hashlock.rs
@@ -15,7 +15,7 @@
 
 use crypto::hash::{self, hash};
 
-use crate::script::HashType;
+use crate::script::HashChallenge;
 
 #[derive(thiserror::Error, Debug, PartialEq, Eq, Clone)]
 pub enum HashlockError {
@@ -37,7 +37,7 @@ pub trait HashlockChecker {
 
     fn check_hashlock(
         &mut self,
-        hash_type: &HashType,
+        hash_challenge: &HashChallenge,
         preimage: &[u8; 32],
     ) -> Result<(), Self::Error>;
 }
@@ -49,7 +49,7 @@ impl HashlockChecker for NoOpHashlockChecker {
 
     fn check_hashlock(
         &mut self,
-        _hash_type: &HashType,
+        _hash_challenge: &HashChallenge,
         _preimage: &[u8; 32],
     ) -> Result<(), Self::Error> {
         Ok(())
@@ -63,19 +63,19 @@ impl HashlockChecker for StandardHashlockChecker {
 
     fn check_hashlock(
         &mut self,
-        hash_type: &HashType,
+        hash_challenge: &HashChallenge,
         preimage: &[u8; 32],
     ) -> Result<(), Self::Error> {
-        match hash_type {
-            HashType::HASH160(expected_hash) => {
+        match hash_challenge {
+            HashChallenge::HASH160(expected_hash) => {
                 let actual_hash = hash::<hash::Ripemd160, _>(hash::<hash::Sha256, _>(preimage));
 
                 ensure_hashes_equal(actual_hash.as_slice(), expected_hash)?;
             }
-            HashType::RIPEMD160(_)
-            | HashType::SHA1(_)
-            | HashType::SHA256(_)
-            | HashType::HASH256(_) => {
+            HashChallenge::RIPEMD160(_)
+            | HashChallenge::SHA1(_)
+            | HashChallenge::SHA256(_)
+            | HashChallenge::HASH256(_) => {
                 unimplemented!()
             }
         }
--- a/mintscript/src/checker/hashlock.rs
+++ b/mintscript/src/checker/hashlock.rs
@@ -67,15 +67,15 @@ impl HashlockChecker for StandardHashlockChecker {
         preimage: &[u8; 32],
     ) -> Result<(), Self::Error> {
         match hash_challenge {
-            HashChallenge::HASH160(expected_hash) => {
+            HashChallenge::Hash160(expected_hash) => {
                 let actual_hash = hash::<hash::Ripemd160, _>(hash::<hash::Sha256, _>(preimage));
 
                 ensure_hashes_equal(actual_hash.as_slice(), expected_hash)?;
             }
-            HashChallenge::RIPEMD160(_)
-            | HashChallenge::SHA1(_)
-            | HashChallenge::SHA256(_)
-            | HashChallenge::HASH256(_) => {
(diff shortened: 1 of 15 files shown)


mintscript/src/checker/hashlock.rs
```diff
@@ -0,0 +1,82 @@
...
+        &mut self,
+        hash_type: HashType,
+        expected_hash: &[u8],
+        preimage: &[u8],
+    ) -> Result<(), Self::Error> {
+        match hash_type {
+            HashType::HASH160 => {
+                ensure!(expected_hash.len() == 20, HashlockError::IncorrectHashSize);
+
+                let actual_hash = hash::<hash::Ripemd160, _>(hash::<hash::Sha256, _>(preimage));
```
iljakuklic, line 71: I think we want to constrain the preimage size too, is it done somewhere?

From [miniscript docs](https://bitcoin.sipa.be/miniscript/):

> Hash preimages are constrained to 32 bytes to disallow various forms of griefing, including making non-standard(un-relayable) transactions, consensus-invalid swaps across blockchains, as well as ensure that satisfaction cost can be accurately calculated.

iljakuklic (commented):
Looks good overall. The preimage size issue needs to be addressed.

mintscript/src/script/mod.rs
```diff
@@ -121,11 +121,25 @@ impl Threshold {
...
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum HashType {
+    RIPEMD160,
+    SHA1,
+    SHA256,
+    HASH160,
+    HASH256,
+}
```
iljakuklic, line 131: A minor suggeston here:

Maybe the hash could be included in the enum arm as data.

```rust
#[derive(Copy, Clone, PartialEq, Eq, Debug)]
pub enum HashType {
    RIPEMD160([u8; 20]),
    SHA256([u8; 32]),
    // ...
}
```
(Or use a stronger types than arrays).

This way the hash size check is pushed into the decoding phase and presumably will not have to be repeated in the implementation. If the hash is a vec, the size is encoded by the vec as well as being implicitly determined by the hash type leading to duplicate information.

Is there a reason why you went for this instead?

azarovh: No specific reason. Your suggestion for enum seems better indeed

mintscript/src/tests/hashlock_checker.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    fn source_time(&self) -> Result<BlockTimestamp, Self::Error> {
+        unreachable!()
+    }
+}
+
+#[rstest::rstest]
+#[case([209, 246, 177, 232, 51, 84, 228, 19, 240, 217, 24, 152, 165, 219, 108, 141, 8, 142, 84, 141, 210, 128, 132, 209, 44, 87, 61, 197, 94, 173, 23, 113],
+       [157, 6, 130, 9, 15, 13, 14, 19, 25, 16, 83, 196, 241, 85, 78, 166, 172, 66, 16, 178])]
```
TheQuantumPhysicist, line 64: Instead of these ugly arrays... use hex!

TheQuantumPhysicist (commented):
First round looks good. I'll take another round after merging the base PR.

mintscript/src/script/mod.rs
```diff
@@ -121,11 +121,24 @@ impl Threshold {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum HashType {
```
iljakuklic, line 125: I'd say `HashType` is no longer an accurate name since it now actually includes the hash itself. Not that I have a great suggestion.

mintscript/src/script/mod.rs
```diff
@@ -121,11 +121,24 @@ impl Threshold {
...
+    HASH256([u8; 32]),
+}
+
 #[derive(Clone, PartialEq, Eq, Debug)]
 enum WitnessScriptInner {
     Threshold(Threshold),
     Signature(Destination, InputWitness),
     Timelock(OutputTimeLock),
+    HashLock {
+        hash_type: HashType,
```
iljakuklic, line 139: Also the field name here.

iljakuklic (approved):

mintscript/src/script/mod.rs
azarovh: Idk, maybe just `Hash`?

iljakuklic: > Idk, maybe just `Hash`?

I would say why not but it conflicts with the `Hash` trait from standard library so not quite sure. How about something like `HashChallenge`? I don't know if this is good terminology since the work challenge is used in cryptography but I don't know how exactly so not sure if this is vaguely appropriate. @TheQuantumPhysicist

TheQuantumPhysicist: Maybe Hashed, just to avoid the collision with std? Challenge seem to be OK as well. I'll have to think about it. Pick what you find appropriate now.

azarovh: renamed to `HashChallenge `

mintscript/src/script/mod.rs
```diff
@@ -122,12 +122,25 @@ impl Threshold {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum HashChallenge {
+    RIPEMD160([u8; 20]),
+    SHA1([u8; 20]),
+    SHA256([u8; 32]),
+    HASH160([u8; 20]),
+    HASH256([u8; 32]),
```
TheQuantumPhysicist, line 131: I don't think these need to be capitalized like this. Hash160, Sha256, etc, is OK.

TheQuantumPhysicist (approved):
Besides the comment, seems fine.

mintscript/src/checker/hashlock.rs
```diff
@@ -0,0 +1,82 @@
...
+
+                ensure_hashes_equal(actual_hash.as_slice(), expected_hash)?;
+            }
+            HashChallenge::Ripemd160(_)
+            | HashChallenge::Sha1(_)
+            | HashChallenge::Sha256(_)
+            | HashChallenge::Hash256(_) => {
+                unimplemented!()
+            }
```
iljakuklic, line 74: What about these?

iljakuklic (approved):

mintscript/src/checker/hashlock.rs
TheQuantumPhysicist: I think this is unreachable now, since hash160 is manually constructed.

PR #1775, opened 2024-06-13
tx-verifier: More informative input check errors

iljakuklic:
* Use full information from errors we get from script
* Attach input index to each error so it's clear which input failed verification
* Prune `ConnectTransactionError` of no longer used arms


test/functional/mempool_timelocked_tx.py
```diff
@@ -81,7 +81,7 @@ def run_test(self):
         node.mempool_submit_transaction(tx1, {})
         node.mempool_submit_transaction(tx3, {})
         # Cannot submit tx2 yet, it spends and unconfirmed time-locked output
-        assert_raises_rpc_error(None, "Timelock rules violated", node.mempool_submit_transaction, tx2, {})
+        assert_raises_rpc_error(None, "locked until", node.mempool_submit_transaction, tx2, {})
```
TheQuantumPhysicist, line 84: This should be more informative with explicit error information. @azarovh can you please fix this?

TheQuantumPhysicist (approved):

test/functional/mempool_timelocked_tx.py
azarovh: The information is there. This is just partial string comparison to check that this string is in the error after rpc call.

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
```diff
@@ -29,7 +31,58 @@ use mintscript::{
...
+impl From<mintscript::script::ScriptError<Infallible, TimelockError>> for InputCheckErrorPayload {
+    fn from(value: mintscript::script::ScriptError<Infallible, TimelockError>) -> Self {
+        Self::Verification(value.errs_into())
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, thiserror::Error, Debug)]
+#[error("Error verifying input #{input_num}: {error}")]
+pub struct InputCheckError {
+    input_num: usize,
```
azarovh, line 63: input number alone won't be that usefull if block contains several txs. Is it easy to add source_id?

azarovh (approved):

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
iljakuklic: > input number alone won't be that usefull if block contains several txs. Is it easy to add source_id?

I'd say that should be done at different level in the transaction verifier by technique similar to here. I.e. blck verification procedure attaches the block ID to the error, transaction verification attaches transaction ID to all transaction verification errors.

The full error would then look something like:
```
Error verifying block 0xaabbcc: error verifying transaction 0xb1ab1a: Error in input #4: Bad signature or whatever
```

test/functional/mempool_timelocked_tx.py
iljakuklic: > This should be more informative with explicit error information. @azarovh can you please fix this?

The full error is `"Spending at height {0}, locked until height {1}"`. Just the test matches only a part of it since I am not sure if the heights are stable across runs.

chainstate/tx-verifier/src/transaction_verifier/input_check/mod.rs
azarovh: ok agree

PR #1778, opened 2024-06-14
Implement get_stake_pool_balances_for_tree

ImplOfAnImpl:


chainstate/src/detail/chainstateref/in_memory_reorg.rs
TheQuantumPhysicist, line 1: Simple and well abstracted concepts are becoming more convoluted and complex... we need to have a discussion. I'm tired now, so let's have that discussion tomorrow.

Generally speaking, we can't just keep adding arbitrarily more and more and more lines of code... maybe I'm not getting why this is necessary, but this trend is dangerous. Things have to remain abstracted well and simple... more code is basically a liability.

If the additional lines of code are unrelated to reorging in memory, then please move them to submodules and test them in isolation. Otherwise, let's talk tomorrow about it.

PR #1786, opened 2024-06-25
Implement in_mem_block_tree

ImplOfAnImpl:
This is the generalized version of the "in-memory block tree" from #1768 . The corresponding `Trees`/`Tree`/`TreeRef` structs are in `tree_wrappers.rs`. However, I also had to wrap all `indextree`'s primitives (which are in the `primitives` module), so that they can be exposed to the user and still maintain the required guarantees (i.e. that a parent/child nodes will always correspond to parent/child blocks).

### DIFF
--- a/utils/in_mem_item_tree/src/primitives/detail.rs
+++ b/utils/in_mem_item_tree/src/primitives/detail.rs
@@ -1,0 +1,262 @@
+// Copyright (c) 2021-2024 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::BTreeMap;
+
+use utils::{debug_assert_or_log, displayable_option::DisplayableOption, ensure};
+
+use super::{arena::Arena, node_id::NodeId, DataItem, Error, Flavor};
+
+/// The trait that is referenced by the `Flavor` trait. `Arena` holds an instance of such
+/// an observer and notifies it about the corresponding modifications.
+pub trait ModificationObserver<T: DataItem>: Clone + std::fmt::Debug + Default {
+    fn on_node_added(&mut self, node_id: NodeId, item_id: <T as DataItem>::Id)
+        -> Result<(), Error>;
+    fn on_node_removed(&mut self, item_id: &<T as DataItem>::Id);
+    fn on_whole_arena_cleared(&mut self);
+}
+
+/// `Arena` uses this trait to get access to the item-id-to-node-id map that is held within
+///  the observer, if any.
+pub trait ItemIdMapHolder<T: DataItem> {
+    fn item_id_map(&self) -> &BTreeMap<<T as DataItem>::Id, NodeId>;
+}
+
+/// This is the observer that always holds the item-id-to-node-id map, which allows `Arena`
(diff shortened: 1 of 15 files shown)


common/src/chain/in_mem_block_tree/tree_wrappers.rs
```diff
@@ -0,0 +1,413 @@
+// Copyright (c) 2021-2024 RBB S.r.l
```
ImplOfAnImpl, line 1: This module needs its own tests (in particular for the `ensure_node_in_subtree` calls).

TheQuantumPhysicist (commented):
Can we move this to its own crate in utils?

TheQuantumPhysicist (approved):
Too generic for my taste, but I'm OK with it. But please move it to utils in its own crate... this doens't belong in primitives.

ImplOfAnImpl:
> Can we move this to its own crate in utils?

`in_mem_block_tree` depends on `common` because it needs primitives related to blocks, such as `Id` and `GenBlock`. So, `utils` is not a good place for it.
But I agree that the current place is not the best for it either.

ImplOfAnImpl:
> > Can we move this to its own crate in utils?
> 
> `in_mem_block_tree` depends on `common` because it needs primitives related to blocks, such as `Id` and `GenBlock`. So, `utils` is not a good place for it. But I agree that the current place is not the best for it either.

I've abstracted `Id<GenBlock>` away from it and then moved it to `utils`.
Some caveats:
1) I've renamed it to `in_mem_item_tree`, because the crate has nothing to do with blocks anymore; I don't really like this new name, but I couldn't come up with anything better.
2) `utils/in_mem_item_tree` depends on `utils` and not vice versa, like its other crates. This doesn't look nice, but we already have similar inconsistency in other places, e.g. in the wallet, so I guess it's ok?

TheQuantumPhysicist:
That's alright.

utils/in_mem_item_tree/src/primitives/detail.rs
```diff
@@ -0,0 +1,262 @@
...
+impl<T: DataItem> ItemIdMapHolder<T> for TrackingModificationObserver<T> {
+    fn item_id_map(&self) -> &BTreeMap<<T as DataItem>::Id, NodeId> {
+        &self.0
+    }
+}
+
+/// This is the observer that only holds the item-id-to-node-id map in debug builds, to perform
+/// the corresponding runtime checks.
+pub struct DebugOnlyTrackingModificationObserver<T: DataItem> {
+    #[cfg(debug_assertions)]
```
TheQuantumPhysicist, line 97: Just note that the overuse of cfg like this complicates testing. I wouldn't recommend doing this. I would prefer to have a run-time configurable flag. Up to you.

TheQuantumPhysicist (approved):

utils/in_mem_item_tree/src/primitives/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+    #[error("Node id {0} is not in arena")]
+    NodeIdNotInArena(NodeId),
+    #[error("Node id {0} is removed")]
+    NodeIsRemoved(NodeId),
+    #[error(
+        "Item {child_item_id} has parent {childs_parent_item_id} instead of expected {expected_parent_item_id}",
+    )]
+    NotParentChild {
+        child_item_id: String,
+        childs_parent_item_id: String,
```
TheQuantumPhysicist, line 102: children?

ImplOfAnImpl: > children?

It's supposed to be read as `child's`.

TheQuantumPhysicist: I would remove the possessive 's', but not a big deal.

utils/in_mem_item_tree/src/primitives/detail.rs
ImplOfAnImpl: > Just note that the overuse of cfg like this complicates testing. I wouldn't recommend doing this. I would prefer to have a run-time configurable flag. Up to you.

Since you don't insist, I'll leave it as is. IMO it won't complicate testing of the user code, because it basically adds some checks in debugs builds.

PR #1787, opened 2024-06-27
Replace `AccountsBalanceTracker` with accounting deltas

azarovh:
`AccountsBalanceTracker` was used to check if multiple account spend in a single tx won't underflow the balance. To avoid duplicating code when adding support for Orders I use Deltas now. They already implement all the constraint check and are easily reusable for the job.


chainstate/constraints-value-accumulator/src/constraints_accumulator.rs
```diff
@@ -72,7 +70,11 @@ impl ConstrainedValueAccumulator {
 
         let mut accumulator = Self::new();
         let mut total_to_deduct = BTreeMap::<CoinOrTokenId, Amount>::new();
-        let mut accounts_balances_tracker = AccountsBalancesTracker::new(pos_accounting_view);
+
+        // Temp deltas are used to check accounting errors like overspends across multiple inputs
+        let mut temp_pos_accounting = pos_accounting::PoSAccountingDelta::new(pos_accounting_view);
+        let mut temp_orders_accounting =
+            orders_accounting::OrdersAccountingCache::new(orders_accounting_view);
```
TheQuantumPhysicist, line 78: What about tokens accounting?

azarovh: done

api-server/scanner-lib/src/sync/tests/simulation.rs
```diff
@@ -113,6 +121,34 @@ impl<'a> PoSAccountingView for PoSAccountingAdapterToCheckFees<'a> {
     }
 }
 
+struct StubTokensAccounting;
```
TheQuantumPhysicist, line 124: I thought we used the word "Mock" instead of "Stub"... it's a matter of convention.

TheQuantumPhysicist (approved):

api-server/scanner-lib/src/sync/tests/simulation.rs
azarovh: Idk it's actually a stub with hardcoded state to satisfy function signature. It doesn't have any logic or expectation as mocks

PR #1788, opened 2024-06-28
Decrease `max_future_block_time_offset`

azarovh:


p2p/src/sync/peer/block_manager.rs
```diff
@@ -688,7 +692,12 @@ where
         {
             let new_block_headers = new_block_headers.clone();
             self.chainstate_handle
-                .call(move |c| Ok(c.preliminary_headers_check(&new_block_headers)?))
+                .call(move |c| {
+                    Ok(c.preliminary_headers_check(
+                        &new_block_headers,
+                        first_header_prev_block_index.block_height(),
```
TheQuantumPhysicist, line 698: I think this isn't the correct way to do this. The right way would make the check depend on the block height, not on the string of blocks and their coincidental first block height.

p2p/src/sync/peer/block_manager.rs
```diff
@@ -606,9 +610,6 @@ where
             return Err(P2pError::ProtocolError(ProtocolError::DisconnectedHeaders));
         }
 
-        let last_header = headers.last().expect("Headers shouldn't be empty");
-        self.wait_for_clock_diff(last_header.timestamp()).await;
```
TheQuantumPhysicist, line 610: Why removing this? Please be careful with such radical changes without studying the consequences.

TheQuantumPhysicist (commented):
Round

p2p/src/sync/peer/block_manager.rs
azarovh: I moved it to the line 644 where the block height is available

p2p/src/sync/peer/block_manager.rs
azarovh: `preliminary_headers_check` only checks the first header in the list, so only one height is required. I made it more explicit though

chainstate/src/detail/chainstateref/mod.rs
```diff
@@ -599,7 +599,11 @@ impl<'a, S: BlockchainStorageRead, V: TransactionVerificationStrategy> Chainstat
     }
 
     #[log_error]
-    pub fn check_block_header(&self, header: &SignedBlockHeader) -> Result<(), CheckBlockError> {
+    pub fn check_block_header(
+        &self,
+        header: &SignedBlockHeader,
+        block_height: BlockHeight,
+    ) -> Result<(), CheckBlockError> {
         self.check_block_parent(header)?;
```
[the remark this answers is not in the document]
azarovh: right, thanks

common/src/chain/config/mod.rs
```diff
@@ -57,6 +57,7 @@ use super::{
 };
 
 const DEFAULT_MAX_FUTURE_BLOCK_TIME_OFFSET: Duration = Duration::from_secs(120);
+const MAX_FUTURE_BLOCK_TIME_OFFSET: Duration = Duration::from_secs(30);
```
ImplOfAnImpl, line 60: From `ChainConfig`'s point of view, both of them are "default". It's just that one of them is the old default and the other one is new.

common/src/chain/config/mod.rs
```diff
@@ -467,8 +468,13 @@ impl ChainConfig {
...
     #[must_use]
-    pub fn max_future_block_time_offset(&self) -> &Duration {
-        &self.max_future_block_time_offset
+    pub fn max_future_block_time_offset(&self, height: BlockHeight) -> Duration {
+        self.max_future_block_time_offset.unwrap_or_else(|| {
+            match self.as_ref().chainstate_upgrades().version_at_height(height).1.htlc_activated() {
+                HtlcActivated::Yes => MAX_FUTURE_BLOCK_TIME_OFFSET,
+                HtlcActivated::No => DEFAULT_MAX_FUTURE_BLOCK_TIME_OFFSET,
+            }
+        })
```
ImplOfAnImpl, line 477: I makes sense to leave a comment that we've switched to the new offset at the same height where we've enabled htlc, and that otherwise they have nothing to do with each other (because judging from this code it seems like they do).

ImplOfAnImpl (approved):

PR #1791, opened 2024-07-05
Introduce `mintscript::SigntureOnlyChecker` and use it in the wallet

azarovh:
This PR addresses a problem that was introduced by HTLC change: `StandardInputSignature::raw_signature` can encode additional layer with `AuthorizedHashedTimelockContractSpend`. Meaning that `common::StandardInputSignature::verify_signtaure` cannot be called directly and `InputWitness` must be first evaluated with mintscript.

- added  `mintscript::SignatureOnlyChecker` and public function in `tx_verifier::verify_tx_signature`
- moved `PartiallySignedTransaction` to common and remove sig verification logic from inside
- removed verify_signature call from inside of `PartiallySIgnedTransaction`
- use new verify_tx_function in the wallet


common/src/chain/transaction/partially_signed_transaction.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        Ok(Self {
+            tx,
+            witnesses,
+            input_utxos,
+            destinations,
+        })
+    }
+
+    pub fn new_witnesses(mut self, witnesses: Vec<Option<InputWitness>>) -> Self {
```
ImplOfAnImpl, line 63: "with_witnesses" ?

mintscript/src/translate.rs
```diff
@@ -321,3 +341,114 @@ impl<C: InputInfoProvider> TranslateInput<C> for TimelockOnly {
...
+pub struct SignatureOnlyReward;
+
+impl<C: SignatureInfoProvider> TranslateInput<C> for SignatureOnlyReward {
+    fn translate_input(_ctx: &C) -> Result<WitnessScript, TranslationError> {
+        // Not used anywhere.
+        // But it's important to outline that if needed the reward implementation must be different
+        // because staking/decommissioning destinations are not the same.
+        unimplemented!()
+    }
+}
```
ImplOfAnImpl, line 454: Why keep it if it's unused?

ImplOfAnImpl (approved):

mintscript/src/translate.rs
azarovh: I wanted to make it explicit for the future. So that noone tried to use implementation from transaction with reward (it might be a problem because both implement Transactable)

common/src/chain/transaction/partially_signed_transaction.rs
```diff
@@ -0,0 +1,144 @@
...
+                (Some(InputWitness::Standard(_)), None) => false,
+                (Some(InputWitness::Standard(_)), Some(_)) => true,
+                (None, _) => false,
+            })
+    }
+
+    pub fn into_signed_tx(self) -> Result<SignedTransaction, TransactionCreationError> {
+        if self.all_signatures_available() {
+            let witnesses = self.witnesses.into_iter().map(|w| w.expect("cannot fail")).collect();
+            Ok(SignedTransaction::new(self.tx, witnesses)?)
```
OBorce, line 113: This function used to check if the signatures are valid or not. Is that not possible now?

wallet/src/wallet/mod.rs
```diff
@@ -970,8 +969,8 @@ impl<B: storage::Backend> Wallet<B> {
             let tx = signer
                 .sign_tx(ptx, account.key_chain())
                 .map(|(ptx, _, _)| ptx)?
-                .into_signed_tx(chain_config)
-                .map_err(error_mapper)?;
+                .into_signed_tx()
+                .map_err(|e| error_mapper(WalletError::TransactionCreation(e)))?;
```
OBorce, line 973: we should verify the signatures after this, just in case the user selected a multisig utxo as an input, in that case the input will not be fully signed.

chainstate/tx-verifier/src/transaction_verifier/input_check/signature_only_check.rs
```diff
@@ -0,0 +1,178 @@
...
+    ChainConfig, DelegationId, Destination, PoolId, SignedTransaction, TxInput, TxOutput,
+};
+use mintscript::{
+    script::ScriptError, translate::InputInfoProvider, InputInfo, SignatureContext, TranslateInput,
+};
+use utils::ensure;
+
+use super::{InputCheckError, InputCheckErrorPayload, PerInputData};
+
+struct InputVerifyContextSignature<'a, T> {
```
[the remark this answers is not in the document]
azarovh: this is just a name in line with others like `InputVerifyContextTimelock`

wallet/src/wallet/mod.rs
azarovh: add verification code

OBorce (approved):

PR #1796, opened 2024-07-22
Transfer tokens from a multisig address

ImplOfAnImpl:
The wallet command `MakeTxToSendTokensFromMultisigAddress` and the corresponding rpc function have been added.


wallet/wallet-controller/src/synced_controller.rs
```diff
@@ -630,6 +649,154 @@ impl<'a, T: NodeInterface, W: WalletEvents> SyncedController<'a, T, W> {
...
+    // TODO: this discrepancy between tokens/coins fee handling is a bit ugly, it's better to unify it.
+    // Note: the reason for this specific fee change behavior is that this function is called from
+    // `make_tx_to_send_tokens_from_multisig_address`, which is supposed to be used in automated
+    // scenarios; issuing a new address for each call seems redundant in such a case.
+    // The token change being sent to first unused address doesn't have any particular reason;
+    // it's just how this function's callee behaves by default.
+    pub async fn make_unsigned_tx_to_send_tokens_to_addresses(
+        &mut self,
+        inputs: Vec<(UtxoOutPoint, TxOutput)>,
+        outputs: BTreeMap<TokenId, Vec<GenericTxOutput>>,
```
azarovh, line 675: I think `GenericTxOutput` adds unnecessary complexity. Providing `BtreeMap<TokenId, (DecimalAmount, Destination)>` would do the job without introducing another TxOutput analog. In terms of wallet `send_*` is always `TxOutput::Transfer` and I don't think we would need anything else.

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -424,6 +424,26 @@ pub enum WalletCommand {
...
     },
 
+    /// Create a transaction for sending tokens from a multisig address to other addresses, returning the change to
+    /// the original multisig address.
+    ///
+    /// The utxos to pay fees from will be selected automatically; these will be normal, single-sig utxos.
+    /// The optional `fee_change_address` specifies the destination for the change for the fee payment;
+    /// If it's unset, the destination will be taken from one of existing single-sig utxos.
+    #[clap(name = "token-make-tx-to-send-from-multisig-address")]
+    MakeTxToSendTokensFromMultisigAddress {
```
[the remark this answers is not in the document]
ImplOfAnImpl: [discussed] `MakeTxToSendTokensFromMultisigAddress` may still be useful for devs to toy with this functionality, so it makes sense to keep it, but perhaps only as a hidden command.

P.S. Also, all additional string parsing that this command requires is in the `parse_generic_token_output` function. The other changes in `wallet/wallet-cli-commands/src/helper_types.rs` can be seen as general improvements and cleanup, and are potentially useful on their own.

wallet/wallet-controller/src/synced_controller.rs
ImplOfAnImpl: > So I'll make it a struct and rename to something like CurrencyTransfer. Also, GenericTxTokenOutput will become TokenTransfer

I've named them `GenericCurrencyTransfer` and `GenericTokenTransfer` respectively.

azarovh (approved):
looks good to me but wait for Boris to review as well

OBorce (approved):

PR #1797, opened 2024-07-30
Add missing wallet functionality to support atomic swaps

azarovh:
- `wallet::SoftwareSigner` now can sign outputs with htlc secret or htlc multisig.
- `PartiallySignedTransaction` was extended with information about htlc secret per input.
- `compose_transaction` accepts htlc secret
- Added functional tests for htlc spend with secret and refund with multisig
- Added new functions to wasm bindings that should cover atomic swap scenario for js clients

So the way I see atomic swap can be implemented now:

1. Alice creates an output with htlc either using `rpc::create_htlc_transaction` or manually
2. Alice manually creates refund transaction, which is signed by Bob with `sign_raw_transaction`
3. Bob performs the same steps
4. Alice spends Bob's output by calling `rpc::compose_transaction` where secret is passed, after that it's signed via `sign_raw_transaction`
5. Bob sees that tx broadcasted and can extract secret from input witness manually or for js via `extract_htlc_secret`
6. Bob now can spend Alice's output with extracted secret and `compose_transaction`
7. In case of refund Alice or Bob signs refund tx with `sign_raw_transaction` and broadcast it after timlock period.


wallet/src/wallet/mod.rs
```diff
@@ -168,13 +170,13 @@ pub enum WalletError {
...
+    #[error("Cannot find UTXO {0:?}")]
     CannotFindUtxo(UtxoOutPoint),
-    #[error("Selected UTXO is already consumed")]
+    #[error("Selected UTXO is already consumed  {0:?}")]
     ConsumedUtxo(UtxoOutPoint),
-    #[error("Selected UTXO is still locked")]
+    #[error("Selected UTXO is still locked  {0:?}")]
     LockedUtxo(UtxoOutPoint),
-    #[error("Selected UTXO is a token v0 and cannot be used")]
+    #[error("Selected UTXO is a token v0 and cannot be used  {0:?}")]
```
ImplOfAnImpl, line 179: 1) Extra space.
2) `Selected UTXO {0:?} ...` would look nicer.

wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -635,6 +636,18 @@ trait WalletRpc {
         options: TransactionOptions,
     ) -> rpc::RpcResult<NewTransaction>;
 
+    /// Creates a transaction that lock a given number of coins or tokens in a Hashed Timelock Contract.
```
ImplOfAnImpl, line 639: locks

wallet/wallet-rpc-lib/src/rpc/mod.rs
```diff
@@ -1274,19 +1277,104 @@ impl<N: NodeInterface + Clone + Send + Sync + 'static> WalletRpc<N> {
...
+            refund_timelock: htlc.refund_timelock,
+            refund_key,
+        };
+
+        let token_id = match token_id {
+            Some(id) => {
+                Some(id.decode_object(&self.chain_config).map_err(|_| RpcError::InvalidTokenId)?)
+            }
+            None => None,
+        };
```
ImplOfAnImpl, line 1313: map + transpose?

Not insisting though.

wallet/wallet-rpc-lib/src/rpc/mod.rs
```diff
@@ -1274,19 +1277,104 @@ impl<N: NodeInterface + Clone + Send + Sync + 'static> WalletRpc<N> {
...
         only_transaction: bool,
     ) -> WRpcResult<(TransactionToSign, Balances), N> {
         ensure!(!inputs.is_empty(), RpcError::ComposeTransactionEmptyInputs);
         let inputs = inputs.into_iter().map(|o| o.into_outpoint()).collect();
+
+        let htlc_secrets = htlc_secrets
+            .map(|s| {
+                s.into_iter()
+                    .map(|s| {
+                        s.map(|s| -> Result<HtlcSecret, RpcError<N>> {
```
ImplOfAnImpl, line 1360: Choosing different names instead of s would make the code a little more readable. E.g. `secrets_opt`, `secret_opt`, `secret`.

wallet/wallet-rpc-lib/src/rpc/server_impl.rs
```diff
@@ -232,7 +234,7 @@ impl<N: NodeInterface + Clone + Send + Sync + Debug + 'static> ColdWalletRpcServ
                     let is_complete = tx.all_signatures_available()
                         && cur_signatures.iter().all(|s| *s == SignatureStatus::FullySigned);
                     let hex = if is_complete {
-                        let tx = tx.into_signed_tx().expect("already checked");
+                        let tx = tx.into_signed_tx().expect("already checked3");
```
ImplOfAnImpl, line 237: extra char in checked3

wallet/wallet-rpc-lib/src/rpc/server_impl.rs
```diff
@@ -396,12 +398,19 @@ impl<N: NodeInterface + Clone + Send + Sync + Debug + 'static> WalletRpcServer f
     async fn get_balance(
         &self,
         account_arg: AccountArg,
+        utxo_states: Vec<RpcUtxoState>,
         with_locked: Option<WithLocked>,
     ) -> rpc::RpcResult<Balances> {
+        let utxo_states = if utxo_states.is_empty() {
+            UtxoStates::ALL
+        } else {
+            utxo_states.iter().map(UtxoState::from).fold(UtxoStates::NONE, |x, y| x | y)
```
ImplOfAnImpl, line 407: It's a 3rd place where this `fold` occurs. I'd add something like `UtxoStates::from_states` accepting `impl IntoIterator<Item = UtxoState>`.

Not insisting on this.

wasm-wrappers/src/error.rs
```diff
@@ -61,6 +70,10 @@ pub enum Error {
     FixedTotalSupply,
     #[error("Invalid token parameters: {0}")]
     TokenIssuanceError(#[from] tx_verifier::error::TokenIssuanceError),
+    #[error("Transaction creation error: {0}")]
+    TransactionCreationError(#[from] TransactionCreationError),
+    #[error("Transaction creation error: {0}")]
```
ImplOfAnImpl, line 75: Wrong description

wasm-wrappers/src/lib.rs
```diff
@@ -725,6 +736,90 @@ pub fn encode_output_data_deposit(data: &[u8]) -> Result<Vec<u8>, Error> {
     Ok(output.encode())
 }
 
+/// Given the parameters needed to create hash timelock contract, and a network type (mainnet, testnet, etc),
+/// this function creates an output.
+#[wasm_bindgen]
+pub fn encode_output_htlc(
+    amount: Amount,
+    token_id: Option<String>,
```
ImplOfAnImpl, line 744: Why not `Option<&str>`?

wasm-wrappers/src/lib.rs
```diff
@@ -725,6 +736,90 @@ pub fn encode_output_data_deposit(data: &[u8]) -> Result<Vec<u8>, Error> {
...
+        SignedTransaction::decode(&mut &signed_tx_bytes[..])
+            .map_err(|_| Error::InvalidTransaction)?
+    };
+
+    let htlc_position = tx.transaction().inputs().iter().position(|input| match input {
+        TxInput::Utxo(outpoint) => *outpoint == htlc_utxo_outpoint,
+        TxInput::Account(_) | TxInput::AccountCommand(_, _) => false,
+    });
+
+    match htlc_position {
```
ImplOfAnImpl, line 804: `htlc_position.ok_or(Error::InvalidWitnessCount)` instead of `match`?

wasm-wrappers/src/lib.rs
azarovh:
```
the trait bound `&str: OptionFromWasmAbi` is not satisfied
the trait `OptionFromWasmAbi` is implemented for `std::string::String`
```

OBorce (approved):
I guess for a future feature, we could try to make the wallet more user friendly, a way to automatically create and store the HTLC's secret when creating a new HTLC, and scan newly spent HTLC's to extract and compare the secret to any HTLC that is being tracked by the wallet.

azarovh:
> we could try to make the wallet more user friendly

The plan is to avoid implementing the atomic swaps process in the wallet and delegate it to external products. In that scenario we provide only essential building blocks like verification, signing etc. Even `rpc::create_htlc_transaction` is not mandatory and was implemented to simplify functional testing.

common/src/chain/transaction/partially_signed_transaction.rs
```diff
@@ -36,27 +39,35 @@ impl PartiallySignedTransaction {
...
-            input_utxos.len() == destinations.len(),
-            TransactionCreationError::InvalidWitnessCount
+            tx.inputs().len() == destinations.len(),
+            TransactionCreationError::InvalidDestinationsCount
+        );
+
+        let htlc_secrets = htlc_secrets.unwrap_or_else(|| vec![None; tx.inputs().len()]);
+        ensure!(
+            htlc_secrets.len() == tx.inputs().len(),
+            TransactionCreationError::InvalidHtlcSecretsCount
```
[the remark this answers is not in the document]
azarovh: good point

wallet/src/send_request/mod.rs
```diff
@@ -311,7 +316,7 @@ where
         | TxOutput::DelegateStaking(_, _)
         | TxOutput::DataDeposit(_)
         | TxOutput::AnyoneCanTake(_) => None,
-        TxOutput::Htlc(_, _) => None, // TODO(HTLC)
+        TxOutput::Htlc(_, htlc) => Some(htlc.spend_key.clone()),
```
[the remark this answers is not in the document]
azarovh: I ended up adding a parameter for htlc and also introducing  new function that returns all possible Destinations. Ideally wallet should not rely on destinations that much and the knowledge should be isolated. But I tried different things and couldn't come up with a solution. So I suggest that for now it's good enough.

wallet/src/signer/software_signer/mod.rs
```diff
@@ -344,5 +372,69 @@ fn get_private_key(
...
+                    AuthorizedHashedTimelockContractSpend::from_data(sig.raw_signature())?;
+                match htlc_spend {
+                    AuthorizedHashedTimelockContractSpend::Secret(_, _) => {
+                        return Err(SignerError::SigningError(
+                            DestinationSigError::IncompleteClassicalMultisigAuthorization,
+                        ));
+                    }
+                    AuthorizedHashedTimelockContractSpend::Multisig(raw_signature) => {
+                        AuthorizedClassicalMultisigSpend::from_data(&raw_signature)?
+                    }
```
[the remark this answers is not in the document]
azarovh: Added new `EvaluatedInputWitness` type as well as moving encode/decode functions to common

wallet/src/account/mod.rs
```diff
@@ -1270,8 +1298,12 @@ impl Account {
 
         Ok((
             txo.clone(),
-            get_tx_output_destination(txo, &|pool_id| self.output_cache.pool_data(*pool_id).ok())
-                .ok_or(WalletError::InputCannotBeSpent(txo.clone()))?,
+            get_tx_output_destination(
+                txo,
+                &|pool_id| self.output_cache.pool_data(*pool_id).ok(),
+                HtlcSpendingCondition::Undefined,
```
ImplOfAnImpl, line 1304: `Undefined` looks cryptic. It's better to use a name that would clearly indicate that HTLC outputs won't be included in the result. E.g. something like `SkipHtlc`?
(not insisting on any particular name)

ImplOfAnImpl (approved):

wallet/src/account/mod.rs
azarovh: Ok I'll rename to `HtlcSpendingCondition::Skip`

PR #1799, opened 2024-08-06
add additional UTXO info to partially signed tx

OBorce:
add additional UTXO info for tokens (number of decimals and ticker) and pools (balance) to partially signed tx, intended to be shown on a hardware wallet UI like Trezor.


common/src/chain/tokens/rpc.rs
```diff
@@ -51,6 +51,13 @@ impl RPCTokenInfo {
             Self::NonFungibleToken(_) => 0,
         }
     }
+
+    pub fn token_ticker(&self) -> Vec<u8> {
```
ImplOfAnImpl, line 55: Just return `&[u8]` and let the caller decide whether to call `to_owned`.

chainstate/src/rpc/mod.rs
```diff
@@ -142,6 +142,12 @@ trait ChainstateRpc {
     #[method(name = "staker_balance")]
     async fn staker_balance(&self, pool_address: String) -> RpcResult<Option<Amount>>;
 
+    /// Returns the pool data of the pool associated with the given pool address.
+    ///
+    /// Returns `None` (null) if the pool is not found.
+    #[method(name = "pool_data")]
+    async fn pool_data(&self, pool_address: String) -> RpcResult<Option<RpcPoolData>>;
```
ImplOfAnImpl, line 149: It's somewhat inconsistent that we have separate functions for various pool-related info, such as `staker_balance`, and then also this `pool_data`. I'd probably add specific functions that you need (`pool_decommission_destination`?) instead of this common one.

common/src/chain/transaction/partially_signed_transaction.rs
```diff
@@ -17,24 +17,35 @@ use super::{
...
 }
 
 impl PartiallySignedTransaction {
     pub fn new(
         tx: Transaction,
         witnesses: Vec<Option<InputWitness>>,
-        input_utxos: Vec<Option<TxOutput>>,
+        input_utxos: Vec<Option<(TxOutput, UtxoAdditionalInfo)>>,
         destinations: Vec<Option<Destination>>,
     ) -> Result<Self, TransactionCreationError> {
```
ImplOfAnImpl, line 50: This is indeed rather hacky. Can we put this info somewhere beside `PartiallySignedTransaction` instead?

chainstate/src/rpc/types/output.rs
```diff
@@ -55,6 +56,36 @@ impl RpcOutputValue {
     }
 }
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, rpc_description::HasValueHint)]
+pub struct RpcPoolData {
```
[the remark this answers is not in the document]
OBorce: Removed

chainstate/src/rpc/mod.rs
OBorce: Replaced with get pool decommission destination

wallet/wallet-node-client/Cargo.toml
```diff
@@ -17,6 +17,7 @@ logging = { path = "../../logging" }
 mempool = { path = "../../mempool" }
 node-lib = { path = "../../node-lib" }
 p2p = { path = "../../p2p" }
+pos-accounting = { path = "../../pos-accounting" }
```
azarovh, line 20: I guess this can be removed?

common/src/chain/transaction/partially_signed_transaction.rs
azarovh: I agree that the way it's done right now is hacky. 
Maybe moving `PartiallySignedTransaction` from `wallet` to `common` was not that good idea after all. I did that few months ago because the struct looked really basic and a good companion for `SignedTransaction`,  also I thought I'd use it in tx-verifier. But maybe I misunderstood it and the type is from wallet layer. Thoughts?

azarovh: I agree that the way it's done right now is hacky.
Maybe moving `PartiallySignedTransaction` from `wallet` to `common` was not that good idea after all. I did that few months ago because the struct looked really basic and a good companion for `SignedTransaction`, also I thought I'd use it in `tx-verifier`. But maybe I misunderstood it and the type is from wallet layer. Thoughts?

wallet/src/send_request/mod.rs
```diff
@@ -279,17 +287,72 @@ impl SendRequest {
...
+                utxo.map(|utxo| {
+                    let additional_info = match &utxo {
+                        TxOutput::Burn(value)
+                        | TxOutput::Htlc(value, _)
+                        | TxOutput::Transfer(value, _)
+                        | TxOutput::LockThenTransfer(value, _, _) => {
+                            find_additional_info(value, additional_info)?
+                        }
+                        TxOutput::AnyoneCanTake(data) => {
+                            find_additional_info(data.as_ref().ask(), additional_info)?
```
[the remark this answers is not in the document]
azarovh: It is spendable by anyone who will provide "asked" currency in exchange. 

P.S. I this `data.as_ref().ask()` is wrong and it should be `give()` which is the value that creator of the output locked with it.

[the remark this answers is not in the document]
azarovh: Also this value can be a token so I think it does make sense

common/src/chain/tokens/tokens_utils.rs
```diff
@@ -87,3 +87,23 @@ pub fn is_token_or_nft_issuance(output: &TxOutput) -> bool {
...
+        }
+        TxOutput::CreateStakePool(_, _)
+        | TxOutput::ProduceBlockFromStake(_, _)
+        | TxOutput::CreateDelegationId(_, _)
+        | TxOutput::DelegateStaking(_, _)
+        | TxOutput::DataDeposit(_)
+        | TxOutput::Htlc(_, _)
+        | TxOutput::IssueFungibleToken(_)
+        | TxOutput::AnyoneCanTake(_) => None,
+        TxOutput::IssueNft(_, _, _) => make_token_id(inputs),
```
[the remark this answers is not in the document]
OBorce: IssueFungibleToken is ignored as it can't be spent as a utxo, this was added to fix the utxo functions, because they were not updated when TokenV1 was added, so they were only handeling the TokensV0 and NFT.

wallet/src/send_request/mod.rs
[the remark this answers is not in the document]
azarovh: Sorry guys I confused everyone here. `AnyoneCanTake` output is not included into utxo set (but rather creates an account) and is not spendable.

common/src/chain/tokens/tokens_utils.rs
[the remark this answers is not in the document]
azarovh: Returning token id only for spendable output is a very specific logic, common crate might not be the best place for it. Why not moving it to the wallet and renaming to `get_token_id_for_spendable_utxo`?

wallet/wallet-controller/src/helpers.rs
```diff
@@ -0,0 +1,276 @@
...
+        | TxOutput::AnyoneCanTake(_)
+        | TxOutput::IssueNft(_, _, _)
+        | TxOutput::IssueFungibleToken(_)
+        | TxOutput::DelegateStaking(_, _)
+        | TxOutput::CreateDelegationId(_, _)
+        | TxOutput::DataDeposit(_) => None,
+    }
+}
+
+pub async fn fetch_utxo_exra_info<T>(
```
azarovh, line 134: type: extra

azarovh (commented):
I have small additional request: `is_v0_token_output` should check that values inside `OrderData` of `AnyoneCanTake` output does not contain tokens v0.

wallet/types/src/partially_signed_transaction.rs
```diff
@@ -13,50 +13,100 @@
...
+    }
+}
+
 #[derive(Debug, Eq, PartialEq, Clone, Encode, Decode)]
 pub struct PartiallySignedTransaction {
     tx: Transaction,
     witnesses: Vec<Option<InputWitness>>,
 
-    input_utxos: Vec<Option<TxOutput>>,
+    input_utxos: Vec<Option<UtxoWithAdditionalInfo>>,
```
azarovh, line 65: why not storing utxo separately from info and zipping if necessary? Right now the data is duplicated. Also it looks cleaner with all the data inside ptx flatten. If such new type is required for trezor I'd move it closer to it.

OBorce: I went with this to preserve the invariant that you can't have additional info if you don't have an UTXO as well. Also the none case of the optional for the additional info does not mean missing it means no additional info. Maybe I should go back to the enum instead of the option + enum?

azarovh: Now I get it. I'd leave a comment about that invariant and new field just to make it clear for the future.

common/src/chain/tokens/tokens_utils.rs
```diff
@@ -87,3 +87,23 @@ pub fn is_token_or_nft_issuance(output: &TxOutput) -> bool {
...
+                OutputValue::Coin(_) | OutputValue::TokenV0(_) => None,
+                OutputValue::TokenV1(token_id, _) => Some(*token_id),
+            }
+        }
+        TxOutput::CreateStakePool(_, _)
+        | TxOutput::ProduceBlockFromStake(_, _)
+        | TxOutput::CreateDelegationId(_, _)
+        | TxOutput::DelegateStaking(_, _)
+        | TxOutput::DataDeposit(_)
+        | TxOutput::Htlc(_, _)
```
azarovh, line 104: Htlc can also contain tokens. Given that the function is in common I'd cover that as well.

wallet/src/send_request/mod.rs
```diff
@@ -279,17 +287,91 @@ impl SendRequest {
...
+                    let additional_info = find_additional_info(&utxo, additional_info)?;
+                    Ok(UtxoWithAdditionalInfo::new(utxo, additional_info))
+                })
+                .transpose()
+            })
+            .collect::<WalletResult<Vec<_>>>()?;
+        let output_additional_infos = self
+            .outputs
+            .iter()
+            .map(|utxo| find_additional_info(utxo, additional_info))
```
azarovh, line 310: So yeah this should be None or even an error because it's not spendable and cannot be a utxo.

common/src/chain/tokens/tokens_utils.rs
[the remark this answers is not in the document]
OBorce: removed the token ids being returned for the utxos, and added a separate call to fetch them at the places where they are needed to be checked for frozen tokens.

common/src/chain/transaction/partially_signed_transaction.rs
```diff
@@ -17,24 +17,50 @@ use super::{
...
+};
 use serialization::{Decode, Encode};
 use utils::ensure;
 
+/// Additional info for UTXOs
+#[derive(Debug, Eq, PartialEq, Clone, Encode, Decode)]
+pub enum UtxoAdditionalInfo {
+    TokenInfo { num_decimals: u8, ticker: Vec<u8> },
+    PoolInfo { staker_balance: Amount },
+    NoAdditionalInfo,
```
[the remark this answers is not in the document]
OBorce: done

wallet/src/send_request/mod.rs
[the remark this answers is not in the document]
OBorce: this is still present as we not only need additional info for UTXOs but also for the output UTXOs

PR #1802, opened 2024-08-27
Testnet htlc activation

azarovh:
- fork for htlc activation on testnet (approximately on 11th of September at 10AM UTC)
- lowered data deposit fee to 20 ML
- increased data deposit max size to 384 bytes
- minor docs improvement for AnyonyCanTake


ImplOfAnImpl (approved):

common/src/chain/config/mod.rs
```diff
@@ -570,13 +569,23 @@ impl ChainConfig {
     }
 
     /// The maximum allowed size for data deposited in DataDeposit output
-    pub fn data_deposit_max_size(&self) -> usize {
-        self.data_deposit_max_size
+    pub fn data_deposit_max_size(&self, height: BlockHeight) -> usize {
+        // Change of the size has nothing to do with htlc, they just come in the same upgrade height
+        self.data_deposit_max_size.unwrap_or_else(|| {
+            match self.chainstate_upgrades.version_at_height(height).1.htlc_activated() {
```
OBorce, line 575: Maybe we can add a separate `deposit_data_version` method and avoid the need for the comment, and will simplify any future changes for deposit data.

OBorce (approved):

common/src/chain/config/mod.rs
azarovh: done

PR #1807, opened 2024-09-04
Mempool rpc events

azarovh:
Allows to subscribe to mempool event over websocket

### DIFF
--- a/mempool/src/interface/mempool_interface.rs
+++ b/mempool/src/interface/mempool_interface.rs
@@ -72,8 +72,11 @@ pub trait MempoolInterface: Send + Sync {
         packing_strategy: PackingStrategy,
     ) -> Result<Option<Box<dyn TransactionAccumulator>>, BlockConstructionError>;
 
-    /// Subscribe to events emitted by mempool
-    fn subscribe_to_events(&mut self, handler: Arc<dyn Fn(MempoolEvent) + Send + Sync>);
+    /// Subscribe to events emitted by mempool subsystem
+    fn subscribe_to_subsystem_events(&mut self, handler: Arc<dyn Fn(MempoolEvent) + Send + Sync>);
+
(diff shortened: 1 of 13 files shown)


mempool/src/interface/mempool_interface.rs
```diff
@@ -72,8 +72,11 @@ pub trait MempoolInterface: Send + Sync {
         packing_strategy: PackingStrategy,
     ) -> Result<Option<Box<dyn TransactionAccumulator>>, BlockConstructionError>;
 
-    /// Subscribe to events emitted by mempool
-    fn subscribe_to_events(&mut self, handler: Arc<dyn Fn(MempoolEvent) + Send + Sync>);
+    /// Subscribe to events emitted by mempool subsystem
+    fn subscribe_to_subsystem_events(&mut self, handler: Arc<dyn Fn(MempoolEvent) + Send + Sync>);
+
+    /// Subscribe to broadcast mempool events
+    fn subscribe_to_rpc_events(&mut self) -> utils_networking::broadcaster::Receiver<MempoolEvent>;
```
ImplOfAnImpl, line 79: I don't think the naming is adequate here, because these 2 functions subscribe to exactly the same events, only the mechanism differ.

The names of the corresponsing methods of `Mempool` - `subscribe_to_events` and `subscribe_to_event_broadcast` - sound better.

(I'd still rename the latter though, e.g. to subscribe_to_event**s**_via_broadcaster or something like that)

ImplOfAnImpl (approved):

OBorce (approved):

mempool/src/interface/mempool_interface.rs
azarovh: Not sure. This is literally reverting changes from #1501. Though I agree that these events aren't subsystem's or rpc's. `subscribe_to_events_subsystem`?

PR #1810, opened 2024-09-10
Fix GH CI workflow

erubboli:
# Fix Github CI workflow

This pull request implements comprehensive CI workflows for building and packaging releases across multiple platforms. The enhancements include:

## Release Types
For each platform, we now produce two distinct release packages:
1. **NodeGUI Package**: Contains the Mintlayer Node GUI application.
2. **Tools Package**: Includes Mintlayer Node and command line utilities.

### macOS
- NodeGUI release:
  - Signed and packaged as a `.dmg` file for easy installation.
- Tools package:
  - Provided as a `.zip` archive.

### Linux
Three package formats are now available:
- `.zip`: For universal compatibility.
- `.deb`: For Debian-based distributions (Ubuntu, Debian, etc.).
- `.rpm`: For Red Hat-based distributions (Fedora, CentOS, etc.).

### Windows
- NodeGUI release:
  - Packaged as a `.zip` file.
  - Also available as an installer (`.exe`).
- Tools package:
  - Provided as a `.zip` archive.

### Docker
- Docker images are built and pushed to our repository.

## Additional Notes
- All builds are automated through our CI pipeline, ensuring consistency across releases.
- Each artifact is accompanied by a SHA256 hash for verification purposes.
- Release notes are automatically generated, listing all included files and their respective hashes.


.github/workflows/build-docker.yml
```diff
@@ -0,0 +1,47 @@
+name: Build Docker
```
ImplOfAnImpl, line 1: I'd suggest to put the release-related `build-xxx` files into a separate `release` subfolder.

build-tools/osx/entitlements.plist
```diff
@@ -0,0 +1,51 @@
...
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <false/>
+    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
+    <false/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+    <false/>
+
+    <!-- Debugging (remove for production) -->
+    <key>com.apple.security.get-task-allow</key>
+    <false/>
```
ImplOfAnImpl, line 41: Should it be removed?

build-tools/osx/sign_and_notarize.sh
```diff
@@ -0,0 +1,231 @@
+#!/bin/bash
+
+set -e
```
ImplOfAnImpl, line 3: I'd also specify `set -o nounset`. Though it'll require refactoring the script a bit.

build-tools/osx/sign_and_notarize.sh
```diff
@@ -0,0 +1,231 @@
...
+    security delete-keychain "$RUNNER_TEMP/$KEYCHAIN_NAME"
+    rm "$RUNNER_TEMP/build_certificate.p12"
+    if [ -z "${RUNNER_TEMP_PRESERVE}" ]; then
+        rm -rf "$RUNNER_TEMP"
+    fi
+}
+
+# Main execution
+main() {
+    if [ $# -eq 0 ]; then
```
ImplOfAnImpl, line 198: `-ne 2` ?

ImplOfAnImpl (approved):

azarovh (approved):

.github/workflows/build-linux.yml
```diff
@@ -0,0 +1,214 @@
...
+        mv rpm-node/RPMS/x86_64/mintlayer-node-${{ steps.get_version.outputs.VERSION }}-1.x86_64.rpm Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_x86_64.rpm
+
+    - name: Package Mintlayer Node (without GUI) as ZIP
+      run: |
+        mkdir -p Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_${{ matrix.arch }}
+        IFS=',' read -ra BINARIES <<< "${{ inputs.binary_list }}"
+        for binary in "${BINARIES[@]}"; do
+          cp target/${{ matrix.arch }}-unknown-linux-gnu/release/$binary Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_${{ matrix.arch }}/
+        done
+        zip -r Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_${{ matrix.arch }}.zip Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_${{ matrix.arch }}
```
azarovh, line 182: btw why zip? Previously we used tar.xz and it looks more linux-friendly

.github/workflows/release.yml
```diff
@@ -1,260 +1,103 @@
...
-# CI that:
-#
-# * checks for a Git Tag that looks like a release
-# * builds artifacts with cargo-dist (archives, installers, hashes)
-# * uploads those artifacts to temporary workflow zip
-# * on success, uploads the artifacts to a Github Release
-#
-# Note that the Github Release will be created with a generated
-# title/body based on your changelogs.
+name: Build, Sign, and Release
```
azarovh, line 1: This is now builds and signs artefacts for every PR. It should be limited to release branches I guess

.github/workflows/build-linux.yml
```diff
@@ -0,0 +1,214 @@
...
+        rpmbuild -bb --define "_topdir $(pwd)/rpm-node" --buildroot $(pwd)/rpm-node/BUILDROOT/mintlayer-node-${{ steps.get_version.outputs.VERSION }}-1.x86_64 rpm-node/SPECS/mintlayer-node.spec
+        mv rpm-node/RPMS/x86_64/mintlayer-node-${{ steps.get_version.outputs.VERSION }}-1.x86_64.rpm Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_x86_64.rpm
+
+    - name: Package Mintlayer Node (without GUI) as ZIP
+      run: |
+        mkdir -p Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_${{ matrix.arch }}
+        IFS=',' read -ra BINARIES <<< "${{ inputs.binary_list }}"
+        for binary in "${BINARIES[@]}"; do
+          cp target/${{ matrix.arch }}-unknown-linux-gnu/release/$binary Mintlayer_Node_linux_${{ steps.get_version.outputs.VERSION }}_${{ matrix.arch }}/
+        done
```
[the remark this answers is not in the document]
ImplOfAnImpl: > The zip archive and packages have inconsistent naming for arm64 - the packages have the suffix "arm64" and the archive "aarch64". Not a big deal, but can be confusing for users.

Same thing for x86_64 actually (the suffixes are x86_64 and amd64)

.github/workflows/build-linux.yml
```diff
@@ -0,0 +1,214 @@
...
+on:
+  workflow_call:
+    inputs:
+      binary_list:
+        required: true
+        type: string
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
ImplOfAnImpl, line 12: I don't think we should use the latest image here, because it effectively determines the minimum version of glibc required by the binaries. E.g. I've checked [your test binaries](https://github.com/erubboli/mintlayer-core-test-build/releases/tag/v0.99.15) - according to `objdump`, `node-daemon` needs at least one symbol from glibc v2.34, while `node-gui` needs a symbol from v2.35.
Glibc v2.35 is the one used by Ubuntu 22.04, which is what "ubuntu-latest" [currently means on GitHub Actions](https://github.com/actions/runner-images).  But they'll update to 24.04 eventually and our next release may suddenly require even a newer version of a Linux distro.
For comparison, Debian 11 (which, as I've googled, will be supported until 08.2026), has glibc v2.31, so the binaries won't work there.
And RHEL 8 (supported until 2029) has glibc v2.28.

So at the very least I'd suggest to specify a concrete Linux distro here.
GitHub actions should still support ubuntu-20.04 (which uses glibc v2.31), we can use this one, so that the binaries work on Debian 11 at least.

An alternative solution might be to use one of the "manylinux" docker images, which are used to compile Python extensions. I've found [this article](https://kobzol.github.io/rust/ci/2021/05/07/building-rust-binaries-in-ci-that-work-with-older-glibc.html) where the author has successfully used one of the manylinux images with GitHub actions. 

---
Or we may want to switch to using musl-libc targets, so that the binaries don't depend on glibc at all. Though those targets are still considered ["Tier 2"](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-with-host-tools) so I'm not sure if it's a good idea (in any case, it's not a good idea to switch to musl in the current release).

P.S. We could also provide musl-based binaries in addition to glibc-based ones, so that people on older systems could have an alternative.

.github/workflows/build-docker.yml
erubboli: like the idea but got this error from gh:
```
invalid value workflow reference: workflows must be defined at the top level of the .github/workflows/ directory
```

.github/workflows/build-linux.yml
```diff
@@ -0,0 +1,214 @@
...
+        cp build-tools/assets/node-gui-icon_512.png debian-gui/usr/share/icons/hicolor/512x512/apps/mintlayer-node-gui.png
+        convert build-tools/assets/node-gui-icon_512.png -resize 256x256 debian-gui/usr/share/icons/hicolor/256x256/apps/mintlayer-node-gui.png
+        convert build-tools/assets/node-gui-icon_512.png -resize 128x128 debian-gui/usr/share/icons/hicolor/128x128/apps/mintlayer-node-gui.png
+        convert build-tools/assets/node-gui-icon_512.png -resize 64x64 debian-gui/usr/share/icons/hicolor/64x64/apps/mintlayer-node-gui.png
+        
+        # Create .desktop file
+        cat << EOF > debian-gui/usr/share/applications/mintlayer-node-gui.desktop
+        [Desktop Entry]
+        Name=Mintlayer Node GUI
+        Exec=/usr/bin/node-gui
```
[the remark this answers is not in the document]
erubboli: I agree with you; binaries should have a prefix to avoid conflicts when installed system-wide. However, implementing this change would break any existing scripts. It's a delicate matter, and I'm not sure we can make this change easily now.

.github/workflows/build-docker.yml
ImplOfAnImpl: > like the idea but got this error from gh

Then I'd suggest to rename the files at least, giving them the "release-" prefix, so that it's clear what those jobs are for.

.github/workflows/build-linux.yml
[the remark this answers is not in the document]
ImplOfAnImpl: > I agree with you; binaries should have a prefix to avoid conflicts when installed system-wide. However, implementing this change would break any existing scripts. It's a delicate matter, and I'm not sure we can make this change easily now.

I wonder whether we should publish the deb/rpm at all in such a case. At least for the command-line tools. Because not only those binaries have generic names, their `--help` output doesn't even mention they are related to Mintlayer. So users may not have any idea what has just been installed into their system.

PR #1818, opened 2024-09-20
add data deposit fee to wasm functions

OBorce:
- normalise parameter types for nft and fungible token functions

### DIFF
--- a/wasm-wrappers/src/lib.rs
+++ b/wasm-wrappers/src/lib.rs
@@ -676,8 +676,8 @@ pub fn token_change_authority_fee(current_block_height: u64, network: Network) -
 #[wasm_bindgen]
 pub fn encode_output_issue_fungible_token(
     authority: &str,
-    token_ticker: &[u8],
-    metadata_uri: &[u8],
+    token_ticker: &str,
+    metadata_uri: &str,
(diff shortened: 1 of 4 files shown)


azarovh (approved):

wasm-wrappers/src/lib.rs
```diff
@@ -718,10 +736,10 @@ pub fn encode_output_issue_nft(
...
     description: &str,
     media_hash: &[u8],
-    creator: Option<String>,
-    media_uri: Option<Vec<u8>>,
-    icon_uri: Option<Vec<u8>>,
-    additional_metadata_uri: Option<Vec<u8>>,
+    creator: Option<Vec<u8>>,
+    media_uri: Option<String>,
+    icon_uri: Option<String>,
+    additional_metadata_uri: Option<String>,
```
ImplOfAnImpl, line 742: Why aren't these `Option<&str>` and `Option<&[u8]>`?

ImplOfAnImpl (approved):

wasm-wrappers/src/lib.rs
OBorce: I think the wasm bindings generator doesn't know how to handle it.

PR #1823, opened 2024-10-04
Support orders in the wallet

azarovh:


wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -676,6 +677,44 @@ trait WalletRpc {
...
+    /// Optionally output address can be provided where the exchanged funds from the order are transferred.
+    #[method(name = "fill_order")]
+    async fn fill_order(
+        &self,
+        account: AccountArg,
+        order_id: RpcAddress<OrderId>,
+        fill_amount: RpcAmountIn,
+        output_address: Option<RpcAddress<Destination>>,
+        options: TransactionOptions,
+    ) -> rpc::RpcResult<NewTransaction>;
```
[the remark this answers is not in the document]
azarovh: went with `fill_amount_in_ask_currency `

common/src/chain/order/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+
+mod rpc;
+pub use rpc::RpcOrderInfo;
+
+/// Order data provides unified data structure to represent an order.
+/// There are no buy or sell types of orders per se but rather exchanges.
+/// The fields represent currencies and amounts to be exchanged and the trading pair can be deducted from it.
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, serde::Serialize, serde::Deserialize)]
+pub struct OrderData {
+    /// The key that can authorize conclusion of an order
```
OBorce, line 31: I think it would be good to explain here what concluding an order will result into, for anyone reading this.

wallet/src/account/mod.rs
```diff
@@ -2189,67 +2366,118 @@ fn group_preselected_inputs(
                     | TxOutput::CreateDelegationId(_, _)
                     | TxOutput::DelegateStaking(_, _)
                     | TxOutput::IssueFungibleToken(_)
-                    | TxOutput::DataDeposit(_) => {
+                    | TxOutput::DataDeposit(_)
+                    | TxOutput::AnyoneCanTake(_) => {
```
OBorce, line 2370: Would CreateOrder be a better name than AnyoneCanTake?

wallet/src/account/mod.rs
```diff
@@ -2132,7 +2308,8 @@ fn group_preselected_inputs(
     chain_config: &ChainConfig,
     block_height: BlockHeight,
     dest_info_provider: Option<&dyn DestinationInfoProvider>,
-) -> Result<BTreeMap<currency_grouper::Currency, (Amount, Amount)>, WalletError> {
+    order_info: Option<&RpcOrderInfo>,
```
OBorce, line 2311: This should be a map <OrderId, &RpcOrderInfo> just in case to know that the order info is for the order from the inputs.

OBorce (approved):

wallet/src/account/mod.rs
azarovh: Originally it was CrateOrder but after some discussions with Sam it was renamed to AnyonCanTake

ImplOfAnImpl: >> Would CreateOrder be a better name than AnyoneCanTake?

> Originally it was CrateOrder but after some discussions with Sam it was renamed to AnyonCanTake

Maybe we should reconsider the name? `AnyoneCanTake` is way too cryptic IMO.

PR #1829, opened 2024-10-25
Fix spending htlc output in api server

azarovh:
Was causing `MissingOutputOrSpent` because Htlc outputs were ignored and not added to the db.

Also enabled htlc for api server's simulation.

### DIFF
--- a/api-server/scanner-lib/src/blockchain_state/mod.rs
+++ b/api-server/scanner-lib/src/blockchain_state/mod.rs
@@ -1957,11 +1957,11 @@ fn get_tx_output_destination(txo: &TxOutput) -> Option<&Destination> {
         | TxOutput::IssueNft(_, _, d)
         | TxOutput::ProduceBlockFromStake(d, _) => Some(d),
         TxOutput::CreateStakePool(_, data) => Some(data.decommission_key()),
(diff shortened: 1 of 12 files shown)


ImplOfAnImpl (approved):

OBorce (commented):
I think the address amounts in the db are not being updated when creating and spending the htlc

azarovh:
> I think the address amounts in the db are not being updated when creating and spending the htlc

Yes, the idea is that amount from htlc is not accounted in the balances, because it's not actually yours unless atomic swap is complete as it can be reverted by refund.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -1932,7 +1961,7 @@ fn get_tx_output_destination(txo: &TxOutput) -> Option<&Destination> {
         | TxOutput::Burn(_)
         | TxOutput::DelegateStaking(_, _)
         | TxOutput::DataDeposit(_)
+        | TxOutput::Htlc(_, _)
```
azarovh, line 1964: this is wrong, will push a fix shortly

azarovh: done

OBorce (approved):
looks ok

PR #1831, opened 2024-11-01
use Vec::with_capacity avoid realloc memory

nkysg:


azarovh (approved):
Thanks

ImplOfAnImpl (commented):
Regarding the changes in the production code - those are useful, thanks.

Regarding those in test code:
1) We usually don't care that much about performance in tests. So e.g. if there is a choice between clarity and performance, clarity wins.
2) You've duplicated some magic numbers in your calls to `Vec::with_capacity`. E.g.
```
let mut vector = Vec::with_capacity(2048);
for _ in 0..1024 {..
```
This is not super nice, because it makes it possible for the numbers to fall out of sync next time someone modifies this code.
So, ideally here you should have figured out what that `1024` actually means, introduce a name for it and use it in both places.
```
let that_thing_count = 1024;
let mut vector = Vec::with_capacity(that_thing_count * 2);
for _ in 0..that_thing_count {..
```
But it's extra work and if you choose a bad name for it, the clarity of the test will suffer. So I'd just revert those changes.

api-server/stack-test-suite/tests/v2/token_ids.rs
```diff
@@ -107,12 +107,13 @@ async fn ok(#[case] seed: Seed) {
                     + (chain_config.nft_issuance_fee(BlockHeight::new(1)) * 10).unwrap())
                 .unwrap();
 
-                let mut token_ids = vec![];
+                const ID_NUMS: usize = 10;
```
ImplOfAnImpl, line 110: "tokens count" is a better name for this.
Also, I'd just make it a variable.
```
let tokens_count = 10;
```

api-server/stack-test-suite/tests/v2/address_delegations.rs
```diff
@@ -152,6 +152,7 @@ async fn ok(#[case] seed: Seed) {
                 let mut blocks = vec![];
                 blocks.push(block.clone());
                 for delegation in &delegations {
+                    blocks.reserve_exact(delegation.3.len());
```
ImplOfAnImpl, line 155: Btw, IMO `reserve_exact` is redundant here, just use `reserve`

api-server/stack-test-suite/tests/v2/pool.rs
```diff
@@ -145,6 +145,7 @@ async fn ok(#[case] seed: Seed) {
                 for pool in &pools {
                     blocks.push(pool.3.clone());
                     for delegation in &pool.2 {
+                        blocks.reserve_exact(delegation.3.len());
```
ImplOfAnImpl, line 148: same

api-server/stack-test-suite/tests/v2/token_ids.rs
```diff
@@ -139,8 +140,9 @@ async fn ok(#[case] seed: Seed) {
                         .unwrap();
                 }
 
-                let mut nft_ids = vec![];
-                for _ in 0..10 {
+                const ID_NUMS: usize = 10;
```
ImplOfAnImpl, line 143: Same.
```
let tokens_count = 10;
```

serialization/core/tests/containers.rs
```diff
@@ -48,8 +48,9 @@ fn test_scale_vectors() {
     let dec = SimpleWrapper::decode_all(&mut &enc[..]).ok();
     assert!(&dec.is_some());
     assert_eq!(dec, Some(SimpleWrapper(vector)));
-    let mut vector = Vec::new();
-    for _ in 0..1024 {
+    const NUM: usize = 1024;
```
ImplOfAnImpl, line 51:
```
let iter_count = 1024;
```

ImplOfAnImpl (approved):

ImplOfAnImpl:
I've merged changes in the production code and the test framework in #1840.
The changes in the tests themselves are redundant so I dropped them.

PR #1834, opened 2024-11-08
Fix missing owning block for transaction in api server

azarovh:


api-server/stack-test-suite/tests/v2/transaction_merkle_path.rs
```diff
@@ -39,142 +34,6 @@ async fn get_transaction_failed() {
     task.abort();
 }
 
-#[rstest]
-#[trace]
-#[case(Seed::from_entropy())]
-#[tokio::test]
-async fn get_block_failed(#[case] seed: Seed) {
```
OBorce, line 46: This test got me thinking, I guess initially the idea was that on reorg blocks are deleted from the mainchain blocks table and transactions owning block is set to NULL. And the test should pass if we set the transaction with a null owning block.

But I guess when I was refactoring the tables instead of 2 tables (one for blocks and one for mainchain blocks) now joined into one, this is no longer true, because only the block height is set to NULL and the owning block reference is still valid even though it is not a mainchain block.
So not sure what should we do, should we set it to NULL when a block is no longer on the mainchain, or set the column to not null as it will always point to a block.

OBorce (approved):

api-server/stack-test-suite/tests/v2/transaction_merkle_path.rs
azarovh: I think making it not null is ok

azarovh: I'd go with not null option. Not sure why would we need transactions without a block

OBorce (approved):

PR #1838, opened 2024-11-14
Feature/trezor common

OBorce:
Move the simplified no-std types used by Trezor firmware to keep it in sync with the core types.
Add randomized tests to check the encodings match between the types.


Cargo.toml
```diff
@@ -60,6 +60,7 @@ members = [
   "test",                               # Integration tests.
   "test-rpc-functions",                 # RPC functions specifically for tests.
   "test-utils",                         # Various utilities for tests.
+  "trezor-common",                      # Code used by Trezor firmware repository.
```
azarovh, line 63: Would `wallet` folder be a better place for it?

trezor-common/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+num-derive.workspace = true
+num-traits = { version = "0.2", default-features = false, features = ["libm"] }
+parity-scale-codec = { version = "3.1", default-features = false, features = ["derive", "chain-error"] }
+
+[dev-dependencies]
+test-utils = { path = "../test-utils" }
+common = { path = "../common" }
+crypto = { path = "../crypto/" }
+
+bitcoin-bech32.workspace = true
```
azarovh, line 20: Looks like only rstest dependency is actually used

azarovh (approved):

trezor-common/Cargo.toml
OBorce: removed

Cargo.toml
OBorce: moved

wallet/trezor-common/src/tests/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+        ),
+        2 => chain::Destination::ScriptHash(primitives::H256(rng.gen()).into()),
+        3 => chain::Destination::ClassicMultisig(
+            (&crypto::key::PrivateKey::new_from_rng(rng, KeyKind::Secp256k1Schnorr).1).into(),
+        ),
+        _ => chain::Destination::PublicKeyHash(
+            (&crypto::key::PrivateKey::new_from_rng(rng, KeyKind::Secp256k1Schnorr).1).into(),
+        ),
+    }
+}
```
ImplOfAnImpl, line 328: I suggest deriving `strum::EnumDiscriminants` for the existing enum types, like I did in [1832](https://github.com/mintlayer/mintlayer-core/pull/1832) for `Destination`:
```
#[derive(..., EnumDiscriminants)]
#[strum_discriminants(name(DestinationTag), derive(EnumIter))]
pub enum Destination {
...
```
This will define an additional enum called `DestinationTag`, which will be like `Destination` but without the data. And this `DestinationTag` will implement `EnumIter`, so it'll be possible to iterate over its values.
Then you'd be able to write:
```
let dest_tag = DestinationTag::iter().choose(rng).unwrap();
match dest_tag {
    DestinationTag::PublicKey => {
        chain::Destination::PublicKey(
            crypto::key::PrivateKey::new_from_rng(rng, KeyKind::Secp256k1Schnorr).1,
        )
    },
    ...
}
```
So, if/when a new variant is added to the enum, we'll get a compilation failure here.

wallet/trezor-common/src/tests/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+    fn from(value: chain::OutPointSourceId) -> Self {
+        match value {
+            chain::OutPointSourceId::Transaction(tx) => Self::Transaction(tx.to_hash().into()),
+            chain::OutPointSourceId::BlockReward(tx) => Self::BlockReward(tx.to_hash().into()),
+        }
+    }
+}
+
+impl From<chain::UtxoOutPoint> for crate::UtxoOutPoint {
+    fn from(value: chain::UtxoOutPoint) -> Self {
```
ImplOfAnImpl, line 40: Though I don't insist on this, I'd suggest to make the `make_random...` functions return a pair of values that are supposed to be encoded to the same bytes, and remove these `From` implementations (which look somewhat fragile to me).

wallet/trezor-common/src/tests/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+
+    let inp = make_random_input(&mut rng);
+
+    let simple_inp: crate::TxInput = inp.clone().into();
+
+    assert_eq!(inp.encode(), simple_inp.encode());
+
+    let decoded_simple_inp =
+        chain::TxInput::decode_all(&mut simple_inp.encode().as_slice()).unwrap();
+    assert_eq!(decoded_simple_inp, inp);
```
ImplOfAnImpl, line 539: It's better to perform the checks multiple times (e.g. 100 or 1000) in a loop, to increase the chance of failure if things get broken.
Same for the other test.

ImplOfAnImpl (approved):

wallet/trezor-common/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    Transaction(H256),
+    #[codec(index = 1)]
+    BlockReward(H256),
+}
+
+#[derive(FromPrimitive)]
+pub enum OutPointSourceIdIndex {
+    Transaction = 0,
+    BlockReward = 1,
+}
```
ImplOfAnImpl, line 293: As I can see, this is similar to AccountCommandIndex, but the difference is that there is already MintlayerUtxoType in the protobuf config which this enum should be consistent with. Remove it and use MintlayerUtxoType instead.

wallet/trezor-common/src/tests/mod.rs
OBorce: I think it is the same, because you would still need to put the same random data in the right fields for both versions when generating them.

trezor-common/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub enum OutputValue {
+    Coin(Amount),
+    TokenV0,
+    TokenV1(H256, Amount),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, EnumDiscriminants)]
+#[strum_discriminants(name(OutputTimeLockIndex), derive(EnumIter, FromPrimitive))]
```
ImplOfAnImpl, line 80: As I can see, `OutputTimeLockIndex` has nothing to do with `OutputTimeLock`, same as it was for `AccountCommandIndex` and `OutPointSourceIdIndex`. But it must be conmsistent with `lock_type` returned by `get_lock` in the trezor repo. So it makes sense to do the same for it as well - move it to rust code in the trezo repo, add same enum to the .proto file, write tests that ensures their consistency.

P.S. probably these extra protobuf enums should go to a separate file, because they have nothing to do with the protocol and only serve as a way to have the same constants in python and rust.

trezor-common/src/tests/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+        ),
+    }
+}
+
+fn make_random_bytes(rng: &mut (impl Rng + CryptoRng)) -> Vec<u8> {
+    (0..rng.gen_range(1..10)).map(|_| rng.gen()).collect()
+}
+
+fn make_random_account_command(rng: &mut (impl Rng + CryptoRng)) -> chain::AccountCommand {
+    match crate::AccountCommandIndex::iter().choose(rng).unwrap() {
```
ImplOfAnImpl, line 338: Btw, the idea was that you add Tag enum to `chain::AccountCommand` so that if it gets more variants, we get a compilation failure here. Same for other "XXXIndex" enums.

Also, I suggest calling them "XXXTag", because "index" is a bit confusing IMO. I.e. we also have those `#[codec(index = n)]` annotations on the "parent" enums, which may not be consecutive, and "XXXIndex" sounds like at has something to do with those indices. And "XXXTag" sounds like it's just a name.

OBorce: It is hard to re-export the AccountCommandTag from the original as it would bring a dependency on that crate, and it is not nostd compliant. As for the handling of new variants it will be caught by the From implementations, they do a match on the original enums.

I replaced Index with Type as it sounds better to me, AccountCommandType than Tag?

ImplOfAnImpl: > It is hard to re-export the AccountCommandTag from the original as it would bring a dependency on that crate, and it is not nostd compliant.

But it's for tests only, which can use std, right? I mean, you already use `chain::DestinationTag`, which also comes from a crate that is not nostd compliant.

> As for the handling of new variants it will be caught by the From implementations, they do a match on the original enums.

Yeah, I guess it will work. But IMO it'd still be a bit nicer if the match was over Tag types (or whatever they will be called) that come with the type that is being generated. Otherwise it's not obvious that the test actually generates all possible "chain::AccountCommand"s

> I replaced Index with Type as it sounds better to me, AccountCommandType than Tag?

Doesn't Type sound ambiguous?
In any case, we should probably use the same suffix for all such discriminant types. I.e. if "tag" is not good enough, `chain::DestinationTag` should be renamed too (I can do it myself or you can do it).
I'm not sure Type is a good alternative though. "Discriminant" might be better, but it's also longer.

PR #1843, opened 2024-11-25
Move node gui backend to a crate

azarovh:
1. Move `node-gui/backend` module to a crate so that it can be reused by external UI clients.
2. Use `HexEncoded<SignedTransaction>` in the backend messages to avoid supporting serialisation for the common type.


node-gui/backend/src/messages.rs
```diff
@@ -164,7 +162,7 @@ pub struct SendDelegateToAddressRequest {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TransactionInfo {
     pub wallet_id: WalletId,
-    pub tx: SignedTransaction,
+    pub tx: HexEncoded<SignedTransaction>,
 }
```
ImplOfAnImpl, line 166: I don't really like when `HexEncoded` is used like that. IMO its usage is only justified if you use it ad-hoc, like e.g. when you need to return a `SignedTransaction` from a function and you need the return type to be serializable, you can return `HexEncoded<SignedTransaction>` instead. But making it a field of a struct looks excessive and it is very intrusive. I suggest that we use serde(with) instead.

For this you would define a module somewhere in `serialization`:
```
pub mod hex_encoded_serialization {
    use serde::Deserialize as _;
    use serialization::{
        hex::{HexDecode, HexEncode},
        Decode, Encode,
    };

    pub fn serialize<T: Encode, S: serde::Serializer>(t: &T, serializer: S) -> Result<S::Ok, S::Error> {
        let hex = t.hex_encode();
        serializer.serialize_str(&hex)
    }

    pub fn deserialize<'de, T: Decode, D: serde::Deserializer<'de>>(
        deserializer: D,
    ) -> Result<T, D::Error> {
        let hex = String::deserialize(deserializer)?;
        let value = T::hex_decode_all(hex).map_err(serde::de::Error::custom)?;
        Ok(value)
    }
}
```
And then you can write:
```
#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
pub struct TransactionInfo {
    pub wallet_id: WalletId,
    #[serde(with = "hex_encoded_serialization")]
    pub tx: SignedTransaction,
}
```

ImplOfAnImpl (approved):

node-gui/backend/src/messages.rs
azarovh: done

OBorce (approved):

node-gui/backend/src/messages.rs
```diff
@@ -62,7 +62,23 @@ impl AccountId {
...
+impl Serialize for AccountId {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        self.0.into_u32().serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for AccountId {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let i = u32::deserialize(deserializer)?;
+        let i = U31::from_u32(i).ok_or(D::Error::custom(format!(
```
ImplOfAnImpl, line 74: 1) `ok_or_else` to avoid constructing the error string when there is no error.

2) it's better to have a test for this.

ImplOfAnImpl (approved):

OBorce (approved):

PR #1851, opened 2024-12-10
Fix erroneous peer discouragement after mempool reorg

ImplOfAnImpl:
Currently there is a possibility for the node to erroneously discourage some of its peers when a transaction that is in the mempool gets included in a block. Such a transaction will fail to validate and p2p will be notified about it; if the corresponding "mempool_ban_score" is non-zero, p2p may discourage the peer that was the first one to send it.

The solution is not to send a `TransactionProcessed` event if a transaction fails during a mempool reorg.


mempool/src/pool/mod.rs
```diff
@@ -326,6 +335,7 @@ struct TxFinalizer<'a> {
     cur_time: Time,
     events_broadcast: &'a mut EventsBroadcast,
     work_queue: &'a mut WorkQueue,
+    called_during_reorg: bool,
```
azarovh, line 338: seems a bit hacky. Bool is hard to read from the outside and easy to miss if modifications required, also information that finalizer is used in a specific situation (reorg) is external. Would passing `Option<&mut EventsBroadcast>` or even `enum EventsMode {Silent, Broadcast(&mut EventsBroadcast)}` be more robust approach here?

azarovh (approved):

PR #1855, opened 2024-12-18
Minor gui backend fixes

azarovh:
1. Added `Transaction` type to backend, which is a wrapper around `SignedTransaction` that allows serialization/deserialization to hex and also methods to convert to text summary or json. Mainly to give UI flexibility and avoid deserializing `SignedTransaction`.
2. Added Serialize trait to some common types.

### DIFF
--- a/node-gui/backend/src/error.rs
+++ b/node-gui/backend/src/error.rs
@@ -21,6 +21,8 @@ pub enum BackendError {
     WalletError(String),
     #[error("RPC error: {0}")]
     RpcError(String),
+    #[error("Json error: {0}")]
+    JsonError(String),
     #[error("Unknown wallet index: {0:?}")]
     UnknownWalletIndex(WalletId),
     #[error("Unknown account index: {0:?}/{0:?}")]
--- a/node-gui/backend/src/error.rs
+++ b/node-gui/backend/src/error.rs
@@ -19,10 +19,8 @@ use super::{account_id::AccountId, messages::WalletId};
 pub enum BackendError {
     #[error("Wallet error: {0}")]
     WalletError(String),
-    #[error("RPC error: {0}")]
-    RpcError(String),
-    #[error("Json error: {0}")]
-    JsonError(String),
+    #[error("Convertion to dehexify json error: {0}")]
+    ConversionToDehexifyJsonError(String),
     #[error("Unknown wallet index: {0:?}")]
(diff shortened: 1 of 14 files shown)


node-gui/backend/src/messages.rs
```diff
@@ -149,11 +150,35 @@ pub struct SendDelegateToAddressRequest {
...
+    }
+
+    pub fn text_summary(&self, config: &ChainConfig) -> String {
+        self.tx.transaction().text_summary(config)
+    }
+
+    pub fn to_json(&self, config: &ChainConfig) -> Result<serde_json::Value, BackendError> {
+        common::address::dehexify::to_dehexified_json(config, self.tx.transaction())
+            .map_err(|e| BackendError::JsonError(e.to_string()))
+    }
```
ImplOfAnImpl, line 175: I kind of missed the fact that `Transaction` is serializable. So now I'm wondering why `SignedTransaction` shouldn't be

node-gui/backend/src/error.rs
```diff
@@ -21,6 +21,8 @@ pub enum BackendError {
     WalletError(String),
     #[error("RPC error: {0}")]
     RpcError(String),
+    #[error("Json error: {0}")]
+    JsonError(String),
```
ImplOfAnImpl, line 25: `JsonError` looks too generic, given the context where it's produced.
I'd suggest introducing a more specific error variant that somehow mentions `to_dehexified_json`.

Btw, `RpcError` above seems to be unused.

node-gui/backend/src/messages.rs
```diff
@@ -149,11 +150,35 @@ pub struct SendDelegateToAddressRequest {
     pub delegation_id: String,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Transaction {
```
ImplOfAnImpl, line 154: Doesn't `Transaction` look confusing? Perhaps `SignedTransaction` would be better?

ImplOfAnImpl (approved):

OBorce (approved):

node-gui/backend/src/messages.rs
azarovh: I don't know, my guess would be that we don't have a use case for serializing witnesses

PR #1856, opened 2024-12-19
Enable logging to a file

azarovh:
For GUI logging to a file is enabled by default. For other apps it's off with an option to enable it explicitly.


azarovh:
I should probably mention why I refactored the functions inside runner.rs
Mainly to separate initialization of folders and logging from actually running the node and to be able to log command line options before everything else. Also now it also seems like a better structure overall then before.

node-lib/src/runner.rs
```diff
@@ -48,6 +49,7 @@ use crate::{
 };
 
 const LOCK_FILE_NAME: &str = ".lock";
+const LOG_FILE_NAME: &str = "mintlayer.log";
```
ImplOfAnImpl, line 52: Perhaps the name should be based on the executable?

node-lib/src/options.rs
```diff
@@ -80,6 +89,11 @@ pub struct RunOptions {
     #[clap(long, short, action = clap::ArgAction::SetTrue)]
     pub clean_data: Option<bool>,
 
+    /// Log to a file
+    #[clap(long, action = clap::ArgAction::Set)]
+    #[arg(hide = true)]
+    pub log_to_file: Option<bool>,
```
ImplOfAnImpl, line 95: Should we hide it?

(If not, the description should be more detailed, mentioning that the default value depends on the application and where the logs will be located).

node-lib/src/runner.rs
```diff
@@ -233,46 +235,65 @@ async fn initialize(
...
+        )?;
+        return Ok(NodeSetupResult::DataDirCleanedUp);
     }
+
+    // Init logging
+    if run_options.log_to_file.is_some_and(|log_to_file| log_to_file) {
+        let log_file = FileRotate::new(
+            data_dir.join(format!("logs/{}", LOG_FILE_NAME)),
+            AppendCount::new(2),              // total 3 file
+            ContentLimit::Bytes(100_000_000), // 100MB each
```
ImplOfAnImpl, line 272: 100 mb seems too much for one file.

IMO we should measure roughly how much logs the node produces in a day and make it the max size. And the number of files can be e.g. 7 (or 10, or 14).

ImplOfAnImpl (approved):

node-lib/src/options.rs
azarovh: but it's already hidden

node-lib/src/runner.rs
azarovh: Changed to 14x10MB

node-lib/src/options.rs
ImplOfAnImpl: > but it's already hidden

Yeah, but I'm wondering whether it should be, because it's potentially useful for node daemon too

azarovh: ok maybe you're right, removed hidden

OBorce (approved):

PR #1864, opened 2025-01-15
Wallet handle txs with conflicting account nonces

azarovh:
Unconfirmed txs in the wallet which uses outdated nonce are marked as conflicted now if confirmed tx is encountered on mainchain and the state of OutputCache is updated accordingly. 

The behaviour is similar to abandoning tx.

Besides tests I had a wallet with conflicting orders which I couldn't open previously because of `InconsistentOrderDuplicateNonce` which is now resolved.

### DIFF
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -2083,7 +2083,8 @@ impl<K: AccountKeyChains> Account<K> {
                         let tx_state =
                             TxState::Confirmed(block_height, block.timestamp(), idx as u64);
                         let wallet_tx = WalletTx::Tx(TxData::new(signed_tx.clone(), tx_state));
-                        self.update_conflicting_txs(&wallet_tx, block, db_tx)?;
+
+                        self.update_conflicting_txs(signed_tx.transaction(), block, db_tx)?;
 
                         new_tx_was_added |= self
                             .add_wallet_tx_if_relevant_and_remove_from_user_txs(
@@ -2111,15 +2112,17 @@ impl<K: AccountKeyChains> Account<K> {
     /// Check for any conflicting txs and update the new state in the DB
     fn update_conflicting_txs<B: storage::Backend>(
         &mut self,
-        wallet_tx: &WalletTx,
+        confirmed_tx: &Transaction,
         block: &Block,
         db_tx: &mut StoreTxRw<B>,
     ) -> WalletResult<()> {
         let acc_id = self.get_account_id();
-        let conflicting_tx = self.output_cache.check_conflicting(wallet_tx, block.get_id().into());
-        for tx in conflicting_tx {
-            let id = AccountWalletTxId::new(acc_id.clone(), tx.id());
-            db_tx.set_transaction(&id, tx)?;
+        let conflicting_txs =
+            self.output_cache.update_conflicting_txs(confirmed_tx, block.get_id().into())?;
+
+        for tx_id in conflicting_txs {
+            let id = AccountWalletCreatedTxId::new(acc_id.clone(), tx_id);
+            db_tx.del_user_transaction(&id)?;
         }
 
         Ok(())
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -1953,16 +1953,10 @@ impl<K: AccountKeyChains> Account<K> {
             .txs_with_unconfirmed()
             .iter()
             .filter_map(|(id, tx)| match tx.state() {
-                TxState::Confirmed(height, _, idx) => {
-                    if height > common_block_height {
-                        Some((
-                            AccountWalletTxId::new(self.get_account_id(), id.clone()),
-                            (height, idx),
-                        ))
-                    } else {
-                        None
-                    }
-                }
+                TxState::Confirmed(height, _, idx) => (height > common_block_height).then_some((
+                    AccountWalletTxId::new(self.get_account_id(), id.clone()),
+                    (height, idx),
+                )),
                 TxState::Inactive(_)
                 | TxState::Conflicted(_)
                 | TxState::InMempool(_)
@@ -1976,7 +1970,7 @@ impl<K: AccountKeyChains> Account<K> {
         for (tx_id, _) in revoked_txs {
             db_tx.del_transaction(&tx_id)?;
             let source = tx_id.into_item_id();
-            self.output_cache.remove_tx(&source)?;
+            self.output_cache.remove_confirmed_tx(&source)?;
             wallet_events.del_transaction(self.account_index(), source);
         }
 
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -2120,9 +2120,9 @@ impl<K: AccountKeyChains> Account<K> {
         let conflicting_txs =
             self.output_cache.update_conflicting_txs(confirmed_tx, block.get_id().into())?;
 
-        for tx_id in conflicting_txs {
-            let id = AccountWalletCreatedTxId::new(acc_id.clone(), tx_id);
-            db_tx.del_user_transaction(&id)?;
+        for (tx_id, tx) in conflicting_txs {
+            db_tx.set_transaction(&AccountWalletTxId::new(acc_id.clone(), tx.id()), &tx)?;
+            db_tx.del_user_transaction(&AccountWalletCreatedTxId::new(acc_id.clone(), tx_id))?;
         }
 
         Ok(())
@@ -2282,9 +2282,9 @@ impl<K: AccountKeyChains> Account<K> {
         let abandoned_txs = self.output_cache.abandon_transaction(tx_id)?;
         let acc_id = self.get_account_id();
 
-        for tx_id in abandoned_txs {
-            let id = AccountWalletCreatedTxId::new(acc_id.clone(), tx_id);
-            db_tx.del_user_transaction(&id)?;
+        for (tx_id, tx) in abandoned_txs {
+            db_tx.set_transaction(&AccountWalletTxId::new(acc_id.clone(), tx.id()), &tx)?;
+            db_tx.del_user_transaction(&AccountWalletCreatedTxId::new(acc_id.clone(), tx_id))?;
         }
 
         Ok(())
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -35,7 +35,6 @@ use common::size_estimation::{
 use common::Uint256;
 use crypto::key::hdkd::child_number::ChildNumber;
 use mempool::FeeRate;
-use output_cache::OrderData;
 use serialization::hex_encoded::HexEncoded;
 use utils::ensure;
 pub use utxo_selector::UtxoSelectorError;
@@ -89,7 +88,8 @@ use wallet_types::{
 };
 
 pub use self::output_cache::{
-    DelegationData, OwnFungibleTokenInfo, PoolData, TxInfo, UnconfirmedTokenInfo, UtxoWithTxOutput,
+    DelegationData, OrderData, OwnFungibleTokenInfo, PoolData, TxInfo, UnconfirmedTokenInfo,
+    UtxoWithTxOutput,
 };
 use self::output_cache::{OutputCache, TokenIssuanceData};
 use self::transaction_list::{get_transaction_list, TransactionList};
@@ -885,6 +885,12 @@ impl<K: AccountKeyChains> Account<K> {
             .token_data(token_id)
             .filter(|data| self.is_destination_mine(&data.authority))
             .ok_or(WalletError::UnknownTokenId(*token_id))
+    }
+
+    pub fn get_orders(&self) -> impl Iterator<Item = (&OrderId, &OrderData)> {
+        self.output_cache
+            .orders()
+            .filter(|(_, data)| self.is_destination_mine(&data.conclude_key))
     }
 
     pub fn find_order(&self, order_id: &OrderId) -> WalletResult<&OrderData> {
--- a/wallet/src/account/output_cache/mod.rs
+++ b/wallet/src/account/output_cache/mod.rs
@@ -815,8 +815,40 @@ impl OutputCache {
         let mut conflicting_txs_with_descendants = vec![];
 
         for conflicting_tx in conflicting_txs {
-            let descendants =
-                self.remove_descendants_and_mark_as(conflicting_tx, TxState::Conflicted(block_id))?;
+            let descendants = self.remove_descendants(conflicting_tx);
+
+            // Mark conflicting tx and its descendants as Conflicting and update OutputCache data accordingly
+            for tx_id in descendants.iter().rev().copied() {
+                match self.txs.entry(tx_id.into()) {
+                    Entry::Occupied(mut entry) => match entry.get_mut() {
+                        WalletTx::Block(_) => Err(WalletError::CannotFindTransactionWithId(tx_id)),
+                        WalletTx::Tx(tx) => match tx.state() {
+                            TxState::Inactive(_) => {
+                                tx.set_state(TxState::Conflicted(block_id));
+                                OutputCache::rollback_tx_data(
+                                    tx,
+                                    &self.unconfirmed_descendants,
+                                    &mut self.consumed,
+                                    &mut self.delegations,
+                                    &mut self.token_issuance,
+                                    &mut self.orders,
+                                );
+                                Ok(())
+                            }
+                            TxState::Abandoned
+                            | TxState::Confirmed(..)
+                            | TxState::InMempool(..)
+                            | TxState::Conflicted(..) => {
+                                Err(WalletError::CannotChangeTransactionState(
+                                    *tx.state(),
+                                    TxState::Conflicted(block_id),
+                                ))
+                            }
+                        },
+                    },
+                    Entry::Vacant(_) => Err(WalletError::CannotFindTransactionWithId(tx_id)),
+                }?;
+            }
 
             conflicting_txs_with_descendants.extend(descendants.into_iter());
         }
@@ -1492,11 +1524,7 @@ impl OutputCache {
         })
     }
 
-    fn remove_descendants_and_mark_as(
-        &mut self,
-        tx_id: Id<Transaction>,
-        new_state: TxState,
-    ) -> WalletResult<Vec<Id<Transaction>>> {
+    fn remove_descendants(&mut self, tx_id: Id<Transaction>) -> Vec<Id<Transaction>> {
         let mut all_txs = Vec::new();
         let mut to_update = BTreeSet::from_iter([OutPointSourceId::from(tx_id)]);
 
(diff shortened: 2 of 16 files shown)


wallet/src/account/output_cache/mod.rs
```diff
@@ -1692,3 +1747,20 @@ fn apply_total_supply_mutations_from_tx(
...
+        TxInput::Utxo(_) => true,
+        TxInput::AccountCommand(nonce, cmd) => {
+            confirmed_account_type == cmd.into() && *nonce <= confirmed_nonce
+        }
+        TxInput::Account(outpoint) => {
+            confirmed_account_type == outpoint.account().into()
+                && outpoint.nonce() <= confirmed_nonce
+        }
+    })
+}
```
OBorce, line 1766: I am a bit confused on this, shouldn't it be any instead of all, and false for Utxo?
If all inputs are Utxos this will return that it is a conflicting nonce tx?

wallet/src/wallet/tests.rs
```diff
@@ -5924,4 +5924,363 @@ fn create_order_fill_partially_conclude(#[case] seed: Seed) {
         );
     }
 }
-// create order, fill partially, conclude
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn conflicting_delegation_account_nonce(#[case] seed: Seed) {
```
OBorce, line 5931: Not sure this test is correct, if I understand correctly,
the spend_from_delegation_tx_2 depends on spend_from_delegation_tx_1,
so when spend_from_delegation_tx_1 is confirmed spend_from_delegation_tx_2 should still be valid not conflicting.

To be conflicting you need to create 3 txs, X1, and X2 that have nonce 0, and than add X2 as unconfirmed, and create Y that depend on X2 with nonce 1. Then confirm X1 which will conflict with X2 and Y.

wallet/src/account/output_cache/mod.rs
azarovh: you are right thanks

wallet/src/account/output_cache/mod.rs
```diff
@@ -680,56 +684,86 @@ impl OutputCache {
...
+                        | AccountCommand::ChangeTokenAuthority(_, _)
+                        | AccountCommand::UnfreezeToken(_)
+                        | AccountCommand::ConcludeOrder(_)
+                        | AccountCommand::FillOrder(_, _, _) => { /* do nothing */ }
+                        | AccountCommand::FreezeToken(token_id, _) => {
+                            frozen_token_id = Some(*token_id);
+                        }
+                    }
+                }
+            }
```
OBorce, line 719: If only one Account or AccountCommand can be present in the tx, maybe we can replace it with a find() that returns the 2 Options.

wallet/src/account/output_cache/mod.rs
```diff
@@ -680,56 +684,86 @@ impl OutputCache {
...
+                        if let WalletTx::Tx(tx) = unconfirmed_tx {
+                            conflicting_txs.push(tx.get_transaction().get_id());
                         }
-                        WalletTx::Block(_) => {}
-                    };
+                    }
+                }
 
-                    conflicting_txs.push(unconfirmed);
+                if let Some((confirmed_account, confirmed_account_nonce)) = confirmed_account_nonce
```
OBorce, line 736: this can probably be an else if, as the transactions that use a token id will be a superset of the transactions that use the token id and have a specific nonce.

wallet/src/account/output_cache/mod.rs
```diff
@@ -784,16 +818,22 @@ impl OutputCache {
...
             TxState::Inactive(_)
             | TxState::InMempool(_)
             | TxState::Conflicted(_)
             | TxState::Abandoned => true,
             TxState::Confirmed(_, _, _) => false,
         };
+
         if is_unconfirmed && !already_present {
             self.unconfirmed_descendants.insert(tx_id.clone(), BTreeSet::new());
+        } else {
```
OBorce, line 835: this should be `else if !is_unconfirmed`, we don't want to remove it if it is unconfirmed and already present.

wallet/src/account/output_cache/mod.rs
```diff
@@ -908,13 +948,14 @@ impl OutputCache {
...
-                    } else {
-                        self.unconfirmed_descendants.remove(tx_id);
+                    if let Some(descendants) =
+                        self.unconfirmed_descendants.get_mut(&outpoint.source_id())
+                    {
+                        if is_unconfirmed {
+                            descendants.insert(tx_id.clone());
+                        } else {
+                            descendants.remove(tx_id);
+                        }
```
OBorce, line 958: It doesn't make sense if the tx is confirmed for it to have a utxo pointing to an unconfirmed tx. Maybe error out in the else instead of descentents.remove ?

wallet/src/account/output_cache/mod.rs
azarovh: I think that it'd make the code more confusing especially when modifying in the future. I replaced the vec with set to avoid duplicates.

wallet/src/account/output_cache/mod.rs
```diff
@@ -680,56 +684,92 @@ impl OutputCache {
...
-                | AccountCommand::ChangeTokenAuthority(_, _)
-                | AccountCommand::UnfreezeToken(_)
-                | AccountCommand::ConcludeOrder(_)
-                | AccountCommand::FillOrder(_, _, _) => None,
-                AccountCommand::FreezeToken(frozen_token_id, _) => Some(frozen_token_id),
-            },
+        let conflict = confirmed_tx.inputs().iter().find_map(|input| {
+            match input {
+                TxInput::Utxo(_) => {
+                    //TODO: check conflicting utxo spends
```
ImplOfAnImpl, line 700: Is there any reason for not implementing it right now?

ImplOfAnImpl (approved):

wallet/src/account/output_cache/mod.rs
```diff
@@ -1364,30 +1411,29 @@ impl OutputCache {
...
+        for tx_id in all_txs.iter().rev().copied() {
             match self.txs.entry(tx_id.into()) {
                 Entry::Occupied(mut entry) => match entry.get_mut() {
                     WalletTx::Block(_) => Err(WalletError::CannotFindTransactionWithId(tx_id)),
                     WalletTx::Tx(tx) => match tx.state() {
-                        TxState::Inactive(_) | TxState::Conflicted(_) => {
-                            tx.set_state(TxState::Abandoned);
+                        TxState::Inactive(_) | TxState::Conflicted(_) | TxState::Abandoned => {
+                            tx.set_state(new_state);
                             for input in tx.get_transaction().inputs() {
```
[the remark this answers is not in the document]
azarovh: done, added an error for setting the same state and also checks to the tests. Also added abandoning a conflicting tx

wallet/src/account/output_cache/mod.rs
azarovh: Just wanted to focus on accounts because it already took me longer than I expected

azarovh:
Fixed the problem when abandoning a tx that is already marked as conflicted leads to invalid state of internal fields of OutputCache (like account nonces).

ImplOfAnImpl (approved):

wallet/src/account/output_cache/mod.rs
```diff
@@ -680,56 +684,122 @@ impl OutputCache {
...
+                            }
+                            TxState::Abandoned
+                            | TxState::Confirmed(..)
+                            | TxState::InMempool(..)
+                            | TxState::Conflicted(..) => {
+                                Err(WalletError::CannotChangeTransactionState(
+                                    *tx.state(),
+                                    TxState::Conflicted(block_id),
+                                ))
+                            }
```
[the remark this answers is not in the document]
azarovh: Looks like `InMempool` -> `Conflicted` is theoretically possible. Changed the logic to mark it as conlflicted.

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,56 +736,129 @@ impl OutputCache {
...
-            .map(|tx_id| self.txs.get(tx_id).expect("must be present"))
-            .collect_vec()
+        // Remove all descendants of conflicting txs
+        let mut conflicting_txs_with_descendants = vec![];
+
+        for conflicting_tx in conflicting_txs {
+            let txs_to_rollback = self.remove_from_unconfirmed_descendants(conflicting_tx);
+
+            // Mark conflicting tx and its descendants as Conflicting and update OutputCache data accordingly
+            for tx_id in txs_to_rollback.iter().rev().copied() {
```
OBorce, line 818: There is a problem that currently I think cannot happen but can happen in the future when we add TXs from Mempool.

The problem is with the rollback_tx_data, as the TXs are iterated in random order by tx_id, in a situation like:
Having unconfirmed txs for a delegation or token:
A(nonce 1) -> B(2) -> C(3) -> D(4)
If for some reason the conflict happens in the middle of the chain e.g. nonce 2 (currently it should always happen at the start of the chain from confirmed blocks, unless I am mistaken)
So if it happens at 2, all 2, 3, and 4 will be inside conflicting_txs, but not sorted by tx_id. So the rollback can happen to finish at 3 or 4, which will be unable to find last_parent, the tx A as a still unconfirmed tx for this delegation or token.

OBorce: Thinking about it maybe you can have filter them out in n * log(d).
something like:
conflicting_descendents = BTreeSet::new();
for tx in conflicting_txs {
  conflicting_descendents += find_descendents(tx);
}
conflicting_txs = conflicting_txs.filter(|tx| conflicting_descendetns.contains(tx))

This should also fix the double Conflict -> Conflict state

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,56 +736,129 @@ impl OutputCache {
...
-                | AccountCommand::ChangeTokenAuthority(_, _)
-                | AccountCommand::UnfreezeToken(_)
-                | AccountCommand::ConcludeOrder(_)
-                | AccountCommand::FillOrder(_, _, _) => None,
-                AccountCommand::FreezeToken(frozen_token_id, _) => Some(frozen_token_id),
-            },
+        let conflict = confirmed_tx.inputs().iter().find_map(|input| {
+            match input {
+                TxInput::Utxo(_) => {
+                    //TODO: check conflicting utxo spends
```
ImplOfAnImpl, line 752: Plz mention the issue that you've created for this

wallet/src/account/output_cache/mod.rs
```diff
@@ -960,13 +1043,14 @@ impl OutputCache {
...
+                        self.unconfirmed_descendants.get_mut(&outpoint.source_id())
+                    {
+                        ensure!(
+                            is_unconfirmed,
+                            WalletError::ConfirmedTxAmongUnconfirmedDescendants(tx_id.clone())
+                        );
+                        descendants.insert(tx_id.clone());
                     }
                 }
                 TxInput::Account(outpoint) => match outpoint.account() {
```
ImplOfAnImpl, line 1056: To me, both the old and the new version seem wrong.
The old one seems wrong because of `else { self.unconfirmed_descendants.remove(tx_id); }` - though `tx` itself is confirmed, it can still have unconfirmed descendants and the `remove` drops them all for no reason.
The new version adds `tx` as an unconmfirmed descendant to its parent only if the parent already has some unconfirmed descendants. Also, the `ensure!` seems to prohibit `update_inputs` from being called on a confirmed tx, if one of its parents has unconfirmed descendants.

I guess the correct way of going about this is to first write 3 tests: the 1st one should fail for the old version due to extra unconfirmed descendants being dropped, the 2nd should fail for the new version due to `tx` not being added as an unconfirmed descendant, the 3rd one should fail for the new version as well due to `update_inputs` failing for a confirmed tx.

wallet/src/account/output_cache/mod.rs
```diff
@@ -1401,83 +1485,108 @@ impl OutputCache {
         })
     }
 
-    /// Mark a transaction and its descendants as abandoned
-    /// Returns a Vec of the transaction Ids that have been abandoned
-    pub fn abandon_transaction(
+    // Removes a tx from unconfirmed descendant.
+    // Returns provided tx and all the descendants.
+    fn remove_from_unconfirmed_descendants(
```
ImplOfAnImpl, line 1490: The other thing that its caller depends on is the fact that children always come after the parents in the result. It makes sense to mention this as well.

wallet/src/account/output_cache/mod.rs
azarovh: Am I getting it right that you suggest filter out conflicting tx to be able to rollback descendants first and then conflicting txs themself?

wallet/src/account/output_cache/mod.rs
azarovh: > The new version adds tx as an unconmfirmed descendant to its parent only if the parent already has some unconfirmed descendants. 

Not exactly. It adds tx to descendants only if parent is unconfirmed (it might not have any descendants yet)

> Also, the ensure! seems to prohibit update_inputs from being called on a confirmed tx, if one of its parents has unconfirmed descendants.

It's not possible to confirm a tx while having unconfirmed parent. Thus the check.

ImplOfAnImpl: > > The new version adds tx as an unconmfirmed descendant to its parent only if the parent already has some unconfirmed descendants.
> 
> Not exactly. It adds tx to descendants only if parent is unconfirmed (it might not have any descendants yet)
> 
> > Also, the ensure! seems to prohibit update_inputs from being called on a confirmed tx, if one of its parents has unconfirmed descendants.
> 
> It's not possible to confirm a tx while having unconfirmed parent. Thus the check.

I guess you are right. I was assuming that `unconfirmed_descendants` may contain unconfirmed descendants of confirmed txs too, but it looks like this is not the case, at least not anymore.

Some comments would be nice to have though. In particular, it's better to document all invariants.
Also, it may make sense to add some debug-only checks that the invariants hold and invoke them in all public functions.

wallet/src/account/output_cache/mod.rs
OBorce: txs_to_rollback  is already sorted with the parent first and all descendants after it, and the loop is in reverse.
only the conflicting_txs are not sorted, so if we filter them out before this loop, everything will be ok and there will be no duplicate txs and no Conflict -> Conflict state.

azarovh: So I checked the collection and I don't think filtering out parents is necessary as a result of `remove_from_unconfirmed_descendants`  function call doesn't contain any duplicates anymore.

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,56 +736,129 @@ impl OutputCache {
...
-                | AccountCommand::UnmintTokens(_)
-                | AccountCommand::LockTokenSupply(_)
-                | AccountCommand::ChangeTokenMetadataUri(_, _)
-                | AccountCommand::ChangeTokenAuthority(_, _)
-                | AccountCommand::UnfreezeToken(_)
-                | AccountCommand::ConcludeOrder(_)
-                | AccountCommand::FillOrder(_, _, _) => None,
-                AccountCommand::FreezeToken(frozen_token_id, _) => Some(frozen_token_id),
-            },
+        let conflict = confirmed_tx.inputs().iter().find_map(|input| {
```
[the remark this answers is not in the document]
azarovh: done

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,56 +736,129 @@ impl OutputCache {
...
-                if self.uses_token(unconfirmed_tx, frozen_token_id) {
-                    let unconfirmed_tx = self.txs.get_mut(unconfirmed).expect("must be present");
-                    match unconfirmed_tx {
-                        WalletTx::Tx(ref mut tx) => {
-                            tx.set_state(TxState::Conflicted(block_id));
+
+                if let WalletTx::Tx(tx) = unconfirmed_tx {
+                    if let Some(frozen_token_id) = conflict.frozen_token_id {
+                        if self.uses_token(unconfirmed_tx, &frozen_token_id) {
+                            conflicting_txs.insert(tx.get_transaction().get_id());
```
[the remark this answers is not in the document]
azarovh: added `continue`

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,56 +736,129 @@ impl OutputCache {
...
+                                    &mut self.orders,
+                                );
+                                Ok(())
+                            }
+                            TxState::Conflicted(..) => {
+                                // It's possible to try to mark descendant as conflicting twice
+                                // because unconfirmed_descendants contains a tx as child and as parent.
+                                // So it's not an error only if done during this function call.
+                                ensure!(
+                                    conflicting_txs_with_descendants.contains(&tx_id),
```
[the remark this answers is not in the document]
azarovh: not applicable anymore as duplicates were fixed

wallet/src/account/output_cache/mod.rs
```diff
@@ -1401,83 +1485,108 @@ impl OutputCache {
...
+    fn rollback_tx_data(
+        tx: &TxData,
+        unconfirmed_descendants: &BTreeMap<OutPointSourceId, BTreeSet<OutPointSourceId>>,
+        consumed: &mut BTreeMap<UtxoOutPoint, TxState>,
+        delegations: &mut BTreeMap<DelegationId, DelegationData>,
+        token_issuance: &mut BTreeMap<TokenId, TokenIssuanceData>,
+        orders: &mut BTreeMap<OrderId, OrderData>,
+    ) {
+        let tx_id = tx.get_transaction().get_id();
+        for input in tx.get_transaction().inputs() {
```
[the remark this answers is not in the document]
azarovh: done

OBorce (approved):

wallet/src/account/output_cache/mod.rs
```diff
@@ -1401,89 +1437,147 @@ impl OutputCache {
...
+                        if let Some(data) = self.orders.get_mut(order_id) {
+                            data.last_nonce = nonce.decrement();
+                            data.last_parent = find_parent(&self.unconfirmed_descendants, &tx_id);
+                        }
+                    }
+                },
+            }
+        }
+
+        for output in tx.outputs() {
```
ImplOfAnImpl, line 1504: Could be nitpicking, but I'd suggest rolling back stuff in the exact opposit order of how it was added. I.e. first roll back outputs (in the reverse order too) and then inputs.
Currently it doesn't matter of course, but maybe our chain will evolve in the future and make the order matter.

wallet/src/account/mod.rs
```diff
@@ -2011,15 +2012,17 @@ impl<K: AccountKeyChains> Account<K> {
...
-        let conflicting_tx = self.output_cache.check_conflicting(wallet_tx, block.get_id().into());
-        for tx in conflicting_tx {
-            let id = AccountWalletTxId::new(acc_id.clone(), tx.id());
-            db_tx.set_transaction(&id, tx)?;
+        let conflicting_txs =
+            self.output_cache.update_conflicting_txs(confirmed_tx, block.get_id().into())?;
+
+        for tx_id in conflicting_txs {
+            let id = AccountWalletCreatedTxId::new(acc_id.clone(), tx_id);
+            db_tx.del_user_transaction(&id)?;
```
ImplOfAnImpl, line 2025: Probably we need `set_transaction` here as well? Othwerwise the tx in the db will have different status and the wallet may behave differently after a restart.

(Same for abandon_transaction?)

wallet/src/wallet/tests.rs
```diff
@@ -6454,4 +6476,1374 @@ fn create_order_fill_partially_conclude(#[case] seed: Seed) {
...
+        WithLocked::Any,
+    );
+    assert_eq!(
+        token_balance_unconfirmed,
+        (received_tokens_1 + received_tokens_2).unwrap()
+    );
+}
+
+// Create a pool and a delegation with some share.
+// Create tx that spend from delegation account multiple times in a single tx. Add it as unconfirmed.
```
ImplOfAnImpl, line 7255: "that spends"
the "in a single tx" part looks redundant

wallet/src/wallet/tests.rs
```diff
@@ -6454,4 +6476,1374 @@ fn create_order_fill_partially_conclude(#[case] seed: Seed) {
...
+
+    let mut delegations = wallet.get_delegations(DEFAULT_ACCOUNT_INDEX).unwrap().collect_vec();
+    assert_eq!(delegations.len(), 1);
+    let (deleg_id, deleg_data) = delegations.pop().unwrap();
+    assert_eq!(*deleg_id, delegation_id);
+    assert_eq!(deleg_data.last_nonce, Some(AccountNonce::new(0)));
+}
+
+// Create a pool and a delegation with some share.
+// Create tx that spend from delegation account and it as unconfirmed.
```
ImplOfAnImpl, line 7524: "that spends"
"and it as unconfirmed" - broken clause

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,59 +745,131 @@ impl OutputCache {
...
+            }
+
+            conflicting_txs_with_descendants.extend(txs_to_rollback.into_iter());
+        }
+
+        Ok(conflicting_txs_with_descendants)
     }
 
-    fn uses_token(&self, unconfirmed_tx: &WalletTx, frozen_token_id: &TokenId) -> bool {
+    fn violates_frozen_token(&self, unconfirmed_tx: &WalletTx, frozen_token_id: &TokenId) -> bool {
```
ImplOfAnImpl, line 872: Btw, probably this should check tx outputs too?

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,59 +745,131 @@ impl OutputCache {
...
+        for conflict_check in conflict_checks {
             for unconfirmed in self.unconfirmed_descendants.keys() {
                 let unconfirmed_tx = self.txs.get(unconfirmed).expect("must be present");
-                if self.uses_token(unconfirmed_tx, frozen_token_id) {
-                    let unconfirmed_tx = self.txs.get_mut(unconfirmed).expect("must be present");
-                    match unconfirmed_tx {
-                        WalletTx::Tx(ref mut tx) => {
-                            tx.set_state(TxState::Conflicted(block_id));
+
+                if let WalletTx::Tx(tx) = unconfirmed_tx {
```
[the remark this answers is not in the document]
azarovh: Seems like it, added

wallet/src/account/output_cache/mod.rs
```diff
@@ -836,16 +921,22 @@ impl OutputCache {
...
             | TxState::Conflicted(_)
             | TxState::Abandoned => true,
             TxState::Confirmed(_, _, _) => false,
         };
+
         if is_unconfirmed && !already_present {
             self.unconfirmed_descendants.insert(tx_id.clone(), BTreeSet::new());
+        } else if !is_unconfirmed {
+            self.unconfirmed_descendants.remove(&tx_id);
         }
```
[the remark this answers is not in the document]
azarovh: fixed

wallet/src/account/output_cache/mod.rs
```diff
@@ -1401,89 +1437,147 @@ impl OutputCache {
...
+    // Returns provided tx and all the descendants in that specific order.
+    fn remove_from_unconfirmed_descendants(
         &mut self,
         tx_id: Id<Transaction>,
-    ) -> WalletResult<Vec<Id<Transaction>>> {
-        let mut all_abandoned = Vec::new();
-        let mut to_abandon = BTreeSet::from_iter([OutPointSourceId::from(tx_id)]);
+    ) -> Vec<Id<Transaction>> {
+        let mut all_txs = Vec::new();
+        let mut to_update = BTreeSet::from_iter([OutPointSourceId::from(tx_id)]);
```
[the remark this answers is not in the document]
azarovh: Fixed, test added

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,59 +745,131 @@ impl OutputCache {
...
-        conflicting_txs
-            .into_iter()
-            .map(|tx_id| self.txs.get(tx_id).expect("must be present"))
-            .collect_vec()
+        // Remove all descendants of conflicting txs
+        let mut conflicting_txs_with_descendants = vec![];
+
+        for conflicting_tx in conflicting_txs {
+            let txs_to_rollback = self.remove_from_unconfirmed_descendants(conflicting_tx);
+
```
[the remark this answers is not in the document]
azarovh: I believe this is fixed in the previous comment with diamond deps

wallet/src/account/output_cache/mod.rs
azarovh: I changed the order to rev(), but processing outputs before inputs break logic that searches for utxo parent, so I'd leave it this way

wallet/src/account/output_cache/mod.rs
azarovh: Not sure that it's possible to have tokens in the outputs without inputs

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,59 +745,143 @@ impl OutputCache {
...
+                    }
+                };
+            }
+        }
+
+        // Remove all descendants of conflicting txs
+        let mut conflicting_tx_ids_with_descendants = vec![];
+        let mut conflicting_txs_with_descendants = vec![];
+
+        for conflicting_tx in conflicting_txs {
```
ImplOfAnImpl, line 838: [discussed] So, there are still problems here - if in `conflicting_txs` a parent comes before a child, on the 2nd iteration of the loop we'll get a tx that is already marked as conflicted, so we'll hit the `CannotMarkTxAsConflictedIfInState` error. The main question was why the existing tests didn't catch this.

wallet/src/account/output_cache/mod.rs
```diff
@@ -1401,90 +1449,157 @@ impl OutputCache {
...
 
+        all_txs
+    }
+
+    // After tx is removed as a result of reorg, abandoning or marking as conflicted
+    // its effect on OutputCache's data fields should be rolled back
+    fn rollback_tx_data(&mut self, tx: &WalletTx) -> WalletResult<()> {
+        let tx_id = tx.id();
+
+        // Iterate in reverse to handle situations where an account is modified twice in the same tx
```
ImplOfAnImpl, line 1482: [discussed] Mention that this situation is not currently possible due to the restriction in the consensus that only one account command is permitted per tx.

wallet/src/account/output_cache/mod.rs
```diff
@@ -732,59 +745,143 @@ impl OutputCache {
...
             .into_iter()
-            .map(|tx_id| self.txs.get(tx_id).expect("must be present"))
-            .collect_vec()
+            .zip_eq(conflicting_txs_with_descendants)
+            .collect())
     }
 
-    fn uses_token(&self, unconfirmed_tx: &WalletTx, frozen_token_id: &TokenId) -> bool {
+    fn violates_frozen_token(&self, unconfirmed_tx: &WalletTx, frozen_token_id: &TokenId) -> bool {
         unconfirmed_tx.inputs().iter().any(|inp| match inp {
```
ImplOfAnImpl, line 885: [discussed] Mention in a comment that technically we should check the outputs too, but currently it's not possible to have a token in an output without having it in an input as well.

ImplOfAnImpl (approved):

wallet/src/account/output_cache/mod.rs
azarovh: Fixed, test added

wallet/src/account/output_cache/mod.rs
```diff
@@ -1144,76 +1251,17 @@ impl OutputCache {
...
-                    | TxOutput::LockThenTransfer(_, _, _)
-                    | TxOutput::CreateDelegationId(_, _)
-                    | TxOutput::IssueFungibleToken(_)
-                    | TxOutput::Htlc(_, _)
-                    | TxOutput::CreateOrder(_) => {}
-                }
-            }
+    pub fn remove_confirmed_tx(&mut self, tx_id: &OutPointSourceId) -> WalletResult<()> {
+        if let Some(tx) = self.txs.remove(tx_id) {
+            matches!(tx.state(), TxState::Confirmed(..));
```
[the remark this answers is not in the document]
azarovh: yes! thanks

PR #1865, opened 2025-01-15
Refactor/upgrade iced version

OBorce:
- update iced 0.12 -> 0.13
- bumped rfd
- all widget styling are functions now
- removed the use of the deprecated Component trait
- fix the overlapping dialogs bug

There are 2 changes from the previous behavior:
- When creating a new wallet (instead of recover wallet) it generates a new seed phrase for you, if you start editing the seed phrase it will automatically switch from Create to Recover mode. Previously it remained Create. Or maybe we should disable editing of the seed phrase in the Create mode?
- As the Model widget was removed from iced_aw no other widget supports the `on_esc` method so I haven't found a way to dismiss dialogs with esc for now, other than writhing a new low-level widget from scratch.


node-gui/src/main_window/mod.rs
```diff
@@ -307,28 +339,34 @@ impl MainWindow {
...
                 TabsMessage::WalletMessage(wallet_id, WalletMessage::Unlock),
             )) => {
-                self.active_dialog = ActiveDialog::WalletUnlock { wallet_id };
-                Command::none()
+                self.active_dialog = ActiveDialog::WalletUnlock {
+                    wallet_id,
+                    state: Default::default(),
+                };
+                Task::none()
             }
```
ImplOfAnImpl, line 354: Wallet unlocking doesn't work - the input field is disabled:
node_gui_wallet_unlock_disabled

Btw, something happened to the close button on all dialogs:
node_gui_close_button

ImplOfAnImpl (changes requested):
1. Wallet recovery seems broken in this PR - the recovery dialog stays open after the file pick dialog has been closed and nothing happens.

2.
> When creating a new wallet (instead of recover wallet) it generates a new seed phrase for you, if you start editing the seed phrase it will automatically switch from Create to Recover mode. Previously it remained Create. Or maybe we should disable editing of the seed phrase in the Create mode?

IMO this will be very confusing, especially for unexperienced users, and also error-prone.
I'd say let's disable editing in the Create mode.

node-gui/src/main_window/main_widget/tabs/wallet/left_panel.rs
```diff
@@ -156,18 +156,17 @@ pub fn view_left_panel(
...
+                    Text::new(iced_fonts::Bootstrap::Question.to_string())
+                        .font(iced_fonts::BOOTSTRAP_FONT),
                     NEW_ACCOUNT_TOOLTIP_TEXT,
                     tooltip::Position::Bottom
                 )
                 .gap(10)
-                .style(iced::theme::Container::Box)
             ]
-            .align_items(Alignment::Center)
+            .align_y(Alignment::Center)
```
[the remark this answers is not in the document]
OBorce: the pick list had a fixed width so the overlap will happen based on the name of the account if someone inputs a very long name or not. I have removed the fixed width, and the widget will calculate it based on the longest name. But now if there is a very long name the button is squashed.

ImplOfAnImpl (changes requested):
1. I've proposed some changes in https://github.com/mintlayer/mintlayer-core/pull/1872
I suggest merging that PR into this one, so that we can review the whole thing one more time.

2. [discussed] IMO it's better to continue using `Component` even though it's deprecated, because:
a) having the dialogs' state inside the app state looks pretty bad.
b) there is a chance that `Component` will be un-deprecated in the future.

node-gui/src/widgets/wallet_mnemonic.rs
```diff
@@ -96,15 +104,32 @@ impl<Message> Component<Message, Theme, iced::Renderer> for WalletMnemonicDialog
...
             )
-            .foot(container(text("Loading...")).width(Length::Fill).center_x())
+            .foot(container(text("Loading...")).center_x(Length::Fill))
         } else {
-            Card::new(
-                Text::new(action_text),
-                iced::widget::column![text_input("Mnemonic", &mnemonic)
+            let body: Element<_> = if self.generated_mnemonic_opt.is_none() {
+                text_input("Mnemonic", &mnemonic)
+                    // only enable edit if there is not pre-generated mnemonic
```
ImplOfAnImpl, line 111: Looks like the comment is in a wrong place.

node-gui/src/widgets/wallet_mnemonic.rs
```diff
@@ -96,15 +104,32 @@ impl<Message> Component<Message, Theme, iced::Renderer> for WalletMnemonicDialog
                 Text::new(action_text),
                 iced::widget::column![text_input("Mnemonic", &mnemonic).padding(15)],
             )
-            .foot(container(text("Loading...")).width(Length::Fill).center_x())
+            .foot(container(text("Loading...")).center_x(Length::Fill))
         } else {
-            Card::new(
-                Text::new(action_text),
-                iced::widget::column![text_input("Mnemonic", &mnemonic)
+            let body: Element<_> = if self.generated_mnemonic_opt.is_none() {
```
ImplOfAnImpl, line 109: Let's rewrite this whole bunch of `if`s as:
```text
let mut element = if self.generated_mnemonic_opt.is_none() {
    use text input element
} else {
    use the non-editable text
};

if state.importing {
    add `text("Loading...")` as a footer
} else {
    add the button as a footer
}
```
so that the dialog remains the same when `state.importing` becomes true

node-gui/Cargo.toml
```diff
@@ -28,7 +28,8 @@ anyhow.workspace = true
 chrono.workspace = true
 futures.workspace = true
 iced = { workspace = true, features = ["canvas", "debug", "tokio", "lazy"] }
-iced_aw = { workspace = true, features = ["cupertino", "icons"] }
+iced_aw = { workspace = true, features = ["cupertino"] }
```
ImplOfAnImpl, line 31: It seems like we don't need this "cupertino" feature anymore.

wallet/types/src/scan_blockchain.rs
```diff
@@ -0,0 +1,40 @@
...
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub enum ScanBlockchain {
+    /// Skips the scanning of the blokchain, used when creating a wallet from a brand new seed
```
ImplOfAnImpl, line 17: type - blok, same below

wallet/types/src/scan_blockchain.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl ScanBlockchain {
+    pub fn skip_syncing(&self) -> bool {
+        match self {
+            Self::SkipScanning => true,
+            Self::ScanNoWait | Self::ScanAndWait => false,
+        }
+    }
+
+    pub fn wait_for_sync(&self) -> bool {
```
ImplOfAnImpl, line 34: If you rename `ImportOrCreate::scan_blockchain`, it makes sense to rename these too.

ImplOfAnImpl (approved):

azarovh (approved):

PR #1870, opened 2025-01-21
Add missing test for orders with nft

azarovh:
Also use `PerThousand::denominator` function instead of hardcoded 1000


OBorce (approved):
Do we have a test that fills the order with ask_amount -1, I guess they will get 0 Tokens/NFT and the next fill order with 1 will get the Token/NFT?

azarovh:
> Do we have a test that fills the order with ask_amount -1, I guess they will get 0 Tokens/NFT and the next fill order with 1 will get the Token/NFT?

It should be covered by other randomized tests but I added explicit one for nft

ImplOfAnImpl (approved):

PR #1872, opened 2025-01-28
A few fixes in node gui related to the `iced` upgrade 

ImplOfAnImpl:
1) Add `esc_handler`, which creates a wrapper for an `Element` that captures Esc key presses.

2) Use an `opaque` wrapper instead of the separate Button-based element on the stack to prevent mouse events from reaching elements under a dialog.
(the problem with the button was that it would change the mouse cursor to a "finger", which looked strange.)
The `opaque` wrapper that comes with `iced` didn't work well enough though, because it only captures mouse clicks but not hovering (the button-based solution had the same problem btw). So e.g. main window tooltips would still respond to mouse movement when a dialog was opened. So I had to add a custom variant of `opaque` that captures all mouse events.

3) Use semi-transparent white color as the background for the `opaque` element, to make the ui under an opened dialog look blurred, like it was before the `iced` upgrade.

4) Fix a minor issue with the "new mnemonic" dialog, where the "copy to clipboard" button would not be laid out properly.
E.g. before:
new_mnemonic_bad_layout
And after:
new_mnemonic_fixed_layout

5) Add background to the "File dialog..." element.
Before:
file_dialog_old_no_bg
After (here the dialog is also blurred because a file dialog is open):
file_dialog_new_with_bg

6) I also got rid of the useless element `Text::new("Nothing to show")`, which wasn't supposed to be shown anyway.

### DIFF
--- a/node-gui/src/widgets/opaque.rs
+++ b/node-gui/src/widgets/opaque.rs
@@ -1,0 +1,168 @@
+// Copyright (c) 2025 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use iced::{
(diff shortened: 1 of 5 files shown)


node-gui/src/widgets/opaque.rs
```diff
@@ -0,0 +1,172 @@
...
+        &self,
+        state: &Tree,
+        layout: Layout<'_>,
+        cursor: mouse::Cursor,
+        viewport: &Rectangle,
+        renderer: &Renderer,
+    ) -> mouse::Interaction {
+        // self.content
+        //     .as_widget()
+        //     .mouse_interaction(state, layout, cursor, viewport, renderer)
```
ImplOfAnImpl, line 149: Commented code

OBorce (approved):

PR #1874, opened 2025-01-30
Feature/trezor reconnect

OBorce:
If the trezor device returns an USB IO error it might mean that the trezor device has been disconnected, in that case try to reconnect to the device and try the failed operation again.

### DIFF
--- a/wallet/src/signer/trezor_signer/mod.rs
+++ b/wallet/src/signer/trezor_signer/mod.rs
@@ -123,19 +123,92 @@ pub enum TrezorError {
     MultipleSignaturesReturned,
     #[error("A multisig signature was returned for a single address from Device")]
     MultisigSignatureReturned,
+    #[error("The file being loaded is a software wallet and does not correspond to the connected hardware wallet")]
+    HardwareWalletDifferentFile,
+    #[error("PublicKeys missmatch. Wrong device or passphrase:\nfile DeviceId \"{file_device_id}\", connected device \"{connected_device_id}\",\nfile label \"{file_label}\" and connected device label \"{connected_device_id}\"")]
+    HardwareWalletDifferentMnemonicOrPassphrase {
+        file_device_id: String,
+        connected_device_id: String,
+        file_label: String,
+        connected_device_label: String,
+    },
+    #[error("The file being loaded correspond to the connected hardware wallet, but public keys are different. Maybe a wrong passphrase was entered?")]
+    HardwareWalletDifferentPassphrase,
 }
 
 pub struct TrezorSigner {
     chain_config: Arc<ChainConfig>,
     client: Arc<Mutex<Trezor>>,
+    session_id: Vec<u8>,
 }
 
 impl TrezorSigner {
-    pub fn new(chain_config: Arc<ChainConfig>, client: Arc<Mutex<Trezor>>) -> Self {
+    pub fn new(
+        chain_config: Arc<ChainConfig>,
+        client: Arc<Mutex<Trezor>>,
+        session_id: Vec<u8>,
+    ) -> Self {
         Self {
             chain_config,
             client,
+            session_id,
         }
+    }
+
+    /// Calls initialize on the device with the current session_id.
+    ///
+    /// If the operation fails due to an I/O error (which may indicate a lost connection to the device),
+    /// the function will attempt to reconnect to the Trezor device once before returning an error.
+    fn check_session(
+        &mut self,
+        db_tx: &impl WalletStorageReadLocked,
+        key_chain: &impl AccountKeyChains,
+    ) -> SignerResult<()> {
+        let mut client = self.client.lock().expect("poisoned lock");
+
+        match client.init_device(Some(self.session_id.clone())) {
+            Ok(_) => Ok(()),
+            // In case of a USB IO error try to reconnect, and try again
+            Err(trezor_client::Error::TransportSendMessage(
+                trezor_client::transport::error::Error::Usb(rusb::Error::Io),
+            )) => {
+                let (mut new_client, data, session_id) = find_trezor_device()?;
+
+                check_public_keys_against_key_chain(
+                    db_tx,
+                    &mut new_client,
+                    &data,
+                    key_chain,
+                    &self.chain_config,
+                )?;
+
+                *client = new_client;
+                self.session_id = session_id;
+                Ok(())
+            }
+            Err(err) => Err(SignerError::TrezorError(TrezorError::DeviceError(
+                err.to_string(),
+            ))),
+        }
+    }
+
+    /// Attempts to perform an operation on the Trezor client.
+    ///
+    /// If the operation fails due to an I/O error (which may indicate a lost connection to the device),
+    /// the function will attempt to reconnect to the Trezor device once before returning an error.
+    fn perform_trezor_operation<F, R>(
+        &mut self,
+        operation: F,
+        db_tx: &impl WalletStorageReadLocked,
+        key_chain: &impl AccountKeyChains,
+    ) -> SignerResult<R>
+    where
+        F: Fn(&mut Trezor) -> Result<R, trezor_client::Error>,
+    {
+        self.check_session(db_tx, key_chain)?;
+
+        let mut client = self.client.lock().expect("poisoned lock");
+        operation(&mut client).map_err(|e| TrezorError::DeviceError(e.to_string()).into())
     }
 
     fn make_signature<F>(
@@ -282,7 +355,7 @@ impl Signer for TrezorSigner {
         &mut self,
         ptx: PartiallySignedTransaction,
         key_chain: &impl AccountKeyChains,
-        _db_tx: &impl WalletStorageReadUnlocked,
+        db_tx: &impl WalletStorageReadUnlocked,
     ) -> SignerResult<(
         PartiallySignedTransaction,
         Vec<SignatureStatus>,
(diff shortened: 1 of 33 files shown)


wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -0,0 +1,1420 @@
...
+        db_tx: &impl WalletStorageReadLocked,
+    ) -> WalletResult<Self> {
+        let (client, data) = find_trezor_device().map_err(SignerError::TrezorError)?;
+
+        let provider = Self {
+            client: Arc::new(Mutex::new(client)),
+            data,
+        };
+
+        check_public_keys_against_db_ones(db_tx, &provider, chain_config)?;
```
azarovh, line 1226: Imo `_ones` suffix is confusing. I'd just remove it

azarovh (approved):
Needs a rebase

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -131,6 +137,50 @@ impl TrezorSigner {
...
+        let mut client = self.client.lock().expect("poisoned lock");
+        let res = operation(&mut client);
+
+        match res {
+            Ok(res) => Ok(res),
+            // In case of a USB IO error try to reconnect, and try again
+            Err(trezor_client::Error::TransportSendMessage(
+                trezor_client::transport::error::Error::Usb(rusb::Error::Io),
+            )) => {
+                let (mut new_client, data) = find_trezor_device()?;
```
ImplOfAnImpl, line 162: Calling `find_trezor_device` on each operation is not a great idea. What if another device is returned?

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -131,6 +137,50 @@ impl TrezorSigner {
...
+    {
+        let mut client = self.client.lock().expect("poisoned lock");
+        let res = operation(&mut client);
+
+        match res {
+            Ok(res) => Ok(res),
+            // In case of a USB IO error try to reconnect, and try again
+            Err(trezor_client::Error::TransportSendMessage(
+                trezor_client::transport::error::Error::Usb(rusb::Error::Io),
+            )) => {
```
ImplOfAnImpl, line 161: IMO instead of calling the operation and then checking for error we should just unconditionally call `Trezor::initialize`, passing to it the previosly obtained session_id (it's inside features). At least this is recommended in [the docs](https://docs.trezor.io/trezor-firmware/common/communication/sessions.html).

ImplOfAnImpl (changes requested):

wallet/src/signer/trezor_signer/mod.rs
OBorce: added an init_device(session_id) call which internally calls initialize before each operation.

wallet/src/signer/trezor_signer/mod.rs
OBorce: we check the public keys again, and if it is a different device we error out.

ImplOfAnImpl: > we check the public keys again, and if it is a different device we error out.

I understand this, the point was that it looks conceptually wrong.

Currently, the question is why do we need it if you already call `init_device` every time. Can't it recover the session after a usb error?

And even if it can't, calling `find_trezor_device` still seems wrong. Perhaps then you should split `find_trezor_device` into 2 functions - the "find" part and the "connect" part, so that the former is still called only once.

ImplOfAnImpl: >  Perhaps then you should split find_trezor_device into 2 functions - the "find" part and the "connect" part, so that the former is still called only once.

I guess having a separate "connect" part won't be possible, because `AvailableDevice` is non-copyable. So we'll have to stick to calling `find_trezor_device` again.

ImplOfAnImpl (commented):
This PR is incomplete without changes to `trezor_signer` from https://github.com/mintlayer/mintlayer-core/pull/1895
I suggest moving those changes here (I mean the parts that are relevant for the "automatic" device selection), so that they can be reviewed together.

wallet/src/signer/trezor_signer/mod.rs
OBorce: > Currently, the question is why do we need it if you already call init_device every time. Can't it recover the session after a usb error?

Mainly for the case if the USB device disconnects for whatever reason.

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -1240,24 +1304,75 @@ fn check_public_keys(
...
+        &chain_config,
+    )
+    .map_err(WalletError::SignerError)
 }
 
-fn find_trezor_device() -> Result<(Trezor, TrezorData), TrezorError> {
+fn find_trezor_device() -> Result<(Trezor, TrezorData, Vec<u8>), TrezorError> {
     let mut devices = find_devices(false)
         .into_iter()
         .filter(|device| device.model == Model::Trezor || device.model == Model::TrezorEmulator)
```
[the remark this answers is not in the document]
OBorce: Added but I don't think it is supported, when I run the emulator it doesn't list the Mintlayer compatibility so it will alwasy be filtered out in the next step.

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -123,21 +123,94 @@ pub enum TrezorError {
...
+        db_tx: &impl WalletStorageReadLocked,
+        key_chain: &impl AccountKeyChains,
+    ) -> SignerResult<()> {
+        let mut client = self.client.lock().expect("poisoned lock");
+
+        match client.init_device(Some(self.session_id.clone())) {
+            Ok(_) => Ok(()),
+            // In case of a USB IO error try to reconnect, and try again
+            Err(trezor_client::Error::TransportSendMessage(
+                trezor_client::transport::error::Error::Usb(rusb::Error::Io),
```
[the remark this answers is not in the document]
OBorce: Added matching on all USB errors

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -123,32 +141,111 @@ pub enum TrezorError {
     MultipleSignaturesReturned,
     #[error("A multisig signature was returned for a single address from Device")]
     MultisigSignatureReturned,
+    #[error("The file being loaded is a software wallet and does not correspond to the connected hardware wallet")]
+    HardwareWalletDifferentFile,
+    #[error("Public keys mismatch. Wrong device or passphrase:\nfile DeviceId \"{file_device_id}\", connected device \"{connected_device_id}\",\nfile label \"{file_label}\" and connected device label \"{connected_device_id}\"")]
```
ImplOfAnImpl, line 146: "DeviceId" -> "device id", "connected device" -> "connected device id"

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -1240,61 +1513,146 @@ fn check_public_keys(
...
+        derivation_path,
+        chain_code,
+        Secp256k1PublicKey::from_bytes(&xpub.public_key.serialize())
+            .map_err(|_| TrezorError::InvalidKey)?,
+    );
+    let account_pubkey = ExtendedPublicKey::new(account_pubkey);
+    Ok(account_pubkey)
+}
+
+/// Check that the public keys in the DB are the same as the ones with the connected hardware
```
ImplOfAnImpl, line 1556: Probably "with" should be "from", same for `check_public_keys_against_key_chain`.

ImplOfAnImpl (approved):

wallet/wallet-cli-commands/src/command_handler/mod.rs
```diff
@@ -195,20 +213,37 @@ where
...
+                            let choices = CreateWalletDeviceSelectMenu::new(
+                                devices,
+                                wallet_path,
+                                CliHardwareWalletType::Trezor,
+                                true,
+                            );
+                            return Ok(ConsoleCommand::ChoiceMenu(Box::new(choices)));
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
OBorce: added

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -73,6 +75,12 @@ pub enum WalletManagementCommand {
         /// and the latter will have to be entered every time the device is connected to the host machine.
         #[arg(long, conflicts_with_all(["mnemonic", "passphrase", "whether_to_store_seed_phrase"]))]
         hardware_wallet: Option<CliHardwareWalletType>,
+
+        /// Optionally specify the ID for the trezor device to connect to in case there
+        /// are multiple trezor devices connected at the same time.
+        /// If not specified and there are multiple devices connected a choice will be presented
+        #[arg(long, conflicts_with_all(["mnemonic", "passphrase", "whether_to_store_seed_phrase"]))]
```
azarovh, line 82: it should require hardware_wallet

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -73,6 +75,12 @@ pub enum WalletManagementCommand {
...
+        trezor_device_id: Option<String>,
```
azarovh, line 83: it seems like this string should be either device_id or included into HardwareWalletType::Trezor. Otherwise what happens when we add new wallet type?

wallet/wallet-controller/src/lib.rs
```diff
@@ -269,30 +269,42 @@ where
...
+                        device_id.map(|device_id| SelectedDevice { device_id }),
+                    )
+                    .map_err(SignerError::TrezorError)?)
+                },
+            )
+            .map_err(ControllerError::WalletError)
+            .map(|w| w.map_wallet(RuntimeWallet::Trezor)),
+        };
+
+        match &res {
```
azarovh, line 292: if let Err(_) = res {}

wallet/wallet-cli-commands/src/lib.rs
OBorce: fixed in separate PR #1907

wallet/wallet-cli-commands/src/lib.rs
OBorce: fixed in separate PR #1907

PR #1878, opened 2025-02-05
optimize postgresql queries for API Server scanner

OBorce:
- added a cache table for the latest delegation states so that we can efficiently retrieve all the latest data for each delegation of a pool
- added delegations balance in the pool endpoints

### DIFF
--- a/api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
+++ b/api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
@@ -212,6 +212,13 @@ impl ApiServerStorageRead for ApiServerInMemoryStorageTransactionalRo<'_> {
         self.transaction.get_delegations_from_address(address)
     }
 
+    async fn get_fungible_tokens_by_authority(
+        &self,
+        authority: Destination,
+    ) -> Result<Vec<TokenId>, ApiServerStorageError> {
+        self.transaction.get_fungible_tokens_by_authority(authority)
+    }
+
     async fn get_fungible_token_issuance(
         &self,
         token_id: TokenId,
--- a/api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
+++ b/api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
@@ -22,12 +22,11 @@ use common::{
     },
     primitives::{Amount, BlockHeight, CoinOrTokenId, Id},
 };
-use pos_accounting::PoolData;
 
 use crate::storage::storage_api::{
     block_aux_data::BlockAuxData, ApiServerStorageError, ApiServerStorageRead, BlockInfo,
     CoinOrTokenStatistic, Delegation, FungibleTokenData, NftWithOwner, Order, PoolBlockStats,
-    TransactionInfo, Utxo, UtxoWithExtraInfo,
+    PoolDataWithExtraInfo, TransactionInfo, Utxo, UtxoWithExtraInfo,
 };
 
 use super::ApiServerInMemoryStorageTransactionalRo;
@@ -116,7 +115,7 @@ impl ApiServerStorageRead for ApiServerInMemoryStorageTransactionalRo<'_> {
         &self,
         len: u32,
         offset: u32,
-    ) -> Result<Vec<(PoolId, PoolData)>, ApiServerStorageError> {
+    ) -> Result<Vec<(PoolId, PoolDataWithExtraInfo)>, ApiServerStorageError> {
         self.transaction.get_latest_pool_ids(len, offset)
     }
 
@@ -124,7 +123,7 @@ impl ApiServerStorageRead for ApiServerInMemoryStorageTransactionalRo<'_> {
         &self,
         len: u32,
         offset: u32,
-    ) -> Result<Vec<(PoolId, PoolData)>, ApiServerStorageError> {
+    ) -> Result<Vec<(PoolId, PoolDataWithExtraInfo)>, ApiServerStorageError> {
         self.transaction.get_pool_data_with_largest_staker_balance(len, offset)
     }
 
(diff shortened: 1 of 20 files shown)


azarovh (approved):

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -1096,6 +1119,33 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+                   spend_destination
+            FROM ml.delegations
+            WHERE pool_id = $2
+              AND block_height = (SELECT max_block_height FROM latest_block_height)
+            ON CONFLICT DO NOTHING;
+            "#,
+            &[&height, &pool_id.as_str()]
+        )
+        .await
+        .map_err(|e| ApiServerStorageError::LowLevelStorageError(e.to_string()))?;
```
ImplOfAnImpl, line 1147: This looks hacky:

1. Copying all the delegations should probably be done once per each block height and not every time a delegation is modified (which may happen multiple times per block).

2. Copying itself feels hacky. Have you investigated other approaches? E.g. what if we had an additional table "delegation_id -> max_block_height", would it help speed up `get_pool_delegation_shares`?

ImplOfAnImpl: Btw, regarding reorgs - can't we just update the "delegation_id -> max_block_height" table in `del_delegations_above_height`? IMO we should be able to do it in the same sql query that deletes the delegations, so it should be relatively efficient. Or am I missing something?

OBorce: added the cache table

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -1834,7 +2081,21 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+                ],
+            )
+            .await
+            .map_err(|e| ApiServerStorageError::LowLevelStorageError(e.to_string()))?;
+
+        let coin_or_token_id = CoinOrTokenId::TokenId(token_id);
+        self.tx
+            .execute(
+                "INSERT INTO ml.coin_or_token_decimals (coin_or_token_id, block_height, number_of_decimals) VALUES ($1, $2, $3) ON CONFLICT (coin_or_token_id) DO NOTHING;",
+                &[&coin_or_token_id.encode(), &height, &(issuance.number_of_decimals as i16)],
```
ImplOfAnImpl, line 2098: "ON CONFLICT (coin_or_token_id) DO NOTHING" IMO looks strange.

In the context of this particular function probably UPDATE would make more sense? I mean, what does `set_fungible_token_issuance` mean in the context of the storage - the token somehow got "re-issued", so probably we should just update the data.

In general, storage methods should make sense on their own. And same methods from the postgres and in-memory storage should probably behave in the same way.
And e.g. the in-memory storage makes no particular assumptions on when a token re-issuance can occur - it just adds a new issuance object for the provided `token_id`/`block_height`.
So I guess postgres should do the same, assuming that the same token can be issued multiple times at different heights. I.e. `coin_or_token_decimals` should have `block_height` as a part of the key and queries that use `coin_or_token_decimals` should take this into account.
But then the next question arises - why isn't token decimals value a part of `ml.fungible_token`? And the coin decimals never change, so there isn't much sense in putting them into the db. So `coin_or_token_decimals` seems redundant.

P.S. another weird thing about `coin_or_token_decimals` is that it's partially exposed from the storage interface via `del_coin_or_token_decimals_above_height`.
Where the in-memory storage implementation just does nothing.
IMO it's another sign that it should go.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -2149,6 +2427,15 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+        let coin_or_token_id = CoinOrTokenId::TokenId(token_id);
+        self.tx
+            .execute(
+                "INSERT INTO ml.coin_or_token_decimals (coin_or_token_id, block_height, number_of_decimals) VALUES ($1, $2, $3);",
+                &[&coin_or_token_id.encode(), &height, &0_i16],
+            )
+            .await
+            .map_err(|e| ApiServerStorageError::LowLevelStorageError(e.to_string()))?;
+
         Ok(())
```
ImplOfAnImpl, line 2439: Same comment.

And btw probably the above "INSERT INTO ml.nft_issuance" should also have "ON CONFLICT ... UPDATE" so that it behaves just like its in-memory counterpart.

ImplOfAnImpl (approved):

api-server/api-server-common/src/storage/impls/postgres/queries.rs
OBorce: Doesn't have an on conflict because it cannot happen same as the insert of the nft above it.
Added a panic in the in-memory counterpart.

api-server/api-server-common/src/storage/impls/in_memory/transactional/read.rs
```diff
@@ -46,6 +45,13 @@ impl ApiServerStorageRead for ApiServerInMemoryStorageTransactionalRo<'_> {
         self.transaction.get_address_balance(address, coin_or_token_id)
     }
 
+    async fn get_address_balances(
+        &self,
+        address: &str,
+    ) -> Result<Vec<(CoinOrTokenId, Amount, u8)>, ApiServerStorageError> {
```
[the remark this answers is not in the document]
OBorce: replaced with DecimalAmount

api-server/web-server/src/api/v2.rs
```diff
@@ -620,11 +616,31 @@ pub async fn address<T: ApiServerStorage>(
...
-    "coin_balance": amount_to_json(coin_balance, state.chain_config.coin_decimals()),
-    "locked_coin_balance": amount_to_json(locked_coin_balance, state.chain_config.coin_decimals()),
-    "transaction_history": transaction_history
-    //TODO "token_balances": destination_summary.token_balances(),
+        "coin_balance": amount_to_json(coin_balance.amount, coin_balance.decimals),
+        "locked_coin_balance": amount_to_json(locked_coin_balance, state.chain_config.coin_decimals()),
+        "transaction_history": transaction_history,
+        "tokens": tokens
     })))
 }
```
ImplOfAnImpl, line 645: Btw, could you fix similar formatting issues in other places in this file (in this or other PR)? Namely in the functions `block`, `chain_tip`, `transaction_merkle_path`. Initially I wanted to fix them myself in a small cleanup PR, but I didn't want to create conflicts.

Issue #1887, opened 2025-02-27
Improve Trezor Integration & Support


OBorce:
Features:
- [x]  Add support for signing with standalone private keys.
- Allow users to select a specific Trezor device when multiple devices are connected simultaneously
  - [x] in wallet-cli;
  - [ ] in node-gui.
- [ ] Support entering the Trezor passphrase directly from the wallet GUI/CLI.

Improvements:
- [ ] Improve handling of communication timeouts if the user doesn't perform the action on the device within 100s
  (Edit: this may not be relevant anymore, but more testing is needed)

CI:
- [x] Implement automated tests for Trezor support using the Trezor Emulator and integrate them into the CI pipeline.

PR #1888, opened 2025-02-28
Introduce `TxInput::OrderAccountCommand` without nonces

azarovh:
- Filled amount is now calculated based on the original ask/give price. This makes fill operations commutative. Though it has a negative consequence as potential remainder amounts, resulting in dust utxos produced on conclude, I don't think it's a major problem.
- It is forbidden to fill order with 0 to enforce spending utxo in a tx making it unique and not replayable.

- ~~Nonce for `AccountCommand::FillOrder` is ignored. This is a bit hacky because nonce is still present making it confusing. But I think that this is more like an exception and refactoring out nonce from `TxInput::AccountCommand` seems like not worth it.~~
- ~~Nonce for `AccountCommand::ConcludeOrder` is kept. Also I'm planning to keep it for the upcoming `AccountCommand::FreezeOrder`. Though it seems like it can be removed I don't see a point in doing it.~~
- ~~Ignoring nonce forced me to add fork logic to `TransactionVerifier::disconnect_transaction` function because we don't need to decrement nonce for fill orders after the fork.~~
- ~~Same fork logic goes into `api-server` and `wallet`. Though for wallet it required changing `OutputCache` logic which looks messy and I'm don't quite like it (alternative solution below can fix that).~~

UPDATE:

- Introduced new `TxInput::OrderAccountCommand` with FillOrder and ConcludeOrder commands.
- Nonces for `OrderAccountCommand` are completely removed.
- `OrderAccountCommand::ConcludeOrder` now contains remaining amounts that can be withdrawn from an account so that these amount can be used by hw wallet and also get into a signature. Also these amounts are checked in TransactionVerifier

### DIFF
--- a/api-server/scanner-lib/src/sync/tests/simulation.rs
+++ b/api-server/scanner-lib/src/sync/tests/simulation.rs
@@ -612,6 +612,7 @@ fn update_statistics(
                 *staked = (*staked - *to_spend).unwrap();
             }
         },
+        TxInput::OrderAccountCommand(..) => todo!(),
         TxInput::AccountCommand(_, cmd) => match cmd {
             AccountCommand::MintTokens(token_id, to_mint) => {
                 statistics
--- a/api-server/scanner-lib/src/sync/tests/simulation.rs
+++ b/api-server/scanner-lib/src/sync/tests/simulation.rs
@@ -612,7 +612,6 @@ fn update_statistics(
                 *staked = (*staked - *to_spend).unwrap();
             }
         },
-        TxInput::OrderAccountCommand(..) => todo!(),
         TxInput::AccountCommand(_, cmd) => match cmd {
             AccountCommand::MintTokens(token_id, to_mint) => {
                 statistics
@@ -647,6 +646,7 @@ fn update_statistics(
             }
             AccountCommand::ConcludeOrder(_) | AccountCommand::FillOrder(_, _, _) => {}
         },
+        TxInput::OrderAccountCommand(..) => {}
     });
 }
 
--- a/api-server/web-server/src/api/json_helpers.rs
+++ b/api-server/web-server/src/api/json_helpers.rs
@@ -24,8 +24,8 @@ use common::{
         block::ConsensusData,
         output_value::OutputValue,
         tokens::{IsTokenUnfreezable, NftIssuance, TokenId, TokenTotalSupply},
-        AccountCommand, AccountSpending, Block, ChainConfig, Destination, OrderId,
-        OutPointSourceId, PoolId, Transaction, TxInput, TxOutput, UtxoOutPoint,
+        AccountCommand, AccountSpending, Block, ChainConfig, Destination, OrderAccountCommand,
+        OrderId, OutPointSourceId, PoolId, Transaction, TxInput, TxOutput, UtxoOutPoint,
     },
     primitives::{Amount, BlockHeight, CoinOrTokenId, Idable},
     Uint256,
@@ -318,7 +318,30 @@ pub fn tx_input_to_json(inp: &TxInput, chain_config: &ChainConfig) -> serde_json
                 })
             }
         },
-        TxInput::OrderAccountCommand(..) => todo!(),
+        TxInput::OrderAccountCommand(cmd) => match cmd {
+            OrderAccountCommand::FillOrder(id, fill, dest) => {
+                json!({
+                    "input_type": "OrderAccountCommand",
+                    "command": "FillOrder",
+                    "order_id": Address::new(chain_config, *id).expect("addressable").to_string(),
+                    "fill_atoms": json!({"atoms": fill.into_atoms().to_string()}),
+                    "destination": Address::new(chain_config, dest.clone()).expect("no error").as_str(),
+                })
+            }
+            OrderAccountCommand::ConcludeOrder {
+                order_id,
+                ask_balance,
+                give_balance,
+            } => {
+                json!({
+                    "input_type": "OrderAccountCommand",
+                    "command": "ConcludeOrder",
+                    "order_id": Address::new(chain_config, *order_id).expect("addressable").to_string(),
+                    "ask_atoms": json!({"atoms": ask_balance.into_atoms().to_string()}),
+                    "give_atoms": json!({"atoms": give_balance.into_atoms().to_string()}),
+                })
+            }
+        },
         TxInput::AccountCommand(nonce, cmd) => match cmd {
             AccountCommand::MintTokens(token_id, amount) => {
                 json!({
--- a/api-server/web-server/src/api/json_helpers.rs
+++ b/api-server/web-server/src/api/json_helpers.rs
@@ -330,15 +330,15 @@ pub fn tx_input_to_json(inp: &TxInput, chain_config: &ChainConfig) -> serde_json
             }
             OrderAccountCommand::ConcludeOrder {
                 order_id,
-                ask_balance,
-                give_balance,
+                filled_amount,
+                remaining_give_amount,
             } => {
                 json!({
                     "input_type": "OrderAccountCommand",
                     "command": "ConcludeOrder",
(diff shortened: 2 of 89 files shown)


common/src/chain/transaction/account_outpoint.rs
```diff
@@ -160,3 +173,32 @@ impl AccountOutPoint {
...
+    FillOrder(OrderId, Amount, Destination),
+    // Close an order and withdraw all remaining funds from both give and ask balances.
+    // Only the address specified as `conclude_key` can authorize this command.
+    // Amounts are important because they get into a signature so that such a tx cannot be applied
+    // to an altered state.
+    #[codec(index = 1)]
+    ConcludeOrder {
+        order_id: OrderId,
+        filled_amount: Amount,
+        remaining_give_amount: Amount,
```
azarovh, line 202: I will remove these fields is we go with `OrderData` in the signature

azarovh: Seems like I cannot do that because `OrderData` does not contain current balances

trezor-common/src/tests/mod.rs
```diff
@@ -146,6 +146,10 @@ impl From<chain::TxInput> for crate::TxInput {
             chain::TxInput::AccountCommand(nonce, command) => {
                 Self::AccountCommand(nonce.value(), command.into())
             }
+            chain::TxInput::OrderAccountCommand(_) => {
+                //TODO: support OrdersVersion::V1
```
ImplOfAnImpl, line 150: I guess this should be put into https://github.com/mintlayer/mintlayer-core/issues/1887

Or maybe some other issue

wallet/src/account/mod.rs
```diff
@@ -981,14 +983,35 @@ impl<K: AccountKeyChains> Account<K> {
...
-            TxInput::AccountCommand(nonce, AccountCommand::ConcludeOrder(order_id)),
-            order_info.conclude_key.clone(),
-        )]);
+        let version = self
+            .chain_config
+            .chainstate_upgrades()
+            .version_at_height(self.account_info.best_block_height())
+            .1
+            .orders_version();
+
```
ImplOfAnImpl, line 992: Tecnically, this can be "best_block_height + 1", because when the tx is mined, the upgrade height will be reached. (Unless a reorg occurs and the tx gets mined at a lower height, but I guess this is always possible).

wallet/src/account/mod.rs
```diff
@@ -2514,6 +2582,67 @@ fn group_preselected_inputs(
...
+    fee: Amount,
+    update_preselected_inputs: &mut impl FnMut(Currency, Amount, Amount, Amount) -> WalletResult<()>,
+) -> WalletResult<()> {
+    let order_info = order_info
+        .as_ref()
+        .and_then(|info| info.get(&order_id))
+        .ok_or(WalletError::OrderInfoMissing(order_id))?;
+
+    let filled_amount = orders_accounting::calculate_filled_amount(
+        order_info.ask_balance,
```
ImplOfAnImpl, line 2598: Probably it's better to use initially_asked/initially_given here and/or assert that they equal ask_balance/give_balance?

wallet/wallet-controller/src/helpers.rs
```diff
@@ -291,42 +291,13 @@ async fn into_utxo_and_destination<T: NodeInterface, B: storage::Backend>(
             (None, TxAdditionalInfo::new(), dest)
         }
         TxInput::AccountCommand(_, cmd) => {
-            // find authority of the token
+            // find authority of the order
```
ImplOfAnImpl, line 294: The comment is still weird.

"find a destination associated with the order" would make more sense (not sure if it's useful though)

ImplOfAnImpl (approved):
Do we have a test for the case when an order is created and partially filled before the upgrade and then filled again and concluded after?

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/purposes_check.rs
```diff
@@ -184,7 +184,7 @@ pub fn check_tx_inputs_outputs_purposes(
         .iter()
         .filter(|input| match input {
             TxInput::Utxo(_) | TxInput::Account(..) => false,
-            TxInput::AccountCommand(..) => true,
+            TxInput::AccountCommand(..) | TxInput::OrderAccountCommand(..) => true,
```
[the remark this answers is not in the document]
ImplOfAnImpl: > Below we should probably also count `CreateOrder` outputs and prohibit having more than one.

Actually now I see that we already prohibit this in a different place (and produce `OrderAlreadyExists`).

common/src/chain/transaction/account_outpoint.rs
azarovh: removed

api-server/scanner-lib/src/sync/tests/simulation.rs
```diff
@@ -646,6 +646,7 @@ fn update_statistics(
             }
             AccountCommand::ConcludeOrder(_) | AccountCommand::FillOrder(_, _, _) => {}
         },
+        TxInput::OrderAccountCommand(..) => {}
```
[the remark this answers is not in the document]
azarovh: We don't do anything with statistics for orders

api-server/web-server/src/api/json_helpers.rs
```diff
@@ -318,6 +318,30 @@ pub fn tx_input_to_json(inp: &TxInput, chain_config: &ChainConfig) -> serde_json
...
             }
         },
+        TxInput::OrderAccountCommand(cmd) => match cmd {
+            OrderAccountCommand::FillOrder(id, fill, dest) => {
+                json!({
+                    "input_type": "OrderAccountCommand",
+                    "command": "FillOrder",
+                    "order_id": Address::new(chain_config, *id).expect("addressable").to_string(),
+                    "fill_atoms": json!({"atoms": fill.into_atoms().to_string()}),
+                    "destination": Address::new(chain_config, dest.clone()).expect("no error").as_str(),
```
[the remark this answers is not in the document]
azarovh: You are right. Originally we used expect because the requirements for api-server were not so strict. But right now exchanges are using so we might want to reconsider. 
I don't want to change anything at the moment, so I just fixed the message inside the expect you mentioned.

chainstate/tx-verifier/src/transaction_verifier/input_output_policy/purposes_check.rs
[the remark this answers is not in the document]
azarovh: I added the check anyway because we want to check it on both levels.

trezor-common/src/tests/mod.rs
azarovh: https://github.com/mintlayer/mintlayer-core/issues/1902

mempool/src/error/ban_score.rs
```diff
@@ -514,6 +519,7 @@ impl MempoolBanScore for orders_accounting::Error {
             Error::InvariantNonzeroGiveBalanceForMissingOrder(_) => 100,
             Error::OrderOverflow(_) => 100,
             Error::OrderOverbid(_, _, _) => 100,
```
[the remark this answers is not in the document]
azarovh: agree, seems ok to not punish for that in mempool

wallet/src/account/mod.rs
```diff
@@ -2514,6 +2569,75 @@ fn group_preselected_inputs(
     Ok(preselected_inputs)
 }
 
+fn handle_fill_order_op(
+    order_id: OrderId,
+    fill_amount_in_ask_currency: Amount,
+    order_info: Option<&BTreeMap<OrderId, &RpcOrderInfo>>,
```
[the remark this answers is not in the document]
azarovh: Not sure, Option is passed from the outside and is more explicit in that context. I could unwrap Option before passing here but it just seems like unnecessary typing.

PR #1890, opened 2025-03-07
Add support for signing with standalone keys to Trezor

OBorce:
- add support for signing with standalone keys with the Trezor signer
- unify software and trezor signer tests

### DIFF
--- a/wallet/src/signer/trezor_signer/mod.rs
+++ b/wallet/src/signer/trezor_signer/mod.rs
@@ -34,10 +34,7 @@ use common::{
                         sign_classical_multisig_spending, AuthorizedClassicalMultisigSpend,
                         ClassicalMultisigCompletionStatus,
                     },
-                    multisig_partial_signature::{
-                        self, PartiallySignedMultisigChallenge,
-                        PartiallySignedMultisigStructureError,
-                    },
+                    multisig_partial_signature::{self, PartiallySignedMultisigChallenge},
                 },
                 htlc::produce_uniparty_signature_for_htlc_input,
                 standard_signature::StandardInputSignature,
@@ -223,7 +220,7 @@ impl TrezorSigner {
     fn make_signature<'a, 'b, F, F2>(
         &self,
         signatures: &[MintlayerSignature],
-        standalone_inputs: Option<&'a [StandaloneInput]>,
+        standalone_inputs: &'a [StandaloneInput],
         destination: &'b Destination,
         sighash_type: SigHashType,
         sighash: H256,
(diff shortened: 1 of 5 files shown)


wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -211,17 +219,21 @@ impl TrezorSigner {
         operation(&mut client).map_err(|e| TrezorError::DeviceError(e.to_string()).into())
     }
 
-    fn make_signature<F>(
+    #[allow(clippy::too_many_arguments)]
+    fn make_signature<'a, 'b, F, F2>(
         &self,
         signatures: &[MintlayerSignature],
-        destination: &Destination,
+        standalone_inputs: Option<&'a [StandaloneInput]>,
```
ImplOfAnImpl, line 226: Why pass an `Option` here? `None` can be represented via an empty slice.

Currently you have the technical possibility of passing `Some(&[])` which doesn't make much sense. Also, it makes the lines `Some(_) => return Err(TrezorError::MultisigSignatureReturned.into())` technically incorrect, because `MultisigSignatureReturned` isn't a correct error type in the case of an empty slice.

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -348,6 +381,55 @@ impl TrezorSigner {
...
+        standalone_inputs.iter().try_fold(
+            (current_signatures, new_status),
+            |(mut current_signatures, mut status), inp| -> SignerResult<_> {
+                if status == SignatureStatus::FullySigned {
+                    return Ok((current_signatures, status));
+                }
+
+                let key_index = inp
+                    .multisig_idx
+                    .ok_or(PartiallySignedMultisigStructureError::InvalidSignatureIndex)?;
```
ImplOfAnImpl, line 403: `InvalidSignatureIndex` currently has the message "The challenge contains a signature for a public key that does not exist", which is not applicable to the current situation.
I suggest either making the current message as generic as the error name `InvalidSignatureIndex` itself, or (better) renaming `InvalidSignatureIndex` to something more specific (so that it better reflects the existing message) and adding a new error variant specific for the current case.

wallet/src/signer/trezor_signer/mod.rs
```diff
@@ -443,24 +537,28 @@ impl Signer for TrezorSigner {
...
-                                            )?;
-
-                                        let sighash_type = SigHashType::all();
-                                        let sig = add_secret_if_needed(StandardInputSignature::new(
-                                            sighash_type,
-                                            current_signatures.encode(),
-                                        ));
-
-                                        return Ok((Some(sig), previous_status, status));
+                                    let (current_signatures, new_status) = if let Some(signature) = new_signatures.get(input_index)
```
ImplOfAnImpl, line 540: signature -> signatures?

ImplOfAnImpl (approved):

OBorce:
Added to #1874

PR #1894, opened 2025-03-18
Add support for pagination in the CLI wallet

OBorce:


wallet/wallet-cli-lib/src/repl/interactive/mod.rs
```diff
@@ -226,3 +231,93 @@ fn handle_response<N: NodeInterface>(
...
+    let (_, end_index) = lines
+        .fold_while((0, 0), |(current_rows, current_index), line| {
+            let new_rows = line.len().div_ceil(cols);
+            if current_rows + new_rows <= page_rows {
+                let new_total_rows = current_rows + new_rows;
+                let new_end_index = current_index + line.len() + 1;
+                FoldWhile::Continue((new_total_rows, new_end_index))
+            } else {
+                let rows_available = page_rows - current_rows;
+                // make sure we cut it on the start of a utf-8 char boundary
```
ImplOfAnImpl, line 311: It's either "on the start of a utf-8 char" or "at a utf-8 char boundary"

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
```diff
@@ -226,3 +231,93 @@ fn handle_response<N: NodeInterface>(
...
+                let new_end_index = current_index
+                    + (1..=rows_available * cols)
+                        .rev()
+                        .find(|&i| line.is_char_boundary(i))
+                        // 0 is always a safe char boundary
+                        .unwrap_or(0);
+                FoldWhile::Done((page_rows, new_end_index))
```
ImplOfAnImpl, line 318: 1. Regaring unicode handling:
    - This won't work correctly in the "k" case, because here you calculate the number of bytes from the beginning of the line to some char boundary and the caller will subtract this value from `current_index`, treating it as the number of bytes from the end of the line to a char boundary.

    - `rows_available * cols` should be the number of characters and not bytes.

2. Even in the ascii case this doesn't work well in the "k" case, because the partial row shown at the top will always have "cols" characters.
    E.g. here is how I see a top row when the size of the terminal is small.
    Initially:
    ```
    | db76bb89b8e14daeed2caba4aeb28493c95eef39cd1ea4dc71c53d77c913d0d5 | 297962
     | 1726820532     |
    ```
    After presing "j":
    ```
    | 1726820532     |
    ```
    And after pressing "j" again and then "k":
    ```
    2caba4aeb28493c95eef39cd1ea4dc71c53d77c913d0d5 | 297962      | 1726820532     |
    ```

I think you should build some kind of "layout" object once (well, maybe rebuilding it on `Resize`), and the "k"/"j" should only scroll over the lines of that "layout" and not do any text parsing.
(In the current implementation the "layout" can be a `Vec<&str>`, where each element represents a separate line of the output)

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
```diff
@@ -226,3 +231,93 @@ fn handle_response<N: NodeInterface>(
...
+            (_, _) => "Press 'j' for next, 'k' for previous, 'q' to quit",
+        };
+        console.print_line(commands);
+
+        // Wait for user input.
+        crossterm::terminal::enable_raw_mode().expect("Should not fail normally");
+        let event = crossterm::event::read().expect("Should not fail normally");
+        crossterm::terminal::disable_raw_mode().expect("Should not fail normally");
+
+        if let crossterm::event::Event::Key(key_event) = event {
```
ImplOfAnImpl, line 273: It would be nice to handle the `Resize` event as well (a TODO is fine for me)

ImplOfAnImpl (approved):

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
```diff
@@ -226,3 +231,93 @@ fn handle_response<N: NodeInterface>(
...
+    console: &mut impl ConsoleOutput,
+) {
+    let mut current_index = 0;
+
+    let (cols, rows) = crossterm::terminal::size().unwrap_or((80, 24));
+    let cols = cols as usize;
+    let page_rows = (rows - 4) as usize; // make room for the header and prompt
+
+    loop {
+        line_editor.clear_screen().expect("Should not fail normally");
```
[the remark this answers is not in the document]
OBorce: done except for the one in the OnceDestructor, not sure about that one

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
```diff
@@ -226,3 +231,93 @@ fn handle_response<N: NodeInterface>(
     }
     None
 }
+
+fn paginate_output(
+    header: String,
+    body: String,
```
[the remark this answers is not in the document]
OBorce: added TODO

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
```diff
@@ -226,3 +231,93 @@ fn handle_response<N: NodeInterface>(
...
+            (0, _) => "Press 'j' for down, 'q' to quit",
+            (_, end) if end == body.len() => "Press 'k' for previous, 'q' to quit",
+            (_, _) => "Press 'j' for next, 'k' for previous, 'q' to quit",
+        };
+        console.print_line(commands);
+
+        // Wait for user input.
+        crossterm::terminal::enable_raw_mode().expect("Should not fail normally");
+        let event = crossterm::event::read().expect("Should not fail normally");
+        crossterm::terminal::disable_raw_mode().expect("Should not fail normally");
```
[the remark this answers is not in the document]
OBorce: Added OnceDestructor for the inner loop, and added TODO for the outer loop

wallet/wallet-cli-lib/src/repl/interactive/mod.rs
OBorce: fixed

azarovh (approved):

PR #1900, opened 2025-04-03
Add option to list change addresses in the wallet

### LINKED ISSUE #1860: [Feature] `address-show` should also display return/change addresses and optionally show address balances
Currently, the  `address-show` command only lists receiving addresses. It would be beneficial to enhance this command with the following features:

---

## Primary Enhancement: Support for return/change addresses

The command should allow users to view change (return) addresses in addition to receiving ones.

**Proposed options:**
- Add a flag like `--include-change` or `--all` to show both receiving and change addresses.
- Clearly label address types in the output (e.g., `receiving`, `change`).

---

##  Secondary Enhancement: Show address balances

By default, the command should display the current balance associated with each address.

Optionally, developers may prefer to guard this behavior behind a flag such as `--with-balances` for performance or design reasons.

This would give users a more complete picture of their wallet's state directly from the CLI.

---

### Current output:
```
Wallet(Account BRIDGE_HOT)> address-show
+-------+----------------------------------------------+--------------------------------+
| Index | Address                                      | Is used in transaction history |
+=======+==============================================+================================+
| 0     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | Yes                            |
| 1     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | No                             |
| 2     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | No                             |
| 3     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | No                             |
| 4     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | No                             |
+-------+----------------------------------------------+--------------------------------+
```

### Proposed extended output (with or without `--with-balances`, depending on design):
```
Wallet(Account BRIDGE_HOT)> address-show --all
+-------+----------------------------------------------+----------+-----------------------------+----------+
| Index | Address                                      | Type     | Is used in transaction hist | Balance  |
+=======+==============================================+==========+=============================+==========+
| 0     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | receiving| Yes                         | 1.234567 |
| 1     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | receiving| No                          | 0.000000 |
| 2     | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | receiving| No                          | 0.000000 |
| 1000  | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | change   | Yes                         | 0.456789 |
| 1001  | mtc1qxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx | change   | No                          | 0.000000 |
+-------+----------------------------------------------+----------+-----------------------------+----------+
```

---

## Use cases

- Better wallet transparency
- Easier wallet reconciliation and audits
- Debugging wallet behavior
- Quick balance checks per address

OBorce:
Add option to show change addresses and also list coin balances for each address

closes #1860


wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -194,7 +194,11 @@ pub enum ColdWalletCommand {
     },
 
     #[clap(name = "address-show")]
-    ShowReceiveAddresses,
+    ShowAddresses {
+        /// Include the change addresses along with the receiving addresses
+        #[arg(long = "include-change")]
+        include_change: bool,
```
azarovh, line 200: looks like it should an optional argument

wallet/wallet-controller/src/read.rs
```diff
@@ -179,30 +194,75 @@ where
...
-    ) -> Result<MapAddressWithUsage<Destination>, ControllerError<T>> {
-        let addresses = self.get_all_issued_addresses()?;
-        let usage = self.get_addresses_usage()?;
-
-        Ok(addresses
-            .into_iter()
-            .map(|(child_number, address)| {
+        include_change_addresses: bool,
+    ) -> Result<Vec<AddressInfo>, ControllerError<T>> {
+        fn map_addresses<'a, I>(
```
azarovh, line 215: why fn and not a closure?

wallet/wallet-controller/src/read.rs
```diff
@@ -179,30 +194,75 @@ where
...
+            .collect()
+        }
+
+        let balances = self.get_address_balances()?;
+
+        let addresses = self.get_all_issued_addresses(KeyPurpose::ReceiveFunds)?;
+        let usage = self.get_addresses_usage(KeyPurpose::ReceiveFunds)?;
+
+        let receiving_addresses = addresses.into_iter().map(|(child_number, address)| {
+            (child_number, address, usage, KeyPurpose::ReceiveFunds)
```
azarovh, line 249: `usage` is the same for every element so it's confusing that it's included into iterator. I'd just pass it to map_addresses as an arg.

wallet/wallet-cli-commands/src/lib.rs
OBorce: it defaults to false if not specified so works as if optional

wallet/wallet-controller/src/read.rs
OBorce: closures can't have generic inputs which is the 2 different iterators one with a Chain and one without.

ImplOfAnImpl (approved):

Issue #1176, opened 2023-09-14
Allow sending the entire wallet balance


pavel-kokolemin:
I don't think it is currently possible to send the entire wallet balance. There should be an option to deduct the network fee from the recipient.

PR #1904, opened 2025-04-10
Feature/wallet sweep all addresses

OBorce:
- add an --all flag to sweep address command that sweeps all addresses for an account
- add a new command for setting the CLI config for broadcasting to true or false
- update all RPC endpoints to return the hex encoded signed transaction, fees and a broadcasted status

### DIFF
--- a/wallet/wallet-cli-commands/src/command_handler/mod.rs
+++ b/wallet/wallet-cli-commands/src/command_handler/mod.rs
@@ -135,11 +135,7 @@ where
                 id_to_hex_string(*new_tx.tx_id.as_hash())
             )
         } else {
-            let hex = new_tx.tx.to_string();
-            let mut summary = new_tx.tx.take().transaction().text_summary(chain_config);
-            format_fees(&mut summary, &new_tx.fees);
-
-            format!("{summary}\nThe transaction was created and is ready to be submitted:\n{hex}",)
+            format_tx_to_be_broadcasted(new_tx.tx, &new_tx.fees, chain_config)
         };
 
         ConsoleCommand::Print(status_text)
@@ -1127,11 +1123,17 @@ where
                     )
                     .await?;
 
-                Ok(ConsoleCommand::Print(format!(
-                    "A new token has been issued with ID: {} in tx: {}",
-                    new_token.token_id,
-                    id_to_hex_string(*new_token.tx_id.as_hash())
-                )))
+                let result = if new_token.broadcasted {
+                    format!(
+                        "A new token has been issued with ID: {} in tx: {}",
+                        new_token.token_id,
+                        id_to_hex_string(*new_token.tx_id.as_hash())
+                    )
+                } else {
+                    format_tx_to_be_broadcasted(new_token.tx, &new_token.fees, chain_config)
+                };
+
+                Ok(ConsoleCommand::Print(result))
             }
 
             WalletCommand::IssueNewNft {
@@ -1161,11 +1163,17 @@ where
                     .issue_new_nft(selected_account, destination_address, metadata, self.config)
                     .await?;
 
-                Ok(ConsoleCommand::Print(format!(
-                    "A new NFT has been issued with ID: {} in tx: {}",
-                    new_token.token_id,
-                    id_to_hex_string(*new_token.tx_id.as_hash())
-                )))
+                let result = if new_token.broadcasted {
+                    format!(
+                        "A new NFT has been issued with ID: {} in tx: {}",
+                        new_token.token_id,
+                        id_to_hex_string(*new_token.tx_id.as_hash())
+                    )
+                } else {
+                    format_tx_to_be_broadcasted(new_token.tx, &new_token.fees, chain_config)
+                };
+
+                Ok(ConsoleCommand::Print(result))
             }
 
             WalletCommand::MintTokens {
@@ -1640,15 +1648,24 @@ where
 
             WalletCommand::CreateDelegation { owner, pool_id } => {
                 let (wallet, selected_account) = wallet_and_selected_acc(&mut self.wallet).await?;
-                let delegation_id = wallet
-                    .create_delegation(selected_account, owner, pool_id, self.config)
-                    .await?
-                    .delegation_id;
+                let new_delegation =
+                    wallet.create_delegation(selected_account, owner, pool_id, self.config).await?;
 
-                Ok(ConsoleCommand::Print(format!(
-                    "Success, the creation of delegation transaction was broadcast to the network. Delegation id: {}",
-                    delegation_id
-                )))
+                let result = if new_delegation.broadcasted {
+                    format!(
+                        "Success, the creation of delegation transaction was broadcast to the network. Delegation id: {} in tx: {}",
+                        new_delegation.delegation_id,
+                        id_to_hex_string(*new_delegation.tx_id.as_hash())
+                    )
+                } else {
+                    format_tx_to_be_broadcasted(
+                        new_delegation.tx,
+                        &new_delegation.fees,
+                        chain_config,
+                    )
+                };
+
+                Ok(ConsoleCommand::Print(result))
             }
 
             WalletCommand::DelegateStaking {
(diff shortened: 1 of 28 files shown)


azarovh (approved):

wallet/wallet-controller/src/synced_controller.rs
```diff
@@ -569,7 +569,7 @@ where
         &mut self,
         destination_address: Destination,
         from_addresses: BTreeSet<Destination>,
```
ImplOfAnImpl, line 571: The fact that empty `from_addresses` now means "all addresses" is super unintuitive and must be documented. And in multiple places too:
*) The doc comment for this function must mention this.
*) Well, probably every `sweep_addresses` function should have a doc comment mentioning this (if it doesn't have an explicit "all" parameter).
*) Need a comment where WalletCommand::SweepFromAddress is handled, because it's supper confusing - you check "all" and then forget about it.

But IMO it's better to just pass the "all" flag explicitly everywhere (and require that if `all` is `true`, then `from_addresses` must be empty).

wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -389,8 +389,9 @@ trait WalletRpc {
         account: AccountArg,
         destination_address: RpcAddress<Destination>,
         from_addresses: Vec<RpcAddress<Destination>>,
+        all: Option<bool>,
```
ImplOfAnImpl, line 392: `Option` looks redundant, why not just pass `bool`?

wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -378,7 +378,7 @@ trait WalletRpc {
         amount: RpcAmountIn,
         selected_utxos: Vec<RpcUtxoOutpoint>,
         options: TransactionOptions,
-    ) -> rpc::RpcResult<NewTransaction>;
+    ) -> rpc::RpcResult<RpcNewTransaction>;
 
     /// Sweep all spendable coins or tokens from an address or addresses to a given address.
     /// Spendable coins are any coins that are not locked, and tokens that are not frozen or locked.
```
ImplOfAnImpl, line 384: Need to mention what "all" means and how it interacts with "from_addresses".

ImplOfAnImpl (changes requested):
Btw, when this is merged, https://github.com/mintlayer/mintlayer-core/issues/1176 can probably be closed.

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -278,6 +278,12 @@ pub enum WalletCommand {
     #[command(flatten)]
     ColdCommands(ColdWalletCommand),
 
+    /// Configure broadcasting to Mempool to true of false.
+    /// By default it is set to true. If set to false, any command that creates a transaction will
+    /// return it to the user and will need to be submitted manually with the command `node-submit-transaction`
+    #[clap(name = "config-broadcast")]
+    ConfigBroadcast { broadcast: bool },
```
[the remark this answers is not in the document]
OBorce: Changed the bool to a YesNo enum

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -527,7 +533,11 @@ pub enum WalletCommand {
         /// The receiving address of the coins or tokens
         destination_address: String,
         /// The addresses to be swept
+        #[arg(required_unless_present("all"))]
         addresses: Vec<String>,
+        /// Sweep all addresses
+        #[arg(long = "all", default_value_t = false, conflicts_with_all(["addresses"]))]
+        all: bool,
     },
```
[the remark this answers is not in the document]
OBorce: fixed in a separate PR

wallet/wallet-cli-commands/src/command_handler/mod.rs
```diff
@@ -1178,23 +1208,23 @@ where
                     .mint_tokens(selected_account, token_id, address, amount, self.config)
                     .await?;
 
-                Ok(Self::new_tx_submitted_command(new_tx))
+                Ok(Self::new_tx_command(new_tx, chain_config))
```
ImplOfAnImpl, line 1211: Plz check all commands in this file that create a transaction that may or may not be broadcast, and ensure that the correct message is printed in the "config-broadcast no" mode.

E.g. as I can see, commands like IssueNewToken, IssueNewNft, CreateDelegation, DelegateStaking, WithdrawFromDelegation (probably some more) will currently report that the tx has been sent even if it hasn't.

wallet/wallet-rpc-lib/src/rpc/server_impl.rs
```diff
@@ -1104,16 +1029,16 @@ where
         token_id: Option<RpcAddress<TokenId>>,
         htlc: RpcHashedTimelockContract,
         options: TransactionOptions,
-    ) -> rpc::RpcResult<HexEncoded<SignedTransaction>> {
-        let config = ControllerConfig {
-            in_top_x_mb: options.in_top_x_mb(),
-            broadcast_to_mempool: true,
-        };
-
+    ) -> rpc::RpcResult<RpcNewTransaction> {
```
ImplOfAnImpl, line 1032: Htlc tx is never broadcasted automatically. I wonder if we should have a separate type similar to NewTransaction that doesn't have the `broadcasted` field and instead implies that it's always false.
Also, we could re-use that type for `make_tx_for_sending_tokens_with_intent` and `make_tx_to_send_tokens_from_multisig_address`.

P.S. we do have the `ComposedTransaction` type, so I guess we could re-use it instead. But probably it should be made consistent with `RpcNewTransaction`. (i.e. contain the same fields except for `broadcasted`).

wallet/wallet-rpc-lib/src/rpc/server_impl.rs
```diff
@@ -609,18 +597,14 @@ where
...
-        };
         rpc::handle_result(
             self.request_send_coins(
                 account_arg.index::<N>()?,
                 address,
                 amount,
                 selected_utxo.into_outpoint(),
                 change_address,
-                config,
+                options.into(),
```
ImplOfAnImpl, line 607: There is a bunch of functions in this file - `transaction_from_cold_input`, `decommission_stake_pool_request`, `make_tx_for_sending_tokens_with_intent`, `make_tx_to_send_tokens_from_multisig_address` and `create_htlc_transaction` (if I haven't missed one) that never attempt to broadcast the tx, so passing `TransactionOptions` to them, which now has the `broadcast_to_mempool` field, looks confusing. Perhaps we should have a separate type (e.g. named `TransactionRequestOptions`) without the `broadcast_to_mempool` field?

wallet/wallet-rpc-lib/src/rpc/interface.rs
```diff
@@ -382,9 +382,13 @@ trait WalletRpc {
...
-    ) -> rpc::RpcResult<NewTransaction>;
+    ) -> rpc::RpcResult<RpcNewTransaction>;
 
-    /// Sweep all spendable coins or tokens from an address or addresses to a given address.
+    /// Sweep all spendable coins or tokens from an address or addresses specified in `from_addresses`
+    /// or all addresses from this account if all is set to true, to the given destination address.
+    /// Either 1 or more addresses need to be specified in `from_addresses` with all set to false, or
+    /// `from_addresses` needs to be empty and all set to true.
+    ///
     /// Spendable coins are any coins that are not locked, and tokens that are not frozen or locked.
```
ImplOfAnImpl, line 392: Put "all" to backticks when it refers to the parameter.

ImplOfAnImpl (approved):

wallet/wallet-controller/src/synced_controller.rs
```diff
@@ -1113,19 +1113,26 @@ where
         output_value: OutputValue,
         htlc: HashedTimelockContract,
         additional_info: TxAdditionalInfo,
-    ) -> Result<SignedTransaction, ControllerError<T>> {
+    ) -> Result<NewTransaction, ControllerError<T>> {
```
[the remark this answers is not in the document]
OBorce: Added a new type similar to NewTransaction without the broadcast

wallet/wallet-rpc-lib/src/rpc/server_impl.rs
OBorce: Added a new type similar to NewTransaction without the broadcast

wallet/wallet-rpc-lib/src/rpc/server_impl.rs
OBorce: fixed and also changed the rpc interface for those methods

PR #1905, opened 2025-04-10
Fix `generate_transaction_graph` test function

azarovh:
Test function didn't account for fee when generating txs, which could fail some tests

### DIFF
--- a/mempool/src/pool/tx_pool/tests/utils.rs
+++ b/mempool/src/pool/tx_pool/tests/utils.rs
@@ -279,15 +279,22 @@ pub fn generate_transaction_graph(
         let mut total = 0u128;
         let mut amts = Vec::new();
 
-        for _ in 0..n_inputs {
+        // the number is chosen to avoid generating empty range below
+        let min_valid_total_amount = 2;
+
+        let mut input_count = 0;
+        while input_count < n_inputs || total < estimated_fee.into_atoms() + min_valid_total_amount
+        {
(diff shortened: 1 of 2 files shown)


OBorce (approved):

mempool/src/pool/tx_pool/tests/utils.rs
```diff
@@ -283,6 +285,8 @@ pub fn generate_transaction_graph(
             builder = builder.add_input(outpt, empty_witness(rng));
         }
 
+        total = total.checked_sub(estimated_fee.into_atoms())?;
```
ImplOfAnImpl, line 288: At least in one of the cases the corresponding test expects a specific number of txs to be generated (`mempool_full_real` in basic.rs has some `assert`s with `num_txs`); this is not always possible with the new logic. Either the test(s) should be fixed, so that they operate on the actual number of txs returned, or (better) the generation logic should be fixed.

Also, it's better to fix the output generation logic below, which will currently panic if `total` becomes 0 or 1 (due to `gen_range` being called with an empty range). It's better to check that `total` is at least 2 at the beginning of that loop and exit early if it's not.

So I guess the result might be something like
```
let min_amount_for_output_generation = 2;

while `input count` < n_inputs OR `total` < `estimated_fee` + `min_amount_for_output_generation` { add a random input }

...

for _ in 0..n_outputs {
    // Make sure that the gen_range call below won't panic.
    if total < min_amount_for_output_generation {
        break;
    }
    let amt = rng.gen_range(...);
}
```

PR #1908, opened 2025-04-15
`make_token_id` should not generate id from account inputs

azarovh:
- Introduced an upgrade so that `make_token_id` produce id not from the first input but from the first utxo input.
- Moved all `make_xxx_id` to a single file to keep it unified.
- Moved logic of finding first utxo input inside `make_xxx_id` functions because the implementation has already diverged between chainstate and wallet: wallet failed if first input is not utxo while chainstate tries to iterate over inputs.
- Added `from_utxo` method to Id types for better usability
- `PosAccountingOperations::create_delegation_id` now accepts DelegationId instead of outpoint. It's now unified with creating a pool. Also it makes more sense since `pos_accounting` doesn't actually care about outpoints.


common/src/chain/pos/delegation_id.rs
```diff
@@ -0,0 +1,94 @@
...
+        Self::new(hasher.finalize().into())
+    }
+
+    pub fn random_using<R: Rng + CryptoRng>(rng: &mut R) -> Self {
+        Self::new(H256::random_using(rng))
+    }
+
+    pub const fn zero() -> Self {
+        Self::new(H256::zero())
+    }
```
ImplOfAnImpl, line 58: Btw maybe we should put them to `Id` directly?

OBorce (approved):

PR #1909, opened 2025-04-16
Fix incorrect rebase of fix/wallet_remove_conflicting_txs on master

azarovh:
Edit: also, `violates_frozen_token` now checks the tx outputs too.


ImplOfAnImpl (approved):

azarovh:
> with CreateOrder you can have a token id only in the output

But `violates_frozen_token` checks TokenData for order inputs, both give and ask currency. What's the case I am missing here?

wallet/src/account/output_cache/mod.rs
```diff
@@ -890,36 +890,33 @@ impl OutputCache {
...
+                OutputValue::TokenV0(_) | OutputValue::Coin(_) => false,
+            }),
+            TxOutput::IssueNft(_, _, _)
+            | TxOutput::DataDeposit(_)
+            | TxOutput::CreateStakePool(_, _)
+            | TxOutput::DelegateStaking(_, _)
+            | TxOutput::CreateDelegationId(_, _)
+            | TxOutput::IssueFungibleToken(_)
+            | TxOutput::ProduceBlockFromStake(_, _) => false,
+        };
```
OBorce, line 912: maybe we can reuse get_referenced_token_ids from token_utils, something like:
get_referenced_token_ids(output).contains(frozen_token_id)

OBorce (approved):

wallet/src/account/output_cache/mod.rs
ImplOfAnImpl: > Or maybe even change get_referenced_token_ids to return an iterator instead of the BTreeSet and let the caller collect if needed.

In the end I decided to return a `SmallVec<[TokenId; 2]>` instead (returning `impl Iterator` is also possible, but it looked less clean IMO).

OBorce (approved):

PR #1911, opened 2025-04-21
fix api server Burn output and tx inspect freezable

OBorce:
- Fix Burn output json helpers type
- Fix wallet CLI tx summary for token freezable status


ImplOfAnImpl (approved):
Let's add some tests for the fixes

PR #1914, opened 2025-04-25
Make checkpoints maps static; update checkpoints

ImplOfAnImpl:
Maps that hold checkpoints data are now lazy_static, to prevent them from being constructed every time a `ChainConfig` is constructed (which happens e.g. every time a function from wasm bindings is called).

I also updates the checkpoints themselves to the latest values.


common/src/chain/config/checkpoints.rs
```diff
@@ -59,22 +93,78 @@ impl Checkpoints {
             .expect("Genesis must be there, at least.");
         (*cp_before.0, (*cp_before.1))
```
OBorce, line 94: we can remove the above exact match get() and just use range(..=height)

OBorce (approved):

common/src/chain/config/checkpoints.rs
ImplOfAnImpl: Done

PR #1935, opened 2025-07-01
Fix compiler error from Rust 1.87, pacify Clippy 1.88

ImplOfAnImpl:


clippy.toml
```diff
@@ -2,3 +2,4 @@ disallowed-methods = []
 disallowed-types = []
 enforced-import-renames = []
 large-error-threshold = 256
+enum-variant-size-threshold = 256
```
ImplOfAnImpl, line 5: FYI: for some reason Clippy started to complain about "large size difference between variants" for multiple enums, even though the default value for `enum-variant-size-threshold` doesn't seem to have changed between the compiler versions. I decided to set it to 256 explicitly, for consistency with `large-error-threshold`, this silenced the complaints.

serialization/core/tests/simple_types.rs
```diff
@@ -201,7 +201,7 @@ fn test_scale_options() {
     assert_eq!(dec, Some(OptionWrapper::new(result)));
 
     // Decode and encode 1_048_576 chars 'X'
-    let result = Some(format!("!{:X<4194304}!", ""));
+    let result = Some(std::iter::repeat_n('X', 1_048_576).collect::<String>());
```
ImplOfAnImpl, line 204: FYI: this was a hard error (probably since 1.87, the version I've skipped) - `integer out of range for u16 in format string`

test/runner/functional.rs
```diff
@@ -70,7 +70,11 @@ where
 fn find_python_exe() -> PathBuf {
     let possible_python_execs = ["python3", "python"];
 
-    let file_suffix = (env::consts::OS == "windows").then_some(".exe").unwrap_or_default();
+    let file_suffix = if env::consts::OS == "windows" {
```
ImplOfAnImpl, line 73: FYI: the complaint was:
```
error: this method chain can be written more clearly with `if .. else ..`
  --> test/runner/functional.rs:73:23
   |
73 |     let file_suffix = (env::consts::OS == "windows").then_some(".exe").unwrap_or_default();
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `if env::consts::OS == "windows" { ".exe" } else { Default::default() }`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#obfuscated_if_else
   = note: `-D clippy::obfuscated-if-else` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::obfuscated_if_else)]`
```

Same for the similar changes in the wallet.

PR #1940, opened 2025-07-22
Tests for compose_transaction and tx_to_partially_signed_tx

ImplOfAnImpl:
1. Tests for `tx_to_partially_signed_tx` and `compose_transaction` were added in `wallet-controller`.
2. `tx_to_partially_signed_tx` now accepts a list of HTLC secrets (though it's only used in tests for now).
3. `compose_transaction` had a bug where `HtlcSpendingCondition::WithSecret` would always be used even if None's were passed for the htlc secrets; this is fixed now.
4. A few utility functions from wallet tests were moved to `wallet/src/wallet/test_helpers.rs`, which is a test-only module that lives in production code. This is ugly, but there is no good alternative when we want to re-use the same code in unit tests of other components (and we already use this approach in p2p).
5. Some minor cleanup here and there.
6. I also removed `Model::TrezorLegacy` from Trezor model lists when searching for a device, because we don't really support it.

### DIFF
--- a/wallet/wallet-controller/src/lib.rs
+++ b/wallet/wallet-controller/src/lib.rs
@@ -21,6 +21,8 @@ pub mod read;
 mod runtime_wallet;
 mod sync;
 pub mod synced_controller;
+#[cfg(test)]
+mod tests;
 pub mod types;
 
 const NORMAL_DELAY: Duration = Duration::from_secs(1);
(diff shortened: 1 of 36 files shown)


wallet/wallet-controller/src/lib.rs
```diff
@@ -1138,12 +1144,10 @@ where
...
-                    let htlc_spending =
-                        htlc_secrets.as_ref().map_or(HtlcSpendingCondition::Skip, |secrets| {
-                            secrets.get(i).map_or(HtlcSpendingCondition::WithMultisig, |_| {
-                                HtlcSpendingCondition::WithSecret
-                            })
-                        });
+                    let htlc_spending = HtlcSpendingCondition::from_opt_secrets_array_item(
+                        htlc_secrets.as_deref(),
+                        i,
+                    );
```
ImplOfAnImpl, line 1151: FYI: this is actually a fix and not just cleanup - in the old code the `_` parameter of the inner closure would have the type `&Option<HtlcSecret>` (due to `secrets.get` returning `Option` of `Option`), so we would always use `HtlcSpendingCondition::WithSecret` when Some(Vec) was passed here for `htlc_secrets`.

OBorce (approved):

PR #1944, opened 2025-07-30
Feature/ledger signer

OBorce:


ImplOfAnImpl (commented):
I haven't dug into the code much yet, will continue next week.

> Tests assume the ledger app repo is cloned next to this one with name ledger-mintlayer

To be honest, I'm not a huge fan of this approach. And also of the fact that the emulator is always started automatically. E.g. in the Trezor case it was sometimes useful to see the emulator logs to understand what went wrong.
Was there any particular reason to do it this way instead of expecting the emuator to be running?

.github/workflows/build.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Update local dependency repositories
         run: sudo apt-get update
       - name: Install build dependencies
-        run: sudo apt-get install -yqq --no-install-recommends build-essential python3 python3-toml podman pkg-config libssl-dev
+        run: sudo apt-get install -yqq --no-install-recommends build-essential python3 python3-toml podman pkg-config libssl-dev libdbus-1-dev libusb-1.0-0-dev
```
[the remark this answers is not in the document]
OBorce: Just for compilation

.github/workflows/code_checks.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Update local dependency repositories
         run: sudo apt-get update
       - name: Install dependencies
-        run: sudo apt-get install -yqq --no-install-recommends build-essential python3 python3-toml
+        run: sudo apt-get install -yqq --no-install-recommends build-essential python3 python3-toml libdbus-1-dev libusb-1.0-0-dev
```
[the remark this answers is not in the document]
OBorce: Just for compilation

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1095 @@
...
+    async fn check_session(
+        &mut self,
+        db_tx: &impl WalletStorageReadLocked,
+        key_chain: &impl AccountKeyChains,
+    ) -> SignerResult<()> {
+        let mut client = self.client.lock().await;
+
+        loop {
+            match get_app_name(&mut *client).await {
+                Ok(_) => return Ok(()),
```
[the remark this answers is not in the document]
OBorce: I did not find any way to get the device name except through Bluetooth connection.

wallet/src/signer/ledger_signer/tests.rs
```diff
@@ -0,0 +1,398 @@
...
+#[trace]
+#[serial_test::serial]
+#[case(Seed::from_entropy())]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_sign_message(
+    #[case] seed: Seed,
+    #[values(
+        MessageToSign::Random,
+        // Special case: an "overlong" utf-8 string (basically, the letter 'K' encoded with 2 bytes
+        // instead of 1). The firmware used to have troubles with this.
```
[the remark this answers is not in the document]
OBorce: Ledger also had issues with displaying some utf-8 characters.

ImplOfAnImpl (changes requested):

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1094 @@
...
+                Err(ledger_lib::Error::Timeout) => {
+                    continue;
+                }
```
ImplOfAnImpl: I don't think it's a good idea to ignore timeouts indefinitely. Perhaps we should only allow a few timeouts and then fail?
But why do we even need to handle a timeout differently that other connection errors that a re handled below?

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1094 @@
...
+    ) -> Self {
+        Self {
+            chain_config,
+            client,
+            sig_aux_data_provider: std::sync::Mutex::new(sig_aux_data_provider),
+            provider,
+        }
+    }
+
+    /// Calls initialize on the device with the current session_id.
```
ImplOfAnImpl: There is neither `initialize` nor `session_id` in the Ledger case.

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,303 @@
...
+}
+
+/// Check that the response ends with the OK status code and return the rest of the response back
+pub fn ok_response(mut resp: Vec<u8>) -> SignerResult<Vec<u8>> {
+    let (_, status_code) = resp.split_last_chunk().ok_or(LedgerError::InvalidResponse)?;
+    let response_status = u16::from_be_bytes(*status_code);
+
+    ensure!(
+        response_status == OK_RESPONSE,
+        LedgerError::ErrorResponse(response_status)
```
ImplOfAnImpl: Let's parse the status code (e.g. using ledger-proto's `StatusCode`, which implements `TryFromPrimitive`) and provide a more informative error instead of this `ErrorResponse(u16)`. E.g. currently I'm getting "Ledger error: Received an error response from the Ledger device: 21781" when trying to connect to a locked device, this is not very informative.

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,326 @@
...
+pub async fn get_app_name<L: Exchange>(ledger: &mut L) -> Result<Vec<u8>, ledger_lib::Error> {
+    let msg_buf = [CLA, Ins::APP_NAME, 0, P2::DONE];
+    ledger.exchange(&msg_buf, Duration::from_millis(100)).await
+}
+
+#[allow(dead_code)]
+pub async fn check_current_app<L: Exchange>(ledger: &mut L) -> SignerResult<()> {
+    let resp = get_app_name(ledger)
+        .await
+        .map_err(|err| LedgerError::DeviceError(err.to_string()))?;
```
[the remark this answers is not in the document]
ImplOfAnImpl: > For some reason it always returns "app" for app name and the OS version instead of the opened app. So, I kept our own instructions.

Well, I tried sending [b0, 1, 0, 0] to my NanoSPlus and got "Mintlayer"/"0.1.0" for our app and "Ethereum"/"1.18.0" for Ethereum.
What device are you using? Can you double check?

In any case, this has to be investigated further. If b0/01 doesn't work indeed, we must at least document that fact, mentioning the particular situations.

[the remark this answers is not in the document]
ImplOfAnImpl: > For some reason it always returns "app"

So I suppose you were using the emulator, for which this is [documented behavior](https://speculos.ledger.com/user/usage.html#app-name-and-version).

In such a case IMO it's better to detect if we're using an emulator (or I guess we could assume that any tcp transport corresponds to an emulator) and skip the app name/version check in that case.

Alternatively, we could go with with a custom APDU. But in this case it has to be handled more carefully. E.g. a) expect that the whole APDU may fail, e.g. with ClaNotSupported or InsNotSupported (because the current app doesn't support it), or with something like WrongP1P2/WrongApduLength (because in the current app this APDU means something else and has different parameters); b) expect that APDU may be handled successfully, but the result is garbage (again, because in the current app this APDU means something else).

The first approach looks easier to implement.

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,321 @@
...
+    H256 as LH256, P1_APP_NAME, P1_GET_VERSION, P1_SIGN_NEXT, P1_SIGN_START, P2_DONE, P2_SIGN_MORE,
+};
+
+const MAX_ADPU_LEN: usize = (u8::MAX - 5) as usize; // 4 bytes for the header + 1 for len
+const TIMEOUT_DUR: Duration = Duration::from_secs(100);
+const OK_RESPONSE: u16 = 0x9000;
+const TX_VERSION: u8 = 1;
+
+#[derive(Decode)]
+pub struct LedgerSignature {
```
ImplOfAnImpl: This is also part of the protocol, why isn't it in the corresponding crate?

node-gui/backend/src/error.rs
```diff
@@ -46,8 +46,8 @@ pub enum BackendError {
     ColdWallet,
     #[error("Cannot interact with a hot wallet when in Cold wallet mode")]
     HotNotSupported,
-    #[error("Cannot use a Trezor wallet in a Cold wallet mode")]
-    ColdTrezorNotSupported,
+    #[error("Cannot use a Hardware wallet in a Cold wallet mode")]
```
ImplOfAnImpl: It's looks a bit strange to capitalize "Hardware".

wallet/src/signer/ledger_signer/speculos/handle.rs
```diff
@@ -0,0 +1,130 @@
...
+            .send()
+            .await?;
+
+        log::debug!("Button request complete: {}", r.status());
+
+        Ok(())
+    }
+}
+
+/// Handle to a Speculos instance running under Podman
```
ImplOfAnImpl: There is nothing specific to Podman here.

Also, IMO having `trait Handle` and its impl is now redundant, `Handle` can just become a concrete struct.

wallet/src/signer/ledger_signer/speculos/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+//
+// https://github.com/mintlayer/mintlayer-core/blob/master/LICENSE
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Rust wrapper for executing Speculos via podman,
```
ImplOfAnImpl: Nothing is related to Podman here.

Also, IMO the whole `speculos` dir can be collapsed into a single `speculos.rs`

wallet/src/signer/ledger_signer/tests.rs
```diff
@@ -0,0 +1,414 @@
...
+
+#[rstest]
+#[trace]
+#[serial_test::serial]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_app_name() {
+    let mut transport = TcpTransport::new().unwrap();
+    let mut device = transport
+        .connect(TcpInfo {
+            addr: SocketAddr::new(std::net::IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 9999),
```
ImplOfAnImpl: Plz don't duplicate the port and the address, factor out the connection creation instead.

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,304 @@
...
+    Bip32Path as LedgerBip32Path, CoinType, GetPublicKeyRespones, GetVersionRespones,
+    InputAdditionalInfoReq, Ins, MsgSignature, OutputValue as LOutputValue, P1SignTx, PubKeyP1,
+    PublicKeyReq, SignMessageReq, SignTxReq, Signature as LedgerSignature, TxInput as LTxInput,
+    TxInputReq, TxMetadataReq, TxOutput as LTxOutput, TxOutputReq, APDU_CLASS, H256 as LH256,
+    P1_APP_NAME, P1_GET_VERSION, P1_SIGN_NEXT, P1_SIGN_START, P2_DONE, P2_SIGN_MORE,
+};
+
+const MAX_ADPU_LEN: usize = (u8::MAX - 5) as usize; // 4 bytes for the header + 1 for len
+const TIMEOUT_DUR: Duration = Duration::from_secs(100);
+const OK_RESPONSE: u16 = 0x9000;
```
ImplOfAnImpl: This should come from mintlayer_ledger_messages

ImplOfAnImpl (changes requested):

common/src/chain/transaction/account_outpoint.rs
```diff
@@ -61,16 +61,22 @@ impl From<&AccountCommand> for AccountType {
     }
 }
 
-impl From<OrderAccountCommand> for AccountType {
-    fn from(cmd: OrderAccountCommand) -> Self {
-        match cmd {
+impl OrderAccountCommand {
+    pub fn order_id(&self) -> OrderId {
```
ImplOfAnImpl: Return a reference instead?

wallet/src/wallet/mod.rs
```diff
@@ -269,8 +269,10 @@ pub enum WalletError {
         "A VRF public key must be specified when creating a staking pool using a hardware wallet"
     )]
     VrfKeyMustBeProvided,
-    #[error("Cannot change a Trezor wallet type")]
-    CannotChangeTrezorWalletType,
+    #[error("Cannot change a {from} wallet type to {to}")]
+    CannotChangeWalletType { from: WalletType, to: WalletType },
+    #[error("Cannot change a Ledger wallet type")]
+    CannotChangeLedgerWalletType,
```
ImplOfAnImpl: Unused?

do_checks.sh
```diff
@@ -14,7 +14,7 @@ cargo fmt --check -- --config newline_style=Unix
 # Note: "--allow duplicate" silences the warning "found x duplicate entries for crate y".
 cargo deny check --allow duplicate --hide-inclusion-graph
 
-cargo vet check --locked
+# cargo vet check --locked
```
ImplOfAnImpl: Don't forget to uncomment this

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1306 @@
...
+    #[error("Received an invalid response from the Ledger device")]
+    InvalidResponse,
+    #[error("Received an error response from the Ledger device: {0}")]
+    ErrorResponse(String),
+    #[error("Device error: {0}")]
+    DeviceError(String),
+    #[error("Missing hardware wallet data in database")]
+    MissingHardwareWalletData,
+    #[error("Derivation path is too long to send to Ledger")]
+    PathToLong,
```
ImplOfAnImpl: Unused?

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1306 @@
...
+    #[error("Invalid public key returned from Ledger")]
+    InvalidKey,
+    #[error("The file being loaded is a software wallet and does not correspond to the connected hardware wallet")]
```
ImplOfAnImpl: Same comment about the "and does not correspond to the connected hardware wallet" part.

wallet/src/signer/ledger_signer/speculos.rs
```diff
@@ -0,0 +1,233 @@
...
+    NanoX,
+    Stax,
+    Flex,
+}
+
+impl Device {
+    /// Returns true if the device has a touch screen
+    pub fn is_touch(&self) -> bool {
+        matches!(self, Device::Stax | Device::Flex)
+    }
```
ImplOfAnImpl: NanoGen5 also has it

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,367 @@
...
+        give_balance: to_ledger_amount(&info.give_balance),
+    }
+}
+
+pub fn to_ledger_output_value(value: &chain::output_value::OutputValue) -> LOutputValue {
+    match value {
+        chain::output_value::OutputValue::Coin(amount) => {
+            LOutputValue::Coin(to_ledger_amount(amount))
+        }
+        chain::output_value::OutputValue::TokenV0(_) => panic!("unsupported V0"),
```
ImplOfAnImpl: Let's not panic and return an error instead (e.g. UnsupportedTokenV0)

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,367 @@
...
+    ledger_decode_all(value.encode().as_slice()).expect("ok")
+}
+
+pub fn to_ledger_account_command(value: &chain::AccountCommand) -> LAccountCommand {
+    ledger_decode_all(value.encode().as_slice()).expect("ok")
+}
+
+pub fn to_ledger_order_account_command(value: &chain::OrderAccountCommand) -> LOrderAccountCommand {
+    ledger_decode_all(value.encode().as_slice()).expect("ok")
+}
```
ImplOfAnImpl: This is not great. Perhaps we should move the code from `common/tests/primitives_repo_consistency/utils/converters.rs` from tests to the `common` crate itself and use it here.But note that we'll need to get rid of the panic for TokenV0 inside the converters and return an error instead. Which means that `ConvertFrom/convert_from` should become `TryConvertFrom/try_convert_from` and return a `Result`. The error type for the result should IMO be a separate type containing only one error variant.

Also, the "L" aliases for the primitives repo types are not needed IMO (or at least they should be "P", because they are specific not to the Ledger, but to the primitives repo).

wallet/src/wallet/mod.rs
```diff
@@ -699,12 +701,32 @@ where
...
+                    from: current_wallet_type,
+                    to: wallet_type,
+                })
             }
             (WalletType::Cold, WalletType::Cold) => {}
             (WalletType::Hot, WalletType::Hot) => {}
             #[cfg(feature = "trezor")]
             (WalletType::Trezor, WalletType::Trezor) => {}
+            #[cfg(feature = "trezor")]
+            (WalletType::Ledger, WalletType::Ledger) => {}
```
ImplOfAnImpl: feature = "ledger"?

common/src/primitives_converters.rs
```diff
@@ -0,0 +1,521 @@
...
+            give: value.give().clone().try_convert_into()?,
+        })
+    }
+}
+
+impl TryConvertFrom<VRFPublicKey> for ml_primitives::VrfPublicKey {
+    fn try_convert_from(value: VRFPublicKey) -> Result<Self, PrimitivesConvertersError> {
+        match value.kind() {
+            VRFKeyKind::Schnorrkel => {
+                let key: SchnorrkelPublicKey = value.try_into().unwrap();
```
ImplOfAnImpl: Return an error instead of unwrapping

.github/workflows/build.yml
```diff
@@ -281,13 +281,109 @@ jobs:
...
+          path: ./mintlayer-core
+
+      - name: Update local dependency repositories
+        run: sudo apt-get update
+
+      - name: Install build dependencies
+        run: sudo apt-get install -yqq --no-install-recommends build-essential pkg-config libdbus-1-dev libusb-1.0-0-dev
+
+      - name: Install rust
+        run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
ImplOfAnImpl: Plz do it the same way as in run_tests_on_trezor_preparation (use Rust version provided by extract.py)

.github/workflows/build.yml
```diff
@@ -281,13 +281,109 @@ jobs:
...
+
+      - name: Run Ledger emulator and execute tests
+        run: |
+          set -e
+
+          sudo docker run -d --rm --name ledger-emulator \
+            -v "$(realpath ./mintlayer-ledger-app):/app" \
+            --publish 5000:5000 --publish 9999:9999 \
+            ghcr.io/ledgerhq/ledger-app-builder/ledger-app-dev-tools:latest \
+            sh -c 'speculos --apdu-port 9999 --api-port 5000 --display headless -s "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"  target/${{ matrix.model }}/release/mintlayer-app'
```
ImplOfAnImpl: Plz wrap the lines to make their length reasonable

common/src/primitives_converters.rs
```diff
@@ -0,0 +1,521 @@
...
+    }
+}
+
+impl TryConvertFrom<TxInput> for ml_primitives::TxInput {
+    fn try_convert_from(value: TxInput) -> Result<Self, PrimitivesConvertersError> {
+        match value {
+            TxInput::Utxo(utxo) => Ok(Self::Utxo(utxo.try_convert_into()?)),
+            TxInput::Account(acc) => Ok(Self::Account(acc.try_convert_into()?)),
+            TxInput::AccountCommand(nonce, command) => Ok(Self::AccountCommand(
+                ml_primitives::AccountNonce(nonce.value()),
```
ImplOfAnImpl: Use `try_convert_into` on the nonce directly

wallet/src/signer/ledger_signer/tests/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+#[rstest_reuse::apply(sign_message_test_params)]
+#[rstest]
+#[trace]
+#[serial_test::serial]
+#[case(Seed::from_entropy())]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+async fn test_sign_message(#[case] seed: Seed, message_to_sign: MessageToSign) {
+    use crate::signer::tests::generic_tests::test_sign_message_generic;
+
+    log::debug!("test_sign_transaction_intent, seed = {seed:?}");
```
ImplOfAnImpl: test_sign_message

wallet/src/signer/ledger_signer/tests/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    let info = device.app_info(Duration::from_millis(500)).await.unwrap();
+
+    let err = check_current_app(&mut device).await.unwrap_err();
+    eprintln!("info: {err:?}");
+    assert_eq!(
+        err,
+        SignerError::LedgerError(LedgerError::DifferentActiveApp(info.name))
+    )
```
ImplOfAnImpl: This will have to be changed to a positive test after upgrading to ledger sdk 1.31

wallet/src/signer/ledger_signer/tests/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+use async_trait::async_trait;
+use ledger_lib::{
+    transport::{TcpDevice, TcpInfo, TcpTransport},
+    Device as _, Transport,
+};
+use mintlayer_ledger_messages::CoinType;
+use randomness::make_true_rng;
+use rstest::rstest;
+use serialization::hex::HexEncode;
+use test_utils::random::{make_seedable_rng, Seed};
```
ImplOfAnImpl: Plz move workspace imports to the other group below

wallet/src/signer/ledger_signer/tests/speculos.rs
```diff
@@ -0,0 +1,239 @@
...
+    NanoX,
+    Stax,
+    Flex,
+    NanoGen5,
+}
+
+impl Device {
+    /// Returns true if the device has a touch screen
+    pub fn is_touch(&self) -> bool {
+        matches!(self, Device::Stax | Device::Flex | Device::NanoGen5)
```
ImplOfAnImpl: Plz use the actual `match`, so that compilation fails if a new enum item is added

wallet/src/signer/ledger_signer/tests/speculos.rs
```diff
@@ -0,0 +1,239 @@
...
+            (ScreenElement::ReviewTap, Device::Stax) => (335, 606),
+            (ScreenElement::ReviewTap, Device::Flex) => (430, 530),
+            (ScreenElement::ReviewTap, Device::NanoGen5) => (295, 370),
+
+            (ScreenElement::ReviewConfirm, Device::Stax) => (335, 515),
+            (ScreenElement::ReviewConfirm, Device::Flex) => (240, 435),
+            (ScreenElement::ReviewConfirm, Device::NanoGen5) => (290, 335),
+
+            // Fallback or unimplemented combinations
+            _ => panic!("Coordinate not mapped for {:?} on {:?}", self, device),
```
ImplOfAnImpl: It's better to have the explicit match `(_, Device::NanoS) | (_, Device::NanoSPlus) | (_, Device::NanoX)`

wallet/src/signer/ledger_signer/tests/speculos.rs
```diff
@@ -0,0 +1,239 @@
...
+        // Flex: 480 x 600
+        match (self, device) {
+            // --- UseCaseReview ---
+            (ScreenElement::ReviewTap, Device::Stax) => (335, 606),
+            (ScreenElement::ReviewTap, Device::Flex) => (430, 530),
+            (ScreenElement::ReviewTap, Device::NanoGen5) => (295, 370),
+
+            (ScreenElement::ReviewConfirm, Device::Stax) => (335, 515),
+            (ScreenElement::ReviewConfirm, Device::Flex) => (240, 435),
+            (ScreenElement::ReviewConfirm, Device::NanoGen5) => (290, 335),
```
ImplOfAnImpl: Btw, where did you get these numbers from? I assumed it was from here - https://github.com/LedgerHQ/ragger/blob/master/src/ragger/firmware/touch/positions.py - but the coordinates differ. Is there any reason for using different coordinates? If yes, the reason should be documented. If not, we should probably use the same coordinated as `rugger` and also mention the `rugger`'s source file where they are defined.

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,301 @@
...
+use utils::ensure;
+
+use ledger_lib::{Device, Exchange};
+use ledger_proto::StatusCode;
+use mintlayer_ledger_messages::{
+    decode_all as ledger_decode_all, encode as ledger_encode, AddrType, Apdu,
+    Bip32Path as LedgerBip32Path, CoinType, GetPublicKeyRespones, Ins, MsgSignature, PubKeyP1,
+    PublicKeyReq, SighashInputCommitment, SignMessageReq, SignP1, SignTxReq, Signature, TxInputReq,
+    TxMetadataReq, TxOutputReq, APDU_CLASS, P2_DONE, P2_MORE,
+};
```
ImplOfAnImpl: Perhaps it's better to import the entire `mintlayer_ledger_messages` e.g. as `ledger_msg` and use qualified names like `ledger_msg::Bip32Path`? Otherwise it looks inconsistent.

wallet/src/signer/ledger_signer/ledger_messages.rs
```diff
@@ -0,0 +1,301 @@
...
+    has_more_signatures: bool,
+}
+
+/// Check that the response ends with the OK status code and return the rest of the response back
+pub fn ok_response(mut resp: Vec<u8>) -> SignerResult<Vec<u8>> {
+    let (_, status_code) = resp.split_last_chunk().ok_or(LedgerError::InvalidResponse)?;
+    let response_status = u16::from_be_bytes(*status_code);
+
+    let code = StatusCode::try_from(response_status)
+        .map_err(|_| LedgerError::ErrorResponse(format!("Unknown error: {response_status}")))?;
```
ImplOfAnImpl: `StatusCode` from rust-ledger doesn't contain app-specific codes, so we shouldn't use it IMO. Instead, we should put app's `StatusWord` to the `messages` crate and use it here. The error messages corresponding to enum variants should ideally be near the enum itself (e.g. you could use `derive_more::Display` or `displaydoc::Display`, both of which AFAIK supports no_std).

But it also makes sense to look at rust-ledger's `StatusCode`

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1336 @@
...
+        ChainType::Regtest => CoinType::Signet,
+    }
+}
+
+async fn find_ledger_device() -> SignerResult<(LedgerHandle, LedgerFullInfo)> {
+    let mut provider = LedgerProvider::init().await;
+    let mut devices = provider
+        .list(Filters::Any)
+        .await
+        .map_err(|err| LedgerError::DeviceError(err.to_string()))?;
```
ImplOfAnImpl: LedgerError already has a From implementation for rust-ledger's Error.

wallet/src/signer/ledger_signer/tests/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+                } else {
+                    // BUTTON DEVICE STRATEGY (Nano S/S+/X)
+                    // 1. Press Right to scroll
+                    // 2. Press Both to confirm
+                    let _ = handle.button(Button::Right, ButtonAction::PressAndRelease).await;
+                    sleep(Duration::from_millis(100)).await;
+                    let _ = handle.button(Button::Both, ButtonAction::PressAndRelease).await;
+                    sleep(Duration::from_millis(100)).await;
+                    let _ = handle.button(Button::Both, ButtonAction::PressAndRelease).await;
+                }
```
ImplOfAnImpl: https://github.com/mintlayer/mintlayer-core/pull/1944#discussion_r2499187307

> t doesn't seem like the second Button::Both press is needed, even though the tests indeed fail if the second one is commented out. Can it be due to the fact that the device has too little time to register the first press? If so, perhaps there should just be a delay after "Button::Right" is pressed?

The question was about why the second "Button::Both" press is needed and whether it can/should be avoided by introducing a delay after "Button::Right".
Instead of looking into the question, you just added delays. What for? Is it the "just do something to make the reviewer bugger off" tactics?

P.S. I've now noticed that the second Button::Both is effectively used to stop signing after the screen "Press right button to continue message or press both to skip" is shown, because if we continue, we'll reject the tx instead of signing it.

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1336 @@
...
+            client,
+            sig_aux_data_provider: std::sync::Mutex::new(sig_aux_data_provider),
+            provider,
+        }
+    }
+
+    /// Tries to confirm the running app name is still matching Mintlayer app
+    /// Also waits after a signing operation for the device to start listening to new instructions
+    ///
+    /// If the operation fails due to an USB error (which may indicate a lost connection to the device),
```
ImplOfAnImpl: "a USB"

wallet/src/signer/ledger_signer/mod.rs
```diff
@@ -0,0 +1,1336 @@
...
+    /// the function will attempt to reconnect to the Ledger device once before returning an error.
+    async fn check_session<T: WalletStorageReadLocked + Send>(
+        &mut self,
+        db_tx: &mut T,
+        key_chain: &impl AccountKeyChains,
+    ) -> SignerResult<()> {
+        let mut client = self.client.lock().await;
+        // Try and wait around 50 * 100ms for the screen to clear after a signing operation ends
```
ImplOfAnImpl: 50 * TIMEOUT_DUR

ImplOfAnImpl (approved):
I don't want to review this again, so I suggest merging it after applying the comments. The ledger tests will have to be disabled on CI, but it's better to enable the code itself, to make sure it compiles.

**Plz make sure that all the comments have been addressed before merging**

PR #1947, opened 2025-08-01
Force fontconfig-parser version to v0.5.7

ImplOfAnImpl:
Today I discovered that when built from master, node-gui UI looks like this on my Ubuntu machine:

node_gui_broken1
node_gui_broken2
node_gui_broken3

The first bad commit is 8dbc6b94a8243f4e3de13c0fcce64425f5065eb1 where the only significant changes are those in `Cargo.lock`. I managed to narrow it down to the `fontconfig-parser` crate which was updated from v0.5.7 to v0.5.8 in that commit.
So in this PR I forcibly set the version back to 0.5.7 via `[patch]`. This is a hacky fix though, ideally we should find the real culprit (which may not be `fontconfig-parser` itself) and at least file a bug report.

Update: it seems that the issue is specific to MATE desktop. I.e. I was able to reproduce it on a fresh install of Ubuntu MATE, but not on a fresh install of the vanilla Ubuntu.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2752,9 +2752,8 @@ dependencies = [
 
 [[package]]
 name = "fontconfig-parser"
-version = "0.5.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bbc773e24e02d4ddd8395fd30dc147524273a83e54e0f312d986ea30de5f5646"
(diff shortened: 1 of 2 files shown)


OBorce (approved):
at least on ubuntu WSL the master version works fine

PR #1949, opened 2025-08-05
Add ledger support in CLI and GUI wallet

OBorce:


node-gui/src/main_window/main_menu.rs
```diff
@@ -125,6 +125,27 @@ fn make_menu_file<'a>(wallet_mode: WalletMode) -> Item<'a, MenuMessage, Theme, i
...
+                #[cfg(feature = "ledger")]
+                {
+                    menu.push(menu_item(
+                        "(Beta) Create new Ledger wallet",
+                        MenuMessage::CreateNewWallet {
+                            wallet_type: WalletType::Ledger,
+                        },
+                    ));
+                    menu.push(menu_item(
+                        "(Beta) Recover from Ledger wallet",
```
ImplOfAnImpl: Btw, in the actual menu the word "wallet" is not visible. I guess we have to increase the width of the menu

PR #1953, opened 2025-08-15
Prettify wallet-cli help; require mnemonic during wallet recovery

ImplOfAnImpl:
1) Previously, in the output of the `help` command, the docs for individual commands were sometimes separated from each other with an empty line (when the doc was taken from RPC) and sometimes not (when it was generated by clap from the comamnd itself).
Now they are always separated by an empty line.
2) Docs taken from RPC used to preserve all original EOLs, and the clap ones didn't. Now the RPC docs are "clapified" before being re-used for CLI.
3) Docs from RPC were put into the "about" part entirely, so they were printed fully in the output of the `help` command; the clap docs only put the first paragraph into "about", and to see the entire doc ("long_about") you need to run "help the_command".
Now the docs from RPC also have the "about"/"long_about" parts.
4) In the output of the `help` command, all commands are now sorted by name, except for "help", which comes at the end.
5) I re-reviewed the output of each "help command" and reformulated some of them.
Many commands now have a shorter first paragraph (which goes to "about" and which is shown in the output of "help"). So, the output of "help" is now less cluttered, but the user will have to run "help command" or "command --help" to see the full help (btw, "command -h" still prints the short version).

Also,
1) The min Rust version was bumped to 1.88, this is because the latest version of `cargo deny` (which we always install on our CI) now requires at least 1.88.
2) During software wallet recovery the mnemonic is now not optional.

### DIFF
--- a/wallet/wallet-cli-commands/src/lib.rs
+++ b/wallet/wallet-cli-commands/src/lib.rs
@@ -55,15 +55,18 @@ pub enum CreateWalletSubCommand {
         /// File path of the wallet file
         wallet_path: PathBuf,
 
-        /// Specifies whether the seed-phrase should be stored in the wallet file or
+        /// Specifies whether the seed phrase should be stored in the wallet file or
         /// only printed on the screen.
         ///
-        /// Not storing the seed-phrase can be seen as a security measure
-        /// to ensure sufficient secrecy in case that seed-phrase is reused
+        /// Not storing the seed phrase can be seen as a security measure
+        /// to ensure sufficient secrecy in case that the seed phrase is reused
         /// elsewhere if this wallet is compromised.
+        ///
+        /// Note: if you decide to store the seed phrase, consider encrypting the wallet with
+        /// the wallet-encrypt-private-keys command, which will also encrypt the seed phrase.
         whether_to_store_seed_phrase: CliStoreSeedPhrase,
 
-        /// Mnemonic phrase (12, 15, or 24 words as a single quoted argument).
+        /// Mnemonic (seed) phrase (12, 15, or 24 words as a single quoted argument).
         ///
         /// If not specified, a new mnemonic phrase will be generated and printed.
         mnemonic: Option<String>,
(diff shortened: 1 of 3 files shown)


wallet/wallet-cli-commands/src/errors.rs
```diff
@@ -20,7 +20,7 @@ use wallet_controller::types::GenericCurrencyTransferToTxOutputConversionError;
 use wallet_rpc_client::{handles_client::WalletRpcHandlesClientError, rpc_client::WalletRpcError};
 use wallet_rpc_lib::RpcError;
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, derive_more::Debug)]
```
ImplOfAnImpl, line 23: FYI: the standard `Debug` macro requires `N` to implement `Debug` as well, so basically it never works. `derive_more::Debug` is more clever and puts no requirements on `N`.

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -1183,30 +1217,113 @@ pub fn get_repl_command(cold_wallet: bool, mutable_wallet: bool) -> Command {
         repl_command
     };
 
-    // Customize the help template for all commands to make it more REPL friendly
+    lazy_static::lazy_static! {
+        static ref CUSTOM_RPC_TO_CLI_NAME_MAPPINGS: BTreeMap<&'static str, &'static str> =
+            BTreeMap::from([("account_extended_public_key", "account-extended-public-key-as-hex")]);
+    }
```
ImplOfAnImpl, line 1223: Note: previously this wasn't needed because we allowed one name to be a prefix of the other (due to the usage of `zip` during comparison). Now we require the names to be identical (except for the separator)

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -960,12 +986,19 @@ pub enum WalletCommand {
...
+        /// A utxo can be from a transaction output or a block reward output:
         /// e.g tx(000000000000000000059fa50103b9683e51e5aba83b8a34c9b98ce67d66136c,1) or
         /// block(000000000000000000059fa50103b9683e51e5aba83b8a34c9b98ce67d66136c,2)
         #[arg(long = "utxos", default_values_t = Vec::<String>::new())]
         utxos: Vec<String>,
 
+        /// This specifies that instead of a (hex encoded) PartiallySignedTransaction
+        /// the result should be a (hex encoded) "simple" transaction.
+        ///
+        /// Note that both variants are accepted by account-sign-raw-transaction,
```
OBorce, line 1000: unless signing with a cold wallet which will need the partially signed tx format as it cannot fetch the utxos.

OBorce (approved):

wallet/wallet-cli-commands/src/lib.rs
```diff
@@ -623,7 +638,7 @@ pub enum WalletCommand {
     /// the original multisig address.
     ///
     /// The utxos to pay fees from will be selected automatically; these will be normal, single-sig utxos.
-    /// The optional `fee_change_address` specifies the destination for the change for the fee payment;
+    /// The optional "fee change address" specifies the destination for the change for the fee payment;
     /// If it's unset, the destination will be taken from one of existing single-sig utxos.
```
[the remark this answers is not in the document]
ImplOfAnImpl: > not sure how clear single-sig utxos is to most users, maybe something like: one of the existing UTXOs that require only a single signature (not multisig)

This is a hidden command for internal use, so I'd leave it as is.

PR #1956, opened 2025-08-18
Add tx decode to js value to wasm lib

OBorce:
Add a new function to wasm lib to decode a signed transaction into a js value

### DIFF
--- a/wasm-wrappers/src/lib.rs
+++ b/wasm-wrappers/src/lib.rs
@@ -31,12 +31,14 @@
 
 use std::{num::NonZeroU8, str::FromStr};
 
+use wasm_bindgen::JsValue;
+
 use bip39::Language;
 use itertools::Itertools as _;
 use wasm_bindgen::prelude::*;
 
 use common::{
-    address::{pubkeyhash::PublicKeyHash, Address},
+    address::{dehexify::dehexify_all_addresses, pubkeyhash::PublicKeyHash, Address},
     chain::{
         block::timestamp::BlockTimestamp,
         classic_multisig::ClassicMultisigChallenge,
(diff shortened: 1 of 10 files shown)


wasm-wrappers/src/lib.rs
```diff
@@ -689,6 +691,23 @@ pub fn encode_transaction(inputs: &[u8], outputs: &[u8], flags: u64) -> Result<V
     Ok(tx.encode())
 }
 
+/// Decodes a signed transaction from its binary encoding into a Json string.
+#[wasm_bindgen]
+pub fn decode_signed_transaction_to_json_str(
```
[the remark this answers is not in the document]
ImplOfAnImpl: > Producing a JsValue from a serde Json value gives a JS Map and not an object as before. Map cannot be put through Json stringify for comparison in the test and accessing fields needs to go through the get() method... With a string you can always do a Json parse and get an object. Not sure what is the best tradeoff.

I've just checked and `js_sys::JSON::parse` seems to do the trick, i.e. it produces an object and not a map.

P.S. `js_sys::JSON::parse` returns `Result<JsValue, JsValue>`, let's not `expect` on it and define an Error variant for its result instead. The reason is (besides `expect`s being bad in general) that it's a complex call, where multiple things could go wrong (as I understand, it calls back into JS to produce a JS object and then returns a handle to it).

wasm-wrappers/src/lib.rs
```diff
@@ -31,12 +31,14 @@
 
 use std::{num::NonZeroU8, str::FromStr};
 
+use wasm_bindgen::JsValue;
```
ImplOfAnImpl, line 34: Move into the other `use wasm_bindgen` clause?

wasm-wrappers/src/lib.rs
```diff
@@ -689,6 +691,22 @@ pub fn encode_transaction(inputs: &[u8], outputs: &[u8], flags: u64) -> Result<V
     Ok(tx.encode())
 }
 
+/// Decodes a signed transaction from its binary encoding into a Json object.
```
ImplOfAnImpl, line 694: "into a JavaScript object"?

ImplOfAnImpl (approved):

PR #1959, opened 2025-08-19
fix standalone private keys encryption and usage

OBorce:
- fix encryption of the standalone private keys when a new password is created or changed
- fix watching/scanning for relevant public key destinations belonging to standalone keys (only public key has destinations used to work before)
- fix wallet balance to show spendable UTXOs belonging to standalone private keys.

### DIFF
--- a/wallet/src/wallet/tests.rs
+++ b/wallet/src/wallet/tests.rs
@@ -298,18 +298,19 @@ fn create_wallet(chain_config: Arc<ChainConfig>) -> DefaultWallet {
 }
 
 #[track_caller]
-fn create_block<B, P>(
+fn create_block_with_address_reward<B, P>(
     chain_config: &Arc<ChainConfig>,
     wallet: &mut Wallet<B, P>,
     transactions: Vec<SignedTransaction>,
     reward: Amount,
     block_height: u64,
+    address: Destination,
 ) -> (Address<Destination>, Block)
 where
     B: storage::Backend + 'static,
     P: SignerProvider,
 {
-    let address = wallet.get_new_address(DEFAULT_ACCOUNT_INDEX).unwrap().1;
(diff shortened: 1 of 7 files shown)


wallet/src/wallet/tests.rs
```diff
@@ -1251,6 +1275,112 @@ fn locked_wallet_cant_sign_transaction(#[case] seed: Seed) {
             .unwrap();
     }
 }
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn locked_wallet_standalone_keys(#[case] seed: Seed) {
```
ImplOfAnImpl, line 1282: 1) I'd say the emphasis here is on "standalone_keys" and not "locked_wallet".
2) Do we still need `test_add_standalone_private_key`? It looks like it does only a subset of what this test does. (I guess the only thing this test doesn't do is check the result of `wallet.get_transaction`, but we could add the check).
3) If it's still needed, let's put the two tests near each other.

ImplOfAnImpl (approved):

wallet/src/wallet/tests.rs
```diff
@@ -298,33 +298,55 @@ fn create_wallet(chain_config: Arc<ChainConfig>) -> DefaultWallet {
...
     block_height: u64,
-) -> (Address<Destination>, Block)
+    address: Destination,
+) -> Block
 where
     B: storage::Backend + 'static,
     P: SignerProvider,
 {
-    let address = wallet.get_new_address(DEFAULT_ACCOUNT_INDEX).unwrap().1;
+    let address = Address::new(chain_config, address).unwrap();
```
ImplOfAnImpl, line 313: You're constructing this only to call `into_object` later. Pass the `Destination` to `make_address_output` directly.

ImplOfAnImpl (approved):

PR #1962, opened 2025-08-20
Prepare release v1.1.0

ImplOfAnImpl:
I also slightly updated README.md, specifying where to find node-gui logs on Windows.

Also, FORK_HEIGHT constants for ORDERS_V1 and SIGHASH_INPUT_COMMITMENTS_V1 were merged into one, because it will be one fork, not two.

The fork height is 566060, which is supposed to be September 30, a few hours after the noon. Plz double-check this.


OBorce (approved):
LGTM

PR #1965, opened 2025-08-25
Fix release Dockerfile's base builder image - use the same distro as in the runner image

ImplOfAnImpl:
Note: I'm going to merge this into `release-v1.1.0`, because this fix is related to the release, but, on the other hand, doesn't change anything that would require a new patch release.


erubboli (approved):
👍

PR #2023, opened 2026-03-05
Fix/ledger ping

OBorce:
- Use the new ping command instead of the get public key
- Add a predefined message test with null and unprintable control characters


ImplOfAnImpl:
Superseded by https://github.com/mintlayer/mintlayer-core/pull/2084

PR #2029, opened 2026-03-24
Feature/api server mempool txs

OBorce:
API Server now scans the mempool after finishing syncing to the latest block.
New address and transaction endpoints for things from the mempool.

New transaction dependency sorter, first builds a graph of dependencies from the mempool transactions and then sorts them for inclusion.


CHANGELOG.md
```diff
@@ -42,6 +42,14 @@ The format is loosely based on [Keep a Changelog](https://keepachangelog.com/en/
...
           its speed at the cost of a potential db corruption if the system crashes in the meantime.
 
+  - API Server:
+    - new endpoints for addresses and transactions from the mempool:
+      - `/mempool/transaction`
+      - `/mempool/transaction/:id`
+      - `/mempool/transaction/:id/output/:idx`
+      - `/mempool/address/:address`
+      - `/mempool/address/:address/all-utxos`
+
```
ImplOfAnImpl, line 52: API server has a separate changelog - api-server/CHANGELOG.md

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3176,568 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
             })
             .collect()
     }
+
+    pub async fn set_mempool_transaction(
+        &mut self,
+        transaction_id: Id<Transaction>,
+        transaction: &TransactionInfo,
+    ) -> Result<(), ApiServerStorageError> {
+        // Because of ON DELETE CASCADE on other tables, we insert the transaction first.
```
ImplOfAnImpl, line 3185: Outdated comment? This function only inserts the transaction and does nothing else, so "we insert the transaction **first**" doesn't seem to make much sense.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -66,6 +66,8 @@ mod pos_adapter;
 pub enum BlockchainStateError {
     #[error("Unexpected storage error: {0}")]
     StorageError(#[from] ApiServerStorageError),
+    #[error("Token decimals not found for token id: {0}")]
```
ImplOfAnImpl: Btw, the ids are printed in contracted form by default (due to how Display is implemented for fixed_hash types), which is not very useful, especially in error messages. You can force the full form by using `{0:x}`.

api-server/scanner-lib/src/sync/remote_node.rs
```diff
@@ -42,6 +42,11 @@ pub trait RemoteNode {
     ) -> Result<Vec<Block>, Self::Error>;
 
     async fn mempool_feerate_points(&self) -> Result<Vec<(usize, FeeRate)>, Self::Error>;
+    async fn mempool_get_transaction(
+        &self,
+        tx_id: Id<Transaction>,
+    ) -> Result<Option<SignedTransaction>, Self::Error>;
+    async fn mempool_get_transactions(&self) -> Result<Vec<SignedTransaction>, Self::Error>;
```
ImplOfAnImpl, line 49: Btw, the doc comment for `MempoolRpc::get_all_transactions` still says "This function is mostly used for testing purposes", which is no longer true.

api-server/api-server-common/src/storage/storage_api/mod.rs
```diff
@@ -800,6 +804,59 @@ pub trait ApiServerStorageRead: Sync {
...
+        &self,
+        address: &str,
+    ) -> Result<Vec<Id<Transaction>>, ApiServerStorageError>;
+
+    async fn get_mempool_address_balances(
+        &self,
+        address: &str,
+    ) -> Result<BTreeMap<CoinOrTokenId, AmountWithDecimals>, ApiServerStorageError>;
+
+    async fn get_utxo_mempool_fallback(
```
ImplOfAnImpl, line 818: `get_mempool_utxo_with_fallback`?

Also,
1) As I can see, a bunch of other functions also fall back to non-mempool ones if the info is not found in the mempool - get_mempool_address_balance, get_mempool_locked_address_balance, get_mempool_token_num_decimals, get_mempool_order. Why aren't they named in the same way?
2) Maybe add doc comments describing what the functions are supposed to do exactly, especially the "fallback" ones?

api-server/api-server-common/src/storage/storage_api/mod.rs
```diff
@@ -800,6 +804,59 @@ pub trait ApiServerStorageRead: Sync {
...
+        outpoint: &UtxoOutPoint,
+    ) -> Result<Option<Utxo>, ApiServerStorageError>;
+
+    async fn get_mempool_address_balance(
+        &self,
+        address: &str,
+        coin_or_token_id: CoinOrTokenId,
+    ) -> Result<Option<Amount>, ApiServerStorageError>;
+
+    async fn get_mempool_locked_address_balance(
```
ImplOfAnImpl, line 829: The name of the non-mempool version of this function says "address_locked_balance", plz make the two names consistent.

Same for the setter.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3176,568 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+
+    pub async fn get_utxo_mempool_fallback(
+        &self,
+        outpoint: &UtxoOutPoint,
+    ) -> Result<Option<Utxo>, ApiServerStorageError> {
+        // TODO: optimize it into a single query
+        let mempool_row = self
+            .tx
+            .query_opt(
+                "SELECT utxo, spent FROM ml.mempool_utxo WHERE outpoint = $1 LIMIT 1;",
```
ImplOfAnImpl: "LIMIT 1" - redundant?

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3176,568 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+            .await
+            .map_err(|e| ApiServerStorageError::LowLevelStorageError(e.to_string()))?;
+        Ok(())
+    }
+
+    pub async fn get_utxo_mempool_fallback(
+        &self,
+        outpoint: &UtxoOutPoint,
+    ) -> Result<Option<Utxo>, ApiServerStorageError> {
+        // TODO: optimize it into a single query
```
ImplOfAnImpl, line 3272: By the second query, do you mean the one performed in `self.get_utxo`? Or is it an outdated comment?

Same for the other 2 functions that have it.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2661,262 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+        ConstrainedValueAccumulator::from_outputs(chain_config, block_height, tx.outputs())
+            .expect("valid mempool block");
+
+    let consumed_accumulator = inputs_accumulator
+        .satisfy_with(outputs_accumulator)
+        .expect("valid mempool block");
+    Ok(consumed_accumulator)
+}
+
+async fn increase_mempool_balance<T: ApiServerStorageWrite + ApiServerStorageRead>(
```
ImplOfAnImpl, line 2856: In this and similar functions below: return ApiServerStorageRead instead of expecting on storage errors.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3176,568 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+        len: u32,
+        offset: u64,
+    ) -> Result<Vec<TransactionInfo>, ApiServerStorageError> {
+        let len = len as i64;
+        let offset = offset as i64;
+        let rows = self
+            .tx
+            .query(
+                "SELECT transaction_data FROM ml.mempool_transactions ORDER BY insertion_time DESC LIMIT $1 OFFSET $2;",
+                &[&len, &offset],
```
ImplOfAnImpl, line 3424: I don't think that offset/len-based approach will work well in the mempool tx case. What if the contents of the mempool changes between requests?

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3176,568 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+                 VALUES ($1, $2, $3, $4)
+                 ON CONFLICT (address, coin_or_token_id) DO UPDATE SET amount = EXCLUDED.amount;",
+                &[&address, &coin_or_token_id.encode(), &amount.encode(), &(decimals as i16)],
+            )
+            .await
+            .map_err(|e| ApiServerStorageError::LowLevelStorageError(e.to_string()))?;
+        Ok(())
+    }
+
+    pub async fn get_mempool_address_balances(
```
ImplOfAnImpl, line 3438: `get_mempool_address_balances` behaves rather differently compared to `get_mempool_address_balance` - the latter returns the cumulative balance for one asset, no matter whether such a tx is present in the mempool or not. So I guess this one should return cumulative balances of all assets too?
If not, then perhaps they should be named differently.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3176,568 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+            .execute(
+                "INSERT INTO ml.mempool_token_transactions (token_id, transaction_id)
+                 VALUES ($1, $2)
+                 ON CONFLICT (token_id, transaction_id) DO NOTHING;",
+                &[&token_id.encode(), &transaction_id.encode()],
+            )
+            .await
+            .map_err(|e| ApiServerStorageError::LowLevelStorageError(e.to_string()))?;
+        Ok(())
+    }
```
ImplOfAnImpl, line 3554: Useless? I don't see any getter for this data.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3176,568 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+        token_id: TokenId,
+        issuance: FungibleTokenData,
+        chain_config: &ChainConfig,
+    ) -> Result<(), ApiServerStorageError> {
+        let authority = Address::new(chain_config, issuance.authority.clone())
+            .map_err(|_| ApiServerStorageError::AddressableError)?;
+
+        self.tx
+            .execute(
+                "INSERT INTO ml.mempool_fungible_token (token_id, ticker, authority, issuance)
```
ImplOfAnImpl, line 3637: This is never read, same for ml.mempool_nft_issuance

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2661,262 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+    block_height: BlockHeight,
+    tx: &Transaction,
+    db_tx: &T,
+) -> Result<AccumulatedFee, ApiServerStorageError> {
+    let mut inputs_utxos = Vec::with_capacity(tx.inputs().len());
+    for input in tx.inputs() {
+        inputs_utxos.push(fetch_mempool_utxo(input, db_tx).await?.map(|utxo| utxo.into_output()));
+    }
+
+    let pools = prefetch_pool_data(&inputs_utxos, db_tx).await?;
```
ImplOfAnImpl, line 2829: This call will panic if the pool data can't be found in the db.
But pool creation and decommissioning txs may appear in the mempool at the same time (especially if prepared by an attacker).

Additionally, there is no guarantee on the order in which mempool txs will be handled (because mempool rpc's `get_all_transactions` doesn't provide any, as I can see), i.e. the decommissioning can come before the pool creation. So simply tracking which pool creation txs have been handled so far won't be enough.

Plz make sure that scanner-lib has a test for this case.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2661,262 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+    let pos_accounting_adapter = PoSAccountingAdapterToCheckFees { pools };
+    let tokens_view = StubTokensAccounting;
+    let orders_store = prefetch_orders(tx.inputs(), db_tx).await?;
```
ImplOfAnImpl, line 2832: Again, order creation and a fill/freeze/conclude may be present in the mempool at the same time, in which case this will panic.

Plz also add a test for this.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2661,262 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+    let inputs_accumulator = ConstrainedValueAccumulator::from_inputs(
+        chain_config,
+        block_height,
+        &orders_db,
+        &pos_accounting_adapter,
+        &tokens_view,
+        tx.inputs(),
+        &inputs_utxos,
+    )
+    .expect("valid mempool block");
```
ImplOfAnImpl: What is "mempool block"? Same below.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2661,262 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+async fn mempool_tx_fees<T: ApiServerStorageRead>(
+    chain_config: &ChainConfig,
+    block_height: BlockHeight,
+    tx: &Transaction,
+    db_tx: &T,
+) -> Result<AccumulatedFee, ApiServerStorageError> {
+    let mut inputs_utxos = Vec::with_capacity(tx.inputs().len());
+    for input in tx.inputs() {
+        inputs_utxos.push(fetch_mempool_utxo(input, db_tx).await?.map(|utxo| utxo.into_output()));
+    }
```
ImplOfAnImpl, line 2827: Btw, this will also not work correctly if both a child and a parent are in the mempool and mempool's `get_all_transactions` returns them in the opposite order.

api-server/scanner-lib/src/sync/remote_node.rs
```diff
@@ -42,6 +42,11 @@ pub trait RemoteNode {
     ) -> Result<Vec<Block>, Self::Error>;
 
     async fn mempool_feerate_points(&self) -> Result<Vec<(usize, FeeRate)>, Self::Error>;
+    async fn mempool_get_transaction(
```
ImplOfAnImpl, line 45: Is this ever used?

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +248,459 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+        let token_decimals = tx_info.additional_info.token_decimals;
+
+        // Process Inputs
+        for (input, sig) in transaction.transaction().inputs().iter().zip(transaction.signatures())
+        {
+            match input {
+                TxInput::Utxo(outpoint) => {
+                    let utxo = fetch_mempool_utxo(input, &db_tx)
+                        .await?
+                        .expect("UTXO must exist for valid mempool tx")
```
ImplOfAnImpl: As I've said elsewhere, both a parent and a child can be in the mempool and they can come in the "wrong" order, in which case this will crash.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +248,459 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+                                    }
+                                    AuthorizedHashedTimelockContractSpend::Refund(_) => {
+                                        htlc.refund_key
+                                    }
+                                };
+                                Address::<Destination>::new(&self.chain_config, dest)
+                                    .expect("Unable to encode destination")
+                            } else {
+                                panic!("Empty signature for htlc")
+                            };
```
ImplOfAnImpl, line 340: Copy-pasted code. Plz move it into a helper function and use in both places. Also, plz don't panic inside the function, make it return a Result.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +248,459 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+                                    )
+                                    .await;
+                                    transaction_tokens.insert(token_id);
+                                }
+                            }
+                        }
+                        TxOutput::ProduceBlockFromStake(_, pool_id)
+                        | TxOutput::CreateStakePool(pool_id, _) => {
+                            let pool_data = db_tx
+                                .get_pool_data(pool_id)
```
ImplOfAnImpl: Again, this will fail if poll creation and decommissioning are in the mempool at the same time.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +248,459 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+                        | AccountCommand::UnfreezeToken(token_id)
+                        | AccountCommand::LockTokenSupply(token_id)
+                        | AccountCommand::ChangeTokenAuthority(token_id, _)
+                        | AccountCommand::ChangeTokenMetadataUri(token_id, _) => {
+                            transaction_tokens.insert(*token_id);
+                        }
+                        AccountCommand::ConcludeOrder(order_id)
+                        | AccountCommand::FillOrder(order_id, _, _) => {
+                            let order =
+                                db_tx.get_mempool_order(*order_id).await?.expect("must exist");
```
ImplOfAnImpl: Same about orders

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +248,459 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+                                increase_mempool_balance(
+                                    &mut db_tx,
+                                    address.as_str(),
+                                    amount,
+                                    CoinOrTokenId::Coin,
+                                    self.chain_config.coin_decimals(),
+                                )
+                                .await;
+                            } else {
+                                increase_mempool_locked_address_balance(
```
ImplOfAnImpl, line 532: I see that you increase locked balances in the mempool but never decrease them.
Mempool has some timelock tolerance defined by constants `FUTURE_TIMELOCK_TOLERANCE`/`FUTURE_TIMELOCK_TOLERANCE_BLOCKS`. So it should be possible to have e.g. a tx with LockThenTransfer output for 2 blocks and then a tx that consumes that output both in the mempool at the same time.

Probably you should decrease locked balances if a LockThenTransfer utxo is spent by a mempool tx.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +248,459 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
 
         Ok(())
     }
+
+    async fn add_mempool_tx(&mut self, transaction: &SignedTransaction) -> Result<(), Self::Error> {
+        let (best_block_height, _, best_block_timestamp) = self.best_block().await?;
+        let next_block_height = best_block_height.next_height();
+
+        let mut db_tx = self.storage.transaction_rw().await.expect("Unable to connect to database");
```
ImplOfAnImpl, line 256: Btw why do you panic on storage errors everywhere, when this function already returns an error type constructible from ApiServerStorageError?

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2661,262 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+        .collect::<Vec<_>>();
+    let addresses_refs: Vec<&str> = addresses.iter().map(|s| s.as_str()).collect();
+    db_tx
+        .set_utxo_at_height(outpoint, utxo, &addresses_refs, block_height)
+        .await
+        .expect("Unable to set utxo");
 }
 
-fn get_tx_output_destination(txo: &TxOutput) -> Option<&Destination> {
+fn get_tx_output_destination(txo: &TxOutput) -> Vec<&Destination> {
```
ImplOfAnImpl, line 2808: destinations

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2661,262 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+                    }
+                }),
+            ),
+            _ => None,
+        })
+        .collect::<Result<BTreeMap<_, _>, _>>()?;
+
+    let input_tasks: FuturesOrdered<_> =
+        tx.inputs().iter().map(|input| fetch_mempool_utxo(input, db_tx)).collect();
+    let input_utxos: Vec<Option<Utxo>> = input_tasks.try_collect().await?;
```
ImplOfAnImpl, line 2725: Again, this will fail if the utxo belongs to a mempool tx that hasn't been handled yet.

api-server/web-server/src/api/v2.rs
```diff
@@ -101,6 +101,14 @@ pub fn routes<
         .route("/transaction/:id/merkle-path", get(transaction_merkle_path))
         .route("/transaction/:id/output/:idx", get(transaction_output));
 
+    let router = router
+        .route("/mempool/transaction", get(mempool_transactions))
+        .route("/mempool/transaction/:id", get(mempool_transaction))
+        .route(
+            "/mempool/transaction/:id/output/:idx",
+            get(mempool_transaction_output),
+        );
```
ImplOfAnImpl, line 110: So this accepts both txs in mempool and already mined ones. Why not add an option to the existing `/transaction/:id/output/:idx` so that it would consider mempool txs as well?

ImplOfAnImpl (commented):
CURRENT_STORAGE_VERSION wasn't updated - is this intended?

api-server/scanner-lib/src/sync/remote_node.rs
OBorce: Removed that comment

api-server/api-server-common/src/storage/impls/postgres/queries.rs
OBorce: yes joining the 2 queries into a single query

api-server/api-server-common/src/storage/impls/postgres/queries.rs
OBorce: Haven't found a better alternative yet.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
OBorce: We can add new endpoints if needed in the future without needing to change the DB and do a rescan.

api-server/scanner-lib/src/blockchain_state/mod.rs
OBorce: Now we are sorting the transactions coming from mempool based on their dependencies.

api-server/scanner-lib/src/blockchain_state/mod.rs
OBorce: I added a test with a tx spending a locked UTXO with 1..FUTURE_TIMELOCK_TOLERANCE_BLOCKS.

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+    chain_config: &ChainConfig,
+    block_height: BlockHeight,
+    dependencies: &mut DependenciesMap,
+    tx: &SignedTransaction,
+) {
+    let inputs = tx.transaction().inputs();
+    for (out_index, out) in tx.transaction().outputs().iter().enumerate() {
+        match out {
+            TxOutput::CreateOrder(order_data) => {
+                let order_id = make_order_id(inputs).expect("Valid TX from mempool");
```
ImplOfAnImpl, line 159: Plz don't `expect`. Create an error type specifically for dependency_graph and return it here.

P.S. also, this module is not mempool-specific, so mempool should not be mentioned here.

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+        let transactions = vec![txa, txb];
+        let dependency_graph = build_dependency_graph(transactions, &chain_config, block_height);
+        assert_eq!(dependency_graph.len(), 2);
+        assert_eq!(dependency_graph[0].id, txa_id);
+        assert_eq!(dependency_graph[1].id, txb_id);
+        assert!(dependency_graph[0].dependencies.is_empty());
+        assert_eq!(dependency_graph[1].dependencies, vec![txa_id]);
+    }
+
+    // test txs not dependednt on UTXO input/outputs but on token creation/command
```
ImplOfAnImpl, line 417: typo - dependednt

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+    #[rstest]
+    #[trace]
+    #[case(Seed::from_entropy())]
+    fn test_token_creation_dependency_chain(#[case] seed: Seed) {
+        let mut rng = make_seedable_rng(seed);
+        let chain_config = create_regtest();
+        let block_height = BlockHeight::new(0);
+
+        // A creates a new token, B uses a comand on it
```
ImplOfAnImpl, line 426: typo - comand

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+        let txb = TransactionBuilder::new()
+            .add_input(
+                TxInput::Utxo(output_from_txa),
+                InputWitness::NoSignature(None),
+            )
+            .add_anyone_can_spend_output(100)
+            .build();
+        let txb_id = txb.transaction().get_id();
+
+        let transactions = vec![txa, txb];
```
ImplOfAnImpl, line 408: Here and in other tests: `shuffle` the tx vector before building the graph?

api-server/scanner-lib/src/sync/tx_dependency_ordering/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+    primitives::BlockHeight,
+};
+
+mod dependency_graph;
+
+use dependency_graph::{build_dependency_graph, DependencyNode};
+
+// Order transactions by dependency between each other.
+// Returns a Vec of transactions starting from the top-most parent transaction
+// which doesn't depend on any other transaction fallowing it, and ending with the leafs.
```
ImplOfAnImpl, line 29: typo - fallowing

api-server/scanner-lib/src/sync/tx_dependency_ordering/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+
+    Ok(sorted_transactions)
+}
+
+/// Errors that can occur during topological sorting.
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
+pub enum TopoSortError {
+    #[error("Circular dependency was detected")]
+    CycleDetected,
+    #[error("A node declared a dependency that is not present in the provided vector.")]
```
ImplOfAnImpl, line 50: We don't use full stops at the end of individual error messages (and you don't use it in `CycleDetected` either).

api-server/scanner-lib/src/sync/tx_dependency_ordering/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+    MissingDependency,
+}
+
+/// Sorts a vector of `DependencyNode`s topologically.
+///
+/// Items with no dependencies (roots) will appear first in the resulting vector.
+fn topological_sort<T>(nodes: Vec<T>) -> Result<Vec<T>, TopoSortError>
```
ImplOfAnImpl, line 57: IMO this must go to `utils`, together with `DependencyNode` and `TopoSortError` (btw, let's rename the latter to `TopologicalSortError`) and the tests.

api-server/scanner-lib/src/sync/tx_dependency_ordering/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+    // Indegree: indegrees[i] is the number of unresolved dependencies node i has.
+    let mut indegrees: Vec<usize> = vec![0; n];
+
+    // Build the graph
+    for (i, node) in nodes.iter().enumerate() {
+        for dep_id in node.dependencies() {
+            let dep_index = id_to_index.get(dep_id).ok_or(TopoSortError::MissingDependency)?;
+
+            dependents[*dep_index].push(i);
+            indegrees[i] += 1;
```
ImplOfAnImpl, line 108: Just do `indegrees[i] = node.dependencies().len()` once outside the loop?

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+                        .push(tx_index);
+                    if let Some(previous_nonce) = nonce.decrement() {
+                        dependencies
+                            .dependents
+                            .entry(Dependency::TokenCommand(*token_id, previous_nonce))
+                            .or_default()
+                            .push(tx_index);
+                    }
+                }
+                AccountCommand::ConcludeOrder(_) | AccountCommand::FillOrder(_, _, _) => {}
```
ImplOfAnImpl, line 296: Perhaps `build_dependency_graph` should have a doc comment stating that it's intended to be used with valid new transactions only. And that e.g. potential dependencies introduced by legacy order commands are ignored.

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+                InputWitness::NoSignature(None),
+            )
+            .add_output(order)
+            .build();
+        let txa_id = txa.transaction().get_id();
+        let order_id = make_order_id(txa.transaction().inputs()).unwrap();
+
+        let random_utxo_outpoint2 =
+            UtxoOutPoint::new(OutPointSourceId::Transaction(random_tx_id), 1);
+        let fill_order = OrderAccountCommand::FillOrder(order_id, Amount::from_atoms(1));
```
ImplOfAnImpl, line 582: Use more than one fill tx?

P.S. The test would be easier to read (and to modify) if you used more descriptive names for the variables. E.g. `txa`/`txa_id` -> `creation_tx`/`creation_tx_id` etc.

api-server/scanner-lib/src/sync/tx_dependency_ordering/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    // Reconstruct the sorted vector without cloning `T`
+    // We wrap the original items in Option, and `take()` them out in sorted order.
+    let mut wrapped_nodes: Vec<Option<T>> = nodes.into_iter().map(Some).collect();
+
+    let sorted_nodes = sorted_indices
+        .into_iter()
+        .map(|idx| wrapped_nodes[idx].take().expect("present"))
+        .collect();
```
ImplOfAnImpl, line 157: Btw, it's possible to do this without the extra allocation. And it'd also allow to change this function's signature to `fn topological_sort<T>(nodes: &mut [T]) -> Result<(), TopoSortError>`, making it even more generic.

Not insisting on this though.

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+                    .push(tx_index);
+            }
+            TxOutput::IssueNft(token_id, _, _) => {
+                dependencies
+                    .providers
+                    .entry(Dependency::TokenCreation(*token_id))
+                    .or_default()
+                    .push(tx_index);
+            }
+            _ => {
```
ImplOfAnImpl, line 203: Plz specify all variants explicitly

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+}
+
+// Highest priority txs are first, then delegation stake, delegation withdrawal and last token freeze
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Clone, Copy)]
+pub enum TxPriorityOrder {
+    Highest = 3,
+    DelegationStake = 2,
+    DelegationWithdrawal = 1,
+    TokenFreeze = 0,
+}
```
ImplOfAnImpl, line 70: I don't understand why this is needed.

1. `TokenFreeze` - as I can see, token txs are already ordered by nonce. So this will only influence `FreezeToken` tx's position compared to some unrelated txs. Or am I missing something?

2. `DelegationStake`/`DelegationWithdrawal` - this could also be handled at the `Dependency` level, like you did for orders.

api-server/scanner-lib/src/sync/tx_dependency_ordering/dependency_graph.rs
```diff
@@ -0,0 +1,661 @@
...
+        self.priority
+    }
+
+    fn dependencies(&self) -> &[Self::Id] {
+        &self.dependencies
+    }
+}
+
+// Build a dependency graph from the provided transactions
+pub fn build_dependency_graph(
```
ImplOfAnImpl, line 117: This function doesn't take all dependencies into account and this has to be documented (or fixed).
E.g. from what I can see:
1. Delegations are not ordered with respect to pool creation and decommissioning (so e.g. if at some point in the future the caller code will start querying pool info when encountering a delegation tx, 
it will become automatically broken).
2. Delegation withdrawals are not ordered by the nonce.

api-server/scanner-lib/src/sync/tx_dependency_ordering/mod.rs
```diff
@@ -0,0 +1,405 @@
...
+};
+
+mod dependency_graph;
+
+use dependency_graph::{build_dependency_graph, DependencyNode};
+
+// Order transactions by dependency between each other.
+// Returns a Vec of transactions starting from the top-most parent transaction
+// which doesn't depend on any other transaction fallowing it, and ending with the leafs.
+pub fn order_transactions_by_dependency(
```
ImplOfAnImpl, line 30: Probably this comment should also mention that certain dependencies are ignored, same as the one for `build_dependency_graph`.

api-server/api-server-common/src/storage/storage_api/mod.rs
```diff
@@ -800,6 +808,83 @@ pub trait ApiServerStorageRead: Sync {
...
+    ) -> Result<Vec<Id<Transaction>>, ApiServerStorageError>;
+
+    async fn get_mempool_address_balances(
+        &self,
+        address: &str,
+    ) -> Result<BTreeMap<CoinOrTokenId, AmountWithDecimals>, ApiServerStorageError>;
+
+    // Returns the UTXO from the mempool if it exists,
+    // otherwise fallbacks to getting it from the confirmed UTXOs.
+    async fn get_utxo_mempool_with_fallback(
```
ImplOfAnImpl, line 824: get_mempool_utxo_with_fallback ?

Same for the intermediate functions with the same name.

api-server/api-server-common/src/storage/impls/in_memory/mod.rs
```diff
@@ -840,6 +865,154 @@ impl ApiServerInMemoryStorage {
...
+    fn get_mempool_address_all_utxos(
+        &self,
+        address: &str,
+    ) -> Result<Vec<(UtxoOutPoint, UtxoWithExtraInfo)>, ApiServerStorageError> {
+        let result = self
+            .mempool_address_utxos
+            .get(address)
+            .unwrap_or(&BTreeSet::new())
+            // TODO:
+            .union(self.address_locked_utxos.get(address).unwrap_or(&BTreeSet::new()))
```
ImplOfAnImpl, line 986: TODO what?

P.S. I guess this `union` should be removed because `mempool_address_utxos` should already contain locked utxos as well?

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -3055,6 +3256,637 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+    }
+
+    pub async fn set_mempool_token_transaction(
+        &mut self,
+        token_id: TokenId,
+        transaction_id: Id<Transaction>,
+    ) -> Result<(), ApiServerStorageError> {
+        self.tx
+            .execute(
+                "INSERT INTO ml.mempool_token_transactions (token_id, transaction_id)
```
ImplOfAnImpl, line 3693: About these tables that no-one reads (`mempool_token_transactions`, `mempool_fungible_token`, `mempool_nft_issuance`): I understand your point, but having code "just in case" that isn't even tested is not great. If you insist on keeping these then let's maybe add the getters too, and also tests in scanner-lib that would ensure that the data is collected properly.

api-server/api-server-common/src/storage/impls/postgres/queries.rs
```diff
@@ -1046,6 +1047,142 @@ impl<'a, 'b> QueryFromConnection<'a, 'b> {
...
+        self.just_execute(
+            "CREATE TABLE ml.mempool_orders (
+                    order_id TEXT NOT NULL,
+                    initially_asked TEXT NOT NULL,
+                    ask_balance TEXT NOT NULL,
+                    ask_currency bytea NOT NULL,
+                    initially_given TEXT NOT NULL,
+                    give_balance TEXT NOT NULL,
+                    give_currency bytea NOT NULL,
+                    next_nonce bytea NOT NULL,
```
ImplOfAnImpl, line 1178: Mempool orders can only be V1, so they can't have nonces.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -66,6 +67,18 @@ mod pos_adapter;
...
     #[error("Unexpected storage error: {0}")]
     StorageError(#[from] ApiServerStorageError),
+    #[error("Token decimals not found for token id: {0:x}")]
+    TokenDecimalsNotFound(TokenId),
+    #[error("UTXO not found in DB")]
+    UtxoNotFound,
+    #[error("Pool not found in DB")]
+    PoolNotFound,
+    #[error("Order not found in DB")]
+    OrderNotFound,
```
ImplOfAnImpl, line 77: Put outpoint, pool and order ids in these errors too?

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2791,303 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
         TxOutput::IssueFungibleToken(_)
         | TxOutput::Burn(_)
         | TxOutput::DelegateStaking(_, _)
         | TxOutput::DataDeposit(_)
-        | TxOutput::CreateOrder(_) => None,
+        | TxOutput::CreateOrder(_) => vec![],
+    }
+}
+
+async fn calculate_mempool_tx_fee_and_collect_token_info<
```
ImplOfAnImpl, line 2827: This file is already pretty big. Let's may be put all mempool-related functions to a mempool-related submodule?

Or perhaps use some other criteria, e.g. so that `calculate_mempool_tx_fee_and_collect_token_info` and `calculate_tx_fee_and_collect_token_info` end up near each other.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -2185,31 +2791,303 @@ async fn set_utxo<T: ApiServerStorageWrite>(
...
+                            token_ids.insert(id);
+                        }
+                    }
+                },
+                TxInput::OrderAccountCommand(cmd) => match cmd {
+                    OrderAccountCommand::FillOrder(order_id, _)
+                    | OrderAccountCommand::FreezeOrder(order_id)
+                    | OrderAccountCommand::ConcludeOrder(order_id) => {
+                        let order = db_tx
+                            .get_order(*order_id)
```
ImplOfAnImpl, line 2900: get_mempool_order_with_fallback?

Same above, where legacy orders are handled.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +259,482 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+                            if let Some(id) = order.ask_currency.token_id() {
+                                transaction_tokens.insert(id);
+                            }
+                            if let Some(id) = order.give_currency.token_id() {
+                                transaction_tokens.insert(id);
+                            }
+                        }
+                    }
+                }
+                _ => {}
```
ImplOfAnImpl, line 447: Using wildcard in matches is dangerous - when the enum changes, you won't notice that the code needs updating.

But here you actually omitted orders v1 entirely.

Plz write a test in scanner-lib for this.

api-server/scanner-lib/src/blockchain_state/mod.rs
```diff
@@ -242,6 +259,482 @@ impl<S: ApiServerStorage + Send + Sync> LocalBlockchainState for BlockchainState
...
+                        .map_err(BlockchainStateError::StorageError)?;
+                }
+                TxOutput::ProduceBlockFromStake(_, _) => {}
+                TxOutput::CreateDelegationId(_, _) => {}
+                TxOutput::DelegateStaking(_, _) => {}
+                TxOutput::DataDeposit(_) => {}
+            }
+
+            // Save new UTXO as unspent in mempool
+            let utxo = Utxo::new(output.clone(), token_decimals_opt, None);
```
ImplOfAnImpl, line 697: Btw, you call `set_mempool_utxo`/`set_mempool_locked_utxo` even for non-spendable outputs, which are not really utxos. As I can see, in `update_tables_from_transaction_outputs` only spendable outputs are recorded as utxos, so I guess here we should do the same?

api-server/CHANGELOG.md
```diff
@@ -9,9 +9,14 @@ The format is loosely based on [Keep a Changelog](https://keepachangelog.com/en/
...
 ### Added
-- New endpoint was added: `/v2/transaction/{id}/output/{idx}`.
+- New endpoint was added: `/v2/transaction/{id}/output/{idx}`, including an optional `with_mempool` parameter that controls the display of mempool spent status.
 - New endpoint was added: `/v2/token/{id}/transactions` will return all transactions related to a token.\
   Pagination works like the new absolute mode in `/v2/transaction`, using `offset` and `items`.
+- New endpoints for addresses and transactions from the mempool:
+  - `/v2/mempool/transaction`
+  - `/v2/mempool/transaction/:id`
+  - `/v2/mempool/address/:address`
+  - `/v2/mempool/address/:address/all-utxos`
```
ImplOfAnImpl, line 19: Wrong place. Add the changes to the `[Unreleased]` section instead.

api-server/scanner-lib/src/sync/mod.rs
```diff
@@ -63,12 +66,31 @@ pub async fn sync_once(
...
+
+            for tx in ordered_txs {
+                local_state
+                    .add_mempool_tx(&tx)
+                    .await
+                    .map_err(|e| SyncError::LocalNode(e.to_string()))?;
+            }
+
             return Ok(());
         }
```
ImplOfAnImpl, line 96: 1. As I can see, the caller function doesn't try to sleep between iterations. Previously most invocations of `sync_once` were cheap - check the block id, see that it's the best one, exit. Now you pull the entire mempool, which in theory may be 300 Mb. Probably the caller function should sleep for some time between iterations, e.g. for 1 sec?

2. Transactions can be evicted from mempool if it reaches the max size. Not a problem for now of course, but it's better to implement things properly from the beginning. So, scanner-lib should check for transactions that are in the mempool-related tables, but no longer in the current mempool, and update the db accordingly, marking the corresponding utxos as unspent etc.

ImplOfAnImpl (commented):
1. The most complex changes are in the scanner-lib, but there are zero new tests in `api-server/scanner-lib/src/sync/tests`. I don't think that the tests at the stack-test-suite level are enough: the current tests are limited and in general tests at that level are limited in what they can check compared to the ones in scanner-lib. Additionally, there is functionality that is not currently exposed via the API, so it's basically not tested.
**Plz add tests at the scanner-lib level too. Plz make sure they're extensive enough.**

2. Regarding the pagination of the mempool transactions in the corresponding endpoint - obviously the current solution just wont work. I guess we should get rid of the pagination entirely in this case and return the entire contents in one call, as they do in bitcoin in `getrawmempool` (by default, they only return tx ids, so the response is not that large even for a full mempool). Also, let's maybe introduce something similar to their `mempool_sequence` counter, which would be incremented every time the mempool contents change, so that the caller can use that to check that the mempool has changed.

erubboli:
Thanks @OBorce for the substantial work on this, the dependency-ordering analysis in particular is excellent and will be reused. Closing rather than merging: since this was opened, the api-server gained a real-time event stream (#2117) whose design (small push events + REST hydration via the node RPC, which already exposes get_transaction/transactions) covers this PR's use cases without persisting mempool state into the api-server database. We'll follow up with a smaller PR under that architecture, based on this one and crediting it.

PR #2041, opened 2026-04-20
Rand upgrade: autoreplace "gen" with "random" etc

ImplOfAnImpl:
Here I auto-replace `gen` and `r#gen` with `random`, `gen_bool` with `random_bool` and `gen_range` with `random_range`.

The purpose of this PR is to make https://github.com/mintlayer/mintlayer-core/pull/2042 smaller. The code here doesn't compile. Once both PRs are approved I'll merge #2042 into this one and then this one to master.

P.S. very few occurrences of "rand" and friends were actually meant to retain their original names, because they are called on a 0.8 RNG. I kept these redundant changes in this PR and reverted them in #2042.


OBorce (approved):

ImplOfAnImpl:
~~I had to push an additional commit after the merge from https://github.com/mintlayer/mintlayer-core/pull/2042, fixing an unrelated typo. This is just to trigger the CI again, which got stuck for some reason (adding a commit and then dropping it didn't help).~~
The problem was in branch protection rules. I removed the extra commit.

PR #2045, opened 2026-04-23
fix handling of conflicting order transactions in the wallet

### LINKED ISSUE #2043: Wallet fails to record confirmed ConcludeOrder when a competing unconfirmed conclude exists
## Wallet fails to record confirmed ConcludeOrder when a competing unconfirmed conclude exists

**Description**

When a confirmed `OrderAccountCommand::ConcludeOrder` transaction arrives in a new block, the wallet incorrectly fails to record it if there is already an unconfirmed conclude transaction for the same order in the output cache. Instead of marking the unconfirmed transaction as conflicted and applying the confirmed one, the wallet throws:

Output cache inconsistency error: `Order <order_id>` is already concluded

**Root cause**

`update_conflicting_txs` in `wallet/src/account/output_cache/mod.rs` explicitly skips conflict detection for `TxInput::OrderAccountCommand`:

```rust
TxInput::OrderAccountCommand(_) => None,  // no conflict detection
```

The old V0 order commands (`TxInput::AccountCommand`) have nonce-based conflict detection and work correctly. The V1 order commands (`OrderAccountCommand`) have no equivalent, so competing unconfirmed conclude transactions are never marked as conflicted when a confirmed conclude arrives.

**Sequence of events**

An unconfirmed V1 conclude tx for order X is added to the wallet  -> cache `sets is_concluded = true`
A different conclude tx for the same order gets confirmed on-chain
`update_conflicting_txs` returns `None` for `OrderAccountCommand` -> the unconfirmed conclude tx is never rolled back
The wallet tries to apply the confirmed conclude tx  -> `finds is_concluded == true`  -> throws `OrderAlreadyConcluded`
Expected behaviour

The confirmed conclude tx is recorded in the wallet. The competing unconfirmed conclude tx is marked as Conflicted and rolled back from the cache.

**Actual behaviour**

The confirmed conclude tx fails to be recorded. The wallet has no knowledge that the order was concluded on-chain. The stale unconfirmed conclude tx remains in the cache indefinitely.

**Fix**

`update_conflicting_txs` needs to handle `OrderAccountCommand::ConcludeOrder(order_id)` and `OrderAccountCommand::FreezeOrder(order_id)`: when such a command is confirmed, any unconfirmed transaction containing the same command for the same `order_id` should be detected as conflicting and rolled back. Since V1 orders are nonce-free, this requires `order-id-based` conflict detection rather than the existing nonce-based approach. `FillOrder` does not need this treatment as multiple fills for the same order can legitimately coexist.

NOTE: `FreezeOrder` likely has the same issue.

OBorce:
Fixes #2043 

Any confirmed Freeze order transaction will now mark any unconfirmed Fill or Freeze order transaction for the same order as conflicting.
Similar for Conclude order it will mark any unconfirmed Fill, Freeze or Conclude order transactions as conflicting.

### DIFF
--- a/wallet/src/account/output_cache/mod.rs
+++ b/wallet/src/account/output_cache/mod.rs
@@ -783,9 +783,11 @@ impl OutputCache {
         confirmed_tx: &Transaction,
         block_id: Id<GenBlock>,
     ) -> WalletResult<Vec<(Id<Transaction>, WalletTx)>> {
+        // TODO: maybe make it an enum
         struct ConflictCheck {
             frozen_token_id: Option<TokenId>,
             confirmed_account_nonce: Option<(AccountType, AccountNonce)>,
+            conflicting_order_command: Option<(OrderAccountCommandTag, OrderId)>,
         }
 
         let conflict_checks = confirmed_tx
@@ -804,6 +806,7 @@ impl OutputCache {
                             outpoint.account().into(),
                             outpoint.nonce(),
                         )),
+                        conflicting_order_command: None,
                     }),
                     TxInput::AccountCommand(nonce, cmd) => match cmd {
                         AccountCommand::MintTokens(_, _)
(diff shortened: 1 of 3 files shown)


wallet/src/account/output_cache/mod.rs
```diff
@@ -2062,6 +2101,41 @@ fn uses_conflicting_nonce(
     })
 }
 
+fn uses_conflicting_order_command(
+    unconfirmed_tx: &WalletTx,
+    cmd_tag: OrderAccountCommandTag,
+    order_id: OrderId,
```
ImplOfAnImpl: Name them `confirmed_tx_cmd_tag` and `confirmed_tx_order_id`?

wallet/src/account/output_cache/tests.rs
```diff
@@ -1267,6 +1267,333 @@ fn orders_state_update(#[case] seed: Seed) {
...
+                unconfirmed_conclude_tx.clone(),
+                TxState::InMempool(2),
+            )),
+        )
+        .unwrap();
+
+    assert!(output_cache.orders[&order_id].is_concluded);
+
+    // A *different* confirmed conclude tx for the same order arrives.
+    // update_conflicting_txs must mark the unconfirmed one as Conflicted.
```
ImplOfAnImpl: "unconfirmed one" - there are multiple txs that are marked as Conflicted here

wallet/src/account/output_cache/tests.rs
```diff
@@ -1267,6 +1267,333 @@ fn orders_state_update(#[case] seed: Seed) {
...
+            BlockHeight::new(2),
+            confirmed_conclude_tx_id.into(),
+            WalletTx::Tx(TxData::new(
+                confirmed_conclude_tx,
+                TxState::Confirmed(BlockHeight::new(2), BlockTimestamp::from_int_seconds(2), 0),
+            )),
+        )
+        .unwrap();
+
+    assert!(output_cache.orders[&order_id].is_concluded);
```
ImplOfAnImpl, line 1448: For completeness, I'd check the state of `is_frozen` too.

Same in the second test - whenever you check `is_frozen`, it's better to also check that `is_concluded` is not affected.

ImplOfAnImpl (approved):

ImplOfAnImpl:
Plz also update the changelog

PR #2051, opened 2026-04-30
Reset InMempool txs to Inactive before connecting to the node

OBorce:
We reset all InMempool transactions back to Inactive on 
1. wallet load, 
2. every time the mempool notifications channel is closed
so that we can rescan them from the node when we connect back.

### DIFF
--- a/wallet/src/account/output_cache/tests.rs
+++ b/wallet/src/account/output_cache/tests.rs
@@ -1608,6 +1608,28 @@ fn update_conflicting_txs_order_v1_freeze(#[case] seed: Seed) {
     assert!(!output_cache.orders[&order_id].is_concluded);
 }
 
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn reset_inmempool_txs_to_inactive(#[case] seed: Seed) {
+    let mut rng = make_seedable_rng(seed);
+    let chain_config = create_unit_test_config();
+    let mut output_cache = OutputCache::empty();
+
+    // add 10 random txs
+    for _ in 0..10 {
+        add_random_transfer_tx(&mut output_cache, &chain_config, &mut rng);
+    }
+
+    // Reset
+    output_cache.reset_inmempool_txs_to_inactive();
+
+    // Check none of the txs are InMempool state
+    for tx in output_cache.txs.values() {
+        assert!(!matches!(tx.state(), TxState::InMempool(_)));
+    }
+}
+
 fn add_random_transfer_tx(
     output_cache: &mut OutputCache,
     chain_config: &ChainConfig,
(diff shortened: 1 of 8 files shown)


wallet/src/account/output_cache/tests.rs
```diff
@@ -1608,6 +1608,28 @@ fn update_conflicting_txs_order_v1_freeze(#[case] seed: Seed) {
...
+#[trace]
+#[case(Seed::from_entropy())]
+fn reset_inmempool_txs_to_inactive(#[case] seed: Seed) {
+    let mut rng = make_seedable_rng(seed);
+    let chain_config = create_unit_test_config();
+    let mut output_cache = OutputCache::empty();
+
+    // add 10 random txs
+    for _ in 0..10 {
+        add_random_transfer_tx(&mut output_cache, &chain_config, &mut rng);
```
ImplOfAnImpl: Plz don't create random transactions, create ones with `InMempool` state explicitly.

For this you can factor out the corresponding logic from `add_random_transfer_tx` into a separate func `add_random_transfer_tx_with_state` and call it here and inside `add_random_transfer_tx`.

wallet/wallet-controller/src/lib.rs
```diff
@@ -1461,7 +1460,7 @@ where
             }
 
             // after the first successful sync to the tip fetch all mempool transactions
-            if !self.finished_initial_sync.test() {
+            if self.should_resecan_mempool_txs {
```
ImplOfAnImpl: The comment is outdated, as mempool fetching is now performed not only after the first successful sync

ImplOfAnImpl (approved):
As I mentioned earlier, mempool can evict transactions - there is a size limit for the entire mempool and an expiration time for each tx. Though at this memoment it's unlikely to happen (the size limit is 300Mb and expiration time 2 weeks), it's better to fix the problem fully from the beginning. At this moment eviction only happens after a new tx has been added to the mempool. So the fix could be that whenever the wallet receives a NewTransaction event, it would obtain all InMempool tx ids from the output cache, check which of them are no longer in mempool and reset their status. (though it's better not to use `MempoolRpc::contains_tx` for checking and instead introduce an additional method that would operate on a set of ids, to avoid the rpc calls overhead).

ImplOfAnImpl:
> As I mentioned earlier, mempool can evict transactions - there is a size limit for the entire mempool and an expiration time for each tx. Though at this memoment it's unlikely to happen (the size limit is 300Mb and expiration time 2 weeks), it's better to fix the problem fully from the beginning. At this moment eviction only happens after a new tx has been added to the mempool. So the fix could be that whenever the wallet receives a NewTransaction event, it would obtain all InMempool tx ids from the output cache, check which of them are no longer in mempool and reset their status. (though it's better not to use `MempoolRpc::contains_tx` for checking and instead introduce an additional method that would operate on a set of ids, to avoid the rpc calls overhead).

I added a TODO for this

PR #2093, opened 2026-07-10
Validate PartiallySignedTransaction after decoding

### LINKED ISSUE #2069: Wallet should validate `PartiallySignedTransaction`
Wallet RPC methods and CLI commands that accept `PartiallySignedTransaction` should call `ensure_consistency` on it after decoding. Otherwise the wallet may crash if a malformed (but still decodable) `PartiallySignedTransaction` is specified.

oliv3rdrt:
## Problem

Wallet RPC methods and CLI commands that accept a `PartiallySignedTransaction` decode it from bytes with SCALE, which skips the checks that `PartiallySignedTransaction::new` performs. A malformed but still decodable transaction (for example one whose witness count does not match the number of inputs) can therefore reach the wallet and make it panic instead of returning an error.

## Fix

Call `ensure_consistency` immediately after decoding in `GenericTransaction::decode_from_untagged_bytes`, the shared path used by `account_sign_raw_transaction` and `transaction_inspect` (and the CLI commands that call them). A new `GenericTransactionError::InconsistentPartiallySignedTransaction` variant is returned on failure.

The `Basic` consistency level is used on purpose: it enforces the structural invariants that prevent the panics, while still accepting partial transactions that are legitimately incomplete during signing. Using `WithAdditionalInfo` here would reject those and also break the existing decode test.

## Tests

Added a unit test that feeds a hand-encoded, inconsistent `PartiallySignedTransaction` through the decode path and asserts it is rejected. The existing tests still pass and clippy is clean for the changed crate.

Closes #2069

### DIFF
--- a/wallet/types/src/generic_transaction.rs
+++ b/wallet/types/src/generic_transaction.rs
@@ -14,7 +14,11 @@
 // limitations under the License.
 
 use common::chain::{
-    SignedTransaction, Transaction, partially_signed_transaction::PartiallySignedTransaction,
+    SignedTransaction, Transaction,
+    partially_signed_transaction::{
+        PartiallySignedTransaction, PartiallySignedTransactionConsistencyCheck,
+        PartiallySignedTransactionError,
+    },
 };
 use serialization::DecodeAll as _;
 
@@ -34,8 +38,18 @@ impl GenericTransaction {
         // has an extra field, it's always bigger than the latter. By using `decode_all` we
         // make sure that these 2 can't be confused with each other either.
         // I.e. the order in which the decoding attempts are performed here doesn't matter.
-        if let Ok(tx) = PartiallySignedTransaction::decode_all(&mut &bytes[..]) {
-            Ok(Self::Partial(tx))
+        if let Ok(ptx) = PartiallySignedTransaction::decode_all(&mut &bytes[..]) {
+            // Decoding bypasses the invariants that `PartiallySignedTransaction::new`
+            // enforces, so a malformed (but still decodable) transaction could later make
+            // the wallet panic, e.g. on a witness/input count mismatch. Reject it here.
+            //
+            // `Basic` only checks the structural invariants that prevent such panics; the
+            // stricter additional-info checks are intentionally not used here so that a
+            // transaction that is still being assembled (and may not carry all of its input
+            // utxos or additional info yet) is not rejected.
+            ptx.ensure_consistency(PartiallySignedTransactionConsistencyCheck::Basic)
+                .map_err(GenericTransactionError::InconsistentPartiallySignedTransaction)?;
+            Ok(Self::Partial(ptx))
         } else if let Ok(stx) = SignedTransaction::decode_all(&mut &bytes[..]) {
             Ok(Self::Signed(stx))
         } else if let Ok(tx) = Transaction::decode_all(&mut &bytes[..]) {
@@ -50,6 +64,9 @@ impl GenericTransaction {
 pub enum GenericTransactionError {
     #[error("The provided bytes are not an encoded transaction")]
     CannotDecodeFromUntaggedBytes,
+
+    #[error("Inconsistent partially signed transaction: {0}")]
+    InconsistentPartiallySignedTransaction(PartiallySignedTransactionError),
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


erubboli:
Thanks for this — the approach looks right to me. I checked that `Basic` is genuinely sufficient for the panics you're targeting: every reachable out-of-bounds index on a decoded PSTx (`wallet-controller/src/lib.rs:1264`, `:1278`, and the three signers) is covered by the `witnesses`/`input_utxos`/`destinations`/`htlc_secrets` == `inputs.len()` invariant, and the sighash-commitment path and `verify_tx_signature` are `Result`-based with no indexing. I also confirmed nothing in tree produces a PSTx that would now fail to decode — every producer goes through `new`/`new_for_wallet` with at least `Basic`.

A few small things:

1. Both callers (`wallet-rpc-lib/src/rpc/mod.rs:913` and `:1118`) do `.map_err(|_| RpcError::InvalidRawTransaction)`, so the new `InconsistentPartiallySignedTransaction(..)` payload is discarded. A user with, say, a wrong `destinations` count gets the generic "invalid raw transaction", indistinguishable from a hex/encoding error — the detail you added never reaches them. Worth a dedicated `RpcError` variant carrying the inner error.

2. The new comment says `WithAdditionalInfo` is skipped so that a transaction still being assembled "may not carry all of its input utxos" — that part is misleading: `Basic` *does* enforce `input_utxos.len() == inputs.len()`, only the entries may be `None`. Only the `additional_info` half of the sentence holds.

3. Optional: the new error field has no `#[from]`/`#[source]`, so `Error::source()` returns `None` for it. Cosmetic, since the message still interpolates.

Unrelated to this diff, but the same bug class as #2069 if you feel like folding it in: `wallet-rpc-client/src/rpc_client/client_impl.rs:1386` does `decode_all(...).expect("valid partially signed tx")` after `hex::decode(...).expect("valid hex")`, so a malformed response from a remote `wallet-rpc` server panics the CLI. This PR hardens the server-side ingress and leaves the client side unguarded.

One process note: this branch is based on `6dce99e4`, which predates the CI fix from #2102 (Ledger app repo revision, pinned `ledger-app-dev-tools`). Could you rebase onto current master so CI has a chance of passing once the workflow run is approved? The same applies to #2094, #2095 and #2096 — #2097–#2100 are already based on a commit that includes it.

oliv3rdrt:
@erubboli Rebased all four onto master.

I folded the error into the existing `InvalidRawTransaction` variant rather than adding a second one, since the old unit variant would have been left unused. Comment is fixed too, you were right that `Basic` does check the input utxo count and only the entries can be None. Added the `#[source]` as well.

Left the client side expects for a follow-up against #2069. There are two of them there rather than one.

nullPointerEnjoyer:

**Overall:** This is a correct, well-motivated robustness fix. `decode_from_untagged_bytes` previously accepted structurally inconsistent `PartiallySignedTransaction`s that `new` would never produce, and downstream code assumes the counts match the input count — e.g. `ptx.input_utxos()[input_num]` in `wallet/wallet-controller/src/lib.rs:1280` and `wallet/src/signer/software_signer/mod.rs:325` — so malformed input to the externally reachable `inspect_transaction`/`sign_raw_transaction` RPCs could panic the wallet. I verified the chosen `Basic` check level is consistent with release-mode wallet construction (`new_for_wallet` uses `Basic` in release, `wallet/types/src/partially_signed_transaction.rs:50`), so no legitimate transaction is rejected; the test's hand-rolled encoding (discriminant `64` per `#[codec(index = 64)]`, field order matching `PartiallySignedTransactionV1`'s declaration, error order matching the first `ensure!` in `v1.rs:90`) is accurate; all three `InvalidRawTransaction` sites are updated consistently with no other constructors or matches in the workspace; and the new error variant breaks no exhaustive matches. I also checked the other decode sites (`wallet-rpc-client/client_impl.rs:1386`, wasm `decode_partially_signed_transaction_to_js`) — neither has a panic path, so not extending the check there is fine. The `map_err(RpcError::InvalidRawTransaction)` refactorings are idiomatic. Only minor test-coverage notes below; no correctness, security, or consensus concerns.

**Findings:**

- **low** · test coverage · `wallet/types/src/generic_transaction.rs:157` — The new test only exercises the `InvalidWitnessCount` case; the other three `Basic` count checks now on the decode path (`InvalidInputUtxosCount`, `InvalidDestinationsCount`, `InvalidHtlcSecretsCount`) are untested. Consider parameterizing the `#[rstest]` over which field vector is encoded with a mismatched length, so all four rejections are covered.
- **low** · maintainability · `wallet/types/src/generic_transaction.rs:163` — The hand-assembled bytes depend on `PartiallySignedTransactionV1`'s field order and the `64` codec index, which are private implementation details; a reorder would only surface as a confusing test failure (or, in theory, an accidentally-different error variant). A brief comment already mitigates this; optionally, asserting `ptx.encode()` of a valid ptx starts with `64u8` (the existing test at line 112 asserts only inequality with the tx byte) would pin the discriminant more explicitly.

PR #2094, opened 2026-07-11
Avoid selecting redundant UTXOs during coin selection

### LINKED ISSUE #1923: UTXO selection algorithms selects redundant UTXOs
This is especially noticeable/reproducible in the `config-broadcast no` mode (introduced in https://github.com/mintlayer/mintlayer-core/pull/1904), in which the new transaction's summary is printed to the console without it being broadcast to the network.

E.g. when trying to create a new transaction I can see that my biggest UTXO (100k TML) is selected, but it's also accompanied by smaller UTXOs, which should not be needed.

There seem to be multiple reasons for this:
1) The selection algirithm to blame seems to be `srd`; in `select_coins_srd` we randomly select UTXOs, populating a heap, and in the end do:
    ```
    if selected_eff_value >= target_value {
        // Result found, add it.
        while let Some(group) = heap.pop() {
            result.add_input(&group.0, pay_fees)?;
        }
        return Ok(result);
    }
    ```
    So the reason for the smaller UTXO being initially selected is that the large UTXO was selected later. But then the above-mentioned code shouldn't put the entire heap into the result; only the top items necessary to meet the target should be put there.

2) In `select_random_coins`, several algoriths (like `select_coins_srd`) are executed and the best result is taken. And the best result selection is currently implemented as
    ```
    results.into_iter().min_by_key(|res| res.waste)
    ```
    In my case, the `waste` value was the same for all algorithms; and since `select_coins_srd` is executed first, it was always winning.
    So:
    * probably the best result should be selected first by `waste`, but if it's equal then the UTXO count should be taken into account as well.
    * perhaps `waste` calculation can be improved.

oliv3rdrt:
## Problem

The wallet's coin selection could pick UTXOs that were not needed. It is easy to see with `config-broadcast no`, where a transaction ends up spending the largest UTXO together with several smaller ones that are not required.

There were two causes:

1. In `select_coins_srd`, once the randomly drawn effective value reached the target, the whole heap was added to the result. If a small UTXO was drawn before a larger one that already covered the target, the small one was included even though it was redundant.

2. In `select_random_coins`, results from the different algorithms were compared by `waste` only. When the waste was equal (which is common), the first result (`srd`) always won, even if another algorithm selected fewer inputs.

## Fix

1. `select_coins_srd` now adds only the most valuable groups needed to reach the target (the heap is already ordered by effective value), leaving out the redundant smaller ones.

2. `select_random_coins` now breaks waste ties by the number of selected inputs, preferring the result with fewer inputs.

## Tests

Added `test_srd_solver_omits_redundant_utxos`: it gives SRD one large UTXO that already covers the target plus several small ones, and asserts only the large UTXO is selected. The existing selection tests still pass.

Closes #1923

### DIFF
--- a/wallet/src/account/utxo_selector/mod.rs
+++ b/wallet/src/account/utxo_selector/mod.rs
@@ -253,8 +253,17 @@ fn select_coins_srd(
 
         // Now check if we are above the target
         if selected_eff_value >= target_value {
-            // Result found, add it.
-            while let Some(group) = heap.pop() {
+            // Result found. The heap is ordered by effective value, so add only the most
+            // valuable groups needed to reach the target. Groups that were drawn earlier but
+            // turned out to be redundant (because a larger one was drawn afterwards) are left
+            // out instead of being included in the selection.
+            let mut added_eff_value = Amount::ZERO;
+            while added_eff_value < target_value {
+                let group = heap
+                    .pop()
+                    .expect("heap total reached the target, so it can't be emptied first");
+                added_eff_value = (added_eff_value + group.0.get_effective_value(pay_fees))
+                    .ok_or(UtxoSelectorError::AmountArithmeticError)?;
                 result.add_input(&group.0, pay_fees)?;
             }
             return Ok(result);
@@ -799,7 +808,9 @@ fn select_random_coins(
 
     results
         .into_iter()
-        .min_by_key(|res| res.waste)
+        // Pick the least wasteful result; when several results have the same waste, prefer the
+        // one that selects fewer inputs so we don't end up with redundant UTXOs.
+        .min_by_key(|res| (res.waste, res.num_selected_inputs()))
         .ok_or_else(|| errors.pop().unwrap_or(UtxoSelectorError::NoSolutionFound))
 }
 
--- a/wallet/src/account/utxo_selector/tests.rs
+++ b/wallet/src/account/utxo_selector/tests.rs
@@ -450,3 +450,36 @@ fn test_srd_solver_max_weight(#[case] seed: Seed) {
 
     assert_eq!(result.unwrap_err(), UtxoSelectorError::MaxWeightExceeded);
 }
+
+#[rstest]
+#[trace]
+#[case(Seed::from_entropy())]
+fn test_srd_solver_omits_redundant_utxos(#[case] seed: Seed) {
+    let mut rng = make_seedable_rng(seed);
+
+    // One large UTXO that on its own covers the target, plus several small UTXOs that
+    // together can't reach it. Whatever order SRD draws them in, the large UTXO has to be
+    // drawn to reach the target, and once it is none of the small ones are needed.
+    let target_value = Amount::from_atoms(50);
+    let mut groups = vec![];
+    add_output(Amount::from_atoms(100), &mut groups);
+    for _ in 0..10 {
+        add_output(Amount::from_atoms(1), &mut groups);
+    }
+
+    let cost_of_change = Amount::ZERO;
+    let max_weight = 100;
+    let result = select_coins_srd(
+        &groups,
+        target_value,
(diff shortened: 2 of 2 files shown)


nullPointerEnjoyer:

**Overall:** The intent is sound and the mechanics of both changes are correct in the common path: the heap is a max-heap on effective value (`OutputGroupOrd::cmp`, wallet/src/account/utxo_selector/mod.rs:198), so popping greedily until the target is reached yields a valid, smaller selection, and the `(waste, num_selected_inputs())` tie-break (both `Ord`, verified) is a sensible wallet-level improvement. However, the new `expect` asserts an invariant (`heap total >= target`) that the *pre-existing* max-weight trimming loop in the same function can violate, making the panic genuinely reachable; the new test avoids exactly that path.

**Findings:**

- **high** · correctness/panic (wallet DoS, not consensus) · `wallet/src/account/utxo_selector/mod.rs:259` — The `expect("heap total reached the target, so it can't be emptied first")` relies on `selected_eff_value` equaling the sum of heap contents, but the unchanged max-weight trim loop (mod.rs:242-252) breaks that invariant: it pops a group *before* checking `weight <= max_weight`, so when the check passes the just-popped group is silently discarded without subtracting its value/weight from `selected_eff_value`/`weight`. Afterwards `selected_eff_value` overcounts the heap, so the `selected_eff_value >= target_value` branch (mod.rs:255) can be entered when the real heap total is below target, and the new loop drains the heap and panics. Concrete reachable case (independently traced and confirmed during this local run): groups `(value, weight)` = [(10,1), (1,1), (1,1), (10,1)], `max_weight = 3`, `target = 12`, draw order as listed. After all four draws: heap = {10a, 1, 1, 10b}, `selected_eff_value` = 22, weight = 4 > 3, so the trim loop pops 10b (subtracting: selected = 12, weight = 3), then pops 10a and hits `weight <= max_weight` → `break` **without subtracting**: 10a is gone from the heap but still counted, so heap total = 2 vs `selected_eff_value` = 12. Since 12 >= target, the new loop pops 1, 1 → `added_eff_value` = 2 < 12, heap empty → `expect` panics. (The old code also misbehaved here — it returned an underfunded result that panicked later in `compute_and_set_waste` via `(effective_value - target).expect(...)` — but this diff adds a direct panic path built on the false invariant.) Fix: repair the trim loop to account for every discarded group (check `weight <= max_weight` before popping, or subtract after every pop), and/or replace the `expect` with `ok_or(UtxoSelectorError::NoSolutionFound)` for defense in depth. Note also the trim loop's own comment ("remove the least valuable inputs") is wrong — `BinaryHeap::pop` removes the *most* valuable group — pre-existing but directly relevant to this invariant.

- **low** · test coverage · `wallet/src/account/utxo_selector/tests.rs:455` — The new test deliberately exercises only the happy path (`max_weight = 100` with 11 weight-1 groups, so the trim loop never runs). There is no test for the interaction between the new target-trimming pop loop and the max-weight trim path, which is precisely where the high finding above lives; add a case with a tight `max_weight` (e.g. the scenario described above) once the trim loop is fixed, asserting either a valid selection or a clean `MaxWeightExceeded`/`NoSolutionFound` error instead of a panic.

The change in `select_random_coins` (`min_by_key` on `(res.waste, res.num_selected_inputs())`, mod.rs:811) is correct as written: `SignedAmount` is `Copy + Ord` (common/src/primitives/amount/signed.rs:28), the tuple key is valid, and preferring fewer inputs on equal waste is a reasonable fee-reducing heuristic with no consensus impact.

---
*Verification note: the high finding above was independently reproduced by hand-tracing `select_coins_srd` on this PR's branch before posting (the counterexample above is the corrected trace — the trim loop discards the most-valuable group, not a small one). The other findings were checked against the diff. Everything else in this PR (the `min_by_key` tie-break) verified clean.*

Issue #2066, opened 2026-05-22
Wallet should take cluster limits into account when creating a transaction


ImplOfAnImpl:
Since https://github.com/mintlayer/mintlayer-core/pull/2063, the mempool has limits for how many txs are allowed to form a cluster and how big in bytes that cluster can be. Currently the wallet doesn't take these limits into account, so it's possible for it to create a tx that will be rejected by the mempool.

At the very least, the wallet should:
* Prefer confirmed utxos during utxo selection, to avoid forming clusters.
* After the tx is constructed, it should ask the mempool for the tx count and byte size of the cluster that will be formed if this tx is added (for which the corresponding methods should be exposed in node RPC) and then, ideally, try some other utxo combination.

Also note that even though currently the mempool only tracks utxo-based relations (https://github.com/mintlayer/mintlayer-core/issues/2065), the wallet should be prepared that at some point this won't be the case anymore and take into account non-utxo relations too.

There are some other possible improvements in utxo selection, see the TODO near `Account::select_inputs_for_send_request`.

P.S. if a general utxo selection revamp is attempted, then this - https://github.com/mintlayer/mintlayer-core/issues/1923 - should probably be fixed as well.

PR #2095, opened 2026-07-11
Prefer confirmed utxos during coin selection

oliv3rdrt:
## Problem

Since #2063 the mempool limits how many transactions a cluster may contain and how big it may be. A transaction that spends unconfirmed (in-mempool) utxos joins the cluster of the transactions that produced them, so a wallet that builds on top of unconfirmed utxos can create a transaction that the mempool then rejects because the cluster grew too large.

## Change

This is the first step of #2066: prefer confirmed utxos during selection.

`select_inputs_for_send_request` now also gathers the confirmed-only utxos, and a new `select_coins_preferring_confirmed` helper tries to reach the target from the confirmed set first, falling back to the full set (which also contains the unconfirmed utxos) only when the confirmed ones are not enough. The fallback is exactly the call that was made before, so behaviour is unchanged whenever confirmed utxos can't cover the amount. The preference is only applied when the caller did not select the inputs explicitly.

## Not included here

The rest of #2066 is left for follow-ups:
- Querying the node/mempool for the resulting cluster's tx count and byte size and retrying with a different utxo combination (needs new node RPC methods).
- Handling non-utxo cluster relations once the mempool tracks them.

## Tests

Added unit tests for `select_coins_preferring_confirmed`: it uses the confirmed set when it is sufficient, falls back to the full set when it is not, and uses the full set directly when there is no preference. The existing selection tests still pass.

Addresses #2066

### DIFF
--- a/wallet/src/account/mod.rs
+++ b/wallet/src/account/mod.rs
@@ -84,7 +84,7 @@ use wallet_types::{
 
 use crate::{
     SendRequest, WalletError, WalletResult,
-    account::utxo_selector::{OutputGroup, select_coins},
+    account::utxo_selector::{OutputGroup, select_coins_preferring_confirmed},
     destination_getters::{HtlcSpendingCondition, get_tx_output_destination},
     key_chain::{AccountKeyChains, KeyChainError, VRFAccountKeyChains},
     send_request::{
@@ -112,7 +112,7 @@ pub use self::{
     utxo_selector::{CoinSelectionAlgo, UtxoSelectorError},
 };
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 pub struct CurrentFeeRate {
     pub current_fee_rate: FeeRate,
     pub consolidate_fee_rate: FeeRate,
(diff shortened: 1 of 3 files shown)


nullPointerEnjoyer:

**Overall:** Sound change implementing the issue #2066 mitigation already sketched in the TODO at `wallet/src/account/mod.rs:217-229`. I verified the key correctness properties: the confirmed-only pool is a strict subset of the full pool (`output_cache`'s `is_in_state` is a per-tx bitflag check over `UtxoState::Confirmed | InMempool | Inactive`), the two pools are partitioned by currency identically so no outpoint can be double-selected, the zero-target short-circuit in `select_coins` keeps `Some(vec![])`-pool behavior identical to before, the fallback surfaces the (more informative) full-pool error, and the new `Copy` derive on `CurrentFeeRate` is genuinely required because `utxo_output_groups_by_currency` takes it by value and it is now consumed twice. The three new unit tests are deterministic despite `Randomize` (single-coin pools, or an assertion that holds for any reaching subset). No correctness, security, or consensus issues found; findings below are minor.

**Findings:**

- **medium** · test-coverage · `wallet/src/account/mod.rs:280-306` — All new tests cover only the `select_coins_preferring_confirmed` unit; the actual wiring in `select_inputs_for_send_request` (building `confirmed_utxos_by_currency`, and the two per-currency `remove` sites feeding it) is untested. A regression such as passing the full pool as the "confirmed" pool, or dropping the fee-currency bucket, would pass CI. Suggest an account/wallet-level test asserting that a send with both a sufficient confirmed utxo and an in-mempool utxo selects the confirmed one (and still succeeds when only in-mempool funds exist).
- **low** · design · `wallet/src/account/mod.rs:288` — Subtle behavior change in an edge case: with empty `input_utxos` and explicit `Some(CoinSelectionAlgo::UsePreselected)`, `select_all_coins` previously spent the entire (confirmed + in-mempool + inactive) pool, but now spends only confirmed utxos whenever they suffice. That combo ("use all specified inputs" with nothing specified) is already ill-defined, and the new behavior arguably matches the PR's goal, but it's undocumented; worth a note in the function's comment or a debug assertion rejecting that combination.
- **low** · performance · `wallet/src/account/mod.rs:281-306` — The auto-selection branch now scans the whole output cache twice (`get_utxos` ×2) and builds output groups twice (`utxo_output_groups_by_currency` ×2, including per-output fee computation). Acceptable as the simple correct approach (state info is lost after grouping, so filtering one grouped set isn't trivial), but consider profiling for wallets with large UTXO sets.

PR #2096, opened 2026-07-13
Add an IsEmpty derive macro and use it in pos-accounting

### LINKED ISSUE #669: Introduce IsEmpty derive trait
Consider following example:
``` rust
struct Foo {
    a: BTreeMap<u32, u32>,
    b: BTreeMap<u32, u32>,
}

impl Foo {
    fn is_empty(&self) -> bool {
        self.a.is_empty() && self.b.is_empty()
    }
}
```
The problem with this code is that if an additional field `c` is added it is easy to forget to update `is_empty` method.

One solution would be to write a procedural macro that would automatically iterate over fields and generate `is_empty` method.

Potential problem is that it should work for types that don't have 'is_empty' method like POD types. To circumvent that an attribute can be added that requires a type to implement the emptiness property.

oliv3rdrt:
## Problem

`PoSAccountingData::is_empty` is written by hand as the conjunction of its five fields' `is_empty` calls, and carries a `// TODO: avoid manual implementation (mintlayer/mintlayer-core#669)`. As the issue notes, adding a field and forgetting to update such a method is an easy mistake.

## Change

Add an `IsEmpty` trait and a derive macro for it, as a new `is-empty` / `is-empty-derive` crate pair that mirrors the existing `typename` / `typename-derive` setup. The derive generates `is_empty` as the AND of each field's own `is_empty`, so a struct is empty exactly when all of its fields are empty (a field-less struct is always empty). Every field type therefore needs an `is_empty` method: the standard collections, `String`, or another `IsEmpty` type.

Apply the derive to `PoSAccountingData`, removing the manual implementation and the TODO. `pos-accounting` re-exports the trait so the existing callers only need it in scope.

## Tests

The `is-empty` crate has unit tests covering named-field, tuple and unit structs, and the empty and non-empty cases. The existing `pos-accounting` and `chainstate` tests still pass.

Closes #669

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4347,6 +4347,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "is-empty"
+version = "1.4.0"
+dependencies = [
+ "is-empty-derive",
+]
+
+[[package]]
+name = "is-empty-derive"
+version = "1.4.0"
+dependencies = [
+ "quote",
+ "syn 2.0.114",
+]
+
+[[package]]
 name = "is-terminal"
 version = "0.4.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6635,6 +6650,7 @@ dependencies = [
  "accounting",
  "common",
  "crypto",
+ "is-empty",
  "parity-scale-codec",
  "randomness",
  "rstest",
--- a/chainstate/test-suite/src/tests/chainstate_accounting_storage_tests.rs
+++ b/chainstate/test-suite/src/tests/chainstate_accounting_storage_tests.rs
@@ -32,7 +32,7 @@ use common::{
     primitives::{Amount, Idable, per_thousand::PerThousand},
 };
 use crypto::vrf::{VRFKeyKind, VRFPrivateKey};
-use pos_accounting::PoolData;
+use pos_accounting::{IsEmpty, PoolData};
 use randomness::{CryptoRng, RngExt as _};
 use rstest::rstest;
 use test_utils::random::{Seed, make_seedable_rng};
--- a/pos-accounting/Cargo.toml
+++ b/pos-accounting/Cargo.toml
@@ -9,6 +9,7 @@ rust-version.workspace = true
 accounting = { path = "../accounting" }
 common = { path = "../common" }
 crypto = { path = "../crypto" }
+is-empty = { path = "../utils/is-empty" }
 randomness = { path = "../randomness" }
 serialization = { path = "../serialization" }
 typename = { path = "../utils/typename" }
--- a/pos-accounting/src/data.rs
+++ b/pos-accounting/src/data.rs
@@ -19,11 +19,12 @@ use common::{
     chain::{DelegationId, PoolId},
     primitives::Amount,
 };
+use is_empty::IsEmpty;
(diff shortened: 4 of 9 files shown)


erubboli:
Nice cleanup. I reproduced the derive in an isolated workspace and checked the generated `PoSAccountingData::is_empty` is equivalent to the manual one it replaces (same five fields, same order), and that the inherent→trait move breaks nothing — the only three call sites are all in `chainstate_accounting_storage_tests.rs`, which the PR already updates. The missing `utils/is-empty*` entries in the workspace `members` list are fine too, consistent with `utils/typename` and auto-included as path deps.

One thing I'd like fixed before merge:

**The generated impl uses an unqualified trait path** (`is-empty-derive/src/lib.rs:70`): `quote! { impl #impl_generics IsEmpty for #ident ... }` resolves `IsEmpty` in the *caller's* scope. Two consequences, both of which I reproduced:

- `#[derive(is_empty::IsEmpty)]` without a separate `use is_empty::IsEmpty;` fails with `error[E0405]: cannot find trait IsEmpty in this scope`. Path-qualified derives are a normal pattern, so this will bite someone.
- Worse, it can be silently wrong: with another trait named `IsEmpty` in scope, the derive implements *that* one, and the type does not satisfy `is_empty::IsEmpty` (rustc even says "implements similarly named trait `IsEmpty`, but not `is_empty::IsEmpty`").

Emitting `::is_empty::IsEmpty` fixes both. Note the per-field call at line 50 must stay unqualified, since `Vec`/`BTreeMap` use an inherent `is_empty` — worth a line in the doc comment saying a nested `IsEmpty`-typed field still needs the trait in scope.

Two smaller ones:

1. `is-empty-derive/src/lib.rs:50` — the doc says "every field type must have one", implying a compile error protects you, but the check is plain method lookup. `struct WithArray { a: [u32; 4] }` compiles (via slice deref) and `[T; N]` with `N > 0` returns `false` unconditionally, so adding a fixed-size array field silently makes `is_empty()` permanently false — the exact silent-bug class #669 is about. Same for types whose `is_empty` means something unrelated (`NetUpgrades`, `Services`, `Script`).

2. `pos-accounting/src/lib.rs:40` — `pub use is_empty::IsEmpty;` leaks a foreign trait (and, since the name covers both namespaces, the derive macro) into an unrelated crate's public API, and ties its availability to pos-accounting continuing to use it. I'd have `chainstate-test-suite` depend on `is-empty` directly. It also makes the shadowing in the finding above more likely, by putting an `IsEmpty` name into `pos_accounting`'s namespace.

Also, could you rebase onto current master? This branch is based on `6dce99e4`, which predates the CI fix from #2102.

oliv3rdrt:
@erubboli All three done!

Worth flagging: emitting ::is_empty::IsEmpty breaks the is-empty crate's own tests, since that path doesn't resolve inside the crate itself. Added extern crate self as is_empty; for that. There's now a test that derives through a qualified path with the trait not imported and a decoy IsEmpty in scope, which fails without the fix.

Dropped the re-export and pointed chainstate-test-suite at is-empty directly. Doc comment now says it's plain method lookup and calls out the fixed size array case.

On the rebase, the branch was already on master with #2102 in it, but it had drifted a couple of commits so it's up to date now.

Issue #762, opened 2023-03-14
Fix leaky tests detected by nextest


TheQuantumPhysicist:
When running unit tests with `nextest`, it detects [leaky tests](https://nexte.st/book/leaky-tests.html), which are defined as tests that don't clean-up properly when they finish. We should investigate this.

Note: For some reason, these are only detected on Windows. For me, I saw them when running the tests in git bash. This isn't found/detected by Linux or Mac.

discosultan:
Also experienced similar warnings on Windows with Git Bash. The warnings are intermittent and not trivially reproducible. Here are a few examples:
```
LEAK [   0.138s] common chain::transaction::signature::tests::mutate_none::case_1
LEAK [   0.187s] crypto hash::tests::test_hash_sha3_512
LEAK [   0.184s] crypto hash::tests::test_ripemd160
LEAK [   0.184s] crypto hash::tests::test_ripemd160_stream
```

Issue #417, opened 2022-09-14
Add the possibility to pass prefix by a reference in storage key prefix iteration routines


iljakuklic:
In the key prefix iteration introduced in #411, the prefix is currently passed by value but it is only read. In general, it would be more appropriate to pass it by reference in some cases. Support for that could be added, e.g. by using the `Borrow` trait to support both.

Some rationale and discussion [here](https://github.com/mintlayer/mintlayer-core/pull/411#discussion_r969370720).

PR #2100, opened 2026-07-28
Pass the prefix by reference in storage key prefix iteration

oliv3rdrt:
The prefix and key passed to `prefix_iter` and `greater_equal_iter` are only read, but they were taken by value, so callers had to hand over an owned `Vec` even when they already had a slice. In the locking adaptor this meant cloning the prefix twice on every call.

This changes the `ReadOps` methods, and the matching helpers in the database module, to take `&[u8]`, and updates the backends (in-memory, lmdb, sqlite, failing, and the locking adaptor) and their callers.

What it removes in practice:

- the two `prefix.clone()` calls and the `key.clone()` in the locking adaptor
- `Vec::new()` and `Data::new()` allocations at call sites that only needed an empty prefix
- a couple of `clone()` calls in the backend test suite

One thing worth a look during review: since the crate is on edition 2024, the returned `impl Iterator` would otherwise capture the prefix lifetime as well as the transaction's. I used the precise capturing syntax (`use<..>`) on the trait methods and the impls so the iterator only borrows from the transaction, which is what the previous signature effectively guaranteed. If you would rather see it done another way I am happy to change it.

Checked with `cargo check --workspace --all-targets`, `cargo fmt`, and the storage tests (`storage`, `storage-core`, `storage-inmemory`, `storage-lmdb`, `storage-sqlite`) plus `chainstate-storage` and `utxo`, all passing. Note that a few of the lmdb property tests fail intermittently when run in parallel both with and without this change, which looks like the flakiness tracked in #762, they pass with `--test-threads=1`.

Addresses #417.


oliv3rdrt:
@erubboli can I be assigned this?

PR #2114, opened 2026-09-15
fix: create Sqlite databases with owner-only permissions (CWE-732)

erubboli:
Fixes a High-severity local vulnerability (CWE-732, suggested CVSS 3.1: 7.8) in wallet database creation.

## Issue

Wallet databases contain highly sensitive material: the root extended private key, the root VRF private key, and — when `store-seed-phrase` is enabled — the complete BIP-39 mnemonic and optional passphrase, in plaintext. With the common `0022` umask:

- missing parent directories were created as `0755` (`create_dir_all`)
- the database file was created by Sqlite as `0644`

Any other local unprivileged user could copy a traversable database and recover full signing authority without the victim's RPC credentials.

## Fix (`storage/sqlite`, Unix)

- **Directories:** missing parent directories are created `0700` — this also shields the auxiliary Sqlite files (rollback journal, WAL, shared-memory, temporary files) under the same private directory boundary. Permissions of pre-existing directories are left untouched, since they may be shared with unrelated data.
- **Database file:** created **atomically** with `0600` (pre-created via `OpenOptions::create_new().mode(0o600)` before Sqlite opens it), so it is never observable with looser permissions, even momentarily.
- **Repair:** opening an existing database created by a pre-fix version tightens the file to `0600`.

Non-Unix platforms keep the previous behavior.

## Tests

Two new Unix-only tests in `storage/sqlite`:
- newly created DB → directory `0700`, file `0600`
- existing DB with `0644` file → repaired to `0600` on open

## Verification

- `cargo test -p storage-sqlite`: 27/27 ✓, `cargo test -p wallet-controller`: 24/24 ✓, `cargo test -p wallet`: 114/114 ✓
- full `./do_checks.sh` green under the CI toolchain (1.92.0)


erubboli:
Addressed the OpenCodeReview findings ([run](https://github.com/mintlayer/mintlayer-core/pull/2114#issuecomment-5682142174)) in bb948a7af:

**Adopted:**
- **Race-free directory ownership** (comments 3+4): `ensure_private_directory` now attempts an atomic `create_dir` first and derives ownership from its result instead of a racy `!dir.exists()` check — this also removes the case where a directory created concurrently between the check and the chmod would have been tightened against the documented intent. Chose `NotFound => create_dir_all` + `AlreadyExists => leave untouched` over the suggested catch-all fallback so the ownership conclusion stays sound (`NotFound` proves the leaf didn't exist).
- **New test** `pre_existing_directory_permissions_are_left_untouched`: creates a `0755` directory, opens the DB inside, and asserts the directory stays `0755` while the file becomes `0600` — pinning the intentional "pre-existing dirs are not modified" behavior (comment 2).

**Declined:**
- **Holding the file handle from `create_private_file`** (comment 1): an open fd doesn't prevent another local process from `unlink`+recreating the path before SQLite opens it by path, so the suggested change doesn't actually narrow the TOCTOU window — it's security theater. The meaningful mitigations (open-by-fd, unsupported by rusqlite; or post-open inode verification) are disproportionate for a wallet directory under the user's own `HOME`.

anyxem (approved):

erubboli:
Addressed the OpenCodeReview findings ([run](https://github.com/mintlayer/mintlayer-core/pull/2114#issuecomment-35059174164-comment) rounds 1–3) in cdad745a5:

**Adopted:**
- **Repair-on-open policy** (rounds 1+2): the file permissions are now repaired to `0600` only when the file is actually exposed to group/other users (`mode & 0o077 != 0`); already-private modes are left untouched, and the policy is documented on `create_private_file` and in the module security notes.
- **Umask-independent creation** (round 2): `0600` is now enforced with an explicit `set_permissions` after `create_new`, since the `OpenOptions::mode` is masked by the umask.
- **Non-regular paths** (round 3): `symlink_metadata` + `is_file` check rejects directories and refuses to follow symlinks at the database path with a clear error instead of a confusing one from Sqlite later.
- **Intermediate directories** (round 3): `ensure_private_directory` now walks the components and creates each missing one with `create_dir` + immediate `set_permissions(0700)`, instead of `create_dir_all` leaving intermediates with umask-derived permissions. This also removes the chmod-after-create window (TOCTOU) for every created component; note std has no atomic mode-at-mkdir and adding a libc dep for this was not warranted.
- **cfg duplication** (round 3): extracted a cfg-gated `ensure_parent_dir` helper; `Backend::open` is now single-line and platform-neutral.

**Documented (declined as code changes):**
- **Sqlite sidecar files** (round 2): journal/WAL/SHM permissions are not controllable from outside Sqlite; documented that they are only protected by the containing directory's permissions (which is why the directory we create is 0700).
- **Non-Unix gap** (round 1): documented in the module security notes and at the `create_private_file` call site; equivalent ACL hardening would be a separate effort.

**New tests:** `nested_directories_are_created_private`, `symlinked_database_path_is_rejected`, `directory_at_database_path_is_rejected`, `already_private_permissions_are_left_untouched`.

PR #2117, opened 2026-09-16
feat(api-server): real-time SSE event stream for block explorers

nullPointerEnjoyer:
## Summary

Adds a real-time event stream to the api-server so block explorers can display transactions and blocks live. New endpoint: `GET /api/v2/stream` (Server-Sent Events), delivering three event kinds as *named* SSE events:

| event | when | payload |
|---|---|---|
| `tx_seen` | a transaction reaches the node's mempool (pre-indexing) | `{tx_id, origin: local\|remote}` |
| `block` | a block has been fully indexed into the api-server database | `{block_id, height, timestamp, tx_ids}` |
| `reorg` | previously indexed blocks are disconnected | `{common_ancestor_height, removed_block_ids, new_tip_height}` |

Payloads are kept small; clients hydrate details through the existing REST endpoints.

## Design

**Block/reorg events — scanner, transactionally consistent.** `BlockchainState::scan_blocks` appends events to a new `ml.emitted_events` table *inside the same RW transaction* that indexes the blocks (reorg detection: local best height > common ancestor, with removed block ids captured before the disconnect marks them). After the appends, the scanner issues `pg_notify('mintlayer_events', <last id>)` inside the transaction, so Postgres delivers the wakeup only on commit: **once a `block` event is delivered, the referenced block is immediately queryable via REST** (asserted by tests).

**tx_seen events — web server.** The web server bridges the node's mempool events from the existing `NodeRpcClient` WebSocket into `TxSeen` events (only `successful: true` transactions; subscription re-established on connection loss).

**Web server event pump.** A background pump wakes on the Postgres notification (dedicated LISTEN connection — with automatic reconnection bounded by the poll interval, since Postgres only delivers notifications on the listening connection) and forwards events into a `tokio::sync::broadcast` channel. Periodic polling (default 30 s) is the safety net for missed notifications; the backlog is drained in bounded batches (1000/batch) and the cursor (`last_seen_id`) only ever moves forward, so a web-server restart cannot produce duplicate deliveries. Events older than the last 10 000 are pruned by the scanner (the stream has no replay, so they are never re-read).

**Endpoint.** Named events for `EventSource.addEventListener`, optional `?types=block,reorg` filter (400 on invalid values), `: keepalive` comments (default 30 s), spec-compliant `retry:` hint as the first frame, `x-accel-buffering: no` for reverse proxies, and a `lag` advisory (`{"skipped": n}`) when a client exceeds the broadcast capacity. No auth, consistent with the other v2 GET endpoints. No replay/Last-Event-ID: clients recover missed events via REST.

## Operational notes

- **Storage version bumped 25 → 26** (new `ml.emitted_events` table). Per the existing convention this triggers the usual full database re-initialization on upgrade (full resync). Required regardless: without the new table the scanner would fail on insert.
- Three new web-server options with working defaults (no config change needed for existing deployments): `--stream-events-broadcast-capacity` (1024), `--stream-events-poll-interval-secs` (30), `--stream-events-keepalive-interval-secs` (30).
- In-memory storage backend treats streaming as a no-op (documented in the trait defaults); its test suite stays green.

## Testing

- `storage-test-suite` (both backends; Postgres via container): events visible only after commit, rollback discards them, ascending monotonic ids, resume-after-last-seen, no-op behavior of the in-memory backend.
- New unit tests: pump backlog/resume/error-survival, event serde roundtrip, `tx_seen` mapping (filters `successful: false` and `NewTip`), filter parsing/matching.
- Stack tests: SSE endpoint contract (content-type, framing, named events, filter, 400 on bad input, keepalive), block event ⇒ immediate `GET /v2/block/:id` consistency, and a Postgres end-to-end test driving the real scanner → `ml.emitted_events` → LISTEN/NOTIFY pump → SSE, including a forced reorg (exactly one `reorg` event with correct removed ids/heights, followed by the new fork's block events, no duplicates), and a dedup check.
- `./do_checks.sh` clean (fmt, cargo-deny, cargo-vet, clippy, codecheck); `cargo test --release` green for every touched crate (stack tests incl. containers re-run after the rebase onto current master).

## Review

Ran two rounds of security review (findings fixed: event-pump hot-spin after listener death → reconnect + sleep fallback; unbounded startup reads → batched reads; poison-row stall → skip undecodable rows; config-value panics → clamped; dead `retry` header → in-stream frame; event-name duplication → single `StreamEventType` source of truth) and a code-quality review (both blockers and warnings addressed; DRY pass on shared types/test helpers).

## Notes for reviewers

- The second commit (`fix: address clippy 1.98 lints across the workspace`) is **unrelated drive-by work**: `do_checks.sh` fails on current master with clippy 1.98's new lints; the fixes are mechanical and behavior-preserving (reviewers can skip it). Happy to split it into its own PR if preferred.
- `CURRENT_STORAGE_VERSION` bump means existing deployments resync on upgrade — called out in the README/CHANGELOG.
- Known gap: `tx_seen` is not covered end-to-end against a real node process (the repo has no harness that spins a node RPC server in tests); the bridge mapping and the SSE delivery path are covered separately.
- The container helper used by the Postgres tests (`storage-test-suite/src/podman.rs`) now falls back to `docker` when `podman` is not installed (identical CLI surface for the commands used).


nullPointerEnjoyer:
Thanks @github-actions[bot] — findings addressed in dcf705b3b:

- **pump replays stale events on restart (high)** — fixed: the pump now initializes its cursor from the most recently committed event id (`initial_last_seen_id` / `latest_stream_event_id`); the backlog is never re-broadcast to connected clients. Committed events remain available via REST.
- **pruning vs lagging pump (medium)** — the retention cap is inherent to the no-replay design; pruning now **logs a warning with the deleted count** for observability. The pump always drains within one wakeup, so losing unread events requires a sustained >10k-event lag.
- **mempool bridge retry loop (medium)** — now uses exponential backoff capped at 60 s, resetting on a successful subscription.
- **dropped JoinHandles in main (medium)** — both background tasks are now supervised: termination logs a CRITICAL error.
- **collector JoinHandle / blocking container startup / timing-based no-duplicates window (test, medium)** — all three fixed in the postgres stream test (collector panics now propagate with their message; startup runs via `spawn_blocking`; windows widened to 3 s).

Intentionally not changed:
- `kind` column written but not read back: kept deliberately for server-side debugging/ops (`GROUP BY kind`) and future server-side filtering.
- listener driver JoinHandle not polled: a driver panic closes the wakeup channel, which the pump already treats as a dead listener and reconnects — no additional supervision needed (documented in code).
- `notify_channel_name_is_stable` test: it intentionally pins the wire value of the NOTIFY channel name against accidental change; the deeper behavior is covered by the container-based suites.
- `stream_events_append_and_read` accepting empty lists: intentional — the in-memory reference backend is a documented no-op for streaming; the strict assertions live in the Postgres suites.

erubboli (approved):

